PR #5467, opened 2022-02-24
[operator] Update ClusterServiceVersion to add missing configurations

### LINKED ISSUE #5466: [operator] ClusterServiceVersion is missing minKubeVersion and nodeSelector
**Describe the bug**
- The ClusterServiceVersion lacks `minKubeVersion`
- There isn't `nodeSelector: {kubernetes.io/os": "linux"}` for OpenShift

**To Reproduce**
1. Deploy Loki Operator.
2. Check the CSV yaml file in the cluster.

**Expected behavior**
nodeSelector and minKubeVersion is present in CSV.

sasagarw:
- Add `minKubeVersion` to csv.
- Add `nodeSelector` patch to overlay/openshift.

Fixes #5466 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operator/config/overlays/openshift/manager_node_selector_patch.yaml
+++ b/operator/config/overlays/openshift/manager_node_selector_patch.yaml
@@ -1,0 +1,9 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+spec:
+  template:
+    spec:
+      nodeSelector:
+        kubernetes.io/os: linux
--- a/operator/config/overlays/openshift/manager_node_selector_patch.yaml
+++ b/operator/config/overlays/openshift/manager_node_selector_patch.yaml
@@ -1,9 +1,0 @@
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  name: controller-manager
-spec:
-  template:
-    spec:
-      nodeSelector:
-        kubernetes.io/os: linux
--- a/operator/bundle/manifests/loki-operator.clusterserviceversion.yaml
+++ b/operator/bundle/manifests/loki-operator.clusterserviceversion.yaml
@@ -645,6 +645,8 @@ spec:
                 volumeMounts:
                 - mountPath: /var/run/secrets/serving-cert
                   name: loki-operator-metrics-cert
+              nodeSelector:
+                kubernetes.io/os: linux
               terminationGracePeriodSeconds: 10
               volumes:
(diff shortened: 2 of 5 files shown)


operator/config/overlays/openshift/manager_node_selector_patch.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+spec:
+  template:
+    spec:
+      nodeSelector:
+        kubernetes.io/os: linux
```
xperimental, line 9: Does it make sense to add the `nodeSelector` to the base deployment instead of just adding it for OpenShift? This should reduce issues when running on "vanilla Kubernetes" as well when there are non-linux nodes in the cluster.

xperimental (commented):
Just one question.

operator/config/overlays/openshift/manager_node_selector_patch.yaml
sasagarw: It makes sense but we won't be able to control that. What I mean is if we define `nodeSelector` to the base deployment then we are putting restriction to the users that our loki operator works only with linux which I don't think is the case? Maybe in future we can extend this through a flag or something else? But I guess it is out of scope for this PR right now. WDYT?

xperimental: The way the build is currently set up there are only images for linux and my guess is that nobody has yet tested the operator on any other operating system yet, so I would assume that it only runs on linux.

I have no experience with mixed-OS clusters but I would assume that every deployment needs a nodeSelector on which operating system it is able to run, otherwise the scheduler might place pods on incompatible nodes where the software is then not able to start causing unnecessary errors and, in the worst case, needing manual intervention. So I wouldn't think of this as a restriction but actually a help for normal users.

People that build their own image for running on a different operating system are probably experienced enough to also change the nodeSelector then.

sasagarw: So should I update this into base deployment?
cc @periklis

periklis: I agree here with @xperimental. As long as we officially support linux only for now, we can keep the node selector on the base deployment.

periklis (approved):

PR #5468, opened 2022-02-24
Stream query responses from boltdb index client

sandeepsukhani:
Previously, we sent just one row per batch in the index query response callback from the boltdb index client.
This, in turn, makes us send just one row at a time per grpc response from the index gateway server.

This PR improves it by streaming the response from the boltdb index client, which would, in turn, lets us send back a larger batch in grpc response from the index gateway server.

Here is the benchmark comparing the performance difference:
```
benchmark                                     old ns/op      new ns/op      delta
Benchmark_QueriesMatchingLargeNumOfRows-8     2027105841     1298895890     -35.92%

benchmark                                     old allocs     new allocs     delta
Benchmark_QueriesMatchingLargeNumOfRows-8     29321207       10141758       -65.41%

benchmark                                     old bytes      new bytes     delta
Benchmark_QueriesMatchingLargeNumOfRows-8     1032313592     507791808     -50.81%
```

This PR also adds the benchmarks to the repository.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/chunk/local/boltdb_index_client.go
+++ b/pkg/storage/chunk/local/boltdb_index_client.go
@@ -268,7 +268,15 @@ func (b *BoltIndexClient) QueryWithCursor(_ context.Context, c *bbolt.Cursor, qu
 
 	rowPrefix := []byte(query.HashValue + separator)
 
-	var batch boltReadBatch
+	// sync.WaitGroup is needed to wait for the caller to finish processing all the index entries being streamed
+	wg := sync.WaitGroup{}
+	batch := newReadBatch()
+	defer func() {
+		batch.done()
+		wg.Wait()
+	}()
+
+	callbackDone := false
 
 	for k, v := c.Seek(start); k != nil; k, v = c.Next() {
 		if !bytes.HasPrefix(k, rowPrefix) {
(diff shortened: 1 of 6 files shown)


pkg/storage/stores/shipper/testutil/testutil.go
```diff
@@ -20,7 +20,7 @@ import (
 	chunk_util "github.com/grafana/loki/pkg/storage/chunk/util"
 )
 
-func AddRecordsToDB(t *testing.T, path string, dbClient *local.BoltIndexClient, start, numRecords int, bucketName []byte) {
+func AddRecordsToDB(t testing.TB, path string, dbClient *local.BoltIndexClient, start, numRecords int, bucketName []byte) {
```
kavirajk, line 23: do we need `testing.TB` here?

cyriltovena (approved):
LGTM

pkg/storage/chunk/local/boltdb_index_client.go
```diff
@@ -282,16 +290,33 @@ func (b *BoltIndexClient) QueryWithCursor(_ context.Context, c *bbolt.Cursor, qu
...
 
-		if !callback(query, &batch) {
-			break
+		err := batch.send(singleResponse{
+			rangeValue: rangeValue,
+			value:      value,
+		})
+
+		if err != nil {
+			return err
```
kavirajk, line 319: may be worth wrapping this error? Because from the `send` method(), the error can be simply "timedout sending response". which I believe lacks context when I see this error in practice.

kavirajk (commented):
LGTM. Suggested few minor nits

pkg/storage/stores/shipper/testutil/testutil.go
sandeepsukhani: Yes, since we use it with both tests and benchmarks.

kavirajk (approved):
LGTM :+1:

PR #5470, opened 2022-02-24
Fixes ingester sample deduping

cyriltovena:

While introducing #5289  to use the original stream hash to correctly dedupe mutated streams from the query, ingesters were still not correctly returning the right stream hash for their respective streams causing deduping to be all over the place.

This does few things:
- Correctly return the streamhash for merge iterator 🤦 it was returning the log hash
- Use always the hash for consistent data ordering, unless it's zero ( in case of query sharding)
- Rework the readBatch when building GRPC response to not overrides the stream hash.
- Add a master test for ingester deduping that sends a log of duplicate to many ingester and ensure we don't end up in this situation anymore since this was very slow and hard to debug.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/iter/entry_iterator.go
+++ b/pkg/iter/entry_iterator.go
@@ -325,6 +325,9 @@ func NewSortEntryIterator(is []EntryIterator, direction logproto.Direction) Entr
 func (i *entrySortIterator) lessByIndex(k, j int) bool {
 	t1, t2 := i.is[k].Entry().Timestamp.UnixNano(), i.is[j].Entry().Timestamp.UnixNano()
 	if t1 == t2 {
+		// The underlying stream hash may not be available, such as when merging LokiResponses in the
+		// frontend which were sharded. Prefer to use the underlying stream hash when available,
+		// which is needed in deduping code, but defer to label sorting when it's not present.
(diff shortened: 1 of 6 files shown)


owen-d (approved):
Wow, this was a wild ride. Nice fix.

pkg/iter/entry_iterator.go
```diff
@@ -329,18 +325,24 @@ func NewSortEntryIterator(is []EntryIterator, direction logproto.Direction) Entr
 func (i *entrySortIterator) lessByIndex(k, j int) bool {
 	t1, t2 := i.is[k].Entry().Timestamp.UnixNano(), i.is[j].Entry().Timestamp.UnixNano()
 	if t1 == t2 {
-		return i.is[k].Labels() < i.is[j].Labels()
+		if i.is[k].StreamHash() == 0 {
```
owen-d, line 328:
```suggestion
		// The underlying stream hash may not be available, such as when merging LokiResponses in the
		// frontend which were sharded. Prefer to use the underlying stream hash when available,
		// which is needed in deduping code, but defer to label sorting when it's not present.
		if i.is[k].StreamHash() == 0 {
```

PR #5472, opened 2022-02-24
Fix potential deadlock in the table manager

MasslessParticle:
I was looking at how the tablemanager worked and noticed that `getOrCreateTable` had a deadlock. I've included a test that illustrates the issue and fixed it.


pkg/storage/stores/shipper/downloads/table_manager.go
```diff
@@ -153,31 +153,27 @@ func (tm *TableManager) query(ctx context.Context, tableName string, queries []c
...
-		}
-		tm.tablesMtx.Unlock()
+	tablePath := filepath.Join(tm.cfg.CacheDir, tableName)
+	err := chunk_util.EnsureDirectory(tablePath)
+	if err != nil {
+		return nil, err
 	}
 
+	table = NewTable(tableName, filepath.Join(tm.cfg.CacheDir, tableName), tm.indexStorageClient, tm.boltIndexClient, tm.metrics)
+	tm.tables[tableName] = table
```
cyriltovena, line 175: don't you need to write lock here for this ?

MasslessParticle: No, I missed that we're using `RLock` at the top. I've changed it to `Lock`

pkg/storage/stores/shipper/downloads/table_manager.go
```diff
@@ -153,31 +153,27 @@ func (tm *TableManager) query(ctx context.Context, tableName string, queries []c
 }
 
 func (tm *TableManager) getOrCreateTable(tableName string) (*Table, error) {
-	// if table is already there, use it.
 	tm.tablesMtx.RLock()
+	defer tm.tablesMtx.RUnlock()
```
owen-d, line 157: I think we're fixing a deadlock and introducing a new one. We use `RLock` first as a lower-cost way to check if the table exists, but must `RUnlock` it after (the previous PR did this). We cannot `defer` it here because we need to `Lock` it later if the table wasn't found. I think we can introduce a 1 line fix:`defer tm.tablesMtx.Unlock()` which will ensure we release the write lock correctly in all cases it's acquired.

MasslessParticle: Yeah, I missed the `Rlock` vs. `Lock`

The latest commit just defers `tm.tablesMtx.Unlock()` after it's created.

pkg/storage/stores/shipper/downloads/table_manager_test.go
```diff
@@ -35,29 +36,50 @@ func buildTestTableManager(t *testing.T, path string) (*TableManager, stopFunc)
...
-	defer stopFunc()
+		tableManager, stopFunc := buildTestTableManager(t, tempDir)
+		defer stopFunc()
 
-	testutil.TestMultiTableQuery(t, testutil.BuildUserID(0), queries, tableManager, 0, 2000)
+		testutil.TestMultiTableQuery(t, testutil.BuildUserID(0), queries, tableManager, 0, 2000)
+	})
+
+	t.Run("it doesn't deadlock when table create fails", func(t *testing.T) {
+		tempDir := os.TempDir()
```
sandeepsukhani, line 65: Just a nit, lets use `t.TempDir` like other places.

sandeepsukhani (approved):
Good catch! Thanks for fixing it!

pkg/storage/stores/shipper/downloads/table_manager_test.go
sandeepsukhani: Sorry, I pushed this change and resolved the merge conflict because I wanted to cut a new release with this fix.

PR #5474, opened 2022-02-24
Disable sharding of count/avg when labels are mutated

### LINKED ISSUE #5461: count() result is inaccurate
**Describe the bug**

When comparing the number of returned series from a query to the same query with `count()` wrapped around it the results are different.

The same `count` query sent directly to a querier and bypassing the query-frontend returns the correct result.

cyriltovena:

Counting distinct values cannot be sharded when mutating labels over multiple shards, since the same label can appears across multiple shard and will be counted twice.

Since avg() uses count then the same logic is applied for it too.

Fixes #5461

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/logql/ast.go
+++ b/pkg/logql/ast.go
@@ -339,10 +339,8 @@ func (e *LineFilterExpr) String() string {
 }
 
 func (e *LineFilterExpr) Filter() (log.Filterer, error) {
-
 	acc := make([]log.Filterer, 0)
 	for curr := e; curr != nil; curr = curr.Left {
-
 		switch curr.Op {
 		case OpFilterIP:
 			var err error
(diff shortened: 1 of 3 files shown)


pkg/logql/ast.go
```diff
@@ -964,6 +962,18 @@ func (e *VectorAggregationExpr) String() string {
 
 // impl SampleExpr
 func (e *VectorAggregationExpr) Shardable() bool {
+	if e.Operation == OpTypeCount || e.Operation == OpTypeAvg {
```
owen-d, line 965: We really need a way to express operations that aren't `always` or `never` shardable, but shardable depending on how they're combined with other operations. This is a great start 👍.

owen-d (approved):

PR #5477, opened 2022-02-25
logcli: Remove port from TLS server name when provided in --addr

atomic77:

If connecting to a TLS secured endpoint with an explicit port with logcli, certificate validation fails:

```
logcli --ca-cert=$TLS_CA --cert=$TLS_CERT --key=$TLS_KEY --addr=https://<host>:<port> labels
error sending request Get "https://<host>:<port>/loki/api/v1/labels?end=1645559631967253910&start=1645556031967253910": x509: certificate is valid for <host>, <SAN>, not <host>:<port>
Error doing request: Run out of attempts while querying the server
```
The problem can be (insecurely) avoided by adding a `--tls-skip-verify` option. 

This PR ensures that the ServerName field of the TLSConfig struct contains only the host if a port is provided to avoid
the failure. 

Per Contribution guide, didn't open an issue since this is a one-liner... but I can create one if needed.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/cmd/logcli/main.go
+++ b/cmd/logcli/main.go
@@ -228,7 +228,7 @@ func newQueryClient(app *kingpin.Application) client.Client {
 		if err != nil {
 			return err
 		}
-		client.TLSConfig.ServerName = u.Host
+		client.TLSConfig.ServerName = strings.Split(u.Host, ":")[0]
 		return nil
 	}
 


cyriltovena (approved):
LGTM

PR #5481, opened 2022-02-25
Add a DeletionMode config variable

MichelHollands:

A new config variable called DeletionMode with a delete mode has been added as well. It's separate from the existing retention setting. It has a default value which uses the existing delete functionality This config variable will be used later on when the v2 deletion functionality is added.

Rename internal fields for future expansion of the delete requests. 

Validate the delete requests.

**Checklist**
- [X] Documentation updated
- [X] Tests updated


MasslessParticle:
It occurred to me yesterday: We're also going to have to store the start and end times for a delete.

pkg/loki/modules.go
```diff
@@ -711,10 +711,9 @@ func (t *Loki) initCompactor() (services.Service, error) {
 	}
 
 	t.Server.HTTP.Path("/compactor/ring").Methods("GET", "POST").Handler(t.compactor)
-	if t.Cfg.CompactorConfig.RetentionEnabled {
-		t.Server.HTTP.Path("/loki/api/admin/delete").Methods("PUT", "POST").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.AddDeleteRequestHandler)))
-		t.Server.HTTP.Path("/loki/api/admin/delete").Methods("GET").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.GetAllDeleteRequestsHandler)))
-		t.Server.HTTP.Path("/loki/api/admin/cancel_delete_request").Methods("PUT", "POST").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.CancelDeleteRequestHandler)))
+	if t.Cfg.CompactorConfig.DeletionEnabled {
+		t.Server.HTTP.Path("/loki/api/v1/delete").Methods("PUT", "POST").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.AddDeleteRequestHandler)))
+		t.Server.HTTP.Path("/loki/api/v1/delete").Methods("GET").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.GetAllDeleteRequestsHandler)))
```
simonswine, line 716: I like the new endpoints naming.

I also matches really well with existing ones using `<verb>`.

pkg/storage/stores/shipper/compactor/compactor.go
```diff
@@ -67,6 +67,7 @@ type Config struct {
 	RetentionEnabled          bool            `yaml:"retention_enabled"`
 	RetentionDeleteDelay      time.Duration   `yaml:"retention_delete_delay"`
 	RetentionDeleteWorkCount  int             `yaml:"retention_delete_worker_count"`
+	DeletionEnabled           bool            `yaml:"deletion_enabled"`
```
simonswine, line 70: I think this should be `DeletionMode`. As I could foresee quite a few combinations in the future, as we are progression deletion. For example: `disabled|filter-only|filter-and-delete`

simonswine (commented):
Thanks for making a start on this. I think it goes into the right direction, but I think for the time being we would still need to have delete functionality for the case when someone using Loki relies on it.

So my suggestion would be: Let use the new API endpoint name (and ideally the new `promql` field), but still allow deletion of "selector-only" logql requests. Wdyt?

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
```diff
@@ -1,24 +1,31 @@
...
-	RequestID string              `json:"request_id"`
-	StartTime model.Time          `json:"start_time"`
-	EndTime   model.Time          `json:"end_time"`
-	Selectors []string            `json:"selectors"`
-	Status    DeleteRequestStatus `json:"status"`
-	CreatedAt model.Time          `json:"created_at"`
+	RequestID    string              `json:"request_id"`
+	StartTime    model.Time          `json:"start_time"`
+	EndTime      model.Time          `json:"end_time"`
+	LogQLRequest string              `json:"logql_requests"`
```
simonswine, line 14: Maybe this could be a better name (or just `LogQL`)

```suggestion
	Query string              `json:"query"`
```

MichelHollands: Done manually, as it required a lot of renames

pkg/storage/stores/shipper/compactor/compactor.go
MichelHollands: Changed to DeletionMode. A v1 mode was added as well for the current functionality. That can be removed once the new delete functionality is available.

dannykopping:
> I think it goes into the right direction, but I think for the time being we would still need to have delete functionality for the case when someone using Loki relies on it.

Agreed. To expand a bit:
We should also be careful about changing API interfaces - can we maintain both the new and the old endpoints? We can mark the old ones as deprecated and remove them with the next major release.

MichelHollands:
> Agreed. To expand a bit: We should also be careful about changing API interfaces - can we maintain both the new and the old endpoints? We can mark the old ones as deprecated and remove them with the next major release.

We can, but this is an API marked as experimental so I would argue it's fine to change them. We're also changing  the parameters for the PUT/POST.

MasslessParticle (approved):
I think this looks awesome

simonswine (approved):
That's coming together really well. LGTM

A couple of nits in the comments.

pkg/storage/stores/shipper/compactor/compactor.go
```diff
@@ -113,6 +120,7 @@ type Compactor struct {
 	metrics               *metrics
 	running               bool
 	wg                    sync.WaitGroup
+	DeleteMode            deletion.Mode
```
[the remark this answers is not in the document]
MichelHollands: A method to get the value has been added and the variable is now unexported.

pkg/storage/stores/shipper/compactor/compactor.go
```diff
@@ -84,6 +86,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.RetentionDeleteWorkCount, "boltdb.shipper.compactor.retention-delete-worker-count", 150, "The total amount of worker to use to delete chunks.")
 	f.DurationVar(&cfg.DeleteRequestCancelPeriod, "boltdb.shipper.compactor.delete-request-cancel-period", 24*time.Hour, "Allow cancellation of delete request until duration after they are created. Data would be deleted only after delete requests have been older than this duration. Ideally this should be set to at least 24h.")
```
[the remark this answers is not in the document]
MichelHollands: Given that we still use the v1 api here which does support cancellation this can stay. The removal of the paragraph in the docs has been reverted instead.

pkg/storage/stores/shipper/compactor/deletion/mode.go
```diff
@@ -0,0 +1,50 @@
...
+
+type Mode int16
+
+var (
+	errUnknownMode = errors.New("unknown deletion mode")
+)
+
+const (
+	Disabled Mode = iota
+	V1            // The existing experimental log deletion that removes whole streams.
```
simonswine, line 15: I think v1 is not really very descriptive. I would call it e.g. `deprecated-delete-whole-streams`

simonswine (approved):
Thanks for incorporating the review. LGTM

I would like you to think about the `v1` name again for the default/legacy, I think we should be more descritive `v1` doesn't really mean anything.

pkg/storage/stores/shipper/compactor/compactor.go
```diff
@@ -201,7 +215,7 @@ func (c *Compactor) init(storageConfig storage.Config, schemaConfig loki_storage
 	c.indexStorageClient = shipper_storage.NewIndexStorageClient(objectClient, c.cfg.SharedStoreKeyPrefix)
 	c.metrics = newMetrics(r)
 
-	if c.cfg.RetentionEnabled {
+	if c.cfg.RetentionEnabled && c.deleteMode == deletion.WholeStreamDeletion {
```
sandeepsukhani, line 218: We need to be careful here since the code under this condition also sets things up specific to retention.

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
```diff
@@ -3,22 +3,34 @@ package deletion
...
 	"github.com/grafana/loki/pkg/storage/stores/shipper/compactor/retention"
 )
 
-// DeleteRequest holds all the details about a delete request.
 type DeleteRequest struct {
 	RequestID string              `json:"request_id"`
 	StartTime model.Time          `json:"start_time"`
 	EndTime   model.Time          `json:"end_time"`
-	Selectors []string            `json:"selectors"`
+	Queries   []string            `json:"queries"`
```
sandeepsukhani, line 14: Not a strong opinion but `Selectors` still sound relevant here to me and less confusing than `Queriers`.

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
```diff
@@ -3,22 +3,34 @@ package deletion
...
+
+func (d *DeleteRequest) AddQueries(queries []string) error {
+	d.Queries = queries
+	for _, query := range queries {
+		matchers, err := parseDeletionQuery(query)
+		if err != nil {
+			return err
+		}
+
+		d.matchers = append(d.matchers, matchers...)
```
sandeepsukhani, line 30: Wouldn't merging the matchers result in incorrect selection of logs? `{namespace="loki-dev"}` and `{container="compactor"}` would select totally different streams than `{namespace="loki-dev", container="compactor"}`

sandeepsukhani (commented):
Added some comments but other than that it looks good to me.

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
MasslessParticle: This is fixed in [this pr](https://github.com/MichelHollands/loki/pull/8) that changes the API to be onle LogQL query at a time.

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
MichelHollands: Yeah, naming is hard. This was changed because we get more than a selector now. A line filter expression can also be present.

pkg/storage/stores/shipper/compactor/compactor.go
MichelHollands: I've moved that condition to later on in that method.

Is the deletion functionality possible without retention? In other words can be separate them. Or do you always need retention enabled to do deletion?

pkg/storage/stores/shipper/compactor/deletion/mode.go
MichelHollands: It's renamed/

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
sandeepsukhani: `Query` sounds weird because it is for fetching the data but that could be just me. We can still call it `Selector` even with filter expression because it is selecting the logs to be deleted. Not strongly opposing it but as you said naming is hard :shrug:

pkg/storage/stores/shipper/compactor/compactor.go
```diff
@@ -222,8 +236,10 @@ func (c *Compactor) init(storageConfig storage.Config, schemaConfig loki_storage
 			return err
 		}
 
-		c.DeleteRequestsHandler = deletion.NewDeleteRequestHandler(c.deleteRequestsStore, time.Hour, r)
-		c.deleteRequestsManager = deletion.NewDeleteRequestsManager(c.deleteRequestsStore, c.cfg.DeleteRequestCancelPeriod, r)
+		if c.deleteMode == deletion.WholeStreamDeletion {
+			c.DeleteRequestsHandler = deletion.NewDeleteRequestHandler(c.deleteRequestsStore, time.Hour, r)
+			c.deleteRequestsManager = deletion.NewDeleteRequestsManager(c.deleteRequestsStore, c.cfg.DeleteRequestCancelPeriod, r)
```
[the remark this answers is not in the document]
MichelHollands: The link in "here" doesn't show a location. Can you double check the change I made?

sandeepsukhani (approved):
Just some minor nits. Thanks for taking care of all the feedback and nice work!

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
```diff
@@ -3,22 +3,30 @@ package deletion
...
 	Status    DeleteRequestStatus `json:"status"`
 	CreatedAt model.Time          `json:"created_at"`
 
-	UserID   string              `json:"-"`
-	Matchers [][]*labels.Matcher `json:"-"`
+	UserID   string            `json:"-"`
+	matchers []*labels.Matcher `json:"-"`
+}
+
+func (d *DeleteRequest) AddQuery(logQL string) error {
```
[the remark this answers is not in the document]
MichelHollands: That indeed makes it clearer. I've updated it manually as it involved changes in other files as well.

PR #5482, opened 2022-02-25
Query Filtering -- Querier

MasslessParticle:
This is the first PR for query filtering for deletes. It adds a delete store to the querier. The querier uses that to lookup any delete request for the current user, finds the ones that overlap with the query range, and forwards them to downstream ingesters and stores.

This is implemented for both Logs and Metrics.


pkg/querier/querier.go
```diff
@@ -191,6 +209,36 @@ func (q *Querier) SelectSamples(ctx context.Context, params logql.SelectSamplePa
...
+	}
+
+	start := startT.UnixNano()
+	end := endT.UnixNano()
+
+	var deletes []*logproto.Delete
+	for _, del := range d {
+		for _, selector := range del.Selectors {
+			if int64(del.StartTime) <= end && int64(del.EndTime) >= start {
+				deletes = append(deletes, &logproto.Delete{
```
MichelHollands, line 230: In a future PR it would be nice to have this as a metrics somewhere so we can see if the different instances are in sync.

MasslessParticle:
This relies on https://github.com/grafana/loki/pull/5481, converting to a draft PR so I can make the necessary changes when it's merged

owen-d (commented):
This is looking really good. When are you planning to remove `draft`?

MasslessParticle:
There's nothing technically stopping it. I think it's going to conflict with #5481 quite a bit, though.

owen-d (approved):

PR #5484, opened 2022-02-26
per user index query readiness with limits overrides

sandeepsukhani:
The current query readiness support works only on table level. With a cluster building per user index and having query readiness configured, it would add a lot of overhead to keep the active index synced since we get a new index from ingesters every 15 mins.
This PR adds support for more fine-grained control over query readiness for per user index using limits overrides config.

I have also changed the metric to track the duration it takes to download the index at query time. It would be tracked per table. I have not added a user id to it to avoid adding too many series. I have added a log line which we can use for viewing per tenant download duration by both user id and table.

I have also dropped the older metric `loki_boltdb_shipper_initial_tables_download_duration_seconds`, which was not very reliable and the name was not very clear too.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/chunk/storage/caching_fixtures.go
+++ b/pkg/storage/chunk/storage/caching_fixtures.go
@@ -12,7 +12,7 @@ import (
 	"github.com/grafana/loki/pkg/storage/chunk/cache"
 	"github.com/grafana/loki/pkg/storage/chunk/gcp"
 	"github.com/grafana/loki/pkg/storage/chunk/testutils"
-	"github.com/grafana/loki/pkg/util/validation"
+	"github.com/grafana/loki/pkg/validation"
 )
 
 type fixture struct {
--- a/pkg/storage/stores/shipper/downloads/metrics.go
+++ b/pkg/storage/stores/shipper/downloads/metrics.go
@@ -1,8 +1,6 @@
 package downloads
 
 import (
-	"sync"
-
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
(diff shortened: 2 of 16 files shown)


pkg/usagestats/reporter_test.go
```diff
@@ -15,20 +15,15 @@ import (
...
 func Test_LeaderElection(t *testing.T) {
 	stabilityCheckInterval = 100 * time.Millisecond
 
 	result := make(chan *ClusterSeed, 10)
-	objectClient, err := storage.NewObjectClient(storage.StorageTypeFileSystem, storage.Config{
-		FSConfig: local.FSConfig{
-			Directory: t.TempDir(),
-		},
-	}, metrics)
+	objectClient, err := local.NewFSObjectClient(local.FSConfig{
```
sandeepsukhani, line 24: All the test changes in `usagestats` package were done to avoid a circular dependency.
It was caused by the import fix I made at https://github.com/grafana/loki/pull/5484/files#diff-8aa5c09f2d873a1f7c7a48d253da4d5d1a434cf2fa32ef84e5c1a68e0617f82aR15

pkg/storage/chunk/storage/caching_fixtures.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/grafana/loki/pkg/storage/chunk/cache"
 	"github.com/grafana/loki/pkg/storage/chunk/gcp"
 	"github.com/grafana/loki/pkg/storage/chunk/testutils"
-	"github.com/grafana/loki/pkg/util/validation"
+	"github.com/grafana/loki/pkg/validation"
```
sandeepsukhani, line 15: The `github.com/grafana/loki/pkg/util/validation` points to a copy of the cortex validation package, which we should get clean up sometime.

pkg/storage/stores/shipper/downloads/metrics.go
```diff
@@ -12,42 +10,19 @@ const (
...
-	}
-
-	m.gauge.Set(totalDuration)
-}
-
 type metrics struct {
-	// metrics for measuring performance of downloading of files per period initially i.e for the first time
-	tablesDownloadDurationSeconds *downloadTableDurationMetric
-
+	queryTimeTableDownloadDurationSeconds  *prometheus.CounterVec
```
cyriltovena, line 14: I wonder if an histogram wouldn't be better here ?

sandeepsukhani: Not sure if that would add much value. Would we be interested in looking at a distribution? I think it would be rare and we can use `quantile_over_time` from the logline I have added [here](https://github.com/grafana/loki/pull/5484/files#diff-3b7a21722d5aa578d861ff5625cc9e841936e7a763e48d0009e196bfc691dd8eR298). wdyt?

pkg/storage/stores/shipper/downloads/table_manager.go
```diff
@@ -232,97 +240,107 @@ func (tm *TableManager) cleanupCache() error {
...
-	var requiredTableNames []string
-
-	for _, tableName := range tablesInStorage {
-		if !re.MatchString(tableName) {
-			continue
+	for _, userID := range usersWithIndexInTable {
+		// use the query readiness config for the user if it exists or use the default config
+		queryReadyNumDays, ok := queryReadinessNumByUserID[userID]
+		if !ok {
+			queryReadyNumDays = tm.cfg.Limits.DefaultLimits().QueryReadyIndexNumDays
```
cyriltovena, line 331: This is not consistent with above where you first check the config and after the limit.

I wonder if we should in fact remove the config ? Or deprecate it ?

cyriltovena (approved):
LGTM.

Left some questions however.

pkg/storage/stores/shipper/downloads/metrics.go
cyriltovena: What's the PromQL you'll use with this ?

sandeepsukhani: something like `sum(quantile_over_time({container="index-gateway"} |= "downloaded index set at query time" | logfmt | unwrap duration(duration)[5m]) by (user_id)`. Makes sense?

pkg/storage/stores/shipper/downloads/table_manager.go
```diff
@@ -71,7 +79,7 @@ func NewTableManager(cfg Config, boltIndexClient BoltDBIndexClient, indexStorage
 	}
 
 	// download the missing tables.
-	err = tm.ensureQueryReadiness()
+	err = tm.ensureQueryReadiness(context.Background())
```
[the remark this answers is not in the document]
sandeepsukhani: yeah, sorry my bad. Good catch!

PR #5490, opened 2022-02-28
Stub multi tenant querier.

### LINKED ISSUE #231: Consider working with CRI-O to deal with byte streams from containers rather than scraping logs from files
Please consider a path laid out with https://github.com/kubernetes-sigs/cri-o/pull/1605.

Instead of scraping logs from disk, work with CRI-O to get a byte stream (no newline interpretation) of the container output chunked through to the storage system.

Then apply what ever interpretation of the logs one wants later when consuming the data from the store.  This avoids all the problems of scraping logs from disk, while efficiently working with the container subsystems to get the container output collected with the right metadata attached.

It also means we can focus on collection with proper metadata efficiently without worry for what interpretation of the container output stream will be applied later.  That is, views of the log data can be considered isolated and independent from how the data is collected.

jeschkies:
- Introduce Queries interface and Querier API.
- Configure a multi-tenant querier that just passes requests through.

Closes #231

**Checklist**
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -80,7 +80,7 @@ type Querier interface {
 	Tail(ctx context.Context, req *logproto.TailRequest) (*Tailer, error)
 }
 
-// SingleTenantQuerier handlers queries.
+// SingleTenantQuerier handles single tenant queries.
 type SingleTenantQuerier struct {
 	cfg             Config
 	store           storage.Store
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -80,8 +80,8 @@ type Querier interface {
 	Tail(ctx context.Context, req *logproto.TailRequest) (*Tailer, error)
 }
 
-// Querier handlers queries.
-type querier struct {
+// SingleTenantQuerier handlers queries.
+type SingleTenantQuerier struct {
 	cfg             Config
 	store           storage.Store
 	limits          *validation.Overrides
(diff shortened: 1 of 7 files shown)


pkg/querier/querier.go
```diff
@@ -72,8 +74,17 @@ func (cfg *Config) Validate() error {
 	return nil
 }
 
+// Querier can select logs and samples and handle query requests.
+type Querier interface {
+	logql.Querier
+	Label(ctx context.Context, req *logproto.LabelRequest) (*logproto.LabelResponse, error)
+	Query(params logql.Params) logql.Query
+	Series(ctx context.Context, req *logproto.SeriesRequest) (*logproto.SeriesResponse, error)
+	Tail(ctx context.Context, req *logproto.TailRequest) (*Tailer, error)
```
jeschkies, line 83: I was hoping it would be enough to only implement `logql.Querier`. However, we might want to support Label and Series as well. Not sure about tail.

jeschkies:
The linter is failing with
```
pkg/querier/http.go:37:6: exported: type name will be used as querier.QuerierAPI by other packages, and that stutters; consider calling this API (revive)

type QuerierAPI struct {

     ^
```
should I follow the advice or ignore it?

pkg/querier/querier.go
```diff
@@ -62,6 +61,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.MaxConcurrent, "querier.max-concurrent", 10, "The maximum number of concurrent queries.")
 	f.BoolVar(&cfg.QueryStoreOnly, "querier.query-store-only", false, "Queriers should only query the store and not try to query any ingesters")
 	f.BoolVar(&cfg.QueryIngesterOnly, "querier.query-ingester-only", false, "Queriers should only query the ingesters and not try to query any store")
+	f.BoolVar(&cfg.MultiTenantQueriesEnabled, "querier.multi-tenant-queries-enabled", false, "Enable queries across multiple tenants.")
```
cyriltovena, line 64: Should we add a note that this is experimental ?

cyriltovena (approved):
LGTM

pkg/querier/querier.go
jeschkies: Yes!

PR #5497, opened 2022-02-28
Fix orphaned metrics in the file tailer

MasslessParticle:
Individual tailers are responsible for cleaning up their own metrics but in the case where no tailer is created because: 
- The file to be tailed exists [here](https://github.com/grafana/loki/blob/cdde6aa49dbb98d6bbfa0f29e0b3b3023211ca08/clients/pkg/promtail/targets/file/filetarget.go#L214) 
- but is gone by the time `startTailing` is called [here](https://github.com/grafana/loki/blob/cdde6aa49dbb98d6bbfa0f29e0b3b3023211ca08/clients/pkg/promtail/targets/file/filetarget.go#L238)

We orphan metrics for the affected paths.

This PR makes sure to cleanup any metrics for the path when there's an error creating the associated tailer.

### DIFF
--- a/clients/pkg/promtail/targets/file/filetarget.go
+++ b/clients/pkg/promtail/targets/file/filetarget.go
@@ -275,15 +275,20 @@ func (t *FileTarget) startTailing(ps []string) {
 		if _, ok := t.tails[p]; ok {
 			continue
 		}
+
 		fi, err := os.Stat(p)
 		if err != nil {
 			level.Error(t.logger).Log("msg", "failed to tail file, stat failed", "error", err, "filename", p)
+			t.metrics.totalBytes.DeleteLabelValues(p)
 			continue
 		}
+
 		if fi.IsDir() {
 			level.Info(t.logger).Log("msg", "failed to tail file", "error", "file is a directory", "filename", p)
+			t.metrics.totalBytes.DeleteLabelValues(p)
 			continue
 		}
+
 		level.Debug(t.logger).Log("msg", "tailing new file", "filename", p)
 		tailer, err := newTailer(t.metrics, t.logger, t.handler, t.positions, p)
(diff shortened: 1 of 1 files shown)


clients/pkg/promtail/targets/file/filetarget.go
```diff
@@ -275,15 +275,20 @@ func (t *FileTarget) startTailing(ps []string) {
 		if _, ok := t.tails[p]; ok {
 			continue
 		}
+
 		fi, err := os.Stat(p)
 		if err != nil {
 			level.Error(t.logger).Log("msg", "failed to tail file, stat failed", "error", err, "filename", p)
+			t.metrics.totalBytes.DeleteLabelValues(p)
```
kavirajk, line 282: wondering why just `totalBytes` and why not clear all metrics just like tailer's `t.cleanupMetrics()`? And good to make it as part of `defer()`? 

https://github.com/grafana/loki/blob/0728237530f9c6d11211a6f30fef2863243406d2/clients/pkg/promtail/targets/file/tailer.go#L221

MasslessParticle: I like the idea of managing metrics together and as close to where they're used as possible. This situation is a little awkward because `totalBytes` is the only metric managed in two places: here and in the tailer. 

I'm not sure `defer` would be a good fit because the only metric this type cares about is `totalBytes` and only in the error case. To successfully cleanup, we'd have to capture the error in the defer, too. I think it could be pretty confusing.

kavirajk: thanks for clarifying. I'm convinced :)

kavirajk (approved):
LGTM. thanks @MasslessParticle

PR #5502, opened 2022-03-01
[Frontend] Log Result Cache

cyriltovena:

Until now we were only caching previous steps results for metrics queries, we never really implemented a log result cache as there are some complexities attached to it:
-  For example since logs are limited (usually 1000) how much should we cache ? 
-  How expensive are actually those queries if they returns logs quickly ? 
-  Can we afford to cache log result and how much can we cache ?

Because of this I decided to actually cache filters queries that returns no result ! This is much easier, and they are known to be more expensive queries.

You can find more about this decision here: https://docs.google.com/document/d/1_mACOpxdWZ5K0cIedaja5gzMbv-m0lUVazqZd2O4mEU/edit

This means if you're running a 12h queries looking for foo (`{app="bar"} |= "foo"`), and no result return, refreshing will only request the last 2h of data, saving 10h of data to be processed.

Fixes #<issue number>

Still need to update the change-log, testing it properly in dev.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,273 @@
...
+	var (
+		startRequest, endRequest *LokiRequest
+		startResp, endResp       *LokiResponse
+		updateCache              bool
+	)
+	g, ctx := errgroup.WithContext(ctx)
+
+	if lokiReq.GetStartTs().Before(cachedResquest.GetStartTs()) {
+		g.Go(func() error {
+			startRequest = lokiReq.WithStartEndTime(lokiReq.GetStartTs(), cachedResquest.GetStartTs())
```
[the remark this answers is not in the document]
cstyan: it looks like you have `cachedResquest` in a few places :)

cstyan (commented):
I don't have any suggestions off the top of my head, but personally I'm finding the code in `log_result_cache.go` a bit hard to follow. Especially `handleHit`, maybe we can split that into some more functions? Like adding another function to `updateCacheResults`?

pkg/querier/queryrange/log_result_cache.go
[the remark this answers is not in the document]
cyriltovena: Yes it's intentional I'm fetching what is missing. From the req to what is in the cache.

cyriltovena:
> I don't have any suggestions off the top of my head, but personally I'm finding the code in `log_result_cache.go` a bit hard to follow. Especially `handleHit`, maybe we can split that into some more functions? Like adding another function to `updateCacheResults`?

I added better comment.

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	return resp, nil
+}
+
+func (l *logResultCache) handleHit(ctx context.Context, cacheKey string, cachedResquest *LokiRequest, lokiReq *LokiRequest) (queryrangebase.Response, error) {
+	l.metrics.CacheHit.Inc()
+	// we start with an empty response
+	result := emptyResponse(cachedResquest)
+	// if the request is the same and cover the whole time range.
+	// we can just return the cached result.
```
simonswine, line 168: Nit

```suggestion
	// if the request is the same and covers the whole time range,
	// we can just return the cached result.
```

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+			Namespace: "loki",
+			Name:      "query_frontend_log_result_cache_miss_total",
+		}),
+	}
+}
+
+// NewLogResultCache creates a new log result cache middleware.
+// Currently it only caches empty filter queries, this is because those are usually easily and freely cacheable.
+// Log hits are difficult to handle because of the limit query parameter and the size of the response.
+// In the future it could be extended to cache log hits.
```
simonswine, line 50: Maybe that is clearer so a "log hit" is not confused with a "log result cache hit".

```suggestion
// In the future it could be extended to cache non-empty query results.
```

pkg/querier/queryrange/log_result_cache.go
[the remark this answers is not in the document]
simonswine: I also thought this is a typo and still have a feeling it might be. Maybe it needs to be clearer what the extra `s` stands for?

```suggestion
			startRequest = lokiReq.WithStartEndTime(lokiReq.GetStartTs(), cachedRequest.GetStartTs())
```

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return nil, httpgrpc.Errorf(http.StatusBadRequest, err.Error())
+	}
+
+	if l.shouldCache != nil && !l.shouldCache(req) {
+		return l.next.Do(ctx, req)
+	}
+
+	maxCacheFreshness := validation.MaxDurationPerTenant(tenantIDs, l.limits.MaxCacheFreshness)
+	maxCacheTime := int64(model.Now().Add(-maxCacheFreshness))
```
simonswine, line 89: I wonder if we need to consider the value of `MaxChunkAge`. In theory it could happen that log delivery is delayed and logs outside of the maxCacheFreshness duration are ingested. Subsequently those results would then not show up if they had been queried earlier.

simonswine (commented):
I think this already can unlock some quicker processing, and we can get more familiar with the results caching efficiencies for log queries.

Also good to see the refactoring. :slightly_smiling_face: 

See my comments in line

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+			}
+			result = mergeLokiResponse(startResp, result)
+		}
+	}
+
+	// if we have data at the end, we need to merge it with the cached data if it's empty and update the cache.
+	// If it's not empty only merge the response.
+	if endResp != nil {
+		if isEmpty(endResp) {
+			cachedResquest = cachedResquest.WithStartEndTime(cachedResquest.GetStartTs(), endRequest.GetEndTs())
```
ssncferreira, line 238: If we also have data on the start (line 224), this will override the start cached request, no?

pkg/querier/queryrange/log_result_cache_test.go
```diff
@@ -0,0 +1,452 @@
...
+	h := lrc.Wrap(fake)
+
+	resp, err := h.Do(ctx, req)
+	require.NoError(t, err)
+	require.Equal(t, emptyResponse(req), resp)
+	resp, err = h.Do(ctx, req)
+	require.NoError(t, err)
+	require.Equal(t, emptyResponse(req), resp)
+
+	fake.AssertExpectations(t)
```
ssncferreira, line 58: Maybe it would also be useful to assert that there are new entries in the cache. Wdyt?

ssncferreira (commented):
Overall LGTM 🎉 
I have just some minor questions 🙂

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// NewLogResultCache creates a new log result cache middleware.
+// Currently it only caches empty filter queries, this is because those are usually easily and freely cacheable.
+// Log hits are difficult to handle because of the limit query parameter and the size of the response.
+// In the future it could be extended to cache non-empty query results.
+// see https://docs.google.com/document/d/1_mACOpxdWZ5K0cIedaja5gzMbv-m0lUVazqZd2O4mEU/edit
+func NewLogResultCache(logger log.Logger, limits Limits, cache cache.Cache, shouldCache queryrangebase.ShouldCacheFn, metrics *LogResultCacheMetrics) queryrangebase.Middleware {
+	if metrics == nil {
+		metrics = NewLogResultCacheMetrics(nil)
```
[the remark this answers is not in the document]
cstyan: I think it's probably because somewhere we make calls to `someMetric.Add/Inc/Set` etc. which would cause issues if the metric struct is nil. OTOH if the struct is instantiated but the `prometheus.Registerer` was nil then it's just like writing to `/dev/null`.

That being said, my opinion is that we should move to a pattern where our constructors either take in an already created metrics struct, or the `prometheus.Registerer` that will be passed to `NewTypeMetrics`.

pkg/querier/queryrange/log_result_cache.go
cyriltovena: It's a good point (although unrelated to `MaxChunkAge`), as for today for the metric result cache we ensure maxCacheFreshness is always 2h which matches the max oldest log timestamp we can receive.

Here I'm using the same as the cache for metrics, I can see how in the future we could avoid this config and set it ourselves based on other configuration.

pkg/querier/queryrange/log_result_cache.go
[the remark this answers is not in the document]
cyriltovena: Callum is correct. I can pass prometheus.Registerer instead of nil if that fits better ? WDYT Callum ?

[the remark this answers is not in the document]
cstyan: I would lean towards the ctor `NewLogResultsCache` taking in an instnace of the metrics struct as a parameter, but passing a registerer in and using that is okay too.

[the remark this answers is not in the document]
cyriltovena: it is actually taking the instance of the metrics.

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+
+	interval := validation.SmallestPositiveNonZeroDurationPerTenant(tenantIDs, l.limits.QuerySplitDuration)
+	// skip caching by if interval is unset
+	if interval == 0 {
+		return l.next.Do(ctx, req)
+	}
+	// The first subquery might not be aligned.
+	alignedStart := time.Unix(0, lokiReq.GetStartTs().UnixNano()-(lokiReq.GetStartTs().UnixNano()%interval.Nanoseconds()))
+	// generate the cache key based on query, tenant and start time.
+	cacheKey := fmt.Sprintf("log:%s:%s:%d", tenant.JoinTenantIDs(tenantIDs), req.GetQuery(), alignedStart.UnixNano()/(interval.Nanoseconds()))
```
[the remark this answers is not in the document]
cyriltovena: I think for multi tenant queries the result is expected to be the same so that should be fine.

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	// cache the result
+	err = l.cache.Store(ctx, []string{cache.HashKey(cacheKey)}, [][]byte{data})
+	if err != nil {
+		level.Warn(l.logger).Log("msg", "error storing cache", "err", err)
+	}
+	return resp, nil
+}
+
+func (l *logResultCache) handleHit(ctx context.Context, cacheKey string, cachedResquest *LokiRequest, lokiReq *LokiRequest) (queryrangebase.Response, error) {
```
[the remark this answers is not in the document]
cyriltovena: Fixed it already

pkg/querier/queryrange/log_result_cache_test.go
cyriltovena: It is kinda already tested by calling it twice and expecting only one call

pkg/querier/queryrange/log_result_cache.go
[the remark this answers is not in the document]
cyriltovena: It was a typo.

pkg/querier/queryrange/log_result_cache.go
cyriltovena: it is using the start of `cachedResquest` so whatever was set is reused here. Am I missing something ?

ssncferreira: Ah ok, got it. I got confused with the start and end times.

So if uncached data exists both at the start and at the end, the cache will be updated with time interval [startRequest.GetStartTs(), endRequest.GetEndTs()]. Couldn't this affect the interval part of the cache key? https://github.com/grafana/loki/blob/c5d27d0e8f696e6f29810f721b882eb8ee413f90/pkg/querier/queryrange/log_result_cache.go#L107 🤔

cyriltovena: It won't affect the key becaue it's forced to be aligned with the interval.

ssncferreira: But it is still possible that `startRequest.GetStartTs()` is before the aligned start interval and `endRequest.GetEndTs()` after the next start interval, right? As the example in the following diagram:
cache_diagram
The updated cache_n time interval is bigger than the expected size based on the split by duration interval.
With the current implementation, it is possible for this to happen, no?

The only issue here is that the updated cache_n could encompass one or more extra subsequent cache blocks, but that shouldn't be a problem.

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+		return nil, httpgrpc.Errorf(http.StatusBadRequest, err.Error())
+	}
+
+	if l.shouldCache != nil && !l.shouldCache(req) {
+		return l.next.Do(ctx, req)
+	}
+
+	maxCacheFreshness := validation.MaxDurationPerTenant(tenantIDs, l.limits.MaxCacheFreshness)
+	maxCacheTime := int64(model.Now().Add(-maxCacheFreshness))
+	if req.GetStart() > maxCacheTime {
```
owen-d, line 90: What about when `req.End > maxCacheTime`? I figure we should avoid caching requests that overlap with the freshness requirement, rather than those that just start in that period. Otherwise, we could cache a request that has no data then receive a push with data.

pkg/querier/queryrange/log_result_cache.go
owen-d, line 107: nit: unlikely, but I think this should have `interval` in the cache key to protect against limits reloading from causing conflicting cache keys (we do this for other cache keys).

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+	if !isEmpty(lokiRes) {
+		return resp, nil
+	}
+	data, err := proto.Marshal(req)
+	if err != nil {
+		level.Warn(l.logger).Log("msg", "error marshalling request", "err", err)
+		return resp, nil
+	}
+	// cache the result
+	err = l.cache.Store(ctx, []string{cache.HashKey(cacheKey)}, [][]byte{data})
```
owen-d, line 156: I think this should be done in the background to avoid introducing extra latency, especially if we're seeing cache-related issues.

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func (l *logResultCache) handleHit(ctx context.Context, cacheKey string, cachedResquest *LokiRequest, lokiReq *LokiRequest) (queryrangebase.Response, error) {
+	l.metrics.CacheHit.Inc()
+	// we start with an empty response
+	result := emptyResponse(cachedResquest)
+	// if the request is the same and covers the whole time range,
+	// we can just return the cached result.
+	if lokiReq.GetStartTs().After(cachedResquest.GetStartTs()) || lokiReq.GetStartTs().Equal(cachedResquest.GetStartTs()) &&
+		lokiReq.GetEndTs().Before(cachedResquest.GetEndTs()) || lokiReq.GetEndTs().Equal(cachedResquest.GetEndTs()) {
```
owen-d, line 170: nit: simplify
```suggestion
	if !lokiReq.GetStartTs().Before(cachedResquest.GetStartTs()) &&
		!lokiReq.GetEndTs().After(cachedResquest.GetEndTs()) {
```

pkg/querier/queryrange/log_result_cache.go
owen-d: I don't think it's possible to be referring to different cache blocks at this point because we calculated `alignedStart` prior to seeing if we had data in the cache. Even if we only had a partial cache here, the range which we're checking will already align with the `alignedStart` cache key.

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// we need to update the cache since we fetched more either at the end or the start and it was empty.
+	if updateCache {
+		data, err := proto.Marshal(cachedResquest)
+		if err != nil {
+			level.Warn(l.logger).Log("msg", "error marshalling request", "err", err)
+			return result, err
+		}
+		// cache the result
+		err = l.cache.Store(ctx, []string{cache.HashKey(cacheKey)}, [][]byte{data})
```
owen-d, line 256: I think we should background this `Store` call

pkg/querier/queryrange/roundtrip.go
```diff
@@ -42,36 +42,46 @@ func NewTripperware(
...
+	var (
+		c   cache.Cache
+		err error
+	)
+	if cfg.CacheResults {
+		c, err = cache.New(cfg.CacheConfig, registerer, log)
+		if err != nil {
+			return nil, nil, err
+		}
+		if cfg.Compression == "snappy" {
```
owen-d, line 56: Do we need to check other compression formats?

owen-d (commented):
This is looking solid.

nit: this would make an easier review if the feature and refactoring were split into two PRs.

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -0,0 +1,280 @@
...
+		return l.next.Do(ctx, req)
+	}
+
+	if len(buff) == 0 {
+		// cache miss
+		return l.handleMiss(ctx, cacheKey, lokiReq)
+	}
+
+	// cache hit
+	var cachedResquest LokiRequest
```
trevorwhitney, line 126: here's another `cachedResquest`. I see others have commented that this should probably be `cachedRequest`?

pkg/querier/queryrange/log_result_cache.go
trevorwhitney: nit: I also wouldn't mind moving this to a method that returns a boolean and clearly states what it's doing, ie `requestWithinCachedRequestTimerange` or something.

pkg/querier/queryrange/queryrangebase/results_cache_test.go
```diff
@@ -751,9 +752,11 @@ func TestResultsCache(t *testing.T) {
 			Cache: cache.NewMockCache(),
 		},
 	}
-	rcm, _, err := NewResultsCacheMiddleware(
+	c, err := cache.New(cfg.CacheConfig, nil, log.NewNopLogger())
+	require.NoError(t, err)
+	rcm, err := NewResultsCacheMiddleware(
 		log.NewNopLogger(),
-		cfg,
+		c,
```
trevorwhitney, line 759: All I see here is changes to the test to satisfy the new signature, have there been any tests added to confirm that empty responses are actually cached? Sorry if I just missed them by only looking at changes.

trevorwhitney (commented):
Looks good, just a few nits and questions.

pkg/querier/queryrange/log_result_cache.go
cyriltovena: Just requiring the end would do then because the start is always before.

pkg/querier/queryrange/log_result_cache.go
cyriltovena: I don't think caching small item like this will be a problem, specially that the timeout is super aggressive. I don't feel like doing this just yet unless it shows to be a problem in the future.

pkg/querier/queryrange/roundtrip.go
cyriltovena: This is copied from the other cache, I don't think we support anything else.

pkg/querier/queryrange/log_result_cache.go
cyriltovena: I would do it if it was used anywhere else, I think the comment is enough there.

pkg/querier/queryrange/log_result_cache.go
cyriltovena: it's already gone :)

pkg/querier/queryrange/queryrangebase/results_cache_test.go
cyriltovena: Yes there is a new test file for this. This is the current test file for metric result.

pkg/querier/queryrange/log_result_cache.go
cyriltovena: I'm thinking about this. I don't think the start is a problem as it will be always either exactly at the start or within never before.

But the end can be problematic.

pkg/querier/queryrange/codec.go
```diff
@@ -909,3 +895,31 @@ func NewEmptyResponse(r queryrangebase.Request) (queryrangebase.Response, error)
 		return nil, fmt.Errorf("unsupported request type %T", req)
 	}
 }
+
+func mergeLokiResponse(responses ...queryrangebase.Response) *LokiResponse {
+	var (
+		lokiRes       = responses[0].(*LokiResponse)
```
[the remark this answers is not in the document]
cyriltovena: This is just refactoring from a code used else where. In our usage we always pass items. In the previous usage they already checked that before, I'll add a quick check just in case.

pkg/querier/queryrange/log_result_cache.go
cyriltovena: In which case that seems to be a non-issue, the end might include more data.

cyriltovena:
I think we're good to go here. /cc @owen-d

owen-d (approved):

PR #5485, opened 2022-02-27
metrics: Add `user` label to `worker_inflight_queries`

kavirajk:
**What this PR does / why we need it**:
Add additional `user`  label to the existing `loki_querier_worker_inflight_queries` metrics.
This enables us observe per tenant in flight requests in the query workers.

/cc @owen-d 

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] Add an entry in the `CHANGELOG.md` about the changes.

PR #5507, opened 2022-03-02
Remove extra param in call for inflightRequests metric

MichelHollands:
**What this PR does / why we need it**:

The call to update the inflightRequests metric incorrectly adds a "user" argument which seems to be used as a label name. This causes a run time panic as it only expects the label value:

```
10:22:22 enterprise-logs: panic: inconsistent label cardinality: expected 1 label values but got 2 in []string{"user", "test_instance"}
10:22:22 enterprise-logs: goroutine 966 [running]:
10:22:22 enterprise-logs: github.com/prometheus/client_golang/prometheus.(*GaugeVec).WithLabelValues(0x28933e0, {0xc0015fb9c0, 0xc001a24510, 0xc0001022d0})
```

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] Add an entry in the `CHANGELOG.md` about the changes.

Signed-off-by: Michel Hollands <michel.hollands@grafana.com>

PR #5508, opened 2022-03-02
fix the broken query worker code for incrementing inflight queries metric with label values

sandeepsukhani:
PR https://github.com/grafana/loki/pull/5485 added `user` label on `worker_inflight_queries` metric. It is using `WithLabelValues` function which requires passing just label values in order of the label names defined in the metric definition but instead it is trying to pass the label name as well. The function name is not super clear which causes a lot of people to fall for it.

This PR fixes the issue since the queriers are panicing with error `panic: inconsistent label cardinality: expected 1 label values but got 2 in []string{"user", "145265"}`.


dannykopping:
I believe this was fixed by https://github.com/grafana/loki/pull/5507

sandeepsukhani:
> I believe this was fixed by #5507 - slightly too late to the party stuck_out_tongue

haha yes, will close this. Thanks!

PR #5512, opened 2022-03-02
Update out_loki.rb

HolgerHees:
this will fix "UndefinedConversionError error="\"\\xC2\" from ASCII-8BIT to UTF-8"" for messages of systemd journal


dannykopping (approved):
LGTM, thanks @HolgerHees

PR #5514, opened 2022-03-02
Ring mode index gateway

DylanGuedes:
TBD

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -81,6 +81,7 @@ const (
 	MemberlistKV             string = "memberlist-kv"
 	Compactor                string = "compactor"
 	IndexGateway             string = "index-gateway"
+	IndexGatewayRing         string = "index-gateway-ring"
 	QueryScheduler           string = "query-scheduler"
 	All                      string = "all"
 	Read                     string = "read"
@@ -391,7 +392,7 @@ func (t *Loki) initStore() (_ services.Service, err error) {
 		}
 	}
 
-	chunkStore, err := chunk_storage.NewStore(t.Cfg.StorageConfig.Config, t.Cfg.ChunkStoreConfig.StoreConfig, t.Cfg.SchemaConfig.SchemaConfig, t.overrides, t.clientMetrics, prometheus.DefaultRegisterer, nil, util_log.Logger)
+	chunkStore, err := chunk_storage.NewStore(t.Cfg.StorageConfig.Config, t.Cfg.ChunkStoreConfig.StoreConfig, t.Cfg.SchemaConfig.SchemaConfig, t.indexGatewayRing, t.overrides, t.clientMetrics, prometheus.DefaultRegisterer, nil, util_log.Logger)
 	if err != nil {
 		return
 	}
@@ -749,6 +750,20 @@ func (t *Loki) initIndexGateway() (services.Service, error) {
 	t.Server.HTTP.Path("/indexgateway/ring").Methods("GET", "POST").Handler(gateway)
 	indexgatewaypb.RegisterIndexGatewayServer(t.Server.GRPC, gateway)
 	return gateway, nil
+}
+
+func (t *Loki) initIndexGatewayRing() (_ services.Service, err error) {
+	t.Cfg.IndexGateway.IndexGatewayRing.KVStore.Multi.ConfigProvider = multiClientRuntimeConfigChannel(t.runtimeConfig)
+	t.Cfg.IndexGateway.IndexGatewayRing.KVStore.MemberlistKV = t.MemberlistKV.GetMemberlistKV
+	ringCfg := t.Cfg.IndexGateway.IndexGatewayRing.ToRingConfig(indexgateway.RingReplicationFactor)
+	reg := prometheus.WrapRegistererWithPrefix("loki_", prometheus.DefaultRegisterer)
+	t.indexGatewayRing, err = ring.New(ringCfg, indexgateway.RingIdentifier, indexgateway.RingKey, util_log.Logger, reg)
+	if err != nil {
+		return
+	}
+
+	t.Server.HTTP.Path("/indexgateway/ring").Methods("GET", "POST").Handler(t.indexGatewayRing)
+	return t.indexGatewayRing, nil
 }
 
 func (t *Loki) initQueryScheduler() (services.Service, error) {
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -392,7 +392,8 @@ func (t *Loki) initStore() (_ services.Service, err error) {
 		}
 	}
 
-	chunkStore, err := chunk_storage.NewStore(t.Cfg.StorageConfig.Config, t.Cfg.ChunkStoreConfig.StoreConfig, t.Cfg.SchemaConfig.SchemaConfig, t.indexGatewayRing, t.overrides, t.clientMetrics, prometheus.DefaultRegisterer, nil, util_log.Logger)
+	t.Cfg.StorageConfig.BoltDBConfig.Ring = t.indexGatewayRing
+	chunkStore, err := chunk_storage.NewStore(t.Cfg.StorageConfig.Config, t.Cfg.ChunkStoreConfig.StoreConfig, t.Cfg.SchemaConfig.SchemaConfig, t.overrides, t.clientMetrics, prometheus.DefaultRegisterer, nil, util_log.Logger)
 	if err != nil {
 		return
 	}
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -392,7 +392,7 @@ func (t *Loki) initStore() (_ services.Service, err error) {
 		}
 	}
 
-	t.Cfg.StorageConfig.BoltDBConfig.Ring = t.indexGatewayRing
+	t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Ring = t.indexGatewayRing
 	chunkStore, err := chunk_storage.NewStore(t.Cfg.StorageConfig.Config, t.Cfg.ChunkStoreConfig.StoreConfig, t.Cfg.SchemaConfig.SchemaConfig, t.overrides, t.clientMetrics, prometheus.DefaultRegisterer, nil, util_log.Logger)
 	if err != nil {
 		return
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -731,7 +731,7 @@ func (t *Loki) initCompactor() (services.Service, error) {
 
 func (t *Loki) initIndexGateway() (services.Service, error) {
 	t.Cfg.StorageConfig.BoltDBShipperConfig.Mode = shipper.ModeReadOnly
-	t.Cfg.IndexGateway.IndexGatewayRing.KVStore.MemberlistKV = t.MemberlistKV.GetMemberlistKV
+	t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Ring.KVStore.MemberlistKV = t.MemberlistKV.GetMemberlistKV
 	objectClient, err := chunk_storage.NewObjectClient(t.Cfg.StorageConfig.BoltDBShipperConfig.SharedStoreType, t.Cfg.StorageConfig.Config, t.clientMetrics)
 	if err != nil {
 		return nil, err
@@ -742,7 +742,7 @@ func (t *Loki) initIndexGateway() (services.Service, error) {
 		return nil, err
 	}
 
-	gateway, err := indexgateway.NewIndexGateway(t.Cfg.IndexGateway, util_log.Logger, prometheus.DefaultRegisterer, shipperIndexClient.(*shipper.Shipper), shipperIndexClient)
+	gateway, err := indexgateway.NewIndexGateway(t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig, util_log.Logger, prometheus.DefaultRegisterer, shipperIndexClient.(*shipper.Shipper), shipperIndexClient)
 	if err != nil {
 		return nil, err
 	}
@@ -753,9 +753,9 @@ func (t *Loki) initIndexGateway() (services.Service, error) {
 }
 
 func (t *Loki) initIndexGatewayRing() (_ services.Service, err error) {
-	t.Cfg.IndexGateway.IndexGatewayRing.KVStore.Multi.ConfigProvider = multiClientRuntimeConfigChannel(t.runtimeConfig)
-	t.Cfg.IndexGateway.IndexGatewayRing.KVStore.MemberlistKV = t.MemberlistKV.GetMemberlistKV
-	ringCfg := t.Cfg.IndexGateway.IndexGatewayRing.ToRingConfig(indexgateway.RingReplicationFactor)
+	t.Cfg.StorageConfig.BoltDBShipperConfig.Mode = shipper.ModeReadOnly
+	t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Ring.KVStore.MemberlistKV = t.MemberlistKV.GetMemberlistKV
+	ringCfg := t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Ring.ToRingConfig(indexgateway.RingReplicationFactor)
 	reg := prometheus.WrapRegistererWithPrefix("loki_", prometheus.DefaultRegisterer)
 	t.indexGatewayRing, err = ring.New(ringCfg, indexgateway.RingIdentifier, indexgateway.RingKey, util_log.Logger, reg)
 	if err != nil {
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -730,6 +730,7 @@ func (t *Loki) initCompactor() (services.Service, error) {
 
 func (t *Loki) initIndexGateway() (services.Service, error) {
 	t.Cfg.StorageConfig.BoltDBShipperConfig.Mode = shipper.ModeReadOnly
+	t.Cfg.IndexGateway.IndexGatewayRing.KVStore.MemberlistKV = t.MemberlistKV.GetMemberlistKV
 	objectClient, err := chunk_storage.NewObjectClient(t.Cfg.StorageConfig.BoltDBShipperConfig.SharedStoreType, t.Cfg.StorageConfig.Config, t.clientMetrics)
 	if err != nil {
 		return nil, err
@@ -744,6 +745,8 @@ func (t *Loki) initIndexGateway() (services.Service, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	t.Server.HTTP.Path("/indexgateway/ring").Methods("GET", "POST").Handler(gateway)
 	indexgatewaypb.RegisterIndexGatewayServer(t.Server.GRPC, gateway)
 	return gateway, nil
 }
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -732,7 +732,9 @@ func (t *Loki) initCompactor() (services.Service, error) {
 
 func (t *Loki) initIndexGateway() (services.Service, error) {
 	t.Cfg.StorageConfig.BoltDBShipperConfig.Mode = shipper.ModeReadOnly
-	t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Ring.KVStore.MemberlistKV = t.MemberlistKV.GetMemberlistKV
+	t.Cfg.IndexGateway.Ring.KVStore.MemberlistKV = t.MemberlistKV.GetMemberlistKV
+	t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Mode = t.Cfg.IndexGateway.Mode
+
 	objectClient, err := chunk_storage.NewObjectClient(t.Cfg.StorageConfig.BoltDBShipperConfig.SharedStoreType, t.Cfg.StorageConfig.Config, t.clientMetrics)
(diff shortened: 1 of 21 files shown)


pkg/storage/stores/shipper/shipper_index_client.go
```diff
@@ -54,18 +55,18 @@ type boltDBIndexClient interface {
...
-	Mode                     int                      `yaml:"-"`
-	IngesterDBRetainPeriod   time.Duration            `yaml:"-"`
+	ActiveIndexDirectory     string              `yaml:"active_index_directory"`
+	SharedStoreType          string              `yaml:"shared_store"`
+	SharedStoreKeyPrefix     string              `yaml:"shared_store_key_prefix"`
+	CacheLocation            string              `yaml:"cache_location"`
+	CacheTTL                 time.Duration       `yaml:"cache_ttl"`
+	ResyncInterval           time.Duration       `yaml:"resync_interval"`
+	QueryReadyNumDays        int                 `yaml:"query_ready_num_days"`
+	IndexGatewayClientConfig indexgateway.Config `yaml:"index_gateway_client"`
```
sandeepsukhani, line 65: I think it would be good to keep `IndexGateway` and `IndexGatewayClient` config separate for better understanding.
You can add the ring config to just `IndexGateway.Config` and handle set a ring reference in `IndexGatewayClient.Config`.

You can see it as an example in [SchedulerWorkerConfig](https://github.com/grafana/loki/blob/main/pkg/querier/worker_service.go#L31) which is used by queriers. And the ring config is only added to [Scheduer.Config](https://github.com/grafana/loki/blob/main/pkg/scheduler/scheduler.go#L125).

Does it make sense?

pkg/storage/stores/shipper/gateway_client.go
```diff
@@ -136,6 +170,56 @@ func (s *GatewayClient) doQueries(ctx context.Context, queries []chunk.IndexQuer
...
+		bufDescs, bufHosts, bufZones := ring.MakeBuffersForGet()
+
+		splittedHashed := strings.Split(q.HashValue, ":")
+		if len(splittedHashed) < 2 {
+			level.Error(util_log.Logger).Log("msg", "query hash value in incorrect format", "hash_value", q.HashValue)
+			continue
+		}
+
+		// tenant is the second parameter.
+		tenantID := splittedHashed[1]
```
sandeepsukhani, line 193: You can get the `tenantID` from the context like this here https://github.com/grafana/loki/blob/6cc503484cb1c12d255ef110a7337f72224e174d/pkg/storage/store.go#L214
And the `tenantID` for a batch of queries would always be the same.

pkg/storage/stores/shipper/gateway_client.go
```diff
@@ -136,6 +170,56 @@ func (s *GatewayClient) doQueries(ctx context.Context, queries []chunk.IndexQuer
...
+		key := util.TokenFor(tenantID, "" /* labels */)
+		rs, err := s.ring.Get(key, ring.WriteNoExtend, bufDescs, bufHosts, bufZones)
+		if err != nil {
+			level.Error(util_log.Logger).Log("msg", "failed to get index key", "error", err)
+			continue
+		}
+
+		_, err = rs.Do(ctx, 0 /* delay, TODO: what is it for? */, func(ctx context.Context, instanceDesc *ring.InstanceDesc) (interface{}, error) {
```
sandeepsukhani, line 201: `rs.Do` would query all the index gateways which is not required. We just want one successful response from one of the index gateway from the replica set.
We should try just one and if it fails try the next one until we have tried all of them.
We should do it with some randomization to avoid querying the same index gateway from a replica set for all the queries.

pkg/storage/stores/shipper/gateway_client.go
```diff
@@ -136,6 +170,56 @@ func (s *GatewayClient) doQueries(ctx context.Context, queries []chunk.IndexQuer
...
+			genericClient, err := s.pool.GetClientFor(instanceDesc.Addr)
+			if err != nil {
+				return nil, errors.Wrap(err, fmt.Sprintf("get client for instance %s", instanceDesc.Addr))
+			}
+
+			client := (genericClient.(indexgatewaypb.IndexGatewayClient))
+			// TODO: does this works as a hedging request?
+			if err := s.clientDoQueries(ctx, gatewayQueries, queryKeyQueryMap, callback, client); err != nil {
```
sandeepsukhani, line 209: We are doing all the `n` queries `n*n` times.

sandeepsukhani (changes requested):
I have done the first pass and have added some suggestions. Please let me know if you need any help addressing any of the feedback.

pkg/storage/stores/shipper/gateway_client.go
DylanGuedes: thanks for the comment!

double checking: are you sure we don't want to execute for all index gateway instances in parallel? Because this way it looks like a hedging request, which I believe is something we want.

Ex: Imagine we have 20 index gateway instances, and instances gateways 4, 5 and 6 are assigned to tenant ABC. Right now we have this problem of having specific index gateway instances taking really long to answer a query (1min+). If we sent the request in parallel to the three gateways (4, 5 and 6), we overcome this issue as long as the three instances aren't problematic at the same time. If we first sent to one index gateway (let's say Index Gateway 4), and only then if it fails we proceed to send to the next one (let's say Index Gateway 5), we will still have the same problem.

Does it make sense? :thinking:

pkg/storage/stores/shipper/gateway_client.go
DylanGuedes: great catch, thanks.

pkg/storage/stores/shipper/gateway_client.go
DylanGuedes: Great catch, thanks. I refactored it to not iterate over all queries anymore (by extracting the tenant from the ctx).

pkg/storage/chunk/storage/factory.go
```diff
@@ -167,6 +168,7 @@ func NewStore(
 	cfg Config,
 	storeCfg chunk.StoreConfig,
 	schemaCfg chunk.SchemaConfig,
+	indexGatewayRing ring.ReadRing,
```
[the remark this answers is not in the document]
DylanGuedes: I managed to make it work the way you suggested, thanks! WDYT?

pkg/storage/stores/shipper/gateway_client.go
DylanGuedes: Done. PTAL?

pkg/storage/stores/shipper/gateway_client.go
DylanGuedes: Done. PTAL?

pkg/storage/stores/shipper/shipper_index_client.go
DylanGuedes: Sorry, I tried it but I have a few questions:
- Are you suggesting to have every configurable YAML definition field in `indexgateway.Config` and to have a copy of them in this `shipper_index_client.go` file but without the YAML references?
- Or are you actually suggesting splitting the YAML definition into both, `indexgateway.Config` and `shipper_index_client.go` file? If so, I found it hard to correctly register the flags.

sandeepsukhani: Sorry for the confusion. I meant splitting the yaml for index gateway client and server. Client config would stay in `pkg/storage/stores/shipper/shipper_index_client.Config` while the index gateway server config(containing just ring config) would be with `pkg/storage/stores/shipper/indexgateway`.

Sorry I gave wrong reference above for [SchedulerWorkerConfig](https://github.com/grafana/loki/blob/main/pkg/querier/worker_service.go#L31). It should be [WorkerServiceConfig](https://github.com/grafana/loki/blob/a2dd73158621441fe2d83346e792913c6770c945/pkg/querier/worker_service.go#L31).

Please let me know if you need any help here.

pkg/storage/stores/shipper/gateway_client.go
sandeepsukhani: Hedging works by redoing only the queries which have not received a response within a specific duration. This could DOS index gateway by 3Xing the QPS(assuming RF=3).

pkg/storage/chunk/storage/factory.go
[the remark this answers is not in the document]
DylanGuedes: oops, forgot to push it. Pushed! Commit [here](https://github.com/grafana/loki/pull/5514/commits/a4f717c31e46ee0f498edc686977d71d5953e583).

pkg/storage/stores/shipper/gateway_client.go
DylanGuedes: Good call! Changed it to not send all requests all the time, WDYT?

pkg/storage/stores/shipper/shipper_index_client.go
DylanGuedes: Thanks for the clarification, it is clear now. I pushed a recent commit changing it, lmk if that's what you were thinking.

pkg/storage/stores/shipper/indexgateway/gateway.go
```diff
@@ -2,40 +2,213 @@ package indexgateway
...
 }
 
-func NewIndexGateway(indexQuerier IndexQuerier) *gateway {
-	g := &gateway{
+// stopping implements the Lifecycler interface and is one of the lifecycle hooks.
+//
+// Only invoked if the Index Gateway is in ring mode.
+func (g *Gateway) stopping(_ error) error {
+	level.Debug(util_log.Logger).Log("msg", "stopping index gateway")
+	g.shipper.Stop()
```
sandeepsukhani, line 139: I think `g.shipper` should be stopped last just for safety since the instance would still be active in the ring which would make the index gateways to keep getting queries.

sandeepsukhani (commented):
I have added some suggestions. The overall approach looks good to me!

pkg/storage/chunk/local/boltdb_index_client.go
```diff
@@ -39,6 +40,8 @@ const (
 // BoltDBConfig for a BoltDB index client.
 type BoltDBConfig struct {
 	Directory string `yaml:"directory"`
+
+	Ring ring.ReadRing
```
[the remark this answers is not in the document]
DylanGuedes: My bad, I think I added it here at first and then forgot to remove it/adapt the rest of the code. I pushed a commit that fixes it, lmk what you think.

pkg/storage/chunk/storage/factory.go
```diff
@@ -245,10 +246,10 @@ func NewStore(
 }
 
 // NewIndexClient makes a new index client of the desired type.
-func NewIndexClient(name string, cfg Config, schemaCfg chunk.SchemaConfig, limits StoreLimits, registerer prometheus.Registerer) (chunk.IndexClient, error) {
+func NewIndexClient(name string, cfg Config, schemaCfg chunk.SchemaConfig, indexGatewayRing ring.ReadRing, limits StoreLimits, registerer prometheus.Registerer) (chunk.IndexClient, error) {
```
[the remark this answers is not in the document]
DylanGuedes: fixed!

pkg/storage/stores/shipper/indexgateway/config.go
```diff
@@ -0,0 +1,68 @@
...
+	// section and the ingester configuration by default).
+	Ring loki_util.RingConfig `yaml:"ring,omitempty"`
+}
+
+// RegisterFlags register all IndexGatewayClientConfig flags and all the flags of its subconfigs but with a prefix (ex: shipper).
+func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	cfg.Mode = SimpleMode // default mode.
+
+	cfg.Ring.RegisterFlagsWithPrefix(prefix+".", "collectors/", f)
+	f.Var(cfg.Mode, prefix+".mode", "Mode to run the index gateway client. Supported modes are simple (default) and ring.")
```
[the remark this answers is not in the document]
DylanGuedes: Fixed!

pkg/storage/stores/shipper/gateway_client.go
```diff
@@ -27,32 +33,26 @@ const (
...
 
-// RegisterFlagsWithPrefix registers flags with prefix.
-func (cfg *IndexGatewayClientConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
-	cfg.GRPCClientConfig.RegisterFlagsWithPrefix(prefix, f)
+	conn       *grpc.ClientConn
+	grpcClient indexgatewaypb.IndexGatewayClient
 
-	f.StringVar(&cfg.Address, prefix+".server-address", "", "Hostname or IP of the Index Gateway gRPC server.")
-}
+	mode indexgateway.Mode
```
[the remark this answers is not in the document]
DylanGuedes: Good catch. Pushed a commit that changes it.

pkg/storage/store.go
```diff
@@ -425,15 +427,15 @@ func RegisterCustomIndexClients(cfg *Config, cm storage.ClientMetrics, registere
...
 
-	storage.RegisterIndexStore(shipper.BoltDBShipperType, func(limits storage.StoreLimits) (chunk.IndexClient, error) {
+	storage.RegisterIndexStore(shipper.BoltDBShipperType, func(limits storage.StoreLimits, indexGatewayRing ring.ReadRing) (chunk.IndexClient, error) {
 		if boltDBIndexClientWithShipper != nil {
 			return boltDBIndexClientWithShipper, nil
 		}
 
-		if cfg.BoltDBShipperConfig.Mode == shipper.ModeReadOnly && cfg.BoltDBShipperConfig.IndexGatewayClientConfig.Address != "" {
-			gateway, err := shipper.NewGatewayClient(cfg.BoltDBShipperConfig.IndexGatewayClientConfig, registerer)
+		if cfg.BoltDBShipperConfig.Mode == shipper.ModeReadOnly {
```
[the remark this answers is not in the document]
DylanGuedes: I think it is a good thing to add as a sanity check so I added it, but at the same time, don't you think it is also good to see things breaking if one of them isn't correctly set? (the address or the ring)

pkg/storage/stores/shipper/indexgateway/gateway.go
```diff
@@ -2,40 +2,213 @@ package indexgateway
...
 		indexQuerier: indexQuerier,
+		shipper:      indexClient,
+		cfg:          cfg,
+		log:          log,
 	}
+
 	g.Service = services.NewIdleService(nil, func(failureCase error) error {
 		g.indexQuerier.Stop()
 		return nil
 	})
```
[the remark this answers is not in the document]
DylanGuedes: my bad, I think it returned during a rebase or something. Fixed!

pkg/storage/stores/shipper/indexgateway/gateway.go
DylanGuedes: I added a defer to `g.shipper.Stop()`, do you think it is enough?

pkg/storage/store.go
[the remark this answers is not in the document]
sandeepsukhani: This condition actually decides whether we will be using the index gateway client or boltdb shipper client for index queries.
Having an incorrect config is a possibility with or without this check.
Does it make sense?

pkg/loki/modules.go
```diff
@@ -740,11 +745,30 @@ func (t *Loki) initIndexGateway() (services.Service, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	t.Server.HTTP.Path("/indexgateway/ring").Methods("GET", "POST").Handler(gateway)
 	indexgatewaypb.RegisterIndexGatewayServer(t.Server.GRPC, gateway)
 	return gateway, nil
 }
 
+func (t *Loki) initIndexGatewayRing() (_ services.Service, err error) {
```
sandeepsukhani, line 758: I think we should have a check for whether the index gateway mode is simple and just return if that is the case.

pkg/storage/stores/shipper/indexgateway/gateway.go
```diff
@@ -2,40 +2,208 @@ package indexgateway
...
+	level.Debug(util_log.Logger).Log("msg", "stopping index gateway")
+	defer g.shipper.Stop()
+	return services.StopManagerAndAwaitStopped(context.Background(), g.subservices)
+}
+
+// NewIndexGateway instantiates a new Index Gateway and start its services.
+//
+// In case it is configured to be in ring mode, a Basic Service wrapping the ring client is started.
+// Otherwise, it starts an Idle Service that doesn't have lifecycle hooks.
+func NewIndexGateway(cfg Config, log log.Logger, registerer prometheus.Registerer, indexClient chunk.IndexClient, indexQuerier IndexQuerier) (*Gateway, error) {
```
sandeepsukhani, line 147: Just a minor nit: Let us change the `indexClient chunk.IndexClient` to `querier Index.Querier` and follow the same wording at all the places.

Also let us move this factory function just below `Gateway` struct for readability.

sandeepsukhani (commented):
Just left few minor nits but the overall changes look good. Thanks for patiently taking care of all the feedback.

pkg/storage/stores/shipper/shipper_index_client.go
```diff
@@ -53,6 +56,45 @@ type boltDBIndexClient interface {
 	Stop()
 }
 
+// IndexGatewayClientConfig configures the Index Gateway client used to
+// communicate with the Index Gateway server.
+type IndexGatewayClientConfig struct {
+	// Mode sets in which mode the client will operate. It is actually defined at the
+	// index_gateway YAML section and reused here.
+	Mode indexgateway.Mode `yaml:"-"`
```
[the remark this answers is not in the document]
DylanGuedes: I tried it but tbh I didn't find a good way of passing the mode to the `gateway_client.go` without keeping it here in the Client config.

pkg/loki/modules.go
```diff
@@ -730,6 +732,9 @@ func (t *Loki) initCompactor() (services.Service, error) {
 
 func (t *Loki) initIndexGateway() (services.Service, error) {
 	t.Cfg.StorageConfig.BoltDBShipperConfig.Mode = shipper.ModeReadOnly
+	t.Cfg.IndexGateway.Ring.KVStore.MemberlistKV = t.MemberlistKV.GetMemberlistKV
+	t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Mode = t.Cfg.IndexGateway.Mode
```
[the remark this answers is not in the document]
DylanGuedes: good point. I moved it to `initStore`, WDYT?

pkg/storage/stores/shipper/indexgateway/gateway.go
DylanGuedes: done. WDYT?

pkg/loki/modules.go
sandeepsukhani: Index gateway is optional to run and it is possible that Loki would break setting up unexpected things.

pkg/loki/modules.go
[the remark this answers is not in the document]
sandeepsukhani: Looks good. Resolving the comment.

pkg/storage/stores/shipper/indexgateway/gateway.go
sandeepsukhani: Now there are 2 fields holding the same reference i.e `shipper` and `indexQuerier`. We should have just one :)

pkg/storage/stores/shipper/shipper_index_client.go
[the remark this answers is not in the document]
sandeepsukhani: I mean we should have this struct defined in `gateway_client.go` file instead of `shipper_index_client.go`. Sorry for the confusion.

pkg/storage/stores/shipper/shipper_index_client.go
```diff
@@ -95,6 +95,10 @@ func (i *IndexGatewayClientConfig) RegisterFlagsWithPrefix(prefix string, f *fla
 	f.StringVar(&i.Address, prefix+".server-address", "", "Hostname or IP of the Index Gateway gRPC server running in simple mode.")
 }
 
+func (i *IndexGatewayClientConfig) RegisterFlags(f *flag.FlagSet) {
```
sandeepsukhani, line 98: I can't find usage of this, can you please help me understand why it was added?

pkg/storage/stores/shipper/indexgateway/gateway.go
DylanGuedes: yep. I noticed that but you said with a lot of confidence so I was not sure :rofl:

pkg/storage/stores/shipper/shipper_index_client.go
DylanGuedes: If we don't implement a `RegisterFlags`, `IndexGatewayClientConfig` won't have support for default flags, which is used by our tests under `stores/shipper`. More specifically, the following block of code is responsible for requiring this:
```go
# pkg/storage/stores/shipper/gateway_client_test.go
	var cfg IndexGatewayClientConfig
	cfg.Mode = indexgateway.SimpleMode
	flagext.DefaultValues(&cfg)
	cfg.Address = storeAddress
```

pkg/loki/modules.go
DylanGuedes: Ooops I overlooked this one, I thought this was the `IndexGateway` service but it is actually the IndexGatewayRing. I added a guard clause as you suggested to not start this if mode != ring.

sandeepsukhani: Looks good. Thanks!

pkg/storage/stores/shipper/shipper_index_client.go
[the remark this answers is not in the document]
DylanGuedes: Got it! Pushed a comment fixing it.

PR #5515, opened 2022-03-02
Change usagestats expvars so they can be set twice

MichelHollands:

The expvars generated in functions like `Edition()`, `Target()` and others in the usagestats package can be called twice. Previously it would panic in that case.

**Checklist**
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/usagestats/stats.go
+++ b/pkg/usagestats/stats.go
@@ -168,7 +168,7 @@ func NewString(name string) *expvar.String {
 	return expvar.NewString(statsPrefix + name)
 }
 
-// Target sets the target name.
+// Target sets the target name. This can be set multiple times.
 func Target(target string) {
 	existing := expvar.Get(statsPrefix + targetKey)
 	if existing != nil {
--- a/pkg/usagestats/stats.go
+++ b/pkg/usagestats/stats.go
@@ -170,6 +170,14 @@ func NewString(name string) *expvar.String {
 
 // Target sets the target name.
 func Target(target string) {
+	existing := expvar.Get(statsPrefix + targetKey)
+	if existing != nil {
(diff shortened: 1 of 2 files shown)


cyriltovena:
This is great I think we should do this for all expvar

pkg/usagestats/stats.go
```diff
@@ -168,13 +168,29 @@ func NewString(name string) *expvar.String {
 	return expvar.NewString(statsPrefix + name)
 }
 
-// Target sets the target name.
+// Target sets the target name. This can be set multiple times.
 func Target(target string) {
+	existing := expvar.Get(statsPrefix + targetKey)
+	if existing != nil {
```
cyriltovena, line 174: Can we apply this to all expvar above: NewFloat, NewInt, NewString, Statistics, Counter  .....

pkg/usagestats/stats.go
```diff
@@ -154,27 +154,67 @@ func memstats() interface{} {
...
+		}
+		panic(fmt.Sprintf("%v is set to a non-string value", name))
+	}
 	return expvar.NewString(statsPrefix + name)
 }
 
-// Target sets the target name.
+// Target sets the target name. This can be set multiple times.
 func Target(target string) {
+	existing := expvar.Get(statsPrefix + targetKey)
```
cyriltovena, line 197: you don't need it on Target and Edition anymore since they rely on NewString

cyriltovena (approved):
LGTM

minus a small change required.

pkg/usagestats/stats.go
MichelHollands: D-oh. That code has been removed.

PR #5517, opened 2022-03-02
Support `SelectLogs` and `SelectSamples` in multi-tenant querier.

jeschkies:
This patch adds some functionality to the `MultiTenantQuerier` stub. `SelectLogs` and `SelectSamples` fan out into multiple calls, one for each tenant. The returned iterators are then merged back together and the proper tenant label is added.

Fixes grafana/loki-private#232

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/querier/multi_tenant_querier.go
+++ b/pkg/querier/multi_tenant_querier.go
@@ -80,9 +80,10 @@ func (q *MultiTenantQuerier) SelectSamples(ctx context.Context, params logql.Sel
 
 		iters[i] = NewTenantSampleIterator(id, iter)
 	}
-	return iter.NewMergeSampleIterator(ctx, iters), nil
+	return iter.NewSortSampleIterator(iters), nil
 }
 
+// TenantEntry Iterator wraps an entry iterator and adds the tenant label.
 type TenantEntryIterator struct {
 	iter.EntryIterator
 	tenantID string
@@ -106,6 +107,7 @@ func (i *TenantEntryIterator) Labels() string {
 	return builder.Labels().String()
 }
 
+// TenantEntry Iterator wraps a sample iterator and adds the tenant label.
 type TenantSampleIterator struct {
 	iter.SampleIterator
 	tenantID string
--- a/pkg/querier/multi_tenant_querier.go
+++ b/pkg/querier/multi_tenant_querier.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/go-kit/log"
+	"github.com/prometheus/prometheus/model/labels"
 	"github.com/weaveworks/common/user"
 
 	"github.com/grafana/loki/pkg/iter"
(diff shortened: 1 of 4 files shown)


pkg/querier/multi_tenant_querier_test.go
```diff
@@ -0,0 +1,44 @@
...
+	ctx := user.InjectOrgID(context.Background(), "1|2")
+	params := logql.SelectLogParams{QueryRequest: &logproto.QueryRequest{
+		Selector:  `{type="test"}`,
+		Direction: logproto.BACKWARD,
+		Limit:     0,
+		Shards:    nil,
+		Start:     time.Unix(0, 1),
+		End:       time.Unix(0, time.Now().UnixNano()),
+	}}
+	iter, err := multi_tenant.SelectLogs(ctx, params)
```
jeschkies, line 33: @JordanRushing it's still a little ugly but it's possible to test the querier this way.

pkg/querier/multi_tenant_querier.go
```diff
@@ -1,15 +1,132 @@
...
+		singleContext := user.InjectUserID(ctx, id)
+		iter, err := q.Querier.SelectLogs(singleContext, params)
+
+		if err != nil {
+			return nil, err
+		}
+
+		iters[i] = NewTenantEntryIterator(id, iter)
+	}
+	return iter.NewSortEntryIterator(iters, params.Direction), nil
```
jeschkies, line 57: I don't think we should deduplicate. What do you think, @cyriltovena

pkg/querier/multi_tenant_querier.go
```diff
@@ -1,15 +1,132 @@
...
+	lbls, _ := logql.ParseLabels(i.EntryIterator.Labels())
+	builder := labels.NewBuilder(lbls.WithoutLabels(defaultTenantLabel))
+
+	// Prefix label if it conflicts with the tenant label.
+	if lbls.Has(defaultTenantLabel) {
+		builder.Set(retainExistingPrefix+defaultTenantLabel, lbls.Get(defaultTenantLabel))
+	}
+	builder.Set(defaultTenantLabel, i.tenantID)
+
+	return builder.Labels().String()
```
jeschkies, line 107: This is code duplication. I'll wait for @JordanRushing's `Label()` implementation. Maybe there is a way to share the code.

pkg/querier/multi_tenant_querier.go
```diff
@@ -1,15 +1,132 @@
...
+	return iter.NewSortSampleIterator(iters), nil
+}
+
+// TenantEntry Iterator wraps an entry iterator and adds the tenant label.
+type TenantEntryIterator struct {
+	iter.EntryIterator
+	tenantID string
+}
+
+func NewTenantEntryIterator(id string, iter iter.EntryIterator) *TenantEntryIterator {
```
cyriltovena, line 92: I would prefer first parameter being the iter.

pkg/querier/multi_tenant_querier.go
```diff
@@ -1,15 +1,132 @@
...
+	return &TenantEntryIterator{EntryIterator: iter, tenantID: id}
+}
+
+func (i *TenantEntryIterator) Labels() string {
+	// TODO: cache manipulated labels
```
cyriltovena, line 97:
```suggestion
	// TODO: cache manipulated labels and add a benchmark.
```

cyriltovena (approved):
LGTM

We should work on the cache and use a benchmark to see where we should focus.

pkg/querier/multi_tenant_querier.go
cyriltovena: No that's good !

jeschkies:
> We should work on the cache and use a benchmark to see where we should focus.

I am tracking this in https://github.com/grafana/loki-private/issues/248.

PR #5520, opened 2022-03-03
fix broken index gateway benchmark due to missing limits

sandeepsukhani:
Index gateway benchmarks were failing due to limits not being initialized. This PR fixes it.
I have also cleaned up an unused method `QueryReadyIndexNumDays` from the limits interface.


dannykopping (approved):
LGTM

Issue #5494, opened 2022-02-28
Promtail: /metrics endpoint fails when multiple clients are used


rfratto:
**Describe the bug**
The `/metrics` endpoint returns an error when multiple clients are used:

```
$ curl localhost:80/metrics
An error has occurred while serving metrics:

collected metric "promtail_stream_lag_seconds" { label:<name:"filename" value:"/tmp/promtail.log" > label:<name:"host" value:"localhost:3100" > gauge:<value:0.287434 > } was collected before with the same name and label values
```

This behavior is visible against the latest Promtail commit (ead9444d7892e893ead34c1c5986dd25b2470cdb).

**To Reproduce**

1. Run Promtail with: `go run ./clients/cmd/promtail -config.file /tmp/promtail.yaml 2>&1 | tee /tmp/promtail.log` and the following config file (replacing the Loki URL as needed): 

```yaml
# /tmp/promtail.yaml
positions:
  filename: /tmp/positions.yaml

clients:
  - url: http://localhost:3100/loki/api/v1/push
    external_labels:
      client: a
  - url: http://localhost:3100/loki/api/v1/push
    external_labels:
      client: b

scrape_configs:
- job_name: self-logging
  static_configs:
  - targets: [localhost]
    labels:
      job: promtail
      __path__: /tmp/promtail.log
```

2. Curl the metrics endpoint of Promtail: `curl localhost:80/metrics`

**Expected behavior**
Metrics should be returned without error.

rfratto:
Note that this is only reproducible if Promtail is actively tailing new data from a log file. Initially, I used a static log file, and could no longer reproduce this issue once `/tmp/positions.yaml` was up to date with the file's contents.

chaudum:
@rfratto Can you verify the issue is solved with #5521 and close the issue in case it is?

rfratto:
I can no longer reproduce as of latest main (b3b96568ed9ee7faf3a8fee61579ff077cb9753b). Closing.

iridos:
Hi,

I think I am seeing the same issue when using a pipeline with a custom counter (called supervisor_total):
`
- job_name:  'cluster'
        pipeline_name: "slurm"
      stages:
      - metrics:
          supervisor_total:
            type: Counter
            description: "total count of slurm supervisor"
            config:
              match_all: True
              action: inc
`
Having two clients very much like described above, this leads to many log messages:

[...]

PR #5521, opened 2022-03-03
WIP: [promtail] Refactor promtail client metrics so that we can't have duplicate metrics collected for the lag metric.

cstyan:
Still some things I need to fix, but some quick reivews/sanity checks would be good.

So, this should solve https://github.com/grafana/loki/issues/5494

The cause of the issue is relatively simple, there was nothing in our code ensuring that for multiple promtail clients in the same binary that this metric would have a unique labelset per client. External labels are not applied to the metrics on `/metrics`. I've done a few things in this PR:
1. refactored the metrics initialization to be done from `main.go`, and the metrics struct is passed in to client constructors from there
2. added a name field to the client config, this will be added as a label to the stream_lag metric (which afaict is the only per-client metric at the moment, the others are for all clients of a type)
3. move the stream lag labels config option from the client config to the overall clients config section, the reason being that you can't register the same metric name with two different label sets


clients/pkg/promtail/client/client.go
```diff
@@ -318,23 +333,20 @@ func (c *client) sendBatch(tenantID string, batch *batch) {
...
 						}
 					}
 				}
 				if lblSet != nil {
 					// always set host
-					lblSet = lblSet.Merge(model.LabelSet{model.LabelName(HostLabel): model.LabelValue(c.cfg.URL.Host)})
+					lblSet[HostLabel] = c.cfg.URL.Host
+					// also set client name since if we have multiple promtail clients configured we will run into a
+					// duplicate metric collected with same labels error when trying to hit the /metrics endpoint
+					lblSet[ClientLabel] = c.name
```
cyriltovena, line 349: id or config id would have work better IMO.

cyriltovena (approved):
LGTM

kavirajk (approved):
LGTM

clients/pkg/promtail/client/client.go
cstyan: for the configuration field name?

Issue #5526, opened 2022-03-03
[bug] lokifrontend: invalid memory address or nil pointer dereference


liguozhong:
![image](https://user-images.githubusercontent.com/9583245/156565124-a5c0bdb4-97c9-4ee1-a275-f0092ca1ebfb.png)

level=debug ts=2022-03-03T09:44:40.881627171Z caller=grpc_logging.go:41 method=/grpc.health.v1.Health/Check duration=44.927µs msg="gRPC (success)"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x22f53aa]

goroutine 359 [running]:
github.com/grafana/loki/pkg/lokifrontend/frontend/v2.(*frontendSchedulerWorker).schedulerLoop(0xc000c10700, {0x3207540, 0xc000771300})

/pkg/lokifrontend/frontend/v2/frontend_scheduler_worker.go:314

PR #5527, opened 2022-03-03
[bugfix] fix nil pointer

liguozhong:
fix [5526](https://github.com/grafana/loki/issues/5526)


pkg/lokifrontend/frontend/v2/frontend_scheduler_worker.go
```diff
@@ -311,7 +311,7 @@ func (w *frontendSchedulerWorker) schedulerLoop(loop schedulerpb.SchedulerForFro
 				req.response <- &frontendv2pb.QueryResultRequest{
 					HttpResponse: &httpgrpc.HTTPResponse{
 						Code: http.StatusInternalServerError,
-						Body: []byte(err.Error()),
+						Body: []byte(resp.Error),
```
kavirajk, line 314: would be nice if could lock this behaviour in any of the tests but non-blocker for this PR.

PR #5417, opened 2022-02-17
[performance] logql：Experimental feature Optimize json parser in limited cases.

liguozhong:
**What this PR does / why we need it**:
Experimental feature Optimize json parser in limited cases.
Implement a high-performance simplejson parser（labels_hints_fast_parse）.
```shell
Benchmark_JsonParserFastParse/json/no_labels_hints-12         	  272154	      4296 ns/op
Benchmark_JsonParserFastParse/json/labels_hints_fast_parse-12 	 4061995	       258.0 ns/op
```
for example:
```go
logql ：
{functionName="feedflow"}  | json   | durationMs > 2003 
transform to
Optimize logql ：
{ functionName="feedflow"} | simplejson | durationMs>2003 | json 
```

In this way, reduce the performance loss caused by full json key parsing

**Which issue(s) this PR fixes**:
Fixes #5405>

**Special notes for your reviewer**:
"linefmt" case logql is more complicated, and another PR will handle it.
This feature is turned off by default, I will turn on this switch in our loki cluster to continuously test and trace this feature

loki.yaml config
```yaml
querier:
  query_timeout: 30m
  engine:
    logql_optimize_enable: true
  query_ingesters_within: 2h
```

**online environment**
I have completed the test in the online environment
tracing
image
+ performance = 50s-39/50s = 22%.  I will continue profile this json issue.

with this pr 39s
image
without this pr 50s
image
**Checklist**
- [x] Have you completed the test in the online environment
- [x] Documentation added
- [x] Tests updated
- [x] Add an entry in the `CHANGELOG.md` about the changes.

Issue #4328, opened 2021-09-14
Improve json parser performance


cyriltovena:
Investigate if we can replace the current jsoniter parser with https://github.com/buger/jsonparser

I already used it for parsing json payload on the push side and it way better for avoiding alloc.

cyriltovena:
When it's a metrics queries we could even use `requiredLabels` to only seek in json only key we need.

cyriltovena:
- https://github.com/francoispqt/gojay
- https://github.com/minio/simdjson-go
More to benchmark

jeschkies:
I was about to pitch simdjson. Do we have a dataset?

splitice:
Simdjson-go looks as good as any and far better than the current?

darxriggs:
still relevant

shaunmansell:
We've been trying Loki with services that log JSON and are seeing significant performance differences between queries with and without the JSON parser. We're seeing orders of difference in performance very similar to what's in https://github.com/grafana/loki/issues/5405.

Has there been any progress on testing the JSON parser switch in this issue?

wtchangdm:
Hi, is there any update? This seems helpful in terms of performance

PR #5538, opened 2022-03-04
Improve json parser performance

liguozhong:
Improve json parser performance, from 2913 ns/op to 1132 ns/op

Fixes [#4328]
reimplement pr: https://github.com/grafana/loki/pull/5417

New  Benchmark_Parser
```shell
Benchmark_Parser/json/no_labels_hints-12            250380        4772 ns/op
Benchmark_Parser/json/labels_hints-12               962606        1172 ns/op
Benchmark_Parser/jsonParser-not_json_line/no_labels_hints-12          81525914          13.54 ns/op
Benchmark_Parser/jsonParser-not_json_line/labels_hints-12              6865554         155.3 ns/op
```
Old  Benchmark_Parser
```shell
Benchmark_Parser/json/no_labels_hints-12            299535        3892 ns/op
Benchmark_Parser/json/labels_hints-12               355948        2913 ns/op
Benchmark_Parser/jsonParser-not_json_line/no_labels_hints-12          38245290          30.84 ns/op
Benchmark_Parser/jsonParser-not_json_line/labels_hints-12             39412749          31.16 ns/op
```

key change.
	use "github.com/buger/jsonparser" replace "github.com/json-iterator/go"
image

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/logql/log/parser.go
+++ b/pkg/logql/log/parser.go
@@ -41,7 +41,7 @@ type JSONParser struct {
 	lbs *LabelsBuilder
 
 	keys                  internedStringSet
-	bugerJsonParserEnable bool
+	bugerJSONParserEnable bool
 	jsonKeyCache          map[string][][]string
 }
 
@@ -50,7 +50,7 @@ func NewJSONParser() *JSONParser {
 	return &JSONParser{
 		buf:                   make([]byte, 0, 1024),
 		keys:                  internedStringSet{},
-		bugerJsonParserEnable: true,
+		bugerJSONParserEnable: true,
 		jsonKeyCache:          make(map[string][][]string),
 	}
 }
@@ -60,7 +60,7 @@ func (j *JSONParser) Process(line []byte, lbs *LabelsBuilder) ([]byte, bool) {
 		return line, true
 	}
 	j.lbs = lbs
-	if j.bugerJsonParserEnable {
+	if j.bugerJSONParserEnable {
 		requiredLabels := j.lbs.ParserLabelHints().RequiredLabels()
 		if len(requiredLabels) > 0 {
 			err := j.parseJSONKeyVal(line, requiredLabels)
@@ -189,11 +189,6 @@ func (j *JSONParser) parseJSONKeyVal(line []byte, requiredLabels []string) error
 		field = sanitizeLabelKey(field, true)
 		keys := make([]string, 0)
 		keys = append(keys, field)
-		//if strings.ContainsRune(field, jsonSpacer) {
-		//	keys = strings.Split(field, "_")
-		//} else {
-		//	keys = append(keys, field)
-		//}
 
 		cacheKeys, ok := j.jsonKeyCache[field]
 		if ok {
@@ -203,7 +198,7 @@ func (j *JSONParser) parseJSONKeyVal(line []byte, requiredLabels []string) error
 				if e != nil {
 					continue
 				}
-				err := j.SetJsonVal(field, v, t)
+				err := j.setJSONVal(field, v, t)
 				if err != nil {
 					return err
 				}
@@ -249,24 +244,26 @@ func (j *JSONParser) parseJSONKeyVal(line []byte, requiredLabels []string) error
 				jsonSpacerIndex++
 				continue
 			}
-			err := j.SetJsonVal(field, v, t)
+			err := j.setJSONVal(field, v, t)
 			if err != nil {
 				return err
 			}
-			cacheKeys, ok := j.jsonKeyCache[field]
-			if !ok {
-				j.jsonKeyCache[field] = make([][]string, 0)
-				cacheKeys = j.jsonKeyCache[field]
+			if len(keys) > 1 { //just cache key >1
+				cacheKeys, ok := j.jsonKeyCache[field]
+				if !ok {
+					j.jsonKeyCache[field] = make([][]string, 0)
+					cacheKeys = j.jsonKeyCache[field]
+				}
+				cacheKeys = append(cacheKeys, keys)
+				j.jsonKeyCache[field] = cacheKeys
 			}
-			cacheKeys = append(cacheKeys, keys)
-			j.jsonKeyCache[field] = cacheKeys
 			break
 		}
 	}
 	return nil
 }
 
-func (j *JSONParser) SetJsonVal(field string, v []byte, t jsonparser.ValueType) error {
+func (j *JSONParser) setJSONVal(field string, v []byte, t jsonparser.ValueType) error {
 	val, err := parseStringVal(v, t)
 	if err != nil {
 		return err
--- a/pkg/logql/log/parser.go
+++ b/pkg/logql/log/parser.go
@@ -22,8 +22,6 @@ import (
 const (
 	jsonSpacer      = '_'
 	duplicateSuffix = "_extracted"
-	trueString      = "true"
-	falseString     = "false"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


cyriltovena:
Can you share the allocations difference ?

liguozhong:
Can I delete these two cases
`go
{
			"escaped",
			[]byte(`{"counter":1,"foo":"foo\\\"bar", "price": {"_net_":5.56909}}`),
			labels.Labels{},
			labels.Labels{
				{Name: "counter", Value: "1"},
				{Name: "price__net_", Value: "5.56909"},
				{Name: "foo", Value: `foo\"bar`},
			},
		},
		{
			"utf8 error rune",
			[]byte(`{"counter":1,"foo":"�", "price": {"_net_":5.56909}}`),
			labels.Labels{},
			labels.Labels{
				{Name: "counter", Value: "1"},
				{Name: "price__net_", Value: "5.56909"},
				{Name: "foo", Value: ""},
			},
		},
`

splitice:
@liguozhong those look like tests of UTF8 and complex json syntax and look valid.

Whats the problem?

The first failure looks like double escaping.

That second test seems to imply that UTF8 error rune should not be propagated into the result. I'm not sure the reasoning behind this decision myself, perhaps someone else can answer?

pkg/logql/log/parser.go
```diff
@@ -38,154 +37,156 @@ type JSONParser struct {
...
-	// the first time we use the field as label key.
-	if len(prefix) == 0 {
-		key, ok := j.keys.Get(unsafeGetBytes(field), func() (string, bool) {
-			field = sanitizeLabelKey(field, true)
-			if !j.lbs.ParserLabelHints().ShouldExtract(field) {
-				return "", false
+		keys := make([]string, 0)
+		keys = append(keys, field)
+
+		cacheKeys, ok := j.jsonKeyCache[field]
```
jeschkies, line 107: Hm, this seems to be an orthogonal improvement to the parser. Could you factor out the cache in another PR? The benchmarks do not seem to be a fair comparison between the different parser implementations. How much improvements does the cache yield?

liguozhong: buger/jsonparser with cache : 1172 ns/op
buger/jsonparser without cache: 2223 ns/op
json-iterator:  2913 ns/op

liguozhong: cache is very effective for this json library, especially when the json key is deep.

but two problems that the test could not pass. give me some hints.
FAIL: Test_jsonParser_Parse/escaped
FAIL: Test_jsonParser_Parse/utf8_error_rune

liguozhong:
> @liguozhong those look like tests of UTF8 and complex json syntax and look valid.
> 
> Whats the problem?
> 
> The first failure looks like double escaping.
> 
> That second test seems to imply that UTF8 error rune should not be propagated into the result. I'm not sure the reasoning behind this decision myself, perhaps someone else can answer?

This json library is hard to implement these two tests

pkg/logql/log/parser.go
```diff
@@ -53,139 +52,110 @@ func (j *JSONParser) Process(line []byte, lbs *LabelsBuilder) ([]byte, bool) {
...
-		}
-		return falseString
-	default:
-		iter.Skip()
-		return ""
+func (j *JSONParser) setJSONVal(field string, v []byte) error {
+	if j.lbs.BaseHas(field) {
+		field = field + duplicateSuffix
 	}
+	j.lbs.Set(field, string(v))
```
splitice, line 153: If JSON rune handling was required a replace could be performed here. Either as a string replace, or probably better as a custom string builder

Perhaps something like:
```
var sb strings.Builder
for _, c := range string(v) {
  if c != '\uFFFD' {
    sb.WriteRune(c)
  }
}
return j.lbs.Set(field, sb.String())
```

splitice: Alternatively to support escaping in general (which currently you don't)

You could try calling `GetString(v)` this should also solve the failing escaping test.

splitice (commented):
I'm not a loki reviewer, but I did a read through and a check of the jsonparser documentation.

pkg/logql/log/parser.go
jeschkies: Thanks for the patience!

pkg/logql/log/parser.go
liguozhong: thanks, my cassandra keeps crashing this week and I'm busy with two issues this week (https://github.com/grafana/loki/issues/5667 and https://github.com/grafana/loki/issues/5595) , when my cassandra is stable, I'll come back to fix the json parser's with your tips

pkg/logql/log/parser.go
```diff
@@ -53,139 +52,110 @@ func (j *JSONParser) Process(line []byte, lbs *LabelsBuilder) ([]byte, bool) {
...
-	// we only care about object and values.
-	if nextType := it.WhatIsNext(); nextType != jsoniter.ObjectValue {
-		return errUnexpectedJSONObject
-	}
-	_ = it.ReadMapCB(j.parseMap(""))
-	if it.Error != nil && it.Error != io.EOF {
-		return it.Error
+func (j *JSONParser) handleJson(preKey string, line []byte) error {
+	err := jsonparser.ObjectEach(line, func(key []byte, val []byte, valueType jsonparser.ValueType, offset int) error {
+		jsonKey := preKey + sanitizeLabelKey(string(key), true)
```
cyriltovena, line 76: This line here, specially string(key) makes the whole PR back to where it was before.

I suggest you take a look at difference in memory allocations, we're trying to lower them and I don't think it's happening.

Use benstats to compare results.

cyriltovena: I suggest you also look at benchmark profile, go test -bench -memprofile mem.out to figure where the work needs to be done.

Again the whole point is to avoid allocation as much as possible.

cyriltovena: From a quick look, it seems that we need to create a pool for the slice of bytes that will contains the potential keys. Then try to work with slice of bytes only, avoiding the string(key) as much as possible.

jeschkies (changes requested):
I took a look at our production profiling and found that the JSON parser has the highest impact on allocations and not CPU. `buger/jsonparser` might be even worse. I hope I get to share the flamegraphs.

That means it is important to share the allocations. However, it made me wonder. What do you think about making `readValue` avoid any copies? It could read `[]byte` slice instead. I've traced the usage of the labels. At some point we call `String()` on the `LabelSet` but there is no need to copy the bytes before.

Sorry opening this rabbit hole. Just wanted to start a discussion.

EDIT: Just saw that Cyril mentions the same. Be aware though. `ReadStringAsSlice` [seems to make a copy](https://github.com/json-iterator/go/blob/e6b9536d3649bda3e8842bb7e4fab489d79a97ea/iter_str.go#L131) as well.

jeschkies:
I've just found that `burger/jsonparer` would [allow](https://github.com/buger/jsonparser#what-makes-it-so-fast) us to access the underlying bytes without allocations.

splitice:
Please don't forget that the underlying bytes may contain escape sequences.

pkg/logql/log/parser.go
jeschkies: What do you think about an unsafe to string here? This would avoid copying the data until `LabelResult.String()` is called. It's tricky though. The underlying `line` must not change and `LabelResult.String()` would have to unescape. In any case, we must start with some memory profiling.

cyriltovena:
The direction is good. The new library seems to allow what we want todo, avoid allocations.

@splitice is right about those tests, we added them as regression test, we had this issue in production with UTF8 characters.

I'm happy to have the whole code rework, but I think we should be cautious about allocation here.

PR #5539, opened 2022-03-04
Refactor LogQL syntax into its own package

### LINKED ISSUE #1854: Create a syntax package for LogQL
Looking at what's coming, I think we should in the future extract the syntax code from the LogQL package and move it to a logql/syntax package so it can be reused by other without having to embed the whole engine. There is a lot places where we need to just parse logql.

This will means also export each supported expression.

/cc @owen-d

cyriltovena:

This is a long overdue refactoring, placing all syntax related code into it's own package. I decided to use the package name `syntax` since this is more in-line with go code convetion (the regexp std package does the same)

There's more refactoring I want to do but I figure this is already a lot.

/cc @ssncferreira you might be affected by this.
/cc @ivanahuckova @torkelo this should make understand logql easier.

Fixes #1854 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


ssncferreira (commented):
LGTM 🎉 

> /cc @ssncferreira you might be affected by this.

Thanks for the heads up 👍 
This PR is also affected by the DownstreamEngine refactor PR: https://github.com/grafana/loki/pull/5546 as it also updates some logql files

jeschkies (approved):

PR #5413, opened 2022-02-16
Revert to working version of Azure client

MasslessParticle:
Due to `context canceled` errors originating in the new Azure SDK, we've decided to stabilize things by forking/patching the deprecated sdk client so we can have time to investigate the new client.

This PR does two things:

- Revert Azure client to deprecated, but working, version
- Add patches from https://github.com/Azure/azure-sdk-for-go/pull/16937

SDK fork exists here: https://github.com/MasslessParticle/azure-storage-blob-go

PR #5315, opened 2022-02-03
filters: use faster regexp package

bboreham:
This PR applies some optimisations I have submitted to the Go project:
* [regexp: allow patterns with no alternates to be one-pass](https://go-review.googlesource.com/c/go/+/353711)
* [regexp: speed up onepass prefix check](https://go-review.googlesource.com/c/go/+/354909)
* [regexp: handle prefix string with fold-case](https://go-review.googlesource.com/c/go/+/358756)
* [regexp: avoid copying each instruction executed](https://go-review.googlesource.com/c/go/+/355789)
* [regexp: allow prefix string anchored at beginning](https://go-review.googlesource.com/c/go/+/377294)

I have a [FOSDEM talk](https://fosdem.org/2022/schedule/event/go_finite_automata/) on Saturday 6th Feb about this.

I collected the optimisations into a new repo https://github.com/grafana/regexp; it exists purely to make PRs like this possible and can be retired once the changes land upstream.

I also added some new benchmark cases, because all the existing ones don't actually reach the regexp package after 'simplification'.
The benchmark is now running each case once for each iteration instead of a million times; this makes it easier to control the time and to read the averages.

```
name                                                           old time/op  new time/op   delta
_LineFilter/simplified_true_foo.*-4                            28.5ns ± 4%   27.8ns ± 4%     ~     (p=0.222 n=5+5)
_LineFilter/simplified_false_foo.*-4                           38.8ns ± 3%   39.0ns ± 5%     ~     (p=1.000 n=5+5)
_LineFilter/simplified_true_(?i)foo-4                           203ns ± 4%    200ns ± 3%     ~     (p=0.881 n=5+5)
_LineFilter/simplified_false_(?i)foo-4                          209ns ± 3%    211ns ± 5%     ~     (p=0.690 n=5+5)
_LineFilter/simplified_true_.*foo.*-4                          27.9ns ± 1%   28.1ns ± 2%     ~     (p=0.746 n=5+5)
_LineFilter/simplified_false_.*foo.*-4                         38.8ns ± 2%   39.0ns ± 3%     ~     (p=1.000 n=5+5)
_LineFilter/simplified_true_.*foo-4                            28.0ns ± 3%   27.9ns ± 2%     ~     (p=1.000 n=5+5)
_LineFilter/simplified_false_.*foo-4                           38.1ns ± 3%   39.0ns ± 2%     ~     (p=0.151 n=5+5)
[...]

PR #5542, opened 2022-03-04
regexp filter: use modified package with optimisations

bboreham:

This is a repeat of #5315, which was reverted in #5413.

See https://github.com/grafana/regexp/tree/speedup#readme

Includes the following changes proposed upstream:
* [regexp: allow patterns with no alternates to be one-pass](https://go-review.googlesource.com/c/go/+/353711)
* [regexp: speed up onepass prefix check](https://go-review.googlesource.com/c/go/+/354909)
* [regexp: handle prefix string with fold-case](https://go-review.googlesource.com/c/go/+/358756)
* [regexp: avoid copying each instruction executed](https://go-review.googlesource.com/c/go/+/355789)
* [regexp: allow prefix string anchored at beginning](https://go-review.googlesource.com/c/go/+/377294)

**Checklist**
- NA Documentation added - not user-visible.
- NA Tests updated - no functionality changes.
- NA Add an entry in the `CHANGELOG.md` about the changes - there is already an entry for #5315.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -104,7 +104,7 @@ require (
 require (
 	github.com/google/renameio/v2 v2.0.0
 	github.com/google/uuid v1.2.0
-	github.com/grafana/regexp v0.0.0-20220304095617-2e8d9baf4ac2
+	github.com/grafana/regexp v0.0.0-20220304100321-149c8afcd6cb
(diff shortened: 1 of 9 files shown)


cyriltovena (approved):
LGTM

PR #5544, opened 2022-03-04
Update vectorAggEvaluator to fail for expressions without grouping

ssncferreira:

Update the `vectorAggEvaluator` to return an error if the vector aggregator expression has a nil `Grouping` property, in order to avoid a nil pointer panic.

NA

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/logql/evaluator.go
+++ b/pkg/logql/evaluator.go
@@ -178,7 +178,7 @@ func (ev *DefaultEvaluator) StepEvaluator(
 					&logproto.SampleQueryRequest{
 						Start:    q.Start().Add(-rangExpr.Left.Interval).Add(-rangExpr.Left.Offset),
 						End:      q.End().Add(-rangExpr.Left.Offset),
-						Selector: e.String(), // intentionally send the the vector for reducing labels.
+						Selector: e.String(), // intentionally send the vector for reducing labels.
 						Shards:   q.Shards(),
 					},
 				})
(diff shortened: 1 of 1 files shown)


pkg/logql/evaluator.go
```diff
@@ -217,6 +217,9 @@ func vectorAggEvaluator(
 	expr *VectorAggregationExpr,
 	q Params,
 ) (StepEvaluator, error) {
+	if expr.Grouping == nil {
```
cyriltovena, line 220: Where does it panic, it's possible to have no groups for vector and range vector operation ?

ssncferreira: It panics if a `VectorAggregationExpr` is created with `Grouping` as nil. The `VectorAggregationExpr` should **always** have a Grouping, even in a simple expression without grouping. Therefore, this should never happen, but it is still possible that a `VectorAggregationExpr` is created without a `Grouping`, so this seems a good safe check.

I came up with this problem in the Split by range epic when creating a `VectorAggregationExpr` from a `RangeAggregationExpr`, because the former always needs a grouping but the latter has the grouping set to nil. The solution is here: https://github.com/grafana/loki/commit/0c291f1f3a3aed5ce7da77b84963ba34a745449f

cyriltovena (approved):
LGTM

PR #5547, opened 2022-03-04
updates clusterrole and clusterrolebinding apiVersions in promtail.sh

owen-d:
Our script currently won't work for anything v1.22+. This PR uses the new apiVersions which will. This change is also backwards compatible for any k8s versions v1.8 or newer: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#rbac-resources-v122


dannykopping (approved):
LGTM

PR #5561, opened 2022-03-07
Split by range on Instant queries POC

ssncferreira:

POC for split by range on instant queries.

**NOTE: do not merge to main until the correctness of this implementation is properly tested.**

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/logql/rangemapper.go
+++ b/pkg/logql/rangemapper.go
@@ -10,17 +10,17 @@ import (
 	util_log "github.com/grafana/loki/pkg/util/log"
 )
 
-func NewRangeVectorMapper(interval time.Duration) (RangeVectorMapper, error) {
-	if interval <= 0 {
-		return RangeVectorMapper{}, fmt.Errorf("Cannot create RangeVectorMapper with <=0 interval. Received %s", interval)
-	}
-	return RangeVectorMapper{
-		interval: interval,
-	}, nil
+type RangeVectorMapper struct {
+	splitByInterval time.Duration
 }
 
-type RangeVectorMapper struct {
-	interval time.Duration
+func NewRangeVectorMapper(interval time.Duration) (RangeVectorMapper, error) {
+	if interval <= 0 {
+		return RangeVectorMapper{}, fmt.Errorf("cannot create RangeVectorMapper with <=0 splitByInterval. Received %s", interval)
+	}
+	return RangeVectorMapper{
+		splitByInterval: interval,
+	}, nil
 }
 
 func (m RangeVectorMapper) Parse(query string) (bool, Expr, error) {
@@ -47,8 +47,6 @@ func (m RangeVectorMapper) Map(expr Expr) (Expr, error) {
 	switch e := expr.(type) {
 	case *VectorAggregationExpr:
 		return m.mapVectorAggregationExpr(e)
-	// case *LabelReplaceExpr:
-	// 	return m.mapLabelReplaceExpr(e, r)
 	case *RangeAggregationExpr:
 		return m.mapRangeAggregationExpr(e), nil
 	case *BinOpExpr:
@@ -76,8 +74,9 @@ func (m RangeVectorMapper) Map(expr Expr) (Expr, error) {
 	}
 }
 
-func (m RangeVectorMapper) mapSampleExpr(expr SampleExpr) SampleExpr {
-	var head *ConcatSampleExpr
+// getRangeInterval returns the interval in the range vector
+// Example: expression `count_over_time({app="foo"}[10m])` returns 10m
+func getRangeInterval(expr SampleExpr) time.Duration {
 	var rangeInterval time.Duration
 	expr.Walk(func(e interface{}) {
 		switch concrete := e.(type) {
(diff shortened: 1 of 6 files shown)


pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,278 @@
...
+		e.RHS = rhsSampleExpr
+		return e, nil
+	default:
+		return nil, errors.Errorf("unexpected expr type (%T) for ASTMapper type (%T) ", expr, m)
+	}
+}
+
+// getRangeInterval returns the interval in the range vector
+// Example: expression `count_over_time({app="foo"}[10m])` returns 10m
+func getRangeInterval(expr SampleExpr) time.Duration {
```
owen-d, line 79: What about 
```
count_over_time({app="foo"}[10m])
/
count_over_time({app="foo"}[30m])
```

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,278 @@
...
+
+	mapped, err := m.Map(parsed)
+	if err != nil {
+		return false, nil, err
+	}
+
+	return parsed.String() == mapped.String(), mapped, err
+}
+
+func (m RangeVectorMapper) Map(expr Expr) (Expr, error) {
```
owen-d, line 40: Do you think it's feasible to refactor the code to avoid duplicating a bunch of this logic?

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,278 @@
...
+	return rangeInterval
+}
+
+// mapSampleExpr transform expr in multiple subexpressions split by offset range interval
+// rangeInterval should be greater than m.splitByInterval, otherwise the resultant expression
+// will have an unnecessary aggregation operation
+func (m RangeVectorMapper) mapSampleExpr(expr SampleExpr, rangeInterval time.Duration) SampleExpr {
+	var head *ConcatSampleExpr
+
+	splitCount := int(rangeInterval / m.splitByInterval)
```
owen-d, line 96: What if `rangeInterval` isn't evenly divisible by `splitBy`?

pkg/logql/rangemapper_test.go
```diff
@@ -0,0 +1,225 @@
...
+					++ downstream<min(rate({app="foo"}[1m] offset 2m0s)), shard=<nil>>
+					++ downstream<min(rate({app="foo"}[1m] offset 1m0s)), shard=<nil>>
+					++ downstream<min(rate({app="foo"}[1m])), shard=<nil>>
+				)
+				/
+				sum(
+					downstream<count(rate({app="bar"}[1m] offset 9m0s)), shard=<nil>>
+					++ downstream<count(rate({app="bar"}[1m] offset 8m0s)), shard=<nil>>
+					++ downstream<count(rate({app="bar"}[1m] offset 7m0s)), shard=<nil>>
+					++ downstream<count(rate({app="bar"}[1m] offset 6m0s)), shard=<nil>>
```
owen-d, line 170: nit: we may need `offset` support in regular LogQL queries to support actually downstreaming these to queriers and thus as a prerequisite to this PR.

pkg/querier/queryrange/split_by_range.go
```diff
@@ -0,0 +1,120 @@
...
+type splitByRange struct {
+	logger log.Logger
+	next   queryrangebase.Handler
+	limits Limits
+
+	ng *logql.DownstreamEngine
+}
+
+// SplitByRangeMiddleware creates a new Middleware that splits log requests by the range interval.
+func SplitByRangeMiddleware(logger log.Logger, limits Limits, metrics *logql.ShardingMetrics) queryrangebase.Middleware {
```
owen-d, line 31: Can you explain why we need a specific middleware for this? I figured we could handle this in AST-mapping and use the existing sharded middleware to dispatch the resulting queries.

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,278 @@
...
+	return expr
+}
+
+func isSplittableByRange(expr SampleExpr) bool {
+	switch e := expr.(type) {
+	case *VectorAggregationExpr:
+		_, ok := SplittableVectorOp[e.Operation]
+		return ok && isSplittableByRange(e.Left)
+	case *BinOpExpr:
+		return true
```
owen-d, line 252: Why does this return `true` when the `Vector` and `Range` options may not?

owen-d (commented):
Cool PR; this is a good start! I've left some questions.

pkg/logql/rangemapper.go
ssncferreira: Each operator of binary expressions is evaluated separately, so when evaluating the first it returns `10m` and when evaluating the second it returns `10m`.

pkg/querier/queryrange/split_by_range.go
ssncferreira: The idea was to have two separate middlewares each one with its own responsibility (sharding or splitting). This way we can have a tripperware that does just sharding whereas another one that does just splitting. This would also facilitate the process of debugging/testing each middleware separately.
But the final idea is to have a tripperware that has both the sharding and the splitting middlewares 👍

kavirajk:
@ssncferreira can we close this in regards to #5662?

pkg/logql/rangemapper.go
ssncferreira: This function will be called in a recursive call, so we want to focus on splitting vector and range vector aggregation expressions. Therefore, a binary operation is considered splittable, because we want to check if their operators are equally splittable.

ssncferreira:
Closing this in favor of a newer version of the POC: https://github.com/grafana/loki/pull/5662

PR #5568, opened 2022-03-08
canary: Adds locking to prevent multiple concurrent invocations of `confirmMissing` from clobbering each other

### LINKED ISSUE #5128: Panic in canaries.
While checking on a cluster I found out a canary was panicking here is stack trace:

```
Connecting to loki at wss://../loki/api/v1/tail?query=%7Bstream%3D%22stdout%22%2Cpod%3D%22loki-canary-7c58648dc8-xwzdr%22%7D, querying for label 'pod' with value 'loki-canary-7c58648dc8-xwzdr'
error reading websocket, will retry in 10 seconds: websocket: close 1011 (internal server error): rpc error: code = Canceled desc = grpc: the client connection is closing
Connecting to loki at wss://l...t/loki/api/v1/tail?query=%7Bstream%3D%22stdout%22%2Cpod%3D%22loki-canary-7c58648dc8-xwzdr%22%7D, querying for label 'pod' with value 'loki-canary-7c58648dc8-xwzdr'
	/src/loki/pkg/canary/comparator/comparator.go:414 +0x3a5
created by github.com/grafana/loki/pkg/canary/comparator.(*Comparator).pruneEntries
	/src/loki/pkg/canary/comparator/comparator.go:446 +0x27d
github.com/grafana/loki/pkg/canary/comparator.(*Comparator).confirmMissing(0xc000230160, {0xc00055f0e0, 0xc00055f0e0, 0x1182f00})
goroutine 1661 [running]:
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa4209d]
panic: runtime error: invalid memory address or nil pointer dereference
```

afayngelerindbx:
nil pointer dereferences in this code are causing panics in our deployment. It seems that confirmMissing wasn't intended to be run concurrently. The locking in this PR prevents accessing `missingEntries` while it is also being set to slice of nils(comparator.go:L482).

Fixes #5128

### DIFF
--- a/pkg/canary/comparator/comparator_test.go
+++ b/pkg/canary/comparator/comparator_test.go
@@ -218,6 +218,45 @@ func TestEntryNeverReceived(t *testing.T) {
 
 }
 
+// Ensure that if confirmMissing calls pile up and run concurrently, this doesn't cause a panic
+func TestConcurrentConfirmMissing(t *testing.T) {
+	found := []time.Time{
+		time.Unix(0, 0),
+		time.UnixMilli(1),
+		time.UnixMilli(2),
+	}
+	mr := &mockReader{resp: found}
+
+	output := &bytes.Buffer{}
+
+	wait := 30 * time.Millisecond
+	maxWait := 30 * time.Millisecond
+
+	c := NewComparator(output, wait, maxWait, 50*time.Hour, 15*time.Minute, 4*time.Hour, 4*time.Hour, 0, 1*time.Minute, 0, 0, 1, make(chan time.Time), make(chan time.Time), mr, false)
+
+	for _, t := range found {
+		c.missingEntries = append(c.missingEntries, &t)
+	}
+
+	wg := sync.WaitGroup{}
+	for i := 0; i < 10; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			assert.NotPanics(t, func() { c.confirmMissing(time.UnixMilli(3)) })
+		}()
+	}
+	assert.Eventually(
+		t,
+		func() bool {
+			wg.Wait()
+			return true
+		},
+		time.Second,
+		10*time.Millisecond,
+	)
+}
+
 func TestPruneAckdEntires(t *testing.T) {
 	actual := &bytes.Buffer{}
 	wait := 30 * time.Millisecond
--- a/pkg/canary/comparator/comparator_test.go
+++ b/pkg/canary/comparator/comparator_test.go
@@ -235,7 +235,8 @@ func TestConcurrentConfirmMissing(t *testing.T) {
 	c := NewComparator(output, wait, maxWait, 50*time.Hour, 15*time.Minute, 4*time.Hour, 4*time.Hour, 0, 1*time.Minute, 0, 0, 1, make(chan time.Time), make(chan time.Time), mr, false)
 
 	for _, t := range found {
-		c.missingEntries = append(c.missingEntries, &t)
+		tCopy := t
+		c.missingEntries = append(c.missingEntries, &tCopy)
 	}
 
 	wg := sync.WaitGroup{}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 ## Main
-
+* [5568](https://github.com/grafana/loki/pull/5568) **afayngelerindbx**: Fix canary panics due to concurrent execution of `confirmMissing`
 * [5552](https://github.com/grafana/loki/pull/5552) **jiachengxu**: Loki mixin: add `DiskSpaceUtilizationPanel`
 * [5541](https://github.com/grafana/loki/pull/5541) **bboreham**: Queries: reject very deeply nested regexps which could crash Loki.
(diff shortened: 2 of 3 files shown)


chaudum (commented):
Generally, lgtm. Could we get this covered with a test, e.g. calling `confirmMissing` concurrently when also calling `pruneEntries`?

kavirajk (commented):
LGTM 👍 Thanks @afayngelerindbx.

Agree with @chaudum, can you please add a test? I will merge after that.

afayngelerindbx:
It's pretty tricky to actually ensure that these tests run with the right sequencing to cause all the panics. Let me know if the test needs to be something smarter.  Also, do I need to add anything to the CHANGELOG? This seems like too small a change for that.

kavirajk:
@afayngelerindbx. It can be simple test that locks this behavior :)

And for CHANGELOG. I don't think we need entry there as this is not user-visible change :)

afayngelerindbx:
I added a fairly simple test that accomplishes two things: 
1) With my comparator.go changes removed, it panics fairly consistently. Sometimes, it even causes the exact panic in the attached issue. With the locking added back in the test passes consistently.
2) With locking removed, the race detector(`go test -race`), shows racy data access for the new test. Once locking is added, the races are resolved.

Please, let me know whether you think this is sufficient test coverage for this change.

pkg/canary/comparator/comparator_test.go
```diff
@@ -218,6 +218,33 @@ func TestEntryNeverReceived(t *testing.T) {
...
+	maxWait := 30 * time.Millisecond
+
+	c := NewComparator(output, wait, maxWait, 50*time.Hour, 15*time.Minute, 4*time.Hour, 4*time.Hour, 0, 1*time.Minute, 0, 0, 1, make(chan time.Time), make(chan time.Time), mr, false)
+
+	for _, t := range found {
+		c.missingEntries = append(c.missingEntries, &t)
+	}
+
+	for i := 0; i < 10; i++ {
+		go func() {
```
kavirajk, line 242: better to add `sync.WaitGroup` to makesure all goroutines are completed by end of the test?

kavirajk:
thanks @afayngelerindbx looks good. Just added one minor suggestion to add waitGroup to the test.

Also looks like `license/cla` pending. Can you sign it? Then I will merge :)

pkg/canary/comparator/comparator_test.go
afayngelerindbx: good suggestion. adding

afayngelerindbx:
> thanks @afayngelerindbx looks good. Just added one minor suggestion to add waitGroup to the test.
> 
> Also looks like `license/cla` pending. Can you sign it? Then I will merge :)

I *think* I  signed. The CLA bot comment says so as well: https://github.com/grafana/loki/pull/5568#issuecomment-1061333145 I'm not sure why the check isn't passing. Any suggestions?

pkg/canary/comparator/comparator_test.go
kavirajk: Thanks for adding :)

1. can we make `wg.Done()` as `defer wg.Done()` at the beginning before `assert.NotPanic` is getting called? Because that way we can make sure `wg.Done()` is run even if that code panics.
2. Not familiar with `assert.Eventually`. But shouldn't just simple `wg.Wait()` after the for loop is sufficient here? Not sure if I miss anything there.

afayngelerindbx: I believe `assert.NotPanic` recovers from the panic and returns false. Adding the defer to make the code easier to reason about.

For 2, I would recommend keeping `Eventually` there. The expectation is that this test should return fairly quickly(within 1s) keeping a `wg.Done()` at the end will make the test suite time out(default 10m) if one of those goroutines hits a deadlock.

pkg/canary/comparator/comparator_test.go
```diff
@@ -218,6 +218,45 @@ func TestEntryNeverReceived(t *testing.T) {
...
+
+	output := &bytes.Buffer{}
+
+	wait := 30 * time.Millisecond
+	maxWait := 30 * time.Millisecond
+
+	c := NewComparator(output, wait, maxWait, 50*time.Hour, 15*time.Minute, 4*time.Hour, 4*time.Hour, 0, 1*time.Minute, 0, 0, 1, make(chan time.Time), make(chan time.Time), mr, false)
+
+	for _, t := range found {
+		c.missingEntries = append(c.missingEntries, &t)
```
chaudum, line 238: All items in the `missingEntries` slice will point to the same `time.Time` object, because `t` inside the range loop is reused for each iteration.

For this test it's irrelevant, but I guess it was not intended.

```suggestion
		tCopy := t
		c.missingEntries = append(c.missingEntries, &tCopy)
```

See https://go.dev/ref/spec#For_statements and https://stackoverflow.com/questions/48826460/using-pointers-in-a-for-loop

chaudum (commented):
Minor thing regarding the usage of the pointer inside a range loop.

pkg/canary/comparator/comparator_test.go
afayngelerindbx: Ah good catch. This gets me in golang all the time.

chaudum (approved):

kavirajk (approved):

Issue #5428, opened 2022-02-18
TSDB Index tracker


owen-d:
This is an issue for keeping track of the experimental work integrating a TSDB-based index into Loki as a boltdb replacement.

## TODO:
- [x] #5581

## Discarded Experiments:
- https://github.com/grafana/loki/pull/5594

bboreham:
Should this be closed as complete?  Or at least remove the word "experimental" from the description?

PR #5577, opened 2022-03-08
Tsdb/index interface

owen-d:
This PR introduces an `Index` interface and an implementation which uses TSDB to service it:

```go
type Interface interface {
	GetChunkRefs(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]ChunkRef, error)
	Series(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]Series, error)
	LabelNames(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]string, error)
	LabelValues(ctx context.Context, userID string, from, through model.Time, name string, matchers ...*labels.Matcher) ([]string, error)
}
```

This is subject to change, but lays a foundation for the interface we expect TSDB to satisfy. I'll be opening followup PRs to query multiple TSDBs, group TSDBs by tenant, instrument/cache the calls, etc.

### DIFF
--- a/pkg/storage/tsdb/index.go
+++ b/pkg/storage/tsdb/index.go
@@ -21,6 +21,6 @@ type ChunkRef struct {
 type Interface interface {
 	GetChunkRefs(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]ChunkRef, error)
 	Series(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]Series, error)
-	LabelNames(ctx context.Context, userID string, from, through model.Time) ([]string, error)
-	LabelValues(ctx context.Context, userID string, from, through model.Time, name string) ([]string, error)
+	LabelNames(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]string, error)
+	LabelValues(ctx context.Context, userID string, from, through model.Time, name string, matchers ...*labels.Matcher) ([]string, error)
 }
--- a/pkg/storage/tsdb/index.go
+++ b/pkg/storage/tsdb/index.go
@@ -1,0 +1,26 @@
+package tsdb
+
+import (
+	"context"
+
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/model/labels"
+)
+
+type Series struct {
+	Labels labels.Labels
+}
+
+type ChunkRef struct {
+	User        string
+	Fingerprint model.Fingerprint
+	Start, End  model.Time
+	Checksum    uint32
+}
+
+type Interface interface {
+	GetChunkRefs(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]ChunkRef, error)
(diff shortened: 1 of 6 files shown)


pkg/storage/tsdb/single_file_index.go
```diff
@@ -0,0 +1,107 @@
...
+
+	for p.Next() {
+		if err := i.reader.Series(p.At(), &ls, &chks); err != nil {
+			return nil, err
+		}
+
+		// TODO(owen-d): use logarithmic approach
+		for _, chk := range chks {
+
+			if chk.From() < through && chk.Through() >= from {
```
cstyan, line 89: does it make sense to use the same check in both places? or is there a reason we would want to ignore a chunk whose time range covers part of `from -> through` but also additional time.
```
if chk.From() > through || chk.Through() < from {
	continue
}
```

pkg/storage/tsdb/index.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+type ChunkRef struct {
+	User        string
+	Fingerprint model.Fingerprint
+	Start, End  model.Time
+	Checksum    uint32
+}
+
+type Interface interface {
```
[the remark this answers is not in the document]
owen-d: The package is actually `tsdb`, but I like the name of `Index` rather than `Interface` :+1:

pkg/storage/tsdb/single_file_index.go
cstyan: Got it. I misunderstood the function a bit when first reading it. This makes sense :+1:

pkg/storage/tsdb/single_file_index.go
```diff
@@ -0,0 +1,114 @@
...
+		// TODO(owen-d): use logarithmic approach
+		for _, chk := range chks {
+
+			// current chunk is outside the range of this request
+			if chk.From() > through || chk.Through() < from {
+				continue
+			}
+
+			if hash == 0 {
+				hash = ls.Hash()
```
cyriltovena, line 54: are we not saving the hash somewhere ?

pkg/storage/tsdb/single_file_index.go
```diff
@@ -0,0 +1,114 @@
...
+				End:         chk.Through(),
+				Checksum:    chk.Checksum,
+			})
+		}
+	}
+	return res, p.Err()
+}
+
+func (i *TSDBIndex) Series(_ context.Context, _ string, from, through model.Time, matchers ...*labels.Matcher) ([]Series, error) {
+	p, err := PostingsForMatchers(i.reader, matchers...)
```
cyriltovena, line 70: This is going to be so fast compare to now.

pkg/storage/tsdb/single_file_index.go
owen-d: Not yet, no. I was looking at this yesterday, but I haven't set it up yet.

cyriltovena (approved):
LGTM

You should add a not or track that this doesn't yet support sharding but should.

PR #5580, opened 2022-03-08
build a new batch while starting to process a bucket

sandeepsukhani:
I noticed that while iterating through buckets for building index batch for writes, I was not clearing the batch while processing the next bucket. The batch gets recreated only when it gets full. This causes us to write duplicate index in other users index files.
This PR fixes the issue by creating a new batch while staring to process a bucket.

### DIFF
--- a/pkg/storage/stores/shipper/compactor/table.go
+++ b/pkg/storage/stores/shipper/compactor/table.go
@@ -430,10 +430,9 @@ func readFile(logger log.Logger, path string, writeBatch func(userID string, bat
 		}
 	}()
 
-	batch := make([]indexEntry, 0, batchSize)
-
 	return db.View(func(tx *bbolt.Tx) error {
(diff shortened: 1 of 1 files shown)


pkg/storage/stores/shipper/compactor/table.go
```diff
@@ -430,10 +430,9 @@ func readFile(logger log.Logger, path string, writeBatch func(userID string, bat
 		}
 	}()
 
-	batch := make([]indexEntry, 0, batchSize)
-
 	return db.View(func(tx *bbolt.Tx) error {
 		return tx.ForEach(func(name []byte, b *bbolt.Bucket) error {
+			batch := make([]indexEntry, 0, batchSize)
```
cyriltovena, line 435: why not batch = batch[:0]

cyriltovena (approved):

PR #5585, opened 2022-03-09
fix possible data loss issue due to failures in uploading compacted user index

sandeepsukhani:
We put the reference to the common index set in a dictionary that contains a reference to all the user index sets.
We iterate this dictionary to call `indexSet.done()` which would upload the compacted db from that index set and clean up the source file from the object store. For user index sets, the index files from common index sets are also a source for compaction.

We could lose data in a case where we cleaned up the common index set first and we failed to upload newly compacted files for user index sets. This PR fixes it by always calling `done()` on the common index set at the end.

### DIFF
--- a/pkg/storage/stores/shipper/compactor/table.go
+++ b/pkg/storage/stores/shipper/compactor/table.go
@@ -223,9 +223,22 @@ func (t *table) done() error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


slim-bean (approved):
LGTM

dannykopping (approved):
LGTM

PR #5587, opened 2022-03-09
Compactor improvements

sandeepsukhani:
It includes two improvements:
1. After we are done compacting the files, the operation of upload and cleanup happens for one tenant at a time. When we have a lot of tenants this takes a lot of time. This PR improves it by concurrently doing upload and cleanup for 10 tenants at a time.
2. While doing compaction, we create a new slice for the index batch when it either gets full or we move to the next tenant bucket. This PR changes the code to reuse the same slice for batching the index.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/shipper/compactor/table.go
+++ b/pkg/storage/stores/shipper/compactor/table.go
@@ -450,9 +450,11 @@ func readFile(logger log.Logger, path string, writeBatch func(userID string, bat
 		}
 	}()
 
+	batch := make([]indexEntry, 0, batchSize)
+
 	return db.View(func(tx *bbolt.Tx) error {
 		return tx.ForEach(func(name []byte, b *bbolt.Bucket) error {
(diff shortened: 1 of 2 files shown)


pkg/storage/stores/shipper/compactor/table.go
```diff
@@ -58,7 +58,8 @@ import (
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 const (
-	uploaderName = "compactor"
+	uploaderName               = "compactor"
+	uploadIndexSetsConcurrency = 10
```
dannykopping, line 62: Do you see any value in making this configurable? I wonder if folks with large numbers of tenants (i.e. us) would benefit from tuning this value.

dannykopping: Works for me :+1: thanks for your thoughts.

dannykopping (approved):

PR #5590, opened 2022-03-09
more tsdb benchmarks + updates script to work with incompatible branches

owen-d:

### DIFF
--- a/tools/tsdb/tsdb-map/main_test.go
+++ b/tools/tsdb/tsdb-map/main_test.go
@@ -1,7 +1,9 @@
 package main
 
 import (
+	"context"
 	"fmt"
+	"math"
 	"math/rand"
 	"os"
 	"testing"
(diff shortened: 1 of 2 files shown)


kavirajk (approved):
LGTM :+1: . One minor suggestion.

tools/tsdb/tsdb-map/main_test.go
```diff
@@ -53,3 +60,28 @@ func BenchmarkQuery(b *testing.B) {
...
+func BenchmarkQuery_GetChunkRefs(b *testing.B) {
+	for _, bm := range cases {
+		indexPath := os.Getenv("LOKI_TSDB_PATH")
+		if indexPath == "" {
+			return
+		}
+
+		reader, err := index.NewFileReader(indexPath)
+		if err != nil {
+			panic(err)
```
kavirajk, line 75: Can we do `b.Fail()` instead? (also other places)

PR #5594, opened 2022-03-09
index wrapper for bounding parallelism

owen-d:
This adds a utility for bounding parallelism for underlying index implementations. 


owen-d:
After talking with @sandeepsukhani, I'm going to close this one. He raised the good point that finding the right parallelism factor depends on a lot of things (cluster size, resources, QPS, component size) and this may be more hassle than benefit.

PR #5598, opened 2022-03-10
Tsdb/multi idx

owen-d:
This PR introduces an index implementation which wraps multiple underlying indices (i.e. multiple TSDB files). In the course of implementing it, I also adjusted the `Series` response type to include the `model.Fingerprint` for use in deduping. I think we can revive the old https://github.com/grafana/loki/pull/5591 PR to make this easier as well.


slim-bean (approved):
amazing

PR #5600, opened 2022-03-10
Tsdb/sharding

owen-d:
This PR adds support for shard-aware index querying, specifically to the `GetChunkRefs` and `Series` methods. This is an initial PR and is relatively naive, but implements the required functionality and leaves us open to improving it in the future.

closes https://github.com/grafana/loki/issues/5581


slim-bean (approved):
LTM

PR #5616, opened 2022-03-14
General: print fatal stacktraces to stderr for improved readability

dannykopping:

Prints fatal stacktraces to `stderr` for improved readability.
Currently our fatal error messages are logged like this:

```
level=error ts=2022-03-14T09:06:10.700672214Z caller=log.go:103 msg="error running loki" err="failed services\ngithub.com/grafana/loki/pkg/loki.(*Loki).Run\n\t/loki/pkg/loki/loki.go:420\nmain.main\n\t/loki/cmd/loki/main.go:108\nruntime.main\n\t/home/danny/sdk/go1.17.4/src/runtime/proc.go:255\nruntime.goexit\n\t/home/danny/sdk/go1.17.4/src/runtime/asm_amd64.s:1581"
```

This is pretty difficult to read.
This log line will be maintained, but we will additionally log the error message to `stderr`:

```
failed services
github.com/grafana/loki/pkg/loki.(*Loki).Run
        /loki/pkg/loki/loki.go:420
main.main
        /loki/cmd/loki/main.go:108
runtime.main
        /home/danny/sdk/go1.17.4/src/runtime/proc.go:255
runtime.goexit
        /home/danny/sdk/go1.17.4/src/runtime/asm_amd64.s:1581
```

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/util/log/log.go
+++ b/pkg/util/log/log.go
@@ -97,7 +97,10 @@ func CheckFatal(location string, err error, logger log.Logger) {
 		logger = log.With(logger, "msg", "error "+location)
 	}
 	// %+v gets the stack trace from errors using github.com/pkg/errors
-	logger.Log("err", fmt.Sprintf("%+v", err))
+	errStr := fmt.Sprintf("%+v", err)
+	fmt.Fprintln(os.Stderr, errStr)
+
+	logger.Log("err", errStr)
 	os.Exit(1)
 }
 


cyriltovena (approved):
LGTM

pkg/util/log/log.go
```diff
@@ -97,7 +97,10 @@ func CheckFatal(location string, err error, logger log.Logger) {
 		logger = log.With(logger, "msg", "error "+location)
 	}
 	// %+v gets the stack trace from errors using github.com/pkg/errors
-	logger.Log("err", fmt.Sprintf("%+v", err))
+	errStr := fmt.Sprintf("%+v", err)
+	fmt.Fprintln(os.Stderr, errStr)
```
kavirajk, line 101: Q: wondering putting same error in log and in stderr can cause scraping it twice? based on container runtime?

My understanding is (for example in OCI) both stderr and stdout of an application goes into pod logs which will be scraped by promtail.

dannykopping: What is your concern, specifically?

The point here is that when we log the error using our standard logger, it replaces all newlines and tabs with `\n` and `\t` respectively, so that it conforms with the `logfmt` format.

kavirajk: Yea I understand that from the change. :+1: 

My concern is before this change, ony `logfmt` format of that error will be scraped and send to Loki, but now both "logfmt" and "plain" format of the error will be send to Loki.

dannykopping: Sure, but what is your actual concern with that? That seems like a perfectly reasonable tradeoff for improved readability.

kavirajk: Not a big concern. Just wanted to clarify. that's all.

PR #5620, opened 2022-03-14
Test multi-tenant context propagation.

jeschkies:
The original change injected the user id when the org id was required to fetch all data.

Fixes grafana/loki-private#230

The changes can be tested as follows
1. Change working directory to `tools/dev/loki-boltdb-storage-s3`.
2. Call `./compose-up.sh` to boot the cluster.
3. Query the logs with
    ```
   curl -v -X GET -G "http://localhost:8007/loki/api/v1/query_range" -H "X-Scope-OrgID: 1|2"   --data-urlencode 'query={compose_service=~"log-gen.*"}' | jq '.data.result[].stream'
    ```
    It will should two streams with `__tenant_id: 1` and `__tenant_id_: 2` as part of the labels.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/tools/dev/loki-boltdb-storage-s3/config/loki.yaml
+++ b/tools/dev/loki-boltdb-storage-s3/config/loki.yaml
@@ -92,7 +92,7 @@ querier:
     multi_tenant_queries_enabled: true
 query_range:
     align_queries_with_step: true
-    cache_results: false 
+    cache_results: true
     max_retries: 5
     parallelise_shardable_queries: true
     results_cache:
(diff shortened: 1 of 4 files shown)


tools/dev/loki-boltdb-storage-s3/config/loki.yaml
```diff
@@ -1,4 +1,4 @@
-auth_enabled: false
+auth_enabled: true
```
jeschkies, line 1: I can change it back. However, I don't think it this example wasn't used much as it wasn't working due to changes in the Minio image.

cyriltovena (approved):
LGTM

Issue #1779, opened 2020-03-06
Interval parameter for log queries


slim-bean:
Recently I created #1773 and after merging realized that Grafana defaults to sending a `step` with log queries.  I would consider this a bug but be that as it may keeping #1773 would result in some pretty nasty unexpected behavior when using Grafana against Loki (even if we fixed Grafana old versions would still be sending this parameter)

#1773 was reverted by #1777 however I would still like to add this feature and I can think of a couple ways to do it:

* Use a new parameter, I am suggesting `log_step`
* Use a new header such as: `Accept: application/json;version=2`

Some more thoughts:

I think the `step` functionality is worth having in Loki.  While Loki is a logs database, more specifically it is a string database extended from a metric database (Cortex) and I think it's a valid use case to store string style metrics in Loki (e.g. GPS coordinates, or a more complex json doc of metrics), as such the `step` parameter is of great value at query time to allow range queries which return sane result sizes over different time periods.

At this point it's much too dangerous to just start using the "step" for log style queries, we use it correctly for metric style queries but previously it was just ignored by Loki for log queries.

I think of the two approaches mentioned above, I prefer the new parameter `log_step` for a few reasons, first it makes URL's easy to write and send (setting a header is tedious), second within our code there were also assumptions made about the use of step only being for metric queries and as a result it has specific rules applied to it which don't apply to log queries.  #1773 handled this by parsing the query very early to determine the type before applying these rules to the incoming step parameter.  Using a new parameter would make it more obvious which rules are applied and why within the Loki codebase.

Looking for thoughts/feedback.

slim-bean:
I've been waffling about the idea of supporting both actually, that way if clients want to be able to issue queries using `step` as intended for both log and metric queries they can do so by setting the header correctly.  I worry that supporting both would be confusing however

rfratto:
Personally between the two, I think the header is the right way to go. In general, I agree that setting a header is a little tedious but tooling helps with making this easy. Arguably, I'd say URLs are already not easy to write if your query is sufficiently complicated. 

Another option might be to introduce the functionality as another API endpoint, something like `/loki/api/v1/query_logs` (as opposed to query_range). We could also introduce `/loki/api/v2/query_range`, although I'm less of a fan here, even though your initial PR was technically backwards incompatible.

slim-bean:
After some discussions and thinking about this for a bit.  I'm moving farther away from using `step` or anything similar to `step`.

[...]

PR #5622, opened 2022-03-14
Preserve interval parameter when splitting queries by time

### LINKED ISSUE #5613: Bug: query_range interval parameter broken when query contains filter expressions
**Describe the bug**
If the LogQL query contains a filter expression (`|=`, `|~`, `!=`, ...) calls to the `query_range` API return the same result, regardless of whether the `interval` parameter set or not and what value it has. E.g. no interval, interval=60s, interval=60000s.

**To Reproduce**
1. Call `query_range` API with a high volume query (several lines per second) and `interval=60s`
2. Add filter expression that matches every line ` |~ ".*"` and repeat.
3. Observe that returned log lines are not equal

**Expected behavior**
Interval should work as documented (i.e., return approximately one line per interval), regardless of the query.

**Environment:**
 - Infrastructure: bare-metal
 - Deployment tool: systemd service
```
$ loki-linux-amd64 -version
loki, version 2.4.2 (branch: HEAD, revision: 525040a32)
  build user:       root@5d9e7a4c92e6
  build date:       2022-01-12T16:48:53Z
  go version:       go1.16.2
  platform:         linux/amd64
```

As seen here in these examples, without the filter expression the difference in timestamps is ~74 seconds. Once the filter is added to the query the returned lines are only 4000ns apart. Removing the interval parameter with the filter in place yields the same result.
```
$ curl -s -G -H 'X-Scope-OrgID: <tenant>' \
 --data-urlencode 'start=1646843100000000000' \
 --data-urlencode 'end=1646903100000000000' \
 --data-urlencode 'interval=60s' \
 --data-urlencode 'limit=5000' \
 --data-urlencode 'query=<query>' \
 'http://127.0.0.1:3117/loki/api/v1/query_range' | jq '.data.result[0].values [0,1]'
[
  "1646903089664581496",
  "<result 0>"
]
[
  "1646903015429412821",
  "<result 1>"
]

$ curl -s -G -H 'X-Scope-OrgID: <tenant>' \
 --data-urlencode 'start=1646843100000000000' \
 --data-urlencode 'end=1646903100000000000' \
 --data-urlencode 'interval=60s' \
 --data-urlencode 'limit=5000' \
 --data-urlencode 'query=<query> |~ ".*"' \
 'http://127.0.0.1:3117/loki/api/v1/query_range' | jq '.data.result[0].values [0,1]'
[
  "1646903089664581496",
  "<result 0>"
]
[
  "1646903089664577339",
  "<result 1>"
]

$ curl -s -G -H 'X-Scope-OrgID: tnguyen' \
> --data-urlencode 'start=1646843100000000000' \
> --data-urlencode 'end=1646903100000000000' \
> --data-urlencode 'limit=5000' \
> --data-urlencode 'query=<query> |~ ".*"' \
> 'http://127.0.0.1:3117/loki/api/v1/query_range' | jq '.data.result[0].values [0,1]'
[
  "1646903089664581496",
  "<result 0>"
]
[
  "1646903089664577339",
  "<result 1>"
]
```

chaudum:

This PR fixes the incorrect behaviour where the `interval` parameter from a query range request is ignored in case the query is internally split by time, and therefore returning much more results than expected.

Fixes #5613

**Checklist**
- [ ] Documentation added
- [x] Tests updated


chaudum:
Adding @slim-bean as reviewer, since the `interval` parameter was initially added by him. The parameter is experimental and the use-cases are questionable, however [users seem to use it](https://github.com/grafana/loki/issues/1779#issuecomment-1066061697).
@cyriltovena was questioning whether we should drop the parameter entirely.

sandeepsukhani (approved):
LGTM

cyriltovena:
Just need to resolve conflict I enable auto merge.

PR #5626, opened 2022-03-15
Support limits in multi-tenant queries.

jeschkies:
This changes adds support for the the following limits in multi-tenant queries:
- `MaxQuerySeries`
- `QuerySplitDuration`
- `MaxEntriesLimitPerQuery`
- `MinShardingLookback`
- `MaxQueryLength`
- `MaxQueryParallelism`
- `MaxQueriersPerUser`
- `MaxCacheFreshness`

It always uses the more restricted limit. That means the smallest `MaxQuerySeries` but biggest `QuerySplitDuration`. The split duration is special. Zero means disabled.

Also, the global default resolver is the multi tenant resolver when multi-tenant queries are enabled. This will result HTTP 400 errors in case the user sends multiple tenants to endpoints that do not support it.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


pkg/logql/downstream.go
```diff
@@ -71,7 +72,8 @@ func (ng *DownstreamEngine) Query(p Params, mapped syntax.Expr) Query {
 		parse: func(_ context.Context, _ string) (syntax.Expr, error) {
 			return mapped, nil
 		},
-		limits: ng.limits,
+		limits:   ng.limits,
+		resolver: tenant.NewMultiResolver(),
```
jeschkies, line 76: We cannot simply use `tenant.TenantIDs` because this will use the single tenant resolver unless we override it. This could be fine but I'm wondering if it could break anything.

pkg/querier/queryrange/split_by_interval.go
```diff
@@ -161,12 +164,12 @@ func (h *splitByInterval) loop(ctx context.Context, ch <-chan *lokiResult, next
...
 
 func (h *splitByInterval) Do(ctx context.Context, r queryrangebase.Request) (queryrangebase.Response, error) {
-	userid, err := tenant.TenantID(ctx)
+	tenantIDs, err := h.resolver.TenantIDs(ctx)
 	if err != nil {
 		return nil, httpgrpc.Errorf(http.StatusBadRequest, err.Error())
 	}
 
-	interval := h.limits.QuerySplitDuration(userid)
+	interval := validation.MaxDurationOrZeroPerTenant(tenantIDs, h.limits.QuerySplitDuration)
```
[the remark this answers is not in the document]
simonswine: I agree it is really hard to determine what the user intends to happen in this case here. 

I think going for the largest duration max sense to me. We probably should add at least some context in a content why we went with largest instead  of the smallest.

pkg/logql/downstream.go
simonswine: Io order to enable any multi tenant features the global resolver should be swapped to a multi tenanted one. See what Cortex is doing there:

https://github.com/cortexproject/cortex/blob/45f043f0d121a04a4da0d60e7791562f93c6aa91/pkg/cortex/cortex.go#L356-L360

If the resolver is not swapped (aka tenant federation disabled), Loki should just treat `x|y|z` as a single tenant named like that. 

I think we should consider enabling tenant federation default.

simonswine (commented):
Thanks for working on this. Figuring out limits relating to multi tenant federation certainly feels quite arbitrary. 

I think we can take a pragmatic stance and accept no ability for defining custom limits for federated tenant for now.

pkg/logql/downstream.go
jeschkies: Ok, so the multi tenant resolver will return an error when only one tenant ID is expected but multiple are set (https://github.com/grafana/loki/blob/main/pkg/tenant/resolver.go#L117). This is exactly the behaviour we want. E.g. if a users sets multiple tenants during ingestion we want Loki to return an error. Ideally `ErrTooManyOrgIDs` is treated as a user error though.

JordanRushing:
I've authored some tests for the new multi-tenant-limits-related util functions as well as a quick fix for `TestMaxDurationOrZeroPerTenant`: 

https://github.com/JordanRushing/loki/compare/karsten/multi-tenant-max-queries%5E%5E...JordanRushing:karsten/multi-tenant-max-queries

I'm happy to push these commits to this PR or to PR the underlying fork; please let me know!

JordanRushing (approved):
LGTM!

PR #5628, opened 2022-03-15
Tsdb/sharding-benchmarks

owen-d:
This PR does the following:
1) Copies `ShardAnnotation` over to the `tsdb/index` package, both where it will remain and for alignment with other sharding work (followup PR)
2) Adds a sharding-aware benchmark to prove sharding optimizations with.


cyriltovena (approved):
Lgtm

PR #5630, opened 2022-03-15
TSDB Index reuses slices, adds pools

owen-d:
This PR changes the index interface to accept slices. This allows us to reuse memory more easily. To make this easier, I've introduced a number of releveant pools for us to use.

It also adds a configurable `rounds` parameter (default 8) for the benchmarking script.

```
benchmarks:

name                                         old time/op    new time/op    delta
Query_PostingsForMatchers/match_ns-4           4.03µs ± 1%    4.02µs ± 0%     ~     (p=0.159 n=7+8)
Query_PostingsForMatchers/match_ns_regexp-4    4.03µs ± 0%    4.02µs ± 0%   -0.28%  (p=0.031 n=8+8)
Query_GetChunkRefs/match_ns-4                  27.4ms ± 0%    14.2ms ± 0%  -48.36%  (p=0.001 n=6+8)
Query_GetChunkRefs/match_ns_regexp-4           27.5ms ± 1%    14.3ms ± 1%  -47.97%  (p=0.000 n=8+7)
Query_GetChunkRefsSharded/match_ns-4            120ms ± 1%     102ms ± 1%  -15.29%  (p=0.000 n=8+8)
Query_GetChunkRefsSharded/match_ns_regexp-4     119ms ± 1%     102ms ± 1%  -14.64%  (p=0.000 n=7+8)

name                                         old alloc/op   new alloc/op   delta
Query_PostingsForMatchers/match_ns-4            80.0B ± 0%     80.0B ± 0%     ~     (all equal)
Query_PostingsForMatchers/match_ns_regexp-4     80.0B ± 0%     80.0B ± 0%     ~     (all equal)
Query_GetChunkRefs/match_ns-4                  40.4MB ± 0%     7.4MB ± 0%  -81.71%  (p=0.000 n=8+8)
Query_GetChunkRefs/match_ns_regexp-4           40.4MB ± 0%     7.4MB ± 0%  -81.70%  (p=0.000 n=8+7)
Query_GetChunkRefsSharded/match_ns-4           48.6MB ± 0%    15.0MB ± 0%  -69.20%  (p=0.001 n=7+7)
Query_GetChunkRefsSharded/match_ns_regexp-4    48.6MB ± 0%    15.0MB ± 0%  -69.20%  (p=0.001 n=8+6)

name                                         old allocs/op  new allocs/op  delta
Query_PostingsForMatchers/match_ns-4             4.00 ± 0%      4.00 ± 0%     ~     (all equal)
Query_PostingsForMatchers/match_ns_regexp-4      4.00 ± 0%      4.00 ± 0%     ~     (all equal)
Query_GetChunkRefs/match_ns-4                    278k ± 0%      278k ± 0%   -0.01%  (p=0.000 n=8+8)
Query_GetChunkRefs/match_ns_regexp-4             278k ± 0%      278k ± 0%   -0.01%  (p=0.000 n=8+8)
Query_GetChunkRefsSharded/match_ns-4             394k ± 0%      394k ± 0%   -0.12%  (p=0.000 n=8+8)
Query_GetChunkRefsSharded/match_ns_regexp-4      394k ± 0%      394k ± 0%   -0.12%  (p=0.000 n=8+8)
```


pkg/storage/tsdb/index.go
```diff
@@ -32,8 +32,8 @@ func (r ChunkRef) Less(x ChunkRef) bool {
 
 type Index interface {
 	Bounded
-	GetChunkRefs(ctx context.Context, userID string, from, through model.Time, shard *index.ShardAnnotation, matchers ...*labels.Matcher) ([]ChunkRef, error)
-	Series(ctx context.Context, userID string, from, through model.Time, shard *index.ShardAnnotation, matchers ...*labels.Matcher) ([]Series, error)
+	GetChunkRefs(ctx context.Context, userID string, from, through model.Time, res *[]ChunkRef, shard *index.ShardAnnotation, matchers ...*labels.Matcher) error
+	Series(ctx context.Context, userID string, from, through model.Time, res *[]Series, shard *index.ShardAnnotation, matchers ...*labels.Matcher) error
```
cyriltovena, line 36: I think we should still returns []Series, just more conventional when using it.

cyriltovena: same for res *[]ChunkRef I don't think we need a pointer though.

pkg/storage/tsdb/multi_file_index.go
```diff
@@ -92,80 +92,68 @@ func (i *MultiIndex) forIndices(ctx context.Context, from, through model.Time, f
 	return results, nil
 }
 
-func (i *MultiIndex) GetChunkRefs(ctx context.Context, userID string, from, through model.Time, shard *index.ShardAnnotation, matchers ...*labels.Matcher) ([]ChunkRef, error) {
+func (i *MultiIndex) GetChunkRefs(ctx context.Context, userID string, from, through model.Time, res *[]ChunkRef, shard *index.ShardAnnotation, matchers ...*labels.Matcher) error {
+	*res = (*res)[:0]
 	groups, err := i.forIndices(ctx, from, through, func(ctx context.Context, idx Index) (interface{}, error) {
-		return idx.GetChunkRefs(ctx, userID, from, through, shard, matchers...)
+		refs := ChunkRefsPool.Get()
+		err := idx.GetChunkRefs(ctx, userID, from, through, &refs, shard, matchers...)
```
cyriltovena, line 99: if the final append was inside here you wouldn't need to Get from the pool multiple times, just reset the slice for each indice.

Same for the Multi Series.

cyriltovena (approved):
LGTM

I think we should keep the interface to return results, feels more natural to me.

pkg/storage/tsdb/multi_file_index.go
```diff
@@ -92,80 +92,68 @@ func (i *MultiIndex) forIndices(ctx context.Context, from, through model.Time, f
 	return results, nil
 }
 
-func (i *MultiIndex) GetChunkRefs(ctx context.Context, userID string, from, through model.Time, shard *index.ShardAnnotation, matchers ...*labels.Matcher) ([]ChunkRef, error) {
+func (i *MultiIndex) GetChunkRefs(ctx context.Context, userID string, from, through model.Time, res *[]ChunkRef, shard *index.ShardAnnotation, matchers ...*labels.Matcher) error {
+	*res = (*res)[:0]
```
chaudum, line 96: Where is `(*MultiIndex).GetChunkRefs()` called from? If a newly retrieved `*[]ChunkRef` is passed to the function, why do we need to reset the slice?

chaudum (approved):

pkg/storage/tsdb/index.go
owen-d: Agreed, I'll do that :+1:

pkg/storage/tsdb/multi_file_index.go
owen-d: This isn't actually called from anywhere yet -- we're just building the prototype for how we will interact with the tsdb-based index. Regarding `*[]ChunkRef`, it was to allow the caller to supply their own slice to avoid allocations. It's encouraged that the slice could be reused across multiple calls to avoid allocations, so we reset it.

pkg/storage/tsdb/multi_file_index.go
owen-d: I used independent slices because `i.forIndices` runs all these calls in parallel.

PR #5635, opened 2022-03-16
RFC: Release Cadence

dannykopping:
This PR adds a Request For Comment (RFC) for a release cadence proposal.

We invite the community to provide feedback.

We ask that you please keep the feedback substantive; comments like "i don't like this" are not very useful; please provide specific detail and make suggestions on what can be improved/changes.

### DIFF
--- a/docs/sources/rfcs/2022-03-RELEASE-CADENCE.md
+++ b/docs/sources/rfcs/2022-03-RELEASE-CADENCE.md
@@ -1,0 +1,64 @@
+## Background
+
+The Loki team currently releases versions of its software without an advertised schedule.
+
+*Release cadences* are used by many large software projects to provide users and operators with a schedule which they can plan around. It also gives contributors an indication of how long it will take for their contributions to become publicly available.
+
+## Scope
+
+This RFC relates to Loki OSS public releases.
+
+See the [Appendix](#Appendix) for the historical effective release cadence.
+
+## Terminology
+
+*VERSION* and *RELEASE* are used interchangeably to refer to a publicly-available, downloadable artifact with a version number. These artifacts are available under https://github.com/grafana/loki/releases and https://hub.docker.com/r/grafana/loki.
+
+## Goals
+
+The goals of this RFC are to:
+
+- get community input on our proposed release cadence
+- describe how we handle old releases (backporting)
+
+## Proposed Release Cadence
+
+The below are our proposals for the target times between releases; we might release sooner if the severity warrants it. We will endeavour to not miss these targets where practical.
+
+- **[PATCH](https://semver.org/#spec-item-6)** releases: every **2 weeks**
+- **[MINOR](https://semver.org/#spec-item-7)** releases: around **once a quarter**
+- **[MAJOR](https://semver.org/#spec-item-8)** releases: no predefined schedule, likely every **12-18 months**
+
+## Backporting
+
+We can't expect all of our users to be running the latest release of Loki OSS at all times. As such, we want to propose a backporting strategy.
+
+For bugfixes that are backportable, we will endeavour to backport these to the **last 2 minor releases**.
+
+By way of example:
+
+- Loki v15.0.0 is released, featuring a time machine
+- Loki v15.1.0 is released, adding an "on/off" switch for the time machine (minor release)
+- Loki v16.0.0 is released, featuring other cool stuff
+- Loki v15.1.1 and v16.0.1 are released: a critical patch was added which fixes the "on/off" switch - allowing it to be flipped multiple times
+
+We will backport the patch to v15.1.0 and v16.0.0 as the feature exists there, but not to v15.0.0 (can't fix a switch which isn't there).
+
+_How do we define backportable bugfixes?_ We've tried to come up with a clean definition, but it's quite a difficult thing to define; one exception here is security-related bugfixes, which should always be backported if relevant. We will use our best judgment, and contributors who submit bugfixes can of course motivate for their PRs to be backported.
+
+## Appendix
+
+### Historical Effective Release Cadence
+
+| Release Date | Version | Type  | Delta        |
+| ------------ | ------- | ----- | ------------ |
+| 12-01-2022   | 2.4.2   | patch | *+ 6 weeks*  |
+| 08-11-2021   | 2.4.1   | patch | *+ 2 days*   |
+| 06-11-2021   | 2.4.0   | minor | *+ 3 months* |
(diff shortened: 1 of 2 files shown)


james-callahan:
>  - **[MINOR](https://semver.org/#spec-item-7)** releases: around **once a quarter**

I think holding back a minor release with a useful feature in it because "it's not time yet" isn't a positive trait: if there is a new feature merged, why not make a release sooner rather than later?

dannykopping:
> > * **[MINOR](https://semver.org/#spec-item-7)** releases: around **once a quarter**
> 
> I think holding back a minor release with a useful feature in it because "it's not time yet" isn't a positive trait: if there is a new feature merged, why not make a release sooner rather than later?

There is a clause just above it which states:

> The below are our proposals for the target times between releases; we might release sooner if the severity warrants it.

Maybe I should replace "severity" with "priority". Would that then communicate the sentiment a bit more clearly, do you think?

james-callahan:
> Maybe I should replace "severity" with "priority". Would that then communicate the sentiment a bit more clearly, do you think?

To take an extreme position: why would the smallest relatively insignificant new feature deserve an immediate release. I think probable answers are:
  - It takes time to do the release process
    e.g. because the release process has manual steps; or there's a significant "pre-release" QA process
  - worried about version number explosion

But to be just slightly more reasonable: why not a minor release e.g. once a week?

periklis:
Look on this I have to say first that this is an good start and finally a good point in time to involve the community. Thank you @dannykopping writing this down. :rocket: 

My comments on improving this document are:
1. I suggest to add a section about handling deprecations. Across the 1.x and 2.x lines we have certainly pilled up config knobs and operating modes that we might want to discuss deprecating. How long to users have to expect that deprecations hold? Removing deprecations usually have an impact on rolling out new releases on production environments and/or add upgrade procedures into the loki-operator and helm charts. From the loki-operator perspective it is easy to pin the version till we've done the job, but helm chart users might need to be aware of the deprecation schedule before bumping a version in a variable.
2. The patch release cadence sounds super aggressive to me and if I count correctly we will have 3 active minor releases at a time supporting patch-releases every two weeks. That counts up to 6 patch releases every second week if and only if we have content to release. Is this currently possible in regards of CI resources and release management bookkeeping?
3. FWIW we should add a section about supporting features that are currently only available behind feature gates. Planning and executing long-term changes (e.g. TSDB) under the hood require to introduce piece by piece at a time. Moving along the timeline means that various minor releases might have include incomplete pieces (e.g. TSDB store in 15.1.0 but no sharding until 15.3.0). In turn this means disabled features under construction should not expect to get supported in any minor releases until we declare them GA. A couple of historical examples are the boltdb_shipper and the memberlist.

WDYT?

jeschkies:
@periklis 

> I suggest to add a section about handling deprecations.

I had similar thoughts. However, this would be related to compatibility IMHO and that discussion is out of scope of the release cadence.

afayngelerindbx:
+1 to explicit deprecation semantics

Another thing to consider is possibly codifying pre-release behavior for minor/major versions: having beta/alpha artifacts consistently available would make it easier to solicit community feedback on larger releases.

periklis:
> @periklis
> 
> > I suggest to add a section about handling deprecations.
> 
> I had similar thoughts. However, this would be related to compatibility IMHO and that discussion is out of scope of the release cadence.

I agree my point is truly optional for this doc and it documenting the deprecation semantics deserves a separate RFC. AFAIK we have not deprecated anything yet, but I bet the timing before a future 3.x becomes reality is good to start deprecating pieces.

dannykopping:
> Look on this I have to say first that this is an good start and finally a good point in time to involve the community. Thank you @dannykopping writing this down. rocket

Thanks for your thoughts @periklis - see my responses below:

> 
> My comments on improving this document are:
> 
> 1. I suggest to add a section about handling deprecations. Across the 1.x and 2.x lines we have certainly pilled up config knobs and operating modes that we might want to discuss deprecating. How long to users have to expect that deprecations hold? Removing deprecations usually have an impact on rolling out new releases on production environments and/or add upgrade procedures into the loki-operator and helm charts. From the loki-operator perspective it is easy to pin the version till we've done the job, but helm chart users might need to be aware of the deprecation schedule before bumping a version in a variable.

We decided to not complicate this RFC with additional details like this; it would've made the review too unwieldy. We can certainly get that RFC started in parallel. Would you be interested in driving this, or contributing in some other way?

> 2. The patch release cadence sounds super aggressive to me and if I count correctly we will have 3 active minor releases at a time supporting patch-releases every two weeks. That counts up to 6 patch releases every second week if and only if we have content to release. Is this currently possible in regards of CI resources and release management bookkeeping?

We intentionally made the schedule on the more aggressive side as a forcing function to drive some automation around releases. The more releases we do and the more painful they are, the more impetus we have to improve the situation.

> if I count correctly we will have 3 active minor releases at a time

How did you come to 3 here? We will only backport fixes to the current and previous minor release, so it'll just be 2 patch releases every ~2 weeks.

> 3. FWIW we should add a section about supporting features that are currently only available behind feature gates. Planning and executing long-term changes (e.g. TSDB) under the hood require to introduce piece by piece at a time. Moving along the timeline means that various minor releases might have include incomplete pieces (e.g. TSDB store in 15.1.0 but no sharding until 15.3.0). In turn this means disabled features under construction should not expect to get supported in any minor releases until we declare them GA. A couple of historical examples are the boltdb_shipper and the memberlist.
> 
> WDYT?

I'd vote for letting this evolve naturally. Adding feature gates will further overload the cognitive burden that operators have when running our rather complex software. To use your example, if folks absolutely need TSDB sharding they should wait until v15.3.0 before using the feature.

dannykopping:
> +1 to explicit deprecation semantics
> 
> Another thing to consider is possibly codifying pre-release behavior for minor/major versions: having beta/alpha artifacts consistently available would make it easier to solicit community feedback on larger releases.

@afayngelerindbx please see my message above regarding the deprecation semantics.

Your second point is a good one: we also release weekly container images (our current one is [k88](https://hub.docker.com/r/grafana/loki/tags?page=1&name=k88)), and it's worth documenting the availability of these. We will probably not do any beta/alpha/RC/GA releases yet, but we will move in this direction in the future if the desire is there.

jeschkies:
@dannykopping i think this should be converted to a LID 🤷🏼‍♂️

MasslessParticle:
I agree with @jeschkies. Also, we haven't discussed this in nearly 9 months so I'm not sure how relevant it still is. I'm going to close it for now.

PR #5636, opened 2022-03-16
WIP: Provide a GetChunkRef API within the index-gateway

simonswine:

This adds a GetChunkRef API within the index-gateway, to reduce the latency of Labels and Series API, which currently transmits a lot of `QueryIndex` calls for smaller queries.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


pkg/storage/chunk/chunk.go
```diff
@@ -37,15 +38,20 @@ func errInvalidChunkID(s string) error {
...
+		// On fetch we will initialise these fields from the DynamoDB key.
+		Fingerprint model.Fingerprint `json:"fingerprint"`
+		UserID      string            `json:"userID"`
+
+		// These fields will be in all chunks, including old ones.
+		From    model.Time    `json:"from"`
+		Through model.Time    `json:"through"`
+	*/
+
+	Metric labels.Labels `json:"metric"`
```
cyriltovena, line 54: I think we should also move the checksum.

PR #5641, opened 2022-03-16
Remove benchmark cron job.

jeschkies:
Summary
- The benchmarks were not stable enough.
- They did not yield much value as we did not give much attention to
  them.

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -371,24 +371,6 @@ local manifest(apps) = pipeline('manifest') {
       },
     ],
   },
-  pipeline('benchmark-cron') {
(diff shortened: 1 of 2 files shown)


dannykopping (approved):
Thanks for this @jeschkies 

To be fair to you, these benchmarks could well have delivered value if we were able to run them on dedicated hardware; we are currently not able to, which is unfortunate. We should totally revive this later when we have this capability.

slim-bean:
I wonder if there is value in running them still with a really high threshold? something outside the noise of not having dedicated hardware? not sure if this would really work or not but could potentially catch a major performance regression?

PR #5643, opened 2022-03-16
Introduce a ChunkRef type as part of logproto 

simonswine:

This adds a ChunkRef protobuf type as a part of logproto.

In follow up PRs I plan to:

- Add a `Checksum` field (keeping the size of the refactoring small per PR)
- Replace the TSDB type `ChunkRef` with the one from logproto: https://github.com/grafana/loki/blob/8c06c546ab15a568f255461f10318dae37e022d3/pkg/storage/tsdb/index.go#L17-L22

**Checklist**
~- [ ] Documentation added~
~- [ ] Tests updated~
~- [ ] Add an entry in the `CHANGELOG.md` about the changes.~


cyriltovena (approved):
LGTM

Issue #2826, opened 2020-10-27
v2 go module semantic versioning


optiman:
**Describe the bug**
I'm using go client library (`/pkg/promtail/client`) instead of promtail to push log entries to Loki. 
After updating Loki to v2.0.0 I cannot use go client library in projects with `go.mod` because Loki's module path does not contain `/v2` suffix as [required by the go command](https://golang.org/cmd/go/#hdr-Module_compatibility_and_semantic_versioning).

> To preserve import compatibility, the go command requires that modules with major version v2 or later use a module path with that major version as the final element. For example, version v2.0.0 of example.com/m must instead use module path example.com/m/v2, and packages in that module would use that path as their import path prefix, as in example.com/m/v2/sub/pkg.

**To Reproduce**

- `mkdir myproject && cd myproject`
- `go mod init myproject`
- `go get github.com/grafana/loki@v2.0.0`

**Expected behavior**
Be able to use Loki v2 go client library in projects with `go.mod` file (module-aware mode).

**Environment:**
 local, linux

**Screenshots, Promtail config, or terminal output**
`go get github.com/grafana/loki@v2.0.0: github.com/grafana/loki@v2.0.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2`

slim-bean:
Yes this was expected and unfortunately is (IMO) a limitation and annoyance with Go mod. I don't think it's appropriate that Go mod scans repos and uses tags to assign versions, this really limits what we can do for versioning a project which was never built with the intention of being a library.

We made the decision to use the 2.0.0 tag knowing it would break go mod. 

Sadly, your use case is completely acceptable and we want to encourage people to use things like the promtail client package in their code.

However, we want the ability to control the Loki version number as it feels appropriate to us and the feature set in the project, not as it pertains to it's use as a go library. 

This creates a dilemma for us, and we chose to go the path of tying the version number to the project state and features, not to the use as a library.

We appreciate this issue, apologies for my long response but you were lucky to be the first to find this 😃 and I want to have something to reference when others find this problem too!

The workaround is to install Loki based on commit hashes now and not version tags.

The commit hash for the 2.0.0 tag is `6978ee5d73878d7406bda68296a3dd8ecda4d538`

```
❯ go get -u github.com/grafana/loki@6978ee5d73878d7406bda68296a3dd8ecda4d538                                                                                
[...]

PR #5646, opened 2022-03-16
VERSIONING: Add v2 suffix to the module path

### LINKED ISSUE #4359: Make package Go Module compatible

The package is not Go Module compatible. The last version that is "go get-able" is `v1.6.1`:

```shell
$ go get github.com/grafana/loki@v2.3.0
go get: github.com/grafana/loki@v2.3.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

Make the package Go Module compatible.

There are no proper alternatives apart from maintaining a fork.

**Additional context**

Screenshot 2021-09-21 at 11 06 48

lukasmalkmus:
It is not possible to get this module using `go get` for the `v2.x` tag. This PR adds the `v2` suffix to the modules import path.

Fixes #4359

The [preferred](https://go.dev/blog/v2-go-modules) approach according to this, is a different one. However, [go-github](https://github.com/google/go-github/blob/master/scrape/apps.go#L20) chooses another approach that seems to be widely accepted.

This PR should not break anyone as `go get` does not work with the `v2.x` tag anyways, right? RIGHT?

**Checklist**
- [x] Documentation added
- [x] Tests updated


kavirajk:
@lukasmalkmus understand your pain :(  We come across this discussion once in a while. There are some effort in the past to do this as well.

Please take a look at this discussion here. https://github.com/grafana/loki/issues/2826#issuecomment-968709560

tl;dr is even if we plan to support module path in Loki, it supposed to be done on next major release say Loki v3 (as recommended by Go wiki. more details on that thread)

kavirajk:
Closing this PR. but feel free to discuss it in the original issue #2826

lukasmalkmus:
Understood. Is there an eta on a next major realize?

From #2826:

> For example, if you are the author of foo, and the latest tag for the foo repository is v2.2.2,  and foo has not yet adopted modules, then the best practice would be to use v3.0.0 for the first release of foo to adopt modules (and hence the first release of foo to contain a go.mod file).

Just wanna point out for people stumbling accross this, that having a `go.mod` file in a `v2.x` tagged project but not having adopted modules (by using the appropriate import path or working with the sub-package approach) is causing confusion for `go get` and resulting in the error pointed out in #4359.

PR #5651, opened 2022-03-17
improve boltdb-shipper logging to help debug index query latency issues

sandeepsukhani:
Improve logging on query path to help debug latency and other possible issues in `boltdb-shipper`.
I have added logs to query traces if the query has to refresh the objects list cache and download the index at query time.

### DIFF
--- a/pkg/storage/stores/shipper/downloads/index_set.go
+++ b/pkg/storage/stores/shipper/downloads/index_set.go
@@ -279,7 +279,7 @@ func (t *indexSet) Sync(ctx context.Context) (err error) {
 
 // sync downloads updated and new files from the storage relevant for the table and removes the deleted ones
 func (t *indexSet) sync(ctx context.Context, lock bool) (err error) {
-	level.Debug(t.logger).Log("msg", "syncing files")
+	level.Debug(t.logger).Log("msg", "syncing index files")
 
 	defer func() {
 		status := statusSuccess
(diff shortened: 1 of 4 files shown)


pkg/storage/stores/shipper/downloads/index_set.go
```diff
@@ -96,13 +96,13 @@ func (t *indexSet) Init() (err error) {
 
 	defer func() {
 		if err != nil {
-			level.Error(util_log.Logger).Log("msg", fmt.Sprintf("failed to initialize table %s, cleaning it up", t.tableName), "err", err)
+			level.Error(t.logger).Log("msg", "failed to initialize table, cleaning it up", "err", err)
```
sandeepsukhani, line 99: `t.logger` already has a table name so removing it from here and a couple of places below.

pkg/storage/stores/shipper/downloads/index_set.go
```diff
@@ -294,7 +294,7 @@ func (t *indexSet) sync(ctx context.Context, lock bool) (err error) {
 		return err
 	}
 
-	level.Debug(util_log.Logger).Log("msg", fmt.Sprintf("updates for table %s. toDownload: %s, toDelete: %s", t.tableName, toDownload, toDelete))
+	level.Debug(t.logger).Log("msg", fmt.Sprintf("sync updates. toDownload: %s, toDelete: %s", toDownload, toDelete))
```
dannykopping, line 297:
```suggestion
	level.Debug(t.logger).Log("msg", "sync updates", "toDownload", toDownload, "toDelete", toDelete)
```

This will make it easier to aggregate on these fields

dannykopping: Also not sure if `sync updates` is the most useful log message here; can we find something more descriptive maybe?

dannykopping (approved):
LGTM with a suggestion

pkg/storage/stores/shipper/downloads/index_set.go
sandeepsukhani: yeah, it makes sense. I will update the message.

sandeepsukhani: thanks for the review!

PR #5658, opened 2022-03-17
Add inflight requests metric for scheduler

salvacorts:
This PR adds a new metric to the query-scheduler component.
The metric, `cortex_query_scheduler_inflight_requests`, reports the number of inflight requests (either queued or processing) sampled at a regular interval.

We'll use this metric as part of our querier autoscaling project.
We need to trigger scaling based on something that affects the querier's workload.
Queriers pull tasks from schedulers’ queues, thereafter, we will take both the size of the queue and the number of tasks being processed.

`cortex_query_scheduler_inflight_requests` is a summary so we can use percentiles to answer the question: For the past 5 mins, what was the size of the queue more than 75% of the time? Counterintuitively, taking just the maximum size of the queue is not a good idea since we may have a spike of jobs that take a very short amount of time to get processed, hence the spike not lasting long enough to be worth scaling up the queries.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -92,6 +92,7 @@ type Scheduler struct {
 	connectedFrontendClients prometheus.GaugeFunc
 	queueDuration            prometheus.Histogram
 	schedulerRunning         prometheus.Gauge
+	inflightRequests         prometheus.Summary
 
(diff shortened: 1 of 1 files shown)


dannykopping (approved):
LGTM

PR #5661, opened 2022-03-17
Invalidate caches on delete

MasslessParticle:
This PR does two things:
1. Generate and store a `generation number` for a tenant when deletes are added, processed, or removed.
2. Creates an implementation of `CacheGenNumLoader` that is passed to our metrics `Tripperware` 

The result is that when deletes are added or change for a user, the caches are queried with an updated generation number on the keys so a user can see filtered/unfiltered results as deletes are created and processed. 

**Checklist**
- [ ] Documentation added
- [x] Tests updated


pkg/loki/loki.go
```diff
@@ -487,6 +490,7 @@ func (t *Loki) setupModuleManager() error {
 	mm.RegisterModule(IndexGateway, t.initIndexGateway)
 	mm.RegisterModule(QueryScheduler, t.initQueryScheduler)
 	mm.RegisterModule(UsageReport, t.initUsageReport)
+	mm.RegisterModule(CacheGenNumberLoader, t.initCacheGenNumberLoader)
```
MichelHollands, line 493: Can this be a UserInvisibleModule?

pkg/storage/stores/shipper/compactor/deletion/gennumber_loader.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func (l *GenNumberLoader) loop() {
+	timer := time.NewTicker(reloadDuration)
+	for {
+		select {
+		case <-timer.C:
+			err := l.reload()
+			if err != nil {
+				level.Error(log.Logger).Log("msg", "error reloading tombstones", "err", err)
```
MichelHollands, line 69: Tombstones doesn't look like the right term here. Can we use cache generation numbers or something else?

pkg/storage/stores/shipper/compactor/deletion/gennumber_loader_test.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+func TestGetCacheGenNumber(t *testing.T) {
+	s := &mockGenNumberGetter{
+		genNumbers: map[string]string{
+			"tenant-a": "1000",
+			"tenant-b": "1050",
+		},
+	}
+	tombstonesLoader := NewGenNumberLoader(s, nil)
```
MichelHollands, line 17:
```suggestion
	cacheGenNumberLoader := NewGenNumberLoader(s, nil)
```
This also needs to change later on.

MichelHollands (commented):
LGTM. A few smaller comments.

pkg/loki/loki.go
```diff
@@ -500,10 +504,10 @@ func (t *Loki) setupModuleManager() error {
 		OverridesExporter:        {Overrides, Server},
 		TenantConfigs:            {RuntimeConfig},
 		Distributor:              {Ring, Server, Overrides, TenantConfigs, UsageReport},
-		Store:                    {Overrides},
+		Store:                    {Overrides, CacheGenNumberLoader},
 		Ingester:                 {Store, Server, MemberlistKV, TenantConfigs, UsageReport},
 		Querier:                  {Store, Ring, Server, IngesterQuerier, TenantConfigs, UsageReport},
-		QueryFrontendTripperware: {Server, Overrides, TenantConfigs},
+		QueryFrontendTripperware: {Server, Overrides, TenantConfigs, CacheGenNumberLoader},
```
sandeepsukhani, line 510: I see a couple of complexities with initializing cache gen number loader from QF:
* QF will now depend on the index store for downloading delete requests db and reading cache gen number from it.
* Deployment of QF will become complex in a case where there are custom mount paths set in yaml config for downloading boltdb files. In jsonnet it is set to `/data/boltdb-cache` and the same yaml is shared by all the components. Since QF do not have any such path, it would require us to change it for QF or provision the same path for QF to be able to write to it.

I was hoping to avoid adding store dependency to QF. The other way I can think of is QF pulling results cache gen numbers from Queriers, using an API. It would only keep gen number in memory for the users whose queries it has got for processing. QF would keep pulling updates from Queriers for all the users gen numbers that it has in memory. 

What do you think?

pkg/storage/stores/shipper/compactor/deletion/gennumber_loader.go
MasslessParticle: Good call. I pulled a lot of this from Cortex and missed this

pkg/storage/stores/shipper/compactor/deletion/gennumber_loader.go
```diff
@@ -0,0 +1,173 @@
...
+	return updatedGenNumbers, nil
+}
+
+func (l *GenNumberLoader) GetResultsCacheGenNumber(tenantIDs []string) string {
+	return l.getCacheGenNumbersPerTenants(tenantIDs)
+}
+
+func (l *GenNumberLoader) GetStoreCacheGenNumber(tenantIDs []string) string {
+	return l.getCacheGenNumbersPerTenants(tenantIDs)
+}
```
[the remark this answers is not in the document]
MasslessParticle: I'm still a little confused on this. Why not just invalidate all the caches whenever a delete happens? Is the idea to preserve the storage cache as long as possible?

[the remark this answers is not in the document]
trevorwhitney: I too am confused. If we invalidate both, won't the results cache get updated as a result of caching the results from the query that hit the actual store on account of the store cache also being invalidated? That seems like less complexity to me, just always invalidate both.

pkg/loki/loki.go
trevorwhitney: I agree that this could be a messy dependency as now there are potentially 2 components downloading chunks to store locally, increasing traffic to object storage, and agree the idea of an API QF can use to Querier is worth pursuing.

I am, however, having trouble following the path issue. I don't see a mount to `/data` for either QF or querier, so I think both are just using ephemeral disk inside the container, and it shouldn't matter?

trevorwhitney (commented):
Looks great, just a couple reactions to @sandeepsukhani's comments.

pkg/loki/loki.go
MasslessParticle: Queriers already depend on Frontends and I'm wary of introducing an API where frontends now depend on queriers. I think a better way to do things might be to make a "shipper" specifically for deletes that can write to ephemeral. I think the only real difference would be the tables that are downloaded.

This is a POC that seems to work where the whole shipper writes to ephemeral disk, https://github.com/MasslessParticle/loki/pull/3 

If we take this approach, making a delete-specific component could be a follow-on PR. Thoughts?

sandeepsukhani: Re API to get results cache gen: We can also makw QF talk to Compactor instead of Queries. 

Re downloading delete requests table in QF: Other than adding a dependency on Store, the thing I am concerned about is whether we will be able to cover all failure modes like QF running as a binary on a baremetal not having write access to temp, which I am not sure is even possible.

pkg/storage/stores/shipper/compactor/deletion/gennumber_loader.go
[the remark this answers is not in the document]
sandeepsukhani: When a new delete request comes in, we will do query time filtering of data until the actual data from the storage gets deleted. So if we invalidate both results and store cache when a new delete request comes in, we would anyways cache the same data since we have not processed the delete requests yet, so it would be a waste to drop and cache the same data again.

Similarly, when the delete request is processed, whatever results we have cached would be built after filtering out deleted data since we would have cleared the results cache already when the delete request was received. So, if we invalidate the results cache again after deleting data from the store, we would be wasting resources to drop and cache the same data again, this time without query time filtering but with the actual data deleted from storage.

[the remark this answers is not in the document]
MasslessParticle: This makes sense. I'm going to implement it in a follow on PR

pkg/loki/modules.go
```diff
@@ -53,6 +52,7 @@ import (
 	"github.com/grafana/loki/pkg/storage/stores/shipper"
 	"github.com/grafana/loki/pkg/storage/stores/shipper/compactor"
 	"github.com/grafana/loki/pkg/storage/stores/shipper/compactor/deletion"
+	"github.com/grafana/loki/pkg/storage/stores/shipper/compactor/generationnumber"
```
cstyan, line 55: nit: indentation

pkg/loki/modules.go
```diff
@@ -510,6 +516,29 @@ func (t *Loki) initQueryFrontendTripperware() (_ services.Service, err error) {
 	return services.NewIdleService(nil, nil), nil
 }
 
+func (t *Loki) frontendGenerationNumberGetter() (generationnumber.CacheGenClient, error) {
```
cstyan, line 519: could we just name this `cacheGenClient()` or `getCacheGenClient()`? we usually don't use `thingGetter` for naming in Go IME

pkg/storage/stores/shipper/compactor/deletion/delete_requests_manager_test.go
```diff
@@ -54,6 +54,14 @@ func (m mockDeleteRequestsStore) Stop() {
 	panic("implement me")
 }
 
+func (m mockDeleteRequestsStore) GetCacheGenerationNumber(ctx context.Context, userID string) (string, error) {
+	panic("implement me")
+}
```
cstyan, line 59: should this be implemented as part of this PR?

pkg/storage/stores/shipper/compactor/deletion/delete_requests_store.go
```diff
@@ -106,6 +111,10 @@ func (ds *deleteRequestsStore) addDeleteRequest(ctx context.Context, userID stri
 	writeBatch.Add(DeleteRequestsTableName, fmt.Sprintf("%s:%s", deleteRequestDetails, userIDAndRequestID),
 		[]byte(rangeValue), []byte(query))
 
+	// create a gen number for this result
+	writeBatch.Add(DeleteRequestsTableName, fmt.Sprintf("%s:%s", cacheGenNum, userID),
+		[]byte{}, []byte(strconv.FormatInt(time.Now().UnixNano(), 10)))
```
cstyan, line 116: do I have this right, the cache gen number is the unix nano timestamp as a base10 integer? is there much of a possibility that the way we generate that value would change? (should there be a `generateCacheGenNum` function?)

pkg/storage/stores/shipper/compactor/deletion/delete_requests_store.go
```diff
@@ -139,6 +148,11 @@ func (ds *deleteRequestsStore) UpdateStatus(ctx context.Context, userID, request
 	writeBatch := ds.indexClient.NewWriteBatch()
 	writeBatch.Add(DeleteRequestsTableName, string(deleteRequestID), []byte(userIDAndRequestID), []byte(newStatus))
 
+	if newStatus == StatusProcessed {
+		// remove runtime filtering for deleted data
+		writeBatch.Add(DeleteRequestsTableName, fmt.Sprintf("%s:%s", cacheGenNum, userID), []byte{}, []byte(strconv.FormatInt(time.Now().UnixNano(), 10)))
+	}
```
cstyan, line 154: seeing as you're generating a value the same way here  via strconv on `time.Now().UnixNano()` I think my earlier comment about there being a `generateCacheGenNumber` function is valid

pkg/storage/stores/shipper/compactor/deletion/delete_requests_manager_test.go
MasslessParticle: Nope. This is just to satisfy the interface for another component. I made it clearer by having the mock embed the interface and only override the used method

pkg/loki/modules.go
MasslessParticle: This looks the same here and in the editor?

owen-d (commented):
This is looking pretty good, although I think it may be better served by the queriers|index-gateways, which aren't singletons like the compactor and are thus less likely to all be unavailable at once.

MasslessParticle:
We discussed where the API should live at length and settled on the compactor because it's the source of truth on deletes and related data. 

The impact of a down compactor is that the caches are stale for a bit. The goal for deletes is that they happen on the order of hours in the worst case so hopefully someone notices their compactor is down before that?

pkg/loki/modules.go
cstyan: yep, just a display issue on my end

pkg/storage/stores/shipper/compactor/deletion/delete_requests_manager_test.go
cstyan: we never need to call this function as a result of a delete test? :thinking:

owen-d (approved):
LGTM.

Looks like there will be a followup as per the conversation with @sandeepsukhani and this only covers the results cache at the moment.

PR #5662, opened 2022-03-17
Split by range of instant queries

ssncferreira:

This PR introduces a new middleware that splits instant queries with a large range into several smaller sub-queries to parallelize their execution.

The interval by which instant queries are split is defined by the setting `split_queries_by_interval`.

_Note: Using a too small split interval may result in a lot of sub-queries, which can cause the query queue to fill up and make requests slower than faster._

The following screenshot shows the p99 query latency for old and new implementation, using a split interval of `1d` and various vector and range aggregations with `7d` and `30d` ranges respectively.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


pkg/querier/queryrange/split_by_range.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (s *splitByRange) Do(ctx context.Context, request queryrangebase.Request) (queryrangebase.Response, error) {
+	logger := util_log.WithContext(ctx, s.logger)
+
+	tenants, err := tenant.TenantIDs(ctx)
+	if err != nil {
+		return nil, httpgrpc.Errorf(http.StatusBadRequest, err.Error())
+	}
+
+	interval := validation.SmallestPositiveNonZeroDurationPerTenant(tenants, s.limits.QuerySplitDuration)
```
chaudum, line 50: The split duration is still subject to change and requires further testing. We may need to introduce a second setting, or even calculate the split duration dynamically based on the range of the query.

pkg/querier/queryrange/split_by_range.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil, err
+	}
+
+	value, err := marshal.NewResultValue(res.Data)
+	if err != nil {
+		return nil, err
+	}
+
+	switch res.Data.Type() {
+	case parser.ValueTypeMatrix:
```
owen-d, line 95: nit: by having two separate middlewares, we're converting back and forth between promtheus types (`Matrix, Vector, Scalar`) and `queryrange.Response` types multiple times in each query (since each middleware has it's own logql engine which downstreams. This is probably fine for simplicity's sake unless we find a better way to handle this.

owen-d (commented):
This is looking really good, but it's in need of some equivalence testing (see `TestMappingEquivalence` in `logql/downstream_test.go`)

chaudum:
> This is looking really good, but it's in need of some equivalence testing (see `TestMappingEquivalence` in `logql/downstream_test.go`)

There are a bunch of range aggregation and vector aggregation queries tested in `TestRangeMappingEquivalence` in `downstream_test.go`.
Or do you mean, adding an additional test with the new range mapper **and** the shard mapper as in `TestMappingEquivalence`?

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,313 @@
...
+}
+
+// hasLabelExtractionStage returns true if an expression contains a stage for label extraction,
+// such as `| json` or `| logfmt`
+func hasLabelExtractionStage(expr syntax.SampleExpr) bool {
+	found := false
+	expr.Walk(func(e interface{}) {
+		switch concrete := e.(type) {
+		case *syntax.LabelParserExpr:
+			if concrete.Op == syntax.OpParserTypeJSON || concrete.Op == syntax.OpParserTypeLogfmt {
```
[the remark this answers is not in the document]
chaudum: We limited to to JSON and LOGFMT only, because these are the ones that let the label cardinality explode and therefore fail if executed as downstream query due to label limit.
For consistency reasons, we could restrict all label extraction stages.

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,313 @@
...
+func (m RangeVectorMapper) mapRangeAggregationExpr(expr *syntax.RangeAggregationExpr, vectorAggrPushdown *syntax.VectorAggregationExpr) syntax.SampleExpr {
+	rangeInterval := getRangeInterval(expr)
+
+	// in case the interval is smaller than the configured split interval,
+	// don't split it.
+	if rangeInterval <= m.splitByInterval {
+		return expr
+	}
+
+	if expr.Grouping == nil && hasLabelExtractionStage(expr) {
```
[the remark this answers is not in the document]
chaudum: Exactly. Will add a comment.

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+}
+
+func isSplittableByRange(expr syntax.SampleExpr) bool {
+	switch e := expr.(type) {
+	case *syntax.VectorAggregationExpr:
+		_, ok := SplittableVectorOp[e.Operation]
+		return ok && isSplittableByRange(e.Left)
+	case *syntax.BinOpExpr:
+		return true
```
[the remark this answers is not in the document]
chaudum: Good point. Atm it works, because for each leave we call Map again and check if anything further down is splittable.
Checking also leaves recursively here will allow to call `isSplittableByRange` once, even before `Map()`.

pkg/logql/rangemapper_test.go
```diff
@@ -0,0 +1,1138 @@
...
+	rvm, err := NewRangeVectorMapper(2 * time.Second)
+	require.NoError(t, err)
+
+	for _, tc := range []struct {
+		expr     string
+		expected string
+	}{
+		{
+			`bytes_over_time({app="foo"}[3s])`,
+			`sum without(
```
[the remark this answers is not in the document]
chaudum: Not sure what you mean with "It won't pass the parser in downstream queriers"

Yes, the expected result is the string representation of the query. Also `downstream<expr ...., shard=<nil>>` is not valid LogQL.

The problem that I see in Loki, is that there is no clear distinction between the expression of a a parsed LogQL string, and the "query plan".

pkg/logql/rangemapper.go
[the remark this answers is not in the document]
owen-d: I'm mostly worried that the function says one thing (tests if the Expr is splittable) but doesn't always do it correctly. This can work _now_, but it seems easily susceptible to accidental refactoring which breaks.

pkg/logql/rangemapper_test.go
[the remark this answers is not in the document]
owen-d: When we execute this code, the `downstream` and `concat` nodes are only used by the query-frontend in scheduling/aggregation concerns, but they dispatch the wrapped expressions to queriers for evaluation. This means a querier will try to handle a query like `sum without (expr)`, but will immediately fail with parsing errors. I'm guessing we'll run into this when trying to run this code.

pkg/logql/rangemapper.go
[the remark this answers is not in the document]
owen-d: Understood; this is worth documenting via code comments I think.

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,313 @@
...
+		Op:   syntax.OpTypeDiv,
+		Opts: &syntax.BinOpOptions{},
+	}
+}
+
+func (m RangeVectorMapper) vectorAggrWithRangeDownstreams(expr *syntax.RangeAggregationExpr, vectorAggrPushdown *syntax.VectorAggregationExpr, op string, rangeInterval time.Duration) syntax.SampleExpr {
+	grouping := expr.Grouping
+	if expr.Grouping == nil {
+		grouping = &syntax.Grouping{
+			Without: true,
```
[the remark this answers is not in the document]
owen-d: Resolving this - it's not passed to downstream queriers but is only used on the aggregating node

owen-d:
I've taken another look and found the equivalence tests -- nice work! I've also dismissed a few of my earlier comments. This is looking about ready, @cyriltovena do you want to give it another pair of 👀 ?

pkg/logql/rangemapper.go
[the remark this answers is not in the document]
chaudum: Yes, makes sense. You are totally right. Implemented fully recursive and moved it before the `Map()` call.

chaudum:
@owen-d thanks for the review! I think this new feature still needs additional instrumentation, like in the shard mapper. Hope it's ok to do that in a separate PR.

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,313 @@
...
+	"time"
+
+	"github.com/go-kit/log/level"
+	"github.com/pkg/errors"
+
+	"github.com/grafana/loki/pkg/logql/syntax"
+	util_log "github.com/grafana/loki/pkg/util/log"
+)
+
+type RangeVectorMapper struct {
```
jeschkies, line 14: I'm a little late to the game so please excuse my ignorance. Why is this called a mapper? What does it do? Could you add some documentation?

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,313 @@
...
+		case *syntax.LabelParserExpr:
+			if concrete.Op == syntax.OpParserTypeJSON || concrete.Op == syntax.OpParserTypeLogfmt {
+				found = true
+			}
+		}
+	})
+	return found
+}
+
+// sumOverFullRange returns an expression that sum up individual downstream queries by preserving labels
```
jeschkies, line 103:
```suggestion
// sumOverFullRange returns an expression that sums up individual downstream queries by preserving labels
```

I'm not sure I understand `by preserving labels and dividing`. The example is clearer to me.

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,313 @@
...
+	return &syntax.BinOpExpr{
+		SampleExpr: &syntax.VectorAggregationExpr{
+			Left: m.mapConcatSampleExpr(downstreamExpr, rangeInterval),
+			Grouping: &syntax.Grouping{
+				Without: true,
+			},
+			Operation: syntax.OpTypeSum,
+		},
+		RHS:  &syntax.LiteralExpr{Val: rangeInterval.Seconds()},
+		Op:   syntax.OpTypeDiv,
```
jeschkies, line 132: Does this always apply?

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,313 @@
...
+		Opts: &syntax.BinOpOptions{},
+	}
+}
+
+func (m RangeVectorMapper) vectorAggrWithRangeDownstreams(expr *syntax.RangeAggregationExpr, vectorAggrPushdown *syntax.VectorAggregationExpr, op string, rangeInterval time.Duration) syntax.SampleExpr {
```
jeschkies, line 137: Could you add an example like above? That helps quite a bit.

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	// Add the remainder offset interval
+	if rangeInterval%m.splitByInterval != 0 {
+		offset := time.Duration(interval) * m.splitByInterval
+		downstreams = appendDownstream(downstreams, expr, rangeInterval-offset, offset)
+	}
+
+	if downstreams == nil {
+		return expr
+	}
```
jeschkies, line 196: How about moving this up and use the `splitCount` instead?

```
if splitCount == 0 {
  return expr
}
```

pkg/querier/queryrange/split_by_range.go
jeschkies: We also parse the query each time. That's probably not as bad. In any case it feels these middlewares could be refactored at some point and "just" pass down the modified query or query plan.

pkg/querier/queryrange/split_by_range.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	value, err := marshal.NewResultValue(res.Data)
+	if err != nil {
+		return nil, err
+	}
+
+	switch res.Data.Type() {
```
jeschkies, line 94: `NewQueryShardMiddleware` has the same dance. What do you think about extracting this logic? Maybe it's not useful when just two function use it.

jeschkies (commented):
Nice. I'm half way through the mapper. Need to go once more through it.

pkg/logql/rangemapper.go
chaudum: Since the rewriting follows the same pattern as in the [ShardMapper](https://github.com/grafana/loki/blob/main/pkg/logql/shardmapper.go#L96), we kept the same naming.

But thinking about it, `Vector` should not really be part of the name.

pkg/logql/rangemapper.go
chaudum: Ok, will try to clarify the comment.

"sum by preserving labels" should mean, that it merges multiple results with a sum operation without dropping labels. In contrary to vector aggregations, such as `sum()` which return a single stream with a single value.

pkg/logql/rangemapper.go
chaudum: Good idea!

pkg/querier/queryrange/split_by_range.go
chaudum: I'll see if I can extract functions in a meaningful way.

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,313 @@
...
+		return expr, nil
+	}
+
+	// Optimization: if the vector aggregator has grouping, the downstream expression can also have the grouping
+	// in order to reduce the label sets. Does not work with count operation.
+	// TODO: handle the case where an additional vector aggregation expression also has grouping.
+	//  Currently, it is sending the last inner expression grouping dowstream.
+	//  Which grouping should be sent downstream?
+	var vectorAggrPushdown *syntax.VectorAggregationExpr
+	if expr.Grouping != nil && expr.Grouping.Groups != nil && expr.Operation != syntax.OpTypeCount {
```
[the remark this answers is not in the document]
chaudum: I'll check.

pkg/querier/queryrange/split_by_range.go
chaudum: After looking at both middlewares, I think it does not make a lot of sense to try to abstract logic used by both. I do think that the middleware structure will change in the future in favour of a more conventional query planner anyway.

pkg/logql/downstream_test.go
```diff
@@ -99,7 +99,207 @@ func TestMappingEquivalence(t *testing.T) {
...
+		{`bytes_over_time({a=~".+"}[3s]) + count_over_time({a=~".+"}[5s])`, time.Second},
+		{`sum(count_over_time({a=~".+"}[3s]) * count(sum_over_time({a=~".+"} | unwrap b [5s])))`, time.Second},
+
+		// Multi vector aggregator layer queries
+		{`sum(max(bytes_over_time({a=~".+"}[3s])))`, time.Second},
+		{`sum(min by (a)(max(sum by (b) (count_over_time({a=~".+"} [2s])))))`, time.Second},
+
+		// Non-splittable vector aggregators
+		// TODO: Fix topk
+		//{`topk(2, count_over_time({a=~".+"}[2s]))`, time.Second},
```
cyriltovena, line 246: I'd like to see some successful topk.

cyriltovena (approved):
LGTM ship it 🚀

pkg/logql/downstream_test.go
chaudum: Let's do it in a 2nd iteration :)

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,332 @@
...
+func (m RangeMapper) mapConcatSampleExpr(expr syntax.SampleExpr, rangeInterval time.Duration) syntax.SampleExpr {
+	interval := 0
+	splitCount := int(rangeInterval / m.splitByInterval)
+
+	if splitCount == 0 {
+		return expr
+	}
+
+	var downstreams *ConcatSampleExpr
+	for interval = 0; interval < splitCount; interval++ {
```
trevorwhitney, line 220: super small nit: `interval` is a bit confusing here since it's an iterator while `rangeInterval` and `m.splitByInterval` are both durations. Maybe call this `split`, since it's an iterator across the splits in the `splitCount`?

pkg/logql/rangemapper_test.go
```diff
@@ -0,0 +1,1138 @@
...
+}
+
+func Test_SplitRangeVectorMapping(t *testing.T) {
+	rvm, err := NewRangeMapper(time.Minute)
+	require.NoError(t, err)
+
+	for _, tc := range []struct {
+		expr       string
+		expected   string
+		expectNoop bool
```
trevorwhitney, line 59: nit: this test is massive. I think we could move all the noop cases to their own test for ease of finding them / organizational sake, which means we could get rid of this property in the struct. WDYT?

trevorwhitney (commented):
Well that was a lesson in LogQL. Very cool stuff, excited for how this will improve our queries. Just a couple nits for you as that's about all I could contribute to this already impressive effort. Nice work!

pkg/logql/rangemapper.go
jeschkies: So it's a sum by labels, isn't it?

jeschkies (approved):

pkg/logql/rangemapper.go
```diff
@@ -0,0 +1,332 @@
...
+		_, ok := splittableRangeVectorOp[e.Operation]
+		return ok
+	default:
+		return false
+	}
+}
+
+// clone returns a copy of the given sample expression
+// This is needed whenever we want to modify the existing query tree.
+func clone(expr syntax.SampleExpr) (syntax.SampleExpr, error) {
```
jeschkies, line 330: Why not `syntax.Clone(expr)`?

chaudum: `syntax.Clone()` does basically the same thing, but with different types: `Expr->Expr`, whereas here I want `SampleExpr->SampleExpr`. This is mainly for convenience.

chaudum:
Hey @owen-d I think we are ready to merge this PR. When you squash the commits, please feel free to remove all the "intermediate" commit messages from the individual commits, and use the PR description instead.

PR #5672, opened 2022-03-18
Configure multi-tenant queries in ksonnet.

jeschkies:
This adds the `multi_tenant_queries_enabled` option to the Loki ksonnet library. It defaults to `null` which means it is not set at all in the rendered Loki config.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


production/ksonnet/loki/config.libsonnet
```diff
@@ -47,6 +47,7 @@
       // A higher value will lead to a querier trying to process more requests than there are available
       // cores and will result in scheduling delays.
       concurrency: 4,
+      multi_tenant_queries_enabled: null,
```
kavirajk, line 50: Q: why not just boolean?
`multi_tenant_queries_enabled: false`. 

which would also make below conditional check bit easier

jeschkies: :see_no_evil:

trevorwhitney: what's the behavior if it's not included?

jeschkies: Depends on the Loki version. In older versions Loki would crash. I'd like to avoid that. If it was not included it would default to `false`, for now.

Overall I want to keep backwards compatibility.

dannykopping: It raises an interesting question: are customers using mixins from the release branch? I don't know if we give any specific advice in that regard.

I think we have so many other BC-breaks in that case. Maybe we need to introduce a `version` field in our mixins and conditionally include config based on that? Or some kind of release config like we do for Grafana Cloud Logs?

jeschkies: I agree. However, this would probably blow up the debate here. Also I'm thinking about our deployments. I'm afraid we would break them.

dannykopping (approved):
LGTM

production/ksonnet/loki/config.libsonnet
```diff
@@ -182,7 +183,9 @@
       querier: {
         max_concurrent: $._config.querier.concurrency,
         query_ingesters_within: '2h',  // twice the max-chunk age (1h default) for safety buffer
-      },
+      } + (if $._config.querier.multi_tenant_queries_enabled != null then
```
owen-d, line 186: This check will error if the field is unset.

jeschkies:
Won't do. Users can override the configuration if required.

PR #5673, opened 2022-03-18
Fix debug image by updating Delve version

salvacorts:
The current version of Delve we are using for the debug Docker image does not support newer versions of Go. This PR updates the version we use of Delve so it works with the current version of Go we use for Loki.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


slim-bean (approved):
LGTM!

PR #5675, opened 2022-03-19
Fix bug with new promtail stream lag metric 

cstyan:
This PR fixes a bug from my last PR (https://github.com/grafana/loki/pull/5521) where labels might not be applied if they don't exist on the source stream, which will cause prometheus to panic due to incorrect label cardinality.


dannykopping (approved):
LGTM

PR #5679, opened 2022-03-20
add experimental mmap api

splitice:
**What this PR does / why we need it**:

Uses Mmap for compressed chunk data to reduce pressure on linux reclaiming blocks due to golang not returning pages to linux system memory until under high memory pressure. This can cause exteremly high CPU load peaks as all the reclaim work is deferred until the system is under pressure.

```
10001          9 31.1 29.8 23733032 2429272 ?    Ssl  Mar19 301:49 /usr/bin/loki -config.file=/etc/loki/config/config.yaml -target=ingester -config.expand-env=true
```

2GB RSS, 23GB virtual. 24GB ram system under high reclaim pressure (process stalled)

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] Add an entry in the `CHANGELOG.md` about the changes.

PR #5680, opened 2022-03-20
Add X4B as an adopter

splitice:

Why not? It's one line and the list was looking a little empty.

I'm actually involved in using Loki in two organisations now. Only one in production so far however.

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -3,3 +3,4 @@
 * [Northflank](https://northflank.com)
 * [Red Hat](https://www.redhat.com)
 * [Digitalist](https://www.digitalist.se/)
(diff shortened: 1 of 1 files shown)


dannykopping (approved):
Thanks! Please accept the CLA so we can merge the PR

splitice:
Already signed.

Other pull requests such as https://github.com/grafana/loki/pull/5679 show this correctly too

slim-bean:
It looks like the CLI got stuck, possible related to the issues github has been having with actions, could you push an empty commit or may even be easier to close and open a new PR for this change.

Sorry for the trouble, would love to get your company on the list!

PR #5684, opened 2022-03-21
Implements the TSDB CacheableIndex

### LINKED ISSUE #5618: Implement/define a TSDB Caching Index wrapper
TSDB index interface is defined [here](https://github.com/grafana/loki/blob/main/pkg/storage/tsdb/index.go#L33-L39). We should define a TSDB caching index wrapper, so that TSDB can be used as a caching index client.

This is how caching works for the chunk store: https://github.com/grafana/loki/blob/main/pkg/storage/chunk/storage/factory.go#L221

This is part of https://github.com/grafana/loki/issues/5428

DylanGuedes:
Implements a new TSDB Index implementation named `CacheableIndex`.
This new implementation has the ability to cache results from all TSDB common operations (ex: Series, GetChunkRefs, LabelNames, LabelValues) while relying on existing implementations (fallbacks).
For instance, if the MultiIndex implementation is set as the fallback Index for `CacheableIndex` and a call to `Series` is made and its result isn't cached, `MultiIndex.Series` will be invoked and its results will be cached.

Fixes #5618 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,303 @@
...
+	mountKeyFn := func(userID string, from, through model.Time, shard *index.ShardAnnotation, matcher *labels.Matcher) string {
+		stringfiedShard := ""
+		if shard != nil {
+			stringfiedShard = shard.String()
+		}
+		if matcher == nil {
+			return userID + sep + from.String() + sep + through.String() + stringfiedShard + sep
+		}
+		return userID + sep + from.String() + sep + through.String() + stringfiedShard + sep + matcher.String()
+	}
```
cstyan, line 61: it looks like `mountKeyFn` is the same for each index call? we could reuse the same function defined once?

cstyan, line 61: not sure if there's any real impact but we should probably use fmt.Sprintf

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/grafana/loki/pkg/storage/chunk/cache"
+	"github.com/grafana/loki/pkg/storage/tsdb/index"
+	"github.com/pkg/errors"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/model/labels"
+)
+
+const (
+	// sep defines the character used to separate the components of the cache key.
+	sep = "/"
```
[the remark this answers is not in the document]
DylanGuedes: great point. yeah, maybe confi

pkg/storage/tsdb/cacheable_index.go
DylanGuedes: hmm do you remember the reasoning behind it? out of curiosity after your suggestion I did some research and found [this](https://dev.to/pmalhaire/concatenate-strings-in-golang-a-quick-benchmark-4ahh), that actually states that using the plus sign is faster than using sprintf.

WDYT?

cstyan: Hmmm, from what I remember unless you're only doing `+` once or maybe twice it tends to always be the slowest. `fmt.Sprintf` should be faster I think, but also `strings.Builder` would be even faster if we quickly computed the final string length to allocate the buffer.

At the end of the day I wouldn't worry too much about which one we choose, we can always optimize for performance later, lets just go with something that reads nicely. How about `strings.Join`?

DylanGuedes: `strings.Join` looks good; pushed a commit using it instead, WDYT?

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -54,10 +55,15 @@ func (i *CacheableIndex) GetChunkRefs(ctx context.Context, userID string, from,
 		if shard != nil {
 			stringfiedShard = shard.String()
 		}
-		if matcher == nil {
-			return userID + sep + from.String() + sep + through.String() + stringfiedShard + sep
+
+		keyMembers := []string{
+			userID, from.String(), through.String(), stringfiedShard,
```
cstyan, line 60: should we declare this above `if shard != nil` and then only append `stringifiedShard` to `keyMembers` inside that same if statement?

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -106,10 +112,14 @@ func (i *CacheableIndex) Series(ctx context.Context, userID string, from, throug
 			stringfiedShard = shard.String()
 		}
 
-		if matcher == nil {
-			return userID + sep + from.String() + sep + through.String() + stringfiedShard + sep
+		keyMembers := []string{
+			userID, from.String(), through.String(), stringfiedShard,
+		}
```
cstyan, line 117: same comment related to `stringifiedShard` here

cstyan (commented):
looking good

pkg/storage/tsdb/cacheable_index.go
DylanGuedes: hmm this was a good change for the matcher but I think it won't necessarily be for the shard because we always want the shard, even if it is nil/empty string.

That said, although the number of lines/logic/readability will be the same, I don't have a strong preference for defining the `keyMembers` before or after it. So, double checking, do you prefer if I do the keyMembers assignment before the shard conditional clause?

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,310 @@
...
+
+	// prepare a mapping of keys to matchers, used to retrieve matchers later.
+	keysMapping := make(map[string][]*labels.Matcher, len(matchers))
+	if len(matchers) == 0 {
+		key := keyFn(userID, from, through, shard, nil)
+		keysMapping[opName+sep+key] = []*labels.Matcher{}
+	}
+	for _, matcher := range matchers {
+		key := keyFn(userID, from, through, shard, matcher)
+		keysMapping[opName+sep+key] = []*labels.Matcher{matcher}
```
owen-d, line 278: I think we're being too clever here. Consider `{foo="bar", bazz="buzz"}`. This query asked for the _intersection_ of the two individual matchers.

If we have the following streams:
```
{foo="bar"}
{foo="bar", bazz="buzz"}
{foo="bar", bazz="buzz", bok="blob"}
{foo="bar", bazz="bonk"}
{bazz="buzz"}
```

the intersection will be
```
{foo="bar", bazz="buzz"}
{foo="bar", bazz="buzz", bok="blob"}
```

However, by requesting each matcher individually, we'd get the _union_ of the two:
```
{foo="bar"}
{foo="bar", bazz="buzz"}
{foo="bar", bazz="buzz", bok="blob"}
{bazz="buzz"}
```

That is not what we want. Instead, we should make one call to the cache for `{foo="bar", bazz="buzz"}`.

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,310 @@
+package tsdb
+
+import (
+	"bytes"
+	"context"
+	"encoding/gob"
```
owen-d, line 6: We'll probably want to use an encoding scheme without the same runtime costs as `gob`, especially when creating an encoder per call. This can be left for a followup PR though.

We can likely do this by creating protobuf definitions for our `Series` and `ChunkRef` types, which will make this even easier to use over the wire via `proto.Marshal()` and allow our index-gateway to expose these methods via grpc (cc @cyriltovena we talked about this recently).

owen-d (commented):
This is a good start!

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,310 @@
...
+// It assumes the cache is ready to be used.
+func NewCacheableIndex(index *TSDBIndex, cache cache.Cache) CacheableIndex {
+	return CacheableIndex{
+		Index: index,
+		Cache: cache,
+	}
+}
+
+// Stop does everything necessary to stop this index safely.
+func (i *CacheableIndex) Stop() error {
```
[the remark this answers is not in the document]
DylanGuedes: Hmm I think I added it just in case we need one but agreed, since the Cache interface doesn't have it I think it is better to start without it too.

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+// GetChunkRefs is a cached implementation of GetChunkRefs.
+//
+// It uses as a key for the cache all parameters (userID, from, through, shard, matchers) separated by a slash (/).
+func (i *CacheableIndex) GetChunkRefs(ctx context.Context, userID string, from, through model.Time, res []ChunkRef, shard *index.ShardAnnotation, matchers ...*labels.Matcher) ([]ChunkRef, error) {
+	const opName = "GetChunkRefs"
+
+	mountKeyFn := func(userID string, from, through model.Time, shard *index.ShardAnnotation, matcher *labels.Matcher) string {
+		stringfiedShard := ""
```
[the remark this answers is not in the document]
DylanGuedes: thanks, didn't know that!

pkg/storage/tsdb/cacheable_index.go
DylanGuedes: Good catch, thanks. I'll be looking for a way to unify all matchers into a single string then.

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+// Stop does everything necessary to stop this index safely.
+func (i *CacheableIndex) Stop() {
+	i.Cache.Stop()
+}
+
+// GetChunkRefs is a cached implementation of GetChunkRefs.
+//
+// It uses as a key for the cache all parameters (userID, from, through, shard, matchers) separated by a slash (/).
```
trevorwhitney, line 50: separator was changed to a colon

```suggestion
// It uses as a key for the cache all parameters (userID, from, through, shard, matchers) separated by a colon (:).
```

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,318 @@
...
+func (i *CacheableIndex) GetChunkRefs(ctx context.Context, userID string, from, through model.Time, res []ChunkRef, shard *index.ShardAnnotation, matchers ...*labels.Matcher) ([]ChunkRef, error) {
+	const opName = "GetChunkRefs"
+
+	mountKeyFn := func(userID string, from, through model.Time, shard *index.ShardAnnotation, matchers ...*labels.Matcher) string {
```
trevorwhitney, line 54: curious about the name, why `mountKey`?

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,318 @@
...
+
+		res = append(res, values...)
+	}
+
+	return res, nil
+}
+
+// Series is a cached implementation of Series.
+//
+// It uses as a key for the cache all parameters (userID, from, through, shard, matchers) separated by a slash (/).
```
trevorwhitney, line 103:
```suggestion
// It uses as a key for the cache all parameters (userID, from, through, shard, matchers) separated by a colon (:).
```

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,318 @@
...
+// Stop does everything necessary to stop this index safely.
+func (i *CacheableIndex) Stop() {
+	i.Cache.Stop()
+}
+
+// GetChunkRefs is a cached implementation of GetChunkRefs.
+//
+// It uses as a key for the cache all parameters (userID, from, through, shard, matchers) separated by a slash (/).
+func (i *CacheableIndex) GetChunkRefs(ctx context.Context, userID string, from, through model.Time, res []ChunkRef, shard *index.ShardAnnotation, matchers ...*labels.Matcher) ([]ChunkRef, error) {
+	const opName = "GetChunkRefs"
```
trevorwhitney, line 52: Why store these operation names in a `const`? I only ever see them used in the call to `CacheableOp` and am curious if they could just be inlined then?

trevorwhitney (commented):
Looks great Dylan!

pkg/storage/tsdb/cacheable_index.go
DylanGuedes: I'm bad at naming :rofl: 
but I picked `mountKey` because this func materializes the key used to insert and retrieve values from the cache.

pkg/storage/tsdb/cacheable_index.go
DylanGuedes: good catch, thanks. I believe I defined it as a const because I was using it in different places but after refactoring it isn't necessary anymore, indeed. I'll change it!

pkg/storage/tsdb/cacheable_index.go
trevorwhitney: gotchya, maybe there's precedence for `mount` in this context, but to me that sounds like a `generateKeyFn`?

pkg/storage/tsdb/cacheable_index.go
DylanGuedes: Done.

pkg/storage/tsdb/cacheable_index.go
DylanGuedes: pushed a commit that fixes this (100% based on our pairing today, thanks! :smile: )

pkg/storage/tsdb/cacheable_index.go
trevorwhitney: @owen-d can you link to something or provide a bit more background on the runtime cost of `gob`?

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,275 @@
...
+		enc := gob.NewEncoder(&encodeBuf)
+		if err := enc.Encode(series); err != nil {
+			return nil, errors.Wrap(err, "Series encoding")
+
+		}
+
+		return [][]byte{encodeBuf.Bytes()}, nil
+	}
+
+	results, err := i.CacheableOp(ctx, "Series", i.buildCacheKey, fallbackFn, userID, from, through, shard, matchers...)
```
owen-d, line 139: I see we're _always_ turning our results into `[][]byte`, then deserializing them again even when there was a cache miss. I'd like to avoid this considering this code is called on every query.

pkg/storage/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,275 @@
...
+
+	var values []Series
+	for _, val := range results {
+		decoderBuf := bytes.NewBuffer(val)
+		dec := gob.NewDecoder(decoderBuf)
+		if err := dec.Decode(&values); err != nil {
+			return nil, errors.Wrap(err, "cacheable index Series decoding")
+		}
+
+		res = append(res, values...)
```
owen-d, line 152: This is dangerous: `res` is an argument passed to this function and is an optional slice that the results may be populated into to prevent unnecessarily reallocating memory (this is an efficiency hack). You've also passed it (rightly) as an argument earlier in the `fallbackFn`: `series, err := i.Index.Series(ctx, userID, from, through, res, shard, matcher...)`. In the case of a cache miss, `res` will _already_ hold the results.

owen-d (commented):
Looking good. A few things:
1) We'll need some metrics, take a look at https://github.com/grafana/loki/blob/main/pkg/storage/chunk/storage/caching_index_client.go#L23-L49 for an example of what I'm thinking.
2) We should ignore cache errors. If the cache is causing problems, it shouldn't stop the underlying index from returning.

pkg/storage/tsdb/cacheable_index.go
DylanGuedes: Jordan pushed a commit fixing this, WDYT?

pkg/storage/tsdb/cacheable_index.go
DylanGuedes: I pushed a commit that tests this behavior - WDYT?

pkg/storage/tsdb/cacheable_index.go
cyriltovena: careful with protobuf it sux for arrays, it's a good idea to keep the size of the array somewhere to speed up unmarshalling.

pkg/storage/stores/tsdb/cacheable_index.go
```diff
@@ -0,0 +1,436 @@
...
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "Series encoding")
+		}
+
+		return chunks, buf, nil
+	}
+
+	key := i.buildCacheKey(userID, from, through, shard, matchers...)
+	keys := []string{key}
+	_, response, misses := i.cacheFetch(ctx, keys)
```
cyriltovena, line 180: I'm trying to understand what are the chance we hit this cache between two index query.

For instance, from 10 to  20, then we do from 0 to 20 ? 

e.g what are the chance that from and through are exactly equal ? Sounds like never ? Am I missing something ?

I'd like to also compare this with Mimir implementation ?  Sounds like they are caching at a lower level to probably increase the chance of hitting the cache between multiple queries.

see https://github.com/grafana/mimir/blob/main/pkg/storegateway/indexcache/cache.go#L42-L76

@pracucci @pstibrany Anything we should know about your index cache ? What works well ? what doesn't ?

DylanGuedes: the chances are fairly low. The idea was to first implement this initial version that rarely hits the cache and then implement the optimization of splitting the timestamp into smaller fixed steps that can be reused by other queries.

The optimization in detail would be:
If the step configured for caching is 10 and you sent a query from 20 to 40, later if you sent a query from 10 to 30, the data between 20~30 will be cached so you'll only have to fetch data from 10 to 20.

pracucci: I haven't reviewed this PR (missing too much context) but Mimir index cache is "high level" from my perspective. We do cache the actual lookup result and not the index sub-object portions (like we do for the chunks cache).

The index cache is very effective in Mimir. For example, looking up the cache hit ratio over the last 24h across all Mimir clusters running at Grafana Labs, we get:

```
sum by(item_type) (rate(thanos_store_index_cache_hits_total[$__range])) /
sum by(item_type) (rate(thanos_store_index_cache_requests_total[$__range]))
* 100
```

Screenshot 2022-05-03 at 12 52 26

sandeepsukhani: The cache hit rate for Mimir is good because they are caching data at block level which we can't do at least for live streams since our index shards are just 15m long so our index files for live data change much often.

I think we should check the feasibility of the proposed follow-up to improve the hit rate i.e smaller steps for lookup and merging the results. This approach works well for the results cache because we do not set any expiry on the cached results and let the underlying cache drop the items if it runs out of memory. I am wondering if this approach would work well for index cache since the expiry is just 5m. If we can't do it or it doesn't work out well then the cache would not be useful at all. Let me see if I can put some thoughts in a doc on what we can do about it.

DylanGuedes:
I'll close this PR for now but I plan to return to work on it once the caching layer is required.

PR #5688, opened 2022-03-22
Require unique client configs when there are multiple clients.

cstyan:
In order to fix the bug with promtail multi clients and metrics scraping I introduced the `name` field/config option to clients. This gets around duplicate metric issues by using the name as a label on client specific metrics (of which there is only one currently, the stream lag metric). Since we're doing this, we should also require that client configs are unique when creating clients.

Note that users who need to have duplicate configs can still do so by changing only the `name` config value.


dannykopping (approved):
LGTM, added a commit to fix a linter error

PR #5691, opened 2022-03-22
operator: Fix immediate reset of degraded condition

sasagarw:
- Degraded condition was getting reset immediately when no pod actions were pending.
- updates how lokistack reconcilation is done.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/operator/controllers/lokistack_controller.go
+++ b/operator/controllers/lokistack_controller.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
@@ -106,6 +107,23 @@ func (r *LokiStackReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	err = handlers.CreateOrUpdateLokiStack(ctx, r.Log, req, r.Client, r.Scheme, r.Flags)
+
+	var degraded *status.DegradedError
+	if errors.As(err, &degraded) {
+		err = status.SetDegradedCondition(ctx, r.Client, req, degraded.Message, degraded.Reason)
+		if err != nil {
+			return ctrl.Result{
+				Requeue:      true,
+				RequeueAfter: time.Second,
+			}, err
+		}
+
+		return ctrl.Result{
+			Requeue:      degraded.Requeue,
+			RequeueAfter: time.Second,
+		}, nil
+	}
+
 	if err != nil {
 		return ctrl.Result{
 			Requeue:      true,
--- a/operator/internal/handlers/lokistack_create_or_update.go
+++ b/operator/internal/handlers/lokistack_create_or_update.go
@@ -25,7 +25,14 @@ import (
 )
 
 // CreateOrUpdateLokiStack handles LokiStack create and update events.
-func CreateOrUpdateLokiStack(ctx context.Context, log logr.Logger, req ctrl.Request, k k8s.Client, s *runtime.Scheme, flags manifests.FeatureFlags) error {
+func CreateOrUpdateLokiStack(
+	ctx context.Context,
+	log logr.Logger,
+	req ctrl.Request,
+	k k8s.Client,
+	s *runtime.Scheme,
+	flags manifests.FeatureFlags,
+) error {
 	ll := log.WithValues("lokistack", req.NamespacedName, "event", "createOrUpdate")
 
 	var stack lokiv1beta1.LokiStack
@@ -52,20 +59,22 @@ func CreateOrUpdateLokiStack(ctx context.Context, log logr.Logger, req ctrl.Requ
 	key := client.ObjectKey{Name: stack.Spec.Storage.Secret.Name, Namespace: stack.Namespace}
 	if err := k.Get(ctx, key, &storageSecret); err != nil {
 		if apierrors.IsNotFound(err) {
-			return status.SetDegradedCondition(ctx, k, req,
-				"Missing object storage secret",
-				lokiv1beta1.ReasonMissingObjectStorageSecret,
-			)
+			return &status.DegradedError{
+				Message: "Missing object storage secret",
+				Reason:  lokiv1beta1.ReasonMissingObjectStorageSecret,
+				Requeue: false,
+			}
 		}
 		return kverrors.Wrap(err, "failed to lookup lokistack storage secret", "name", key)
 	}
 
 	storage, err := secrets.ExtractStorageSecret(&storageSecret, stack.Spec.Storage.Secret.Type)
 	if err != nil {
-		return status.SetDegradedCondition(ctx, k, req,
-			"Invalid object storage secret contents",
-			lokiv1beta1.ReasonInvalidObjectStorageSecret,
-		)
+		return &status.DegradedError{
+			Message: "Invalid object storage secret contents",
+			Reason:  lokiv1beta1.ReasonInvalidObjectStorageSecret,
+			Requeue: false,
+		}
 	}
 
 	var (
@@ -74,16 +83,18 @@ func CreateOrUpdateLokiStack(ctx context.Context, log logr.Logger, req ctrl.Requ
 		tenantConfigMap map[string]openshift.TenantData
 	)
 	if flags.EnableGateway && stack.Spec.Tenants == nil {
-		return status.SetDegradedCondition(ctx, k, req,
-			"Invalid tenants configuration - TenantsSpec cannot be nil when gateway flag is enabled",
-			lokiv1beta1.ReasonInvalidTenantsConfiguration,
-		)
+		return &status.DegradedError{
+			Message: "Invalid tenants configuration - TenantsSpec cannot be nil when gateway flag is enabled",
+			Reason:  lokiv1beta1.ReasonInvalidTenantsConfiguration,
+			Requeue: false,
+		}
 	} else if flags.EnableGateway && stack.Spec.Tenants != nil {
 		if err = gateway.ValidateModes(stack); err != nil {
-			return status.SetDegradedCondition(ctx, k, req,
-				fmt.Sprintf("Invalid tenants configuration: %s", err),
-				lokiv1beta1.ReasonInvalidTenantsConfiguration,
-			)
+			return &status.DegradedError{
+				Message: fmt.Sprintf("Invalid tenants configuration: %s", err),
+				Reason:  lokiv1beta1.ReasonInvalidTenantsConfiguration,
+				Requeue: false,
+			}
 		}
 
 		if stack.Spec.Tenants.Mode != lokiv1beta1.OpenshiftLogging {
@@ -96,7 +107,7 @@ func CreateOrUpdateLokiStack(ctx context.Context, log logr.Logger, req ctrl.Requ
 		if stack.Spec.Tenants.Mode == lokiv1beta1.OpenshiftLogging {
 			baseDomain, err = gateway.GetOpenShiftBaseDomain(ctx, k, req)
 			if err != nil {
-				return nil
+				return err
 			}
 
 			// extract the existing tenant's id, cookieSecret if exists, otherwise create new.
--- a/operator/internal/handlers/lokistack_create_or_update_test.go
+++ b/operator/internal/handlers/lokistack_create_or_update_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/grafana/loki/operator/internal/external/k8s/k8sfakes"
 	"github.com/grafana/loki/operator/internal/handlers"
 	"github.com/grafana/loki/operator/internal/manifests"
+	"github.com/grafana/loki/operator/internal/status"
 
 	"github.com/ViaQ/logerr/log"
 	routev1 "github.com/openshift/api/route/v1"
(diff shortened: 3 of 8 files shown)


sasagarw:
cc @periklis @xperimental

operator/controllers/lokistack_controller.go
```diff
@@ -105,15 +105,16 @@ func (r *LokiStackReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, nil
 	}
 
-	err = handlers.CreateOrUpdateLokiStack(ctx, req, r.Client, r.Scheme, r.Flags)
+	var degraded *status.DegradedCondition
+	degraded, err = handlers.CreateOrUpdateLokiStack(ctx, req, r.Client, r.Scheme, r.Flags)
```
xperimental, line 109:
```suggestion
	degraded, err := handlers.CreateOrUpdateLokiStack(ctx, req, r.Client, r.Scheme, r.Flags)
```

operator/internal/status/status.go
```diff
@@ -30,18 +30,26 @@ func Refresh(ctx context.Context, k k8s.Client, req ctrl.Request) error {
 
 	cs := s.Status.Components
 
+	if degraded != nil {
```
xperimental, line 33: Would put it before the `cs := ` assignment, so that it is clear the assignment belongs to the instructions below.

xperimental (changes requested):
As far as I can see the degraded condition is not used in any tests. Do we have no tests that create such a condition yet or is just the test for the condition missing?

As far as I can see, there are also some `SetDegradedCondition` calls remaining that would not trigger the new code-path.

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -26,17 +26,23 @@ import (
...
 
 // CreateOrUpdateLokiStack handles LokiStack create and update events.
-func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client, s *runtime.Scheme, flags manifests.FeatureFlags) error {
+func CreateOrUpdateLokiStack(
+	ctx context.Context,
+	req ctrl.Request,
+	k k8s.Client,
+	s *runtime.Scheme,
+	flags manifests.FeatureFlags,
+) (*status.DegradedCondition, error) {
```
Red-GV, line 35: Something seems strange - for lack of a better word - about returning the optional condition next to an error. It like they both signify that something has gone wrong. I wonder if it might be better to pass in the `secrets` rather than extract it in this method.

sasagarw: I don't see that as a good option. We should extract secrets here itself and pass the errors out of it.
DegradedCondition seems to be optional condition but I don't see any other naming for doing this. I will leave it this way for the time being.

operator/controllers/lokistack_controller.go
sasagarw: Since `err` is already defined before hence I can't use `:` as you suggesting. I need to create a `degraded` var or define `err` with some new variable name.

sasagarw:
@xperimental The new path won't be triggered for few because we are setting `SetDegradedCondition` directly for them for requeue purpose. For those calls, the lokistack was already running correctly and hence no need to take through new code-path.

sasagarw:
@xperimental the thing is we cannot follow the new code-path for those which needs to be requeue. Because those will never reach `status.Refresh` method and hence we should just use `SetDegradedCondition` at that place itself.

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -132,9 +134,11 @@ func TestCreateOrUpdateLokiStack_WhenGetReturnsAnErrorOtherThanNotFound_ReturnsT
 		return badRequestErr
 	}
 
-	err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, flags)
+	degraded, err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, flags)
 
 	require.Equal(t, badRequestErr, errors.Unwrap(err))
+	// make sure degraded is nil
+	require.Nil(t, degraded)
```
xperimental, line 141:
```suggestion

	require.Nil(t, degraded)
```

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -209,8 +213,10 @@ func TestCreateOrUpdateLokiStack_SetsNamespaceOnAllObjects(t *testing.T) {
 		return nil
 	}
 
-	err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, flags)
+	degraded, err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, flags)
 	require.NoError(t, err)
+	// make sure degraded is nil
+	require.Nil(t, degraded)
```
xperimental, line 219:
```suggestion

	require.Nil(t, degraded)
```

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -307,8 +313,10 @@ func TestCreateOrUpdateLokiStack_SetsOwnerRefOnAllObjects(t *testing.T) {
...
```
xperimental, line 319:
```suggestion

	require.Nil(t, degraded)
```

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -357,10 +365,12 @@ func TestCreateOrUpdateLokiStack_WhenSetControllerRefInvalid_ContinueWithOtherOb
 		return nil
 	}
 
-	err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, flags)
+	degraded, err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, flags)
 
 	// make sure error is returned to re-trigger reconciliation
 	require.Error(t, err)
+	// make sure degraded is nil
+	require.Nil(t, degraded)
```
xperimental, line 373: Usually the contract is that when an error is returned, the result of the other values is undefined. Because of that I would not check for the value of `degraded` in this case.

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -450,8 +460,10 @@ func TestCreateOrUpdateLokiStack_WhenGetReturnsNoError_UpdateObjects(t *testing.
 		return nil
 	}
 
-	err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, flags)
+	degraded, err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, flags)
 	require.NoError(t, err)
+	// make sure degraded is nil
+	require.Nil(t, degraded)
```
xperimental, line 466:
```suggestion

	require.Nil(t, degraded)
```

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -508,10 +520,12 @@ func TestCreateOrUpdateLokiStack_WhenCreateReturnsError_ContinueWithOtherObjects
 		return apierrors.NewTooManyRequestsError("too many create requests")
 	}
 
-	err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, flags)
+	degraded, err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, flags)
 
 	// make sure error is returned to re-trigger reconciliation
 	require.Error(t, err)
+	// make sure degraded is nil
+	require.Nil(t, degraded)
```
xperimental, line 528: Similar to the other case where `err != nil`, I would not check for `degraded` in this test.

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -607,10 +621,12 @@ func TestCreateOrUpdateLokiStack_WhenUpdateReturnsError_ContinueWithOtherObjects
...
```
xperimental, line 629: Similar to the other case where err != nil, I would not check for degraded in this test.

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -655,14 +674,19 @@ func TestCreateOrUpdateLokiStack_WhenMissingSecret_SetDegraded(t *testing.T) {
...
+	// make sure error is not returned to re-trigger reconciliation
 	require.NoError(t, err)
+	// make sure degraded is non-nil
+	require.NotNil(t, degraded)
 
-	// make sure status and status-update calls
-	require.NotZero(t, k.StatusCallCount())
-	require.NotZero(t, sw.UpdateCallCount())
+	require.Equal(t, msg, degraded.Message)
+	require.Equal(t, reason, degraded.Reason)
```
xperimental, line 685: Does it make sense to define a `DegradedCondition` instead and compare the whole struct in one assertion?

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -866,10 +906,12 @@ func TestCreateOrUpdateLokiStack_WhenMissingGatewaySecret_SetDegraded(t *testing
 
 	k.StatusStub = func() client.StatusWriter { return sw }
 
-	err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, ff)
+	degraded, err := handlers.CreateOrUpdateLokiStack(context.TODO(), r, k, scheme, ff)
 
 	// make sure error is returned to re-trigger reconciliation
 	require.Error(t, err)
+	// make sure degraded is nil
+	require.Nil(t, degraded)
```
xperimental, line 914:
```suggestion

	require.Nil(t, degraded)
```

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -950,10 +992,12 @@ func TestCreateOrUpdateLokiStack_WhenInvalidGatewaySecret_SetDegraded(t *testing
...
```
xperimental, line 1000:
```suggestion

	require.Nil(t, degraded)
```

operator/internal/status/lokistack.go
```diff
@@ -13,6 +13,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// DegradedCondition creates the condition Degraded to the lokistack status conditions.
+type DegradedCondition struct {
```
xperimental, line 17: I think `status.Degraded` sounds like a better type name, what do you think?
```suggestion
// Degraded contains information about why the managed LokiStack has an invalid configuration.
type Degraded struct {
```

operator/internal/handlers/lokistack_create_or_update.go
xperimental: @Red-GV we had thought about a few alternatives to the current approach. One of them was to define a custom error type that contains the information about the degraded state, but in the end we settled on just returning an additional value.

Does the custom error type sound like a better alternative to you?

operator/controllers/lokistack_controller.go
xperimental: As long as there is at least one undefined value on the left-hand side you can still use `:=`, so this change should be fine.

xperimental (changes requested):
> @xperimental the thing is we cannot follow the new code-path for those which needs to be requeue. Because those will never reach status.Refresh method and hence we should just use SetDegradedCondition at that place itself.

I think one of the goals of this PR is also to centralize the handling of the conditions. For me, it would make sense to change the handling of retries, so that the remaining degraded conditions can be handled centrally as well. What do you think?

sasagarw:
> > @xperimental the thing is we cannot follow the new code-path for those which needs to be requeue. Because those will never reach status.Refresh method and hence we should just use SetDegradedCondition at that place itself.
> 
> I think one of the goals of this PR is also to centralize the handling of the conditions. For me, it would make sense to change the handling of retries, so that the remaining degraded conditions can be handled centrally as well. What do you think?

It definitely makes sense but I am still thinking on how can we incorporate the retry logic.
The condition in retry case would be `degraded != nil && err != nil` but if requeue is triggered then it will never go to next `status.Refresh` call to `SetDegradedCondition`. :thinking:

operator/internal/handlers/lokistack_create_or_update.go
Red-GV: Yeah @xperimental, a custom error type is probably the better option in my opinion. The degraded condition already acts like its own error. So checking two things when they both say "An error has occurred which stopped the creation or updated of the objects" is kinda odd/clunky to me.

xperimental: Sometimes I just need to write the actual code to get a feel for it. Have a look: [comparison with PR](https://github.com/sasagarw/loki/compare/update-statuses...xperimental:degraded-as-err)

sasagarw: Thanks @xperimental :+1:

xperimental (changes requested):
Looks okay to me, just needs a final rebase on the latest changes.

@periklis Do you want to take a look as well? Aside from the fix to the handling of the conditions this code should behave the same as before. I'd do changes to the retries in a separate PR, if we want to do any.

periklis (commented):
Awesome :rocket:

xperimental (approved):

PR #4881, opened 2021-12-06
Add loki operator subproject

periklis:
This PR addresses the first step to upstream the [loki-operator](https://github.com/viaq/loki-operator) project into the official Loki repository as a separate sub-project. The Loki Operator subproject is going to be maintained by newly formed SIG group. More details about the subproject under:

https://docs.google.com/document/d/1dSyDwUHpFdAjdS46X4RNx9OkewY2JimRWq3Xs39LScU/edit?usp=sharing

More details to be shared on the next official community call.

**What this PR does / why we need it**:
The Loki Operator is a complementary solution to manage Loki clusters on Kubernetes and OpenShift clusters. Until now it lives in a separate repository without any attractive route to engage community members. Since Loki provides a multi-purpose mono-repo for Loki itself as well as for clients/client-plugins (e.g. fluentd, promtail, fluentbit) and utilities (e.g. logcli, loki-canary, chunks-inspect), it is the next logical step to host the operator under the same repository. The benefits of this approach are to elevate visibility and actively engage the community in designing and implementing the future of the operator as well as maintaining it. To be successful on this journey we require to set a solid repository migration path including history, a licensing definition as well as a governance process compatible to Loki’s.

**Special notes for your reviewer**:
The PR basically imports the history from one git repository into the official one. The most noteworthy changes beyond the merge commit are the last commits merging:
- Github Actions from the operator into the top-level github directory
- Add @periklis to the approve config

PR #5693, opened 2022-03-22
operator: Replace frontend_worker parallelism with match_max_concurrent

periklis:
This PR implements a long-standing recommendation to switch from using the `frontend_worker.parallelism` config option for `frontend_worker.match_max_concurrent: true` that sync the value from `querier.max_concurrent`. Later represents the upper bound of parallel queries to process per querier based on available CPU cores.

Fixes [Recommendation request](https://github.com/grafana/loki/pull/4881#discussion_r770303369)

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/operator/internal/manifests/internal/config/options.go
+++ b/operator/internal/manifests/internal/config/options.go
@@ -19,7 +19,7 @@ type Options struct {
 	Querier          Address
 	IndexGateway     Address
 	StorageDirectory string
-	QueryParallelism Parallelism
+	MaxConcurrent    MaxConcurrent
 	WriteAheadLog    WriteAheadLog
 
 	ObjectStorage storage.Options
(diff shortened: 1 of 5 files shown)


operator/internal/manifests/internal/config/options.go
```diff
@@ -46,11 +44,9 @@ type WriteAheadLog struct {
 	IngesterMemoryRequest int64
 }
 
-// Value calculates the floor of the division of
-// querier cpu limits to the query frontend replicas
-// available.
-func (p Parallelism) Value() int32 {
-	return int32(math.Floor(float64(p.QuerierCPULimits) / float64(p.QueryFrontendReplicas)))
+// Value returns the number of available CPU cores per querier.
+func (p MaxConcurrent) Value() int32 {
```
xperimental, line 48: Now that there is no calculation needed anymore, does it make sense to directly use the attribute value and get rid of this function?

If we want to keep the function:
- Do we want to bound this value to a certain minimum/maximum?
- I'm unsure of the coding style: Does it need to be moved up one block to be grouped with its type or are all methods grouped together?

xperimental (approved):

xperimental:
CLA check has failed to trigger again. A re-push usually fixes this. It would be nice if there was an easier fix, like a button to re-trigger that check.

PR #5698, opened 2022-03-22
Promtail: Fix retry/stop when erroring for out of cloudflare retention range (e.g. over 168 hours old)

### LINKED ISSUE #5696: Cloudflare stops running after 168 hours
**Describe the bug**
With the new cloudflare scraper after 168 hours it stops running as it tries to fill in log holes for older times.

The error below prints for a stream that has this issue and then it stops trying to get anything again from that zone:
`5 errors: HTTP status 400: bad query: error parsing time: invalid time range: too early: logs older than 168h0m0s are not available ; HTTP status 400: bad query: error parsing time: invalid time range: too early: logs older than 168h0m0s are not available ; HTTP status 400: bad query: error parsing time: invalid time range: too early: logs older than 168h0m0s are not available ; HTTP status 400: bad query: error parsing time: invalid time range: too early: logs older than 168h0m0s are not available ; HTTP status 400: bad query: error parsing time: invalid time range: too early: logs older than 168h0m0s are not available`

I will also note that If I clear loki it will start running again but this is not a complete solution either and retention settings to less than 168h or reject_old_samples of less don't seem to help.

**To Reproduce**
1. Started Loki 
2. Started Promtail main-4f2f088 with cloudflare scrape targets (mutiple of them)
```
          - job_name: cloudflare/cloudflare_site-test
            cloudflare:
              pull_range: 1m
              fields_type: all
              api_token: ${CLOUDFLARE_TOKEN}
              zone_id: <site.test zoneId>
              labels:
                job: cloudflare
                cloudflare_zone_id: <site.test zoneId>
                cloudflare_site_name: site.test
```
3. Query: `{job="cloudflare"}`

**Expected behavior**
Logs to be streaming in, which does work for the first few days.

**Environment:**
 - Infrastructure: Kubernetes - azure storage backed loki
 - Deployment tool: helm

paullryan:
Fixes the new cloudflare scrape in promtail for when loki tries to pull logs that are older than 168 hours, this is not allowed by the cloudflare api and right now is causing loki to stop reading logs in the future instead of ignoring and moving on.

Fixes #5696 

@cyriltovena I'm unsure if I missed a case here, please let me know.

**Checklist**
- [N/A] Documentation added
- [X] Tests updated


clients/pkg/promtail/targets/cloudflare/target.go
```diff
@@ -151,7 +151,10 @@ func (t *Target) pull(ctx context.Context, start, end time.Time) error {
 
 	for backoff.Ongoing() {
 		it, err = t.client.LogpullReceived(ctx, start, end)
-		if err != nil {
+		if err != nil && err.Error() == "HTTP status 400: bad query: error parsing time: invalid time range: too early: logs older than 168h0m0s are not available" {
```
jeschkies, line 154: Hm, could we make this a constant and maybe match it a little bit better. Is there a chance the client exposes the HTTP status?

clients/pkg/promtail/targets/cloudflare/target.go
```diff
@@ -151,7 +151,10 @@ func (t *Target) pull(ctx context.Context, start, end time.Time) error {
...
+			level.Debug(t.logger).Log("msg", "failed iterating over logs, out of cloudflare range, not retrying", "err", err, "start", start, "end", end, "retries", backoff.NumRetries())
```
jeschkies, line 155: That should probably be a warning.

jeschkies (commented):
Thanks for fixing it :slightly_smiling_face:

clients/pkg/promtail/targets/cloudflare/target.go
dannykopping: Agreed, this is a bit flimsy

paullryan: Sure I'll make a constant, I'm not sure the best match maybe just for the invalid time range: too early: logs older than 168h0m0s part... Any suggestions. As for status there are a ton of valid retry 400 error's but only this one should be fully ignored as far as I can see from the cloudflare docs.

But will add the 400 portion.

paullryan:
@dannykopping or @jeschkies not sure about the recent test failure appears to be a cleanup issue with that one run, any way to restart it without changing something in code?

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## Main
+* [5626](https://github.com/grafana/loki/pull/5696) **paullryan** Ignore and move one when cloudflare reports errors about too early logs.
```
cstyan, line 2: `move one` should be `move on`, but maybe an entry like `don't block scraping of new logs from cloudflare within promtail if an error is received from cloudflare about too early logs`? `too early` also sounds a bit off to me but from the looks of it that's the actual message they report from their end right?

clients/pkg/promtail/targets/cloudflare/target.go
```diff
@@ -28,6 +28,8 @@ import (
 // The minimun window size is 1 minute.
 const minDelay = time.Minute
 
+const cloudflareTooEarlyError = "invalid time range: too early: logs older than 168h0m0s are not available"
```
cstyan, line 31: ah maybe I misread the initial messages in this PR, I think we could just remove `too early` and say `invalid time range: logs older than 168h0m0s are not available` or `invalid start time: logs older than 168h0m0s are not available`

CHANGELOG.md
paullryan: Yes that is the language from cloudflare. Updated per your suggestion.

clients/pkg/promtail/targets/cloudflare/target.go
paullryan: Too early is the message from cloudlfare we're matching on so no cannot be anything else. I did remove invalid start time so that it is tighter.

paullryan:
@dannykopping, @cstyan  or @jeschkies anything else you need me to do to accept this?

clients/pkg/promtail/targets/cloudflare/target.go
```diff
@@ -28,6 +28,8 @@ import (
 // The minimun window size is 1 minute.
 const minDelay = time.Minute
 
+const cloudflareTooEarlyError = "too early: logs older than 168h0m0s are not available"
```
dannykopping, line 31: I'm still hesitant about including specific details here, because it's too fragile. If CF updates their max lookback period from `168h0m0s` to `168h0m1s` (a silly example), this will break. Maybe we should just have this as:

```suggestion
const cloudflareTooEarlyError = "too early: logs older than \S+ are not available"
```

and use regex to match.

That should hopefully be a bit more resilient, given that the CF client is quite flimsy already in that it doesn't give us access to any other fields besides the iterator and the err AFAICS.

paullryan: Updated to regexp as requested.

PR #5432, opened 2022-02-18
Add support for additional storage backends in operator

Red-GV:
**What this PR does / why we need it**:
This PR adds Azure, GCS, and Swift as storage options for the Loki Operator. Loki already provides support for these storage types and an easy configuration for multiple components through the `common_config`.  This PR implements the usage of a `ObjectStorageSecretType` to allow the operator to understand what type of storage it will be utilizing. Like the existing storage configuration, the human operator would need to know what components need to be in the secret for Loki to be able to access the storage.

**Which issue(s) this PR fixes**:
Fixes #4974 

**Special notes for your reviewer**:
Local File System was left out of this PR as it didn't make sense to have a path directory in a secret.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] Add an entry in the `CHANGELOG.md` about the changes.

PR #5699, opened 2022-03-22
operator: Configuring multiple storage stypes for boltdb-shipper and schema

Red-GV:
https://github.com/grafana/loki/pull/5432 added additional storage types for the installation to use. However, the PR was missing configuration changes to the `boltdb_shipper` and `schema_config` fields. This caused some components, like the ingester, to look for `S3` storage when none was specified.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/operator/internal/manifests/storage/options.go
+++ b/operator/internal/manifests/storage/options.go
@@ -1,12 +1,17 @@
 package storage
 
+import (
+	lokiv1beta1 "github.com/grafana/loki/operator/api/v1beta1"
+)
+
 // Options is used to configure Loki to integrate with
 // supported object storages.
 type Options struct {
-	Azure *AzureStorageConfig
-	GCS   *GCSStorageConfig
-	S3    *S3StorageConfig
-	Swift *SwiftStorageConfig
+	SharedStore lokiv1beta1.ObjectStorageSecretType
+	Azure       *AzureStorageConfig
+	GCS         *GCSStorageConfig
+	S3          *S3StorageConfig
+	Swift       *SwiftStorageConfig
 }
 
 // AzureStorageConfig for Azure storage config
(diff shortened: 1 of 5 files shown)


operator/internal/manifests/storage/options.go
```diff
@@ -1,12 +1,28 @@
...
+var (
+	// SharedStoreAzure storage
+	SharedStoreAzure SharedStoreOptions = "azure"
+	// SharedStoreGCS storage
+	SharedStoreGCS SharedStoreOptions = "gcs"
+	// SharedStoreS3 storage
+	SharedStoreS3 SharedStoreOptions = "s3"
+	// SharedStoreSwift storage
+	SharedStoreSwift SharedStoreOptions = "swift"
+)
```
periklis, line 16: Why do we need to replicate the same types here, as per API defs deliver the enum already? You could pass the type as an additional field into each `StorageConfig` WDYT?
```yaml
// ObjectStorageSecretType defines the type of storage which can be used with the Loki cluster.
//
// +kubebuilder:validation:Enum=azure;gcs;s3;swift
type ObjectStorageSecretType string

const (
	// ObjectStorageSecretAzure when using Azure for Loki storage
	ObjectStorageSecretAzure ObjectStorageSecretType = "azure"

	// ObjectStorageSecretGCS when using GCS for Loki storage
	ObjectStorageSecretGCS ObjectStorageSecretType = "gcs"

	// ObjectStorageSecretS3 when using S3 for Loki storage
	ObjectStorageSecretS3 ObjectStorageSecretType = "s3"

	// ObjectStorageSecretSwift when using Swift for Loki storage
	ObjectStorageSecretSwift ObjectStorageSecretType = "swift"
)
```

periklis (changes requested):

PR #5705, opened 2022-03-23
build generic index shipper to be shared between boltdb-shipper and tsdb-shipper

sandeepsukhani:
Build a generic index shipper which would be shared by `boltdb-shipper` and `tsdb-shipper` for shipping index files.
Since tsdb and boltdb based index files would have different query APIs and the way the index would be built, `index-shipper` would just manage the immutable index with following interface:
```
type Index interface {
	Name() string
	Path() string
	Close() error
	Reader() (io.ReadSeeker, error)
}
```
It has all the properties of `boltdb-shipper` like query readiness, query time downloading of index, syncs etc.
This does not have any impact yet since the code is not wired. I will slowly start wiring `boltdb-shipper` to start using `index-shipper`.

The thing that is still left to be decided is whether we will have different directories for storing boltdb and tsdb files to differentiate between type of files or we will have an extension to determine the type of file.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/indexshipper/downloads/index_set.go
+++ b/pkg/storage/stores/indexshipper/downloads/index_set.go
@@ -14,7 +14,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/grafana/dskit/concurrency"
 
-	chunk_util "github.com/grafana/loki/pkg/storage/chunk/util"
+	"github.com/grafana/loki/pkg/storage/chunk/client/util"
 	"github.com/grafana/loki/pkg/storage/stores/indexshipper/index"
 	"github.com/grafana/loki/pkg/storage/stores/shipper/storage"
 	shipper_util "github.com/grafana/loki/pkg/storage/stores/shipper/util"
@@ -59,7 +59,7 @@ func NewIndexSet(tableName, userID, cacheLocation string, baseIndexSet storage.I
 		return nil, fmt.Errorf("userID must be empty")
 	}
 
-	err := chunk_util.EnsureDirectory(cacheLocation)
+	err := util.EnsureDirectory(cacheLocation)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/storage/stores/indexshipper/downloads/index_set.go
+++ b/pkg/storage/stores/indexshipper/downloads/index_set.go
@@ -117,7 +117,7 @@ func (t *indexSet) Init() (err error) {
 		}
 
 		fullPath := filepath.Join(t.cacheLocation, fileInfo.Name())
-		// if we fail to open a boltdb file, lets skip it and let sync operation re-download the file from storage.
+		// if we fail to open an index file, lets skip it and let sync operation re-download the file from storage.
 		idx, err := t.openIndexFileFunc(fullPath)
 		if err != nil {
 			level.Error(util_log.Logger).Log("msg", fmt.Sprintf("failed to open existing index file %s, removing the file and continuing without it to let the sync operation catch up", fullPath), "err", err)
--- a/pkg/storage/stores/indexshipper/downloads/index_set.go
+++ b/pkg/storage/stores/indexshipper/downloads/index_set.go
@@ -121,9 +121,9 @@ func (t *indexSet) Init() (err error) {
 		idx, err := t.openIndexFileFunc(fullPath)
 		if err != nil {
 			level.Error(util_log.Logger).Log("msg", fmt.Sprintf("failed to open existing index file %s, removing the file and continuing without it to let the sync operation catch up", fullPath), "err", err)
-			// Sometimes files get corrupted when the process gets killed in the middle of a download operation which causes boltdb client to panic.
-			// We already recover the panic but the lock on the file is not released by boltdb client which causes the reopening of the file to fail when the sync operation tries it.
-			// We want to remove the file failing to open to get rid of the lock.
+			// Sometimes files get corrupted when the process gets killed in the middle of a download operation which can cause problems in reading the file.
+			// Implementation of openIndexFileFunc should take care of gracefully handling corrupted files.
+			// Let us just remove the file and let the sync operation re-download it.
 			if err := os.Remove(fullPath); err != nil {
 				level.Error(util_log.Logger).Log("msg", fmt.Sprintf("failed to remove index file %s which failed to open", fullPath))
 			}
--- a/pkg/storage/stores/indexshipper/downloads/index_set.go
+++ b/pkg/storage/stores/indexshipper/downloads/index_set.go
@@ -1,0 +1,361 @@
+package downloads
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/concurrency"
+
+	chunk_util "github.com/grafana/loki/pkg/storage/chunk/util"
+	"github.com/grafana/loki/pkg/storage/stores/indexshipper/index"
+	"github.com/grafana/loki/pkg/storage/stores/shipper/storage"
+	shipper_util "github.com/grafana/loki/pkg/storage/stores/shipper/util"
+	util_log "github.com/grafana/loki/pkg/util/log"
+	"github.com/grafana/loki/pkg/util/spanlogger"
+)
+
+type IndexSet interface {
+	Init() error
+	Close()
+	ForEach(ctx context.Context, callback index.ForEachIndexCallback) error
+	DropAllDBs() error
+	Err() error
+	LastUsedAt() time.Time
+	UpdateLastUsedAt()
+	Sync(ctx context.Context) (err error)
+	AwaitReady(ctx context.Context) error
+}
+
+// indexSet is a collection of multiple files created for a same table by various ingesters.
+// All the public methods are concurrency safe and take care of mutexes to avoid any data race.
+type indexSet struct {
+	openIndexFileFunc index.OpenIndexFileFunc
+	baseIndexSet      storage.IndexSet
+	tableName, userID string
+	cacheLocation     string
+	logger            log.Logger
+
+	lastUsedAt time.Time
+	index      map[string]index.Index
+	indexMtx   *mtxWithReadiness
+	err        error
+
+	cancelFunc context.CancelFunc // helps with cancellation of initialization if we are asked to stop.
+}
+
+func NewIndexSet(tableName, userID, cacheLocation string, baseIndexSet storage.IndexSet, openIndexFileFunc index.OpenIndexFileFunc,
+	logger log.Logger) (IndexSet, error) {
+	if baseIndexSet.IsUserBasedIndexSet() && userID == "" {
+		return nil, fmt.Errorf("userID must not be empty")
+	} else if !baseIndexSet.IsUserBasedIndexSet() && userID != "" {
+		return nil, fmt.Errorf("userID must be empty")
+	}
+
+	err := chunk_util.EnsureDirectory(cacheLocation)
+	if err != nil {
+		return nil, err
+	}
+
+	is := indexSet{
+		openIndexFileFunc: openIndexFileFunc,
+		baseIndexSet:      baseIndexSet,
+		tableName:         tableName,
+		userID:            userID,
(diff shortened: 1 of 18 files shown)


pkg/storage/stores/indexshipper/shipper.go
```diff
@@ -0,0 +1,165 @@
...
+			CacheDir:          s.cfg.CacheLocation,
+			SyncInterval:      s.cfg.ResyncInterval,
+			CacheTTL:          s.cfg.CacheTTL,
+			QueryReadyNumDays: s.cfg.QueryReadyNumDays,
+			Limits:            limits,
+		}
+		downloadsManager, err := downloads.NewTableManager(cfg, s.openIndexFileFunc, indexStorageClient)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
sandeepsukhani: With single binary, it is possible to run both shipper and downloader both which requires a single interface for querying.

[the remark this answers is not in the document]
sandeepsukhani: Under the hood they are separate packages, this file just provides a way to interact with both.

pkg/storage/stores/indexshipper/uploads/index_set.go
```diff
@@ -0,0 +1,251 @@
...
+		t.indexMtx.RLock()
+		indexUploadTime, ok := t.indexUploadTime[name]
+		t.indexMtx.RUnlock()
+
+		if ok && indexUploadTime.Before(cutoffTime) {
+			filesToCleanup = append(filesToCleanup, name)
+		}
+	}
+
+	t.indexMtx.RUnlock()
```
cstyan, line 203: did you mean to wrap this loop in the `RLock` call as well as use `RLock` again on the same mutex within the loop? I would think you only meant to keep one of those?

pkg/storage/stores/indexshipper/uploads/table.go
```diff
@@ -0,0 +1,137 @@
...
+type Table interface {
+	Name() string
+	AddIndex(userID string, idx index.Index) error
+	ForEach(userID string, callback index.ForEachIndexCallback) error
+	Upload(ctx context.Context) error
+	Cleanup(indexRetainPeriod time.Duration) error
+	Stop()
+}
+
+// table is a collection of multiple dbs created for a same table by the ingester.
```
cstyan, line 29: Can we add a bit more detail here, as someone new to the index and shipper part of the code base I still don't know what a table is in this context. Is the name a side effect of having used boltdb up until this point? does it refer to all indexes for a given configuration, such as by time range or tenant?

pkg/storage/stores/indexshipper/uploads/table.go
```diff
@@ -0,0 +1,137 @@
...
+	defer lt.indexSetMtx.Unlock()
+
+	for _, indexSet := range lt.indexSet {
+		indexSet.Close()
+	}
+}
+
+func loggerWithUserID(logger log.Logger, userID string) log.Logger {
+	if userID == "" {
+		return logger
```
cstyan, line 133: should we `level.Warn` log if user ID was empty?

pkg/storage/stores/indexshipper/downloads/index_set.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+// LastUsedAt returns the time at which table was last used for querying.
+func (t *indexSet) LastUsedAt() time.Time {
+	return t.lastUsedAt
+}
+
+func (t *indexSet) UpdateLastUsedAt() {
+	t.lastUsedAt = time.Now()
+}
```
cstyan, line 214: Should these use the mutex, I saw an earlier comment that all `IndexSet` public functions are thread safe but it looks like we just wrap calls to this function within this file with mutex calls, so it wouldn't be thread safe. It's a relatively minor call, given that there's no computation happening here or in `LastUsedAt`.

pkg/storage/stores/indexshipper/downloads/table.go
```diff
@@ -0,0 +1,347 @@
...
+	"github.com/grafana/loki/pkg/storage/stores/indexshipper/index"
+	"github.com/grafana/loki/pkg/storage/stores/shipper/storage"
+	util_log "github.com/grafana/loki/pkg/util/log"
+)
+
+// timeout for downloading initial files for a table to avoid leaking resources by allowing it to take all the time.
+const (
+	downloadTimeout        = 5 * time.Minute
+	maxDownloadConcurrency = 50
+)
```
cstyan, line 28: should these be configurable?

cstyan (commented):
Just some minor comments, mostly hoping to clarify some gaps in my understanding of this area of the codebase. Thanks for all the code comments, they were useful when reading this.

pkg/storage/stores/indexshipper/shipper.go
```diff
@@ -0,0 +1,165 @@
...
+	// UploadInterval defines interval for when we check if there are new index files to upload.
+	// It's also used to snapshot the currently written index tables so the snapshots can be used for reads.
+	UploadInterval = 1 * time.Minute
+)
+
+type Index interface {
+	Close() error
+}
+
+type IndexShipper interface {
```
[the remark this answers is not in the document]
sandeepsukhani: `ExternalIndex` sounds like something that is not a part of Loki but that could be just me or the naming being hard :shrug: 
I will think more but please feel free to suggest any other names.

pkg/storage/stores/indexshipper/uploads/index_set.go
sandeepsukhani: Sorry, my bad. I should be using `indexUploadTimeMtx`. This mistake showed us that you can take multiple read locks from the same goroutine which doesn't make a lot of sense.

pkg/storage/stores/indexshipper/uploads/table.go
sandeepsukhani: `table` holds index for logs within a specific interval, `24h` in case of `boltdb-shipper`. I will add more comments.

pkg/storage/stores/indexshipper/uploads/table.go
sandeepsukhani: We can add it but not sure if it will help much since it is expected to have an empty user id in the common index set.

pkg/storage/stores/indexshipper/downloads/index_set.go
sandeepsukhani: I didn't add a mutex for this because technically it is okay to concurrently overwrite a variable; it is just that we won't be sure which of the values would get assigned. Since we are not looking for a very precise last used time it is okay as long as the value gets updated. I do not have a strong opinion for this but happy to change it if you strongly feel we should add a mutex. What do you think?

pkg/storage/stores/indexshipper/downloads/table.go
sandeepsukhani: I try to avoid adding new configurations since we have more than 1000 of them now. This value has been working fine across all our cells ranging from tiny to too big. The dynamics with the `tsdb` index might change, and we can add it later?

sandeepsukhani:
Thanks for the review @cstyan!

pkg/storage/stores/indexshipper/downloads/index_set.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+
+	// open all the locally present files first to avoid downloading them again during sync operation below.
+	for _, fileInfo := range filesInfo {
+		if fileInfo.IsDir() {
+			continue
+		}
+
+		fullPath := filepath.Join(t.cacheLocation, fileInfo.Name())
+		// if we fail to open a boltdb file, lets skip it and let sync operation re-download the file from storage.
```
cyriltovena, line 120:
```suggestion
		// if we fail to open an index file, lets skip it and let sync operation re-download the file from storage.
```

pkg/storage/stores/indexshipper/downloads/index_set.go
```diff
@@ -0,0 +1,361 @@
...
+		idx, err := t.openIndexFileFunc(fullPath)
+		if err != nil {
+			level.Error(util_log.Logger).Log("msg", fmt.Sprintf("failed to open existing index file %s, removing the file and continuing without it to let the sync operation catch up", fullPath), "err", err)
+			// Sometimes files get corrupted when the process gets killed in the middle of a download operation which causes boltdb client to panic.
```
cyriltovena, line 124: Needs an update.

pkg/storage/stores/indexshipper/uploads/index_set.go
```diff
@@ -0,0 +1,251 @@
...
+		return err
+	}
+
+	err = compressedWriter.Close()
+	if err != nil {
+		return err
+	}
+
+	// flush the file to disk and seek the file to the beginning.
+	if err := f.Sync(); err != nil {
```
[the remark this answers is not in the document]
sandeepsukhani: yes, we need to pass a `ReadSeeker` to `PutFile` which unfortunately is not supported by gzip compressor. We need to compress the file first. On the read path, we decompress the file as we download but can't do it on the uploads part.

pkg/storage/stores/indexshipper/uploads/table.go
cstyan: ` it is expected to have an empty user id in the common index set.` can you elaborate? I don't understand still

pkg/storage/stores/indexshipper/downloads/index_set.go
cstyan: If, for the sake of usage of the stored value, it's not important that the value is 100% accurate, then I think not using a mutex is fine.

pkg/storage/stores/indexshipper/downloads/table.go
cstyan: `I try to avoid adding new configurations since we have more than 1000 of them now.` Fair, can we at least include in docs somewhere (if there's a spot where we describe how the shipper works) that these are the values used?

pkg/storage/stores/indexshipper/shipper.go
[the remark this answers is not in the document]
cstyan: Hmm, when/what is the `ForEach` use on the general read path? or am I misunderstanding, is the comment below only related to `ForEach` used for reads but still within the context of also eventually uploading index files?

I guess I was thinking `ExternalIndex` being the opposite of `LocalIndex` made sense but I don't have a very strong opinion on changing the name. I still don't fully understand the index shipper usage so I'll defer to you and Cyril here about what makes more sense.

pkg/storage/stores/indexshipper/shipper.go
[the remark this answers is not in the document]
cstyan: Got it, thanks for explaining!

pkg/storage/stores/indexshipper/uploads/table.go
```diff
@@ -0,0 +1,139 @@
...
+	AddIndex(userID string, idx index.Index) error
+	ForEach(userID string, callback index.ForEachIndexCallback) error
+	Upload(ctx context.Context) error
+	Cleanup(indexRetainPeriod time.Duration) error
+	Stop()
+}
+
+// table is a collection of one or more index files created by the ingester or compacted by the compactor.
+// A table would provide a logical grouping for index by a period. This period is controlled by `schema_config.configs.index.period` config.
+// It contains index for all the users sending logs to Loki.
```
cstyan, line 31: is this still correct? I thought I heard we were using per tenant indices now?

pkg/storage/stores/indexshipper/shipper.go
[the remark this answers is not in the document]
sandeepsukhani: `ForEach` is just for iterating over all the index files in a table, mostly just for reads.

pkg/storage/stores/indexshipper/uploads/table.go
sandeepsukhani: Sorry for the confusion, my point is just that it is okay to have an empty user id since it is used for managing the common index set which belongs to all the users and not a specific user.

pkg/storage/stores/indexshipper/uploads/table.go
sandeepsukhani: yeah, a table still contains index for all the users. We just build an index file per tenant which is still stored within a table.

owen-d (approved):
LGTM

PR #5706, opened 2022-03-23
Update retention.md

nvanheuverzwijn:
Retention policy actually accept suffix such as `d` and `w` but Go's time.ParseDuration does not.
See https://github.com/grafana/loki/blob/a2dd73158621441fe2d83346e792913c6770c945/pkg/storage/chunk/table_manager.go#L93

The retention_period actually accept suffix such as `d` and `w`. Go's time package ParseDuration does not. This is confusing.
No issues for that

I don't think this is worth noting in changelog but let me know if it should.

**Checklist**
- [X] Documentation added
- [ ] Tests updated


kavirajk (approved):
Thanks @nvanheuverzwijn. Much clear now :)

PR #5709, opened 2022-03-24
Add `syslog` parser as pipeline stage

chaudum:

This PR adds a parser for RFC5424 syslog messages as pipeline stage to LogQL.
This will allow users to parse raw syslog messages into labels.

Prerequisite in Promtail to push RFC5424 messages: https://github.com/grafana/loki/pull/5715

**Not to the reviewers**:

Unfortunately, the syslog parser we use does not perform very well.
 
```
$ go test ./pkg/logql/log -test.bench=Benchmark_Parser -test.run=^$
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/logql/log
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
Benchmark_Parser/json/no_labels_hints-8         	  403036	      2986 ns/op
Benchmark_Parser/json/labels_hints-8            	  549129	      2337 ns/op
Benchmark_Parser/jsonParser-not_json_line/no_labels_hints-8         	51310922	        23.52 ns/op
Benchmark_Parser/jsonParser-not_json_line/labels_hints-8            	51239817	        24.83 ns/op
Benchmark_Parser/unpack/no_labels_hints-8                           	 1317918	       845.1 ns/op
Benchmark_Parser/unpack/labels_hints-8                              	 1346008	       889.3 ns/op
Benchmark_Parser/unpack-not_json_line/no_labels_hints-8             	48373522	        26.34 ns/op
Benchmark_Parser/unpack-not_json_line/labels_hints-8                	46241360	        26.00 ns/op
Benchmark_Parser/logfmt/no_labels_hints-8                           	  778330	      1506 ns/op
Benchmark_Parser/logfmt/labels_hints-8                              	  815367	      1513 ns/op
Benchmark_Parser/regex_greedy/no_labels_hints-8                     	  418903	      2964 ns/op
Benchmark_Parser/regex_greedy/labels_hints-8                        	  412575	      3015 ns/op
Benchmark_Parser/regex_status_digits/no_labels_hints-8              	 4998464	       236.7 ns/op
Benchmark_Parser/regex_status_digits/labels_hints-8                 	 5097273	       234.4 ns/op
Benchmark_Parser/pattern/no_labels_hints-8                          	 8696990	       133.9 ns/op
Benchmark_Parser/pattern/labels_hints-8                             	10077829	       115.7 ns/op
Benchmark_Parser/syslog/no_labels_hints-8                           	  399288	      2853 ns/op
Benchmark_Parser/syslog/labels_hints-8                              	  420158	      2858 ns/op
PASS
ok  	github.com/grafana/loki/pkg/logql/log	24.625s
```
**Checklist**
- [x] Documentation added
- [x] Tests updated


dannykopping (approved):
LGTM, thanks @chaudum!

chaudum:
Need to put this PR back to draft, because I realized that when setting up the `syslog` scrape config in Promtail, only the `message` part of the syslog is pushed to Loki, making the parser stage not really useful.

chaudum:
> Need to put this PR back to draft, because I realized that when setting up the `syslog` scrape config in Promtail, only the `message` part of the syslog is pushed to Loki, making the parser stage not really useful.

I'm wondering whether there is a use-case to allow passing the whole RFC5424 formatted syslog message to be pushed to Loki from Promtail. What do you think @grafana/loki-team ?

chaudum:
Something like that:

```diff
diff --git a/clients/pkg/promtail/scrapeconfig/scrapeconfig.go b/clients/pkg/promtail/scrapeconfig/scrapeconfig.go
index 738723ec5..d5aec8811 100644
--- a/clients/pkg/promtail/scrapeconfig/scrapeconfig.go
+++ b/clients/pkg/promtail/scrapeconfig/scrapeconfig.go
@@ -183,6 +183,8 @@ type SyslogTargetConfig struct {
 	// MaxMessageLength sets the maximum limit to the length of syslog messages
 	MaxMessageLength int `yaml:"max_message_length"`
 
+	UseRFC5424Message bool `yaml:"use_rfc5424_message"`
+
 	TLSConfig promconfig.TLSConfig `yaml:"tls_config,omitempty"`
 }
 
diff --git a/clients/pkg/promtail/targets/syslog/syslogtarget.go b/clients/pkg/promtail/targets/syslog/syslogtarget.go
index bcd3443bc..c51ce0dcf 100644
--- a/clients/pkg/promtail/targets/syslog/syslogtarget.go
+++ b/clients/pkg/promtail/targets/syslog/syslogtarget.go
@@ -271,7 +271,14 @@ func (t *SyslogTarget) handleMessage(connLabels labels.Labels, msg syslog.Messag
 	} else {
 		timestamp = time.Now()
 	}
-	t.messages <- message{filtered, *rfc5424Msg.Message, timestamp}
+
+	m := *rfc5424Msg.Message
+	if t.config.UseRFC5424Message {
+		if fullMsg, err := rfc5424Msg.String(); err == nil {
+			m = fullMsg
+		}
+	}
+	t.messages <- message{filtered, m, timestamp}
 }
 
 func (t *SyslogTarget) messageSender(entries chan<- api.Entry) {
```

dannykopping:
> > Need to put this PR back to draft, because I realized that when setting up the `syslog` scrape config in Promtail, only the `message` part of the syslog is pushed to Loki, making the parser stage not really useful.
> 
> I'm wondering whether there is a use-case to allow passing the whole RFC5424 formatted syslog message to be pushed to Loki from Promtail. What do you think @grafana/loki-team ?

Sounds reasonable to me :+1: thanks for creating #5715

pkg/logql/log/parser.go
```diff
@@ -486,3 +500,76 @@ func (u *UnpackParser) unpack(it *jsoniter.Iterator, entry []byte, lbs *LabelsBu
...
+	return &SyslogParser{maxMessageLength: 1 << 12}
+}
+
+func (l *SyslogParser) Process(line []byte, lbs *LabelsBuilder) ([]byte, bool) {
+	if lbs.ParserLabelHints().NoLabels() {
+		return line, true
+	}
+	l.lbs = lbs
+	b := line[0]
+	r := bytes.NewReader(line)
```
cyriltovena, line 519: This could be in the syslogparser struct, and reused using Reset.

Which make me think can you compare this parser with other parser within the parser benchmark I'm curious how it performs.

pkg/logql/log/parser.go
```diff
@@ -486,3 +500,76 @@ func (u *UnpackParser) unpack(it *jsoniter.Iterator, entry []byte, lbs *LabelsBu
...
+	if b == '<' {
+		nontransparent.NewParser(
+			syslog.WithListener(l.handleMessage),
```
cyriltovena, line 522: syslog.WithListener(l.handleMessage) could also be part of the struct for reuse.

pkg/logql/log/parser.go
```diff
@@ -486,3 +500,76 @@ func (u *UnpackParser) unpack(it *jsoniter.Iterator, entry []byte, lbs *LabelsBu
...
+
+func (l *SyslogParser) Process(line []byte, lbs *LabelsBuilder) ([]byte, bool) {
+	if lbs.ParserLabelHints().NoLabels() {
+		return line, true
+	}
+	l.lbs = lbs
+	b := line[0]
+	r := bytes.NewReader(line)
+	if b == '<' {
+		nontransparent.NewParser(
```
cyriltovena, line 521: In fact I also think the two parser could be part of the struct, if we set a MaxMessageLength high enough ?

cyriltovena: One of the parser actually doesn't use the length.

cyriltovena: maxMessageLength: 8192, // size as per RFC5425#section-4.3.1
may be ?

cyriltovena (commented):
Looking good.

I think ultimately we may want to build our own parser, I don't know enough about the RFC but it's using a state machine and a pattern might actually work too. But let's get something working first and see about perf later.

pkg/logql/log/parser.go
chaudum: > Which make me think can you compare this parser with other parser within the parser benchmark I'm curious how it performs.

Unfortunately, it performs very badly :(

```console
$ go test ./pkg/logql/log -test.bench=Benchmark_Parser -test.run=^$
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/logql/log
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
Benchmark_Parser/json/no_labels_hints-8         	  403036	      2986 ns/op
Benchmark_Parser/json/labels_hints-8            	  549129	      2337 ns/op
Benchmark_Parser/jsonParser-not_json_line/no_labels_hints-8         	51310922	        23.52 ns/op
Benchmark_Parser/jsonParser-not_json_line/labels_hints-8            	51239817	        24.83 ns/op
Benchmark_Parser/unpack/no_labels_hints-8                           	 1317918	       845.1 ns/op
Benchmark_Parser/unpack/labels_hints-8                              	 1346008	       889.3 ns/op
Benchmark_Parser/unpack-not_json_line/no_labels_hints-8             	48373522	        26.34 ns/op
Benchmark_Parser/unpack-not_json_line/labels_hints-8                	46241360	        26.00 ns/op
Benchmark_Parser/logfmt/no_labels_hints-8                           	  778330	      1506 ns/op
Benchmark_Parser/logfmt/labels_hints-8                              	  815367	      1513 ns/op
Benchmark_Parser/regex_greedy/no_labels_hints-8                     	  418903	      2964 ns/op
Benchmark_Parser/regex_greedy/labels_hints-8                        	  412575	      3015 ns/op
Benchmark_Parser/regex_status_digits/no_labels_hints-8              	 4998464	       236.7 ns/op
Benchmark_Parser/regex_status_digits/labels_hints-8                 	 5097273	       234.4 ns/op
Benchmark_Parser/pattern/no_labels_hints-8                          	 8696990	       133.9 ns/op
Benchmark_Parser/pattern/labels_hints-8                             	10077829	       115.7 ns/op
Benchmark_Parser/syslog/no_labels_hints-8                           	  399288	      2853 ns/op
Benchmark_Parser/syslog/labels_hints-8                              	  420158	      2858 ns/op
PASS
ok  	github.com/grafana/loki/pkg/logql/log	24.625s
```

pkg/logql/log/parser.go
chaudum: I made the two parsers part of the struct.

pkg/logql/log/parser.go
cyriltovena: Do you know why :) ?

pkg/logql/log/parser.go
```diff
@@ -486,3 +500,77 @@ func (u *UnpackParser) unpack(it *jsoniter.Iterator, entry []byte, lbs *LabelsBu
...
+}
+
+func (l *SyslogParser) Process(line []byte, lbs *LabelsBuilder) ([]byte, bool) {
+	if lbs.ParserLabelHints().NoLabels() {
+		return line, true
+	}
+	l.lbs = lbs
+	l.reader.Reset(line)
+
+	if line[0] == '<' {
```
kavirajk, line 527: small nit: switch case would be bit nicer IMHO. e.g: one case for `nontransparent parser`, one for `octetcounting` parser and finally default one.

pkg/logql/log/parser_test.go
```diff
@@ -720,6 +720,88 @@ func Test_logfmtParser_Parse(t *testing.T) {
...
+		lbs  labels.Labels
+		want labels.Labels
+	}{
+		{
+			"not syslog",
+			[]byte("not a syslog"),
+			labels.Labels{
+				{Name: "foo", Value: "bar"},
+			},
+			labels.Labels{
```
kavirajk, line 736: Q: passing random string to syslog `parser.Process()` returning no error? sounds bit weird. is it expected?

pkg/logql/syntax/ast_test.go
```diff
@@ -410,6 +409,9 @@ func Test_parserExpr_Parser(t *testing.T) {
 		{"json", OpParserTypeJSON, "", log.NewJSONParser(), false},
 		{"unpack", OpParserTypeUnpack, "", log.NewUnpackParser(), false},
 		{"logfmt", OpParserTypeLogfmt, "", log.NewLogfmtParser(), false},
+		// NewSyslogParser() returns a struct pointer with pointer fields
+		// Two instances cannot be compared on equality, even reflect.DeepEqual() returns false
+		// {"syslog", OpParserTypeSyslog, "", log.NewSyslogParser(), false},
```
kavirajk, line 414: Either we uncomment the last line or remove the whole block?

kavirajk (commented):
Looks super nice 👍. Left one minor suggestions and one clarification. 
Let me know once they are addressed, I can merge.

pkg/logql/log/parser_test.go
chaudum: You're right, this should add a syslog error to the labels.

slim-bean:
I think we should think about this a little more, I'm concerned about the performance but also perhaps more concerned about what we add to the language. 

While verbose, our existing parsing tools can handle this type of line and I wonder if the solution we want for something like this should take a different form completely (leverage our existing parsers and wrap them in some kind of reusable macro?)

If we add this, we can't take it back out... I think we should sit on this longer and have some bigger conversations about the future of the query language first.

owen-d:
I'm worried about the bench comparisons to other parsers, although I think it'd be nice to have this feature. Given it's close to greedy regexp matching currently, I don't feel comfortable accepting it. I think writing our own parser could be a way forward.

darxriggs:
please keep it open

docs/sources/logql/log_queries.md
```diff
@@ -387,6 +387,30 @@ will get those labels extracted:
 "status" => "200"
 ```
 
+#### syslog
+
+The **syslog** parser can be added using the `| syslog` and will extract values from [RFC5424](https://datatracker.ietf.org/doc/html/rfc5424) formatted syslog lines.
+
+For example, parsing the following log line
```
KMiller-Grafana, line 394: Excellent addition.
These suggestions attempt to make this prose read easier, and in a similar way to the Pattern parser prose.

```suggestion
The **syslog** parser (`| syslog`) extracts values from [RFC5424](https://datatracker.ietf.org/doc/html/rfc5424)-formatted syslog lines.

Consider this syslog log line:
```

docs/sources/logql/log_queries.md
```diff
@@ -387,6 +387,30 @@ will get those labels extracted:
...
+
+```
+<165>1 2003-10-11T22:14:15.003Z mymachine.example.com evntslog - ID47 [exampleSDID@32473 iut="3" eventSource="Application" eventID="1011"] An application event log entry
+```
+
+will get those labels extracted:
```
KMiller-Grafana, line 400:
```suggestion
This example extracts these labels:
```

KMiller-Grafana (changes requested):

MasslessParticle:
We haven't updated this for about 7 months. I'm going to close it for cleanliness. Please reopen if this is in error!

PR #5715, opened 2022-03-27
Allow promtail to push RFC5424 formatted syslog messages

chaudum:

This PR introduces a new flag `use_rfc5424_message` for syslog scrape
targets, which allows to push the full
[RFC5424](https://datatracker.ietf.org/doc/html/rfc5424) formatted
syslog message received by Promtail to be pushed to Loki, instead of
only the message content.

Storing the full message allows users to parse the message in Loki
instead of relabelling internal labels in Promtail.

See also #5709 

**Checklist**
- [ ] Documentation added
- [x] Tests updated


clients/pkg/promtail/targets/syslog/syslogtarget.go
```diff
@@ -271,7 +271,14 @@ func (t *SyslogTarget) handleMessage(connLabels labels.Labels, msg syslog.Messag
 	} else {
 		timestamp = time.Now()
 	}
-	t.messages <- message{filtered, *rfc5424Msg.Message, timestamp}
+
+	m := *rfc5424Msg.Message
+	if t.config.UseRFC5424Message {
+		if fullMsg, err := rfc5424Msg.String(); err == nil {
```
dannykopping, line 277: Maybe we should log a `debug` message here stating that the full message could not be retrieved. It'd be quite confusing if we enabled this setting and it had no effect because an error was being silently discarded.

chaudum: Good point!

cyriltovena (approved):
LGTM

PR #5738, opened 2022-03-31
Docs: Fix promtail <client_config> block heading

DylanGuedes:
Fix promtail <client_config> block heading, wrongly named <client>

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/clients/promtail/configuration.md
+++ b/docs/sources/clients/promtail/configuration.md
@@ -157,10 +157,9 @@ The `server` block configures Promtail's behavior as an HTTP server:
 [health_check_target: <bool> | default = true]
 ```
 
-## clients
+## client_config
 
-The `clients` block configures how Promtail connects to instances of
-Loki:
+`client_config` configures how Promtail connects to instances of Loki:
 
 ```yaml
 # The URL where Loki is listening, denoted in Loki as http_listen_address and


docs/sources/clients/promtail/configuration.md
```diff
@@ -157,10 +157,9 @@ The `server` block configures Promtail's behavior as an HTTP server:
 [health_check_target: <bool> | default = true]
 ```
 
-## clients
+## client_config
```
trevorwhitney, line 160: Are we just changing the header to match the `<client_config>` referenced above? If so should we do the same for `server` -> `server_config`, and `positions` -> `position_config`?

docs/sources/clients/promtail/configuration.md
```diff
@@ -157,10 +157,9 @@ The `server` block configures Promtail's behavior as an HTTP server:
...
 
-The `clients` block configures how Promtail connects to instances of
-Loki:
+`client_config` configures how Promtail connects to instances of Loki:
```
trevorwhitney, line 162: I'm not sure this is correct.

The top level config keys appear to be `client` and `clients` (https://github.com/grafana/loki/blob/51fbc95bbd27b9edc584c66fd7d18256e11ec33a/clients/pkg/promtail/config/config.go#L30), so we shouldn't change that here.

trevorwhitney (changes requested):
I'm not sure this change is correct, and it breaks the pattern we have for the other sections, WDYT?

docs/sources/clients/promtail/configuration.md
DylanGuedes: Omg you right :facepalm: I got so used to the old syntax that I forgot to look at the other sections for comparison. Thank you!

docs/sources/clients/promtail/configuration.md
DylanGuedes: True, my suggestion is misleading. My intent was actually to say that the `<client_config>` type configures Promtail connection to Loki.

DylanGuedes:
> I'm not sure this change is correct, and it breaks the pattern we have for the other sections, WDYT?

You're 100% right :smile: believe I was having a brain fog when I created this lool. Closing...

PR #5746, opened 2022-04-01
[TSDB] Ensure tsdb is sorted by hash then labelset

owen-d:
Small PR to ensure that when fingerprints collide, we sort by labelset.


cyriltovena (approved):
LGTM

PR #5767, opened 2022-04-04
Loki: Add more details about usage reporting to the docs and default config files.

slim-bean:

Want to be very transparent about the new usage reporting


cyriltovena (approved):
Thanks

docs/sources/configuration/_index.md
```diff
@@ -2513,6 +2513,14 @@ This way, one doesn't have to replicate configuration in multiple places.
...
 
 ```yaml
+# By default, Loki will send anonymous but uniquely identifiable usage and configuration analytics to Grafana Labs.
+# Stats are sent to https://stats.grafana.org/
+#
+# Stats are very helpful for us to better understand how Loki is used, and what performance for most users is like.
+# This helps us prioritize features and documentation.
+# For more information on what's sent look at https://github.com/grafana/loki/blob/main/pkg/usagestats/stats.go
+# Check out the buildReport method to see what goes into a report.
+#
```
KMiller-Grafana, line 2523: I'm injecting newlines so that the rendered web pages won't need a horizontal scroll bar. And, I've reworded. If you're ok with the wording, can you propagate this change to both of the YAML example files?

```suggestion
# By default, Loki will send anonymous, but uniquely-identifiable usage and configuration
# analytics to Grafana Labs. These statistics are sent to https://stats.grafana.org/
#
# Statistics help us better understand how Loki is used, and they show us performance
# levels for most users. This helps us prioritize features and documentation.
# For more information on what's sent, look at
# https://github.com/grafana/loki/blob/main/pkg/usagestats/stats.go
# Refer to the buildReport method to see what goes into a report.
#
```

KMiller-Grafana (changes requested):

KMiller-Grafana (approved):
Thanks for the quick changes! LGTM.

PR #5770, opened 2022-04-05
Fixes deduping issues when using LogQL parser.

cyriltovena:

This PR fixes an issue when deduping logs and samples when using LogQL parser but also when streams and timestamps are equal.

The LogQL parser would extract labels and shuffle the order when batching via GRPC, however when reconstructing the order using the heap and go sorting, because the hash and the timestamp are equal, the less function would be false for all case leading to unstable sort.

Unfortunately if the sort is unstable then deduping will most likely not work, as we need to see entry tuples together.

The solution basically is to force the sort to be stable UNLESS it's a real duplicate (ts + stream hash + line) in which case we don't really care if it's unstable, since the natural order of duplicates isn't important. See the snippet below:

```diff
func (h iteratorSortHeap) Less(i, j int) bool {
        t1, t2 := h.iteratorHeap[i].Entry().Timestamp.UnixNano(), h.iteratorHeap[j].Entry().Timestamp.UnixNano()
        if t1 == t2 {
+               if h.iteratorHeap[i].StreamHash() == h.iteratorHeap[j].StreamHash() {
+                       return h.iteratorHeap[i].Entry().Line < h.iteratorHeap[j].Entry().Line
+               }
                return h.iteratorHeap[i].StreamHash() < h.iteratorHeap[j].StreamHash()
        }
```

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Main
+* [5770](https://github.com/grafana/loki/pull/5770) **cyriltovena** Fixes deduping issues when using LogQL parser.
 * [5696](https://github.com/grafana/loki/pull/5696) **paullryan** don't block scraping of new logs from cloudflare within promtail if an error is received from cloudflare about too early logs.
 * [5685](https://github.com/grafana/loki/pull/5625) **chaudum** Fix bug in push request parser that allowed users to send arbitrary non-string data as "log line".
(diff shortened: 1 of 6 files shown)


cyriltovena:
There's actually more to fix but this gets us into a better state. I will keep working on this.

cyriltovena:
Closing in favor of https://github.com/grafana/loki/pull/5799

PR #4845, opened 2021-11-30
Set correct content type header when returning 4xx errors

chaudum:
Signed-off-by: Christian Haudum <christian.haudum@gmail.com>

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] Add an entry in the `CHANGELOG.md` about the changes.

PR #5772, opened 2022-04-05
Loki: Revert #4845 which changed the format of errors from the API

slim-bean:

PR #4845 changed the return type of error messages from plain text to JSON. 

This PR was actually fixing a problem with the current API which returns plain text and sets a `Content-Type` header of `application/json` but then returns plain text. The problem is it's significantly changing the behavior of the API which should be considered a breaking change and we should wait to do this at the next major release.

It's entirely possible anyone who's implemented their own code against the Loki API are expecting the error messages in their current format and even though this is a "fix" it would break those users so we want to wait to do this until a future major release.

### DIFF
--- a/docs/sources/upgrading/_index.md
+++ b/docs/sources/upgrading/_index.md
@@ -84,22 +84,6 @@ Meanwhile, the legacy format is a string in the following format:
    [ ANNOTATIONS <label set> ]
 ```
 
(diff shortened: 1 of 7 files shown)


KMiller-Grafana:
Should the change log entry also be reverted?

PR #5775, opened 2022-04-05
golangci-lint: Check for pointers to enclosing loop variables

chaudum:
Using the pointer to the enclosing loop variable in a range loop likely
leads to unwanted behaviour. Introducing the check [`exportloopref`](https://github.com/kyoh86/exportloopref) helps
to prevent the usage of such pointers.

> The iteration variables may be declared by the "range" clause using a
> form of short variable declaration (:=). In this case their types are
> set to the types of the respective iteration values and their scope is
> the block of the "for" statement; they are re-used in each iteration.

https://go.dev/ref/spec#For_statements

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -74,6 +74,7 @@ linters:
     - unused
     - typecheck
     - depguard
+    - exportloopref
 
 issues:
   exclude:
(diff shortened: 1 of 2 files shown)


cyriltovena (approved):
LGTM

PR #5776, opened 2022-04-05
Build: update package for fuzzer after refactoring

slim-bean:
We refactored the logql into its own package and missed updating the ossfuzz packages


cyriltovena (approved):
Thanks

PR #5780, opened 2022-04-06
Update alpine images to 3.15.4

simonswine:

This updates all references to alpine images to the latest released 3.15.4

It was released yesterday to address busybox [CVE-2022-28391](https://security.alpinelinux.org/vuln/CVE-2022-28391)

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Main
+* [5780](https://github.com/grafana/loki/pull/5780) **simonswine**: Update alpine image to 3.15.4.
 * [5715](https://github.com/grafana/loki/pull/5715) **chaudum** Add option to push RFC5424 syslog messages from Promtail in syslog scrape target.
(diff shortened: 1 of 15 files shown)


simonswine:
Command used:
```
git ls-files  '*[Dd]ockerfile*' | grep -v ^vendor/ | xargs sed -i "s/alpine:[a-z0-9\\.]*/alpine:3.15.4/g"
```

slim-bean:
do we want to pin this to the `.4`? or should we just do `3.15`?

slim-bean:
I think it would be better to do `3.15` and then we would auto get the latest patch fixes

simonswine:
> I think it would be better to do `3.15` and then we would auto get the latest patch fixes

I don't think we want that. E.g if we would have built `3.15` let's say two weeks ago, how would we clearly understand that we are vulnerable to [CVE-2022-28391](https://security.alpinelinux.org/vuln/CVE-2022-28391) which was only released 2 days ago. I think we want immutable tags used as base image, so we clearly have our dependecies pinned down.

slim-bean (approved):

PR #5789, opened 2022-04-06
production config: add dot to some DNS address to reduce lookups

bboreham:

Kubernetes sets up a DNS search path like `namespace.svc.cluster.local, svc.cluster.local, cluster.local, google.internal`; any unqualified name is first tried as a prefix to each of these. These requests, which will never succeed, add latency and increase load on the DNS service.

Adding a `.` at the end makes the name fully-qualified, so Go's DNS client will skip the search path and use the name as-is.

Note this only changes a subset of addresses in the config; others didn't seem to be causing a problem and I don't have a good way to test them.

**Checklist**
- NA Documentation added
- NA Tests updated


sandeepsukhani (approved):
LGTM

PR #5790, opened 2022-04-06
Add UDP transport option to Promtail's syslog target

### LINKED ISSUE #5370: Promtail Syslog UDP transport
When using Promtail with Syslog we would like to use UDP transport to prevent promtail from becoming a potential bottleneck for syslog. Implementing UDP transport would help us with that.

We would like to see UDP transport being supported by the promtail syslog receiver.

Keep using the TCP receiver. Alternatively, we could also think about unix socket transport, so that forwarding on localhost is even faster.

**Additional context**
None.

chaudum:

Until now, the syslog target in Promtail did only support TCP. This PR adds support for sending syslog messages to Promtail (via ng-syslog or rsyslog) via UDP.

UPD for syslogs can be used if you prefer speed over guaranteed delivery.

Fixes #5370 

**Checklist**
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/clients/pkg/promtail/targets/syslog/transport.go
+++ b/clients/pkg/promtail/targets/syslog/transport.go
@@ -1,0 +1,406 @@
+package syslog
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/grafana/dskit/backoff"
+	"github.com/mwitkow/go-conntrack"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/grafana/loki/clients/pkg/promtail/scrapeconfig"
+	"github.com/grafana/loki/clients/pkg/promtail/targets/syslog/syslogparser"
+	"github.com/influxdata/go-syslog/v3"
+	"github.com/prometheus/prometheus/model/labels"
+)
+
+var (
+	protocolUDP = "udp"
+	protocolTCP = "tcp"
+)
+
+type Transport interface {
+	Run() error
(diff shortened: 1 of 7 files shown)


clients/pkg/promtail/targets/syslog/transport.go
```diff
@@ -0,0 +1,375 @@
...
+			continue
+		}
+
+		// TODO: is there a better way to pass a copy to the handleRcv() function?
+		b := make([]byte, n)
+		copy(b, buf[:n])
+		r := t.readerPool.Get().(*bytes.Reader)
+		r.Reset(b)
+		t.openConnections.Add(1)
+		go t.handleRcv(addr, r)
```
chaudum, line 344: @cyriltovena Is there a better way to do this without doing too many allocations?

I even think it does not make much difference to just pass `b` as `[]byte` to `t.handleRcv()` and create a new reader instance in the goroutine.

chaudum, line 344: @cyriltovena Is there a better way to do this without doing too many allocations?

I even think it does not make much difference to just pass `b` as `[]byte` to `t.handleRcv()` and create a new reader instance in the goroutine.

chaudum:
I still need to investigate some issues I discovered when benchmarking UDP transport, so I put the PR back to draft.

martensson:
This is really great! We are running promtail on hundreds of servers and recently activated syslog assuming UDP was supported. This will be fantastic addition and save running rsyslog (or similar software) just to forward logs.

clients/pkg/promtail/targets/syslog/transport.go
cyriltovena: taking a look

cyriltovena (approved):
LGTM

Let's resolve conflict for the changelog.

PR #5791, opened 2022-04-06
Update references to the latest build image

simonswine:

This uses the latest alpine:3.15.4 image build image with tag `0.20.0` everywhere. Esp. promtail seems to have been stuck on an older tag.


simonswine:
Sorry I thought those image references are `FROM` statements rather than `BUILD_ARGS` :roll_eyes:

PR #5800, opened 2022-04-07
Return HTTP 400 when multiple tenants are present in push.

### LINKED ISSUE #5756: `POST /loki/api/v1/push` returns an HTTP 400 when more than one tenant is specified.

jeschkies:
When multi-tenant query support is enabled we want `/loki/api/v1/push` fail when more than one tenant is configured.

Fixes #5756

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


pkg/loki/modules.go
```diff
@@ -205,6 +205,12 @@ func (t *Loki) initDistributor() (services.Service, error) {
 		logproto.RegisterPusherServer(t.Server.GRPC, t.distributor)
 	}
 
+	// If the querier module is not part of this process we need to check if multi-tenant queries are enabled.
+	// If the querier module is part of this process the querier module will configure everything.
+	if !t.Cfg.isModuleEnabled(Querier) && t.Cfg.Querier.MultiTenantQueriesEnabled {
+		tenant.WithDefaultResolver(tenant.NewMultiResolver())
+	}
```
jeschkies, line 212: I'm not really a fan of this. Ideally the multi-tenant resolver is configured globally based on the config. I actually assume that we are going to enable it by default. The only difference it makes is that `|` becomes an invalid character.

pkg/distributor/http.go
```diff
@@ -19,10 +19,15 @@ import (
 // PushHandler reads a snappy-compressed proto from the HTTP body.
 func (d *Distributor) PushHandler(w http.ResponseWriter, r *http.Request) {
 	logger := util_log.WithContext(r.Context(), util_log.Logger)
-	userID, _ := tenant.TenantID(r.Context())
-	req, err := push.ParseRequest(logger, userID, r, d.tenantsRetention)
+	tenantID, err := tenant.TenantID(r.Context())
 	if err != nil {
-		if d.tenantConfigs.LogPushRequest(userID) {
+		level.Warn(logger).Log("msg", "error getting tenant id", "err", err)
```
sandeepsukhani, line 24: should be `level.Error`?

pkg/loki/modules.go
sandeepsukhani: I am wondering if we should do it for all the modules in a commonplace. What do you think?

sandeepsukhani (approved):
Looks good to me. Just added some minor non blocking suggestions.

pkg/distributor/http.go
jeschkies: I was wondering the same. IMHO a user error should not be an error.

pkg/loki/modules.go
jeschkies: I'd like that. Where would that be?

pkg/distributor/http.go
trevorwhitney: I feel like because we return an error http status code this should be level error as well, but I could see the case it's a warn because we handle it. guess it depends what a user will go looking for if they're trying to figure out why the api is returning 400s.

pkg/loki/modules.go
trevorwhitney: agreed this gets messy, `isModuleEnabled` is just doing a string contain, so we also have to make sure we're not running in `All` or `Read` as well if we take this approach.

trevorwhitney (approved):
looks good, though I like the idea of changing how we configure the multi-tenant resolver, and not having it based on the target string.

pkg/distributor/http.go
jeschkies: Ok. I've changed the level. I think I saw `level.Error` somewhere else.

pkg/loki/modules.go
trevorwhitney: Maybe I misunderstood, what I meant to say is that the Querier is running in `Read` and `All`, but isn't detectable via the target string, so we might need to update this check to include those cases? If we took the approach of always registering this independent of whether the Querier is running or not it avoids all that messy checking.

PR #5818, opened 2022-04-07
fix missing file during rpm and deb packaging step

trevorwhitney:

The `packages` step in our Makefile has never worked as far as I can tell, this fixes it so the Circle CI release job will work.

The `-m` flag during the `zip` stage of `dist` will move the file into the resulting zip, meaning it is no longer in the `dist` folder. As a result, the `packages` step, which is looking for the compiled `logcli-linux-amd64` binary in the `dist` folder will always fail. 

This change unzips the `logcli` zip to a tmp folder so the `packages` step can find it.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -236,8 +236,11 @@ dist: clean
 	pushd dist && sha256sum * > SHA256SUMS && popd
 
 packages: dist
+	mkdir -p dist/tmp
+	unzip dist/logcli-linux-amd64.zip -d dist/tmp
 	nfpm package -f tools/nfpm.yaml -p rpm -t dist/
 	nfpm package -f tools/nfpm.yaml -p deb -t dist/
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -236,8 +236,11 @@ dist: clean
 	pushd dist && sha256sum * > SHA256SUMS && popd
 
 packages: dist
+	mkdir -p dist/tmp
+	unzip dist/logcli-linux-amd64.zip -d dist/tmp
```
jeschkies, line 240: Will this remove the zip file again?

trevorwhitney: I don't think so... but I can double check

trevorwhitney: confirmed, zip is still there

PR #5830, opened 2022-04-08
improve mutex locking when writing index in boltdb-shipper

sandeepsukhani:
We currently take an exclusive lock for each index write to check whether desired index file exists which can cause contention. This PR improves the code by checking index file exists with a read lock and then takes a write lock to create if it does not exist.

### DIFF
--- a/pkg/storage/stores/shipper/uploads/table.go
+++ b/pkg/storage/stores/shipper/uploads/table.go
@@ -226,15 +226,18 @@ func (lt *Table) MultiQueries(ctx context.Context, queries []chunk.IndexQuery, c
 }
 
 func (lt *Table) getOrAddDB(name string) (*bbolt.DB, error) {
+	lt.dbsMtx.RLock()
+	db, ok := lt.dbs[name]
+	lt.dbsMtx.RUnlock()
+
+	if ok {
+		return db, nil
+	}
+
 	lt.dbsMtx.Lock()
 	defer lt.dbsMtx.Unlock()
 
-	var (
-		db  *bbolt.DB
-		err error
-		ok  bool
-	)
-
+	var err error
 	db, ok = lt.dbs[name]
 	if !ok {
 		db, err = shipper_util.SafeOpenBoltdbFile(filepath.Join(lt.path, name))


liguozhong:
The code of this PR would be helpful. After the PR is merged, we will release boltdb-shipper again to observe the optimization effect.

liguozhong:
I am waiting for this PR to be merged, please help to speed up the merge of this PR.

liguozhong:
@sandeepsukhani Thanks for the quick response

liguozhong:
It would be great if this PR could be merged in the 2.5 release.

liguozhong:
@cyriltovena Please help to review this PR as soon as possible，thanks!

dannykopping (approved):
LGTM

sandeepsukhani:
cla stuck, will see if reopening PR fixes it.

pkg/storage/stores/shipper/uploads/table.go
```diff
@@ -226,15 +226,18 @@ func (lt *Table) MultiQueries(ctx context.Context, queries []chunk.IndexQuery, c
...
 
-	var (
-		db  *bbolt.DB
-		err error
-		ok  bool
-	)
-
+	var err error
 	db, ok = lt.dbs[name]
 	if !ok {
```
bboreham, line 242: I might invert this `if` so that the read-lock and write-lock cases look the same.  Also this would outdent the remainder of the function which is more idiomatic Go.

pkg/storage/stores/shipper/uploads/table.go
```diff
@@ -226,15 +226,18 @@ func (lt *Table) MultiQueries(ctx context.Context, queries []chunk.IndexQuery, c
...
 	lt.dbsMtx.Lock()
 	defer lt.dbsMtx.Unlock()
 
-	var (
-		db  *bbolt.DB
-		err error
-		ok  bool
-	)
-
+	var err error
```
bboreham, line 240: This could be moved inside the scope where it is used.

pkg/storage/stores/shipper/uploads/table.go
sandeepsukhani: Sorry, I merged the PR already. Will open a follow-up PR.

PR #5833, opened 2022-04-08
Refactoring of the Loki storage

cyriltovena:
This PR refactors the Loki storage so that we can extract a better interface for integrating TSDB but also a better index-gateway API.

To inject a new store now all you need is to implement those interface:

```golang
type Index interface {
	GetChunkRefs(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]logproto.ChunkRef, error)
	GetSeries(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]labels.Labels, error)
	LabelValuesForMetricName(ctx context.Context, userID string, from, through model.Time, metricName string, labelName string, matchers ...*labels.Matcher) ([]string, error)
	LabelNamesForMetricName(ctx context.Context, userID string, from, through model.Time, metricName string) ([]string, error)
	SetChunkFilterer(chunkFilter chunk.RequestChunkFilterer)
}

type ChunkWriter interface {
	Put(ctx context.Context, chunks []chunk.Chunk) error
	PutOne(ctx context.Context, from, through model.Time, chunk chunk.Chunk) error
}
```

For example this means we can easily change the index behaviour when communicating with the index-gateway to avoid sending and receiving many queries/entries.

While doing the refactoring, I had many circular dependencies issues so I had to also:
- Merge the old chunk storage. This means we have only a single storage now. Not one depending on another one.
- Improve the package architecture.
- Delete a lot of unused code.

The PR is huge but I couldn't make this works without all those refactoring, apologies for this.

I believe this will also make the code clearer as to what the storage does and why.

I wanted to do even more, but figured that was already hard enough to review.

/cc @simonswine @owen-d @sandeepsukhani @periklis


pkg/querier/queryrange/roundtrip_test.go
```diff
@@ -24,14 +24,14 @@ import (
...
-	"github.com/grafana/loki/pkg/storage/chunk"
 	"github.com/grafana/loki/pkg/storage/chunk/cache"
+	"github.com/grafana/loki/pkg/storage/config"
 	util_log "github.com/grafana/loki/pkg/util/log"
 	"github.com/grafana/loki/pkg/util/marshal"
 )
 
 var (
-	testTime   = time.Date(2019, 12, 02, 11, 10, 10, 10, time.UTC)
+	testTime   = time.Date(2019, 12, 0o2, 11, 10, 10, 10, time.UTC)
```
owen-d, line 34:
```suggestion
	testTime   = time.Date(2019, 12, 02, 11, 10, 10, 10, time.UTC)
```
Guessing you didn't mean to use an octal literal here, even though it's the same :)

owen-d (approved):
Nice work.

pkg/querier/queryrange/roundtrip_test.go
cyriltovena: LOL

PR #5840, opened 2022-04-08
usagestats tests: don't check unpredictable number

bboreham:

If this test is run alongside a number of other tests then the number of goroutines can change between when the report is generated and when the result is checked, which will cause the test to fail.

**Checklist**
- NA Documentation added
- [x] Tests updated
- NA Add an entry in the `CHANGELOG.md` about the changes.


cyriltovena (approved):
LGTM

PR #5842, opened 2022-04-08
[promtail] Mark promtail single client config flags and config file block as deprecated

cstyan:
It can be confusing to end users to try and understand the interaction between setting `client`, `clients` (in the config file), and cli flags `client.*` (such as `client.url`) as to which has priority and what will happen if multiple are set. See [this](https://grafana.slack.com/archives/CEPJRLQNL/p1649334559202859) community slack thread.

Here I am proposing that we mark the config file `client` block and CLI `client.*` flags as deprecated, including a warning log line when they are used, in preparation for removing them in a future release.

**Checklist**
- [x] Documentation added

### DIFF
--- a/docs/sources/clients/promtail/configuration.md
+++ b/docs/sources/clients/promtail/configuration.md
@@ -1797,7 +1797,7 @@ server:
 positions:
   filename: /var/log/positions.yaml # This location needs to be writeable by Promtail.
 
-client:
+clients:
   url: http://ip_or_hostname_where_Loki_run:3100/loki/api/v1/push
 
 scrape_configs:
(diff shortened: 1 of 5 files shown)


docs/sources/clients/promtail/configuration.md
```diff
@@ -1797,7 +1797,7 @@ server:
 positions:
   filename: /var/log/positions.yaml # This location needs to be writeable by Promtail.
 
-client:
+clients:
```
kavirajk, line 1800: Q: if we are chaning `client` -> `clients` then below `url` should also be changed to array?

kavirajk (commented):
LGTM. left one small comment.

docs/sources/clients/promtail/configuration.md
cstyan: yes, thanks for catching this

kavirajk (approved):

Issue #5194, opened 2022-01-20
Flush queue workers can race and flush the same chunks


slim-bean:
While investigating something else, I noticed that the metric `loki_chunk_store_deduped_chunks_total` for a Loki cluster with `replication_factor: 1` was not zero like I would expect.

The metric is incremented whenever a chunk is sent to flush but is already found in the chunk cache.

When running a replication factor > 1 the cache is used as a fast check to de-dupe chunks being sent to the store, if the chunk is in the cache it was already successfully written to the store by someone else (typically another ingester which is in the replicaset for the given stream)

For `replication_factor: 1` I would not expect to see this metric ever incremented, upon a cursory investigation it looks like the flush queue workers can race over the same chunks, we lock the list of chunks to gather a list that need to be flushed, then release the lock and attempt to flush them. It looks like this creates an opportunity for another worker to then gather the same chunk and also attempt to flush it.

The impact here is small, if you don't have a chunk cache the same chunk is sent to the store twice which is effectively a no-op (although does take some network/cpu etc), if you do have a chunk cache it's quickly found as a dupe and not sent to the store.

It does throw off the metric of how effective the dedupe code is working though which is probably the biggest impact.

This graph is from the environment I found this, the green line is the number of chunks flushed and the yellow is the deduped chunks, it's a small number that get caught in this race but we should still fix this to make it zero.

![image](https://user-images.githubusercontent.com/10332331/150364562-5c555137-fd27-4916-98c0-887d88ffa3e4.png)

JStickler:
Closing as stale. This issue was opened against a version of Loki that is no longer supported. If this issues still exists in more recent versions of Loki, please leave a comment and we can reopen the issue.

PR #5849, opened 2022-04-10
Ingester: limit max chunks per flush

afayngelerindbx:
It seems possible for Loki ingesters to build up streams that are have too many unflushed chunks to successfully flush within a single flush op timeout period. This diff caps the number of chunks that a single flush will attempt to commit to storage. Ideally, this will allow ingesters to incrementally reclaim memory after or during storage degradation.

This may address some of the reports in #5267 

**Checklist**
- [ ] Tests updated
- [ ] Metrics Added
- [ ] Documentation added

### DIFF
--- a/pkg/ingester/flush.go
+++ b/pkg/ingester/flush.go
@@ -238,11 +238,10 @@ func (i *Ingester) flushUserSeries(userID string, fp model.Fingerprint, immediat
 		return nil
 	}
 
-	chunks, labels, chunkMtx := i.collectChunksToFlush(instance, fp, immediate)
+	chunks, labels, chunkMtx, truncated := i.collectChunksToFlush(instance, fp, immediate)
 	if len(chunks) < 1 {
 		return nil
 	}
-
 	ctx := user.InjectOrgID(context.Background(), userID)
 	ctx, cancel := context.WithTimeout(ctx, i.cfg.FlushOpTimeout)
 	defer cancel()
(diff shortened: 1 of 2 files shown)


afayngelerindbx:
@slim-bean This is a very rough draft of what I was talking about in #5267 

I'm fairly new to Loki so I'd appreciate directional feedback here e.g. am I breaking some assumption/future enhancement/misunderstanding something etc.

slim-bean:
Thanks @afayngelerindbx! Let me regroup with the team a little, this is helpful in narrowing down where to look, but I'm also wondering now that this relates to #5194 and we should not only limit the number of chunks per flush, but also look at a way to make sure multiple flush workers aren't trying to flush the same chunks.

afayngelerindbx:
@slim-bean yea #5194 could definitely play a part here. It's what I was ranting about here: https://grafana.slack.com/archives/CTJ4EN6MU/p1649357960760409

I have gotten extra evidence that the issue(at least my issue) is sequential flushes. I added 

``` 
                "-store.chunks-cache.cache-stubs=true",
                "-store.chunks-cache.cache.enable-fifocache=true",
```

to the config and noticed that the cache hit rates go up dramatically before OOM

Screen Shot 2022-04-12 at 9 31 04 AM

(left is number of cache hits per second, right is used RAM in bytes)

Zooming in on the RHS graph, the cache hit rate does hit zero periodically, which seems to indicate that these cache hits are from individual flush loop iterations(e.g. not #5194). 

Screen Shot 2022-04-12 at 9 40 18 AM

Please, let me know, once ya'll decide, whether this PR is worth pursuing further. I'd love to get it merged and tested in our deployment.

PR #5861, opened 2022-04-11
Add more function comments to the rangemapper implementation

chaudum:

Helps developers understand how the rangemapper works.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/logql/rangemapper.go
+++ b/pkg/logql/rangemapper.go
@@ -37,12 +37,12 @@ var splittableRangeVectorOp = map[string]struct{}{
 // A rewrite is performed using the following rules:
 // 1) Check if query is splittable based on the range.
 // 2) Check if the query is splittable based on the query AST
-// 3) Range aggregations are split into multiple downstream range aggregations
-//    that are merged with a vector aggregation with an appropriate operator.
-//    If the range aggregation has a grouping, the grouping is also applied on
-//    the merge aggregation.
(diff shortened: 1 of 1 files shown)


pkg/logql/rangemapper.go
```diff
@@ -30,10 +30,16 @@ var splittableRangeVectorOp = map[string]struct{}{
 	syntax.OpRangeTypeMin:       {},
 }
 
+// RangeMapper is used to rewrite LogQL sample expressions into multiple
+// downstream sample expressions with a smaller time range that can be executed
+// using the downstream engine.
+// A rewrite is performed using the following rules:
```
kavirajk, line 36: > A rewrite is performed using the following rules:

but rules are missing?

kavirajk: nit: A simple example would also be very helpful IMO :) (also for `Parse` and `Map`)

kavirajk (commented):
LGTM 👍 . Need to adding missing rules I think.

pkg/logql/rangemapper.go
chaudum: 🤦‍♂️ I wanted to update the comment after writing the comments for `Map()` and `Parse()` and forgot about it.

ssncferreira (approved):
LGTM ✍️ 
Thank you @chaudum 🙏 it looks really good ❤️ 
Just a couple of minor suggestions

PR #5863, opened 2022-04-11
Docs: add schema docs

slim-bean:

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


sandeepsukhani (approved):

docs/sources/operations/storage/schema.md
```diff
@@ -0,0 +1,49 @@
+---
+title: Storage Schema
+---
+# Loki Storage Schema
+
+To support iterations over the storage layer, Loki has a configurable storage schema. The schema is defined in configuration with a `from` value that is used to mark the starting point of that schema, it is considered the active schema until another entry is inserted defining a new schema and a new `from` date.
```
KMiller-Grafana, line 6:
```suggestion
title: Storage schema
---
# Loki storage schema

To support iterations over the storage layer contents, Loki has a configurable storage schema. The schema is defined to apply over periods of time. A `from` value marks the starting point of that schema. The schema is  active until another entry defines a new schema with a new `from` date.
```

docs/sources/operations/storage/schema.md
```diff
@@ -0,0 +1,49 @@
...
+
+Be aware of using the current date and your relation to UTC to make sure UTC 00:00:00 has not already passed for the date you are using!
+
+Example, the date is 2022-04-10 and you want to update to the v12 schema so you restart Loki with 2022-04-11 as the `from` date for the new schema, but you forgot to take into account that your timezone is UTC -5:00 and it’s currently 20:00 hours in your local timezone!  Which is actually 2022-04-11T01:00:00 UTC. When Loki starts it will see the new schema and begin to write and store objects following that new schema. If you try to query data that was written between 00:00:00 and 01:00:00 UTC, however, Loki will use the new schema and the data will be unreadable because it was actually created with the previous schema.
+
+### You cannot “undo” or “rollback” a schema change.
+
+Any data written with an active schema can only be read by that schema, instead you can add another new entry with the previous schema settings if required.
```
KMiller-Grafana, line 28:
```suggestion
Loki uses the defined schemas to determine which format to use when querying the data.

Use of a schema allows Loki to iterate over the storage layer without requiring migration of existing data.

## Changing the schema

Here are items to consider when changing the schema; if not done properly, a newly created scenario can prevent data from being read.

- Always set the `from` date in the new schema to a date in the future.

    The `from` date is interpreted by Loki to start at 00:00:00 UTC. Therefore, Loki must have a date in the future to be able to transition to the new schema when that date and time arrives.

    Be aware of your relation to UTC when using the current date. Make sure that UTC 00:00:00 has not already passed for your current date.

    As an example, assume that the current date is 2022-04-10, and you want to update to the v12 schema, so you restart Loki with 2022-04-11 as the `from` date for the new schema. If you forget to take into account that your timezone is UTC -5:00 and it’s currently 20:00 hours in your local timezone,  that is actually 2022-04-11T01:00:00 UTC. When Loki starts it will see the new schema and begin to write and store objects following that new schema. If you then try to query data that was written between 00:00:00 and 01:00:00 UTC, Loki will use the new schema and the data will be unreadable, because it was created with the previous schema.

- You cannot undo or roll back a schema change.

    Any data written with an active schema can only be read by that schema. You can add another new entry with the previous schema settings, if required.
```

docs/sources/operations/storage/schema.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Examples
```
KMiller-Grafana, line 30:
```suggestion
## Schema configuration example
```

KMiller-Grafana (changes requested):

PR #5864, opened 2022-04-11
Implements TSDB Head

owen-d:
This PR, largely inspired by Prometheus' TSDB, implements a mutable TSDB Head. This can be both appended to and queried and will be used to accumulate in memory index writes in the ingester before they're flushed to disk. Querying this will also satisfy the index methods of our Store when the index hasn't been flushed to object storage yet.


cyriltovena:
Having two inverted index type in the ingester will be hard operationally why is the current index could not be adapted to flush TSDB indices ?

cyriltovena:
> This makes it easy to accumulate index writes in memory whenever the flush code `Puts` chunks, 

Currently we actually saves in a nmap files on every Puts for the index. The index file being mutable. Does that means we are not crash resilient anymore for the index if everything is in memory ? 

> and every checkpoint cycle we can serialize the accumulated values to disk.

I wonder if we actually need a disk implementation that support accumulation and out of order instead.


>We'd need to store/expire flushed chunk references for all streams according to index_cache_validity, likely in a new structure so we don't keep the entire chunk in memory.
We'd need to handle retaining references to inactive series that have been removed from ingester instances, but still have index entries that haven't been flushed to storage.

This is already solved by the current implementation.

cyriltovena (approved):
LGTM

PR #5871, opened 2022-04-11
doc: fix default table for 2.4 release

nvanheuverzwijn:
The new default for :
- `chunk_retain_period` is `0s`, not `30s`
- `chunk_idle_period` is `30m`, not `1h`
- `chunk_target_size ` is `1572864`, not `1048576`

According to https://grafana.com/docs/loki/v2.4/configuration/#ingester_config

The documentation is misleading.


owen-d (approved):
Oh, nice catch, looks like these were put in the wrong columns 😱 :
https://github.com/grafana/loki/blob/main/pkg/ingester/ingester.go#L115-L118

Thanks!

PR #5874, opened 2022-04-11
Gob vs protobuf benchmark comparison

DylanGuedes:
[This is a draft just to share to others the difference in performance between Gob vs Protobuf encoding]

Results: 
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -tags integration,requires_docker -bench ^BenchmarkCacheableIndexGobEncoding$ github.com/grafana/loki/pkg/storage/stores/tsdb

goos: linux
goarch: amd64
BenchmarkCacheableIndexGobEncoding/GetChunkRef/gob_encoding-12         	   27756	     43275 ns/op	   10136 B/op	     304 allocs/op
BenchmarkCacheableIndexGobEncoding/GetChunkRef/protbuf_encoding-12     	  125461	      9535 ns/op	    1577 B/op	      74 allocs/op
PASS
ok  	github.com/grafana/loki/pkg/storage/stores/tsdb	2.965s

```


jeschkies:
Interesting just yesterday I was wondering if gob would be an alternative.

DylanGuedes:
> Btw how do you feel about benchmarking flat buffers as well?

actually I've never heard of it :joy: but looking at the docs looks like it shouldn't be too hard to try it. I'll have a look at it this week then, let's see how it goes :eye:

jeschkies:
> actually I've never heard of it

I haven't used them myself but they [claim](https://google.github.io/flatbuffers/) to be faster and avoid copying but are bigger.

jeschkies:
While you are at it. Could you try [Cap'n Proto](https://github.com/capnproto/go-capnproto2) as well :slightly_smiling_face:

jeschkies:
@DylanGuedes  I'm curious. Did you find anything interesting? I think it's worthwhile investigating different serializers.

DylanGuedes:
I didn't, sorry. I don't even remember why I closed it :smile:

jeschkies:
No worries. Will you keep working on it? Should we create a ticket instead?

DylanGuedes:
> No worries. Will you keep working on it? Should we create a ticket instead?

I'd like to work on it, but I'll only have time once I finish the indexgateway work :(
Yes, maybe we should ticket this to not lose track of it

PR #5879, opened 2022-04-12
Add filter parameter to rebound so lines can be deleted by the compactor

MichelHollands:
Add filter functionality to the delete functionality of the compactor. This is hidden behind the "filter-and-delete" value of the deletion_mode config setting.

A filter function is added to the intervals sent to the Rebound function. This removes lines that match the given DeleteRequest.

**Checklist**
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/cmd/loki/loki-docker-config.yaml
+++ b/cmd/loki/loki-docker-config.yaml
@@ -39,14 +39,3 @@ ruler:
 # If you would like to disable reporting, uncomment the following lines:
 #analytics:
 #  reporting_enabled: false
-
-ingester:
-  max_chunk_age: 10m
-
-compactor:
-  working_directory: /loki/retention
-  compaction_interval: 10m
-  retention_enabled: true
-  deletion_mode: "filter-and-delete"
-  retention_delete_delay: 20m
-  delete_request_cancel_period: 20m
--- a/cmd/loki/loki-docker-config.yaml
+++ b/cmd/loki/loki-docker-config.yaml
@@ -39,3 +39,14 @@ ruler:
 # If you would like to disable reporting, uncomment the following lines:
 #analytics:
 #  reporting_enabled: false
+
+ingester:
+  max_chunk_age: 10m
+
+compactor:
+  working_directory: /loki/retention
+  compaction_interval: 10m
+  retention_enabled: true
+  deletion_mode: "filter-and-delete"
+  retention_delete_delay: 20m
+  delete_request_cancel_period: 20m
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -795,7 +795,8 @@ func (t *Loki) initCompactor() (services.Service, error) {
 
 	t.Server.HTTP.Path("/compactor/ring").Methods("GET", "POST").Handler(t.compactor)
 
-	if t.Cfg.CompactorConfig.RetentionEnabled && t.compactor.DeleteMode() != deletion.Disabled {
+	deleteMode := t.compactor.DeleteMode()
+	if t.Cfg.CompactorConfig.RetentionEnabled && (deleteMode == deletion.WholeStreamDeletion || deleteMode == deletion.FilterAndDelete) {
 		t.Server.HTTP.Path("/loki/api/v1/delete").Methods("PUT", "POST").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.AddDeleteRequestHandler)))
 		t.Server.HTTP.Path("/loki/api/v1/delete").Methods("GET").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.GetAllDeleteRequestsHandler)))
 		t.Server.HTTP.Path("/loki/api/v1/delete").Methods("DELETE").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.CancelDeleteRequestHandler)))
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -795,13 +795,15 @@ func (t *Loki) initCompactor() (services.Service, error) {
 
 	t.Server.HTTP.Path("/compactor/ring").Methods("GET", "POST").Handler(t.compactor)
 
-	deleteMode := t.compactor.DeleteMode()
-	if t.Cfg.CompactorConfig.RetentionEnabled && (deleteMode == deletion.WholeStreamDeletion || deleteMode == deletion.FilterAndDelete) {
-		t.Server.HTTP.Path("/loki/api/v1/delete").Methods("PUT", "POST").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.AddDeleteRequestHandler)))
-		t.Server.HTTP.Path("/loki/api/v1/delete").Methods("GET").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.GetAllDeleteRequestsHandler)))
-		t.Server.HTTP.Path("/loki/api/v1/delete").Methods("DELETE").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.CancelDeleteRequestHandler)))
-
-		t.Server.HTTP.Path("/loki/api/v1/cache/generation_numbers").Methods("GET").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.GetCacheGenerationNumberHandler)))
+	if t.Cfg.CompactorConfig.RetentionEnabled {
+		switch t.compactor.DeleteMode() {
+		case deletion.WholeStreamDeletion, deletion.FilterOnly, deletion.FilterAndDelete:
+			t.Server.HTTP.Path("/loki/api/v1/delete").Methods("PUT", "POST").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.AddDeleteRequestHandler)))
+			t.Server.HTTP.Path("/loki/api/v1/delete").Methods("GET").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.GetAllDeleteRequestsHandler)))
+			t.Server.HTTP.Path("/loki/api/v1/delete").Methods("DELETE").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.CancelDeleteRequestHandler)))
+		default:
+			break
+		}
 	}
 
 	return t.compactor, nil
--- a/pkg/storage/stores/shipper/compactor/compactor.go
+++ b/pkg/storage/stores/shipper/compactor/compactor.go
@@ -229,7 +229,8 @@ func (c *Compactor) init(storageConfig storage.Config, schemaConfig config.Schem
 			return err
 		}
 
-		if c.deleteMode == deletion.WholeStreamDeletion || c.deleteMode == deletion.FilterOnly || c.deleteMode == deletion.FilterAndDelete {
+		switch c.deleteMode {
+		case deletion.WholeStreamDeletion, deletion.FilterOnly, deletion.FilterAndDelete:
 			deletionWorkDir := filepath.Join(c.cfg.WorkingDirectory, "deletion")
 
 			c.deleteRequestsStore, err = deletion.NewDeleteStore(deletionWorkDir, c.indexStorageClient)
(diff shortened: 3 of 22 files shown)


pkg/storage/stores/shipper/compactor/deletion/delete_request.go
```diff
@@ -15,21 +17,60 @@ type DeleteRequest struct {
...
+		return nil, err
+	}
+
+	f := p.ForStream(labels).ProcessString
+	return func(s string) bool {
+		if !allMatch(d.matchers, labels) {
+			return false
+		}
+		result, _, skip := f(s)
+		if len(result) != 0 || skip {
```
MasslessParticle, line 55: we can just `return len(result) != 0 || skip`

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
```diff
@@ -15,21 +17,60 @@ type DeleteRequest struct {
...
+		}
+	}
+	p, err := d.logSelectorExpr.Pipeline()
+	if err != nil {
+		return nil, err
+	}
+
+	f := p.ForStream(labels).ProcessString
+	return func(s string) bool {
+		if !allMatch(d.matchers, labels) {
```
MasslessParticle, line 51: The labels should be stable so we probably want to do `allMatch` outside of the filter function and return a no-op if they don't match. This keeps us from matching all the labels on every line in every chungk

pkg/storage/stores/shipper/compactor/compactor.go
```diff
@@ -229,15 +229,20 @@ func (c *Compactor) init(storageConfig storage.Config, schemaConfig config.Schem
 			return err
 		}
 
-		if c.deleteMode == deletion.WholeStreamDeletion {
+		if c.deleteMode == deletion.WholeStreamDeletion || c.deleteMode == deletion.FilterAndDelete {
```
MasslessParticle, line 232: I think we need this for any mode (except disabled) so the cache invalidation can work in `FilterOnly` mode

pkg/loki/modules.go
```diff
@@ -750,7 +750,8 @@ func (t *Loki) initCompactor() (services.Service, error) {
 	t.Server.HTTP.Path("/compactor/ring").Methods("GET", "POST").Handler(t.compactor)
 
 	// TODO: update this when the other deletion modes are available
-	if t.Cfg.CompactorConfig.RetentionEnabled && t.compactor.DeleteMode() == deletion.WholeStreamDeletion {
+	deleteMode := t.compactor.DeleteMode()
+	if t.Cfg.CompactorConfig.RetentionEnabled && (deleteMode == deletion.WholeStreamDeletion || deleteMode == deletion.FilterAndDelete) {
```
MasslessParticle, line 754: Same here. The query frontend always needs this API unless deletes are disabled altogether

cmd/loki/loki-docker-config.yaml
```diff
@@ -39,3 +39,14 @@ ruler:
 # If you would like to disable reporting, uncomment the following lines:
 #analytics:
 #  reporting_enabled: false
+
```
MasslessParticle, line 42: Do we want compactors to always come with our docker config, now?

MasslessParticle (changes requested):
This looks awesome. I think there are just a couple of QOL and efficiency changes

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
MichelHollands: Indeed.

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
MichelHollands: Good idea. This was added.

pkg/storage/stores/shipper/compactor/compactor.go
MichelHollands: Updated to use the switch statement as in https://github.com/grafana/loki/pull/5922/files

pkg/loki/modules.go
MichelHollands: Updated to use the switch statement as in https://github.com/grafana/loki/pull/5922/files

cmd/loki/loki-docker-config.yaml
MichelHollands: Probably. I'll remove the compactor settings as they are set for testing and not for production.

pkg/storage/stores/shipper/compactor/deletion/validation_test.go
```diff
@@ -8,32 +8,34 @@ import (
...
 	t.Run("pipeline expression with line filter", func(t *testing.T) {
-		matchers, err := parseDeletionQuery(`{env="dev", secret="true"} |= "social sec number"`)
-		require.Nil(t, matchers)
-		require.ErrorIs(t, err, errUnsupportedQuery)
+		logSelectorExpr, err := parseDeletionQuery(`{env="dev", secret="true"} |= "social sec number"`)
+		require.NotNil(t, logSelectorExpr)
+		require.NoError(t, err)
 	})
 
+	/* syntax.ParseLogSelector does not reject these
```
MasslessParticle, line 28: Why are these commented out? If they're not useful we should probably delete them

pkg/loki/modules.go
```diff
@@ -766,11 +766,15 @@ func (t *Loki) initCompactor() (services.Service, error) {
 
 	t.Server.HTTP.Path("/compactor/ring").Methods("GET", "POST").Handler(t.compactor)
 
-	// TODO: update this when the other deletion modes are available
-	if t.Cfg.CompactorConfig.RetentionEnabled && t.compactor.DeleteMode() == deletion.WholeStreamDeletion {
-		t.Server.HTTP.Path("/loki/api/v1/delete").Methods("PUT", "POST").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.AddDeleteRequestHandler)))
-		t.Server.HTTP.Path("/loki/api/v1/delete").Methods("GET").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.GetAllDeleteRequestsHandler)))
-		t.Server.HTTP.Path("/loki/api/v1/delete").Methods("DELETE").Handler(t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.compactor.DeleteRequestsHandler.CancelDeleteRequestHandler)))
+	if t.Cfg.CompactorConfig.RetentionEnabled {
```
MasslessParticle, line 769: Do we still want these endpoints if the mode is `disabled`?

MasslessParticle (approved):
A nit about commented out code and a question about when to add the endpoints.

Otherwise LGTM

pkg/loki/modules.go
MichelHollands: I don't really see the point. If you can't delete why provide a way to add delete requests?

MasslessParticle: That makes sense. Do we ever expect both retention-enabled to be true and delete-mode to be `disabled`?

MichelHollands: The way I understand https://grafana.com/docs/loki/latest/operations/storage/retention/ is that retention removes older data based on the `retention_period` setting. This is separate from explicit delete requests. So the scenario you mention is valid. @sandeepsukhani is that correct?

MasslessParticle: Cool deal. That means we want our condition to be 

```go
if t.Cfg.CompactorConfig.RetentionEnabled && deleteMode != Disabled
```

MichelHollands: Yes, but you could argue that the current way is more future proof. If we add a new deletion mode that doesn't use the API this won't have to change. That is probably not a likely scenario though.

MichelHollands: I must be missing something here. If retention is enabled and delete mode = Disabled then the default branch in the switch statement is taken and the API is not added using `t.Server.HTTP.Path(...)`.

pkg/storage/stores/shipper/compactor/retention/retention.go
```diff
@@ -341,16 +349,16 @@ func (c *chunkRewriter) rewriteChunk(ctx context.Context, ce ChunkEntry, interva
...
+			end,
 		)
 
 		err = newChunk.Encode()
 		if err != nil {
 			return false, err
 		}
 
-		entries, err := c.seriesStoreSchema.GetChunkWriteEntries(interval.Start, interval.End, userID, "logs", newChunk.Metric, c.scfg.ExternalKey(newChunk.ChunkRef))
+		entries, err := c.seriesStoreSchema.GetChunkWriteEntries(start, end, userID, "logs", newChunk.Metric, c.scfg.ExternalKey(newChunk.ChunkRef))
```
sandeepsukhani, line 361: I think we should be using the intervals from the new chunk and not the rebound interval since with line filters we might have new chunk bounds.

sandeepsukhani (commented):
Overall changes look good to me. Left a few minor suggestions.

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
```diff
@@ -15,21 +17,60 @@ type DeleteRequest struct {
...
+	}
+	p, err := d.logSelectorExpr.Pipeline()
+	if err != nil {
+		return nil, err
+	}
+
+	if !allMatch(d.matchers, labels) {
+		return func(s string) bool {
+			return false
+		}, nil
```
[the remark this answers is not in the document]
MichelHollands: Yes, this wasn't planned to be used. In fact I was tempted to make that branch print a warning (or error) and return an error.

pkg/storage/stores/shipper/compactor/deletion/delete_request_test.go
```diff
@@ -180,3 +199,29 @@ func mustParseLabel(input string) labels.Labels {
 
 	return lbls
 }
+
+func TestDeleteRequest_FilterFunction(t *testing.T) {
+	dr := DeleteRequest{
+		Query: `{foo="bar"} |= "some"`,
```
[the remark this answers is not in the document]
MichelHollands: Good idea. It's been added.

pkg/storage/stores/shipper/compactor/retention/retention.go
MichelHollands: Done.

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
```diff
@@ -52,19 +93,30 @@ func (d *DeleteRequest) IsDeleted(entry retention.ChunkEntry) (bool, []model.Int
 		return true, nil
```
sandeepsukhani, line 93: I think we should be returning a non `nil` when the whole chunk is covered by a delete request which has a line filter? Without a filter, it would delete the whole chunk.
Let us also add a test case for that as well?

sandeepsukhani: I think you can use [HasFilter](https://github.com/grafana/loki/blob/79f0e349f56c3ec6663a3dc0457173c49a2765b4/pkg/logql/syntax/ast.go#L41) to decide whether `FilterFunc` is required or not. I think we should cover the cases of both a chunk being fully or partially deleted with/without a filter.

sandeepsukhani (commented):
Sorry, I missed a case and found a small problem. Can you please take care of that as well? Please let me know if it is not clear or if I got something wrong.

pkg/storage/stores/shipper/compactor/deletion/delete_request.go
MichelHollands: Good catch. The check using `HasFilter` has been added plus an extra test for this. 

I've also added 1 test for a partial deletion with a filter. The code didn't have to change so this 1 test will suffice hopefully.

pkg/storage/stores/shipper/compactor/deletion/delete_request_test.go
```diff
@@ -167,7 +218,16 @@ func TestDeleteRequest_IsDeleted(t *testing.T) {
...
+			for idx := range tc.expectedResp.nonDeletedIntervals {
+				require.Equal(t,
+					tc.expectedResp.nonDeletedIntervals[idx].Interval.Start,
+					nonDeletedIntervals[idx].Interval.Start,
+				)
+				require.Equal(t,
+					tc.expectedResp.nonDeletedIntervals[idx].Interval.End,
+					nonDeletedIntervals[idx].Interval.End,
+				)
+			}
```
sandeepsukhani, line 230: Let us also add a nil vs non nil check when the delete request is with `lblWithFilter`?

pkg/util/filter/filter_function.go
```diff
@@ -0,0 +1,3 @@
+package filter
+
+type Func func(string) bool
```
sandeepsukhani, line 3: Is it possible to either move this somewhere relevant like chunks or deletion package? It looks out of place to me.
The other option is to add more context to it if moving it is not possible due to circular dependency.

sandeepsukhani (approved):
Just some very minor non-blocking nits.
Nice work, the changes look great!

PR #5884, opened 2022-04-12
operator: Update Loki operand to v2.5.0

periklis:
This PR updates the loki operand from v2.4.2 to v2.5.0

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/operator/internal/manifests/internal/config/build_test.go
+++ b/operator/internal/manifests/internal/config/build_test.go
@@ -160,6 +160,8 @@ storage_config:
       server_address: dns:///loki-index-gateway-grpc-lokistack-dev.default.svc.cluster.local:9095
 tracing:
   enabled: false
+analytics:
+  reporting_enabled: false
 `
 	expRCfg := `
 ---
@@ -382,6 +384,8 @@ storage_config:
       server_address: dns:///loki-index-gateway-grpc-lokistack-dev.default.svc.cluster.local:9095
 tracing:
   enabled: false
+analytics:
+  reporting_enabled: false
 `
 	expRCfg := `
 ---
--- a/operator/internal/manifests/internal/config/build_test.go
+++ b/operator/internal/manifests/internal/config/build_test.go
@@ -161,7 +161,7 @@ storage_config:
 tracing:
   enabled: false
 analytics:
-  reporting_enabled: false
+  reporting_enabled: true
 `
 	expRCfg := `
 ---
@@ -225,6 +225,7 @@ overrides:
 				AccessKeySecret: "test123",
 			},
 		},
+		EnableRemoteReporting: true,
 	}
 	cfg, rCfg, err := Build(opts)
 	require.NoError(t, err)
--- a/operator/internal/manifests/internal/config/build_test.go
+++ b/operator/internal/manifests/internal/config/build_test.go
@@ -45,6 +45,7 @@ ingester:
   chunk_idle_period: 1h
   chunk_retain_period: 5m
   chunk_target_size: 2097152
+  flush_op_timeout: 10m
   lifecycler:
     final_sleep: 0s
     heartbeat_period: 5s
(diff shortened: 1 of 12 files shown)


operator/main.go
```diff
@@ -64,8 +65,10 @@ func main() {
 	flag.BoolVar(&enableGateway, "with-lokistack-gateway", false,
 		"Enables the manifest creation for the entire lokistack-gateway.")
 	flag.BoolVar(&enableGatewayRoute, "with-lokistack-gateway-route", false,
-		"Enables the usage of Route for the lokistack-gateway instead of Ingress (OCP Only!)")
-	flag.BoolVar(&enablePrometheusAlerts, "with-prometheus-alerts", false, "Enables prometheus alerts")
+		"Enables the usage of Route for the lokistack-gateway instead of Ingress (OCP Only!).")
+	flag.BoolVar(&enablePrometheusAlerts, "with-prometheus-alerts", false, "Enables prometheus alerts.")
+	flag.BoolVar(&enableGrafanaLabsStats, "with-grafana-labs-stats", false,
```
slim-bean, line 70: Would you consider enabling this by default? Nobody ever opts in to this kind of thing and setting the default to false essentially means we'll never get stats from anyone using the operator :( 

It's really helpful to us to know how people are using Loki with this kind of functionality to be able to steer the direction of the project.

periklis: Sure this came in by default typing to `false` accidentally.

slim-bean (approved):
LGTM!

periklis:
@slim-bean FYI, i have not added the schema-config for V12 in this PR, because from an operator perspective it is not easy to know when users will opt-in upgrading the operator and thus have a good value for the `from` field. I am currently thinking to pursue in one of the two following directions:
1. **Full automation**: After an operator upgrade the operator can check the current config map and make an entry into the `schema_config` for V12 starting e.g. on the next day. This is a bit tricky because we need to get the UTC math correct and other considerations (e.g. disconnected clusters w/o time sync, shutdown clusters)
2. **Opt-In automation**: The user will get an entry-point in the API to select which schema he/she want to upgrade next and when. This is simpler and persistent with the custom resource on etcd side.

WDYT?

operator/internal/manifests/internal/config/build_test.go
```diff
@@ -159,6 +160,8 @@ storage_config:
       server_address: dns:///loki-index-gateway-grpc-lokistack-dev.default.svc.cluster.local:9095
 tracing:
   enabled: false
+analytics:
+  reporting_enabled: false
```
xperimental, line 164: Should we enable the reporting in the test, so that we see the field updated (false=default)?

operator/main.go
```diff
@@ -64,8 +65,10 @@ func main() {
 	flag.BoolVar(&enableGateway, "with-lokistack-gateway", false,
 		"Enables the manifest creation for the entire lokistack-gateway.")
 	flag.BoolVar(&enableGatewayRoute, "with-lokistack-gateway-route", false,
-		"Enables the usage of Route for the lokistack-gateway instead of Ingress (OCP Only!)")
-	flag.BoolVar(&enablePrometheusAlerts, "with-prometheus-alerts", false, "Enables prometheus alerts")
+		"Enables the usage of Route for the lokistack-gateway instead of Ingress (OCP Only!).")
+	flag.BoolVar(&enablePrometheusAlerts, "with-prometheus-alerts", false, "Enables prometheus alerts.")
+	flag.BoolVar(&enableGrafanaLabsStats, "with-grafana-labs-stats", true,
```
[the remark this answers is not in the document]
periklis: I wanted to make `Grafana Labs` as the collecting entity visible and `--with-grafana-labs-analytics` seem to me quite long.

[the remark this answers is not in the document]
periklis: Changed this to `--with-grafana-labs-analytics`

PR #5888, opened 2022-04-12
loki: Fix common config net interface name overwritten by ring common config

Papawy:
In the case where `instance_interface_names` is declared in the common
configuration AND the common ring configuration is also present,
`instance_interface_names` from common configuration is overwritten by
default ring configuration for all ring components.

This behavior is not the one described in documentation. 

This commit fix this issue by overwriting common ring net interface
config IF it is deeply equal to the default one.

Example:

```
common:
  instance_interface_names:
  - interface
  ring:
    kvstore:
      store: inmemory
```

Then (example with Distributor configuration),
```
Distributor.DistributorRing.InstanceInterfaceNames == defaults.Common.Ring.InstanceInterfaceNames
```

but should be

```
Distributor.DistributorRing.InstanceInterfaceNames == []string{"interface"}
```

None.

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/loki/config_wrapper.go
+++ b/pkg/loki/config_wrapper.go
@@ -121,20 +121,16 @@ func (c *ConfigWrapper) ApplyDynamicConfig() cfg.Source {
 // - "instance-interface-names", a list of net interfaces used when looking for addresses.
 func applyInstanceConfigs(r, defaults *ConfigWrapper) {
 	if !reflect.DeepEqual(r.Common.InstanceAddr, defaults.Common.InstanceAddr) {
-		r.Ingester.LifecyclerConfig.Addr = r.Common.InstanceAddr
-		r.CompactorConfig.CompactorRing.InstanceAddr = r.Common.InstanceAddr
-		r.Distributor.DistributorRing.InstanceAddr = r.Common.InstanceAddr
-		r.Ruler.Ring.InstanceAddr = r.Common.InstanceAddr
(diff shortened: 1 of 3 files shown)


DylanGuedes:
Hey thanks for catching this, we noticed that this also happened for the replication factor so we're fixing this for the replication factor on another PR.

pkg/loki/config_wrapper.go
```diff
@@ -121,20 +121,16 @@ func (c *ConfigWrapper) ApplyDynamicConfig() cfg.Source {
 // - "instance-interface-names", a list of net interfaces used when looking for addresses.
 func applyInstanceConfigs(r, defaults *ConfigWrapper) {
 	if !reflect.DeepEqual(r.Common.InstanceAddr, defaults.Common.InstanceAddr) {
-		r.Ingester.LifecyclerConfig.Addr = r.Common.InstanceAddr
-		r.CompactorConfig.CompactorRing.InstanceAddr = r.Common.InstanceAddr
-		r.Distributor.DistributorRing.InstanceAddr = r.Common.InstanceAddr
-		r.Ruler.Ring.InstanceAddr = r.Common.InstanceAddr
-		r.QueryScheduler.SchedulerRing.InstanceAddr = r.Common.InstanceAddr
+		if reflect.DeepEqual(r.Common.Ring.InstanceAddr, defaults.Common.Ring.InstanceAddr) {
```
DylanGuedes, line 124: hey this was really smart :smile:

kavirajk (approved):
LGMT 👍

PR #5358, opened 2022-02-09
Add a Ring to IndexGateway

JordanRushing:
Signed-off-by: Jordan Rushing <jordan.rushing@grafana.com>

**What this PR does / why we need it**:

This PR introduces a ring to Loki's `IndexGateway` to provide better horizontal scalability via per-tenant hashing with changes to the associated Index files provided by per-tenant compaction introduced in #5026 which will reduce overall startup time for new `IndexGateway` instances.

`IndexGateway` now includes a new configuration, `RingMode`. When this mode is enabled each `IndexGateway` instance becomes responsible for a subset of tenant indices.

In PRs to follow, we will update `TableManager` to reduce the amount of tenant data downloaded and add metrics.

**Which issue(s) this PR fixes**:

N/A

**Special notes for your reviewer**:

N/A

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] Add an entry in the `CHANGELOG.md` about the changes.

PR #5892, opened 2022-04-12
Introduce a more scalable index-gateway API.

cyriltovena:
Because of the storage refactoring (https://github.com/grafana/loki/pull/5833) we can now introduce a better
index-gateway API that fits more our index interface.

Such as :

```proto
    rpc GetChunkRef(GetChunkRefRequest) returns (GetChunkRefResponse) {};
    rpc LabelNamesForMetricName(LabelNamesForMetricNameRequest) returns (LabelResponse)  {};
    rpc LabelValuesForMetricName(LabelValuesForMetricNameRequest) returns (LabelResponse) {};
```

This will avoid sending thousands of index queries to the index-gateway but instead them just a single request.
The index caching, parsing and filtering is now happening all on the index-gateway side.

Loki queriers will first check if the new API exists before using it, this way update can be done transparently.
However the check happens only on startup, so if you want to start using the new API you need to restart queriers after fully rolling out index-gateways.

CC @sandeepsukhani @simonswine @slim-bean 

Note: Now that I think about it if one index gateway is rolled out the querier might think they all are. So I might want to test ALL IPs, instead of a random one, assuming I can do that.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -404,7 +404,7 @@ func (t *Loki) initStore() (_ services.Service, err error) {
 			// and queried as part of live data until the cache TTL expires on the index entry.
 			t.Cfg.Ingester.RetainPeriod = t.Cfg.StorageConfig.IndexCacheValidity + 1*time.Minute
 			t.Cfg.StorageConfig.BoltDBShipperConfig.IngesterDBRetainPeriod = boltdbShipperQuerierIndexUpdateDelay(t.Cfg) + 2*time.Minute
-		case t.Cfg.isModuleEnabled(Querier), t.Cfg.isModuleEnabled(Ruler), t.Cfg.isModuleEnabled(Read):
+		case t.Cfg.isModuleEnabled(Querier), t.Cfg.isModuleEnabled(Ruler), t.Cfg.isModuleEnabled(Read), t.isModuleActive(IndexGateway):
 			// We do not want query to do any updates to index
 			t.Cfg.StorageConfig.BoltDBShipperConfig.Mode = shipper.ModeReadOnly
 		default:
@@ -425,6 +425,8 @@ func (t *Loki) initStore() (_ services.Service, err error) {
 			// Use AsyncStore to query both ingesters local store and chunk store for store queries.
 			// Only queriers should use the AsyncStore, it should never be used in ingesters.
 			asyncStore = true
+		case t.Cfg.isModuleEnabled(IndexGateway):
+			t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Address = ""
 		case t.Cfg.isModuleEnabled(All):
 			// We want ingester to also query the store when using boltdb-shipper but only when running with target All.
 			// We do not want to use AsyncStore otherwise it would start spiraling around doing queries over and over again to the ingesters and store.
@@ -760,18 +762,14 @@ func (t *Loki) initCompactor() (services.Service, error) {
 }
 
 func (t *Loki) initIndexGateway() (services.Service, error) {
-	t.Cfg.StorageConfig.BoltDBShipperConfig.Mode = shipper.ModeReadOnly
-	objectClient, err := storage.NewObjectClient(t.Cfg.StorageConfig.BoltDBShipperConfig.SharedStoreType, t.Cfg.StorageConfig, t.clientMetrics)
+	cfg := t.Cfg.StorageConfig
+	cfg.BoltDBShipperConfig.IndexGatewayClientConfig.Address = ""
+
+	indexClient, err := storage.NewIndexClient(config.BoltDBShipperType, cfg, t.Cfg.SchemaConfig, t.overrides, t.clientMetrics, prometheus.DefaultRegisterer)
 	if err != nil {
 		return nil, err
 	}
-
-	shipperIndexClient, err := shipper.NewShipper(t.Cfg.StorageConfig.BoltDBShipperConfig, objectClient, t.overrides, prometheus.DefaultRegisterer)
-	if err != nil {
-		return nil, err
-	}
-
-	gateway := indexgateway.NewIndexGateway(shipperIndexClient)
+	gateway := indexgateway.NewIndexGateway(t.Store, indexClient)
 	indexgatewaypb.RegisterIndexGatewayServer(t.Server.GRPC, gateway)
 	return gateway, nil
 }
--- a/pkg/storage/store.go
+++ b/pkg/storage/store.go
@@ -25,6 +25,7 @@ import (
 	"github.com/grafana/loki/pkg/storage/stores"
 	"github.com/grafana/loki/pkg/storage/stores/series"
 	"github.com/grafana/loki/pkg/storage/stores/series/index"
+	"github.com/grafana/loki/pkg/storage/stores/shipper"
 	"github.com/grafana/loki/pkg/usagestats"
 	"github.com/grafana/loki/pkg/util"
 )
@@ -196,8 +197,30 @@ func (s *store) storeForPeriod(p config.PeriodConfig, chunkClient client.Client,
 		schema = index.NewSchemaCaching(schema, time.Duration(s.storeCfg.CacheLookupsOlderThan))
 	}
 
-	return series.NewWriter(f, s.schemaCfg, idx, schema, s.writeDedupeCache, s.storeCfg.DisableIndexDeduplication),
-		series.NewIndexStore(s.schemaCfg, schema, idx, f, s.cfg.MaxChunkBatchSize),
+	var (
+		writer       stores.ChunkWriter = series.NewWriter(f, s.schemaCfg, idx, schema, s.writeDedupeCache, s.storeCfg.DisableIndexDeduplication)
+		seriesdIndex *series.IndexStore = series.NewIndexStore(s.schemaCfg, schema, idx, f, s.cfg.MaxChunkBatchSize)
+		index        stores.Index       = seriesdIndex
+	)
+
+	if s.cfg.BoltDBShipperConfig.Mode == shipper.ModeReadOnly && s.cfg.BoltDBShipperConfig.IndexGatewayClientConfig.Address != "" {
+
+		hasRefsAPI, err := shipper.HasGetRefsAPI(s.cfg.BoltDBShipperConfig.IndexGatewayClientConfig)
+		if err != nil {
+			return nil, nil, nil, err
+		}
+		if hasRefsAPI {
+			gw, err := shipper.NewGatewayClient(s.cfg.BoltDBShipperConfig.IndexGatewayClientConfig, indexClientReg)
+			if err != nil {
+				return nil, nil, nil, err
+			}
+			index = series.NewIndexGatewayClientStore(gw, seriesdIndex)
+		}
+
+	}
+
+	return writer,
+		index,
 		func() {
 			chunkClient.Stop()
 			f.Stop()
--- a/pkg/storage/stores/series/series_index_gateway_store.go
+++ b/pkg/storage/stores/series/series_index_gateway_store.go
@@ -1,0 +1,72 @@
+package series
+
+import (
+	"context"
+
+	"github.com/prometheus/common/model"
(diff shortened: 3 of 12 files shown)


periklis (commented):
Excellent :rocket:

pkg/storage/store.go
```diff
@@ -196,8 +197,30 @@ func (s *store) storeForPeriod(p config.PeriodConfig, chunkClient client.Client,
...
+		if err != nil {
+			return nil, nil, nil, err
+		}
+		if hasRefsAPI {
+			gw, err := shipper.NewGatewayClient(s.cfg.BoltDBShipperConfig.IndexGatewayClientConfig, indexClientReg)
+			if err != nil {
+				return nil, nil, nil, err
+			}
+			index = series.NewIndexGatewayClientStore(gw, seriesdIndex)
+		}
```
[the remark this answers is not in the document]
cyriltovena: Thanks this is a great suggestions, I'll merge your commit if you don't mind and add some tests.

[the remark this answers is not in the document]
simonswine: I was playing around with a test like this maybe that can also be useful: https://github.com/grafana/loki/commit/c50f2df33bf8e015d37199606326808b02bc3a98

(should have really committed in the first place)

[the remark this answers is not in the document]
cyriltovena: on it.

periklis (commented):
Do I get this right, if the RPC methods are not implemented yet, we cascade to the index store? Nice!

periklis:
> > Do I get this right, if the RPC methods are not implemented yet, we cascade to the index store? Nice!
> 
> Yes so we don't need an option or to rollout in order. The index-store uses the old index-gateway queries API.

This kind of handling is a huge plus for rolling out via the operator. Anything we can go to make rollouts order-agnostic is a huge win for keeping the operator complexity as low as possible

pkg/storage/store.go
```diff
@@ -196,8 +197,22 @@ func (s *store) storeForPeriod(p config.PeriodConfig, chunkClient client.Client,
...
-	return series.NewWriter(f, s.schemaCfg, idx, schema, s.writeDedupeCache, s.storeCfg.DisableIndexDeduplication),
-		series.NewIndexStore(s.schemaCfg, schema, idx, f, s.cfg.MaxChunkBatchSize),
+	var (
+		writer       stores.ChunkWriter = series.NewWriter(f, s.schemaCfg, idx, schema, s.writeDedupeCache, s.storeCfg.DisableIndexDeduplication)
+		seriesdIndex *series.IndexStore = series.NewIndexStore(s.schemaCfg, schema, idx, f, s.cfg.MaxChunkBatchSize)
+		index        stores.Index       = seriesdIndex
+	)
+	if s.cfg.BoltDBShipperConfig.Mode == shipper.ModeReadOnly && s.cfg.BoltDBShipperConfig.IndexGatewayClientConfig.Address != "" {
+		// inject the index-gateway client into the index store
+		gw, err := shipper.NewGatewayClient(s.cfg.BoltDBShipperConfig.IndexGatewayClientConfig, indexClientReg)
```
sandeepsukhani, line 207: Wouldn't `idx` already be a gateway client instance created by `NewIndexClient` [here](https://github.com/JordanRushing/loki/blob/main/pkg/storage/factory.go#L148)?

cyriltovena: Yes but a different one. This one wrap the index-store, the other one call the old API.

sandeepsukhani:
> Note: Now that I think about it if one index gateway is rolled out the querier might think they all are. So I might want to test ALL IPs, instead of a random one, assuming I can do that.

In non ring mode I think we don't have to worry about it since we point to a service, right? With the ring mode, it is already taken care of [here](https://github.com/JordanRushing/loki/blob/b118248d26bd743598d8ae64bf9f4c6346afbe46/pkg/storage/stores/shipper/gateway_client.go#L238).

Sorry for the trouble, it seems you need to rebase your PR since I merged #5358

pkg/storage/store.go
sandeepsukhani: yeah, but that part is done in `IndexGatewayClientStore`, right?
you can type cast `idx` to `*shipper.GatewayClient` and pass it to `NewIndexGatewayClientStore` like below:
`series.NewIndexGatewayClientStore(idx.(*shipper.GatewayClient), seriesdIndex)`

cyriltovena:
> ow that I think about it if one index gateway is rolled out the querier might think they all are. So I might want to test ALL IPs, instead of a random one, assuming I

We decided to go ahead with doing a call each time, since this is a temporary cluster state.

pkg/storage/store.go
cyriltovena: it might work but this is more risky.

cyriltovena:
it's ok, I was able to merge, can you verify what I did is fine ?

sandeepsukhani (approved):

pkg/storage/stores/series/series_index_gateway_store.go
```diff
@@ -0,0 +1,101 @@
...
+	response, err := c.client.GetChunkRef(ctx, &indexgatewaypb.GetChunkRefRequest{
+		From:     from,
+		Through:  through,
+		Matchers: (&syntax.MatchersExpr{Mts: allMatchers}).String(),
+	})
+	if err != nil {
+		if isUnimplementedCallError(err) {
+			return c.IndexStore.GetChunkRefs(ctx, userID, from, through, allMatchers...)
+		}
+		return nil, err
```
simonswine, line 45: Nit: Maybe it is worth commenting on those statements like:

```
 // Handle communication with older index gateways gracefully, by falling back to the index store calls.
```

simonswine (approved):
LGTM

pkg/storage/stores/shipper/indexgateway/indexgatewaypb/gateway.proto
```diff
@@ -2,10 +2,46 @@ syntax = "proto3";
...
+    rpc LabelNamesForMetricName(LabelNamesForMetricNameRequest) returns (LabelResponse)  {};
+    rpc LabelValuesForMetricName(LabelValuesForMetricNameRequest) returns (LabelResponse) {};
+}
+
+message LabelValuesForMetricNameRequest {
+    string metric_name = 1;
+    string label_name = 2;
+    int64 from = 3 [ (gogoproto.customtype) = "github.com/prometheus/common/model.Time", (gogoproto.nullable) = false];
+    int64 through = 4 [ (gogoproto.customtype) = "github.com/prometheus/common/model.Time", (gogoproto.nullable) = false];
+    string matchers = 5;
```
DylanGuedes, line 24: double checking: shouldn't this be a repeated matcher?

pkg/storage/stores/shipper/indexgateway/gateway.go
```diff
@@ -269,6 +290,62 @@ func buildResponses(query index.Query, batch index.ReadBatchResult, callback fun
...
+func (g *Gateway) LabelValuesForMetricName(ctx context.Context, req *indexgatewaypb.LabelValuesForMetricNameRequest) (*indexgatewaypb.LabelResponse, error) {
+	instanceID, err := tenant.TenantID(ctx)
+	if err != nil {
+		return nil, err
+	}
+	matchers, err := syntax.ParseMatchers(req.Matchers)
+	if err != nil {
+		return nil, err
+	}
+	names, err := g.indexQuerier.LabelValuesForMetricName(ctx, instanceID, req.From, req.From, req.MetricName, req.LabelName, matchers...)
```
DylanGuedes, line 340: double checking
```suggestion
	names, err := g.indexQuerier.LabelValuesForMetricName(ctx, instanceID, req.To, req.From, req.MetricName, req.LabelName, matchers...)
```

pkg/storage/stores/shipper/indexgateway/gateway.go
```diff
@@ -269,6 +290,62 @@ func buildResponses(query index.Query, batch index.ReadBatchResult, callback fun
...
+	}
+	return result, nil
+}
+
+func (g *Gateway) LabelNamesForMetricName(ctx context.Context, req *indexgatewaypb.LabelNamesForMetricNameRequest) (*indexgatewaypb.LabelResponse, error) {
+	instanceID, err := tenant.TenantID(ctx)
+	if err != nil {
+		return nil, err
+	}
+	names, err := g.indexQuerier.LabelNamesForMetricName(ctx, instanceID, req.From, req.From, req.MetricName)
```
DylanGuedes, line 322: ditto
```suggestion
	names, err := g.indexQuerier.LabelNamesForMetricName(ctx, instanceID, req.From, req.Through, req.MetricName)
```

pkg/storage/stores/shipper/indexgateway/gateway.go
cyriltovena: 🤦

pkg/storage/stores/shipper/indexgateway/indexgatewaypb/gateway.proto
cyriltovena: We;re passing the stringified version.

pkg/loki/modules.go
```diff
@@ -437,6 +437,9 @@ func (t *Loki) initStore() (_ services.Service, err error) {
 			// Use AsyncStore to query both ingesters local store and chunk store for store queries.
 			// Only queriers should use the AsyncStore, it should never be used in ingesters.
 			asyncStore = true
+		case t.Cfg.isModuleEnabled(IndexGateway):
+			// we want to use the actual storage when running the index-gateway, so we remove the Addr from the config
+			t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Address = ""
```
DylanGuedes, line 442: I didn't get this part - doesn't it basically means that we can't run boltdbshipper with an Index Gateway that has no ring?

DylanGuedes (approved):
a few comments but LGTM!

pkg/storage/store.go
```diff
@@ -196,8 +197,22 @@ func (s *store) storeForPeriod(p config.PeriodConfig, chunkClient client.Client,
...
 	}
 
-	return series.NewWriter(f, s.schemaCfg, idx, schema, s.writeDedupeCache, s.storeCfg.DisableIndexDeduplication),
-		series.NewIndexStore(s.schemaCfg, schema, idx, f, s.cfg.MaxChunkBatchSize),
+	var (
+		writer       stores.ChunkWriter = series.NewWriter(f, s.schemaCfg, idx, schema, s.writeDedupeCache, s.storeCfg.DisableIndexDeduplication)
+		seriesdIndex *series.IndexStore = series.NewIndexStore(s.schemaCfg, schema, idx, f, s.cfg.MaxChunkBatchSize)
+		index        stores.Index       = seriesdIndex
+	)
+	if s.cfg.BoltDBShipperConfig.Mode == shipper.ModeReadOnly && s.cfg.BoltDBShipperConfig.IndexGatewayClientConfig.Address != "" {
```
sandeepsukhani, line 205: We need an additional check when running index gateway in ring mode since there won't be an address set. Can you please add this change?
```suggestion
	if s.cfg.BoltDBShipperConfig.Mode == shipper.ModeReadOnly &&
			(s.cfg.BoltDBShipperConfig.IndexGatewayClientConfig.Address != "" || s.cfg.BoltDBShipperConfig.IndexGatewayClientConfig.Ring != nil) {
```

pkg/loki/modules.go
sandeepsukhani: It is just clearing the address so that `NewIndexClient` initializes `boltdb-shipper` index client instead of index gateway client.

pkg/storage/store.go
sandeepsukhani: Dylan has already taken care of the other check in PR https://github.com/grafana/loki/pull/5902

pkg/loki/modules.go
sandeepsukhani: Basically we do not want to initialize index gateway client from index gateway.

DylanGuedes: ohhhh got it

PR #5893, opened 2022-04-12
operator: Align PVC storage size requests for all lokistack t-shirt sizes

periklis:
This PR aligns the PVC storage size requests of `1x.extra-small` with the requests of `1x.small` and `1x.medium`. This allows seamless upgrades of all stateful Loki components w/o the need to implement an particular migration handler in the operator. Since `1x.extra-small` is used only for demostration purposes, it would have been an overkill to write a handler that cleans up PVCs and statefulsets before re-creating the in the next bigger t-shirt size.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Main
 
+- [5893](https://github.com/grafana/loki/pull/5893) **periklis**: Align PVC storage size requests for all lokistack t-shirt sizes
 - [5884](https://github.com/grafana/loki/pull/5884) **periklis**: Update Loki operand to v2.5.0
 - [5748](https://github.com/grafana/loki/pull/5748) **Red-GV**: Update Prometheus go client to 12.1
 - [5739](https://github.com/grafana/loki/pull/5739) **sasagarw**: Change UUIDs to tenant name in doc
--- a/operator/internal/manifests/internal/sizes.go
+++ b/operator/internal/manifests/internal/sizes.go
@@ -36,7 +36,7 @@ var ResourceRequirementsTable = map[lokiv1beta1.LokiStackSizeType]ComponentResou
 			},
 		},
 		Ingester: ResourceRequirements{
-			PVCSize: resource.MustParse("1Gi"),
+			PVCSize: resource.MustParse("10Gi"),
 			Requests: map[corev1.ResourceName]resource.Quantity{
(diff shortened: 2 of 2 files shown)


xperimental:
Should there even be a migration path from `extra-small` to one of the other sizes?

I'm wondering if announcing >220G storage need for a small test deployment might be overkill. It seems to be irrelevant for `kind` because it does not actually try to allocate the space and this might apply to many other testing tools as well.

But if an environment actually tries to allocate the space this will probably rule out the "local testing" use-case that extra-small currently occupies as well as the demonstration target.

periklis:
> Should there even be a migration path from `extra-small` to one of the other sizes?
> 
> I'm wondering if announcing >220G storage need for a small test deployment might be overkill. It seems to be irrelevant for `kind` because it does not actually try to allocate the space and this might apply to many other testing tools as well.
> 
> But if an environment actually tries to allocate the space this will probably rule out the "local testing" use-case that extra-small currently occupies as well as the demonstration target.

IMHO the `1x.extra-small` started out as a target for `kind` only, but people use it on their stage clusters to test out Loki for the first time. Once they have a feel how things work, they just feel eager to bump to `1x.small` after allocating enough CPU and Memory. In addition the advertisement drum "we need only object storage for Loki" does its best to make people forgetful about PVC sizes which are used for temporary local storage on the query side or the WAL on the ingester side.

In summary I suggest we should provide users (currently early-adopters) an easy way to create a first stage cluster of `1x.extra-small` and then bump it to `1x.small | 1x.medium` when they have the resources. Either instructing them to delete the CR and PVC first or implement automation for such in the operator is IMHO a waste of time right now. Thus biting on the larger PVC requirement was for me the cost/benefit winner.

cyriltovena (approved):

xperimental:
> In summary I suggest we should provide users (currently early-adopters) an easy way to create a first stage cluster of `1x.extra-small` and then bump it to `1x.small | 1x.medium` when they have the resources. Either instructing them to delete the CR and PVC first or implement automation for such in the operator is IMHO a waste of time right now. Thus biting on the larger PVC requirement was for me the cost/benefit winner.

I suppose there will be a time when we need to think about the automatic migration between different PVC configurations but now is probably not that time. I'll keep my fingers crossed that this will not have a negative impact on testing, but so far it seems fine.

xperimental (approved):

PR #5894, opened 2022-04-12
Loki: Flush chunks one at a time

DylanGuedes:
This modifies our ingesters flushing strategy to flush one chunk at a time instead of flushing them in a bulk. This has two benefits/fixes in relation to the current implementation:
1. Imagine we have 250 chunks to flush, but chunk 125 fails to flush. In the previous implementation, we would first encode all 250 chunks, but flushing chunk 125 would fail so the time encoding the last 125 chunks would be a waste
2. Flushes are marked with `cs[i].flushed = time.Now()`. In the previous implementation, a chunk flush could only be marked if all other flushes also succeeded. This makes it more granular, helping chunks to not be flushed multiple times.

This PR also refactor related functions into smaller pieces.

N/A

For reference, see https://github.com/grafana/loki/issues/5267#issuecomment-1096919667.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/ingester/flush.go
+++ b/pkg/ingester/flush.go
@@ -276,11 +276,12 @@ func (i *Ingester) collectChunksToFlush(instance *instance, fp model.Fingerprint
 			}
 			// Flush this chunk if it hasn't already been successfully flushed.
 			if stream.chunks[j].flushed.IsZero() {
-				result = append(result, &stream.chunks[j])
 				if immediate {
 					reason = flushReasonForced
 				}
-				chunksFlushedPerReason.WithLabelValues(reason).Add(1)
+				stream.chunks[j].reason = reason
+
+				result = append(result, &stream.chunks[j])
 			}
 		}
 	}
@@ -373,7 +374,7 @@ func (i *Ingester) flushChunks(ctx context.Context, fp model.Fingerprint, labelP
 			lastTime,
 		)
 
-		if err := i.encodeChunk(ch, c); err != nil {
+		if err := i.encodeChunk(ctx, ch, c); err != nil {
 			return err
 		}
 
@@ -383,7 +384,14 @@ func (i *Ingester) flushChunks(ctx context.Context, fp model.Fingerprint, labelP
 
 		i.markChunkAsFlushed(cs[j], chunkMtx)
 
-		i.reportFlushedChunkStatistics(ch, c, sizePerTenant, countPerTenant)
+		reason := func() string {
+			chunkMtx.Lock()
+			defer chunkMtx.Unlock()
+
+			return c.reason
+		}()
+
+		i.reportFlushedChunkStatistics(ch, c, sizePerTenant, countPerTenant, reason)
 	}
 
 	return nil
@@ -410,7 +418,10 @@ func (i *Ingester) closeChunk(desc *chunkDesc, chunkMtx sync.Locker) error {
 //
 // If the encoding is unsuccessful the flush operation is reinserted in the queue which will cause
 // the encoding for a given chunk to be evaluated again.
-func (i *Ingester) encodeChunk(ch chunk.Chunk, desc *chunkDesc) error {
+func (i *Ingester) encodeChunk(ctx context.Context, ch chunk.Chunk, desc *chunkDesc) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
 	start := time.Now()
 	chunkBytesSize := desc.chunk.BytesSize() + 4*1024 // size + 4kB should be enough room for cortex header
 	if err := ch.EncodeTo(bytes.NewBuffer(make([]byte, 0, chunkBytesSize))); err != nil {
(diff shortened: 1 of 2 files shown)


owen-d (commented):
Nice work! This isn't easy code and it's a great surprise to wake up and see a bug fix for this. I've left one suggestion around locking, then LGTM

pkg/ingester/flush.go
```diff
@@ -352,87 +357,100 @@ func (i *Ingester) flushChunks(ctx context.Context, fp model.Fingerprint, labelP
...
+			chunkenc.NewFacade(c.chunk, i.cfg.BlockSize, i.cfg.TargetChunkSize),
+			firstTime,
+			lastTime,
+		)
 
-		numEntries := cs[i].chunk.Size()
-		byt, err := wc.Encoded()
-		if err != nil {
-			continue
+		if err := i.encodeChunk(ch, c); err != nil {
```
afayngelerindbx, line 380: Chunk encoding can take time and memory. I'd recommend checking whether `ctx` has timed out before doing it.

afayngelerindbx:
I'm also not sure that this is in scope for this PR(let me know if I should open a PR to do this after this one is merged):

We're incrementing `chunksFlushedPerReason` in `collectChunksToFlush`. This means that even if a particular chunk isn't flushed or even encoded, we count it in that metric. This is very misleading when debugging flush throughput e.g. the metric says that flushes are happening when, in fact, they are backlogged. It'd be nice to move that reporting into `reportFlushedChunkStatistics` 

Should we also measure the number of chunks that were not flushed?

pkg/ingester/flush.go
```diff
@@ -352,87 +357,100 @@ func (i *Ingester) flushChunks(ctx context.Context, fp model.Fingerprint, labelP
...
+			return err
 		}
 
-		compressedSize := float64(len(byt))
-		uncompressedSize, ok := chunkenc.UncompressedSize(wc.Data)
-
-		if ok && compressedSize > 0 {
-			chunkCompressionRatio.Observe(float64(uncompressedSize) / compressedSize)
+		if err := i.flushChunk(ctx, ch); err != nil {
```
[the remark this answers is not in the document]
DylanGuedes: Thanks for catching this. Pushed a commit addressing it, and now I'm checking how it is going.

pkg/ingester/flush.go
DylanGuedes: WDYT of this?
```go
func (i *Ingester) encodeChunk(ctx context.Context, ch chunk.Chunk, desc *chunkDesc) error {
		return err
	}
	start := time.Now()
	chunkBytesSize := desc.chunk.BytesSize() + 4*1024 // size + 4kB should be enough room for cortex header
	if err := ch.EncodeTo(bytes.NewBuffer(make([]byte, 0, chunkBytesSize))); err != nil {
		return fmt.Errorf("chunk encoding: %w", err)
	}
	chunkEncodeTime.Observe(time.Since(start).Seconds())
	return nil
}
```

slim-bean: This feels like an unnecessary optimization to me, at most you would save encoding one chunk I believe

DylanGuedes:
> I'm also not sure that this is in scope for this PR(let me know if I should open a PR to do this after this one is merged):
> 
> We're incrementing `chunksFlushedPerReason` in `collectChunksToFlush`. This means that even if a particular chunk isn't flushed or even encoded, we count it in that metric. This is very misleading when debugging flush throughput e.g. the metric says that flushes are happening when, in fact, they are backlogged. It'd be nice to move that reporting into `reportFlushedChunkStatistics`
> 
> Should we also measure the number of chunks that were not flushed?

I'm pushing a commit addressing this problem with `chunksFlushedPerReason`, lmk what you think.

Regarding measuring the number of chunks not flushed, I think we can derive it from the total number of chunks (I think it is `ingester_chunks_stored_total`) minus the number of flushed chunks

pkg/ingester/flush.go
```diff
@@ -352,87 +357,100 @@ func (i *Ingester) flushChunks(ctx context.Context, fp model.Fingerprint, labelP
...
+	return nil
+}
+
+// flushChunk flushes the given chunk to the store.
+//
+// If the flush is successful, metrics for this flush are to be reported.
+// If the flush isn't successful, the operation for this userID is requeued allowing this and all other unflushed
+// chunk to have another opportunity to be flushed.
+func (i *Ingester) flushChunk(ctx context.Context, ch chunk.Chunk) error {
+	if err := i.store.Put(ctx, []chunk.Chunk{ch}); err != nil {
```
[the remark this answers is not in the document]
afayngelerindbx: Ah I got confused with the different store interfaces. Definitely not worth adding an interface method if it isn't there already.

pkg/ingester/flush.go
afayngelerindbx: This allocates up to sightly more than the target chunk size(>1MB by default) and in benchmarks encoding takes 10s of ms per stream so definitely seemed worth the extra 3LOC.

slim-bean (approved):
LGTM!

afayngelerindbx:
> Regarding measuring the number of chunks not flushed, I think we can derive it from the total number of chunks (I think it is `ingester_chunks_stored_total`) minus the number of flushed chunks

So `ingester_chunks_stored_total` is broken down by tenant while `ingester_chunks_flushed_total` is broken down by reason. So you can't really get a per-tenant count of "unflushed" chunks.  Also, with this change as it stands, the total aggregated value of these counters should be the same since both are incremented once in `reportFlushedChunkStatistics` 

Oof rationalizing these seems like a bigger change(probably out of scope here). Maybe revert the change to move the `flushed` metric. I'm sorry I've caused confusion.

pkg/ingester/flush.go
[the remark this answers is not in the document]
afayngelerindbx: Ah I got confused with the different store interfaces. Definitely not worth adding an interface method if it isn't there already.

slim-bean:
I think moving the flushed metric still makes sense to be consistent on only updating the metric if the chunk is actually flushed. 

There is also `cortex_ingester_flush_queue_length` for tracking the queues to see if they get backed up. It's not per tenant but I'm not sure you really need to know about per tenant stats on flushing as all the chunks are going to the same place?

afayngelerindbx:
> There is also `cortex_ingester_flush_queue_length` for tracking the queues to see if they get backed up. It's not per tenant but I'm not sure you really need to know about per tenant stats on flushing as all the chunks are going to the same place?

flush queue length tracks the number of *streams* enqueued to be flushed. That number can stay flat while ingester is falling behind because a large number of chunks is going unflushed.  Breaking down by tenant would be preferable since that is how one would apply limits/shed load, I think.

Definitely outside the scope of this patch. Glad this was merged. Going to test the new version soon. Thanks @slim-bean and @DylanGuedes for getting this in so quickly!

PR #5902, opened 2022-04-13
Loki: Don't use boltdb-shipper if gateway communication isn't available

DylanGuedes:
This adds a new condition to our conditional clause that checks if boltdb-shipper should be used for the index or not.

Without this check, there's a chance of boltdb being enabled even if no index gateway ring or index gateway address are available, which would lead to problems as no communication with the index gateway would be possible.

N/A

This was previously suggested by Sandeep on https://github.com/grafana/loki/pull/5514#discussion_r828977931, we applied it but somehow lost this during a rebase.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


sandeepsukhani (approved):
LGTM

PR #5914, opened 2022-04-13
Reduces mutex locking during flushes

owen-d:
This is done by iterating a few times, but should be negligible with small n-values (we don't flush many chunks at a time as it's a per stream operation).

Now we lock/unlock two times for a set of chunks, rather than twice for _every_ chunk.

Followup to https://github.com/grafana/loki/pull/5894

### DIFF
--- a/pkg/ingester/flush.go
+++ b/pkg/ingester/flush.go
@@ -437,19 +437,6 @@ func encodeChunk(ctx context.Context, ch chunk.Chunk, desc *chunkDesc) error {
 	return nil
 }
 
-// flushChunk flushes the given chunk to the store.
-//
-// If the flush is successful, metrics for this flush are to be reported.
-// If the flush isn't successful, the operation for this userID is requeued allowing this and all other unflushed
-// chunk to have another opportunity to be flushed.
-func (i *Ingester) flushChunk(ctx context.Context, ch chunk.Chunk) error {
-	if err := i.store.Put(ctx, []chunk.Chunk{ch}); err != nil {
-		return fmt.Errorf("store put chunk: %w", err)
-	}
-	flushedChunksStats.Inc(1)
-	return nil
-}
-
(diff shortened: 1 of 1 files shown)


pkg/ingester/flush.go
```diff
@@ -361,64 +361,70 @@ func (i *Ingester) flushChunks(ctx context.Context, fp model.Fingerprint, labelP
...
-		if err := i.encodeChunk(ctx, ch, c); err != nil {
-			return err
+	}
+	chunkMtx.Unlock()
+
+	// Without the lock, encode & store
+	flushTimes := make([]struct {
+		t  time.Time
+		ch chunk.Chunk
+	}, 0, len(cs))
```
slim-bean, line 378: This PR kind of feels like trading CPU for memory with the side effect of more complexity/harder to maintain.  

The overhead of calling a mutex lock/unlock a few hundred or even a few thousand times feels better to me than allocating a slice over all the chunks which I would wonder has a higher garbage collection CPU overhead than the mutex calls.  I think if we want to go this route it might be better to store a pointer in this slice than copy the chunk.

It also feels a little more brittle to me now that future maintainers have to closely track this lock instead of relying on `defer` to unlock.

I guess I'm not convinced this would actually save an CPU and I think it makes the code harder to maintain

cyriltovena:
You should write a benchmark to measure the improvement

owen-d:
I'm not sure if it's worth the time/complexity to bench this. There's a lot of stateful code + I/O here and the main reason I wanted to reduce lock contention was to benefit reads on the ingesters. If we were to bench this, I think the two variables should be `concurrent_readers` and `n_chunks`. Since there isn't clear consensus and I don't see myself writing a benchmark, I'll close this.

pkg/ingester/flush.go
owen-d: Fair enough :+1:

PR #5918, opened 2022-04-13
allow read only components to *not* configure an index gateway client…

afayngelerindbx:
…, if they so choose

This PR fixes a recently introduced bug. The bug being fixed does not allow loki to run read path components without an index gateway client.

This is due to a nil check in https://github.com/grafana/loki/blob/435391661f4e0f03eb378f77c07aaad5009cd6e9/pkg/storage/store.go#L205 that returns true when a nil struct typed reference is assigned to an interface reference.

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -423,7 +423,14 @@ func (t *Loki) initStore() (_ services.Service, err error) {
 	}
 
 	t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Mode = t.Cfg.IndexGateway.Mode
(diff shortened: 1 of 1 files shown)


cyriltovena:
me and @DylanGuedes where looking at this. Thanks let's see

cyriltovena:
I think it works, we might need to reintroduce the nil check in the store. Let's me test it and push commit if that works.

Thanks for helping btw.

cyriltovena:
Great spot thank you so much for the hint, I've merged another PR to solve this but again thanks for helping, appreciated.

afayngelerindbx:
Thanks for fixing! Your patch looks much cleaner.

PR #5920, opened 2022-04-14
support custom label selector in kubernetes_sd_configs

hoozecn:

make promtail scrape logs of pod with speciifc field or label selector, so that it will watch less logs and make fewer calculations and get a better performance in some scenarios.

for example, we only want to scape the pods in a specific namespace or with specific label in other namespaces

```
      - job_name: lcc-system-pods
        pipeline_stages:
          {{- toYaml .Values.config.snippets.pipelineStages | nindent 4 }}
        kubernetes_sd_configs:
          - role: pod
            selectors:
            - role: pod
              field: metadata.namespace=lcc-system
        relabel_configs: ....
     -  job_name: lcc-pods-not-in-lcc-system
        pipeline_stages:
          {{- toYaml .Values.config.snippets.pipelineStages | nindent 4 }}
        kubernetes_sd_configs:
          - role: pod
            selectors:
            - role: pod
              field: metadata.namespace!=lcc-system
              label: app.lccomputing.com/product=lcc
        relabel_configs: ....
```

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/clients/promtail/configuration.md
+++ b/docs/sources/clients/promtail/configuration.md
@@ -1547,18 +1547,20 @@ namespaces:
   names:
     [ - <string> ]
 
-# Optional label and field selectors to limit the discovery process to a subset of available resources.
-# See https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/
-# and https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ to learn more about the possible
-# filters that can be used. The endpoints role supports pod, service and endpoints selectors.
-# Roles only support selectors matching the role itself (e.g. node role can only contain node selectors).
-
-# Note: When making decision about using field/label selector make sure that this
-# is the best approach - it will prevent Promtail from reusing single list/watch
-# for all scrape configs. This might result in a bigger load on the Kubernetes API,
-# because per each selector combination there will be additional LIST/WATCH. On the other hand,
-# if you just want to monitor small subset of pods in large cluster it's recommended to use selectors.
-# Decision, if selectors should be used or not depends on the particular situation.
+# Optional label and field selectors to limit the discovery process to a subset of available
+#  resources. See
+# https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/
+# and https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ to learn
+# more about the possible filters that can be used. The endpoints role supports pod,
+# service, and endpoint selectors. Roles only support selectors matching the role itself;
+# for example, the node role can only contain node selectors.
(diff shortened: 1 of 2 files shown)


cyriltovena (commented):
Looks good should we document it ?

hoozecn:
> Looks good should we document it ?

yeah sure. will do it asap

hoozecn:
> Looks good should we document it ?

Make a copy from https://github.com/prometheus/prometheus/blob/main/docs/configuration/configuration.md

hoozecn:
@cyriltovena If there any other reason block this pr, please let me know :)

docs/sources/clients/promtail/configuration.md
```diff
@@ -1498,6 +1498,23 @@ tls_config:
...
+# and https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ to learn more about the possible
+# filters that can be used. The endpoints role supports pod, service and endpoints selectors.
+# Roles only support selectors matching the role itself (e.g. node role can only contain node selectors).
+
+# Note: When making decision about using field/label selector make sure that this
+# is the best approach - it will prevent Promtail from reusing single list/watch
+# for all scrape configs. This might result in a bigger load on the Kubernetes API,
+# because per each selector combination there will be additional LIST/WATCH. On the other hand,
+# if you just want to monitor small subset of pods in large cluster it's recommended to use selectors.
+# Decision, if selectors should be used or not depends on the particular situation.
```
KMiller-Grafana, line 1513: Line length needs to be such that we don't need a horizontal scroll bar.

```suggestion
# Optional label and field selectors to limit the discovery process to a subset of available
#  resources. See
# https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/
# and https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ to learn
# more about the possible filters that can be used. The endpoints role supports pod,
# service, and endpoint selectors. Roles only support selectors matching the role itself;
# for example, the node role can only contain node selectors.
# Note: When making decisions about using field/label selectors, make sure that this
# is the best approach. It will prevent Promtail from reusing single list/watch
# for all scrape configurations. This might result in a bigger load on the Kubernetes API,
# because for each selector combination, there will be additional LIST/WATCH.
# On the other hand, if you want to monitor a small subset of pods of a large cluster,
# we recommend using selectors. The decision on the use of selectors or not depends
# on the particular situation.
```

KMiller-Grafana (changes requested):

hoozecn:
@KMiller-Grafana Changes applied

MichelHollands (approved):
LGTM

PR #5923, opened 2022-04-14
Fixes the index-gateway when using the ring

cyriltovena:
As spotted in #5918 the index-gw should not use the index-gw client but the local storage instead.

I'm taking another path than relying on the ring, but instead use a new config.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


slim-bean (approved):
LGTM

PR #5925, opened 2022-04-14
Other: Improve check-drone-drift script

chaudum:

Instead of checking the line count of the current branch an main, run
`make drone` and check if there are changes to the target
`.drone/drone.yml` file.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/tools/check-drone-drift.sh
+++ b/tools/check-drone-drift.sh
@@ -4,18 +4,29 @@ set -uo pipefail
 
 command -v drone >/dev/null 2>&1 || { echo "drone is not installed"; exit 1; }
 
-TARGET_BRANCH="$1"
 DRONE_JSONNET_FILE=".drone/drone.jsonnet"
 DRONE_CONFIG_FILE=".drone/drone.yml"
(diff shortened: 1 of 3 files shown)


tools/check-drone-drift.sh
```diff
@@ -4,18 +4,25 @@ set -uo pipefail
...
-      exit 1
-    fi
-  fi
+drone jsonnet \
+    --stream \
+    --stdout \
+    --format=false \
+    --source "${DRONE_JSONNET_FILE}" \
+    --target "${DRONE_CONFIG_FILE}" \
+    | diff "${DRONE_CONFIG_FILE}" -
```
jeschkies, line 17: I did try this in the past but had some issues :shrug: If this works, great!

jeschkies (approved):

tools/check-drone-drift.sh
chaudum: The problem was that the check-drone-drift script did not sign the output.

kavirajk (approved):
Didn't have much context. but changes LGTM 👍

PR #5932, opened 2022-04-14
Loki: Fix issue with recent change to flush code

slim-bean:
#5894 changed how we flush chunks but had a small bug which resulted in the checksum and some other values not being set properly because we passed a copy of the chunk and not a pointer to it and were mutating it.


owen-d (approved):
I think we should change these others to also use pointers:
https://github.com/grafana/loki/blob/main/pkg/ingester/flush.go#L439-L448

PR #5933, opened 2022-04-15
[promtail] stream lag delete fix

cstyan:
_Note: to reproduce the bug you can run promtail on your laptop and have a single static config that reads a local log file. Delete that file, and then see if the metric still exists on `/metrics`, it should. With the changes in this PR the metric will properly be deleted when you delete that local log file._

So because of the way the metric is registered now we couldn't delete the metric for a file stream that went away in the same way. There's a PR open in the prometheus client_golang that would allow us to do this in a nicer way (delete based on a partial label set), but for now this should work. Alternatively we would refactor the filestream code to track this metric instead of it being part of the client code. 

There was another issue related to the actual `InstrumentedEntryHandler` interface. I had a hard time fully grokking the way the client(s) are passed around between the promtail client code, and then used as `api.EntryHandler` in the file target and target manager code. If I understand correctly, because of another change I made which deprecated the individual `config.Client` and only now allow for multiple clients via `config.Clients`, we now always have a `MultiClient` that's being passed. For some reason `MultiClient`, even with `UnregisterLatencyMetric` implemented, would never match the type assertion for `t.handler.(api.InstrumentedEntryHandler)`, and so we would never call the function to delete the metric. By refactoring the interfaces so that `EntryHandler` and `LatencyMetricHandler` are separate parameters to `NewFileTarget` we shouldn't run into this same bug again.

The commits should be standalone (at least starting from [8d99c61](https://github.com/grafana/loki/pull/5933/commits/8d99c611dfdd39c5e142857ec2da8aa0b140dc76)) and could be read in order for easier reviewing, though some of your comments may be resolved along the way.

CC @MasslessParticle

### DIFF
--- a/clients/pkg/promtail/client/client.go
+++ b/clients/pkg/promtail/client/client.go
@@ -465,6 +465,12 @@ func (c *client) UnregisterLatencyMetric(labels prometheus.Labels) bool {
 		level.Debug(c.logger).Log("msg", "no stream lag metric found", "filename", labels["filename"], "client", c.name, "host", c.cfg.URL.Host)
 		return false
 	}
+	// If we found an entry in the map it's always safe to delete it, this is the
+	// only place we try to look for that entry. This guarantees we don't endlessly
+	// grow the map entries + if the entire labelset is seen again we would just
+	// write this entry back into the map the next time we send a batch for the stream.
+	delete(c.streamLagMetricsMap, key)
+
 	if !c.metrics.streamLag.Delete(lset) {
 		level.Debug(c.logger).Log("msg", "no stream lag metric deleted, mismatch between stored metrics map and client library registered metrics", "filename", labels["filename"], "client", c.name, "host", c.cfg.URL.Host)
 		return false
--- a/clients/pkg/promtail/client/client.go
+++ b/clients/pkg/promtail/client/client.go
@@ -139,13 +139,16 @@ type Client interface {
 
 // Client for pushing logs in snappy-compressed protos over HTTP.
 type client struct {
-	name            string
-	metrics         *Metrics
-	streamLagLabels []string
-	logger          log.Logger
-	cfg             Config
-	client          *http.Client
-	entries         chan api.Entry
+	name    string
+	metrics *Metrics
+	// TODO: callum, replace with new client_golang functionality `DeletePartialMatch(labels Labels)`
+	// once https://github.com/prometheus/client_golang/pull/1013 is merged.
+	streamLagMetricsMap map[string]prometheus.Labels
+	streamLagLabels     []string
+	logger              log.Logger
+	cfg                 Config
+	client              *http.Client
+	entries             chan api.Entry
 
 	once sync.Once
 	wg   sync.WaitGroup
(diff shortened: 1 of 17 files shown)


clients/pkg/promtail/client/client.go
```diff
@@ -451,11 +458,24 @@ func (c *client) processEntry(e api.Entry) (api.Entry, string) {
...
+	lset, ok := c.streamLagMetricsMap[key]
+	if !ok {
+		level.Debug(c.logger).Log("msg", "no stream lag metric found", "filename", labels["filename"], "client", c.name, "host", c.cfg.URL.Host)
+		return false
+	}
+	if !c.metrics.streamLag.Delete(lset) {
+		level.Debug(c.logger).Log("msg", "no stream lag metric deleted, mismatch between stored metrics map and client library registered metrics", "filename", labels["filename"], "client", c.name, "host", c.cfg.URL.Host)
+		return false
+	}
+	return true
```
cstyan, line 472: todo: also delete the entry from the map when we delete the metric itself

clients/pkg/promtail/client/client.go
```diff
@@ -451,11 +458,30 @@ func (c *client) processEntry(e api.Entry) (api.Entry, string) {
...
 }
 
 func (c *client) Name() string {
 	return c.name
 }
+
+func (c *client) streamLagMetricsMapKey(labels prometheus.Labels) string {
```
owen-d, line 485: Does this happen before relabeling/etc? If a config drops the `filename` label, is this ok?

clients/pkg/promtail/client/client.go
```diff
@@ -352,6 +357,8 @@ func (c *client) sendBatch(tenantID string, batch *batch) {
 					// also set client name since if we have multiple promtail clients configured we will run into a
 					// duplicate metric collected with same labels error when trying to hit the /metrics endpoint
 					lblSet[ClientLabel] = c.name
+					key := c.streamLagMetricsMapKey(lblSet)
```
owen-d, line 360: nit: This can probably check `if key, ok := map[key]; !ok`

clients/pkg/promtail/client/client.go
```diff
@@ -174,12 +178,13 @@ func newClient(metrics *Metrics, cfg Config, streamLagLabels []string, logger lo
...
-		cfg:             cfg,
-		entries:         make(chan api.Entry),
-		metrics:         metrics,
-		streamLagLabels: streamLagLabels,
-		name:            asSha256(cfg),
+		logger:              log.With(logger, "component", "client", "host", cfg.URL.Host),
+		cfg:                 cfg,
+		entries:             make(chan api.Entry),
+		metrics:             metrics,
+		streamLagMetricsMap: make(map[string]prometheus.Labels),
```
owen-d, line 185: I think we need this new map to be behind a mutex to avoid panics with getting/setting a hashmap concurrently.

clients/pkg/promtail/client/client.go
cstyan: Good catch. This happens after relabelling.

By the time we receive an entry on the clients channel the streams labelset has already had relabelling rules applied. So then we can't guarantee that we have the filename (or other unique target identifying label).

Without my recent PRs we expose the same metric twice, which results in Prometheus being unable to scrape `/metrics`. With my recent PRs and someone using relabelling or the `labeldrop` pipeline stage they would likely end up with one stream lag metric per scrape job. That doesn't change with this PR. We would simply track all stream lag for files with `filename` dropped as if they were just one metric for the client.

This PR would allow us to at least not continue reporting the stream lag metric for a file that goes away in **most** cases. We still need additional changes/refactoring to handle all cases for this metric properly. Ed and I are going to discuss further on Monday.

PR #5943, opened 2022-04-18
promtail: Add support for exclusion patterns in `static_config`

### LINKED ISSUE #5744: Allow exclusion patterns on Promtail's `static_config`
There's currently no easy way to _exclude_ files from Promtail's `static_config`, (or as part of a `file_sd_config` definition). 

This functionality might be useful in a number of scenarios:
- The target directory contains files with different formats, and we want to set up separate `pipeline_stages` for a subset of them
- The target directory might contain files which Promtail doesn't have access to. This would avoid polluting logs with `msg="failed to start tailer" error="open /tmp/log/root.log: permission denied" filename=/tmp/log/root.log`
- For security reasons, one might want to avoid tailing specific files or subdirectories
- We might want to exclude heavy/spammy log files

There was a prior discussion over at #3804 which didn't have any negative feedback, but was shot down by stalebot. I think that the suggestion there was pretty reasonable.

I propose that we add a new `__path_exclude__` field under static_configs labels, which would use the same package for glob patterns. The config could look like
```
      scrape_configs:
        - job_name: system
          static_configs:
          - targets:
              - localhost
            labels:
              job: tmplogs
              __path__: /tmp/log/*.log
              __path_exclude__: /tmp/log/root.log
```

This new field would be used to perform a second doublestar.Match call and remove any found files from the `matches` slice before t.startTailing.

https://github.com/grafana/loki/blob/62876c57f76797df401190b5df6c0c9940b006b1/clients/pkg/promtail/targets/file/filetarget.go#L192-L197

An alternative would be to use `relabel_configs` to drop files with specific `__path__` labels.   
One drawback here is that people are hesitant to use relabelling rules, plus I think that doing it that way is more confusing (having to use both regexes for exclusion, and glob patterns for inclusion).

Another alternative would be working with the upstream bmatcuk/doublestar package to add a negation pattern there. If we don't want to support either option, we should mention at least in the docs that a user should consider stricter pattern matching to exclude files.

**Additional context**
I wouldn't think that glob patterns are _that_ performance-intensive, but still this change would only affect users who choose to use them.

If this seems reasonable, I'd be more than happy to start the implementation. Let me know what you think! ^^

tpaschalis:
Signed-off-by: Paschalis Tsilias <paschalis.tsilias@grafana.com>

As stated in the linked Issue, there's currently no easy way to _exclude_ files from a Promtail config, except for overly restrictive and/or complex glob patterns.

This addition may be helpful in a number of scenarios, from avoiding usage spikes in Grafana cloud, to having separate pipeline_stages for specific file formats, or filtering directories which Promtail couldn't tail due to permissions.

Since the upstream bmatcuk/doublestar package doesn't support negation patterns, this is another way of achieving the same result.

Fixes #5744

From my local tests, this seems to work as intended both in scrape_config and file_sd_config blocks; let me know if I've missed anything ^^

**Checklist**
- [x] Documentation added
- [X] Tests updated


clients/pkg/promtail/targets/file/filetarget.go
```diff
@@ -196,8 +199,26 @@ func (t *FileTarget) sync() error {
 		}
 	}
 
+	if fi, err := os.Stat(t.pathExclude); err == nil && !fi.IsDir() {
+		matchesExcluded = []string{t.pathExclude}
+	} else {
+		matchesExcluded, err = doublestar.Glob(t.pathExclude)
```
tpaschalis, line 205: If path.Exclude is empty, then doublestar.Glob returns `nil, nil`, just skipping the loop below.

cyriltovena (approved):
LGTM

adam0s9:
is it please possible to make it exclude two files resp. two different logfiles with no common root (how about 3)?

tpaschalis:
@adam0s9 hey there 👋 

Right now this is not possible with the `static_configs` targets. The workaround is to define more multiple, more fine-grained targets in your scrape configuration to handle nested directories.

If you'd like to propose an enhancement to the way `__path_exclude__` works, I'd recommend opening a new issue to start the conversation from scratch and get more eyes on it.

PR #5945, opened 2022-04-18
store: fix bug in getting chunk ids from ingesters when using boltdb-shipper

sandeepsukhani:
We have a [ChunkStore](https://github.com/grafana/loki/blob/8f024954539b5eb706ab3ef967c4b1696ad04fd6/pkg/storage/stores/composite_store.go#L16) type which has a `GetChunkRefs` method and [LokiStore](https://github.com/grafana/loki/blob/5ea1951c6e3e2690bdfb993e6a71dc4e8d8efe50/pkg/storage/store.go#L41) type which is actually a wrapper on top of `ChunkStore` type, providing methods like `SelectLogs`, `SelectSamples` etc.
We also have a [AsyncStore](https://github.com/grafana/loki/blob/5ea1951c6e3e2690bdfb993e6a71dc4e8d8efe50/pkg/storage/async_store.go#L34) type which is used when `boltdb-shipper` is used as an index store so that we can get chunk ids of recently flushed chunks from ingesters. However, it only implements the `GetChunkRefs` method, which gets chunk ids from both `ChunkStore` and Ingesters.

Previously all 3 had separate factory functions, but in PR #5833, we merged factory functions for `ChunkStore` and `LokiStore`. The problem now is `AsyncStore` now embeds `LokiStore` instead of `ChunkStore`. So when any of the `LokiStore` methods are called, control shifts to `LokiStore`, which calls `ChunkStore` methods since it has no idea about the presence of `AsyncStore`. Unfortunately, this causes us not to query ingesters for recently flushed chunk ids, so we miss some of the data from queries until the index gets synced.

This PR fixes the issue by making the factory function for `LokiStore` initialise `AsyncStore` as well so that it can query chunk ids from ingesters.

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -458,15 +458,17 @@ func (t *Loki) initStore() (_ services.Service, err error) {
 		}
 	}
 
+	if asyncStore {
+		t.Cfg.StorageConfig.EnableAsyncStore = true
+		t.Cfg.StorageConfig.AsyncStoreConfig = storage.AsyncStoreCfg{
+			IngesterQuerier:      t.ingesterQuerier,
+			QueryIngestersWithin: calculateAsyncStoreQueryIngestersWithin(t.Cfg.Querier.QueryIngestersWithin, boltdbShipperMinIngesterQueryStoreDuration(t.Cfg)),
+		}
+	}
+
 	t.Store, err = storage.NewStore(t.Cfg.StorageConfig, t.Cfg.ChunkStoreConfig, t.Cfg.SchemaConfig, t.overrides, t.clientMetrics, prometheus.DefaultRegisterer, util_log.Logger)
 	if err != nil {
 		return
-	}
-
-	if asyncStore {
-		t.Store = storage.NewAsyncStore(t.Store, t.Cfg.SchemaConfig, t.ingesterQuerier,
-			calculateAsyncStoreQueryIngestersWithin(t.Cfg.Querier.QueryIngestersWithin, boltdbShipperMinIngesterQueryStoreDuration(t.Cfg)),
-		)
 	}
 
 	return services.NewIdleService(nil, func(_ error) error {
--- a/pkg/storage/async_store.go
+++ b/pkg/storage/async_store.go
@@ -23,7 +23,8 @@ type IngesterQuerier interface {
 }
 
 type AsyncStoreCfg struct {
-	IngesterQuerier      IngesterQuerier
+	IngesterQuerier IngesterQuerier
+	// QueryIngestersWithin defines maximum lookback beyond which ingesters are not queried for chunk ids.
 	QueryIngestersWithin time.Duration
 }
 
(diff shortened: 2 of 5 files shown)


pkg/storage/async_store.go
```diff
@@ -21,23 +22,28 @@ type IngesterQuerier interface {
 	GetChunkIDs(ctx context.Context, from, through model.Time, matchers ...*labels.Matcher) ([]string, error)
 }
 
+type AsyncStoreCfg struct {
+	IngesterQuerier      IngesterQuerier
+	QueryIngestersWithin time.Duration
```
kavirajk, line 27: nit: I think it would really be helpful to add doc string about what this `QueryIngestersWithin` config does exactly.

kavirajk (commented):
LGTM. Added one minor suggestion.

> Unfortunately, this causes us not to query ingesters for recently flushed chunk ids, so we miss some of the data from queries until the index gets synced.

curious can we add a test for that if possible?

sandeepsukhani:
> LGTM. Added one minor suggestion.
> 
> > Unfortunately, this causes us not to query ingesters for recently flushed chunk ids, so we miss some of the data from queries until the index gets synced.
> 
> curious can we add a test for that if possible?

I think an `e2e` test would make sense here but we don't have a setup for them :shrug:

pkg/storage/async_store.go
sandeepsukhani: Thanks for the review! I have taken care of the suggestion.

pkg/loki/modules.go
```diff
@@ -458,17 +458,19 @@ func (t *Loki) initStore() (_ services.Service, err error) {
...
 	}
 
+	if asyncStore {
+		t.Cfg.StorageConfig.EnableAsyncStore = true
+		t.Cfg.StorageConfig.AsyncStoreConfig = storage.AsyncStoreCfg{
+			IngesterQuerier:      t.ingesterQuerier,
+			QueryIngestersWithin: calculateAsyncStoreQueryIngestersWithin(t.Cfg.Querier.QueryIngestersWithin, boltdbShipperMinIngesterQueryStoreDuration(t.Cfg)),
+		}
+	}
+
```
slim-bean, line 468: It feels a little odd to me having an ingesterquerier inside a config object, would it be possible instead to wrap the storage object here in modules.go rather than inside the init function so we don't have to pass the ingester querier at all?

call the storage.NewStore below this but then kind of like before, if asyncStore then replace the Store inside that object with an async store? WDYT?

owen-d: Agreed, this seems much nicer.

sandeepsukhani: We will have to split the factory function for initialising ChunkStore and LokiStore like how it was before. Is that what you mean?

owen-d: I guess so. I'm also fine with merging this as is to fix the bug even though I don't like the pattern we're using and then refactor afterwards.

slim-bean: after some discussion we have decided to merge this as is and do a follow up to refactor

slim-bean (approved):
LGTM, we can do a follow up to refactor a little per our discussions

PR #5946, opened 2022-04-18
Add better observability to queryReadiness

DylanGuedes:
~~Adds a new `query_readiness_duration_seconds` metric, that reports query readiness duration of a tablemanager/index gateway instance. We should use it later to report performance against the ring mode.~~ 

~~Adds a new `usersToBeQueryReadyForTotal` metric, that reports number of users involved in the query readiness operation. We should use it later to correlate number of users with the query readiness duration.~~

Logs the users involved in the query readiness operation. Will be especially useful for the ring mode so that we can track down the users assigned to each instance.

Logs the duration spent in the query readiness operation.

N/A

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/storage/stores/shipper/downloads/metrics.go
+++ b/pkg/storage/stores/shipper/downloads/metrics.go
@@ -28,7 +28,7 @@ func newMetrics(r prometheus.Registerer) *metrics {
 		ensureQueryReadinessDurationSeconds: promauto.With(r).NewHistogram(prometheus.HistogramOpts{
 			Namespace: "loki_boltdb_shipper",
 			Name:      "query_readiness_duration_seconds",
-			Help:      "Time (in seconds) spent making an index gateway ready to be queried",
+			Help:      "Time (in seconds) spent making this instance ready to be queried",
 			Buckets:   instrument.DefBuckets,
 		}),
 		tablesSyncOperationTotal: promauto.With(r).NewCounterVec(prometheus.CounterOpts{
--- a/pkg/storage/stores/shipper/downloads/metrics.go
+++ b/pkg/storage/stores/shipper/downloads/metrics.go
@@ -16,7 +16,6 @@ type metrics struct {
 	tablesSyncOperationTotal               *prometheus.CounterVec
 	tablesDownloadOperationDurationSeconds prometheus.Gauge
 	ensureQueryReadinessDurationSeconds    prometheus.Histogram
-	usersToBeQueryReadyForTotal            prometheus.Gauge
 }
 
 func newMetrics(r prometheus.Registerer) *metrics {
@@ -31,11 +30,6 @@ func newMetrics(r prometheus.Registerer) *metrics {
 			Name:      "query_readiness_duration_seconds",
 			Help:      "Time (in seconds) spent making an index gateway ready to be queried",
 			Buckets:   instrument.DefBuckets,
-		}),
-		usersToBeQueryReadyForTotal: promauto.With(r).NewGauge(prometheus.GaugeOpts{
-			Namespace: "loki_boltdb_shipper",
-			Name:      "users_to_be_query_ready_for_total",
-			Help:      "Total number of users an index gateway instance has to be query ready for.",
 		}),
 		tablesSyncOperationTotal: promauto.With(r).NewCounterVec(prometheus.CounterOpts{
 			Namespace: "loki_boltdb_shipper",
--- a/pkg/storage/stores/shipper/downloads/metrics.go
+++ b/pkg/storage/stores/shipper/downloads/metrics.go
@@ -3,6 +3,7 @@ package downloads
 import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/weaveworks/common/instrument"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/storage/stores/shipper/downloads/table_manager.go
```diff
@@ -309,6 +317,9 @@ func (tm *TableManager) ensureQueryReadiness(ctx context.Context) error {
 			return err
 		}
 
+		usersToBeQueryReadyLen += len(usersToBeQueryReadyFor)
```
sandeepsukhani, line 320: I am not sure I see much value with this since the final count would be the sum of all the user indexes from all the tables that are within query readiness. What do you think?

DylanGuedes: For now, yes (so you're right!).

But I was wondering, once we adapt the table manager to only download data from tenants assigned to it, the number of reported users is expected to be different. Don't you think this will be useful to check if the different instances are being assigned a balanced number of tenants? Or do you think there's a better way of doing it?

pkg/storage/stores/shipper/downloads/metrics.go
```diff
@@ -23,6 +26,17 @@ func newMetrics(r prometheus.Registerer) *metrics {
...
+			Namespace: "loki_boltdb_shipper",
+			Name:      "query_readiness_duration_seconds",
+			Help:      "Time (in seconds) spent making an index gateway ready to be queried",
+			Buckets:   instrument.DefBuckets,
+		}),
+		usersToBeQueryReadyForTotal: promauto.With(r).NewGauge(prometheus.GaugeOpts{
+			Namespace: "loki_boltdb_shipper",
+			Name:      "users_to_be_query_ready_for_total",
+			Help:      "Total number of users an index gateway instance has to be query ready for.",
+		}),
```
cstyan, line 39: it could be confusing for some people that the metrics are namespaced with `loki_boltdb_shipper` but the help text mentions `index gateway`

pkg/storage/stores/shipper/downloads/table_manager.go
cstyan: > once we adapt the table manager to only download data from tenants assigned to it, the number of reported users is expected to be different

Is the plan to do this :point_up: soon?

pkg/storage/stores/shipper/downloads/table_manager.go
```diff
@@ -309,6 +317,9 @@ func (tm *TableManager) ensureQueryReadiness(ctx context.Context) error {
...
+
+		level.Debug(util_log.Logger).Log("msg", "instance should be query ready for users", "users", usersToBeQueryReadyFor)
```
cstyan, line 322: we could also log `time.Since(start)` here?

DylanGuedes: good call. I'll add it.

DylanGuedes:
> Is the plan to do this point_up soon?

I'm working on it right now, I'm just not sure how long it will take to have this merged as I'm having some difficulties in the implementation :sweat_smile: do you think it makes sense to remove this metric for now and only add it in a more appropriate time?

cstyan:
Personally I would err on the side of adding it earlier rather than later. It's not that big of a deal to remove a metric later on (especially given that the metric wouldn't make it into a big release for quite a while). OTOH, it can be a pain in the ass to build and deploy a new image in a live environment just because you don't have a metric you need to debug something during an incident.

pkg/storage/stores/shipper/downloads/metrics.go
DylanGuedes: removed this one and adjusted the other helper text.

DylanGuedes:
> Personally I would err on the side of adding it earlier rather than later. It's not that big of a deal to remove a metric later on (especially given that the metric wouldn't make it into a big release for quite a while). OTOH, it can be a pain in the ass to build and deploy a new image in a live environment just because you don't have a metric you need to debug something during an incident.

I agree! But at the end I decided to remove it just because I already added a log message that will serve a similar purpose (i.e: we can derive/debug how is the balancing going with Loki instead of Prometheus).

JordanRushing (approved):
LGTM

pkg/storage/stores/shipper/downloads/table_manager.go
```diff
@@ -309,6 +315,7 @@ func (tm *TableManager) ensureQueryReadiness(ctx context.Context) error {
 			return err
 		}
 
+		level.Debug(util_log.Logger).Log("msg", "instance should be query ready for users", "users", usersToBeQueryReadyFor, "query_readiness_duration", time.Since(start))
```
sandeepsukhani, line 318: I see 2 problems here:
1. We are logging the duration before doing query readiness.
2. We will be logging cumulative query readiness duration for tables since the `start` is never reset.

pkg/storage/stores/shipper/downloads/table_manager.go
```diff
@@ -241,6 +241,12 @@ func (tm *TableManager) cleanupCache() error {
 
 // ensureQueryReadiness compares tables required for being query ready with the tables we already have and downloads the missing ones.
 func (tm *TableManager) ensureQueryReadiness(ctx context.Context) error {
+	start := time.Now()
+	defer func() {
+		duration := time.Since(start)
+		tm.metrics.ensureQueryReadinessDurationSeconds.Observe(duration.Seconds())
```
sandeepsukhani, line 247: Since duration is not reused, you can make it concise.
```suggestion
		tm.metrics.ensureQueryReadinessDurationSeconds.Observe(time.Since(start).Seconds())
```

pkg/storage/stores/shipper/downloads/table_manager.go
DylanGuedes: ops, good catch. I've split it into two durations now: one per table, that I reset and only counts for `EnsureQueryReadiness`, and `start` is now only used to report the metric that reports the total time we spent here.

DylanGuedes:
fyi: after some suggestion I decided to remove the queryReadinessDuration metric as it was redundant with one of the new log lines.

slim-bean (approved):
LGTM

PR #5952, opened 2022-04-18
operator: Storage schema update

Red-GV:
This PR allows the operator to update the `schema_config` section of the configuration. It adds `AutoUpdateStorageSchema` to the API to allow a user to define whether they want the operator to automatically update their schemas or not between versions. This PR does not have the ability to apply a previous version after an upgrade.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated


operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	last := existing[count-1]
+
+	if last.Version != version {
+		// Apply the changes to the index for logs starting the next day.
+		// Since the timestamps work by UTC, the goal is to avoid the scenario
+		// in which logs might be inaccessible because they were created with the
+		// wrong schema.
+		from := date.Add(time.Hour * 24).Format("2006-01-02")
```
simonswine, line 32: I am wondering if we should have a check for a grace period of a couple of hours (or however long we think it is gonna take to roll-out the changes in the worst case to all components). 

With this code let's say you would update the SchemaVersion at 1 second to UTC midnight, you would get the new schema 1 second later, but the change hasn't been rolled out to all components.

operator/api/v1beta1/lokistack_types.go
```diff
@@ -343,9 +343,30 @@ type ObjectStorageSecretSpec struct {
...
+)
+
 // ObjectStorageSpec defines the requirements to access the object
 // storage bucket to persist logs by the ingester component.
 type ObjectStorageSpec struct {
+	// Type of object storage that should be used
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=v12
```
simonswine, line 366: I wonder why the default is `v12`, while the example stacks in hack have still `schemaVersion: v11`.

simonswine (commented):
Thanks for your work on this. I think this looks quite ready. I would like to actually spin it up myself, but till then some feedback.

operator/api/v1beta1/lokistack_types.go
Red-GV: Good catch. I was testing the appending schema functionality and forgot to change it back.

operator/internal/manifests/storage/schema.go
Red-GV: Maybe it might be best to do a two day gap instead of a one day one.

Red-GV: I do have a question about multiple schemas being applied for the same day. For instance, if I make a schema change to be applied on a day in the future and then make another to apply on the same day, is the most recent change the only one that matters?

simonswine: Yes there can only be a single schema that is valid for the full UTC day. 

I think we should make sure we do never change a schema that is active or will become active in the next n hours (e.g. 24 hours).

simonswine (approved):
I have given the latest changes a try in the real world and it seemed to behave all as expected. 

I did notice this edge case though:

https://github.com/Red-GV/loki/compare/operator-storage-schema-update...simonswine:20220425_operator-storage-schema-update-add-test?expand=1

I think it is more on the nice to have end, as I don't think it will impact the loki operation at all.

simonswine:
> Thanks for the review, @simonswine! Though I do have a question. Maybe I'm missing something, but I think we're testing the same edge case already?
> 
> https://github.com/grafana/loki/blob/cabf0d222f161ceb154197095aa9279b3cb71ab0/operator/internal/manifests/storage/schema_test.go#L80

Maybe have a try and run it. This one should actually fail. (If I didn't make a mistake).

Red-GV:
Ah, now I see it. Thanks, @simonswine! The latest commit should fix that case now.

operator/api/v1beta1/lokistack_types.go
```diff
@@ -343,9 +343,30 @@ type ObjectStorageSecretSpec struct {
...
+	ObjectStorageSchemaV11 ObjectStorageSchemaVersion = "v11"
+
+	// ObjectStorageSchemaV12 when using v12 for the storage schema
+	ObjectStorageSchemaV12 ObjectStorageSchemaVersion = "v12"
+)
+
 // ObjectStorageSpec defines the requirements to access the object
 // storage bucket to persist logs by the ingester component.
 type ObjectStorageSpec struct {
+	// Type of object storage that should be used
```
periklis, line 362: Doc requires update explaining schema version. In addition we should highlight here that using/updating this field is an append only operation, means we support only future log chunks being stored in the new schema version. Further on we need to give here a thorough explanation of what `future log chunks` means, e.g. 24h after successful reconciliation of the schema version on all components.

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,71 @@
...
+		// case is not in consideration:
+		//
+		// Day 0: Accept and set update to v12 for two days in the future
+		// Day 1: Accept and set update to v11 (v12 may not have been applied yet)
+		// Day 2: v11 Update will be applied
+		//
+		// Since the UTC timing is important and to avoid risk of multiple
+		// updates at wrong intervals, the v12 update would be applied on Day 2
+		// and the v11 update would be applied on Day 3. However, changes on the same
+		// day (as implied by the date being the same) should be fine.
```
periklis, line 51: Is this doc comment valid? I don't quite get the "case is not in consideration" vs the code below. In addition a `Day 3` is mentioned. Maybe we use case use example dates to enhance comprehension.

operator/internal/manifests/storage/schema_test.go
```diff
@@ -0,0 +1,145 @@
...
+	actual := UpdateSchemas(lokiv1beta1.ObjectStorageSchemaV11, test)
+
+	require.Len(t, actual, len(expected))
+	for i := range expected {
+		require.Equal(t, expected[i].From, actual[i].From)
+		require.Equal(t, expected[i].Version, actual[i].Version)
+	}
+}
+
+func TestUpdateSchema_ModifyUnappliedSchema(t *testing.T) {
```
periklis, line 81: This test case assumes in the `test` schema that we have a partially applies schema for `v10`. This case should be by all means not possible. The reconciliation loop should either fail or succeed on each individual create/update op to the apiserver but never store partial results. Thus what is the purpose of this test?

operator/internal/manifests/storage/schema_test.go
```diff
@@ -0,0 +1,145 @@
...
+	actual := UpdateSchemas(lokiv1beta1.ObjectStorageSchemaV12, test)
+
+	require.Len(t, actual, len(expected))
+	for i := range expected {
+		require.Equal(t, expected[i].From, actual[i].From)
+		require.Equal(t, expected[i].Version, actual[i].Version)
+	}
+}
+
+func TestUpdateSchema_ModifyUnappliedSchema_ReduceSameVersion(t *testing.T) {
```
periklis, line 117: This test assumes that in the `test` schema slice we have two entries where the `v11` has a later `From` date than `v12`. This case should not be possible as per meaning somehow the user tricked the system to allow downgrades. Can you elaborate on the purpose?

operator/hack/lokistack_dev.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
 spec:
   size: 1x.extra-small
   storage:
+    schemaVersion: v12
```
periklis, line 8: Since we use the `+kubebuilder:default` annotation we should not need this here right?

operator/hack/lokistack_gateway_dev.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
 spec:
   size: 1x.extra-small
   storage:
+    schemaVersion: v12
```
periklis, line 8: Same here as in `lokistack_dev.yaml`

operator/hack/lokistack_gateway_ocp.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
 spec:
   size: 1x.extra-small
   storage:
+    schemaVersion: v12
```
periklis, line 8: Same here as in `lokistack_dev.yaml`

operator/internal/migrations/storageschema/storageschema_configmap.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+type lokiSchemaConfigJSON struct {
+	Configs []lokiStorageSchemaJSON `json:"configs"`
+}
+
+type lokiStorageSchemaJSON struct {
+	From    string `json:"from"`
+	Version string `json:"schema"`
+}
```
periklis, line 32: nit. What is the purpose of the `JSON` suffix here? These types are neither exported nor used anywhere else. The should be named after the config sections that pass through unmarshaling.

periklis (changes requested):
This is really great work @Red-GV ! :rocket: I really enjoyed reviewing this PR a lot. Solid separation of concerns and usage of modules that enhance readability as well as serving the original intention.

Regardless I would like to open here the conversion on user experience of the proposed CRD changes (cc @simonswine, @xperimental). First of all let's summarize the intention here as the CRD should enable users to use the newer schema version for future log chunks. We certainly know better here that users need to keep in mind (which is not always the case) the following trade-offs when switching to a newer schema version:
1. This operation affects only **future** log chunks (currently proposed 2 days in future)
2. This operation can **not** be reverted.

Looking on the proposed enumeration for `SchemaVersion` and considering the above I have the following reservations:
1. The enumeration field can be flipped back to `v11` as of today which clearly breaks with the second trade-off above. We know this should not be possible!
2. The enumeration field itself does not serve the append behavior we are implementing. I.e. users flip from `v11` to `v12` and if they don't read docs or any other auxilliary points they might expect that `v12` is applied to their data entirely. Cautious users might be concerned of live migrations. Uncautious users might report a bug for missing live migration.

FWIW and as in previous offline discussion about this PR, we may solve the first issue with a validating webhook (adds some extra complexity and bumps requirements on the running clusters, e.g. PKI for TLS certs available, e.g. cert-manager or OpenShift Cert-Signing-Service). But I still want to pock here if there is no better type instead of the enumeration to solve the second issue.

WDYT?

operator/internal/manifests/storage/schema_test.go
Red-GV: This test was to make sure that the operator changes the last element if for some reason a user flips between v11 and v12 on the same day. However when doing examples by hand, I found it difficult to write this case without a third version. I agree this is not a realistic test; it's purely to test the functional case.

operator/internal/manifests/storage/schema_test.go
Red-GV: Actually, this can happen. Before this PR, we have the `v11` schema that takes effect from "2020-10-01". In the same way as above, this is more functionality driven than realistic. This was testing if a user could flip back between the two versions on the same day and there was already a defined version, the earliest definition should be preferred.

However, as you mentioned, this is might be something to nit upon the API discussion. The test case above too might also be nit as well.

operator/internal/manifests/storage/schema.go
Red-GV: Yes, adding dates for clarity make sense.

With the idea of "downgrading" the schema version, there is a case which could happen:

2022-4-27: Apply change to v12, this will be landed on 2022-4-29
2022-4-28: Apply change to v11, this will be landed on 2022-4-30

If this change happened early enough, the operator could have removed that v12 change in place of the v11 one. To keep things simple, I opted to just apply both changes on their given days rather than attempt to modify the v12 change to v11. Since there is already a two day gap to try and avoid timing troubles, it didn't make sense.

operator/config/crd/bases/loki.grafana.com_lokistacks.yaml
```diff
@@ -207,6 +207,13 @@ spec:
                 description: Storage defines the spec for the object storage endpoint
                   to store logs.
                 properties:
+                  schemaVersion:
+                    default: v12
```
xperimental, line 211: Is the default set correctly?

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -77,6 +79,13 @@ func CreateOrUpdateLokiStack(
 		}
 	}
 
+	schemas, err := storageschema.GetLokiStorageSchemaData(ctx, k, req)
+	if err != nil {
+		log.Error(err, "Failed to retrieve config map for storage schemas.")
+		return err
+	}
+	objStore.Schemas = storage.UpdateSchemas(stack.Spec.Storage.SchemaVersion, schemas)
```
xperimental, line 87: Does the current code handle the situation when the `SchemaVersion` is not set in the `LokiStack`? I think, if we want to keep existing stacks then this would need to be handled like it was set to `v11`.

xperimental (commented):
This looks quite good. My only real concern is treating the generated ConfigMap as source of truth for the migration dates, I'd rather keep this separate. For existing deployments this information could be synthesized from the creation timestamp as the desired "SchemaVersion" will be empty for those, or we just keep the static timestamp we have used in the configuration previously.

In general, I also do not see an issue with "rolling back" to a previous schema. This might be needed in case the change to the new schema has unintended consequences (for example due to cost reasons or a bug).

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,71 @@
...
+// schema if there are no existing, the last schema is not the latest, or the last schema
+// is not using the right object storage.
+func UpdateSchemas(version lokiv1beta1.ObjectStorageSchemaVersion, existing []Schema) []Schema {
+	count := len(existing)
+	date := time.Now().UTC()
+
+	if count == 0 {
+		return []Schema{
+			{
+				From:    date.Format("2006-01-02"),
```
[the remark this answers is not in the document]
Red-GV: Probably. I have not really looked into this, but I've seen that the date itself is important for formatting.

operator/internal/migrations/storageschema/storageschema_configmap.go
```diff
@@ -0,0 +1,88 @@
...
+
+const (
+	// LokiConfigFileName is the name of the config file in the configmap
+	LokiConfigFileName = "config.yaml"
+)
+
+// GetLokiStorageSchemaData returns the previous and current storage schemas that
+// need to be applied to the lokistack.
+func GetLokiStorageSchemaData(ctx context.Context, k k8s.Client, req ctrl.Request) ([]storage.Schema, error) {
```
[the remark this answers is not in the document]
Red-GV: I think using the status will be better than the `configmap`.

operator/internal/handlers/lokistack_create_or_update.go
Red-GV: The default covers the case where the user does not specifically set the version.

xperimental:
@Red-GV and I had a conversation about this pull-request and came across an interesting use-case: Let's say a cluster has been through a few schema changes along its lifetime but now the LokiStack resource is deleted (user error, or maybe even on purpose) and subsequently recreated. What do we do about the schema versions of existing data?
The user-controlled part of the LokiStack does not track the changes over time, so there is no way to express "this cluster is new, but there is already data in the storage" and which schema versions to expect for which date ranges. For me it looks like if we want to have schema updates controlled by the operator there is no way around having the dates in the LokiStack spec as well.

Side-question: What happens to the old "unreadable" data? Will the cleanup (retention period) still be able to access / remove it or will the data just stay in storage indefinitely?

operator/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - [6195](https://github.com/grafana/loki/pull/6195) **periklis**: Add ruler config support
 - [6198](https://github.com/grafana/loki/pull/6198) **periklis**: Add support for custom S3 CA
 - [6199](https://github.com/grafana/loki/pull/6199) **Red-GV**: Update GCP secret volume path
+- [5952](https://github.com/grafana/loki/pull/5952) **Red-GV**: Add api to change storage schema version
```
periklis, line 6: Reminder before squash-and-merge: This needs to go to the top.

operator/api/v1beta1/lokistack_types.go
```diff
@@ -314,6 +314,17 @@ type LokiTemplateSpec struct {
...
+// ObjectStorageTLSSpec is the TLS configuration for reaching the object storage endpoint.
+type ObjectStorageTLSSpec struct {
+	// CA is the name of a ConfigMap containing a CA certificate.
+	// It needs to be in the same namespace as the LokiStack custom resource.
+	//
+	// +optional
+	// +kubebuilder:validation:optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:io.kubernetes:ConfigMap",displayName="CA ConfigMap Name"
+	CA string `json:"caName,omitempty"`
+}
```
periklis, line 326: Is this a rebase/merge main mistake? IMHO this was introduced by #6198 .

operator/api/v1beta1/lokistack_types.go
```diff
@@ -350,20 +361,50 @@ type ObjectStorageSecretSpec struct {
...
 // ObjectStorageSpec defines the requirements to access the object
 // storage bucket to persist logs by the ingester component.
 type ObjectStorageSpec struct {
+
+	// Schemas that have been applied to the cluster.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems:=1
+	// +kubebuilder:default:={{version:v11,effectiveDate:"2020-10-11"}}
```
periklis, line 405: :rocket:

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// BuildSchemaConfigList creates a list of schemas to be used to configure
+// the storage schemas for the cluster. This method assumes that the following
+// validation has been done to the statuses and specs:
+//
+// 1. All EffectiveDate fields are able to be parsed
+// 2. All EffectiveDate fields are unique in their respective list
+//
+func BuildSchemaConfigList(
```
periklis, line 34: nit. `BuildSchemaConfig` is imho sufficient as we re-use the loki configuration terminology. `List` as a suffix is not adding more information to the calling side to execute as per return value already a slice. Suffixes like `List` for functions/methods are usually used in dynamic typed programming languages. Here is Go-Land and we say what we do with static types :)

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,200 @@
...
+		// Schemas applied in the future can be ignored cause they have not
+		// taken effect yet. The delay is added to avoid edge cases where UTC
+		// and local time zones may cause interference.
+		//
+		// For example, say there is a future change which will occur on 6/02/22.
+		// If change occurs close to midnight on 6/01/22, there might be timing
+		// issues between the local and UTC. Thus, there is a chance of corruption.
+		// To avoid this, the change on 6/02/22 should be considered as already
+		// applied and an error should be sent back to change the date of the
+		// in-flight change.
```
periklis, line 57: nit. to improve reability of comments I suggest to keep to the date format we parse, i.e. `2022-06-02`.

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,200 @@
...
+	sort.SliceStable(specSchemas, func(i, j int) bool {
+		return specSchemas[i].effectiveDate.Before(specSchemas[j].effectiveDate)
+	})
+
+	reducedSpecSchemas := reduceSortedSchemas(specSchemas)
+	return specList(reducedSpecSchemas), nil
+}
+
+// statusSchemaConfigList creates a list of schemaConfigs based on the given statuses
+func statusSchemaConfigList(statuses []lokiv1beta1.StorageSchemaStatus) []schemaConfig {
```
periklis, line 79: Same here `List` suffix is not adding value. Also I suggest to rename this to `schemaConfigFromStatus` as it says what this function is doing `I return a slice of schemaConfig objects reading from LokiStack Status`.

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,200 @@
...
+			version:       status.Version,
+			effectiveDate: date,
+		}
+	}
+
+	return configs
+}
+
+// specSchemaConfigList creates a list of schemaConfigs based on the given specs
+func specSchemaConfigList(specs []lokiv1beta1.ObjectStorageSchemaSpec) []schemaConfig {
```
periklis, line 95: Same here `List` suffix is not adding value. Also I suggest to rename this to `schemaConfigFromSpec` as it says what this function is doing `I return a slice of schemaConfig objects reading from LokiStack Spec`.

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,200 @@
...
+			version:       spec.Version,
+			effectiveDate: date,
+		}
+	}
+
+	return configs
+}
+
+// specList creates a list of ObjectStorageSchemaSpecs based on the given configs
+func specList(configs []schemaConfig) []lokiv1beta1.ObjectStorageSchemaSpec {
```
periklis, line 111: This is effectively a builder function, let's stick with using the `build` prefix as `buildSpecs`

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+	}
+
+	return true
+}
+
+// reduceSortedSchemas returns a list of configs that have removed redundant entries.
+func reduceSortedSchemas(configs []schemaConfig) []schemaConfig {
+	if len(configs) == 0 {
+		return []schemaConfig{}
```
periklis, line 186: Any specific reason on why we need to return an empty slice instead of nil?

operator/internal/manifests/storage/schema_test.go
```diff
@@ -0,0 +1,613 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func BuildSchemaConfigList_NoSchemas(t *testing.T) {
+	specs := []lokiv1beta1.ObjectStorageSchemaSpec{}
+	statuses := []lokiv1beta1.StorageSchemaStatus{}
+
+	_, err := BuildSchemaConfigList(time.Now().UTC(), specs, statuses)
+
+	require.Error(t, err)
```
periklis, line 18: Shouldn't we check for nil-ness or emptyness here?

operator/internal/manifests/storage/schema_test.go
```diff
@@ -0,0 +1,613 @@
...
+			Version:       lokiv1beta1.ObjectStorageSchemaV11,
+			EffectiveDate: "2020-10-01",
+		},
+	}
+
+	require.NoError(t, err)
+	require.Equal(t, expected, actual)
+}
+
+func BuildSchemaConfigList_AddSchema_WithPriorSchemas_OutsideUpdateWindow(t *testing.T) {
```
periklis, line 42: nit. `Outside` can have ambiguous meaning, i.e.:
- It can mean outside before time window
- It can mean outside after time window

I suggest to rename this to switch to something `Valid/Invalid` or `Matching/Conflicting`.

operator/internal/manifests/storage/schema_test.go
```diff
@@ -0,0 +1,613 @@
...
+			Version:       lokiv1beta1.ObjectStorageSchemaV12,
+			EffectiveDate: "2021-10-01",
+		},
+	}
+
+	require.NoError(t, err)
+	require.Equal(t, expected, actual)
+}
+
+func BuildSchemaConfigList_AddSchema_WithPriorSchemas_InsideUpdateWindow(t *testing.T) {
```
periklis, line 77: As said above, I suggest to rename this to switch to something `Valid/Invalid` or `Matching/Conflicting`.

periklis (changes requested):

operator/api/v1beta1/lokistack_types.go
Red-GV: It's not reintroducing it. I had moved the block above all of the new things and I think that is confusing `git` for producing the diff.

operator/api/v1beta1/lokistack_types.go
```diff
@@ -350,9 +350,30 @@ type ObjectStorageSecretSpec struct {
...
+	ObjectStorageSchemaV11 ObjectStorageSchemaVersion = "v11"
+
+	// ObjectStorageSchemaV12 when using v12 for the storage schema
+	ObjectStorageSchemaV12 ObjectStorageSchemaVersion = "v12"
+)
+
 // ObjectStorageSpec defines the requirements to access the object
 // storage bucket to persist logs by the ingester component.
 type ObjectStorageSpec struct {
+	// Type of schema which should be used for
```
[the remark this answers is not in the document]
Red-GV: Is this an older comment?

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -77,6 +79,20 @@ func CreateOrUpdateLokiStack(
...
+		time.Now().UTC(),
+		stack.Spec.Storage.SchemaVersion,
+		stack.Status.Storage.Schemas,
+	)
+
+	err = status.SetStorageSchemaStatus(ctx, k, req, storageSchemas)
+	if err != nil {
+		ll.Error(err, "failed to set storage schema status")
+		return err
+	}
```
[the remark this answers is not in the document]
Red-GV: The reason I put the status update before the `createOrUpdate` method is because I wanted to avoid a scenario where, for some reason, one resource fails the push and thus the entire function exits. If the `configmap` got updated, then we might have some sync issues between the `configmap` and the `status`.

periklis (commented):
Looks good to me. Let's wait for @xperimental to take a look too.

xperimental (commented):
Have not gotten very far with today's review yet, but here are already a few comments.

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,200 @@
...
+
+const (
+	// DateTimeFormat is the datetime string need to format the time.
+	DateTimeFormat = "2006-01-02"
+
+	// UpdateDelay is the time before a change to the storage schema
+	// is applied to a cluster. This will avoid scenarios in which
+	// the difference between UTC and local time zones will cause
+	// logs to be unavailable.
+	UpdateDelay = time.Hour * 24
```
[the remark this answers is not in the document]
Red-GV: The reason for the two day delay in the hit is to make sure we clear past the UTC/Local midnight scenario. I'm not sure if just a few hours of time will clear the time in some region. However, I don't know the max value of difference there can be between the two. However, the two day window this provides should clear the differences.

(It's two days because the anything within a day of the current time is counted as an applied change. Since the schema only works in year-month-day formats, the delay is actually two days.)

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,200 @@
...
+	"sort"
+	"time"
+
+	"github.com/ViaQ/logerr/v2/kverrors"
+	lokiv1beta1 "github.com/grafana/loki/operator/api/v1beta1"
+)
+
+const (
+	// DateTimeFormat is the datetime string need to format the time.
+	DateTimeFormat = "2006-01-02"
```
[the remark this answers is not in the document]
Red-GV: I'll try to move this somewhere were the project can benefit rather than make a new const every time it needs to be referred to.

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,200 @@
...
+		// taken effect yet. The delay is added to avoid edge cases where UTC
+		// and local time zones may cause interference.
+		//
+		// For example, say there is a future change which will occur on 2022-06-02.
+		// If change occurs close to midnight on 2022-06-01, there might be timing
+		// issues between the local and UTC. Thus, there is a chance of corruption.
+		// To avoid this, the change on 2022-06-02 should be considered as already
+		// applied and an error should be sent back to change the date of the
+		// in-flight change.
+		cutoff := utcTime.Add(UpdateDelay)
```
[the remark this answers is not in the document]
Red-GV: As far as I can tell, I cannot access the `status` field in the validating webhook. So, I don't think I can reject these any earlier.

xperimental (changes requested):
For me, the main remaining issue is the ability for the user to discover errors with the schema-config directly on the LokiStack resource.

operator/internal/manifests/storage/schema.go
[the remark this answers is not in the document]
Red-GV: It's not that I mean there is a delay in the configuration. What I'm concerned about allowing is the given example: https://grafana.com/docs/loki/latest/operations/storage/schema/#changing-the-schema

Granted, this part of the overhead should be assumed by the person adding the schema. However, the enforcing of this delay should at least make this scenario hard to cause. Though I agree that the timing capture here is not obvious. What I'm trying to avoid is someone telling Loki, make a change tomorrow (local time) and it is already tomorrow for Loki (utc time). Adding the delay should stop anyone from making the `effectiveDate` either the current date.

I guess it should be stated here that all of the dates entered on the spec are assumed to be in UTC. While I could do some assumptions on the location via `time.Now().Location()`, that may have areas where it might not work out. Assuming that the location could be trusted, I suppose doing a validation check against that time converted to UTC would be the most ideal way to stop this from occurring.

Then in that case, we'd probably only need to budget in an hour to avoid any surprise midnight change.

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -78,6 +79,18 @@ func CreateOrUpdateLokiStack(
 		}
 	}
 
+	storageSchemas, err := storageoptions.BuildSchemaConfig(
+		time.Now().UTC(),
+		stack.Spec.Storage.Schemas,
+		stack.Status.Storage.Schemas,
+	)
+	if err != nil {
+		ll.Error(err, "failed to create storage schema configurations")
```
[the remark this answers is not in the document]
Red-GV: Mmm, though on that note, is it even required to add in a boolean. If it configured successfully, one would see all the configurations added to the status. Or are you suggesting possibly adding a DegradedStatus for the schema configuration?

xperimental (commented):
Seems like I overlooked the parsing of the dates the last review...

operator/api/v1beta1/lokistack_types.go
```diff
@@ -725,6 +768,35 @@ type LokiStackComponentStatus struct {
 	Ruler PodStatusMap `json:"ruler,omitempty"`
 }
 
+// StorageSchemaStatus defines the observed state of a single
+// change to the Loki Storage Schema.
+type StorageSchemaStatus struct {
```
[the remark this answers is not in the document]
Red-GV: Although they are the same, it think it's fine to have the two objects. Putting a `Spec` labeled object in `Status` or vice versa seems confusing. It might be odd to have a non-spec labeled object in the `Spec`.

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return configs
+}
+
+// schemaConfigFromSpec creates a list of schemaConfigs based on the given specs
+func schemaConfigFromSpec(specs []lokiv1beta1.ObjectStorageSchemaSpec) []schemaConfig {
+	configs := make([]schemaConfig, len(specs))
+
+	for i, spec := range specs {
+		date, _ := time.Parse(lokiv1beta1.StorageSchemaEffectiveDateFormat, string(spec.EffectiveDate))
```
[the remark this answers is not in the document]
Red-GV: `Parse` already defaults to `time.UTC`.

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,197 @@
...
+	reducedSpecSchemas := reduceSortedSchemas(specSchemas)
+	return buildSpecs(reducedSpecSchemas), nil
+}
+
+// schemaConfigFromStatus creates a list of schemaConfigs based on the given statuses
+func schemaConfigFromStatus(statuses []lokiv1beta1.StorageSchemaStatus) []schemaConfig {
+	configs := make([]schemaConfig, len(statuses))
+
+	for i, status := range statuses {
+		date, _ := time.Parse(lokiv1beta1.StorageSchemaEffectiveDateFormat, status.EffectiveDate)
```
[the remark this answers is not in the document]
Red-GV: `Parse` already defaults to `time.UTC`.
Ref: https://pkg.go.dev/time@go1.17.11#ParseInLocation

operator/internal/manifests/storage/schema.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return specs
+}
+
+// isApplied returns whether a schemaConfig has been applied or not
+func isApplied(config schemaConfig, effectiveDate time.Time) bool {
+	year, month, day := config.effectiveDate.Date()
+	utcTime := time.Date(year, month, day, 0, 0, 0, 0, time.UTC)
+
+	return config.effectiveDate.Before(effectiveDate) || utcTime.Equal(effectiveDate)
```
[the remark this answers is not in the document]
Red-GV: Yes, however we cannot make any modifications to the spec if they occur on the same day. `effectiveDate` should generally be before any configuration change, but checking explicitly for the same day ensures that it is rejected.

operator/api/v1beta1/lokistack_webhook_test.go
```diff
@@ -0,0 +1,139 @@
...
+							Version:       v1beta1.ObjectStorageSchemaV12,
+							EffectiveDate: "2020-10-13",
+						},
+					},
+				},
+			},
+		},
+	},
+	{
+		desc: "not unique schema effective dates",
```
xperimental, line 40: For some reason this check does not work for me on the test-cluster. I am able to write a spec with two schema-configs with the same effective date. If they are in the right order (v11, v12, v11, for example) this will lead to a configuration that causes Loki to immediately fail to start up.

operator/internal/manifests/var.go
```diff
@@ -108,6 +118,11 @@ func GossipLabels() map[string]string {
 	}
 }
 
+// LokiConfigMapName is the name of the configmap for Loki
+func LokiConfigMapName(stackName string) string {
```
xperimental, line 122: Can be private again.

xperimental (changes requested):
Like the recent changes :+1: 

The issue with duplicated schema-configs is the only remaining blocker, I'd say.

operator/api/v1beta1/lokistack_webhook_test.go
Red-GV: Strange, the last time I tested this, it worked as expected (by that I mean that it blocked the change from occurring).

xperimental (approved):

PR #5954, opened 2022-04-19
boltdb-shipper: common index set should not expire before any of the user index sets

sandeepsukhani:
While cleaning up expired index, we look at the last used time to decide whether an index set has expired or not. The last used time gets set to `now` while initializing or querying the index set. The common index set should always be the last one to expire because it is the parent directory of all the user index sets and while querying we anyways need both user and common index set.

To make sure common index set never expires before user index set, we query user index set first so that its last used time is smaller than common index set. However that would not work always because if user cancels the query between index set querying, the last used time of common index set would not get updated and break our expectation.

This PR fixes the issue by making sure we only remove common index set when all the user index sets have been removed.


kavirajk (approved):
LGTM.

PR #5957, opened 2022-04-19
index-gateway: fix index gateway chunk refs response

sandeepsukhani:
While building `GetChunkRefs` response from index gateways, we use the address of loop variable. This causes us to have the same values for all the chunks. Lint somehow missed catching that mistake. This PR fixes the issue by using index for getting the address of the slice item.

### DIFF
--- a/pkg/storage/stores/shipper/indexgateway/gateway.go
+++ b/pkg/storage/stores/shipper/indexgateway/gateway.go
@@ -307,8 +307,8 @@ func (g *Gateway) GetChunkRef(ctx context.Context, req *indexgatewaypb.GetChunkR
 		Refs: make([]*logproto.ChunkRef, 0, len(chunks)),
 	}
 	for _, cs := range chunks {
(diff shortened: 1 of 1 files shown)


chaudum:
Great find! Do we have a test where we check the correctness of the chunk refs response?

sandeepsukhani:
> Great find! Do we have a test where we check the correctness of the chunk refs response?

No, sorry, we don't have any such tests for index gateway. I just want to unblock our internal release for now. I won't be able to work on tests today. We can open an issue so that we don't forget adding them.

chaudum (approved):

sandeepsukhani:
Here is the issue for tracking https://github.com/grafana/loki/issues/5958

slim-bean (approved):
LGTM!

cyriltovena:
Awesome I think we should have added a test too. If anyone want to take a stab at it.

PR #5961, opened 2022-04-19
redact s3 credential values when printing config to logs

afayngelerindbx:
This PR makes sure that we don't commit S3 credential values to logs when using `-print-config-stderr` and `-log-config-reverse-order` flags. 

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/chunk/client/aws/s3_storage_client.go
+++ b/pkg/storage/chunk/client/aws/s3_storage_client.go
@@ -14,6 +14,8 @@ import (
 	"strings"
 	"time"
 
+	"gopkg.in/yaml.v2"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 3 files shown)


pkg/storage/chunk/client/aws/s3_storage_client.go
```diff
@@ -69,8 +85,8 @@ type S3Config struct {
 	BucketNames      string
 	Endpoint         string              `yaml:"endpoint"`
 	Region           string              `yaml:"region"`
-	AccessKeyID      string              `yaml:"access_key_id"`
-	SecretAccessKey  string              `yaml:"secret_access_key"`
+	AccessKeyID      secret              `yaml:"access_key_id"`
```
kavirajk, line 88: I'm failing to understand how this will redact when printing these configs in logs. Can you clarify?

1. Can you tell me where exactly it's getting logged?
2. Even with this change, if we log the config (either logfmt or json), it's still going to be visible correct?
3. Only advantage of adding this `secret` type I see is when we print the config after marshalling into YAML, it won't be visible. Is that your real use case? can you help me understand how it's useful?

afayngelerindbx: Thanks for taking a look!

When I set the `-print-config-stderr` or `-log-config-reverse-order`, loki marshals the config to YAML and dumps it to stderr. This diff forces S3 credentials to be redacted from this output. I will update the PR description as well.

afayngelerindbx: For reference, loki prints config using [this](https://github.com/grafana/loki/blob/a2dd73158621441fe2d83346e792913c6770c945/pkg/util/config.go#L34) function [here](https://github.com/grafana/loki/blob/a2dd73158621441fe2d83346e792913c6770c945/cmd/loki/main.go#L57)

kavirajk: Thanks @afayngelerindbx for the clarification :)

kavirajk (approved):
LGTM!

Issue #5965, opened 2022-04-20
Index Gateway: Unable to handle GetChunkRefs/LabelValuesForMetricName with empty matchers


simonswine:
#5892 introduced a bug when label values and potentially GetChunkRefs are called with empty matchers.

```
$ curl -H 'localhost:3100/loki/api/v1/label/namespace/values'                                                                                                                                                                                                                                                                                               
rpc error: code = Unknown desc = parse error at line 1, col 2: syntax error: unexpected }, expecting IDENTIFIER
```

Expected behavior is to show all label values in that case.

PR #5968, opened 2022-04-20
Add integration tests which test happy path

simonswine:
This PR adds simple go based integration tests, which can run Loki
in two modes:

- single-binary
- micro-services with index-gateway

After standing up it performs the most basic duties like:
- ingest logs
- run a range query
- query label names and label values

This is meant to catch more significant problem as early as possible in
the PR.

We should aim to keep this as simple as possible.

Note: _`integration/client/client.go` is copied from the closed source Granana Enterprise Logs code based_

### DIFF
--- a/integration/client/client.go
+++ b/integration/client/client.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -150,7 +149,7 @@ func (c *Client) pushLogLine(line string, timestamp time.Time, extraLabelList ..
 		return nil
 	}
 
-	buf, err := ioutil.ReadAll(res.Body)
+	buf, err := io.ReadAll(res.Body)
 	if err != nil {
 		return fmt.Errorf("reading request failed with status code %v: %w", res.StatusCode, err)
 	}
@@ -430,7 +429,7 @@ func (c *Client) run(u string) ([]byte, int, error) {
 	}
 	defer res.Body.Close()
 
-	buf, err := ioutil.ReadAll(res.Body)
+	buf, err := io.ReadAll(res.Body)
 	if err != nil {
 		return nil, 0, fmt.Errorf("request failed with status code %v: %w", res.StatusCode, err)
 	}
--- a/integration/client/client.go
+++ b/integration/client/client.go
@@ -1,0 +1,439 @@
+package client
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type roundTripper struct {
+	instanceID    string
+	token         string
+	injectHeaders map[string][]string
+	next          http.RoundTripper
+}
+
+func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Set("X-Scope-OrgID", r.instanceID)
+	if r.token != "" {
+		req.SetBasicAuth(r.instanceID, r.token)
+	}
+
+	for key, values := range r.injectHeaders {
+		for _, v := range values {
+			req.Header.Add(key, v)
+		}
+
+		fmt.Println(req.Header.Values(key))
+	}
+
+	return r.next.RoundTrip(req)
+}
+
+type CortexClientOption interface {
+	Type() string
+}
+
+type InjectHeadersOption map[string][]string
+
+func (n InjectHeadersOption) Type() string {
+	return "headerinject"
+}
+
+// Client is a HTTP client that adds basic auth and scope
+type Client struct {
+	Now time.Time
+
+	httpClient *http.Client
+	baseURL    string
+	instanceID string
+}
+
+// NewLogsClient creates a new client
+func New(instanceID, token, baseURL string, opts ...CortexClientOption) *Client {
+	rt := &roundTripper{
+		instanceID: instanceID,
+		token:      token,
+		next:       http.DefaultTransport,
+	}
+
+	for _, opt := range opts {
+		switch opt.Type() {
+		case "headerinject":
+			rt.injectHeaders = opt.(InjectHeadersOption)
+		}
+	}
+
+	return &Client{
+		Now: time.Now(),
+		httpClient: &http.Client{
+			Transport: rt,
+		},
+		baseURL:    baseURL,
+		instanceID: instanceID,
+	}
+}
+
+// PushLogLine creates a new logline with the current time as timestamp
+func (c *Client) PushLogLine(line string, extraLabels ...map[string]string) error {
+	return c.pushLogLine(line, c.Now, extraLabels...)
+}
+
+// PushLogLineWithTimestamp creates a new logline at the given timestamp
+// The timestamp has to be a Unix timestamp (epoch seconds)
+func (c *Client) PushLogLineWithTimestamp(line string, timestamp time.Time, extraLabelList ...map[string]string) error {
+	return c.pushLogLine(line, timestamp, extraLabelList...)
+}
+
+func formatTS(ts time.Time) string {
+	return strconv.FormatInt(ts.UnixNano(), 10)
+}
+
+type stream struct {
+	Stream map[string]string `json:"stream"`
+	Values [][]string        `json:"values"`
+}
+
+// pushLogLine creates a new logline
+func (c *Client) pushLogLine(line string, timestamp time.Time, extraLabelList ...map[string]string) error {
+	apiEndpoint := fmt.Sprintf("%s/loki/api/v1/push", c.baseURL)
+
+	s := stream{
+		Stream: map[string]string{
+			"job": "varlog",
+		},
+		Values: [][]string{
+			{
+				formatTS(timestamp),
+				line,
+			},
+		},
+	}
+	// add extra labels
+	for _, labelList := range extraLabelList {
+		for k, v := range labelList {
+			s.Stream[k] = v
+		}
+	}
+
+	data, err := json.Marshal(&struct {
+		Streams []stream `json:"streams"`
+	}{
+		Streams: []stream{s},
+	})
+	if err != nil {
+		return err
+	}
+	req, err := http.NewRequest("POST", apiEndpoint, bytes.NewReader(data))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("X-Scope-OrgID", c.instanceID)
+
+	// Execute HTTP request
+	res, err := c.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	if res.StatusCode/100 == 2 {
+		defer res.Body.Close()
+		return nil
+	}
+
+	buf, err := ioutil.ReadAll(res.Body)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


simonswine:
@chaudum this is also reproducing the bug mentioned in #5965

integration/loki_micro_services_test.go
```diff
@@ -0,0 +1,106 @@
...
+	cliIngester.Now = now
+	cliQueryFrontend := client.New(tenantID, "", tQueryFrontend.HTTPURL().String())
+	cliQueryFrontend.Now = now
+
+	t.Run("ingest-logs-store", func(t *testing.T) {
+		// ingest some log lines
+		require.NoError(t, cliDistributor.PushLogLineWithTimestamp("lineA", now.Add(-45*time.Minute), map[string]string{"job": "fake"}))
+		require.NoError(t, cliDistributor.PushLogLineWithTimestamp("lineB", now.Add(-45*time.Minute), map[string]string{"job": "fake"}))
+
+		// TODO: Flushing is currently causing a panic, as the boltdb shipper is shared using a global variable in:
```
dannykopping, line 70: TODO?

dannykopping (commented):
Thanks for getting us started on this @simonswine!
Added a couple comments

integration/loki_micro_services_test.go
simonswine: Yes removing this is not trivial, I would like to get the tests in the current form and address this later in a targeted PR.

The PR is complex enough as it is.

integration/cluster/cluster.go
```diff
@@ -0,0 +1,348 @@
...
+compactor:
+  working_directory: %s/retention
+  shared_store: filesystem
+  retention_enabled: true
+
+analytics:
+  reporting_enabled: false
+
+ruler:
+  alertmanager_url: http://localhost:9093
```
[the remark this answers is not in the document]
simonswine: I guess I just copied it from `./cmd/loki/loki-local-config.yaml`. I will remove it

integration/cluster/cluster.go
```diff
@@ -0,0 +1,348 @@
...
+
+type Cluster struct {
+	sharedPath string
+	components []*Component
+	waitGroup  sync.WaitGroup
+}
+
+func New() *Cluster {
+	wrapRegistry()
+	sharedPath, err := ioutil.TempDir("", "loki-shared-data")
```
MichelHollands, line 118: The `ioutil` package is somewhat deprecated: https://pkg.go.dev/io/ioutil#pkg-overview.

```suggestion
	sharedPath, err := os.MkdirTemp("", "loki-shared-data")
```

integration/cluster/cluster.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+	if err := errs.Err(); err != nil {
+		return err
+	}
+
+	// wait for all process to close
+	c.waitGroup.Wait()
+
+	// cleanup dirs/files
+
```
MichelHollands, line 160: This empty line can be removed.

integration/cluster/cluster.go
```diff
@@ -0,0 +1,348 @@
...
+	return &url.URL{
+		Host:   fmt.Sprintf("localhost:%d", c.grpcPort),
+		Scheme: "grpc",
+	}
+}
+
+func (c *Component) writeConfig() error {
+	var err error
+
+	configFile, err := ioutil.TempFile("", "loki-config")
```
MichelHollands, line 223:
```suggestion
	configFile, err := os.CreateTemp("", "loki-config")
```

integration/cluster/cluster.go
```diff
@@ -0,0 +1,348 @@
...
+	if err != nil {
+		return fmt.Errorf("error creating config file: %w", err)
+	}
+
+	c.dataPath, err = ioutil.TempDir("", "loki-data")
```
MichelHollands, line 228:
```suggestion
	c.dataPath, err = os.MkdirTemp("", "loki-data")
```

integration/cluster/cluster.go
```diff
@@ -0,0 +1,348 @@
...
+	if err != nil {
+		return fmt.Errorf("error creating config file: %w", err)
```
MichelHollands, line 230:
```suggestion
		return fmt.Errorf("error creating data path: %w", err)
```

integration/cluster/cluster.go
```diff
@@ -0,0 +1,348 @@
...
+		c.dataPath,
+		c.dataPath,
+		c.dataPath,
+		c.grpcPort,
+		c.grpcPort,
+	))); err != nil {
+		return fmt.Errorf("error writing config file: %w", err)
+	}
+	if err := configFile.Close(); err != nil {
+		return fmt.Errorf("error writing config file: %w", err)
```
MichelHollands, line 244:
```suggestion
		return fmt.Errorf("error closing config file: %w", err)
```

integration/loki_micro_services_test.go
```diff
@@ -0,0 +1,106 @@
...
+			"-target=index-gateway",
+		)
+		tDistributor = clu.AddComponent(
+			"distributor",
+			"-target=distributor",
+		)
+		tIngester = clu.AddComponent(
+			"ingester",
+			"-target=ingester",
+			//			"-boltdb.shipper.index-gateway-client.server-address="+tIndexGateway.GRPCURL().Host,
```
MichelHollands, line 32: Is this still needed?

integration/cluster/cluster.go
```diff
@@ -0,0 +1,348 @@
...
+		c.httpPort,
+		c.grpcPort,
+		c.dataPath,
+		c.cluster.sharedPath,
+		c.cluster.sharedPath,
+		c.dataPath,
+		c.dataPath,
+		c.dataPath,
+		c.grpcPort,
+		c.grpcPort,
```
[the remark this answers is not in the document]
simonswine: In retrospect it is surprising that I didn't get more into trouble with this `Sprintf` :joy: 

acb3e90c7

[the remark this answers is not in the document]
dannykopping: Super :+1: thanks

dannykopping (approved):
I'll approve this, but I think we need some consensus before merging.

Integration tests has been a hotly-contested topic, and AFAICS these tests cover the same surface-area as the Loki Canary. The scope here will naturally diverge/expand over time, and this has added benefits over the Canary (can be automated), so I really want this personally.

integration/loki_micro_services_test.go
simonswine: Oh yes in fact it should be uncommented

jeschkies:
Did we agree to add integration tests? I know @slim-bean was skeptical. I'd be afraid that this opens the door for more evolved tests.

integration/loki_micro_services_test.go
```diff
@@ -0,0 +1,106 @@
...
+		)
+		tQueryScheduler = clu.AddComponent(
+			"query-scheduler",
+			"-target=query-scheduler",
+			"-boltdb.shipper.index-gateway-client.server-address="+tIndexGateway.GRPCURL().Host,
+		)
+		tQueryFrontend = clu.AddComponent(
+			"query-frontend",
+			"-target=query-frontend",
+			"-frontend.scheduler-address="+tQueryScheduler.GRPCURL().Host,
```
trevorwhitney, line 42: ahh, so this is how you overcome the different grpc port for the scheduler. since you will only have 1 instance for this test, I feel like we could pass this to the cluster setup so we don't have to remember this here. WDYT?

trevorwhitney (commented):
Thanks for getting this conversation started. I like the idea of adding integration tests, and I think this is the right approach. I really like there is no docker or building of binaries involved here, as this also makes this a really useful debug tool. I have 2 thoughts.

1) How would you feel about a gateway and a single client, limiting us to only being able to test "front-door" behavior in this suite?
2) How would you feel about an SSD test? I think this could either be a 3rd test or it could replace single binary?

cyriltovena:
> I'll approve this, but I think we need some consensus before merging.
> 
> Integration tests has been a hotly-contested topic, and AFAICS these tests cover the same surface-area as the Loki Canary. The scope here will naturally diverge/expand over time, and this has added benefits over the Canary (can be automated), so I really want this personally.

There's no doubt for me with regards to canary this is debuggable and gives faster feedback, I love both and I want this too.

integration/client/client.go
```diff
@@ -0,0 +1,438 @@
...
+}
+
+// Response holds the status and data
+type Response struct {
+	Status string
+	Data   DataType
+}
+
+// RunRangeQuery runs a query and returns an error if anything went wrong
+func (c *Client) RunRangeQuery(query string) (*Response, error) {
```
cyriltovena, line 294: I would love that we do a follow up PR and change this to use either logcli client or https://github.com/grafana/loki-client-go

integration/cluster/cluster.go
```diff
@@ -0,0 +1,342 @@
...
+	component.grpcPort, err = getFreePort()
+	if err != nil {
+		panic(fmt.Errorf("error allocating GRPC port: %w", err))
+	}
+
+	c.components = append(c.components, component)
+	return component
+}
+
+type Component struct {
```
cyriltovena, line 191: Not using docker is a gift.

integration/loki_micro_services_test.go
cyriltovena: Can you share the panic, I might be able to help getting this fixed.

cyriltovena:
Tests are taking 5s.

> ok  	github.com/grafana/loki/integration	5.367s	coverage: [no statements]

Really want this to be a thing.

cyriltovena (approved):
LGTM

There's definitively improvement we can do but I feel like this is such a great starting point and already has helped debug issue.

slim-bean:
I support this, my concerns about integration tests largely stem from two things:

Runtime and scope creep.

> We should aim to keep this as simple as possible.

> Tests are taking 5s.

I agree with @simonswine on that first point, let's scrutinize additions carefully and see if we can maintain that amazing 5s runtime!

This LGTM, nice work!

Fun fact, I wrote [promtail_test.go](https://github.com/grafana/loki/blob/main/clients/pkg/promtail/promtail_test.go) which is also really an integration test disguised as a unit test. I *think* it's largely been a net positive but we've also chased a lot of failures/flakes in it over the years. 

I think it's tricky to get this kind of testing right, but I definitely think we should merge this and give this a go.

slim-bean (approved):

slim-bean:
I appreciate the discourse in this thread and appreciate y'all valuing my opinions here. I feel like I've come down a bit too hard on integration tests but I have some strong opinions here. 

I really like this approach as others have mentioned, no docker or binaries and you can run the debugger. This is awesome, nice work @simonswine 

I want to merge this to get this rolling, but I also think we should add the SSD mode 👍

integration/client/client.go
simonswine: Yes that's a great idea :+1:

integration/cluster/cluster.go
```diff
@@ -0,0 +1,342 @@
...
+	var a *net.TCPAddr
+	if a, err = net.ResolveTCPAddr("tcp", "localhost:0"); err == nil {
+		var l *net.TCPListener
+		if l, err = net.ListenTCP("tcp", a); err == nil {
+			defer l.Close()
+			return l.Addr().(*net.TCPAddr).Port, nil
+		}
+	}
+	return
+}
```
jeschkies, line 342: Nice one. I was about to hint at this.

jeschkies (commented):
Thanks. I just have two comments.

PR #5970, opened 2022-04-20
fix ingester: grpc method=/logproto.Querier/Query fail.

### LINKED ISSUE #5964: ingester: grpc method=/logproto.Querier/Query fail.
**Describe the bug**
rpc error: code = Internal desc = transport: transport: the stream is done or WriteHeader was already called.

image

image

**To Reproduce**
1. Started Loki (SHA or version)
2. Started Promtail (SHA or version) to tail '...'
3. Query: `{} term`

**Expected behavior**

**Environment:**
 - Infrastructure: [e.g., Kubernetes, bare-metal, laptop]
 - Deployment tool: [e.g., helm, jsonnet]

If applicable, add any output to help explain your problem.

liguozhong:
When ingester's ctx times out, don't try to use grpc to return data to the querier again. 
This PR fixes the ingester error rate > 0 for our loki cluster.
We rely on this error rate indicator for alarming
image

Fixes #5964

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


kavirajk (commented):
changes looks reasonable. Would wait for someone more familiar with ingester to do final check. 
Just to double check this doesn't break any existing behaviour (e.g: querier may still need stats about what happened)

bboreham:
It's not something you added in this PR, but I wanted to note there is a simpler way to do `isDone()`: https://github.com/grafana/loki/blob/557e6e528059691911945bd7a26e8efded84cd08/pkg/ingester/instance.go#L683-L690

Just check `ctx.Err() != nil`, as in the [docs](https://pkg.go.dev/context#Context):
```
	// If Done is not yet closed, Err returns nil.
	// If Done is closed, Err returns a non-nil error explaining why:
```

bboreham:
I find it slightly weird that there are now four calls to `isDone()` in `sendBatches()`, three looking at `ctx` and one looking at `queryServer.Context()`.

If there is a good reason for this, please put it in a comment.

Possibly both loops should check both contexts.
Arguably both loops should only check at one point.

pkg/ingester/instance.go
```diff
@@ -734,7 +736,9 @@ func sendBatches(ctx context.Context, i iter.EntryIterator, queryServer QuerierQ
 
 		stats.AddIngesterBatch(int64(batchSize))
 		batch.Stats = stats.Ingester()
-
+		if isDone(ctx) {
+			break
+		}
 		if err := queryServer.Send(batch); err != nil {
 			return err
```
bboreham, line 743: If the intention is that context cancellation should never result in an error return from this function then I would think you'd want to swallow that here too. (and in similar places)

liguozhong:
> It's not something you added in this PR, but I wanted to note there is a simpler way to do `isDone()`:
> 
> https://github.com/grafana/loki/blob/557e6e528059691911945bd7a26e8efded84cd08/pkg/ingester/instance.go#L683-L690
> 
> Just check `ctx.Err() != nil`, as in the [docs](https://pkg.go.dev/context#Context):
> 
> ```
> 	// If Done is not yet closed, Err returns nil.
> 	// If Done is closed, Err returns a non-nil error explaining why:
> ```

thanks , done.

liguozhong:
> I find it slightly weird that there are now four calls to `isDone()` in `sendBatches()`, three looking at `ctx` and one looking at `queryServer.Context()`.
> 
> If there is a good reason for this, please put it in a comment.
> 
> Possibly both loops should check both contexts. Arguably both loops should only check at one point.

I checked that ctx comes from queryServer.Context() , so I deleted queryServer.Context() and uniformly used ctx as the parameter of isDone() func.

splitice:
I like this commit. These messages have similarly annoyed us during monitoring.

bboreham (approved):
Thanks for responding to my comments.  LGTM.

DylanGuedes:
Do you mind rebasing it? Other than that, LGTM

DylanGuedes (approved):

splitice:
Go away useless bot

MichelHollands:
@liguozhong Can you rebase? Once that is done this can be merged.

liguozhong:
> @liguozhong Can you rebase? Once that is done this can be merged.

done,thanks

PR #5972, opened 2022-04-20
Loki: Modifies TableManager to use IndexGateway ring

DylanGuedes:
- Adds a new entity indexgateway.RingManager, responsible for taking care of the index gateway ring for both, index gateway clients and servers. Now, the IndexGatewayRing service is basically a wrapper for the indexgateway.RingManager.
- Add the IndexGatewayRing service as a dependency for the IndexGateway
- Adds a new OwnsTenant function as a parameter to the shipper. This function replies to whether a tenant is assigned to some instance or not.
- Modifies the TableManager to only download data from tenants assigned to it through the new OwnsTenant function.

N/A

This depends on PR https://github.com/grafana/loki/pull/5946.

**Checklist**
- [x] Documentation added
- [x] Tests updated


pkg/storage/stores/shipper/indexgateway/gateway.go
```diff
@@ -211,12 +237,18 @@ func (g *Gateway) stopping(_ error) error {
...
+		if g.ready {
+			g.indexClient.Stop()
+		}
 	}()
 	return services.StopManagerAndAwaitStopped(context.Background(), g.subservices)
 }
 
 func (g *Gateway) QueryIndex(request *indexgatewaypb.QueryIndexRequest, server indexgatewaypb.IndexGateway_QueryIndexServer) error {
+	if !g.ready {
+		return fmt.Errorf("index gateway not ready for not assigning index client")
```
JordanRushing, line 249:
```suggestion
		return fmt.Errorf("index gateway not yet ready due to index client")
```

JordanRushing (approved):
LGTM!

sandeepsukhani (commented):
I haven't thoroughly reviewed the PR but wanted to point out a problem I see:
We need the read ring passed to `boltdb-shipper` client when it is initialized. With the new storage refactoring, we now have `store` as a dependency for `indexGateway` which means the store would have already initialized the `boltdb-shipper` client before `initIndexGateway` gets called.

Maybe you can modify `initIndexGatewayRing` which gets calls as a dependency on `Store`. Based on whether it is being initialized for `indexGateway` or not, you can initialize a read ring or a rw ring with lifecycler.

DylanGuedes:
> I haven't thoroughly reviewed the PR but wanted to point out a problem I see: We need the read ring passed to `boltdb-shipper` client when it is initialized. With the new storage refactoring, we now have `store` as a dependency for `indexGateway` which means the store would have already initialized the `boltdb-shipper` client before `initIndexGateway` gets called.

But the index client used by the index gateway is initialized inside the `initIndexGateway` function. Do you mind going through how I could debug this? I managed to run the indexgateway and the queriers just fine with this code and we were able to test this in a dev environment.

> Maybe you can modify `initIndexGatewayRing` which gets calls as a dependency on `Store`. Based on whether it is being initialized for `indexGateway` or not, you can initialize a read ring or a rw ring with lifecycler.

The `initIndexGatewayRing` doesn't have access to the lifecycler, though. As of now, only the IndexGateway is starting the lifecycler. That said, are you suggesting us to instead instantiate both the lifecycler and the ring in the `initIndexGatewayRing` and to use them in `initIndexGateway`? If so, we can try that.

pkg/storage/stores/shipper/downloads/table_manager.go
```diff
@@ -309,9 +321,11 @@ func (tm *TableManager) ensureQueryReadiness(ctx context.Context) error {
 			return err
 		}
 
+		perTableStart := time.Now()
 		if err := table.EnsureQueryReadiness(ctx, usersToBeQueryReadyFor); err != nil {
 			return err
 		}
+		level.Debug(util_log.Logger).Log("msg", "instance query ready for users", "users", usersToBeQueryReadyFor, "query_readiness_duration", time.Since(perTableStart), "table", tableName)
```
slim-bean, line 328:
```suggestion
		level.Info(util_log.Logger).Log("msg", "index pre-download for query readiness completed", "users", usersToBeQueryReadyFor, "duration", time.Since(perTableStart), "table", tableName)
```

pkg/storage/stores/shipper/downloads/table_manager.go
```diff
@@ -241,6 +248,11 @@ func (tm *TableManager) cleanupCache() error {
 
 // ensureQueryReadiness compares tables required for being query ready with the tables we already have and downloads the missing ones.
 func (tm *TableManager) ensureQueryReadiness(ctx context.Context) error {
+	start := time.Now()
+	defer func() {
+		tm.metrics.ensureQueryReadinessDurationSeconds.Observe(time.Since(start).Seconds())
```
slim-bean, line 253:
```suggestion
		level.Info(util_log.Logger).Log("msg", "query readiness setup completed", "duration", time.Since(start))
		tm.metrics.ensureQueryReadinessDurationSeconds.Observe(time.Since(start).Seconds())
```

DylanGuedes:
@slim-bean I'll apply your suggestions to PR https://github.com/grafana/loki/pull/5946 instead since this one relies on it.

sandeepsukhani:
> But the index client used by the index gateway is initialized inside the initIndexGateway function, not inside the store service. Do you mind going through how I could debug this? I managed to run the indexgateway and the queriers just fine with this code and we were able to test this in a dev environment.

`boltdb-shipper` index client is a singleton i.e you would always [get the same instance back](https://github.com/grafana/loki/blob/d1aff7b29f94cfb2a2ccb655ca582edaf2688aba/pkg/storage/factory.go#L149). Since the store is now initialized before index gateway, the `boltdb-shipper` client would have been initialized already. You can verify this with some logging or debugger.
The reason I pointed it out is that the ring should be passed to `boltdb-shipper` for query readiness during initialization with the ring tokens already created. If we don't do that we would be doing query readiness on unwanted tables and might miss the ones that are actually required.

> The initIndexGatewayRing doesn't have access to the lifecycler, though. As of now, only the IndexGateway is starting the lifecycler. That said, are you suggesting us to instead instantiate both the lifecycler and the ring in the initIndexGatewayRing and to use them in initIndexGateway? If so, we can try that.
Yes, that is correct. I can't think of any other better way than this.

Please let me know if you need any help with this!

DylanGuedes:
> > But the index client used by the index gateway is initialized inside the initIndexGateway function, not inside the store service. Do you mind going through how I could debug this? I managed to run the indexgateway and the queriers just fine with this code and we were able to test this in a dev environment.
> 
> `boltdb-shipper` index client is a singleton i.e you would always [get the same instance back](https://github.com/grafana/loki/blob/d1aff7b29f94cfb2a2ccb655ca582edaf2688aba/pkg/storage/factory.go#L149). Since the store is now initialized before index gateway, the `boltdb-shipper` client would have been initialized already. You can verify this with some logging or debugger. The reason I pointed it out is that the ring should be passed to `boltdb-shipper` for query readiness during initialization with the ring tokens already created. If we don't do that we would be doing query readiness on unwanted tables and might miss the ones that are actually required.
> 
> > The initIndexGatewayRing doesn't have access to the lifecycler, though. As of now, only the IndexGateway is starting the lifecycler. That said, are you suggesting us to instead instantiate both the lifecycler and the ring in the initIndexGatewayRing and to use them in initIndexGateway? If so, we can try that.
> > Yes, that is correct. I can't think of any other better way than this.
> 
> Please let me know if you need any help with this!

Thanks for the clarification! I pushed a few commits tackling this: the tl;dr is:
- I added a new entity named RingManager
- The initIndexGatewayRing starts the ringmanager
- The ringmanager is then passed to both, shipper and indexgateway

I tested this locally with a k3d kubernetes cluster with both, ring mode and simple mode and had no issues. Let me know if there's something you would like me to test in my env.

pkg/loki/modules.go
```diff
@@ -836,29 +837,16 @@ func (t *Loki) initIndexGatewayRing() (_ services.Service, err error) {
...
-	reg := prometheus.WrapRegistererWithPrefix("loki_", prometheus.DefaultRegisterer)
 
-	logger := util_log.Logger
-	ringStore, err := kv.NewClient(
-		ringCfg.KVStore,
-		ring.GetCodec(),
-		kv.RegistererWithKVName(prometheus.WrapRegistererWithPrefix("loki_", reg), "index-gateway"),
-		logger,
-	)
+	rm, err := indexgateway.NewRingManager(t.Cfg.IndexGateway, util_log.Logger, prometheus.DefaultRegisterer)
```
sandeepsukhani, line 841: Since we are initializing `RingManager` for both index gateway and index gateway client, wouldn't index gateway clients(running in queriers) also register themselves in the ring?

pkg/storage/stores/shipper/downloads/table_manager.go
```diff
@@ -32,6 +32,10 @@ type Limits interface {
 	DefaultLimits() *validation.Limits
 }
 
+type TenantBoundariesClient interface {
+	TenantInBoundaries(tenant string) bool
+}
+
```
sandeepsukhani, line 38: `TenantBoundaries` sounds confusing and does not give much context. What do you think about using function type instead of interface and calling it something like `IndexGatewayOwnsTenant func(tenant string) bool`?

pkg/loki/modules.go
DylanGuedes: shit, you right. I'll look forward on how to fix it.

docs/sources/configuration/_index.md
```diff
@@ -2334,6 +2337,27 @@ backoff_config:
...
+The `index_gateway` block configures the Loki index gateway server, responsible for serving index queries
+without the need to constantly interact with the object store.
+
+```yaml
+# Defines in which mode the index gateway server will operate (default to 'simple').
+# It supports two modes:
+# 'simple': an index gateway server instance is responsible for handling,
+#     storing and returning requests for all indices for all tenants.
+# 'ring': an index gateway server instance is responsible for a subset of tenants instead
+#     of all tenants, such that they don't have to be query ready for all existing index
```
sandeepsukhani, line 2351: We don't have to call out query readiness because it is an optional feature and is disabled by default.

pkg/loki/modules.go
```diff
@@ -836,29 +837,21 @@ func (t *Loki) initIndexGatewayRing() (_ services.Service, err error) {
...
-	ringStore, err := kv.NewClient(
-		ringCfg.KVStore,
-		ring.GetCodec(),
-		kv.RegistererWithKVName(prometheus.WrapRegistererWithPrefix("loki_", reg), "index-gateway"),
-		logger,
-	)
-	if err != nil {
-		return nil, gerrors.Wrap(err, "kv new client")
+	managerMode := indexgateway.ClientMode
+	if t.isModuleActive(IndexGateway) && !t.isModuleActive(Ingester) {
```
sandeepsukhani, line 842: Why do we have to check for `!t.isModuleActive(Ingester)`? I think we can just do `if t.isModuleEnabled(IndexGateway)`?

pkg/storage/stores/shipper/indexgateway/ringmanager.go
```diff
@@ -0,0 +1,249 @@
...
+	log log.Logger
+}
+
+func NewRingManager(managerMode ManagerMode, cfg Config, log log.Logger, registerer prometheus.Registerer) (*RingManager, error) {
+	rm := &RingManager{
+		cfg: cfg, log: log, managerMode: managerMode,
+	}
+
+	if cfg.Mode != RingMode {
+		return rm, nil
```
sandeepsukhani, line 67: I think we should return an error because the code might panic or some other problems can happen which might make it hard to understand the real problem.

pkg/storage/stores/shipper/indexgateway/ringmanager.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	rm.buffers.bufDescs, rm.buffers.bufHosts, rm.buffers.bufZones = ring.MakeBuffersForGet()
```
sandeepsukhani, line 70: This can be moved to `startServerMode` I guess. Also I am not sure if it is a good idea to reuse buffers if `TenantInBoundaries` gets called concurrently. To be safe we can create buffers in `IsAssignedKey` itself?

pkg/loki/modules.go
DylanGuedes: Oops, I mistakenly thought that the `isModuleEnabled` behavior is actually the one implemented by `isModuleActive`. Fixed!

pkg/storage/stores/shipper/indexgateway/ringmanager.go
DylanGuedes: good call. Changed it, lmk what you think.

pkg/loki/loki_test.go
```diff
@@ -120,7 +120,7 @@ func TestLoki_CustomRunOptsBehavior(t *testing.T) {
 	httpPort := ports[0]
 	grpcPort := ports[1]
 
-	yamlConfig := fmt.Sprintf(`target: querier
+	yamlConfig := fmt.Sprintf(`target: distributor
```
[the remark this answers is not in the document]
DylanGuedes: Reverted!

pkg/util/ring.go
```diff
@@ -31,3 +31,25 @@ func IsInReplicationSet(r *ring.Ring, ringKey uint32, address string) (bool, err
...
+		if !rs.Includes(addr) {
+			return false
+		}
+	}
+
+	return true
+}
+
+// HasDedicatedAddress specify an instance support for replying its own address.
+type HasDedicatedAddress interface {
```
sandeepsukhani, line 53: Naming is hard :)
It doesn't sound correct to me. What do you think about using `RingInstanceInfoReader` or simply going back to previous implementation to pass the address directly?

sandeepsukhani (changes requested):
I have added few minor nits but overall code looks great to me.

pkg/util/ring.go
DylanGuedes: Hahaha it is!

Agreed, it looks odd. I returned to pass the address directly, lmk what you think.

pkg/storage/stores/shipper/indexgateway/ringmanager.go
```diff
@@ -0,0 +1,256 @@
...
+
+	svcs := []services.Service{rm.Ring}
+	rm.subservices, err = services.NewManager(svcs...)
+	if err != nil {
+		return errors.Wrap(err, "new index gateway services manager in client mode")
+	}
+
+	rm.subservicesWatcher = services.NewFailureWatcher()
+	rm.subservicesWatcher.WatchManager(rm.subservices)
+	rm.Service = services.NewBasicService(rm.starting, rm.running, rm.stopping)
```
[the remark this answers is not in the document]
DylanGuedes: oops, thanks for catching it. Fixed.

pkg/storage/stores/shipper/gateway_client.go
```diff
@@ -310,6 +315,10 @@ func (s *GatewayClient) ringModeDo(ctx context.Context, callback func(client ind
 			continue
 		}
 
+		if s.cfg.LogGatewayRequests {
+			level.Debug(util_log.Logger).Log("msg", "sending request to gateway", "gateway", addr, "tenant", userID)
+		}
+
```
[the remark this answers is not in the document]
DylanGuedes: lol that explains a lot. I remember having a hard time figuring out why some requests were not being logged :man_facepalming: good catch, ty.

pkg/loki/modules.go
```diff
@@ -864,7 +865,7 @@ func (t *Loki) initIndexGateway() (services.Service, error) {
 		return nil, err
 	}
 
-	t.Server.HTTP.Path("/indexgateway/ring").Methods("GET", "POST").Handler(gateway)
+	t.Server.HTTP.Path("/indexgateway/ring").Methods("GET", "POST").Handler(t.indexGatewayRingManager)
```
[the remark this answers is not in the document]
DylanGuedes: +1, good catch

pkg/util/ring.go
```diff
@@ -32,24 +34,15 @@ func IsInReplicationSet(r *ring.Ring, ringKey uint32, address string) (bool, err
...
-	rs, err := ringClient.Get(token, ring.WriteNoExtend, bufDescs, bufHosts, bufZones)
-	if err == nil {
-		addr := instance.GetInstanceAddr()
-		if !rs.Includes(addr) {
-			return false
-		}
+	inSet, err := IsInReplicationSet(ringClient, token, instanceAddress)
+	if err != nil {
+		level.Error(util_log.Logger).Log("msg", "error checking if key is in replicationset", "error", err, "key", key)
+		return true
```
sandeepsukhani, line 45:
```suggestion
		return false
```
Let us return `false` here to avoid downloading the index for all the tenants when the ring is down. It is okay to not do query readiness for tenants when ring is down since we anyways would download their index at query time but downloading it for all users might make us run out of disk if there are too many tenants.

sandeepsukhani (approved):
just a minor nit, otherwise it LGTM

pkg/util/ring.go
DylanGuedes: Hmmm are you sure it would be better? I think true fits better here because:
- We're not exactly going to download all index for all tenants when the ring is down; what will happen instead is that if the ring is down, the `simple` mode behavior will take place, so we won't have any regressions AFAIK
- If we return false here, we'll do nothing during query readiness in case the ring is down. I'm not sure what is the best behavior though.

That said, if you still think false looks better here just lmk, I think it would work nicely too.

sandeepsukhani: In `simple` mode we do download index for all the tenants, of course if it is configured so :)

DylanGuedes: got it, changed!

sandeepsukhani (approved):
LGTM! Thanks for patiently taking care of all the feedback!

PR #5979, opened 2022-04-21
Call correct method on index store

chaudum:

when index gateway does not implement the new API.

This bug did not yield an error but only returned incorrect results
because both functions `c.IndexStore.LabelNamesForMetricName` and
`c.IndexStore.LabelValuesForMetricName` return `([]string, error)`.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/storage/stores/series/series_index_gateway_store.go
+++ b/pkg/storage/stores/series/series_index_gateway_store.go
@@ -88,7 +88,7 @@ func (c *IndexGatewayClientStore) LabelValuesForMetricName(ctx context.Context, 
 	})
 	if isUnimplementedCallError(err) {
 		// Handle communication with older index gateways gracefully, by falling back to the index store calls.
-		return c.IndexStore.LabelNamesForMetricName(ctx, userID, from, through, metricName)
+		return c.IndexStore.LabelValuesForMetricName(ctx, userID, from, through, metricName, labelName, matchers...)
 	}
 	if err != nil {
 		return nil, err


kavirajk (approved):
LGTM

pkg/storage/stores/series/series_index_gateway_store.go
```diff
@@ -88,7 +88,7 @@ func (c *IndexGatewayClientStore) LabelValuesForMetricName(ctx context.Context,
 	})
 	if isUnimplementedCallError(err) {
 		// Handle communication with older index gateways gracefully, by falling back to the index store calls.
-		return c.IndexStore.LabelNamesForMetricName(ctx, userID, from, through, metricName)
+		return c.IndexStore.LabelValuesForMetricName(ctx, userID, from, through, metricName, labelName, matchers...)
```
kavirajk, line 91: is it tested?

simonswine (approved):
LGTM Thanks for catching this :+1:

pkg/storage/stores/series/series_index_gateway_store.go
simonswine: No it is not currently, but I have #5968 prepared, that covers this aspect.

chaudum: I'll try to find a way to test this without a huge amount of boilerplate code.

PR #5982, opened 2022-04-21
Add string representation of boltdb shipper mode

simonswine:
Makes the log output significantly more meaningful.

Before:

```
level=info ts=2022-04-21T07:35:01.687187409Z caller=shipper_index_client.go:107 msg="starting boltdb shipper in 0 mode"
```

After:

```
level=info ts=2022-04-21T07:35:01.687187409Z caller=shipper_index_client.go:107 msg="starting boltdb shipper in read-write mode"
```


kavirajk (approved):
Neat :)

periklis (approved):

Issue #5843, opened 2022-04-08
operator: Support for reconciling Loki Rules


mladedav:
**Is your feature request related to a problem? Please describe.**
I want to declaratively create Loki alert rules for my applications.

**Describe the solution you'd like**
For the Loki operator to be able to watch for a new CR LokiRule, or to somehow reuse and watch for PrometheusRule

**Describe alternatives you've considered**
Using current imperative tooling possibly through Helm hooks to create these during installation.

**Additional context**
Similar requests that have died out:
https://github.com/grafana/loki/issues/2990
https://github.com/grafana/loki/issues/3456
Implementation that could be used as a starting point:
https://github.com/opsgy/loki-rule-operator

I would be happy to help out with this

periklis:
@mladedav Thank you for reporting this issue, the SIG operator is currently working on a design proposal for integration a CRD for Loki Rules in the operator. Please stay tuned!

FYI, the loki-operator hosted here in this repository goes beyond reconciling rules. It is a fully fledged operator managing Loki installations on Kubernetes and OpenShift clusters. I will provide an overview on the next community meeting.

rosscdh:
+1 something similar to the service monitor annotations would be nice; also the ability to design an write the alert in the ui but then export the json (as per dashboards) would be super awesome

Jeinhaus:
Wouldn't a sidecar, that syncs labeled configmaps to a local volume work as well? Similar to the way the Grafana provisioning for dashboards works?
I added [this PR](https://github.com/grafana/helm-charts/pull/1625) in the old repository to provide that functionality.
Do you think I should add it here as well?

Issue #5211, opened 2022-01-23
operator: Add support for ruler component


eranra:
Loki stack includes a new GA'ed component called Ruler. The `Ruler` component marriage metrics with logs closing the gap for customers to achieve a real **consolidated observability** solution. The Ruler component is responsible for continually evaluating a set of configurable queries and performing actions based on the result.  There are two key types of rules that the `Ruler` component  supports  [Alerting Rules](https://grafana.com/docs/loki/latest/rules/#alerting-rules) and [Recording Rules](https://grafana.com/docs/loki/latest/rules/#recording-rules) 

Note: In addition to the direct business value, this PR also aligns Loki Operator functionality with Loki Helm functionality. Loki Helm already provides the ability to deploy the `Ruler` component [here](https://github.com/grafana/helm-charts/tree/main/charts/loki-distributed/templates/ruler)

With this new feature, customers will be able to selectively deploy Loki  `Ruler` component as part of LokiStack operator deployments and be able to define a set of rules and any additional `Ruler` configuration needed for the `Ruler` component to function as described in the `Ruler` API documentation [here](https://grafana.com/docs/loki/latest/configuration/#ruler_config).

As an alternative customers can use helm charts and not use Loki operator when the `Ruler` component is required.

Customer-facing API:

It is suggested that the Operator configuration for the `Ruler` component will be common to the configuration used by other deployments such as the Helm charts. This includes a config-map that will be created by customers. The config-map points to a yaml file holding the `Ruler` rules. This is the same approach used by the Helm chart solution [here](production/helm/loki/templates/configmap-alert.yaml)

Example for  such `ruler_rules.yaml` file::

```
groups:
  - name: PlatformRules
    interval: 1m
    rules:
    - record: logs:openshift:platform:rate1m
      expr: |
        sum(rate({namespace=~"openshift+"}[1m]))
      labels:
        metric: "rate"
        source: "logs"
```

eranra:
Initial implementation in PR https://github.com/grafana/loki/pull/5188

eranra:
@periklis @Red-GV @jacobsakran please review

eranra:
👍

Red-GV:
I believe this was done with: https://github.com/grafana/loki/pull/6195 and https://github.com/grafana/loki/pull/5986

PR #5985, opened 2022-04-21
operator: Ruler enhancement proposal

periklis:
The following PR adds a enhancement proposal for adding Ruler support into the Loki Operator.

Addresses #5211 and #5843

**Checklist**
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/operator/docs/enhancements/ruler_support.md
+++ b/operator/docs/enhancements/ruler_support.md
@@ -70,7 +70,7 @@ type RulesSpec struct {
     // +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch",displayName="Enable"
     Enabled bool `json:"enabled"`
 
-    // A selector to select which LokiRules to mount for loading alerting/recording
+    // A selector to select which Loki rules to mount for loading alerting/recording
     // rules from.
     //
     // +optional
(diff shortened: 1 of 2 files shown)


simonswine (approved):
Thanks for putting the design document together. I think this is mostly LGTM.

I would like to see a more flexible mapping of Rules to Stacks, but that is not a hard requirement for merging.

cyriltovena (approved):

operator/docs/enhancements/ruler_support.md
```diff
@@ -0,0 +1,686 @@
...
+
+**Mitigation**: Introduce some sort of sharding of LokiRules into multiple ConfigMap resources.
+
+## Design Details
+
+### Open Questions [optional]
+
+1. Do we need to split the `LokiRule` into two types `RecordingRule` and `AlertingRule`?
+2. Do we need to add support for Remote Write Configuration per tenant? (See `ruler_remote_write_*` parameters in [limits_config](https://grafana.com/docs/loki/latest/configuration/#limits_config))
+3. Do we need to add support for the `ruler_evaluation_delay_duration`, `ruler_max_rules_per_rule_group` and `ruler_max_rule_groups_per_tenant` limits?
```
periklis, line 671: @cyriltovena @simonswine What is your opinion on these open question?

operator/docs/enhancements/ruler_support.md
```diff
@@ -0,0 +1,686 @@
...
+    // Defines the configuration for remote write client queue.
+    //
+    // +optional
+    // +kubebuilder:validation:Optional
+    // +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:advanced",displayName="Client Queue"
+    QueueSpec *RemoteWriteClientQueueSpec `json:"queueSpec"`
+}
+
+// LokiRulerSpec defines the desired state of LokiRuler
+type LokiRulerSpec struct {
```
xperimental, line 541: I think this would also be a good place for the selectors that Simon proposed for `LokiRule` instead of using the `StackName`. I like the idea of using matchers to find the rules, so that we are able to re-use rules across multiple stacks.

operator/docs/enhancements/ruler_support.md
```diff
@@ -0,0 +1,686 @@
...
+
+**Risk**: The proposed reconciliation approach manifests that all `LokiRule` instances are transformed to individual entries in a single ConfigMap. Alerting/Recording rules can be become quite big (e.g. large LogQL expresssions). As per ConfigMaps represent Kubernetes resource stored in etcd, this might hit some store limits (See [1MiB limit](https://kubernetes.io/docs/concepts/configuration/configmap/#motivation)). Therefore storing all rules in a single ConfigMap might become impossible over time or on large clusters.
+
+**Mitigation**: Introduce some sort of sharding of LokiRules into multiple ConfigMap resources.
+
+## Design Details
+
+### Open Questions [optional]
+
+1. Do we need to split the `LokiRule` into two types `RecordingRule` and `AlertingRule`?
```
xperimental, line 669: This sounds like a good idea to me. Currently there is the possibility to specify both `Alert` and `Record` in a rule.

operator/docs/enhancements/ruler_support.md
```diff
@@ -0,0 +1,686 @@
...
+
+In detail the separation of concerns between both CRDs looks like:
+1. `LokiStack`: Defines all required aspects to spin up a Loki ruler component, i.e. size, node placement. In addition it controls the common config settings for the Work-Ahead-Log and Ring configuration.
+2. `LokiRulerConfig`: Defines only the global runtime settings for the ruler, i.e. evaluation/poll intervals, AlertManager configuration, remote write client configuration.
+
+#### General constraints
+
+1. The above `LokiRulerConfig` is limited to a single global remote write endpoint for exporting metrics from recording rules. This leaves solving multi-tenancy issues on the remote write server side and provide means (e.g. headers) for spliting/amending metrics ingestion per tenant. On the other hand it simplifies ruler operations as it does not require to spin concurrent remote write clients per tenant.
+2. Additionally the `LokiRulerConfig** requires a user-provided Kubernetes Secret for sensitive information. This adds an extra dependency that requires validation inside the controller-loop.
```
xperimental, line 643:
```suggestion
2. Additionally the `LokiRulerConfig` requires a user-provided Kubernetes Secret for sensitive information. This adds an extra dependency that requires validation inside the controller-loop.
```

operator/docs/enhancements/ruler_support.md
```diff
@@ -0,0 +1,686 @@
...
+// +kubebuilder:validation:Pattern:="((([0-9]+)y)?(([0-9]+)w)?(([0-9]+)d)?(([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?|0)"
+type EvaluationDuration string
+
+// LokiRuleSpec defines the desired state of LokiRule
+type LokiRuleSpec struct {
+    // Name of the LokiStack to reconcile the rules for
+    //
+    // +required
+    // +kubebuilder:validation:Required
+    StackName string `json:"stackName"`
```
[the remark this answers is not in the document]
periklis: Thanks for raising this topic. I have wondered myself how the Prometheus-operator solved this topic. PTAL again I have updated the `LokiRulesSpec` to support a rule Selector and a NamespaceSelector. Same goes for the implement PR.

operator/docs/enhancements/ruler_support.md
jan--f: I agree, if you can, provide two distinct types.
I don't know the history or any arguments that made Prometheus go with one "type" (in quotes since for prometheus its just yaml) that can be rule or alert, so I might be missing upsides of this.
 
One downside we are experiencing is that when one would like to give users the option to create alerting rules, but explicitly not recording rules. With a type that can be either one has to resort to either a validating webhook which muddies the user experience a bit or one ends up creating a new type. See https://github.com/openshift/enhancements/pull/1030#discussion_r854971161 for more context.

The ability to specify which rule variant one expects can be very handy. And you'll probably end up with a union type anyway.

periklis: Thanks for the feedback @jan--f . I have split the `LokiRule` type into an `AlertingRule` and a `RecordingRule` types. This brings a lot of clarity in using the APIs in distinct scenarios. Plus it helps users to apply different RBAC on each one.

bwplotka: Eh, forgot to submit my feedback... too late 3c:

I think in some way it makes sense to keep the semantics of https://github.com/prometheus-operator/prometheus-operator PrometheusRule. At the end simplicity of the API is important and less components = the better. Also the ecosystem got use to this, so it's might be easier to use.

@periklis @jan--f  the history is most likely coming from Prometheus itself: one type is easier than two and both types are somewhat "rules" (one is recording, second alering). In Prometheus we use same rules manager, same rules API (/api/v1/rules) for both on Prometheus given similar semantics (`expr`).

I think both ways work, but consistency is key here. It's weird I can list all my `lokiAlertRule` and `lokiRecordRule` in Kube API for Loki, but I have only `PrometeusRule` in Prom operator. It would be amazing to move either way for both projects (: (and ideally have Prom Alert easily convertible to Loki Alert for ease of use and so on...)

bwplotka (approved):
LGTM! Great work. For my taste, there are too many CR details discussed here that might be discussed in PR.

I think I would vote for fewer API components and have LokiRule for a start - no need to confuse people and break consistency here. Similar as it was done in PromRule.

operator/docs/enhancements/ruler_support.md
```diff
@@ -0,0 +1,844 @@
...
+* 2022-04-28: Update draft to split `LokiRule` into two distinct types `AlertingRule` and `RecordingRule`. (See [PR](https://github.com/grafana/loki/pull/5986))
+* 2022-04-28: Renamed `LokiRulerConfig` types to `RulerConfig` as per `Loki` prefix is part of the API group `loki.grafana.com`, i.e. fully qualified type is `ruler.loki.grafana.com`
+
+## Drawbacks
+
+The above proposed design and implementation for LokiStack Ruler support adds a significant amount of new APIs as well as complexity into exposing a declarative approach **only** for the ruler component. Regardless the hard effort to minimize the amount of configuration settings in the proposed CRDs it remains still a huge addition to the operator code base. In contrast to the existing `LokiStack` CRD that is a very slim set of Loki settings (Note: without considering the gateway tenant configuration), the `RulerConfig` is almost one-to-one identical to the [ruler config](https://grafana.com/docs/loki/latest/configuration/#ruler).
```
owen-d, line 838: I would consider slimming down some of these specs  and instead hardcoding/using defaults until the need for such complexity is proven. This would reduce the new surface area by a large margin and options could be added incrementally as needed.

owen-d (commented):
Left a few small suggestions, but overall looks reasonable to me.

owen-d (approved):

operator/docs/enhancements/ruler_support.md
```diff
@@ -0,0 +1,844 @@
...
+```
+/tmp/rules/application/ns-a-alerting-rule-a-alerts.yaml
+          /application/ns-a-recording-rule-a-recs.yaml
+          /infrastructure/ns-b-alerting-rule-b-alerts.yaml
+          /infrastructure/ns-b-recording-rule-b-recs.yaml
+```
+
+5. The `AlertingRuleController` listens for `AlertingRule` create/update/delete events and it applies the `loki.grafana.com/rulesDiscoveredAt: time.Now().Format(time.RFC3339)` on each `LokiStack` instance on the cluster. This ensures that the `LokiStack` reconciliation loop starts anew.
+6. The `RecordingRuleController` listens for `RecordingRule` create/update/delete events and it applies the `loki.grafana.com/rulesDiscoveredAt: time.Now().Format(time.RFC3339)` on each `LokiStack` instance on the cluster. This ensures that the `LokiStack` reconciliation loop starts anew.
+7. The `LokiStackController` compiles on each create/update/delete event a SHA1 of the ConfigMap data entries and appends dedicated annotations `loki.grafana.com/alerting-rules-hash` and `loki.grafana.com/recording-rules-hash` to the ruler pod spec. If any of the hash changes, the change applied to the annotation value and the ruler pods get restarted.
```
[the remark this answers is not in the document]
periklis: Good point, I forgot that configmap content is auto-populated into the containers by kubernetes.

operator/docs/enhancements/ruler_support.md
periklis: @owen-d From a user experience perspective the amount of non-defaulted fields the user needs to declare thanks to `+kubebuilder:default` annotation is limited to a minimum. Do you foresee any upcoming changes for `remote_write` or `alertmanager` (e.g. fields that will be deprecated, removed or replaced)?

operator/docs/enhancements/ruler_support.md
periklis: Good point, I've update the proposal by removing this field.

operator/docs/enhancements/ruler_support.md
```diff
@@ -0,0 +1,686 @@
...
+}
+
+// LokiRuleGroup defines a group of Loki alerting and/or recording rules.
+type LokiRuleGroup struct {
+    // Name defines a name of the present recoding/alerting rule. Must be unique
+    // within all loki rules.
+    //
+    // +required
+    // +kubebuilder:validation:Required
+    Name string `json:"name"`
```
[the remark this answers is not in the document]
periklis: > > Must be unique within all loki rules
> 
>     * What are the consequences of duplicate group names? (warning, duplicates are ignored, hard error)
> 

I am not sure where/why Prometheus maintains this limitation and Loki adopts it. But I consider that rule group names are used like IDs from tools like cortextool, e.g. examplary output from cortextool:
```
Namespace                                                                              | Rule Group
openshift-operators-redhat-loki-operator-dev-163b29f5-bc65-4289-9457-c8f94269d0e2.yaml | should_operator_record
openshift-operators-redhat-loki-operator-dev-d908b623-b32e-4cd6-8800-75323199c706.yaml | should_operator_fire_on_errors
openshift-operators-redhat-loki-operator-dev-d908b623-b32e-4cd6-8800-75323199c706.yaml | credentials_operator_leak
```
Currently the proposed validation webhook would disallow creating multiple groups in the same rule.

>     * Would it be an idea to auto-generate the group names in the operator based on the namespace/name of the `LokiRule`? What about only having one group per `LokiRule`?

I suggest to support user-defined group names as well as multiple groups inside a rule to maintain similiarity as close as possible to `PrometheusRule`.

operator/docs/enhancements/ruler_support.md
periklis: > Eh, forgot to submit my feedback... too late 3c:
> 
> I think in some way it makes sense to keep the semantics of https://github.com/prometheus-operator/prometheus-operator PrometheusRule. At the end simplicity of the API is important and less components = the better. Also the ecosystem got use to this, so it's might be easier to use.
> 
> @periklis @jan--f the history is most likely coming from Prometheus itself: one type is easier than two and both types are somewhat "rules" (one is recording, second alering). In Prometheus we use same rules manager, same rules API (/api/v1/rules) for both on Prometheus given similar semantics (`expr`).

This will continue to be the case with Loki if you use the Rules API. You will be able to retrieve both types of rules from the Loki. The `AlertingRule` and `RecordingRule` CRD are just a storage form for k8s and both get translated into rules that Loki get read, e.g. examplary output from cortex tool:

```
Namespace                                                                 | Rule Group
openshift-logging-lokistack-dev-17f964d8-35bc-4586-89d2-4f071b36c090.yaml | should_record
openshift-logging-lokistack-dev-623fa43e-3083-450c-959d-db575016bc33.yaml | should_fire
openshift-logging-lokistack-dev-623fa43e-3083-450c-959d-db575016bc33.yaml | credentials_leak
```
where `should_record` is a k8s `RecordingRule` and `should_fire` is a k8s `AlertingRule`

> 
> I think both ways work, but consistency is key here. It's weird I can list all my `lokiAlertRule` and `lokiRecordRule` in Kube API for Loki, but I have only `PrometeusRule` in Prom operator. It would be amazing to move either way for both projects (: (and ideally have Prom Alert easily convertible to Loki Alert for ease of use and so on...)

I would say this is a call for the prometheus-operator to re-consider splitting into two types in future. I won't say consistency is bad or neglible, but AFAICS and as @jan--f mentions having RBAC-control on different types is something that users want to have. For example alerts are simple and can on misuse just be noisy on firing. OTH recording rules need attention before processing because they might overload a remote write endpoint. Thus admins would possible want to limit who can create recording rules, while leave alerting rules for general use free.

operator/docs/enhancements/ruler_support.md
```diff
@@ -98,142 +124,218 @@ type LokiStackComponentStatus struct {
 }
 ```
 
-#### LokiRule definition
+#### AlertingRule definition
```
xperimental, line 127: Should `AlertingRule` and `RecordingRule` be prefixed with `Loki`? They will not conflict with Prometheus (should the other operator ever decide to split the rules up as well) because of the group, but this might be more convenient / visible for users.

operator/docs/enhancements/ruler_support.md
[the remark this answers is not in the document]
xperimental: > Currently the proposed validation webhook would disallow creating multiple groups in the same rule.

Yes, that should be possible. Also groups with the same name across multiple rules, but this makes the validation harder and less transparent. The offending other rule might be in a different namespace that I as a user do not have write access to but want to use in my Loki using the label-based configuration. This might also mean that the validation hook would need to know about that configuration (which LokiStack matches which Rules, so that conflicts can be found correctly).

> I suggest to support user-defined group names as well as multiple groups inside a rule to maintain similiarity as close as possible to PrometheusRule.

Similarity to the existing prometheus-operator / PrometheusRule is a valid argument. I need to check if and how the issue is addressed there.

For me, auto-generated group names still sound like a convenient and efficient way of getting rid of the validation issue.

xperimental (commented):
A few more comments. Sorry for forgetting to click the send button last week.

operator/docs/enhancements/ruler_support.md
```diff
@@ -61,15 +61,41 @@ In summary the `LokiStack` ruler component will support by default minimum the f
...
 ...
-    // EnableRuler defines a flag to enable/disable the ruler component
+    // Rules defines the spec for the ruler component
     //
     // +optional
     // +kubebuilder:validation:Optional
-    // +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch",displayName="Enable Ruler"
-    EnableRuler bool `json:"enableRuler"`
+    // +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:advanced",displayName="Rules"
+    Rules *RulesSpec `json:"rules,omitempty"`
```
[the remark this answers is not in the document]
periklis: Nope this is correct. The `LokiStackSpec` includes a spec for enabling the ruler and on how to select rules based on selectors and this spec is called `RulesSpec`. The name maintains coherency with other Loki runtime aspects, e.g. `LimitsSpec` for limits_config, `TenantsSpec` for gateway tenants

operator/docs/enhancements/ruler_support.md
periklis: The proposal started to use `Loki` as a prefix, but toward refinement I realized that the typical kubernetes naming follows the GroupVersion approach. Thus when we speak of `AlertingRule` we mean `alertingrule.loki.grafana.com/v1beta1` and similar for `RecordingRule` we speak of `recordingrule.loki.grafana.com/v1beta1`. Ergo prefixes make names longer and double the amount of `loki` references in the full name.

operator/docs/enhancements/ruler_support.md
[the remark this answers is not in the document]
periklis: > > Currently the proposed validation webhook would disallow creating multiple groups in the same rule.
> 
> Yes, that should be possible. Also groups with the same name across multiple rules, but this makes the validation harder and less transparent. The offending other rule might be in a different namespace that I as a user do not have write access to but want to use in my Loki using the label-based configuration. This might also mean that the validation hook would need to know about that configuration (which LokiStack matches which Rules, so that conflicts can be found correctly).
> 

The requirement of unique group names applies only within **a single** rule and not across rules. TBH as far as I understand tooling handling/output, e.g. `cortextool` the rule name is creating some sort of namespace and within this all group names are required to be unique, e.g.:

```
Namespace                                                                              | Rule Group
openshift-operators-redhat-loki-operator-dev-d908b623-b32e-4cd6-8800-75323199c706.yaml | should_operator_fire_on_errors
openshift-operators-redhat-loki-operator-dev-d908b623-b32e-4cd6-8800-75323199c706.yaml | credentials_operator_leak
```

> > I suggest to support user-defined group names as well as multiple groups inside a rule to maintain similiarity as close as possible to PrometheusRule.
> 
> Similarity to the existing prometheus-operator / PrometheusRule is a valid argument. I need to check if and how the issue is addressed there.
> 
> For me, auto-generated group names still sound like a convenient and efficient way of getting rid of the validation issue.

I believe auto-generating group names is defeating the purpose of users giving names that mean something to them and their workloads. As longs as we don't need to provide uniqueness across rules, I believe we can leave this field for user edits.

operator/docs/enhancements/ruler_support.md
cyriltovena: 1. Not sure we do.
2. Could be useful at some point.
3. I guess this is more for expert type of operations, so not having at first is fine.

operator/docs/enhancements/ruler_support.md
owen-d: I don't see any changes coming, but @dannykopping may have a better sense than I.

operator/docs/enhancements/ruler_support.md
xperimental: This might be just me, but I use the "unqualified names" of Kubernetes types a lot, so I would just write out "alertingrule" when using kubectl. As I said this is just for convenience, so it's fine the way it currently is.

PR #5986, opened 2022-04-21
operator: Add basic support for Loki ruler and rules reconciliation

periklis:
This PR is first-shot implementation for the proposed enhancement: #5985

Implements #5985

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


cyriltovena:
Went through it all where is the rules API reconciliation with the ruler ?

periklis:
> Went through it all where is the rules API reconciliation with the ruler ?

You mean the `RulerConfig` reconciliation for AlertManager and RemoteWrite? If yes this will be a separate PR. This one is focusing on `AlertingRule` and `RecordingRule`.

operator/Makefile
```diff
@@ -103,7 +103,7 @@ help: ## Display this help.
 
 .PHONY: deps
 deps: go.mod go.sum
-	go mod tidy
+	go mod tidy -compat=1.17
```
xperimental, line 106: I have a really bad feeling about the error message that `go mod tidy` is outputting when used without the compat option. Maybe we need to discuss splitting out the API dependency as a separate module from Loki?

operator/hack/lokistack_gateway_ocp.yaml
```diff
@@ -3,11 +3,19 @@ kind: LokiStack
 metadata:
   name: lokistack-dev
 spec:
-  size: 1x.extra-small
+  size: 1x.small
```
xperimental, line 6: Is this an intended change? It will make our testing deployments much larger...

operator/hack/lokistack_gateway_ocp.yaml
```diff
@@ -3,11 +3,19 @@ kind: LokiStack
...
   storage:
     secret:
       name: test
       type: s3
   storageClassName: gp2
   tenants:
     mode: openshift-logging
+  rules:
+    enabled: true
```
xperimental, line 15: Ruler enabled for all our testing clusters in the future? Or should we have a separate config file?

operator/internal/manifests/mutate.go
```diff
@@ -121,6 +121,7 @@ func mergeWithOverride(dst, src interface{}) error {
 
 func mutateConfigMap(existing, desired *corev1.ConfigMap) {
 	existing.BinaryData = desired.BinaryData
+	existing.Data = desired.Data
```
xperimental, line 124: Do you know why `Data` was omitted in the original code? This seems like a subtle but significant change...

xperimental (commented):
I'm about half-way through the content of the PR, but here's the comments so far.

operator/PROJECT
```diff
@@ -5,15 +5,39 @@ plugins:
   manifests.sdk.operatorframework.io/v2: {}
   scorecard.sdk.operatorframework.io/v2: {}
 projectName: loki-operator
-repo: github.com/grafana/loki
+repo: github.com/grafana/loki/operator
```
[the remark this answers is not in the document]
periklis: Unfortunately `operator-sdk`'s kubebuilder scaffolding facility requires that this value is equal to go.mod's module:
```
module github.com/grafana/loki/operator
```
This enables to use:
```
operator-sdk-v1.19.1 create api --group loki --version v1beta1 --kind AlertingRule --namespaced=true --controller --resource --verbose
```

operator/api/v1beta1/alertingrule_webhook.go
```diff
@@ -0,0 +1,124 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
+var alertingrulelog = logf.Log.WithName("alertingrule-resource")
```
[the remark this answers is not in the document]
periklis: You are right, this came in from the code generator.

operator/config/crd/patches/webhook_in_alertingrules.yaml
```diff
@@ -0,0 +1,16 @@
...
+  name: alertingrules.loki.grafana.com
+spec:
+  conversion:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          namespace: system
+          name: webhook-service
+          path: /convert
```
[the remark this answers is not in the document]
periklis: This a generated file in case we want later to enable a mutating webhook for conversions. This is not referenced anywhere. This is auto-generated TBH. In fact if we use it ever, the namespace will be replace by the kustomization options of each overlay, e.g.:
```
# Adds namespace to all resources.
namespace: openshift-logging
```

operator/config/crd/patches/webhook_in_alertingrules.yaml
```diff
@@ -0,0 +1,16 @@
...
+      conversionReviewVersions:
+      - v1
```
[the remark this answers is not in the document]
periklis: This is auto-generated and in fact unused. I leave this to the generated whenever we introduce a mutating webhook for conversions.

operator/config/default/manager_webhook_patch.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: apps/v1
```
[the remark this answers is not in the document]
periklis: Yes they are auto-generated and in fact we need to delete these files because we use the overlays.

operator/hack/lokistack_gateway_ocp.yaml
periklis: Nope this change is a mistake. Will return this back to 1x.extra-small

operator/hack/lokistack_gateway_ocp.yaml
periklis: I opt for using the ruler in testing by default in the `_ocp` suffixed hack manifests as per ruler will be an important aspect for user buy-in.

operator/internal/manifests/mutate.go
periklis: We use to store everything in `BinaryData`, i.e. the loki config, the tenant overrides, etc. because there was the fear that Loki might introduced config options that are non-UTF8. This makes reading configuration a bit harder because we need to pipe everything into a bas64 decoder. I opted here in to make the rules configmap easier to read.

operator/api/v1beta1/alertingrule_types.go
```diff
@@ -0,0 +1,134 @@
...
+	// List of groups for alerting rules.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Groups"
+	Groups []*AlertingRuleGroup `json:"groups"`
+}
+
+// AlertingRuleGroup defines a group of Loki alerting rules.
+type AlertingRuleGroup struct {
```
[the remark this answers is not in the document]
periklis: What is the benefit of such a struct? IMHO it makes writing the golang types a bit harder, because one needs to type the extra embedded types for the common fields.

operator/Makefile
periklis: The problem here is our new dependency against Loki itself, in particular we reference: 

```
github.com/grafana/loki/operator/internal/manifests/internal/rules imports
        gopkg.in/yaml.v2 tested by
        gopkg.in/yaml.v2.test imports
        gopkg.in/check.v1 loaded from gopkg.in/check.v1@v1.0.0-20200902074654-038fdea0a05b,
        but go 1.16 would select v1.0.0-20201130134442-10cb98267c6c
github.com/grafana/loki/operator/api/v1beta1 imports
        github.com/grafana/loki/pkg/logql/syntax imports
        github.com/grafana/loki/pkg/util imports
        github.com/grafana/dskit/kv imports
        github.com/grafana/dskit/kv/consul imports
        github.com/hashicorp/consul/api loaded from github.com/hashicorp/consul/api@v1.12.0,
        but go 1.16 would fail to locate it:
        ambiguous import: found package github.com/hashicorp/consul/api in multiple modules:
        github.com/hashicorp/consul v1.2.1 (/home/periklis/.gvm/pkgsets/go1.17.8/global/pkg/mod/github.com/hashicorp/consul@v1.2.1/api)
        github.com/hashicorp/consul/api v1.12.0 (/home/periklis/.gvm/pkgsets/go1.17.8/global/pkg/mod/github.com/hashicorp/consul/api@v1.12.0)

To upgrade to the versions selected by go 1.16, leaving some packages unresolved:
        go mod tidy -e -go=1.16 && go mod tidy -e -go=1.17
If reproducibility with go 1.16 is not needed:
        go mod tidy -compat=1.17
For other options, see:
        https://golang.org/doc/modules/pruning
make: *** [Makefile:106: deps] Error 1
```

operator/controllers/recordingrule_controller.go
```diff
@@ -0,0 +1,52 @@
...
+			RequeueAfter: time.Second,
+		}, err
+	}
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *RecordingRuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&lokiv1beta1.RecordingRule{}).
```
xperimental, line 50: Do we need to watch Namespaces as well, so that the controller gets notified when labels change possibly resulting in different rules being applied?

operator/internal/handlers/internal/rules/rules.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	return lrl, nil
+}
+
+// ListRecordingRules returns a k8s resource list of AlertingRules for the given spec or an error. Three cases apply:
+// - Return only matching rules in the stack namespace if no namespace selector given.
+// - Return only matching rules in the stack namespace and in namespaces matching the namespace selector.
+// - Return no rules if rules selector does not apply at all.
+func ListRecordingRules(ctx context.Context, k k8s.Client, stackNs string, rs *lokiv1beta1.RulesSpec) (lokiv1beta1.RecordingRuleList, error) {
```
xperimental, line 47: Do we really need two functions for this? I think a single `ListRules()` which returns both alerting and recording rules might be all that is needed.

operator/internal/manifests/config.go
```diff
@@ -42,6 +42,11 @@ func LokiConfigMap(opt Options) (*corev1.ConfigMap, string, error) {
 
 // ConfigOptions converts Options to config.Options
 func ConfigOptions(opt Options) config.Options {
+	var rulerEnabled bool
+	if opt.Stack.Rules != nil && opt.Stack.Rules.Enabled {
+		rulerEnabled = true
+	}
```
xperimental, line 48: Convert to direct bool assignment?
```suggestion
	rulerEnabled := opt.Stack.Rules != nil && opt.Stack.Rules.Enabled
```

operator/internal/manifests/distributor.go
```diff
@@ -15,12 +15,14 @@ import (
...
-	dataDirectory     = "/tmp/loki"
-	secretDirectory   = "/etc/proxy/secrets"
+	walVolumeName          = "wal"
+	configVolumeName       = "config"
+	rulesStorageVolumeName = "rules"
+	storageVolumeName      = "storage"
+	walDirectory           = "/tmp/wal"
+	dataDirectory          = "/tmp/loki"
+	rulesStorageDirectory  = "/tmp/rules"
+	secretDirectory        = "/etc/proxy/secrets"
```
xperimental, line 25: Does it make sense to group this section by "component"? (for example putting the two `rules` constants together)

I also would not have expected this section to be in `distributor.go`, if I am honest :grinning:

operator/internal/manifests/internal/gateway/options.go
```diff
@@ -28,6 +28,5 @@ type Secret struct {
 
 // TenantData defines the existing tenantID and cookieSecret for lokistack reconcile.
 type TenantData struct {
-	TenantID     string
```
xperimental, line 31: It looks to me as if `CookieSecret` is also not read from here anymore but directly from `OpenShiftOptions.Authentication`. On the other hand I'm also not sure if `TenantID` is still read somewhere.

xperimental (commented):
And another round, still not fully through the files (61/93). The good news is that I combined it with actually running the new version of the operator on a testing cluster.

Generally I have the feeling that there is some "de-duplication" of the alertingrule/recordingrule code possible, maybe by introducing some interfaces?

I'll continue tomorrow.

operator/hack/lokistack_gateway_ocp.yaml
```diff
@@ -11,3 +11,11 @@ spec:
...
+    selector:
+      matchLabels:
+        loki.grafana.com/cluster-monitoring: "true"
+    namespaceSelector:
+      matchLabels:
+        loki.grafana.com/cluster-monitoring: "true"
```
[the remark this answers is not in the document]
periklis: Indeed this makes sense as it aligns with OCP's cluster-monitoring norms: `openshift.io/cluster-monitoring: "true"`

operator/internal/manifests/distributor.go
periklis: You are right, this belongs into `var.go`, however I saw the other day that @Red-GV schema-migration PR did the right move. So keeping this here for now.

operator/config/overlays/openshift/kustomization.yaml
```diff
@@ -31,60 +26,15 @@ labels:
...
+- auth_proxy_service_annotations_patch.yaml
 - manager_auth_proxy_patch.yaml
 - manager_related_image_patch.yaml
 - manager_run_flags_patch.yaml
+- manager_webhook_patch.yaml
 - prometheus_service_monitor_patch.yaml
+- webhook_service_annotations_patch.yaml
 
-# apiVersion: kustomize.config.k8s.io/v1beta1
-# kind: Kustomization
```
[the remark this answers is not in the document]
periklis: I am not sure I understand the ask here. I removed these lines because they were unused/obsolete before. What do I miss here? Was the issue already in place by having the commented?

operator/api/v1beta1/alertingrule_types.go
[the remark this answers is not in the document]
periklis: Can you draft an alternative in a couple of lines of code here? I personally tolerate the duplication here, because APIs and their code should move version independently, e.g. one gets promoted to v1 before the other.

operator/internal/manifests/internal/gateway/options.go
periklis: Oh `CookieSecret` is the only option that gets read in `operator/internal/manifests/gateway.go` specifically when we translate from `TenantConfigs` to `TenantData`. TenantID was obsolete since we switched from UUIDs to names, i.e. `application`, `infrastructure`, `audit`.

operator/Makefile
periklis: Requires a patch into loki go.mod to resolve this issue.

operator/api/v1beta1/alertingrule_types.go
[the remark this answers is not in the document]
periklis: Per our online discussion, this will be resolved for the `internal/manifests` code path in a separate issue that addresses configmap sharding for alerting/recording rules.

operator/internal/manifests/rules_config.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+
+	"github.com/grafana/loki/operator/internal/manifests/internal/rules"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// RulesConfigMap returns a ConfigMap resource that contains
+// all loki alerting and recording rules as YAML data.
+func RulesConfigMap(opts *Options) (*corev1.ConfigMap, error) {
```
xperimental, line 13: Would it make sense to move this code to `github.com/grafana/loki/operator/internal/manifests/internal/rules` as well?

xperimental (commented):
Finally through all the files. :tada:

operator/internal/manifests/rules_config.go
periklis: No this belongs in the `manifests` package as it is the actual resource builder. We generally move only helpers and templates into the internal package.

operator/internal/manifests/ruler.go
```diff
@@ -0,0 +1,278 @@
...
+				Type: appsv1.RollingUpdateStatefulSetStrategyType,
+			},
+			RevisionHistoryLimit: pointer.Int32Ptr(10),
+			Replicas:             pointer.Int32Ptr(opts.Stack.Template.Ruler.Replicas),
+			Selector: &metav1.LabelSelector{
+				MatchLabels: labels.Merge(l, GossipLabels()),
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        fmt.Sprintf("loki-ruler-%s", opts.Name),
```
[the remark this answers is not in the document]
periklis: I am not hundert percent sure why we added this in the first place. It is though something like a pattern we seem to use across all components.

xperimental (approved):

PR #5997, opened 2022-04-22
Parallize label queries to both stores

simonswine:

This executes label queries against ingesters and store in parallel, rather than sequentially.

**Checklist**
~- [ ] Documentation added~
- [ ] Tests updated

### DIFF
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -378,23 +378,18 @@ func (q *SingleTenantQuerier) Label(ctx context.Context, req *logproto.LabelRequ
 	var storeValues []string
 	if !q.cfg.QueryIngesterOnly && storeQueryInterval != nil {
 		g.Go(func() error {
-			var err error
-			from := model.TimeFromUnixNano(storeQueryInterval.start.UnixNano())
-			through := model.TimeFromUnixNano(storeQueryInterval.end.UnixNano())
+			var (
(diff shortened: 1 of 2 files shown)


pkg/querier/querier.go
```diff
@@ -353,36 +352,48 @@ func (q *SingleTenantQuerier) Label(ctx context.Context, req *logproto.LabelRequ
...
+				}
+			} else {
+				storeValues, err = q.store.LabelNamesForMetricName(ctx, userID, from, through, "logs")
+				if err != nil {
+					return err
+				}
 			}
-		}
+
+			return nil
```
trevorwhitney, line 391: could we just return `err` here and remove the 2 nested `if err != nil` checks? Since they're in an if/else only one branch should execute, yeah?

trevorwhitney (approved):
LGTM, one small nit

cyriltovena (approved):
LGTM

pkg/querier/querier.go
simonswine: [965fec7](https://github.com/grafana/loki/pull/5997/commits/965fec75a83f7882eccb88b768abbb4f8ebd5674)

trevorwhitney (commented):
thanks for that change, let's get this merged!

jeschkies (approved):

PR #6003, opened 2022-04-24
Promote policyRule to v1

xaque208:

In Kubernetes 1.22, policyRule is promoted to v1 and out of v1beta.  Here we make the adjustment to the jsonnet call to reflect that change and allow users on 1.22 to make use of this jsonnet.  It looks like this was available at least back to 1.18.

https://jsonnet-libs.github.io/k8s-libsonnet/1.18/rbac/v1/policyRule/

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


simonswine (approved):
LGTM

Thanks that works for me

PR #6025, opened 2022-04-26
Disable calling new index-gateway client's API.

kavirajk:
We see some memory issue with new Index Gateway Client introduced in #5892, This PR just disables using new API by sticking with old API calls.

Fixes #<issue number>

**Checklist**

- [x] Tests updated


cyriltovena (approved):
LGTM

PR #6030, opened 2022-04-26
Add metrics to range mapper

ssncferreira:

* Refactor ShardingMetrics to MapperMetrics
* Add MapperMetrics to range mapper: each mapper contains the same set of metrics with a constant label that determines the mapper type, i.e., "shard" and "range" for shardMapper and rangeMapper, respectively.

Note: the metric names were not changed for backward compatibility ([see comment for more detail](https://github.com/grafana/loki/pull/6030/files#r858950266)).

For example, for the following configuration:
* `row_shards: 2`
* `split_queries_by_interval: 1m`

running the **instant query** `bytes_over_time({app="foo"}[5m])` and taking into account that the split by range middleware is executed prior to the sharding middleware (i.e., the downstream queries outputted by the range middleware are inputted to the sharding middleware), the metrics would result in the following values: 
* rangeMapper:
  * `loki_query_frontend_shards_total{mapper="range",type="metrics"} 5` - (counter) the query generates 5 downstream queries
  * `loki_query_frontend_sharding_parsed_queries_total{mapper="range",type="success"} 1` - (counter) the query is successfully parsed
  * `loki_query_frontend_shard_factor_sum{mapper="range"} 5` - (histogram) the query request generates 5 downstream range queries which are passed downwards to the shard mapper
* shardMapper:
  * `loki_query_frontend_shards_total{mapper="shard",type="metrics"} 10` - (counter) each of the 5 downstream queries generated by the split by range middleware generates 2 more downstream queries, one for each shard
  * `loki_query_frontend_sharding_parsed_queries_total{mapper="shard",type="success"} 5` - (counter) all the 5 downstream queries generated by the split by range middleware are successfully parsed
  * `loki_query_frontend_shard_factor_sum{mapper="shard"} 10` - (histogram) the query request generates 5 downstream shard queries

NA

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/logql/mapper_metrics.go
+++ b/pkg/logql/mapper_metrics.go
@@ -25,31 +25,27 @@ type MapperMetrics struct {
 	DownstreamFactor  prometheus.Histogram   // per request downstream factor
 }
 
-func NewMapperMetrics(registerer prometheus.Registerer) *MapperMetrics {
+func newMapperMetrics(registerer prometheus.Registerer, mapper string) *MapperMetrics {
 	return &MapperMetrics{
 		DownstreamQueries: promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
-			Namespace: "loki",
-			Name:      "query_frontend_shards_total",
-			Help:      "Number of downstream queries by expression type",
+			Namespace:   "loki",
+			Name:        "query_frontend_shards_total",
+			Help:        "Number of downstream queries by expression type",
+			ConstLabels: prometheus.Labels{"mapper": mapper},
 		}, []string{"type"}),
 		ParsedQueries: promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
-			Namespace: "loki",
-			Name:      "query_frontend_sharding_parsed_queries_total",
-			Help:      "Number of parsed queries by evaluation type",
+			Namespace:   "loki",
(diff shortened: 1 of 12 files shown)


pkg/logql/mapper_metrics.go
```diff
@@ -0,0 +1,81 @@
...
+			Help:        "Number of parsed queries by evaluation type",
+			ConstLabels: prometheus.Labels{"mapper": mapper},
+		}, []string{"type"}),
+		DownstreamFactor: promauto.With(registerer).NewHistogram(prometheus.HistogramOpts{
+			Namespace:   "loki",
+			Name:        "query_frontend_shard_factor",
+			Help:        "Number of downstream queries per request",
+			Buckets:     prometheus.LinearBuckets(0, 16, 4),
+			ConstLabels: prometheus.Labels{"mapper": mapper},
+		}),
```
ssncferreira, line 48: Since these metric names are now shared amongst all mappers, the names should be updated to make them more general, i.e., replacing "shards" with "downstream". However, this would result in a **breaking change**. This is my preferred option, but also don't know the impact this might have on users.
An alternative would be to remove this generalization by keeping the shard and the range mapper (and future mappers) metrics separate from their own names. Wdyt?

Issue #3374, opened 2021-02-24
Promtail not following symlinks as expected


fchiorascu:
**Is your feature request related to a problem? Please describe.**
Possibility to collect (tail) logs from stdout/stderr, port?

The current static_config uses **__path__** (log files) https://grafana.com/docs/loki/latest/clients/promtail/configuration/#static_configs that for `symlinks ` is not working as expected.

**Using: PLG Stack**
Grafana: 7.4.0
Loki: 2.1.0
Promtail: 2.1.0

**Describe the solution you'd like**
- **Promtail** being able to tail logs from: path file logs - symlinks, port, stdout, stderr of a pod/container in RHEL OpenSHift Container > 3.9 from a namespace.

**Describe alternatives you've considered**
- At least being able to tail path file logs - symlinks.

**Additional context**
- https://docs.fluentbit.io/manual/pipeline/filters/standard-output
- https://docs.fluentbit.io/manual/pipeline/inputs/tcp

bramevo:
+1 For this request. I need to be able parse compressed log files (gzip) and this would be possible using the above I assume.
Now I need a dedicated container just for decompressing the logfiles before using the promtail container to parse them.

proffalken:
Isn't this what Fluentd/fluent-bit are for?

Promtail seems like a good starting place to slurp logs, but if you want to do processing on the data as it comes in then there are plenty of existing tools for that and we should be careful not to re-invent the wheel.

Admittedly neither of them are able to parse compressed files, but it feels we should add the features to something like that rather than adding it to promtail?

fchiorascu:
@proffalken thank you for your point of view.

I understand that is not a need of reinventing the wheel or covering the same aspects with multiple options/similar tools, this is clear and no doubt there. However I don't need promtail for log processing or to use in complex ecosystems, it was a good fit for us due to the fact that is lightweight and we could integrate in a small RHEL OpenShift Container Platform (namespace) to collect logs from 1 application (backend/frontend) but he is not behaving well with symlinks and doesn't have the maturity or alternatives to file log __path__ such as stdout, stderr, port. Why to use fluentbit/fluentd instead of using entire PLG Stack, reducing the complexity and being sure that the pieces integrates seamlessly.

proffalken:
@fchiorascu that's fair enough, objection removed ;)

proffalken:
Commenting to keep this open, I think it's a really useful addition to Loki/Promtail.

proffalken:
[...]

PR #6034, opened 2022-04-26
Revendor `grafana/tail` to fix promtail symlink behavior

DylanGuedes:
Revendors `grafana/tail` to introduce a recent fix to how we tail symlinks. With the current implementation, promtail is only able to work with symlinks if they're in the same folder as the promtail binary; if they're inside a nested folder, tail is unable to find the files.

- Fixes https://github.com/grafana/loki/issues/3374

- The fix is based on moving to [filepath/EvalSymlinks method](https://pkg.go.dev/path/filepath#EvalSymlinks) instead of `os.Readlink`
- This was already tested [here](https://github.com/grafana/tail/issues/11)

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -284,7 +284,7 @@ require (
 // Upgrade to run with gRPC 1.3.0 and above.
 replace github.com/sercand/kuberesolver => github.com/sercand/kuberesolver v2.4.0+incompatible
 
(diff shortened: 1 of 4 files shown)


slim-bean (approved):
LGTM

slim-bean:
@DylanGuedes sorry I didn't think about this until after clicking merge. Can you fire up another PR and put an entry in the change log for this in the promtail section?

PR #6044, opened 2022-04-28
Improve batching in boltdb index

cyriltovena:
This removes the timer and sync code, and replace it with a batch cursor that iterate when required.

Also improve the memory usage.

```
name       old time/op    new time/op    delta
_Query-16    3.16µs ± 4%    1.01µs ± 5%  -67.96%  (p=0.008 n=5+5)

name       old alloc/op   new alloc/op   delta
_Query-16    1.08kB ± 1%    0.66kB ± 0%  -38.95%  (p=0.008 n=5+5)

name       old allocs/op  new allocs/op  delta
_Query-16      15.0 ± 0%      10.0 ± 0%  -33.33%  (p=0.008 n=5+5)
```

### DIFF
--- a/pkg/storage/chunk/client/local/boltdb_index_client.go
+++ b/pkg/storage/chunk/client/local/boltdb_index_client.go
@@ -258,69 +258,103 @@ func (b *BoltIndexClient) QueryDB(ctx context.Context, db *bbolt.DB, bucketName 
 }
 
 func (b *BoltIndexClient) QueryWithCursor(_ context.Context, c *bbolt.Cursor, query index.Query, callback index.QueryPagesCallback) error {
-	var start []byte
-	if len(query.RangeValuePrefix) > 0 {
-		start = []byte(query.HashValue + separator + string(query.RangeValuePrefix))
-	} else if len(query.RangeValueStart) > 0 {
-		start = []byte(query.HashValue + separator + string(query.RangeValueStart))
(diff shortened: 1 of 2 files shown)


pkg/storage/chunk/client/local/boltdb_index_client.go
```diff
@@ -258,69 +258,103 @@ func (b *BoltIndexClient) QueryDB(ctx context.Context, db *bbolt.DB, bucketName
...
-		start = []byte(query.HashValue + separator + string(query.RangeValuePrefix))
-	} else if len(query.RangeValueStart) > 0 {
-		start = []byte(query.HashValue + separator + string(query.RangeValueStart))
-	} else {
-		start = []byte(query.HashValue + separator)
-	}
+	batch := batchPool.Get().(*cursorBatch)
+	batch.reset()
+	batch.cursor = c
+	batch.query = &query
```
sandeepsukhani, line 264: Should we pass these to `batch.reset`? Would make it a bit clearer I feel.

sandeepsukhani (approved):
Nice one! Just added a minor non blocking suggestion.

pkg/storage/chunk/client/local/boltdb_index_client.go
cyriltovena: yep

PR #6045, opened 2022-04-28
Improve series store index queries

cyriltovena:
This PR improve performance of index queries of the series store by:

- Removing useless debug log in the hot path.
- Re-use slice of index entries to reduce allocations
- Use yolostring from index entries since those byte slice will never changed.


sandeepsukhani (approved):
LGTM

kavirajk (approved):
LGTM :+1:

PR #6047, opened 2022-04-28
improve index list cache in boltdb-shipper to avoid refreshing cache at query time

sandeepsukhani:
To reduce the number of list calls we make to the object store when using `boltdb-shipper`, we have a cache in place. The cache stays valid for 1 min and is refreshed when a list call is made and it finds that cache needs to be rebuilt. The list call could take a couple of seconds to finish in a large cluster which can add up to the query latency when downloading index at query time.

This PR improves it by doing the cache auto refreshing in the background and the list call returning whatever we have in cache without worrying about its validity. The cache auto refreshing is only necessary for background operations like sync and compaction. For query time index download, it is okay to use the listing built by last sync operation. So the caching layer now exposes 2 new methods for enabling/disabling cache auto sync by background operations:
```
	EnableCacheAutoRefresh()
	DisableCacheAutoRefresh()
```
The implementation takes care of multiple simultaneous requests that can come in to enable/disable cache auto refresh when running multiple Loki services in a single process.

I also realized that caching can be also used with `filesystem` object store due to it now supporting flat object listing so I have enabled it for that as well. I have also added `RefreshCache` to let the tests refresh the cache without which they would fail to see updates done in the storage since they use `filesystem` for storing index and it now supports caching of index list.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/shipper/downloads/table_test.go
+++ b/pkg/storage/stores/shipper/downloads/table_test.go
@@ -396,6 +396,9 @@ func TestTable_Sync(t *testing.T) {
 	require.NoError(t, os.Remove(filepath.Join(tablePathInStorage, deleteDB)))
 	testutil.AddRecordsToDB(t, filepath.Join(tablePathInStorage, newDB), boltdbClient, 20, 10, nil)
 
+	// refresh the object list cache to get latest list of files
+	table.storageClient.RefreshCache()
+
 	// sync the table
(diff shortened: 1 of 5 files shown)


MasslessParticle:
This looks good! Who's job is it to call?

```
	EnableCacheAutoRefresh()
	DisableCacheAutoRefresh()
```

PR #6049, opened 2022-04-28
TSDB shipper + WAL

owen-d:
This PR does a few things and is very large. I'm opening as a draft to get some initial feedback. It:
* Adds a TSDB WAL for persistence on the ingesters
* Adds a TSDB Head, which wraps a WAL and makes written index entries immediately available for querying in memory
* Adds a TSDB Store implementation
* Integrates TSDB with the generic indexshipper
* Builds a TSDB Manager, which combines the `(TSDB Head, TSDB  WAL, indexshipper)`
* Some refactoring

This is by no means complete, but it's getting rather large and this is a relatively good place to PR so we can have multiple authors focus on different parts.

On disk, it looks like this when building/shipping the TSDB multitenant indices (first phase):
```
/tmp/loki/
├── data
│   ├── fake
│   │   └── ce802667485e7a77
│   │       ├── MTgwNzFjM2Y1YTY6MTgwNzFjNDQzZmY6ZGI1NzRh
│   │       ├── MTgwNzFjMzc1NjQ6MTgwNzFjM2NlN2U6YTRlYThkNzk=
│   │       ├── MTgwNzFjNDZiMjI6MTgwNzFjNGI5NmQ6NmFmNjQ1YWY=
│   │       ├── MTgwNzFjNGUwOTQ6MTgwNzFjNTJlZTY6YTE5OGJiZmM=
│   │       ├── MTgwNzFjNTU1MWI6MTgwNzFjNWEzN2M6YWZkY2ZiOWU=
│   │       └── MTgwNzFjNWNhYWI6MTgwNzFjNjE5MGM6Yjk2Mjc1Nzg=
│   ├── index
│   │   └── 19110
│   │       └── v1
│   │           └── multitenant
│   │               ├── 1651176108-.tsdb.gz
│   │               ├── 1651176138-.tsdb.gz
│   │               └── 1651176168-.tsdb.gz
│   └── loki_cluster_seed.json
├── rules
├── tsdb-shipper-active
│   └── v1
│       ├── multitenant
│       │   └── 19110
│       │       ├── 1651176108-.tsdb
│       │       ├── 1651176138-.tsdb
│       │       ├── 1651176168-.tsdb
│       │       ├── 1651176198-.tsdb
│       │       └── 1651176228-.tsdb
│       ├── per_tenant
│       ├── scratch
│       └── wal
│           └── 1651176258
│               └── 00000000
├── tsdb-shipper-cache
└── wal
    ├── 00000000
    ├── 00000001
    ├── 00000002
    └── checkpoint.000001.tmp
        └── 00000000

19 directories, 20 files

```

cc @sandeepsukhani @cyriltovena @slim-bean 


pkg/loki/modules.go
```diff
@@ -412,22 +417,54 @@ func (t *Loki) initStore() (_ services.Service, err error) {
...
 			t.Cfg.Ingester.RetainPeriod = t.Cfg.StorageConfig.IndexCacheValidity + 1*time.Minute
-			t.Cfg.StorageConfig.BoltDBShipperConfig.IngesterDBRetainPeriod = boltdbShipperQuerierIndexUpdateDelay(t.Cfg) + 2*time.Minute
+
+			// We do not want ingester to unnecessarily keep downloading files
+			t.Cfg.StorageConfig.BoltDBShipperConfig.Mode = shipper.ModeWriteOnly
+			t.Cfg.StorageConfig.BoltDBShipperConfig.IngesterName = t.Cfg.Ingester.LifecyclerConfig.ID
+			t.Cfg.StorageConfig.BoltDBShipperConfig.IngesterDBRetainPeriod = shipperQuerierIndexUpdateDelay(t.Cfg.StorageConfig.IndexCacheValidity, t.Cfg.StorageConfig.BoltDBShipperConfig.ResyncInterval) + 2*time.Minute
+
+			t.Cfg.StorageConfig.TSDBShipperConfig.Mode = indexshipper.ModeWriteOnly
+			t.Cfg.StorageConfig.TSDBShipperConfig.IngesterName = t.Cfg.Ingester.LifecyclerConfig.ID
```
sandeepsukhani, line 427: The reason ingester name is not being set with target `all` is because `t.Cfg.isModuleEnabled` would only look at explicitly specified targets in `-target` flag. We will have to move it to a separate block with or without the check using `t.Cfg.isModuleActive("ingester")` check which actually checks whether the module is active or not .

sandeepsukhani (approved):
One thing I noticed is that the new shipper code is not adding ingester names to the files, having just the timestamp. This can cause multiple ingesters uploading files at exactly the same time to override each other's files. I will take care of this issue in a follow-up PR that I will work on to make the `boltdb-shipper` client use the new shipper code.

PR #6056, opened 2022-04-29
Add comments to the Func definition

MichelHollands:
Explain the filter func better. This was mentioned as a review comment in https://github.com/grafana/loki/pull/5879.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/util/filter/filter_function.go
+++ b/pkg/util/filter/filter_function.go
@@ -1,3 +1,5 @@
 package filter
 
+// Func is a function type that is the signature of a filter function used in log deletion.
+// This is used as a parameter in the Rebound function.
 type Func func(string) bool


sandeepsukhani (approved):
Would it be possible to move `Func` to either of `deletion` or `chunkenc` package i.e where it is used?

pkg/util/filter/filter_function.go
```diff
@@ -1,3 +1,5 @@
 package filter
 
+// Func is a function type that is the signature of a filter function used in log deletion.
+// This is used as a parameter in the Rebound function.
 type Func func(string) bool
```
DylanGuedes, line 5: WDYT of naming it `Predicate`? I think it is how Apache Spark and Java name a similar thing.

PR #6058, opened 2022-04-29
ignores reporting cancellations as errors in the usage-reporter module

owen-d:
After noticing some unexpected module failures, I realized the usage reporter is returning context cancellation as errors. This doesn't work well with the service code which [cancels](https://github.com/grafana/loki/blob/main/vendor/github.com/grafana/dskit/services/basic_service.go#L242-L246) the context itself. I've adjusted the code to ignore cancellations for this reason.

### DIFF
--- a/pkg/usagestats/reporter.go
+++ b/pkg/usagestats/reporter.go
@@ -3,6 +3,7 @@ package usagestats
 import (
 	"bytes"
 	"context"
+	"errors"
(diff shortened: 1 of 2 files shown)


trevorwhitney (approved):
seems reasonable to me, LGTM!

cyriltovena (approved):
LGTM

PR #6059, opened 2022-04-29
Add tailing metrics

### LINKED ISSUE #6027: Additional metrics for monitoring tailer component

At the moment there is no good way to monitor log tailing connections.

Add new metrics for:
* active tail connection
* amount of active streams tailed
* amount of data sent to clients

Some of the metrics could be derived from gRPC connection logs.

MasslessParticle:
Add tail metrics.

Fixes #6027

### DIFF
--- a/pkg/querier/tail.go
+++ b/pkg/querier/tail.go
@@ -41,6 +41,10 @@ type Tailer struct {
 	currEntry  logproto.Entry
 	currLabels string
 
+	// keep track of the streams for metrics about active streams
+	seenStreams    map[uint64]struct{}
+	seenStreamsMtx sync.Mutex
+
 	tailDisconnectedIngesters func([]string) (map[string]logproto.Querier_TailClient, error)
 
 	querierTailClients    map[string]logproto.Querier_TailClient // addr -> grpc clients for tailing logs from ingesters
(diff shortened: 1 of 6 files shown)


pkg/querier/tail.go
```diff
@@ -106,6 +114,7 @@ func (t *Tailer) loop() {
 				continue
 			}
 
+			entriesSize += len(t.currEntry.Line)
```
MasslessParticle, line 117: This isn't necessarily the number of bytes, but it is consistent with how we calculate similar metrics in other components.

trevorwhitney (approved):
LGTM!

pkg/querier/tail.go
MasslessParticle: You're right: Golang `len` gives back bytes rather characters. I can't believe I didn't know this.

I wonder how many bugs I've caused with it?

trevorwhitney: haha, go was clearly designed for dummies like me who only use English characters and still think ASCII is the only encoding out there.

pkg/querier/tail.go
```diff
@@ -264,25 +279,48 @@ func (t *Tailer) getCloseErrorChan() <-chan error {
 	return t.closeErrChan
 }
 
+func (t *Tailer) recordStream(id uint64) {
+	t.seenStreamsMtx.Lock()
+	defer t.seenStreamsMtx.Unlock()
+
+	if _, ok := t.seenStreams[id]; ok {
```
cyriltovena, line 286: If you're only writing once there's a lot of chances that `sync.Map` will perform better. Not sure this is important in this case but at least I shared the options for later.

cyriltovena (approved):
LGTM

PR #6063, opened 2022-05-01
Loki: Change live tailing to only allow mutating the log line not the number of streams.

slim-bean:

Stops live tailing from splitting a stream into more streams when using parsers while still maintaining the ability to mutate the log line using parsers and `line_format`

The current implementation will allow creating many new streams from a tailed stream when using parsers, but this multiplicative affect can make it difficult to then tail the stream because it creates a lot more overhead in the amount of data pushed down the websocket.

For the purposes of live tailing which is meant primarily for human consumption, returning the parsed and extracted labels and manipulating the streams isn't helpful for the end user visualization and instead makes the process less efficient and more likely to struggle to keep up with higher volume streams.

Update: Also included in this PR is a change to no longer reuse the same pipeline for parsing every push request, the pipeline has a label cache built into it which, when reused, would lead to unbounded memory growth when the log lines being processed had extremely high cardinality data being parsed.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/ingester/tailer.go
+++ b/pkg/ingester/tailer.go
@@ -121,53 +121,31 @@ func (t *tailer) send(stream logproto.Stream, lbs labels.Labels) {
 		return
 	}
 
-	streams := t.processStream(stream, lbs)
-	if len(streams) == 0 {
-		return
-	}
-	for _, s := range streams {
-		select {
-		case t.sendChan <- s:
-		default:
-			t.dropStream(*s)
-		}
+	t.processStream(&stream, lbs)
+	select {
+	case t.sendChan <- &stream:
+	default:
+		t.dropStream(stream)
 	}
 }
 
-func (t *tailer) processStream(stream logproto.Stream, lbs labels.Labels) []*logproto.Stream {
+func (t *tailer) processStream(stream *logproto.Stream, lbs labels.Labels) {
 	// Optimization: skip filtering entirely, if no filter is set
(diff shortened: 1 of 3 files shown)


pkg/ingester/tailer.go
```diff
@@ -121,53 +121,31 @@ func (t *tailer) send(stream logproto.Stream, lbs labels.Labels) {
...
-		}
-		stream.Entries = append(stream.Entries, logproto.Entry{
-			Timestamp: e.Timestamp,
-			Line:      newLine,
-		})
-	}
-	streamsResult := make([]*logproto.Stream, 0, len(streams))
-	for _, stream := range streams {
-		streamsResult = append(streamsResult, stream)
+		stream.Entries[i].Line = newLine
```
owen-d, line 147: I don't think we can mutate the line here because the stream is passed to multiple tailers:

https://github.com/grafana/loki/blob/main/pkg/ingester/stream.go#L311-L317

owen-d (commented):
Ooh, nice find. How'd you catch this?

owen-d (approved):

pkg/ingester/tailer.go
slim-bean: Good catch! fixed.

sandeepsukhani (approved):
Nice one! LGTM!

PR #6064, opened 2022-05-01
renames skip -> ok in ProcessString

owen-d:
This confuses me every now and then, so I fixed the method var name and the last 2 references that used it as the name `skip`. Looking at how it's _used_, it should be named `ok` or `keep` or `passes`.


slim-bean:
What about 'matched' instead of 'ok' ?

cyriltovena:
I hear you but not sure ok does a better job no ? The point is that the line is filtered out.

owen-d:
I was being lazy trying to avoid replacing all the other references which already use `ok` 😆 . Pushed an update to fix them 👍

cyriltovena (approved):
LGTM

PR #6071, opened 2022-05-02
allow more time for boltdb-shipper index syncs to finish

sandeepsukhani:
With `boltdb-shipper`, we have some timers configured to manage the index across various services. This PR looks at the time we allow sync operations to finish in the worst-case scenario.
With the current timers set, in the worst-case scenario, we should be finishing the index sync operation in up to 2 mins and if we fail, users might see gaps in their query response. This PR tunes some timers to allow syncs to have more time to run, specifically allowing sync to get the updates in up to 3 attempts. This change impacts the following timers with default values:

* retains index files with ingester for 20m instead of 12m after uploading.
* sets GetChunkIDs lookback to 2h41m instead of 2h28m.

The impact of this is ingesters getting more queries from queries for `GetChunkIDs` since we are increasing the lookback


cyriltovena (approved):
LGTM

PR #5971, opened 2022-04-20
`metrics.go` support for metadata queries(labels and series)

kavirajk:
**What this PR does / why we need it**:
We use `metrics.go`  (a convention) to record some useful statistics about each of the `instance` and `range` queries. 

So that we can query something like
`{foo="bar"}|= "metrics.go" |logfmt|latency=slow `

Where `latency` comes from statistics recorded in `pkg/logql/metric.go`.

This PR add support those experience for series and labels API as well.

Examples:
1. Labels API (names)
```
curl -G -s  "http://127.0.0.1:3100/loki/api/v1/label"
```

Generates following log lines via `metrics.go`
```
level=info ts=2022-05-03T19:51:37.417252626Z caller=metrics.go:170 component=frontend org_id=145265 traceID=5bdfb2e229890ea2 latency=fast query_type=labels length=1h0m0s duration=65.140762ms status=200 label= throughput=0B total_bytes=0B total_entries=14
```

2. Labels API (values)
```
curl -G -s  "http://127.0.0.1:3100/loki/api/v1/label/container/values"
```

Generates following log lines via `metrics.go`
```
level=info ts=2022-05-03T19:50:07.710590524Z caller=metrics.go:170 component=frontend org_id=145265 traceID=11349be1207cbdba latency=fast query_type=labels length=1h0m0s duration=11.239903ms status=200 label=container throughput=0B total_bytes=0B total_entries=19
```

3. Series API
```
curl -G -s  "http://127.0.0.1:3100/loki/api/v1/series" --data-urlencode 'match[]={app="loki"}'
```

Generates following log lines via `metrics.go`
```
level=info ts=2022-05-03T19:51:40.272480868Z caller=metrics.go:215 component=frontend org_id=145265 traceID=396c8cdce34bb53e latency=fast query_type=series length=1h0m0s duration=85.968193ms status=200 match="{app=\"loki\"}" throughput=0B total_bytes=0B total_entries=249
```

**Which issue(s) this PR fixes**:
Fixes https://github.com/grafana/loki/issues/4503

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] Add an entry in the `CHANGELOG.md` about the changes.

**Future Work**
- [ ] Get rid of `latency=slow|fast` label from metric.go entries.
- [ ] deprecate and remove `returned_lines` when `query_type=filtered`. Because now we have `total_entries` common for all the `query_type`.

PR #6072, opened 2022-05-02
Fix(query-frontend): `/label/<name>/values` endpoint to use right mid…

kavirajk:

Previously, only `/label/` and `/labels` endpoint were using correct tripperware middleware
on the query frontend(e.g: stats collector middleware)

This PR fixes it for `/label/<name>/values` endpoint as well.

This is useful to support `metrics.go` stats middleware for `label/{name}/values` endpoint as well.

Fixes #<issue number>

Related to #5971 

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/querier/queryrange/roundtrip.go
+++ b/pkg/querier/queryrange/roundtrip.go
@@ -232,7 +232,7 @@ func getOperation(path string) string {
 		return QueryRangeOp
 	case strings.HasSuffix(path, "/series"):
 		return SeriesOp
-	case strings.HasSuffix(path, "/labels") || strings.HasSuffix(path, "/label"):
+	case strings.HasSuffix(path, "/labels") || strings.HasSuffix(path, "/label") || strings.HasSuffix(path, "/values"):
 		return LabelNamesOp
 	case strings.HasSuffix(path, "/v1/query"):
 		return InstantQueryOp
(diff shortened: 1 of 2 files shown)


pkg/querier/queryrange/roundtrip.go
```diff
@@ -232,7 +232,7 @@ func getOperation(path string) string {
 		return QueryRangeOp
 	case strings.HasSuffix(path, "/series"):
 		return SeriesOp
-	case strings.HasSuffix(path, "/labels") || strings.HasSuffix(path, "/label"):
+	case strings.HasSuffix(path, "/labels") || strings.HasSuffix(path, "/label") || strings.HasSuffix(path, "/values"):
```
dannykopping, line 235: This function feels a bit flimsy, and could introduce some _interesting_ bugs in the future.
We should refactor this at some point

dannykopping: Thanks for adding a test!

dannykopping (approved):
LGTM

pkg/querier/queryrange/roundtrip.go
kavirajk, line 235: Q: Does anyone know why middeware chain was disabled for `/labels/<name>/values` endpoint on the query frontend?

We in fact have a test to lock that behaviour.
https://github.com/grafana/loki/blob/main/pkg/querier/queryrange/roundtrip_test.go#L390-L400

I wanted this support so that middlewares like `metrics.go` stats collector can work on this endpoints too. (#5971)

Any thoughts?

kavirajk: If no protest, I can remove that test!

PR #6084, opened 2022-05-03
Fix(query-frontend): LokiCodec encode/decode works correctly with label values endpoint

kavirajk:

When encode/decode `http.Request` into `querybase.Request` we re-create request.Path.

Before this PR, both `/labels/` and `/labels/{name}/values` endpoint was decoding into same URL path `/labels/` before
forwarding to the querier. So it always returns label names as response.

This bug became visible after enabling all the middleware for `/labels/{name}/values` endpoint as well via #6072

Fixes # NA. 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


salvacorts (approved):
LGTM

cyriltovena (approved):

PR #6089, opened 2022-05-03
Docs: fix pic location in `sources/fundamentals/architecture/deployme…

n4mine:


kavirajk (approved):
LGTM.

Although the current tip of the main seems to render the image just fine.
https://grafana.com/docs/loki/next/fundamentals/architecture/deployment-modes/

Better to add broken links as reference to PR(s) like this; makes review much easy :)

PR #6090, opened 2022-05-03
Fix Labels when no logs exist

### LINKED ISSUE #6011: Label named "__name__" exists in empty Loki database
**Describe the bug**
when running an empty Loki database via Docker, when i query the `/loki/api/v1/labels` HTTP endpoint, it returns a result saying there is a label-name,  `"__name__"`, even though this is an empty Loki database.

**To Reproduce**
1. `docker run -p 3100:3100 grafana/loki:2.5.0`
2. `curl 'http://127.0.0.1:3100/loki/api/v1/labels'`
3. the returned JSON says `{"status":"success","data":["__name__"]}` , meaning there is 1 label already, named `__name__`

**Expected behavior**
the result from Loki should say there are no label-names in the database.

**Environment:**
 - Infrastructure: docker on my local computer
 - Deployment tool: none

MasslessParticle:
The `__name__` label was still being added to results even when no series were selected for the time period in a call to the `Labels` endpoint. This also meant that `__name__` was being included when Loki contained no logs.

In addition, this PR changes the ingester to return a `__name__` label when it is queried for labels. Before, the querier was relying on the store to always return the `__name__` label

Fixes #6011


pkg/storage/stores/series/series_store_test.go
```diff
@@ -301,6 +301,14 @@ func TestChunkStore_LabelNamesForMetricName(t *testing.T) {
 						t.Fatalf("%s: future query should yield empty resultset ... actually got %v label names: %#v",
 							tc.metricName, len(labelNames3), labelNames3)
 					}
+
+					// Query asking for a non-existant metric name should return an empty resultset
```
kavirajk, line 305:
```suggestion
					// Query asking for a non-existent metric name should return an empty resultset
```

kavirajk (approved):
LGTM

owen-d (approved):
Nice find.

pkg/ingester/index/index.go
```diff
@@ -171,11 +171,20 @@ func (ii *InvertedIndex) LabelNames(shard *astmapper.ShardAnnotation) ([]string,
...
+
+	var (
+		shards     = ii.getShards(shard)
+		hasResults = false
+		results    = make([][]string, 0, len(shards))
+	)
 	for i := range shards {
 		shardResult := shards[i].labelNames()
 		results = append(results, shardResult)
+		hasResults = hasResults || len(shardResult) > 0
```
sandeepsukhani, line 183: Why can't we just do this check below instead of each iteration unless I am missing something?

pkg/ingester/ingester_test.go
```diff
@@ -588,7 +588,7 @@ func Test_InMemoryLabels(t *testing.T) {
 
 	res, err = i.Label(ctx, &logproto.LabelRequest{})
 	require.NoError(t, err)
-	require.Equal(t, []string{"bar", "foo"}, res.Values)
+	require.Equal(t, []string{"__name__", "bar", "foo"}, res.Values)
```
chaudum, line 591: Do I read that right, that there were cases that label queries on the ingester did _not_ contain the `__name__` label, even though labels were present?
If so, I think it would be good to also add this in the PR description.

MasslessParticle: There were. It was relying on the fact that a `__name__` label always came from the store: whether there were series or not. This was actually found as a result of the new integration tests.

I'll add it to the PR description

MasslessParticle:
Closing this in favor of removing `__name__` from label results altogether

PR #6093, opened 2022-05-04
Revert "Docs: fix pic location in `sources/fundamentals/architecture/deployme…"

n4mine:
Reverts grafana/loki#6089

i think it's break https://grafana.com/docs/loki/next/fundamentals/architecture/deployment-modes/ pic

and i see the desc at `https://github.com/grafana/loki/tree/main/docs`

> `The files in this directory are used to generate it but consequently the links don't work in Github.`

so reverts that pr


KMiller-Grafana (approved):
The hosted docs need to be right over looking at docs on GitHub. That's why this revert needs to happen.  Thanks.

Issue #5784, opened 2022-04-06
Promtail to support validating JSON before sending off to Loki


wengelbrecht-grafana:
**Is your feature request related to a problem? Please describe.**
When logs are generated in JSON format and sent of to Loki, it happens sometimes that logs are not in valid JSON format. For example this could be due to misconfigured applications typically; init and "sidecar" containers; bugs, misconfiguration of the agent, or any reason. 

This means that logs are ingested, potentially at cost and immediately ignored again when querying them. 

It would be ideal if in promtail drop stage, there could be the option to validate if a logline is in valid JSON format, and if not, just drop it there and never send it off. 

Although there is an existing Promtail JSON stage, it assumes the logline is already in valid JSON format and does not give you the option to drop invalid/unparsable JSON log lines. 

**Describe the solution you'd like**

In promtail drop stage, to have the ability to specify that the logline should pass the "valid JSON parsable check", and if it fails, then drop the log line. 

**Describe alternatives you've considered**
There isn't really an alternative at the moment. You are sort of forced to ingest invalid JSON log lines and then immediately discard them at query time. However this means you're paying costs to ingest and store. 

**Additional context**
None

mrmartan:
It would also be useful to be able to emit the count of invalid logs as a metric.

09jvilla:
Can some of the folks who thumbs up'd this issue help give some examples of why log lines are generated in invalid JSON format?

mrmartan:
Sometimes they are not json at all. Imagine scraping from a pod with multiple containers, some you have under control (those can be misconfigured), other you don't (3rd party sidecars or init containers that even dont have the ability to write logs in json, e.g. injected service mesh overlay). 

Managing the log shpping middlewares in a larger organizations you run into these situations and this could generate a lot of unusable data before rectified, if it can be.

09jvilla:
And you don't find yourself using any of those non-json logs @mrmartan ? You wouldn't be able to use the json parser to pull out information for these lines, but you could still manipulate them with the regex or the pattern parser.

mrmartan:
I would though I don't know I can. I am getting JSON perse error entries in Loki/Grafana. Which means I don't have the labels I want attached. Is it possible to setup branching pipelines in primtail?

That's a bit off-topic. The pipeline configuration expects JSON log lines and isn't getting it and then ships unprocessed log lines to storage. It is valid to not want them shipped, isn't it? It is also necessary to be aware that this is happening.

[...]

PR #6099, opened 2022-05-04
[Promtail] drop lines with malformed json

cstyan:
Fixes https://github.com/grafana/loki/issues/5784

In this PR I've added a new config field to the JSON pipeline stage which allows for dropping of log lines that cannot properly be unmarshaled to JSON objects (malformed json or just not json at all). In the future we could extend this to other log formats with new pipeline stages such as logfmt. 

The commits are relatively standalone, first I refactored the json pipeline stage to match the new(ish) pipeline stage interface with a `Run` function, then implemented the new `drop malformed json` option.


clients/pkg/logentry/stages/json.go
```diff
@@ -92,8 +93,28 @@ func parseJSONConfig(config interface{}) (*JSONConfig, error) {
...
+	}()
+	return out
+}
+
 // Process implements Stage
-func (j *jsonStage) Process(labels model.LabelSet, extracted map[string]interface{}, t *time.Time, entry *string) {
+func (j *jsonStage) Process(_ model.LabelSet, extracted map[string]interface{}, _ *time.Time, entry *string) {
+	//nolint:errcheck
+	j.process(extracted, entry)
+}
```
chaudum, line 115: The function `Process()` is obsolete, because `jsonStage` now implements the `Stage` interface.

clients/pkg/logentry/stages/json.go
```diff
@@ -130,7 +151,7 @@ func (j *jsonStage) Process(labels model.LabelSet, extracted map[string]interfac
 		if Debug {
 			level.Debug(j.logger).Log("msg", "failed to unmarshal log line", "err", err)
 		}
-		return
+		return fmt.Errorf("malformed json")
```
chaudum, line 154: Can we replace this with a variable to avoid allocations?

clients/pkg/logentry/stages/json_test.go
```diff
@@ -365,3 +365,33 @@ func TestJSONParser_Parse(t *testing.T) {
...
+func TestValidateJSONDrop(t *testing.T) {
+	labels := map[string]string{"foo": "bar"}
+	matchConfig := JSONConfig{
+		DropMalformed: true,
+		Expressions:   map[string]string{"page": "page"},
+	}
+	s, err := newJSONStage(util_log.Logger, matchConfig)
+	if err != nil {
+		t.Errorf("withMatcher() error = %v", err)
+		return
```
chaudum, line 378:
```suggestion
		t.Fatalf("withMatcher() error = %v", err)
```

owen-d (approved):
Approving, but please address @chaudum's feedback. Nice work.

vlad-diachenko (approved):
Looks great! also, I think that @chaudum 's suggestion to use testify lib for assertions will simplify the test for reading...

clients/pkg/logentry/stages/json_test.go
```diff
@@ -365,3 +365,33 @@ func TestJSONParser_Parse(t *testing.T) {
...
+	}
+	if s != nil {
```
vlad-diachenko, line 380: I would prefer to fail the test if `s` is nil. the current implementation passes the test if `s` is nil and does not run any assertions.

cstyan:
Thanks for the reviews guys! I fixed some of the comments in one commit and included a second that refactors all of the `json_test.go` tests to use testify `assert` consistently.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Main
 
+* [6099](https://github.com/grafana/loki/pull/6099/files) **cstyan**: drop lines with malformed json in promtail json pipeline stage
```
KMiller-Grafana, line 3:
```suggestion
* [6099](https://github.com/grafana/loki/pull/6099/files) **cstyan**: Drop lines with malformed JSON in Promtail JSON pipeline stage
```

docs/sources/clients/promtail/stages/json.md
```diff
@@ -22,6 +22,10 @@ json:
 
   # Name from extracted data to parse. If empty, uses the log message.
   [source: <string>]
+
+  # If true then any lines that cannot be successfully parsed as valid JSON objects
+  # will be dropped instead of being sent to Loki.
+  drop_malformed: <bool>
```
KMiller-Grafana, line 28: Is there a default value?   I think that the syntax here might be incomplete.  I think that there should be "[]" surrounding the configuration variable if this variable is optional.

If there is a default value and this configuration variable is optional, then the definition ought to be revised to something like

```suggestion
  # When true, any lines that cannot be successfully parsed as valid JSON objects
  # will be dropped instead of being sent to Loki (the default).
  [drop_malformed: <bool> | default = false]
```

KMiller-Grafana (approved):
I've left comments/suggestions, but am approving this PR such that you don't need a re-review from me.

chaudum (approved):

PR #6101, opened 2022-05-05
[corner case] putChunk: use the HeadObject s3 api to detect whether the chunk is exist when retry.

liguozhong:
ingester flush.go putChunk: When the s3 write queue is slow, the client will time out, but in fact, the final s3 write queue will succeed. 
Therefore, this PR attempts to record a "failed" flag for the chunk. When retrying next time, use the HeadObject s3 api to lightly detect whether the chunk is successfully processed in the queue. Since HeadObject and putObject are different processing loads, when putObject is busy, HeadObject can also respond normally, thus solving this corner case.

Fixes #<issue number>
image

This is a 5-hour failure of our online loki cluster from last week.
API_HeadObject： https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html
HeadObject prometheus metrics snapshot.
image

why hard code "NotFount",http msg snapshot. error msg = "Not Found"
image

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/ingester/flush.go
+++ b/pkg/ingester/flush.go
@@ -378,8 +378,11 @@ func (i *Ingester) flushChunks(ctx context.Context, fp model.Fingerprint, labelP
 		if err := i.encodeChunk(ctx, &ch, c); err != nil {
 			return err
 		}
-
+		if c.failed {
+			ch.PreCheck = true
+		}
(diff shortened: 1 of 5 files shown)


liguozhong:
image

Is this bug introduced by this PR? Request for help.

PR #6105, opened 2022-05-05
Promtail: Add metrics for journal target

### LINKED ISSUE #5554: Promtail missing internal telemetry about systemd-journal target
We have a promtail configured with two scrape configs: `consulagent_sd_configs` & `systemd-journal`
Inspecting the output from the metrics at `/metrics`, it seems that Promtail only exposes data relevant to file-based log tailing. 

There is no data about metrics that were collected via the `systemd-journal` target at all. 
I'm not even sure if the total metrics such as 
- `promtail_targets_active_total`, 
- `promtail_dropped_bytes_total` 
- `promtail_dropped_entries_total`
- `promtail_encoded_bytes_total`
Include data from `systemd-journal` or not.

This makes it difficult to monitor log volume and errors from promtail.

RutgerKe:

There are no metrics for the journal target yet, this adds two basic ones:
- Total number of lines processed
- Total number of errors processing lines (with a reason in a label)

Because of the way the journal works, tailing one file with many possible processes writing to it, I chose note to expose counts per unit.

Fixes #5554

Doing this PR mostly to learn, not because I really need this feature. So all feedback/opinions are welcome!
I looked at the other targets and did not see any tests for the metrics produced/no pattern to copy here, but I'd be happy to figure out how to add them if requested. 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/clients/pkg/promtail/targets/journal/journaltarget.go
+++ b/clients/pkg/promtail/targets/journal/journaltarget.go
@@ -90,6 +90,7 @@ var defaultJournalEntryFunc = func(c sdjournal.JournalReaderConfig, cursor strin
 // JournalTarget tails systemd journal entries.
 // nolint
 type JournalTarget struct {
+	metrics       *Metrics
 	logger        log.Logger
 	handler       api.EntryHandler
 	positions     positions.Positions
@@ -104,6 +105,7 @@ type JournalTarget struct {
 
 // NewJournalTarget configures a new JournalTarget.
 func NewJournalTarget(
+	metrics *Metrics,
 	logger log.Logger,
 	handler api.EntryHandler,
 	positions positions.Positions,
@@ -113,6 +115,7 @@ func NewJournalTarget(
 ) (*JournalTarget, error) {
 
 	return journalTargetWithReader(
+		metrics,
 		logger,
 		handler,
 		positions,
@@ -125,6 +128,7 @@ func NewJournalTarget(
 }
 
 func journalTargetWithReader(
+	metrics *Metrics,
 	logger log.Logger,
 	handler api.EntryHandler,
 	pos positions.Positions,
@@ -147,6 +151,7 @@ func journalTargetWithReader(
 
 	until := make(chan time.Time)
 	t := &JournalTarget{
+		metrics:       metrics,
 		logger:        logger,
 		handler:       handler,
 		positions:     pos,
@@ -264,6 +269,7 @@ func (t *JournalTarget) formatter(entry *sdjournal.JournalEntry) (string, error)
 		bb, err := json.Marshal(entry.Fields)
 		if err != nil {
 			level.Error(t.logger).Log("msg", "could not marshal journal fields to JSON", "err", err)
+			t.metrics.journalErrors.WithLabelValues("cannot_marshal").Inc()
 			return journalEmptyStr, nil
 		}
 		msg = string(bb)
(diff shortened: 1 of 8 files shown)


clients/pkg/promtail/targets/journal/journaltarget.go
```diff
@@ -264,6 +269,7 @@ func (t *JournalTarget) formatter(entry *sdjournal.JournalEntry) (string, error)
 		bb, err := json.Marshal(entry.Fields)
 		if err != nil {
 			level.Error(t.logger).Log("msg", "could not marshal journal fields to JSON", "err", err)
+			t.metrics.journalErrors.WithLabelValues("cannot_marshal").Inc()
```
trevorwhitney, line 272: do we want to move these reasons to constants so someone can quickly see how many there are? we want to make sure not to explode cardinality too much by adding too many of them, but 3 seems pretty reasonable for now.

trevorwhitney (commented):
Thanks for the contribution! Curious about the decision to not include unit as a label though, could you elaborate on that a bit? I feel like if I was getting errors, I would want to know which unit was emitting these bad logs so I could try to fix it, or remove it from my scrape targets. What value are the error metrics without that, are there other ways to investigate what's wrong?

RutgerKe:
> Thanks for the contribution! Curious about the decision to not include unit as a label though, could you elaborate on that a bit? I feel like if I was getting errors, I would want to know which unit was emitting these bad logs so I could try to fix it, or remove it from my scrape targets. What value are the error metrics without that, are there other ways to investigate what's wrong?

Sure, I was mainly thinking about the number of units: this can easily be in the hundreds, I'm not sure if we should create a label for each of them? It kind of goes towards analyzing the content of the data, instead of providing metrics if the software is working. I personally like providing as much insights into data as possible, but should this be a concern of promtail? This might be better done in grafana by querying the loki data itself?

I do get your use case, I'm happy to add them to all metrics, a comprise could be to just add them to the errors.

clients/pkg/promtail/targets/journal/journaltarget.go
```diff
@@ -264,6 +269,7 @@ func (t *JournalTarget) formatter(entry *sdjournal.JournalEntry) (string, error)
 		bb, err := json.Marshal(entry.Fields)
 		if err != nil {
 			level.Error(t.logger).Log("msg", "could not marshal journal fields to JSON", "err", err)
```
trevorwhitney, line 271: do we have access to the unit here? could we include it in the log message since we're not going to include it in the metric?

clients/pkg/promtail/targets/journal/journaltarget.go
```diff
@@ -272,6 +278,7 @@ func (t *JournalTarget) formatter(entry *sdjournal.JournalEntry) (string, error)
 		msg, ok = entry.Fields["MESSAGE"]
 		if !ok {
 			level.Debug(t.logger).Log("msg", "received journal entry with no MESSAGE field")
```
trevorwhitney, line 280: and here as well?

trevorwhitney (approved):
I buy your argument for not including the unit in the metric labels. How would you feel about emitting a log line with the unit though for the purposes of investigating the problem?

kavirajk (commented):
LGTM. Thanks @RutgerKe :)

Just few minor nits. Let me know once you fix those. I can merge.

clients/pkg/promtail/targets/journal/journaltarget.go
RutgerKe: Added it, its not always available, but I'd say most of the time it is & adds some value

kavirajk:
> Turns out I couldn't really trigger the error returned from the marshal function, so I deleted that counter. (Though: If you have any idea how to trigger it / in what situations this occurs, let me know and I can add it back in).

Given that filed is `map[string]string` I think it always marshall well into JSON. So I also think we may not need that error count.

kavirajk (approved):
LGTM 👍

PR #6107, opened 2022-05-05
Tsdb/head wal feedback

owen-d:
Followup to https://github.com/grafana/loki/pull/6049 which was merged before I addressed feedback

I ran into some issues trying out a `sync.Map` approach, so I reverted it and will leave it for later as a potential optimization.

Ref https://github.com/grafana/loki/issues/5428


cyriltovena (approved):
LGTM

chaudum (approved):

Issue #6023, opened 2022-04-26
During rollout ingester GRPC call are too long to timeout


cyriltovena:
When doing a rollout of ingester, querier might have already initiated a GRPC call, however it seems that it can takes up to 13s for the querier to realized that and that adds up to the latency before retrying.

We should try to investigate if we can kill the connection as soon as an ingester is restarting or increase the healthcheck loop.

The best way to go about this is to write a test using in-memory connection and try the use case. see https://github.com/grafana/loki/blob/main/pkg/ingester/ingester_test.go#L751

<img width="2298" alt="image" src="https://user-images.githubusercontent.com/1053421/165293759-dabeaf1b-0606-45ac-92f6-6fc540351aa3.png">

chaudum:
gRPC healthcheck can still return `SERVING` even though the instance is in state `LEAVING`, see https://github.com/grafana/loki/blob/main/pkg/ingester/ingester.go#L781-L783

kavirajk:
@chaudum nice catch. Are we able to reproduce it anywhere?

JStickler:
Closing as stale. This issue was opened against a version of Loki that is no longer supported. If this issues still exists in more recent versions of Loki, please leave a comment and we can reopen the issue.

PR #6113, opened 2022-05-06
Remove unused grpc health check methods

kavirajk:
We have generic `grpc_healthcheck` endpoint that get registered for all the components
during initialization stage.
https://github.com/grafana/loki/blob/main/pkg/loki/loki.go#L364

Looks like these old health checks that are part of actual component's struct itself came from legacy.

NOTE: This should be **no** op. No build or test failure as **there are no consumer for these methods**

Fixes # NA

We figured this out while investigating this ingester/querier grpc issue https://github.com/grafana/loki/issues/6023

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -778,15 +778,6 @@ func (i *Ingester) Series(ctx context.Context, req *logproto.SeriesRequest) (*lo
 	return instance.Series(ctx, req)
 }
 
-// Check implements grpc_health_v1.HealthCheck.
-func (i *Ingester) Check(ctx context.Context, req *grpc_health_v1.HealthCheckRequest) (*grpc_health_v1.HealthCheckResponse, error) {
-	status := grpc_health_v1.HealthCheckResponse_SERVING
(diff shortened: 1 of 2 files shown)


chaudum (approved):

pkg/ingester/ingester.go
```diff
@@ -778,15 +778,6 @@ func (i *Ingester) Series(ctx context.Context, req *logproto.SeriesRequest) (*lo
 	return instance.Series(ctx, req)
 }
 
-// Check implements grpc_health_v1.HealthCheck.
-func (i *Ingester) Check(ctx context.Context, req *grpc_health_v1.HealthCheckRequest) (*grpc_health_v1.HealthCheckResponse, error) {
-	status := grpc_health_v1.HealthCheckResponse_SERVING
-	if (i.State() != services.Running) || (i.lifecycler.GetState() != ring.ACTIVE) {
```
owen-d, line 784: I'm a bit worried that we don't seem to be checking the lifecycler state in our blanket version, but it doesn't look like these are used either.

owen-d (approved):

PR #6115, opened 2022-05-06
Improve JSON parser allocations

cyriltovena:
This is a refactoring of the json parser to avoid a maximun of allocations.

bench so far:

```
name           old time/op    new time/op    delta
JSONParser-16    1.53µs ± 1%    1.63µs ± 4%   +6.38%  (p=0.008 n=5+5)

name           old alloc/op   new alloc/op   delta
JSONParser-16      952B ± 0%      872B ± 0%   -8.40%  (p=0.008 n=5+5)

name           old allocs/op  new allocs/op  delta
JSONParser-16      12.0 ± 0%       6.0 ± 0%  -50.00%  (p=0.008 n=5+5)
```

### DIFF
--- a/pkg/logql/log/json.go
+++ b/pkg/logql/log/json.go
@@ -1,0 +1,165 @@
+package log
+
+import (
+	"bytes"
+	"fmt"
+	"unicode/utf8"
+
+	"github.com/buger/jsonparser"
+	jsoniter "github.com/json-iterator/go"
+)
+
+const (
+	jsonSpacer      = '_'
+	duplicateSuffix = "_extracted"
+	trueString      = "true"
+	falseString     = "false"
+)
+
+var (
+	_                       Stage = &JSONParser{}
+	errUnexpectedJSONObject       = fmt.Errorf("expecting json object(%d), but it is not", jsoniter.ObjectValue)
+)
+
+type JSONParser struct {
+	buf    []byte // buffer used to build json keys
+	prefix []byte
+	lbs    *LabelsBuilder
+
+	keys internedStringSet
+}
+
+// NewJSONParser creates a log stage that can parse a json log line and add properties as labels.
+func NewJSONParser() *JSONParser {
+	return &JSONParser{
+		buf:    make([]byte, 0, 1024),
+		prefix: make([]byte, 0, 1024),
+		keys:   internedStringSet{},
+	}
+}
+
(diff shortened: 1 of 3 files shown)


pkg/logql/log/json.go
```diff
@@ -0,0 +1,165 @@
...
+		if bytes.ContainsRune(value, utf8.RuneError) {
+			return ""
+		}
+		// todo if j.buf has grown it won't be reused.
+		// unlikely to happen.
+		v, err := jsonparser.Unescape(value, j.buf)
+		if err != nil {
+			return ""
+		}
+		return string(j.buf)
```
jeschkies, line 130: This creates a copy right?

pkg/logql/log/json.go
```diff
@@ -0,0 +1,165 @@
...
+	case jsonparser.Number:
+		return string(value)
```
jeschkies, line 132: Same here. I was wondering if it would be possible to use an unsafe to string here. This should work as long as the `line` slice is not changing.

pkg/logql/log/json.go
```diff
@@ -0,0 +1,165 @@
...
+		key, ok := j.keys.Get(field, func() (string, bool) {
+			j.buf = sanitizeLabelTo(field, j.buf, true)
+			if j.lbs.BaseHas(unsafeGetString(j.buf)) {
+				j.buf = append(j.buf, duplicateSuffix...)
+			}
+			if !j.lbs.ParserLabelHints().ShouldExtract(unsafeGetString(j.buf)) {
+				return "", false
+			}
+
+			return string(j.buf), true
```
jeschkies, line 88: If I read `sanitizeLabelTo` correctly it makes a copy. Won't `string(j.buf)` make another copy here?

pkg/logql/log/json.go
cyriltovena: Yes this one is impossible to avoid right now.

pkg/logql/log/json.go
cyriltovena: It will change, this required a wider refactoring.

pkg/logql/log/json.go
cyriltovena: I think `sanitizeLabelTo` does avoid a copy, that's why it's named like this it sanitize into a buffer.

PR #6117, opened 2022-05-06
loads previously created TSDBs into shipper on startup

owen-d:
This PR ensures that on startup, the index-shipper is loaded with any previously created tsdb indices, which may or may not have already been flushed to storage. This is important for two reasons:
1) If they were not flushed, this will re-register them to be flushed to storage
2) It will make previously built TSDBs available for querying in the `ingester` and `write` targets, which use the `uploadmanager` for reads.


slim-bean (approved):
LGTM, not sure if we might want an info level log line to output some statistics? like how many tables were loaded and how long it took? This has been kind of handy with our current boltdb

PR #6118, opened 2022-05-06
Add upgrade guide regarding azure container name change

trevorwhitney:

In https://github.com/grafana/loki/pull/6015 we changed the default value of the Azure container name from `cortex` to `loki`. If anyone was relying on this value, they will need to manually specify it when upgrading to the latest version.


KMiller-Grafana (approved):
Approving this PR and hoping that you'll put in my suggested change.

owen-d (approved):

PR #6125, opened 2022-05-09
operator: Add method to get authenticated from GCP

sasagarw:
There isn't an authentication mechanism in the provided Loki configuration. The Loki GCS authentication is handled via the GCS SDK which is typically done with environment variables(For example:GOOGLE_APPLICATION_CREDENTIALS).

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/operator/internal/manifests/compactor.go
+++ b/operator/internal/manifests/compactor.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path"
 
+	lokiv1beta1 "github.com/grafana/loki/operator/api/v1beta1"
 	"github.com/grafana/loki/operator/internal/manifests/internal/config"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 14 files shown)


sasagarw:
cc @periklis

operator/internal/manifests/compactor.go
```diff
@@ -99,6 +100,26 @@ func NewCompactorStatefulSet(opts Options) *appsv1.StatefulSet {
...
+			Name:      opts.Stack.Storage.Secret.Name,
+			ReadOnly:  false,
+			MountPath: secretDirectory,
+		})
+		podSpec.Containers[0].Env = append(podSpec.Containers[0].Env, corev1.EnvVar{
+			Name:  EnvGoogleApplicationCredentials,
+			Value: path.Join(secretDirectory, GCSFileName),
+		})
+	}
+
```
Red-GV, line 122: Maybe this should just be a callable function in `var.go`? Similar to how we do the modifications for the service monitor?

Red-GV (commented):
Looks overall fine to me. A small comment though.

operator/internal/manifests/var.go
```diff
@@ -273,3 +279,23 @@ func lokiReadinessProbe() *corev1.Probe {
...
+	})
+	podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, corev1.VolumeMount{
+		Name:      secretName,
+		ReadOnly:  false,
+		MountPath: secretDirectory,
+	})
+	podSpec.Containers[0].Env = append(podSpec.Containers[0].Env, corev1.EnvVar{
+		Name:  EnvGoogleApplicationCredentials,
+		Value: path.Join(secretDirectory, GCSFileName),
+	})
```
periklis, line 300: I suggest to continue using `mergo.Merge` for mutating manifests as we do in other places, e.g. [service monitor PKI](https://github.com/grafana/loki/blob/main/operator/internal/manifests/service_monitor.go#L136).

In addition I suggest we move the whole function out of `var.go` to `storage/configure.go` file and follow-up the configure pattern as in `gateway_tenants.go`, i.e.
- First in `storage/configure.go` you place the `ensureCredentialsFor GCS`
- Second in `manifests/object_storage.go` you place the switch based on `opts.Stack.Storage.Secret.Type`
- Third you call the `configureDeploymentForStorageType` and `configureStatefulSetForStorageType` only in the Builder functions (e.g. `BuildQuerier` w/o nesting if blocks in the original constructor functions (e.g. `NewQuerierDeployment`)

This ensures that we have dedicated configure/mutate functions with proper error handling (thanks to `mergo.Merge` while keeping our manifest constructors are pure and minimal as possible.

e.g.
```go
func configureDeploymentForStorageType(d *appsv1.Deployment, t lokiv1beta1.ObjectStorageType) error {
	switch t {
	case lokiv1beta1.ObjectStorageSecretGCS:
		return storage.ConfigureDeployment(d)
	default:
		return nil
	}
}
func configureStatefulSetForStorageType(d *appsv1.Deployment, t lokiv1beta1.ObjectStorageType) error {
	switch t {
	case lokiv1beta1.ObjectStorageSecretGCS:
		return storage.ConfigureStatefulSet(d)
	default:
		return nil
	}
}
```

periklis (changes requested):

periklis (approved):
Looks really good to me.

PR #6132, opened 2022-05-10
Fix certain binary expression queries by optimizing through push down

### LINKED ISSUE #6130: Instant query returns `unimplemented` error
**Describe the bug**

The following instant query returns an error `unimplemented`:

```logql
sum (count_over_time({foo="bar"} | logfmt | duration > 2s [3s]))
/
sum (count_over_time({foo="bar"} [3s]))
```

**To Reproduce**

Add the following case to the `TestRangeMappingEquivalence` (downstream_test.go) test:

```
{`sum (count_over_time({a=~".+"} | logfmt | b > 2 [3s])) / sum (count_over_time({a=~".+"} [3s]))`, time.Second},
```

**Expected behavior**

Should return correct result.

chaudum:

This PR optimizes instant queries that use a vector aggregation around a range aggregation with log extraction stage (`json`, `logfmt`), such as `sum(count_over_time({foo="bar"} | logfmt | duration > 30s [24h]))`.

Since the vector aggregation can be pushed down to the downstream query, the downstream query does not create a massive amount of streams, even though it contains a generic label extraction stage.

This would also fix binary expression queries that use said aggregation on one of its leave nodes, such as:

```
sum(count_over_time({foo="bar"} | logfmt | duration > 2s [3s]))
/
sum(count_over_time({foo="bar"} [3s]))
```

If either the left hand side or the right hand side of a binary expression is a noop, we need to return the original expression so the whole expression is a noop as well, and thus not executed using the downstream engine.
Otherwise, a binary expression that has a noop on either side, results in an `unimplemented` error when executed using the downstream engine.

Fixes #6130 

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/logql/downstream_test.go
+++ b/pkg/logql/downstream_test.go
@@ -132,14 +132,19 @@ func TestRangeMappingEquivalence(t *testing.T) {
 
 		// sum
 		{`sum(bytes_over_time({a=~".+"}[2s]))`, time.Second},
+		{`sum(bytes_over_time({a=~".+"} | logfmt | line > 5 [2s]))`, time.Second},
 		{`sum(count_over_time({a=~".+"}[2s]))`, time.Second},
+		{`sum(count_over_time({a=~".+"} | logfmt | line > 5 [2s]))`, time.Second},
 		{`sum(sum_over_time({a=~".+"} | unwrap b [2s]))`, time.Second},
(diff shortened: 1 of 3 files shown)


sandeepsukhani (approved):
LGTM

PR #6135, opened 2022-05-10
promtail: Fix infinite loop in file target (fixes #5350)

### LINKED ISSUE #5350: promtail causing very high cpu load when running and stops
On several of my servers i want to use promtail like i did on a bunch of other machines.
But when i start promtail on only the var/log logs, the CPU load goes trough the roof (like 15 to 20) and then promtail stops.
Looks like kswap is joining the party in being high on CPU when running promtail.

Is there anybody out there that has this same issue?

It's on RHEL7 by the way.

tcolgate:

This attempt to fix an issue where promtail consumes 100% CPU on
servers when some pods stop.

Fixes #5350

This may not be the best fix as a subsequent write to 
quit could block (though if quit where/is buffered, that
may not be a problem).

### DIFF
--- a/clients/pkg/promtail/targets/file/filetarget.go
+++ b/clients/pkg/promtail/targets/file/filetarget.go
@@ -168,7 +168,11 @@ func (t *FileTarget) run() {
 
 	for {
 		select {
-		case event := <-t.fileEventWatcher:
+		case event, ok := <-t.fileEventWatcher:
+			if !ok {
+				// fileEventWatcher has been closed
+				return
+			}
 			switch event.Op {
 			case fsnotify.Create:
 				t.startTailing([]string{event.Name})


tcolgate:
> How often has the issue occured in your environments? How long until you can confirm the change _probably_ fixes the problem? I guess then I'd build off your branch and deploy that promtail version in our environment.

The issue has been trivially reproducible for me. It was guaranteed to occur when one of our CI jobs finished, so I've been testing as I go. This fix definitely fixes the issue I've been having. That said, I have seen the `for { select { } }` in several other places (it's quite a common idiom), and it's not impossible that there are other variants that could be responsible for the same problem.

clients/pkg/promtail/targets/file/filetarget.go
```diff
@@ -168,7 +168,11 @@ func (t *FileTarget) run() {
 
 	for {
 		select {
-		case event := <-t.fileEventWatcher:
+		case event, ok := <-t.fileEventWatcher:
+			if !ok {
+				// fileEventWatcher has been closed
```
DylanGuedes, line 173: WDYT of adding a debug or info message here saying the file was closed?

Mario-Hofstaetter:
@tcolgate Do you have time to carry on with this PR? 🙏

tcolgate:
Sorry, I've been on leave for 2 weeks. I'll pick this up tomorrow.
FWIW, I suspect this fix could be problematic. It works well for my case, but ultiamtely I think it is caused by the fact that the map of open channels only tracks the file path pattern, not the entirety of the actual target (because the path is unique in the positions file). Essentially two targets are working on the same file, they just happen to not overlap, and the second incarnation of the target will never see any writes to the file.
  This fix works for me because once my jobs have finished, I never write to the file again, so the fact that the channel is closed doesn't matter. But in other scenarios, it might be perfectly reasonable to expect writes to the file which switched between targets, and with this patch, I think those writes would be lost

cyriltovena (approved):
LGTM

@DylanGuedes feel free to follow up with the change you had in mind.

Mario-Hofstaetter:
@cyriltovena Has this not been released with version `2.6` ? The commit was merged into `main` already 25 days ago?

```text
$ git log main --grep '#6135'
commit 21f69643b3cc99c85ea48c983c1967916139b95e
Author: Tristan Colgate-McFarlane <tcolgate@gmail.com>
Date:   Thu Jun 16 08:55:19 2022 +0100

    promtail: fix infinite loop in file target (fixes #5350) (#6135)

$ git log v2.6.0 --grep '#6135'

# <nope>
```

DylanGuedes:
> @cyriltovena Has this not been released with version `2.6` ? The commit was merged into `main` already 25 days ago?
> 
> ```
> $ git log main --grep '#6135'
> commit 21f69643b3cc99c85ea48c983c1967916139b95e
> Author: Tristan Colgate-McFarlane <tcolgate@gmail.com>
> Date:   Thu Jun 16 08:55:19 2022 +0100
> 
>     promtail: fix infinite loop in file target (fixes #5350) (#6135)
> 
> $ git log v2.6.0 --grep '#6135'
> 
> # <nope>
> ```

looks like it was not :(

anarcher:
```
$ git log v2.7.0 --grep '#6135'
commit 21f69643b3cc99c85ea48c983c1967916139b95e
Author: Tristan Colgate-McFarlane <tcolgate@gmail.com>
Date:   Thu Jun 16 08:55:19 2022 +0100

    promtail: fix infinite loop in file target (fixes #5350) (#6135)
```

PR #6139, opened 2022-05-10
Documentation: Explain --config.expand-env=true double slash with slash substitution

### LINKED ISSUE #3635: Promtail --config.expand-env=true parameter breaks Promtail configuration if references to regex capture groups exist
**Describe the bug**
References to capture groups (`$1` or `${1}`) in Promtail stage configurations (for example, template with `regexReplaceAll`) are treated as environment variables when `--config.expand-env=true` Promtail parameter is used. This breaks the stage configurations, and Promtail fails to start because of invalid syntax.

**To Reproduce**
1. Given this json log file
```
{"hostname": "host-10-1-2-3.prod.example.com"}
```
2. Intention is to remove `.prod.example.com` from `hostname` value. Here is a corresponding Promtail configuration: 
```
server:
  http_listen_port: 9080

positions:
  filename: /tmp/positions.yaml

client:
  url: http://localhost:3100/loki/api/v1/push

scrape_configs:
  - job_name: test-job
    pipeline_stages:
    - json:
        expressions:
          hostname: hostname
    - template:
        source: hostname
        template: '{{ regexReplaceAll "(.*?)\\.(.*)" .Value "$1" }}'
    static_configs:
    - targets:
        - localhost
      labels:
        __path__: /tmp/json.log
```
3. Run in dry run mode
```
$ cat json.log | promtail --stdin --dry-run --config.file promtail.yaml --config.expand-env=true
Clients configured:
----------------------
url: http://localhost:3100/loki/api/v1/push
batchwait: 1s
batchsize: 1048576
backoff_config:
  min_period: 500ms
  max_period: 5m0s
  max_retries: 10
timeout: 10s
tenant_id: ""

level=error ts=2021-04-21T12:15:49.822312Z caller=main.go:106 msg="error creating promtail" error="invalid template stage config: template: pipeline_template:1: invalid syntax"
```

**Expected behavior**
The template syntax is valid in this case, and this can be confirmed by running the same command but without passing the `--config.expand-env=true`:
```
$ cat json.log | promtail --stdin --dry-run --config.file promtail.yaml
Clients configured:
----------------------
url: http://localhost:3100/loki/api/v1/push
batchwait: 1s
batchsize: 1048576
backoff_config:
  min_period: 500ms
  max_period: 5m0s
  max_retries: 10
timeout: 10s
tenant_id: ""

level=info ts=2021-04-21T12:14:43.518236Z caller=server.go:225 http=[::]:9080 grpc=[::]:9095 msg="server listening on addresses"
level=info ts=2021-04-21T12:14:43.518533Z caller=main.go:110 msg="Starting Promtail" version="(version=, branch=, revision=)"
2021-04-21T15:14:43	{__path__="/tmp/json.log"}	{"hostname": "host-10-1-2-3.prod.example.com"}
```

**Environment:**
 - Infrastructure: none is required to reproduce
 - Deployment tool: none is required to reproduce

V-Mann-Nick:

When I was configuring Promtail to use a regex stage I escaped slashes in regular expressions when using double quotes in yaml files. But no where did it explain that when using the option `--config.expand-env=true` the program `envsubst` would replace double slashes with slashes. I found the issue #3635 to find out this is behavior as expected. While I accept this, I thought it would be useful to warn the next person about just this case.

Adds two paragraphs to the Promtail configuration docs:
- in `docs/sources/clients/promtail/configuration.md`: use double slashes for single slashes when using `expand-env=true`
- in `docs/sources/clients/promtail/stages/regex.md`: explained again as it's more likely to use back slashes in regex

Although already closed it's related to the following: Fixes #3635

**Checklist**
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/clients/promtail/configuration.md
+++ b/docs/sources/clients/promtail/configuration.md
@@ -60,6 +60,11 @@ ${VAR:default_value}
 
 Where default_value is the value to use if the environment variable is undefined.
 
+**Note**: With `expand-env=true` the configuration will first run through
+[envsubst](https://linux.die.net/man/1/envsubst) which will replace double
+slashes with single slashes. Because of this every use of a slash `\` needs to
+be replaced with a double slash `\\`
+
 ### Generic placeholders:
 
 - `<boolean>`: a boolean that can take the values `true` or `false`
--- a/docs/sources/clients/promtail/configuration.md
+++ b/docs/sources/clients/promtail/configuration.md
@@ -61,7 +61,7 @@ ${VAR:default_value}
 Where default_value is the value to use if the environment variable is undefined.
 
 **Note**: With `expand-env=true` the configuration will first run through
-[envsubst](https://linux.die.net/man/1/envsubst) which will replace double
+[envsubst](https://pkg.go.dev/github.com/drone/envsubst) which will replace double
 slashes with single slashes. Because of this every use of a slash `\` needs to
 be replaced with a double slash `\\`
 
--- a/docs/sources/clients/promtail/stages/regex.md
+++ b/docs/sources/clients/promtail/stages/regex.md
@@ -38,6 +38,16 @@ But these are not:
 - `expression: '\\w*'` (only escape backslashes when using double quotes)
 - `expression: "\w*"` (backslash must be escaped)
 
+
+If you run Promtail with the `--config.expand-env=true` flag the configuration
+will run through [envsubst](https://linux.die.net/man/1/envsubst) which will
+replace double slashes with single slashes. Because of this when using
+`expand-env=true` you need to use double slashes for each single slash. For
+example:
+
+- `expression: '\w*'` must be `expression: '\\w*'`
+- `expression: "\\w*"` must be `expression: "\\\\w*"`
+
 ## Example
 
 ### Without `source`
--- a/docs/sources/clients/promtail/stages/regex.md
+++ b/docs/sources/clients/promtail/stages/regex.md
@@ -40,7 +40,7 @@ But these are not:
 
 
 If you run Promtail with the `--config.expand-env=true` flag the configuration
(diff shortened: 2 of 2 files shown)


docs/sources/clients/promtail/stages/regex.md
```diff
@@ -38,6 +38,16 @@ But these are not:
 - `expression: '\\w*'` (only escape backslashes when using double quotes)
 - `expression: "\w*"` (backslash must be escaped)
 
+
+If you run Promtail with the `--config.expand-env=true` flag the configuration
+will run through [envsubst](https://linux.die.net/man/1/envsubst) which will
```
kavirajk, line 43:
```suggestion
will run through [envsubst](https://pkg.go.dev/github.com/drone/envsubst)  which will
```
same here.

docs/sources/clients/promtail/configuration.md
```diff
@@ -60,6 +60,11 @@ ${VAR:default_value}
 
 Where default_value is the value to use if the environment variable is undefined.
 
+**Note**: With `expand-env=true` the configuration will first run through
+[envsubst](https://linux.die.net/man/1/envsubst) which will replace double
```
kavirajk, line 64:
```suggestion
[envsubst](https://pkg.go.dev/github.com/drone/envsubst) which will replace double
```
would recommend to link the actual package we use to avoid missing any __real__ behaviour of `envsubst`.

kavirajk (commented):
thanks @V-Mann-Nick :+1: LGTM. Added one small suggestion with linking reference.

V-Mann-Nick:
Thanks for the suggestions @kavirajk. Didn't know that. I committed them.

kavirajk (approved):
Thanks @V-Mann-Nick LGTM

PR #6142, opened 2022-05-11
tsdb query fixes

sandeepsukhani:
Doing some minor fixes for the issues that I found on my local setup:
1. Uploads manager was never queried when downloads manager was initialized in index shipper.
2. Files created from nodes having `-` in their names were failing to be opened.
3. `__name__` matcher was not removed from queries passed to tsdb head.

### DIFF
--- a/pkg/storage/stores/indexshipper/shipper.go
+++ b/pkg/storage/stores/indexshipper/shipper.go
@@ -139,11 +139,15 @@ func (s *indexShipper) AddIndex(tableName, userID string, index index.Index) err
 
 func (s *indexShipper) ForEach(ctx context.Context, tableName, userID string, callback func(index index.Index) error) error {
 	if s.downloadsManager != nil {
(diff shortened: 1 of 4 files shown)


pkg/storage/stores/indexshipper/shipper.go
```diff
@@ -139,11 +139,15 @@ func (s *indexShipper) AddIndex(tableName, userID string, index index.Index) err
 
 func (s *indexShipper) ForEach(ctx context.Context, tableName, userID string, callback func(index index.Index) error) error {
 	if s.downloadsManager != nil {
-		return s.downloadsManager.ForEach(ctx, tableName, userID, callback)
+		if err := s.downloadsManager.ForEach(ctx, tableName, userID, callback); err != nil {
```
owen-d, line 142: I thought this was by design: only using either the downloads mgr or the uploads mgr so we didn't query the same indices in `R/W` mode for the shipper.

PR #6152, opened 2022-05-12
Loki: When processing logs during tailing, don't reuse the query pipeline

slim-bean:

The query pipeline will build a label cache to improve performance when processing queries, the normal use of a pipeline for queries is ephemeral so the cache is cleared when the pipeline is discarded.

When tailing, however, the pipeline was long lived and reused which lead to an unbounded growth of cached labels for certain tail queries.

This PR creates a new pipeline for every processed push into Loki vs reusing. 

There is probably future potential for a mechanism to clear the cache but not sure it's worth the additional effort vs creating and disposing of pipelines.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@
 ##### Enhancements
 
 ##### Fixes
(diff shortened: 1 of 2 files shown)


sandeepsukhani (approved):
LGTM

PR #6160, opened 2022-05-13
Look up config file in default locations

chaudum:

This change allows to start Loki without `-config.file` argument, by making it optional and by default look up the config file in the current directory or in a config/ sub-directory.

This makes first use of Loki even simpler, especially when we distribute a tarball/zip that contains both the binary and a config file for use with local filesystem. Then Loki can be invoked simply with `./bin/loki` within the extracted folder.

Pr-requisite for https://github.com/grafana/loki/pull/6159

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/util/cfg/files.go
+++ b/pkg/util/cfg/files.go
@@ -66,7 +66,7 @@ func dYAML(y []byte) Source {
 	}
 }
 
-func YAMLFlag(args []string, name string) Source {
+func ConfigFileLoader(args []string, name string) Source {
 
 	return func(dst Cloneable) error {
 		freshFlags := flag.NewFlagSet("config-file-loader", flag.ContinueOnError)
--- a/pkg/util/cfg/files.go
+++ b/pkg/util/cfg/files.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/drone/envsubst"
 	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


pkg/util/cfg/files.go
```diff
@@ -92,13 +93,18 @@ func YAMLFlag(args []string, name string) Source {
...
+			expandEnv := false
+			expandEnvFlag := freshFlags.Lookup("config.expand-env")
+			if expandEnvFlag != nil {
+				expandEnv, _ = strconv.ParseBool(expandEnvFlag.Value.String()) // Can ignore error as false returned
+			}
+			if _, err := os.Stat(val); err == nil {
+				return YAML(val, expandEnv)(dst)
+			}
+		}
+		return fmt.Errorf("%s does not exist", f.Value.String())
```
kavirajk, line 108:
```suggestion
		return fmt.Errorf("%s config does not exist, set %s for custom config path", f.Value.String(), name)
```

kavirajk: Added this suggestion because, currently it just prints
`config.yaml does not exist` if expected config file not found. Better to give some context where this is coming from and how to change it.

kavirajk (commented):
LGTM. Added one small suggestion to make error more useful.

owen-d (commented):
I think this is pretty reasonable, but want to let others dissent before we merge (cc @cyriltovena @slim-bean). Also, can we add some docs describing the new behavior?

cstyan:
I think having a default location for the config file makes sense

kavirajk (approved):
Nice addition :). Would wait what @slim-bean think before merging it though.

vlad-diachenko (approved):
LGTM

owen-d (approved):

PR #6171, opened 2022-05-16
Logql Analyzer

vlad-diachenko:
It's API server that returns detailed information about execution LogQL expression against user's log line. This API is used on https://grafana.com/docs/loki/latest/logql/analyzer/ page
Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated


pkg/logql/log/pipeline.go
```diff
@@ -118,6 +119,85 @@ func NewPipeline(stages []Stage) Pipeline {
...
+}
+
+type streamPipelineAnalyzer struct {
+	origin       *pipeline
+	stagesCount  int
+	streamLabels labels.Labels
+}
+
+func NewPipelineAnalyzer(origin Pipeline, streamLabels labels.Labels) PipelineAnalyzer {
+	if o, ok := origin.(*pipeline); ok {
```
owen-d, line 140: type switching here will work, but has the tendency to break when we refactor or extend the code (such as renaming `type pipeline struct` or adding another pipeline variant.

Instead, we can add a new interface

```go
type PipelineStages interface {
  Pipeline
  Stages() []Stage
}
```

That way, we can do two things:
1) Make the `PipelineAnalyzer` work for all types that implement `PipelineStages`, including those we may add in the future
2) Let's us move the `PipelineAnalyzer` code into the `logqlanalyzer` pkg as it no longer depends on the unexported `pipeline` struct.

What do you think?

pkg/logqlanalyzer/analyzer.go
```diff
@@ -0,0 +1,83 @@
...
+			LabelsBefore: mapAllToLabelsResponse(record.LabelsBefore),
+			LineAfter:    record.LineAfter,
+			LabelsAfter:  mapAllToLabelsResponse(record.LabelsAfter),
+			FilteredOut:  record.FilteredOut,
+		})
+	}
+	return LineResult{originLine, stageRecords}
+}
+
+func mapAllToLabelsResponse(labels labels.Labels) []Label {
```
owen-d, line 77: This can be replaced with `Labels.Copy()`

owen-d (commented):
Looking great, left a few notes.

cmd/logql-analyzer/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+WORKDIR /src/loki
+RUN make clean && CGO_ENABLED=0 go build ./cmd/logql-analyzer/
+
+FROM alpine:3.15.4
+
+RUN apk add --no-cache ca-certificates
+
+COPY --from=build /src/loki/logql-analyzer /usr/bin/logql-analyzer
+
+ENTRYPOINT [ "/usr/bin/logql-analyzer", "--server.http-listen-port=3001" ]
```
jeschkies, line 13: Should we rather pull the port from a config file?

pkg/logql/log/pipeline.go
```diff
@@ -118,6 +119,85 @@ func NewPipeline(stages []Stage) Pipeline {
 	}
 }
 
+type PipelineAnalyzer interface {
+	AnalyzeLine(line string) []StageAnalysisRecord
+}
+type noopPipelineAnalyzer struct {
+	_ *pipeline
```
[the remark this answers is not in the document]
vlad-diachenko: removed

cmd/logql-analyzer/Dockerfile
vlad-diachenko: we do not use the config files, just a few args to define the port and log level....anyway this port here is not needed, moved it to `docker-compose` file.

pkg/logql/log/pipeline.go
vlad-diachenko: done

owen-d (approved):
LGTM!

PR #6176, opened 2022-05-17
fix(querier): Record `result` length after response validation.

kavirajk:

Without this fix, it causes panic, if series or label endpoint returns error.

Fixes NA

@sandeepsukhani found querier panics in one of our internal Loki cluster because of this bug.

**Checklist**


sandeepsukhani (approved):
LGTM

PR #6177, opened 2022-05-17
Promtail: Support all `cri-o` tags (multiline tags).

### LINKED ISSUE #5854: Promtail: Support CRI-O multi-lines
Promtail has top-level support for CRI-O logs using the [cri stage](https://grafana.com/docs/loki/latest/clients/promtail/stages/cri/). However, this does not support multi-line CRI-O logs.  

The `cri` stage just handles multi-lines.

A custom `multiline` step. This is possible but the `cri` step removes some of the log tags so it doesn't work in conjunction.

**Additional context**
CRI-O has two log tags, `P` indicating a partial logline (one that will be continued by another partial logline, or by a 'full log line'), and `F` indicating a full logline. 
Also, Fluent Bit already supports this. The documentation is [here](https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/multiline-parsing#built-in-multiline-parsers) with some additional context in the [feature request.](https://github.com/fluent/fluent-bit/issues/1316)

kavirajk:
Context matters in log lines. 

Any log lines that are scrapped by promtail that is generated by `cri-0` runtime on Kubernetes can either be partial or full long line. Partial long lines are the result of "too" big (if it exceeds 8192 bytes per line) log lines that was split by the runtime.

And these partial and full log lines can be identified by "tags" that are part of log lines. Currently promtail ignores it. This PR make promtail supports `cri-o` tags.

Fixes #5854 

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/clients/pkg/logentry/stages/extensions.go
+++ b/clients/pkg/logentry/stages/extensions.go
@@ -1,6 +1,9 @@
 package stages
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/go-kit/log"
 	"github.com/prometheus/client_golang/prometheus"
 )
@@ -35,9 +38,44 @@ func NewDocker(logger log.Logger, registerer prometheus.Registerer) (Stage, erro
 	return NewPipeline(logger, stages, nil, registerer)
 }
 
+type cri struct {
+	// bounded buffer for CRI-O Partial logs lines (identified with tag `P` till we reach first `F`)
+	partialLines []string
+	base         *Pipeline
+}
+
+// implement Stage interface
+func (c *cri) Name() string {
+	return "cri"
+}
+
+// implements Stage interface
+func (c *cri) Run(entry chan Entry) chan Entry {
+	entry = c.base.Run(entry)
+
+	in := RunWithSkip(entry, func(e Entry) (Entry, bool) {
+		fmt.Println(e.Extracted)
+		if e.Extracted["flags"] == "P" {
+			// TODO(kavi): Make `c.partialLines` bounded to some upper length. Flush it any partialLines more than that.
+			c.partialLines = append(c.partialLines, e.Line)
+			return Entry{}, true
+		}
+		if len(c.partialLines) > 0 {
+			cpy := e.copy()
+			c.partialLines = append(c.partialLines, e.Line)
+			cpy.Line = strings.Join(c.partialLines, "\n")
+			e = *cpy
+			c.partialLines = c.partialLines[:0]
+		}
+		return e, false
+	})
+
+	return in
+}
+
 // NewCRI creates a CRI format specific pipeline stage
 func NewCRI(logger log.Logger, registerer prometheus.Registerer) (Stage, error) {
-	stages := PipelineStages{
+	base := PipelineStages{
 		PipelineStage{
 			StageTypeRegex: RegexConfig{
 				Expression: "^(?s)(?P<time>\\S+?) (?P<stream>stdout|stderr) (?P<flags>\\S+?) (?P<content>.*)$",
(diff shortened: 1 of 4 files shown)


clients/pkg/logentry/stages/extensions.go
```diff
@@ -35,9 +41,46 @@ func NewDocker(logger log.Logger, registerer prometheus.Registerer) (Stage, erro
...
+			if len(c.partialLines) >= MaxPartialLinesSize {
+				c.partialLines = c.partialLines[:0]
+			}
+			c.partialLines = append(c.partialLines, e.Line)
+			return Entry{}, true
+		}
+		if len(c.partialLines) > 0 {
+			cpy := e.copy()
+			c.partialLines = append(c.partialLines, e.Line)
+			cpy.Line = strings.Join(c.partialLines, "\n")
```
owen-d, line 71: Why are we copying the entry here?

owen-d (commented):
Some questions, but this is looking great!

clients/pkg/logentry/stages/extensions.go
```diff
@@ -35,9 +41,46 @@ func NewDocker(logger log.Logger, registerer prometheus.Registerer) (Stage, erro
...
+func (c *cri) Run(entry chan Entry) chan Entry {
+	entry = c.base.Run(entry)
+
+	in := RunWithSkip(entry, func(e Entry) (Entry, bool) {
+		fmt.Println(e.Extracted)
+		if e.Extracted["flags"] == "P" {
+			if len(c.partialLines) >= MaxPartialLinesSize {
+				c.partialLines = c.partialLines[:0]
+			}
+			c.partialLines = append(c.partialLines, e.Line)
```
trevorwhitney, line 65: `RunWithSkip` runs the passed process function in a go routine. Do we need to therefore but any locking around the `partialLines` slice that we're mutating here?

clients/pkg/logentry/stages/pipeline.go
```diff
@@ -74,6 +74,22 @@ func RunWith(input chan Entry, process func(e Entry) Entry) chan Entry {
 	return out
 }
 
+func RunWithSkip(input chan Entry, process func(e Entry) (Entry, bool)) chan Entry {
+	out := make(chan Entry)
+	go func() {
+		defer close(out)
+		for e := range input {
+			ee, skip := process(e)
```
trevorwhitney, line 82: see above, but since we're running process async, do we need to put any locking around the `partialLines` slices it's modifying?

trevorwhitney (commented):
This will be a great feature for peiple using `cri-o`, thanks Kavi! Just a couple questions (one original, one copycat).

clients/pkg/logentry/stages/extensions.go
kavirajk: Good point. But AFAIK, we don't need any such sync mechanism. Because
1. I think It is using go routine there just because it wants to return the channel directly without waiting for entries to come in.
2. Each pipeline stages in promtail will create separate CRI stage (via `NewCRI`) and all the stages are processed one by one as a pipeline. So I don't think calling CRI pipeline (that has partialline slice) concurrently make any sense.

That's my understanding. But someone can double check. May be @cyriltovena ?

owen-d (approved):

clients/pkg/logentry/stages/extensions.go
```diff
@@ -35,9 +41,50 @@ func NewDocker(logger log.Logger, registerer prometheus.Registerer) (Stage, erro
...
+				c.partialLines = c.partialLines[:0]
+				c.partialLines = append(c.partialLines, newPartialLine)
+				return e, false
+			}
+			c.partialLines = append(c.partialLines, e.Line)
+			return e, true
+		}
+		if len(c.partialLines) > 0 {
+			c.partialLines = append(c.partialLines, e.Line)
+			e.Line = strings.Join(c.partialLines, "\n")
```
dewe, line 76: I know this has already been merged, but I've got a question: what's the reason for joining with a `\n` here? 

Given that we originally had a message, split by cri-o into partial lines, adding a linefeed is effectively changing the resulting message.

For example, a json log message

```text
2022-09-20T17:51:01.214529932Z stdout P {"a": "very long json field 
2022-09-20T17:51:01.214697597Z stdout F split in two lines"}
```

will result in something that cannot be handled by a json parser

```text
2022-09-20T17:51:01.214697597+0000	{stream="stdout"}	{"a": "very long json field 
split in two lines"}
```

since control characters isn't valid in a json string.

This can be "fixed" by an extra stage replacing all linefeeds, but in this case it shouldn't be needed.

kavirajk: I think you are right. Let me fix that.

kavirajk: Fix: https://github.com/grafana/loki/pull/7997

PR #6179, opened 2022-05-17
feature: Add ingester handler for shutdown and forget tokens

chaudum:

This handler can be used to gracefully shut down a Loki instance and
delete the file that persists the tokens of the ingester ring.

In production environments you usually want to persist ring tokens so
that during a restart of an ingester instance, or during rollout, the
tokens from that instance are not re-distributed to other instances, but
instead kept so that the same streams end up on the same instance once
it is up and running again. For that, the tokens are written to a file
that can be specified via the `-ingester.tokens-file-path` argument.

In certain cases, however, you want to forget the tokens and
re-distribute them when shutting down an ingester instance. This was
already possible by calling `/ingester/flush_shutdown`, deleting the
tokens file and terminating the instance. The new handler
`/ingester/shutdown_and_forget` combines these manual steps into a
single handler.

Fixes #<issue number>

The first commit with the vendor changes can be removed once the PR against [grafana/dskit](https://github.com/grafana/dskit/pull/167) has been merged.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


docs/sources/api/_index.md
```diff
@@ -47,6 +47,7 @@ These endpoints are exposed by the ingester:
 
 - [`POST /flush`](#post-flush)
 - [`POST /ingester/flush_shutdown`](#post-ingesterflush_shutdown)
+- [`POST /ingester/shutdown_and_forget`](#post-ingestershutdown_and_forget)
```
sandeepsukhani, line 50: `shutdown_and_forget` doesn't make it clear what it would forget. What do you think about calling it `shutdown_and_forget_ring_tokens` or `shutdown_and_remove_ring_tokens`?
We can also add a query param to `flush_shutdown` like `/ingester/flush_shutdown?forget_ring_tokens=true`.

pkg/ingester/ingester.go
```diff
@@ -506,6 +508,12 @@ func (i *Ingester) stopping(_ error) error {
 	}
 	i.flushQueuesDone.Wait()
 
+	// In case the flag to clear tokens on shutdown is set we need to mark the
+	// ingester service as "failed", so Loki will shut down entirely.
+	// The module manager logs the failure `modules.ErrStopProcess` in a special way.
+	if i.lifecycler.ClearTokensOnShutdown() && errs.Err() == nil {
+		return modules.ErrStopProcess
```
sandeepsukhani, line 515: Why can't we just rely on `nil` error? We can return internal server error from `ShutdownAndForgetHandler` handler when error is non-nil or am I missing something?

sandeepsukhani (commented):
The changes look great. Just left some comments.

pkg/storage/stores/series/series_index_store_test.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func defaultLimits() (*validation.Overrides, error) {
+	var defaults validation.Limits
+	flagext.DefaultValues(&defaults)
+	defaults.CardinalityLimit = 5
+	return validation.NewOverrides(defaults, nil)
+}
+
+func BenchmarkIndexStore(b *testing.B) {
```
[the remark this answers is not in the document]
chaudum: Totally irrelevant. :thinking: I have no idea how this change found its way into the PR

docs/sources/api/_index.md
chaudum: On the one hand I like the idea of a single endpoint, but on the other hand, I find it weird to have query parameters in URLs that only allow `POST` requests.

chaudum: It would make sense then to change the URL to `/ingester/shutdown` and make both `flush` and `delete_tokens` URL parameters.

```
/ingester/shutdown?flush=true
/ingester/shutdown?delete_tokens=true
```

(Of course keep the existing URL and deprecate it.)

What do you think @sandeepsukhani ?

pkg/ingester/ingester.go
chaudum: No, because we need to be able to distinguish between stopping of the ingester service that happened through the signal handler, and stopping of the service that was triggered by the shutdown handler. In case of the former, the error will be `nil` as well.

docs/sources/api/_index.md
dannykopping: What about `/ingester/shutdown/?unregister=true`?

Keep in mind as well that documentation can clarify the purpose and mechanism for an API endpoint, if need be.

sandeepsukhani: > I find it weird to have query parameters in URLs that only allow POST requests
I agree it looks ugly but it helps us keep it tidy. If we keep adding different paths for various use cases, it would start getting messier.

I think `delete_ring_tokens` or `reset_ring_tokens` or something similar would be a bit self-descriptive.

pkg/ingester/ingester.go
sandeepsukhani: In both cases, we are returning `w.WriteHeader(http.StatusNoContent)` so I don't see any special handling in `ShutdownAndForgetHandler` for either of the cases. Am I missing something?

chaudum:
@sandeepsukhani @dannykopping I opted for `/ingester/shutdown` and restructured the handler a bit.

pkg/loki/modules.go
```diff
@@ -341,9 +341,15 @@ func (t *Loki) initIngester() (_ services.Service, err error) {
 	httpMiddleware := middleware.Merge(
 		serverutil.RecoveryHTTPMiddleware,
 	)
-	t.Server.HTTP.Path("/flush").Methods("GET", "POST").Handler(httpMiddleware.Wrap(http.HandlerFunc(t.Ingester.FlushHandler)))
-	t.Server.HTTP.Methods("POST").Path("/ingester/flush_shutdown").Handler(httpMiddleware.Wrap(http.HandlerFunc(t.Ingester.ShutdownHandler)))
-
+	t.Server.HTTP.Methods("GET", "POST").Path("/flush").Handler(
```
dannykopping, line 344: Howcome this isn't `/ingester/flush`? I see that it wasn't before, but we should probably namespace it like the others for consistency

docs/sources/api/_index.md
```diff
@@ -805,13 +806,31 @@ In microservices mode, the `/flush` endpoint is exposed by the ingester.
...
+endpoint, but accepts two URL query parameters `flush` and `delete_ring_tokens`.
+
+**URL query parameters:**
+
+* `flush=<bool>`:
+  Flag to control whether to flush any in-memory chunks the ingester holds. Defaults to `true`.
+* `delete_ring_tokens=<bool>`:
+  Flag to control whether to delete the file that contains the ingester ring tokens of the instance if the `-ingester.token-file-path` is specified.
+
+This handler, in contrast to the `/ingester/flush_shutdown` handler, terminates the Loki process.
```
dannykopping, line 829: I'm curious why we don't expose a flag for the shutdown - can you help me understand pls?

pkg/ingester/ingester.go
```diff
@@ -526,10 +535,16 @@ func (i *Ingester) loop() {
...
 //     * Change the state of ring to stop accepting writes.
 //     * Flush all the chunks.
-func (i *Ingester) ShutdownHandler(w http.ResponseWriter, r *http.Request) {
+// Note: This handler does not trigger a termination of the Loki process,
+// despite its name. Instead, the ingester service is stopped, so an external
+// source can trigger a safe termination through a signal to the process.
+// The handler is deprecated and usage is discouraged. Use ShutdownHandler
+// instead.
+func (i *Ingester) LegacyShutdownHandler(w http.ResponseWriter, r *http.Request) {
+	level.Warn(util_log.Logger).Log("msg", "This handler is deprecated and usage is discouraged. Please use /ingester/shutdown?flush=true instead.")
```
dannykopping, line 547: This message doesn't mention which handler is deprecated

pkg/ingester/ingester.go
```diff
@@ -538,6 +553,36 @@ func (i *Ingester) ShutdownHandler(w http.ResponseWriter, r *http.Request) {
...
+	err := services.StopAndAwaitTerminated(context.Background(), i)
+	// Stopping the module will return the modules.ErrStopProcess error. This is
+	// needed so the Loki process is shut down completely.
+	if err == nil || err == modules.ErrStopProcess {
+		w.WriteHeader(http.StatusNoContent)
+	} else {
+		w.WriteHeader(http.StatusInternalServerError)
+		_, _ = w.Write([]byte(err.Error()))
+	}
+}
```
dannykopping, line 584: `<nit>`I would prefer that this function doesn't do any HTTP work; that's the responsibility of the calling function`</nit>`

docs/sources/api/_index.md
chaudum: Not sure what you exactly mean with that.
* Adding a flag to the existing `/ingester/flush_shutdown` handler to allow the process to be terminated?
* Or having a flag on `/ingester/shutdown` to allow either stopping the service only, or terminating the process completely?

IMO, "shutdown" suggests that the process is terminated, that's why I wanted to keep the behaviour of the old handler but deprecate it, and the new handler _always_ terminates the process (which I think is almost always what you want).

pkg/ingester/ingester.go
chaudum: True :see_no_evil:

docs/sources/api/_index.md
dannykopping: I think it's the "almost always" there that's making me hesitate. We're changing behaviour now, which some users may rely on. I think we need to maintain the ability to shutdown the ingester but not terminate; it can default to `true` if this is what we recommend.

chaudum:
I'll squash commits once approved.

dannykopping:
> I'll squash commits once approved.

No need, we squash on merge

pkg/ingester/ingester.go
```diff
@@ -556,10 +562,16 @@ func (i *Ingester) LegacyShutdownHandler(w http.ResponseWriter, r *http.Request)
 // ShutdownHandler handles a graceful shutdown of the ingester service and
 // termination of the Loki process.
 func (i *Ingester) ShutdownHandler(w http.ResponseWriter, r *http.Request) {
+	// Don't allow calling the shutdown handler multiple times
+	if i.State() != services.Running {
+		w.WriteHeader(http.StatusNotFound)
+		_, _ = w.Write([]byte("Ingester is stopping or already stopped."))
```
dannykopping, line 567: Should we `return` here?

dannykopping: I'd also argue `404 Not Found` is not the right status code here. Maybe `410 Gone` is better.

chaudum: Indeed :man_facepalming:

chaudum: I was thinking about `410` as well, but I'm not sure if it's better either.
Yes, for this specific instance the endpoint is gone, but after restart, it is available again, making it only temporarily unavailable.
> If you don't know whether this condition is temporary or permanent, a [404](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404) status code should be used instead.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410

dannykopping: In our API, `404` is mainly used (AFAIK) when an API endpoint doesn't exist, so I think 404 would be a confusing choice here even if it's the most correct.

docs/sources/api/_index.md
```diff
@@ -805,13 +806,34 @@ In microservices mode, the `/flush` endpoint is exposed by the ingester.
...
 
 In microservices mode, the `/ingester/flush_shutdown` endpoint is exposed by the ingester.
 
-### `GET /distributor/ring`
+## `POST /ingester/shutdown`
+
+`/ingester/shutdown` is similar to the [`/ingester/flush_shutdown`](#post-ingesterflush_shutdown)
+endpoint, but accepts three URL query parameters `flush`, `delete_ring_tokens`, and `terminate`.
+
+**URL query parameters:**
```
trevorwhitney, line 822: since this is a POST, curious why we opt for url params intead of a request body?

trevorwhitney (commented):
Just did a quick review, but looks good!

pkg/ingester/ingester.go
chaudum: I thought about it a bit more and I think 503 is more accurate, because the server is not ready to handle the request.

dannykopping: That works for me too :+1:

docs/sources/api/_index.md
chaudum: Good point. [We've discussed this a bit](https://github.com/grafana/loki/pull/6179#discussion_r876827482). Query params are I guess easier to write when you execute the request using `curl`.

chaudum: Also, I think we do not follow any strict, restful HTTP API design guides in our endpoint design.

dannykopping (approved):
LGTM, nice work!

trevorwhitney (approved):
LGTM!

PR #6181, opened 2022-05-17
strips out tenant label from LabelNames queries

owen-d:
This PR prevents sending the internal tenant label used in `LabelNames` queries. This label is only used to pack multiple tenants into a single TSDB index until we compact to multitenant ones.

### DIFF
--- a/pkg/storage/stores/tsdb/multitenant.go
+++ b/pkg/storage/stores/tsdb/multitenant.go
@@ -2,6 +2,7 @@ package tsdb
 
 import (
 	"context"
+	"sort"
 
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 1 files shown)


pkg/storage/stores/tsdb/multitenant.go
```diff
@@ -66,7 +67,18 @@ func (m *MultiTenantIndex) Series(ctx context.Context, userID string, from, thro
...
+	// Strip out the tenant label in response.
+	i := sort.SearchStrings(res, TenantLabel)
+	if i == len(res) || res[i] != TenantLabel {
+		return res, nil
+	}
+
+	return append(res[:i], res[i+1:]...), nil
 }
 
 func (m *MultiTenantIndex) LabelValues(ctx context.Context, userID string, from, through model.Time, name string, matchers ...*labels.Matcher) ([]string, error) {
```
DylanGuedes, line 84: double checking: shouldn't we have a similar check for `labelValues`?

owen-d: Good question!
I left it out for speed benefits and because it's unlikely to be queried as it's never returned by the `series` or `labelnames` calls. It should be possible to get a response for this value by querying `labelvalues(LokiTenantLabel)`, but it's unusable outside of that since it's a synthetic label that's only used for packing multiple tenants into a single TSDB pre-compaction.

It _is_ technically correct to remove it though :)

owen-d: followup: it's probably negligible in latency benefits, so I should remove it. Will do!

sandeepsukhani (commented):
LGTM! Let us handle it for `LabelValues` as well for consistency!

PR #6186, opened 2022-05-18
security: Redact credentials when marshalled to YAML

### LINKED ISSUE #6184: Secrets leak through the config endpoint
**Describe the bug**
Sensitive information like aws `access_key_id` and `secret_access_key` are listed in plain text when calling the `[GET] /config` endpoint. It seems to be a security violation for most companies. 

**Expected behavior**
They are type of `secret` in code, so I guess they can be masked as ******* when displayed. Or, maybe add an option to disable the `/config` endpoint.

chaudum:
## What this PR does / why we need it

**security: Redact BOS credentials when marshalled to YAML**

Change the type of the credentials flags for the BOS client configuration from string to flagext.Secret, which automatically redacts to `********` when marshalled to YAML.
    
**Use flagext.Secret for S3 object client credentials**

instead of a package private, similar implementation of a secret type flag. The redacted output of the value changes from `redacted` to `********`.

## Which issue(s) this PR fixes
Fixes #6184

## Special notes for your reviewer

No changelog entry because the BOS storage feature is unreleased.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/storage/chunk/client/aws/s3_storage_client.go
+++ b/pkg/storage/chunk/client/aws/s3_storage_client.go
@@ -69,7 +69,7 @@ type S3Config struct {
 	BucketNames      string
 	Endpoint         string              `yaml:"endpoint"`
 	Region           string              `yaml:"region"`
-	AccessKeyID      flagext.Secret      `yaml:"access_key_id"`
+	AccessKeyID      string              `yaml:"access_key_id"`
 	SecretAccessKey  flagext.Secret      `yaml:"secret_access_key"`
 	Insecure         bool                `yaml:"insecure"`
 	SSEEncryption    bool                `yaml:"sse_encryption"`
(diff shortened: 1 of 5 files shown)


pkg/storage/chunk/client/aws/s3_storage_client.go
```diff
@@ -85,8 +69,8 @@ type S3Config struct {
 	BucketNames      string
 	Endpoint         string              `yaml:"endpoint"`
 	Region           string              `yaml:"region"`
-	AccessKeyID      secret              `yaml:"access_key_id"`
-	SecretAccessKey  secret              `yaml:"secret_access_key"`
+	AccessKeyID      flagext.Secret      `yaml:"access_key_id"`
```
simonswine, line 72: I would like us to stay consistent with Mimir and Tempo and not consider AccessKeyID as a secret.

https://github.com/grafana/mimir/blob/c248b4e81e8fed041b188749719bb8167aa77a1f/pkg/storage/bucket/s3/config.go#L78

https://github.com/grafana/tempo/blob/4e0f12da3cd2d665e5933bb402b06ca66efe5d7a/tempodb/backend/s3/config.go#L13

(It is actually called `access_key` there :sob:)

pkg/storage/chunk/client/baidubce/bos_storage_client.go
```diff
@@ -36,10 +37,10 @@ func init() {
 }
 
 type BOSStorageConfig struct {
```
simonswine, line 39: Nit: It might be worth also adding a check like we have for S3 redacting / parsing

simonswine (approved):
Thanks for fixing this so quickly.

LGTM, one non-blocking nit

kavirajk (approved):
👍

PR #6192, opened 2022-05-18
Update delete integration test

MichelHollands:

Update the deletion integration test so it waits until the delete request has been processed. Note that at the moment no lines are removed due to flush not working. This will be fixed later on in a separate PR.
Add a metric to show how many lines are deleted.
Remove mention of Cortex in a helper function.
Change the various timeouts so the test runs in 10 seconds.
Add code to parse and check metrics.

**Checklist**
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/integration/cluster/cluster.go
+++ b/integration/cluster/cluster.go
@@ -301,7 +301,8 @@ func (c *Component) run() error {
 		defer c.cluster.waitGroup.Done()
 		err := c.loki.Run(loki.RunOpts{})
 		if err != nil {
-			errCh <- err
+			newErr := fmt.Errorf("error starting component %v: %w", c.name, err)
+			errCh <- newErr
 		}
 	}()
 
--- a/integration/loki_micro_services_delete_test.go
+++ b/integration/loki_micro_services_delete_test.go
@@ -81,7 +81,7 @@ func TestMicroServicesDeleteRequest(t *testing.T) {
 
 		// TODO: Flushing is currently causing a panic, as the boltdb shipper is shared using a global variable in:
 		// https://github.com/grafana/loki/blob/66a4692423582ed17cce9bd86b69d55663dc7721/pkg/storage/factory.go#L32-L35
-		//require.NoError(t, cliIngester.Flush())
+		// require.NoError(t, cliIngester.Flush())
 	})
 
 	t.Run("ingest-logs-ingester", func(t *testing.T) {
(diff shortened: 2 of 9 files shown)


integration/loki_micro_services_delete_test.go
```diff
@@ -115,4 +117,49 @@ func TestMicroServicesDeleteRequest(t *testing.T) {
 		require.Equal(t, `{job="fake"} |= "lineB"`, deleteRequests[0].Query)
 		require.Equal(t, "received", deleteRequests[0].Status)
 	})
+
+	// Wait until delete request is finished
+	t.Run("wait-until-delete-request-processed", func(t *testing.T) {
+		require.Eventually(t, func() bool {
+			deleteRequests, err := cliCompactor.GetDeleteRequests()
+			require.NoError(t, err)
+			require.NotEmpty(t, deleteRequests)
```
MasslessParticle, line 126: Nit: `require.NotEmpty` is redundant if we're `require.Len(t,...,1)` on the next line

MasslessParticle (approved):
This looks awesome

integration/loki_micro_services_delete_test.go
MichelHollands: Indeed. That check has been removed.

integration/loki_micro_services_delete_test.go
```diff
@@ -115,4 +117,48 @@ func TestMicroServicesDeleteRequest(t *testing.T) {
...
+		assert.Equal(t, "streams", resp.Data.ResultType)
+
+		var lines []string
+		for _, stream := range resp.Data.Stream {
+			for _, val := range stream.Values {
+				lines = append(lines, val[1])
+			}
+		}
+		// Remove lineB once flush works
+		assert.ElementsMatch(t, []string{"lineA", "lineB", "lineC", "lineD"}, lines)
```
trevorwhitney, line 151: when we get to the point that `lineB` is actually removed, does `ElementsMatch` accept a 4th argument for a "reason" or "comment" string? If so I'd like to move the comment on 138 into there so that's printed out for the user when the test fails WDYT?

trevorwhitney (approved):
This looks great! LGTM

integration/loki_micro_services_delete_test.go
MichelHollands: Good idea. It has been added.

integration/cluster/cluster.go
```diff
@@ -301,7 +301,8 @@ func (c *Component) run() error {
 		defer c.cluster.waitGroup.Done()
 		err := c.loki.Run(loki.RunOpts{})
 		if err != nil {
-			errCh <- err
+			newErr := fmt.Errorf("error starting component %v: %w", c.name, err)
+			errCh <- newErr
```
chaudum, line 305: nit: since `newErr` is used nowhere else ...

```suggestion
			errCh <- fmt.Errorf("error starting component %v: %w", c.name, err)
```

chaudum (approved):

owen-d (approved):

PR #6194, opened 2022-05-18
Fix delete updates

MasslessParticle:
Previously the querier got deletes from the index gateway.  The `delete_requests` table was never updated because the component assumes that filenames never change once written (which is how the index files work). Rather than changing the index gateway's assumption about indicies or carving out an exception, this PR makes the querier get delete requests from the delete API itself. That way, there is only one source of truth for deletes and everything just pulls from there.

This PR also changes the location of the `compactor_address` configuration. Rather than have it on each component that needs it, the address is now part of the compactor config where any component can just grab it.


pkg/storage/stores/shipper/compactor/compactor.go
```diff
@@ -87,6 +88,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.DeleteRequestCancelPeriod, "boltdb.shipper.compactor.delete-request-cancel-period", 24*time.Hour, "Allow cancellation of delete request until duration after they are created. Data would be deleted only after delete requests have been older than this duration. Ideally this should be set to at least 24h.")
 	f.IntVar(&cfg.MaxCompactionParallelism, "boltdb.shipper.compactor.max-compaction-parallelism", 1, "Maximum number of tables to compact in parallel. While increasing this value, please make sure compactor has enough disk space allocated to be able to store and compact as many tables.")
 	f.StringVar(&cfg.DeletionMode, "boltdb.shipper.compactor.deletion-mode", "whole-stream-deletion", fmt.Sprintf("(Experimental) Deletion mode. Can be one of %v", strings.Join(deletion.AllModes(), "|")))
+	f.StringVar(&cfg.Address, "boltdb.shipper.compactor.address", "", "host and port where the compactor API is listening")
```
sandeepsukhani, line 91: Adding this config to the compactor looks out of place to me. New users already get confused by our configs, so it is better to duplicate it a little than add to the confusion. It could be just me. What do you think?

pkg/storage/stores/shipper/compactor/deletion/delete_requests_client.go
```diff
@@ -0,0 +1,174 @@
...
+
+type DeleteRequestsStoreOption func(c *deleteRequestsClient)
+
+func WithRequestClientCacheDuration(d time.Duration) DeleteRequestsStoreOption {
+	return func(c *deleteRequestsClient) {
+		c.cacheDuration = d
+	}
+}
+
+func NewDeleteRequestsClient(addr string, c doer, opts ...DeleteRequestsStoreOption) (DeleteRequestsClient, error) {
```
sandeepsukhani, line 50: Instead of taking an HTTP client as a parameter, what do you think about accepting a config?

sandeepsukhani (commented):
Overall the changes look great to me. I just added 2 comments. Irrelevant to this PR, I also think we should add an in-memory cache for `CacheGenClient` same as `DeleteRequestsClient`.

pkg/storage/stores/shipper/compactor/deletion/delete_requests_client.go
```diff
@@ -0,0 +1,174 @@
...
+	req.Header.Set(orgHeaderKey, userID)
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		level.Error(log.Logger).Log("msg", "error getting delete requests from the store", "err", err)
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
MichelHollands, line 161: This can be written like this as well:

```suggestion
	if resp.StatusCode/100 != 2 {
```

pkg/storage/stores/shipper/compactor/compactor.go
MasslessParticle: I went back and forth on this. If we leave it off the compactor config, it becomes a required property on:
- frontend
- querier
- ruler

Do you think that's clearer for the user than having the compactor know it's own address?

pkg/storage/stores/shipper/compactor/deletion/delete_requests_client.go
MasslessParticle: Nice, committed

pkg/storage/stores/shipper/compactor/deletion/delete_requests_client.go
MasslessParticle: I'd prefer to take the client, if it's not too disruptive. I really like this pattern because it makes if very easy to unit test without worrying about the network: [pkg/storage/stores/shipper/compactor/deletion/delete_requests_client_test.go](https://github.com/grafana/loki/pull/6194/files/51d1535c4d9adf786c76e59f1352dd1d1b6a28dd#diff-1f659c597f41d7bd102f9f32644059a6172892ca7905eb176b0972386f5ed3fe)

pkg/loki/modules.go
```diff
@@ -951,16 +959,16 @@ func (t *Loki) deleteRequestsStore() (deletion.DeleteRequestsStore, error) {
 		return nil, err
 	}
 
-	deleteStore := deletion.NewNoOpDeleteRequestsStore()
-	if config.UsingBoltdbShipper(t.Cfg.SchemaConfig.Configs) && filteringEnabled {
```
sandeepsukhani, line 955: This check needs to stay, right?

pkg/storage/stores/shipper/compactor/compactor.go
trevorwhitney: is there also a case to be made for the `common` config, since it's common across multiple components?

pkg/storage/stores/shipper/compactor/deletion/delete_requests_client.go
```diff
@@ -0,0 +1,174 @@
...
+	httpClient doer
+	mu         sync.RWMutex
+
+	cache         map[string][]DeleteRequest
+	cacheDuration time.Duration
+
+	stopChan chan struct{}
+}
+
+type doer interface {
```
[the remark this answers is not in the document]
trevorwhitney: agreed that `doer` is not a very clear name.

pkg/storage/stores/shipper/compactor/compactor.go
MasslessParticle: Common Config says `// Values defined under this common configuration are supersede if a more specific value is defined.`

Does that imply that there's also some more specific override elsewhere? If not, it would be a fine place for this to go.

pkg/storage/stores/shipper/compactor/deletion/delete_requests_client.go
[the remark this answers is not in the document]
MasslessParticle: I'm happy to change this. It's an attempt to follow Golang interface naming conventions: https://go.dev/doc/effective_go#interface-names

pkg/storage/stores/shipper/compactor/compactor.go
trevorwhitney: I don't think that would apply in this case, if we moved it to the common config so that was the only place it's defined, then there would be no superseding. What that comment is meant to convey is that after things from the common config (like storage or ring configs) are applied to the many places those are used, the `config.yaml` is read again, so if there is say, a specific ring config for the scheduler, that will supersede whatever was put in common.

sandeepsukhani: yeah, I thought about using `common` config but due to that comment I thought it was not the right place, same as Travis. If we can use it for this use case then it would be a better option.

docs/sources/configuration/_index.md
```diff
@@ -2208,6 +2204,10 @@ The `limits_config` block configures global and per-tenant limits in Loki.
 # CLI flag: -store.max-query-length
 [max_query_length: <duration> | default = 721h]
 
+# Address, including port, where the compactor api is served
+# CLI flag: -store.compactor-address
+[compactor_address: <string> | default = ""]
+
```
sandeepsukhani, line 2210: It seems the documentation has been added under `limits_config` instead of `storage_config`.

sandeepsukhani (commented):
Left very minor suggestions. Let us see if we want to move `compactor_address` to `common` config.

sandeepsukhani (approved):
LGTM

PR #6196, opened 2022-05-19
doc(storage-example-configs): AWS single store configs

kavirajk:
Our example configs are bit misleading here. Particularly AWS deployment.
https://grafana.com/docs/loki/latest/storage/#examples

As a new user, I come here and all I can conclude is, we cannot use s3 as a single store under AWS deployment. Doc seems to recommend AWS deployment always  needs S3 + DynamoDB.
(in fact it was real confusion for new user from the community)

https://grafana.slack.com/archives/CEPJRLQNL/p1652867888720979

Fixes #<issue number>

**Checklist**
- [x] Documentation added

### DIFF
--- a/docs/sources/storage/_index.md
+++ b/docs/sources/storage/_index.md
@@ -213,7 +213,7 @@ storage_config:
 schema_config:
   configs:
     - from: 2020-07-01
-      store: aws
+      store: boltdb-shipper
       object_store: aws
       schema: v11
       index:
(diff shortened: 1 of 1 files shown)


sandeepsukhani (changes requested):
Thanks for fixing it! One minor change, `store` under `schema_config.configs` needs to be update to `boltdb-shipper`

sandeepsukhani (approved):
LGTM

PR #6203, opened 2022-05-19
creates a single tsdb for entire wal recovery

owen-d:
This simplifies our TSDB WAL replay to build a single TSDB rather than a TSDB per rotation period (15m) on restart. It also fixes a bug during replay where the WALs for a new rotation period didn't include the series reference (likely in a previous period).

It also builds TSDBs during replay with un-rounded timestamps to avoid clobbering in the case of multiple replays (crashlooping) in the same rotation period (15m)


sandeepsukhani (approved):
LGTM

Issue #6202, opened 2022-05-19
Query returns unimplemented error when nesting `avg(avg_over_time())`


lmickh:
**Describe the bug**
After upgrading from 2.3 to 2.5, this query will result in an `unimplemented` error:
```logql
avg(avg_over_time({job=~"myapps.*"} |= "stats" | json busy="utilization" | unwrap busy [$__interval]))
```

The `utilization` field being extracted is a float (ex: `55.4`).  The `avg_over_time()` appears to properly calculate the values for each individual stream, but when wrapped with a `avg()` to get the average across all of the streams the `unimplemented` error is returned instead. 

**To Reproduce**
1. Started Loki v2.5.0
2. Started Promtail v2.5.0 to tail Kubernetes pods
3. Query via Grafana v8.4.5: `avg(avg_over_time{job="somejob" | json somefield="field" | unwrap somefield [$__interval]))`

**Expected behavior**
While wrapping an `avg_over_time()` with a `avg()` is probably not as good as `avg_over_time() by (label)`, there is not an obvious reason why it should not work.  I would expect an unwrapped float per log stream could be aggregated with `avg()`.

**Environment:**
 - Infrastructure: EKS
 - Deployment tool: Raw manifests via ArgoCD

**Work-around**

Using `avg_over_time(...) by (label)` to aggregate instead of wrapping it with `avg()`.

This might be related to #6130 and the changes in 2.4, but I haven't had a chance to test that yet.

owen-d:
Confirmed, this is in `main` as well as `2.5` (I'm pretty sure it's been around for a long time).

PR #6204, opened 2022-05-19
prevents sharding avg and avg_over_time for unshardable children

owen-d:
This fixes a bug in query planning which didn't check if the child-expressions of `avg` were shardable, which led to some `unimplemented` errors on the frontend. It's pretty unlikely to run into this (i.e. `avg(avg_over_time({job=~"myapps.*"} |= "stats" | json busy="utilization" | unwrap busy [5m]))`).

I've run this in a dev cluster as well, everything looking :+1: .

In the future, we can consider mapping `avg_over_time` -> `sum_over_time / count_over_time` :thinking: 

Closes https://github.com/grafana/loki/issues/6202

### DIFF
--- a/pkg/logql/syntax/ast.go
+++ b/pkg/logql/syntax/ast.go
@@ -909,12 +909,19 @@ func (e *VectorAggregationExpr) String() string {
 // impl SampleExpr
 func (e *VectorAggregationExpr) Shardable() bool {
 	if e.Operation == OpTypeCount || e.Operation == OpTypeAvg {
-		// count is shardable is labels are not mutated
+		if !e.Left.Shardable() {
+			return false
+		}
+		// count is shardable if labels are not mutated
 		// otherwise distinct values can be counted twice per shard
 		shardable := true
-		e.Walk(func(e interface{}) {
+		e.Left.Walk(func(e interface{}) {
 			switch e.(type) {
-			case *LabelParserExpr, LabelFmtExpr:
+			// LabelParserExpr is normally shardable, but not in this case.
+			// TODO(owen-d): I think LabelParserExpr is shardable
+			// for avg, but not for count. Let's refactor to make this
+			// cleaner. For now I'm disallowing sharding on both.
+			case *LabelParserExpr:
 				shardable = false
 			}
 		})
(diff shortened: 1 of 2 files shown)


pkg/logql/syntax/ast.go
```diff
@@ -909,12 +909,19 @@ func (e *VectorAggregationExpr) String() string {
...
 		shardable := true
-		e.Walk(func(e interface{}) {
+		e.Left.Walk(func(e interface{}) {
 			switch e.(type) {
-			case *LabelParserExpr, LabelFmtExpr:
+			// LabelParserExpr is normally shardable, but not in this case.
+			// TODO(owen-d): I think LabelParserExpr is shardable
+			// for avg, but not for count. Let's refactor to make this
+			// cleaner. For now I'm disallowing sharding on both.
+			case *LabelParserExpr:
```
ssncferreira, line 924: Should we remove the `LabelFmtExpr` case?
```suggestion
			case *LabelParserExpr, LabelFmtExpr:
```

ssncferreira (commented):
Overall LGTM, just have a couple of questions ❓

pkg/logql/syntax/ast.go
```diff
@@ -909,12 +909,19 @@ func (e *VectorAggregationExpr) String() string {
...
 	if e.Operation == OpTypeCount || e.Operation == OpTypeAvg {
-		// count is shardable is labels are not mutated
+		if !e.Left.Shardable() {
+			return false
+		}
+		// count is shardable if labels are not mutated
 		// otherwise distinct values can be counted twice per shard
 		shardable := true
-		e.Walk(func(e interface{}) {
+		e.Left.Walk(func(e interface{}) {
```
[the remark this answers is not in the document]
owen-d: It won't actually matter in this case (`e.Left.Walk` only walks over the child nodes as we already know what the current node is). In an earlier revision I was playing with it caused an infinite loop as I called `e.Shardable()` within the walk function -- not a good idea :sweat_smile: . I ended up putting the `e.Left.Shardable()` check outside the walk function because we only need call it once there rather than once for each child node recursively.

pkg/logql/syntax/ast.go
owen-d: `LabelFmtExpr` already returns `false` for `Shardable()`, so it was simpler to remove it.

kavirajk (approved):
LGTM :+1:

PR #6214, opened 2022-05-20
Runtime reloadable config; ring migration setup

cstyan:
This adds the few missing pieces of multi KV runtime reloadable config to loki components that use the ring, based on Peters mimir PR [here](https://github.com/grafana/mimir/issues/996). Note that a lot of config was necessary to get the test to pass (many loki components will fail to start/be registered via the modules setup without things like storage config).

cc @JordanRushing @DylanGuedes @pstibrany


slim-bean (approved):
LGTM

PR #6216, opened 2022-05-21
Reduce to a single code path through sendBatches regardless of limit

splitice:

DRY up send batches

Fixes #<issue number>

This should have no great effect on performance. Primarily this makes the code more maintainable.

It's likely not measurable but there is likely a minute cost for the path that had no limit in the past, but this would be absolutely minute.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


splitice:
This function could be made even simpler if I changed the signature to accept limit as a signed int (where -1 was unlimited, not 0)

Thoughts?

owen-d (approved):
LGTM, nice catch!

PR #6217, opened 2022-05-21
[RFC] Feature/ingester stream backpressure

### LINKED ISSUE #5804: [bug] ingester goroutine leak
**Describe the bug**
ingester goroutine leak. I am trying to reproduce this problem
**To Reproduce**
1. Started Loki (SHA or version)
2. Started Promtail (SHA or version) to tail '...'
3. Query: `{} term`

**Expected behavior**

**Environment:**
 - Infrastructure: [e.g., Kubernetes, bare-metal, laptop]
 - Deployment tool: [e.g., helm, jsonnet]

image

splitice:

This PR add backpressure to the response to ingester `Query()`. A similar aproach could also be used for `QuerySample()`

This is done through the addition of an `Ack(id)` method and a limit to the max in-flight responses in the stream (currently 20)

This is built upon my branch for #6216

I suspect would fix #5804

This needs a cleanup and represents a RFC on the concept which it will get if positively reviewed.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


pkg/ingester/instance.go
```diff
@@ -695,51 +699,44 @@ type QuerierQueryServer interface {
 	Send(res *logproto.QueryResponse) error
 }
 
-func sendBatches(ctx context.Context, i iter.EntryIterator, queryServer QuerierQueryServer, limit uint32) error {
+func sendBatches(ctx context.Context, i iter.EntryIterator, queryServer QuerierQueryServer, q *IngesterQuery, limit int32) error {
```
splitice, line 702: I'm tempted to make this `SendBatches` as a member function of IngesterQuery.

pkg/ingester/ingester.go
```diff
@@ -569,6 +569,27 @@ func (i *Ingester) GetOrCreateInstance(instanceID string) *instance { //nolint:r
...
+	instance := i.GetOrCreateInstance(instanceID)
+	if err != nil {
+		return &logproto.AckResponse{}, err
+	}
+
+	instance.queryMtx.Lock()
+	queryIngester := instance.queries[req.Id]
+	instance.queryMtx.Unlock()
+	if queryIngester != nil {
+		queryIngester.ReleaseAck()
```
DylanGuedes, line 587: hmm nitpicking maybe you should call it `SendAck()` instead of `ReleaseAck()`? (since SendAck will send an empty struct message to the channel)

pkg/ingester/ingester_query.go
```diff
@@ -0,0 +1,130 @@
...
+type QuerierQueryServer interface {
+	Context() context.Context
+	Send(res *logproto.QueryResponse) error
+}
+
+func (q *IngesterQuery) SendBatches(ctx context.Context, i iter.EntryIterator, queryServer QuerierQueryServer, limit int32) error {
+	stats := stats.FromContext(ctx)
+
+	// send until the limit is reached.
+	for limit != 0 && !isDone(ctx) {
```
DylanGuedes, line 64: can the limit be negative? if so, I think it should be limit <= 0 instead

DylanGuedes (commented):
a few (not so important) things. Just for curiosity, what would the backpressure improve? and since you used the term RFC, do you have a doc or anything explaining the reasoning behind the idea?

pkg/ingester/ingester_query.go
splitice: if it is negative it's an error. Theres no precedent of assertions in loki that I've found so alas it's a silent failure.

splitice:
I dare say between #6241 and this that these are the two main causes of ingester memory peaks. Whether or not this is the direction loki team want to go is worthy of comment.

splitice:
Go away

splitice:
Go away

MasslessParticle:
Hey @splitice, thanks for your contribution! 

We've added a new way to get visibility on issues like this. Please check out [LIDs](https://grafana.com/docs/loki/next/lids/). I'm going to close this for now. If you'd still like to discuss it, please open a lid.

PR #6229, opened 2022-05-23
Remove __name__ from label calls

MasslessParticle:
`__name__` wasn't being being applied to `Label` calls when it's not meaningful for user. This PR always filters `__name__` from Label calls.


owen-d (approved):
LGTM

PR #6233, opened 2022-05-24
Tsdb/inverted index

owen-d:
In order to maintain query consistency, we must ensure the same shard (i.e. `1_of_16`) corresponds to the same series. Since TSDB calculates shard inclusion differently (by using bit prefixes instead of modulos), we need to account for this change when resolving streams in the ingester's inverted index.

This is the first PR which doesn't use the new code paths.
Also, because the bit prefixed inverted index will support dynamic sharding of higher _or_ lower shard factors, it occasionally needs to do additional computation when translating higher granularity shard factors to lower granularity ones (i.e. 16 -> 4).

### DIFF
--- a/pkg/ingester/index/bitprefix_test.go
+++ b/pkg/ingester/index/bitprefix_test.go
@@ -5,12 +5,13 @@ import (
 	"sort"
 	"testing"
 
-	"github.com/grafana/loki/pkg/logproto"
-	"github.com/grafana/loki/pkg/querier/astmapper"
-	"github.com/grafana/loki/pkg/storage/stores/tsdb/index"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/loki/pkg/logproto"
+	"github.com/grafana/loki/pkg/querier/astmapper"
+	"github.com/grafana/loki/pkg/storage/stores/tsdb/index"
 )
 
 func Test_BitPrefixGetShards(t *testing.T) {
--- a/pkg/ingester/index/bitprefix_test.go
+++ b/pkg/ingester/index/bitprefix_test.go
@@ -2,50 +2,198 @@ package index
 
 import (
 	"fmt"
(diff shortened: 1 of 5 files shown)


pkg/ingester/index/bitprefix_test.go
```diff
@@ -0,0 +1,200 @@
...
+	// non factor of 2 shard factor
+	_, err := NewBitPrefixWithShards(6)
+	require.Error(t, err)
+
+	// valid shard factor
+	_, err = NewBitPrefixWithShards(4)
+	require.Nil(t, err)
+}
+
+func Test_BitPrefixDeleteAddLoopkup(t *testing.T) {
```
chaudum, line 72: nit: typo

```suggestion
func Test_BitPrefixDeleteAddLookup(t *testing.T) {
```

pkg/ingester/index/bitprefix_test.go
```diff
@@ -0,0 +1,200 @@
...
+		labels.Label{Name: "compose_project", Value: "loki-boltdb-storage-s3"},
+		labels.Label{Name: "compose_service", Value: "ingester-2"},
+		labels.Label{Name: "container_name", Value: "loki-boltdb-storage-s3_ingester-2_1"},
+		labels.Label{Name: "filename", Value: "/var/log/docker/790fef4c6a587c3b386fe85c07e03f3a1613f4929ca3abaa4880e14caadb5ad1/json.log"},
+		labels.Label{Name: "host", Value: "docker-desktop"},
+		labels.Label{Name: "source", Value: "stderr"},
+	}
+
+	// for _, shard := range []uint32{2, 4, 8, 16, 32, 64, 128} {
+	for _, shard := range []uint32{2} {
```
chaudum, line 102: Is there a reason why you only tested for 2 shards?

chaudum (approved):
LGTM, have a view questions though.

chaudum:
@owen-d I was late to the game and submitted my review only after the PR was already merged. Any chance you could answer the questions if it's not too much effort? :innocent:

PR #4346, opened 2021-09-17
add logfmt promtail stage to be able to extract data from logfmt formatted log

sandeepsukhani:
**What this PR does / why we need it**:
It would be good to have a `logfmt` stage, the same as `json` stage, for extracting data from a log entry. It works the same as `json` stage where the user can specify the mapping for extracted data to define which kvs to extract and set a different key than the one in logs for the extracted data.

**Special notes for your reviewer**:
I would open a flow up PR for adding documentation.

**Checklist**
- [x] Tests updated

PR #5921, opened 2022-04-14
docs: Add logfmt promtail parsing stage

alanmcg:
Fixes #5384

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] Add an entry in the `CHANGELOG.md` about the changes.

PR #6238, opened 2022-05-24
[docs] document logfmt pipeline stage

cstyan:
I noticed earlier while writing a design doc that the `logfmt` pipeline stage is not documented. It was originally added here: https://github.com/grafana/loki/pull/4346

I've also fixed a minor typo in the json pipeline stage docs.

### DIFF
--- a/docs/sources/clients/promtail/stages/logfmt.md
+++ b/docs/sources/clients/promtail/stages/logfmt.md
@@ -1,5 +1,7 @@
 ---
 title: logfmt
+menuTitle: logfmt
+description: The logfmt parsing stage reads logfmt log lines and extracts the data into labels.
 ---
 # `logfmt` stage
 
--- a/docs/sources/clients/promtail/stages/logfmt.md
+++ b/docs/sources/clients/promtail/stages/logfmt.md
@@ -72,7 +72,7 @@ For the given pipeline:
 And the given log line:
 
 ```
-time=2012-11-01T22:08:41+00:00 app=loki level=WARN duration=125 message="this is a log line" extra="user=foo""
+time=2012-11-01T22:08:41+00:00 app=loki level=WARN duration=125 message="this is a log line" extra="user=foo"
 ```
 
 The first stage would create the following key-value pairs in the set of
--- a/docs/sources/clients/promtail/stages/logfmt.md
+++ b/docs/sources/clients/promtail/stages/logfmt.md
@@ -21,12 +21,12 @@ logfmt:
 
 This stage uses the [go-logfmt](https://github.com/go-logfmt/logfmt) unmarshaler, which means non-string types like
 numbers or booleans will be unmarshaled into those types. The extracted data
-can hold non-string values and this stage does not do any type conversions;
+can hold non-string values, and this stage does not do any type conversions;
 downstream stages will need to perform correct type conversion of these values
-as necessary. Please refer to the [the `template` stage](../template/) for how
+as necessary. Please refer to the [`template` stage](../template/) for how
 to do this.
 
-If the value extracted is a complex type it's value is extracted as a string.
+If the value extracted is a complex type, its value is extracted as a string.
 
 ## Examples
 
--- a/docs/sources/clients/promtail/stages/logfmt.md
+++ b/docs/sources/clients/promtail/stages/logfmt.md
@@ -1,0 +1,86 @@
+---
+title: logfmt
+---
+# `logfmt` stage
+
+The `logfmt` stage is a parsing stage that reads the log line as [logfmt](https://brandur.org/logfmt) and allows extraction of data into labels.
+
+## Schema
+
+```yaml
+logfmt:
+  # Set of key/value pairs for mapping of logfmt fieds to extracted labels. The key will be
+  # the key in the extracted data while the expression will be the value. If the value
+  # is empty then the logfmt field with the same name is extracted.
+  mapping:
+    [ <string>: <string> ... ]
+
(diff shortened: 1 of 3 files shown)


kavirajk:
@cstyan duplicate of this https://github.com/grafana/loki/pull/5921.

cyriltovena (approved):
LGTM

DylanGuedes (approved):
looks wonderful, lgtm!

docs/sources/clients/promtail/stages/logfmt.md
```diff
@@ -0,0 +1,86 @@
...
+If the value extracted is a complex type it's value is extracted as a string.
+
+## Examples
+
+### Using log line
+
+For the given pipeline:
+
+```yaml
+- logfmt:
```
DylanGuedes, line 38: Do you mind wrapping this logfmt inside a `pipeline_stages` key?

docs/sources/clients/promtail/stages/logfmt.md
```diff
@@ -0,0 +1,86 @@
...
+    mapping:
+      timestamp: time
+      app:
+      duration:
+      unknown:
```
DylanGuedes, line 43: Do you mind using the `labels` stage in conjunction and adding a disclaimer telling that if not listed under `labels`, extracted labels are to be ignored?

docs/sources/clients/promtail/stages/logfmt.md
```diff
@@ -0,0 +1,86 @@
+---
+title: logfmt
```
DylanGuedes, line 2: Do you mind adding `logfmt` to the list of existing stages in the stages page? (this one: https://grafana.com/docs/loki/latest/clients/promtail/pipelines/)

DylanGuedes (changes requested):
I reviewed it one more time after trying it better, I think now I have a few suggestions.

docs/sources/clients/promtail/stages/logfmt.md
```diff
@@ -0,0 +1,86 @@
...
+---
+# `logfmt` stage
```
KMiller-Grafana, line 4: As we add new content to the Loki documentation, we're trying to standardize what we include in the YAML front matter.  This suggestion adds 2 new fields to implement that.  I'm leaving out the weight field, as adding it only to this stage's file would mess up our (fine) alphabetized organization.

```suggestion
title: logfmt
menuTitle: logfmt
description: The logfmt parsing stage reads logfmt log lines and extracts the data into labels.
---
# `logfmt` stage
```

docs/sources/clients/promtail/stages/logfmt.md
```diff
@@ -0,0 +1,86 @@
...
+
+The `logfmt` stage is a parsing stage that reads the log line as [logfmt](https://brandur.org/logfmt) and allows extraction of data into labels.
+
+## Schema
+
+```yaml
+logfmt:
+  # Set of key/value pairs for mapping of logfmt fieds to extracted labels. The key will be
+  # the key in the extracted data while the expression will be the value. If the value
+  # is empty then the logfmt field with the same name is extracted.
```
KMiller-Grafana, line 14: What do you think of being more explicit?
```suggestion
  # Set of key/value pairs for mapping of logfmt fields to extracted labels. The YAML key will be
  # the key in the extracted data, while the expression will be the YAML value. If the value
  # is empty, then the logfmt field with the same name is extracted.
```

docs/sources/clients/promtail/stages/logfmt.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+This stage uses the [go-logfmt](https://github.com/go-logfmt/logfmt) unmarshaler, which means non-string types like
+numbers or booleans will be unmarshaled into those types. The extracted data
+can hold non-string values and this stage does not do any type conversions;
+downstream stages will need to perform correct type conversion of these values
+as necessary. Please refer to the [the `template` stage](../template/) for how
+to do this.
+
+If the value extracted is a complex type it's value is extracted as a string.
```
KMiller-Grafana, line 29: Just a bit of grammar and spelling suggestions.

```suggestion
This stage uses the [go-logfmt](https://github.com/go-logfmt/logfmt) unmarshaler, which means non-string types like
numbers or booleans will be unmarshaled into those types. The extracted data
can hold non-string values, and this stage does not do any type conversions;
downstream stages will need to perform correct type conversion of these values
as necessary. Please refer to the [`template` stage](../template/) for how
to do this.

If the value extracted is a complex type, its value is extracted as a string.
```

docs/sources/clients/promtail/stages/logfmt.md
```diff
@@ -0,0 +1,86 @@
...
+- logfmt:
+    mapping:
+      user:
+    source: extra
+```
+
+And the given log line:
+
+```
+time=2012-11-01T22:08:41+00:00 app=loki level=WARN duration=125 message="this is a log line" extra="user=foo""
```
KMiller-Grafana, line 75: Is there an extra double quote mark at the end of this log line?  If yes, this suggestion removes it.
```suggestion
time=2012-11-01T22:08:41+00:00 app=loki level=WARN duration=125 message="this is a log line" extra="user=foo"
```

KMiller-Grafana (changes requested):
Thanks for adding in this missing content!  A few suggestions for you.

docs/sources/clients/promtail/stages/logfmt.md
DylanGuedes: agreed.

PR #6252, opened 2022-05-25
Tsdb/inverted index wiring

owen-d:
Followup to https://github.com/grafana/loki/pull/6233

This allows us to run two incompatible inverted indices in the ingesters, one for bit-prefixed sharding (tsdb) and one for modulo sharding (everything else). It also adds some more config validation and fixes a bug in the config-wrapper that hasn't surfaced yet, but will as soon as you try to run a non-current boltdb-shipper period config: https://github.com/grafana/loki/commit/d9f1e06838979120d84c1ef2732ae18b4aa8572a


pkg/ingester/index/multi.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func NewMultiInvertedIndex(periods []config.PeriodConfig, indexShards uint32) (*Multi, error) {
+	var (
+		err error
+
+		ii          Interface // always stored in 0th index
+		bitPrefixed Interface // always stored in 1st index
+
+		periodIndices []periodIndex
```
DylanGuedes, line 31: maybe you can assign `periodIndices := make([]periodIndex, len(periods))`

pkg/ingester/index/multi.go
```diff
@@ -0,0 +1,123 @@
...
+	return
+}
+
+func (m *Multi) Delete(labels labels.Labels, fp model.Fingerprint) {
+	for _, i := range m.indices {
+		if i != nil {
+			i.Delete(labels, fp)
+		}
+	}
+
```
DylanGuedes, line 79: nit
```suggestion
```

pkg/ingester/index/multi.go
```diff
@@ -0,0 +1,123 @@
...
+	return &Multi{
+		periods: periodIndices,
+		indices: []Interface{ii, bitPrefixed},
+	}, nil
+}
+
+func (m *Multi) Add(labels []logproto.LabelAdapter, fp model.Fingerprint) (result labels.Labels) {
+	for _, i := range m.indices {
+		if i != nil {
```
DylanGuedes, line 66: for curiosity, why would an index inside m.indices be nil?

pkg/ingester/index/multi.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (m *Multi) LabelNames(t time.Time, shard *astmapper.ShardAnnotation) ([]string, error) {
+	return m.indexFor(t).LabelNames(shard)
+}
+
+func (m *Multi) LabelValues(t time.Time, name string, shard *astmapper.ShardAnnotation) ([]string, error) {
+	return m.indexFor(t).LabelValues(name, shard)
+}
+
+// Query planning is responsible for ensuring no query spans more than one inverted index.
```
DylanGuedes, line 94: is this comment out of place?

pkg/ingester/index/multi.go
```diff
@@ -0,0 +1,123 @@
...
+// Therefore we don't need to account for both `from` and `through`.
+func (m *Multi) indexFor(t time.Time) Interface {
+	for i := range m.periods {
+		if !m.periods[i].Time.After(t) && (i+1 == len(m.periods) || t.Before(m.periods[i+1].Time)) {
+			return m.indices[m.periods[i].idx]
```
DylanGuedes, line 99: the first index that matches the given `t` will be returned; shouldn't this return a collection of all matches instead?

pkg/ingester/index/multi_test.go
```diff
@@ -0,0 +1,179 @@
...
+			},
+			{
+				From:      MustParseDayTime("2021-01-01"),
+				IndexType: config.TSDBType,
+			},
+		},
+		factor,
+	)
+	require.Error(t, err)
+
```
DylanGuedes, line 74: nit
```suggestion
```

pkg/ingester/index/multi_test.go
```diff
@@ -0,0 +1,179 @@
...
+	_, err := NewMultiInvertedIndex(
+		[]config.PeriodConfig{
+			{
+				From:      MustParseDayTime("2020-01-01"),
+				IndexType: config.StorageTypeBigTable,
+			},
+		},
+		factor,
+	)
+	require.Nil(t, err)
```
DylanGuedes, line 58: nitpicking:
I noticed we use `require.Nil(t, err)` frequently, but `require.NoError(t, err)` looks better I think.  [require.NotNil](https://github.com/stretchr/testify/blob/083ff1c0449867d0d8d456483ee5fab8e0c0e1e6/assert/assertions.go#L505) implementation vs [require.NoError implementation](https://github.com/stretchr/testify/blob/083ff1c0449867d0d8d456483ee5fab8e0c0e1e6/assert/assertions.go#L1334)
```suggestion
	require.NoError(t, err)
```

DylanGuedes (commented):
few nits/questions

pkg/ingester/index/multi.go
```diff
@@ -0,0 +1,123 @@
...
+		periodIndices []periodIndex
+	)
+
+	for _, pd := range periods {
+		switch pd.IndexType {
+		case config.TSDBType:
+			if bitPrefixed == nil {
+				bitPrefixed, err = NewBitPrefixWithShards(indexShards)
+				if err != nil {
+					return nil, err
```
DylanGuedes, line 40: nit: maybe you could wrap this error with `errors.Wrap` and add more details to the message, like which period erroed and its index type). Motivation:
there's a chain of calls that aren't wrapping errors that will end up here, so it might be hard to identify the root of the problem. The chain is: 
- `GetOrCreateInstance` is invoked at multiple places without wrapping errors and triggers `newInstance`
  - `newInstance` will invoke `index.NewMultiInvertedIndex` without wrapping errors
    - we end up here, with `NewMultiInvertedIndex` invoking `NewBitPrefixWithShards`

pkg/ingester/index/multi.go
owen-d: In the case where there's only `tsdb` period configs or non-tsdb period configs (such as `boltdb-shipper`), only one of the sub-indices will be needed. If that's the case, we can avoid the overhead of running an unnecessary one.

pkg/ingester/index/multi.go
owen-d: I intended to put it here because most queries in loki use a `from, through`, but in this case we're using another part of the code base to ensure we'd never query more than one period at a time, so we don't need to account for the range `from->through`, only the first point (`from`).

pkg/ingester/index/multi.go
owen-d: There can only be one match for any `t`, so we return the first one we find which will be the only one.

pkg/ingester/index/multi_test.go
owen-d: I prefer `require.Nil`'s error message when it doesn't pass (helpful debugging regressions/etc), so I usually use that.

pkg/ingester/index/multi.go
owen-d: Yeah, that would work, but the overhead here is basically zero (a 2-length slice that's instantiated once on startup)

pkg/ingester/index/multi.go
owen-d: Good call!

pkg/ingester/index/multi.go
DylanGuedes: got it, I thought we could have a huge number of periods

pkg/ingester/index/multi.go
```diff
@@ -0,0 +1,123 @@
...
+type Multi struct {
+	periods []periodIndex
+	indices []Interface
+}
+
+func NewMultiInvertedIndex(periods []config.PeriodConfig, indexShards uint32) (*Multi, error) {
+	var (
+		err error
+
+		ii          Interface // always stored in 0th index
```
trevorwhitney, line 28: does `ii` stand for `invertedIndex`? if so I might opt for the longer version of that variable as it's not that long.

pkg/ingester/index/multi.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/prometheus/prometheus/model/labels"
+
+	"github.com/grafana/loki/pkg/logproto"
+	"github.com/grafana/loki/pkg/querier/astmapper"
+	"github.com/grafana/loki/pkg/storage/config"
+)
+
+type periodIndex struct {
+	time.Time
+	idx int // address of the index to use
```
trevorwhitney, line 16: a bit confused about what this represents. it seems to be a type more than an address to me, ie 0 is a regular inverted index, whereas 1 is a tsdb inverted index. could we reflect that in the name and/or comment?

trevorwhitney: or is this the index in the `indices` property on `Multi`? If so, and there can only be 2, I'm curious if instead of storing these in a fixed length slice with magic indices, maybe `Multi` should have a field for each type?

pkg/loki/config_wrapper.go
```diff
@@ -99,12 +99,12 @@ func (c *ConfigWrapper) ApplyDynamicConfig() cfg.Source {
 			return err
 		}
 
-		if len(r.SchemaConfig.Configs) > 0 && config.UsingBoltdbShipper(r.SchemaConfig.Configs) {
-			betterBoltdbShipperDefaults(r, &defaults)
+		if i := lastBoltdbShipperConfig(r.SchemaConfig.Configs); i != len(r.SchemaConfig.Configs) {
```
trevorwhitney, line 102: What was the reason for using `len(r.SchemaConfig.Configs` as your sentinel value when a config can't be found here? I wonder if `-1` might be a more canonical?

pkg/loki/config_wrapper.go
```diff
@@ -116,6 +116,27 @@ func (c *ConfigWrapper) ApplyDynamicConfig() cfg.Source {
 	}
 }
 
+func lastConfigFor(configs []config.PeriodConfig, predicate func(config.PeriodConfig) bool) int {
+	for i := len(configs) - 1; i >= 0; i-- {
+		if predicate(configs[i]) {
+			return i
+		}
+	}
+	return len(configs)
```
trevorwhitney, line 125: same question as above, but asked closer to the source:
What was the reason for using `len(r.SchemaConfig.Configs` as your sentinel value when a config can't be found here? I wonder if `-1` might be a more canonical?

trevorwhitney: I see the motivation for providing the last config when a more specific config of the correct type can't be found, but since we're only using this to find configs the apply better BoltdbShipper or TSDBConfig defaults for, I don't think we need to care about using the last config for anything.

trevorwhitney (commented):
Looks good Owen, I have a few comments which are mostly questions. Thanks!

pkg/ingester/index/multi.go
owen-d: That's a fair point. I wrote it like this in case we had period configs like `[boltdb, tsdb, boltdb, tsdb]`, in which case we'd want the first period to point to the index of the `boltdb` inverted index, the second to point to the `tsdb` inverted index and so on. Decoupling the period from the index it points to allows us to instantiate a maximum of one inverted index for each type, regardless of the number of period configs.

The way it's written now will allow us to add other index types in the future if necessary. Does that make sense?

pkg/loki/config_wrapper.go
owen-d: I used the convention from [`sort.Search`](https://go.dev/src/sort/search.go) in the stdlib.

trevorwhitney: ahh, all these go conventions I don't know. sounds good!

pkg/loki/config_wrapper.go
trevorwhitney: your comment above explained it, following the convention from `sort.Search` :+1:

pkg/ingester/index/multi.go
trevorwhitney: it does, but we could also come back and update this struct with the necessary types when we get to that point. just feels like we're overloading slice indexing as a mechanism for expressing type, whereas there are clearer ways to do that.

that being said I have no interest in dying on this hill, and I'm cool with keeping it this way if you think it's more flexible.

slim-bean (approved):
LGTM

PR #6256, opened 2022-05-27
promtail: Allow docker_sd_config to handle log lines longer than 64k

### LINKED ISSUE #6257: promtail docker_sd_config: long lines crash promtail parsing
**Describe the bug**
@jeschkies Using promtail's `docker_sd_config`, lines longer than than 64k result in an error that prevents further log lines from being consumed. The error is:

```
bufio.Scanner: token too long
```

This occurs because by default `bufio.Scanner` uses [`MaxScanTokenSize = 64 * 1024`](https://pkg.go.dev/bufio#pkg-constants), and lines longer than the buffer will result in the above error.

See #6256 for a potential fix.

**To Reproduce**
1. Started Promtail with `docker_sd_config` listening for Docker container logs
2. Start a Docker container which generates log lines longer than 64k bytes
3. Observe that parsing of the logs for the container fail when parsing long log lines and do not parse further lines

**Expected behavior**
Promtail parses Docker container log lines longer than 64k without issue.

**Environment:**
 - Infrastructure: bare metal
 - Deployment tool: ansible

From `clients/pkg/promtail/targets/docker/targetmanager_test.go` when a long log line is added to `clients/pkg/promtail/targets/docker/testdata/flog.log` and included in the tests:

```
level=warn msg="finished scanning logs lines with an error" err="bufio.Scanner: token too long"
```

shane-axiom:
In promtail's docker_sd_config, parsing of a container's
logs would fail if any log line was over 64k bytes in length.
This is because bufio.Scanner uses a 64k buffer by default.
Since there doesn't appear to be a way to elegantly catch
`bufio.ErrTooLong` and carry on, the parsing approach was changed
to bufio.Reader which can handle lines of arbitrary length.

Fixes #6257

Originally I considered increasing the buffer or allowing for a config parameter to do the same, but that would only raise the ceiling of parsable log line lengths. Since there doesn't seem to be a way to catch `bufio.ErrTooLong` when using `Scanner`, it seemed better to just go with the `Reader` approach and handle lines of arbitrary length.

Log lines of this length can be problematic by nature and it may be prudent to add a configuration to truncate parsed log lines to a certain maximum length, but I think that concern is separate enough to avoid conflating it here.

**Checklist**
- [-] Documentation added
- [x] Tests updated
- [-] Is this an important fix or new feature? Add an entry in the `CHANGELOG.md`.
- [-] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`


clients/pkg/promtail/targets/docker/target.go
```diff
@@ -148,14 +148,38 @@ func extractTs(line string) (time.Time, string, error) {
...
+		line, ln []byte
+	)
+
+	for isPrefix && err == nil {
+		line, isPrefix, err = r.ReadLine()
+		ln = append(ln, line...)
+	}
+
+	return string(ln), err
+}
```
chaudum, line 165: Do you mind putting this function into a utility package, e.g. `clients/pkg/promtail/common`? I think it may be used in various places.

chaudum (commented):
Thanks for the fix!

clients/pkg/promtail/targets/docker/target.go
trevorwhitney: I'm fine either way on that actually, we can also wait until we need it somewhere else to figure out the best package for it to go into, or what the interface should look like, ie. there might be more duplication to reduce if we refactor our the section in `process` for processing a whole file?

clients/pkg/promtail/targets/docker/targetmanager_test.go
```diff
@@ -89,7 +89,7 @@ func Test_TargetManager(t *testing.T) {
 	require.True(t, ta.Ready())
 
 	require.Eventually(t, func() bool {
-		return len(entryHandler.Received()) >= 5
+		return len(entryHandler.Received()) >= 6
```
trevorwhitney, line 92: what caused this value to change? I don't see any changes to the test setup.

trevorwhitney (approved):
Thanks for the improvement, LGTM!

clients/pkg/promtail/targets/docker/targetmanager_test.go
shane-axiom: I inserted a Very Long Log line into `flog.log` to test against (over 65k). I had to increase by one here to read in that extra line.

clients/pkg/promtail/targets/docker/target.go
shane-axiom: Yeah I'm fine either way, happy to move it if that's the desired approach...

clients/pkg/promtail/targets/docker/targetmanager_test.go
trevorwhitney: got it, and that's a binary file which is why I'm not seeing the diff, thanks!

PR #6263, opened 2022-05-30
update drone branch match for 3 digit k release

sandeepsukhani:
This week we are building release `k100` but the images have not been pushed automatically since the branch name matcher is set to `k??` which means only 2 characters after prefix `k` would match. This PR updates the drone config to look for 3 letters after prefix `k`.

Since drone only supports glob for pattern matching which is not as feature right as regex we can't have a pattern matching variable lengths. See https://discourse.drone.io/t/extended-glob-patterns/7047

I think this is good enough for now and we will have ~2.5 years before we need to update this again :D

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -9,7 +9,7 @@ local condition(verb) = {
       [verb]:
         [
           'refs/heads/main',
-          'refs/heads/k??',
+          'refs/heads/k???',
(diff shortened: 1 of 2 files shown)


vlad-diachenko (approved):
LGTM

danmdinu (approved):

kavirajk (approved):
LGTM :+1:

PR #6267, opened 2022-05-30
use tsdb config for initializing index gateway client

sandeepsukhani:
We were temporarily using `boltdb-shipper` config for initializing the index gateway client. This PR fixes it by doing relevant updates.

### DIFF
--- a/pkg/storage/store.go
+++ b/pkg/storage/store.go
@@ -186,12 +186,12 @@ func (s *store) chunkClientForPeriod(p config.PeriodConfig) (client.Client, erro
 	return chunks, nil
 }
 
-func shouldUseBoltDBIndexGatewayClient(cfg Config) bool {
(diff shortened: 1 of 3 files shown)


pkg/storage/store.go
```diff
@@ -252,7 +251,7 @@ func (s *store) storeForPeriod(p config.PeriodConfig, chunkClient client.Client,
 	)
 
 	// (Sandeep): Disable IndexGatewayClientStore for stores other than tsdb until we are ready to enable it again
-	/*if shouldUseBoltDBIndexGatewayClient(s.cfg) {
+	/*if shouldUseIndexGatewayClient(s.cfg) {
```
kavirajk, line 254: I understand it's a comment, but still shouldn't this be now something like this?
```suggestion
	/*if s.cfg.BoltDBShipperConfig != nil && shouldUseIndexGatewayClient(s.cfg.BoltDBShipperConfig) {
```

kavirajk (approved):
LGTM!. One minor nit.

PR #6269, opened 2022-05-31
feat(logql): support histogram in logql

### LINKED ISSUE #2953: Support creating histograms using logql
I'd like to be able to create histograms using logql based on an extracted field. We want to track job execution times with histograms on a per-user basis and doing this with labels would have too high of a cardinality.

The design in the original draft of logql v2 seems reasonable:
https://github.com/cyriltovena/loki/blob/a87cf44829841958a3dd78030266ac93eb7269c3/docs/design-documents/2020-03-logql-fields.md#series--histogram-operators 

Using cloudwatch, which has a `bin` function for this. But we're already sending logs to Loki so that's not ideal.

**Additional context**

adityacs:
Adds histogram function to logql. 

Fixes #2953 

Output looks as following.

```json
[
  {
    "metric": {
      "path": "/api/v2/"
    },
    "histvalues": [
      [1653907528,[{"UpperBound":0.005,"Count":6},{"UpperBound":0.025,"Count":2}]],
      [1653907542,[{"UpperBound":0.005,"Count":7},{"UpperBound":0.025,"Count":2}]],
    ]
  }
]
```
Can be changed to other format. Something like below.
```json
[
  {
    "metric": {
      "path": "/api/v2/"
    },
    "histvalues": [
      [1653907528,[{"0.005": 6},{"0.025": 2}]],
      [1653907542,[{"0.005": 7},{"0.025": 2}]],
    ]
  }
]
```

**Checklist**
- [x] Documentation added
- [x] Tests updated


docs/sources/logql/metric_queries.md
```diff
@@ -114,6 +115,58 @@ sum by (org_id) (
 
 This calculates the amount of bytes processed per organization ID.
 
+```logql
+buckets_over_time((.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10),
```
cyriltovena, line 119: This is great but I think it would be better if we could find bucket automatically like sparse histogram https://grafana.com/blog/2021/11/03/how-sparse-histograms-can-improve-efficiency-precision-and-mergeability-in-prometheus-tsdb/

pkg/logcli/query/query.go
```diff
@@ -161,6 +161,8 @@ func (q *Query) printResult(value loghttp.ResultValue, out output.LogOutput, las
 		q.printScalar(value.(loghttp.Scalar))
 	case loghttp.ResultTypeMatrix:
 		q.printMatrix(value.(loghttp.Matrix))
+	case loghttp.ResultTypeHistogramMatrix:
```
cyriltovena, line 164: using the labels we should be able to not have to use a new response model WDYT ?

adityacs:
Response will be in below format

```json
[
  {
    "metric": {
      "filename": "/api/v2",
      "le": "0.025"
    },
    "values": [
      [
        1654670374,
        "1"
      ],
      [
        1654670388,
        "1"
      ]
    ]
  },
  {
    "metric": {
      "filename": "/api/v2",
      "le": "0.005"
    },
    "values": [
      [
        1654670374,
        "1"
      ],
      [
        1654670388,
        "2"
      ]
    ]
  }

```

pkg/logcli/query/query.go
adityacs: @cyriltovena Have removed custom response format. Output will be of the form

```json
[
  {
    "metric": {
      "path": "/api/v2/"
      "le": "0.05",
    },
    "values": [
      [
        1653907528, 
        "2"
      ],
      [
        1653907542, 
        "3"
      ]
    ]
  },
  {
    "metric": {
      "path": "/api/v2/"
      "le": "0.025",
    },
    "values": [
      [
        1653907528, 
        "2"
      ],
      [
        1653907542, 
        "3"
      ]
    ]
  }
]
```

docs/sources/logql/metric_queries.md
cyriltovena: I think if finding buckets is doable as per the doc I shared then I would love to have only this

MasslessParticle:
I going to close this due to inactivity. If you still think this is valuable, It's a great candidate for a LID: https://grafana.com/docs/loki/next/lids/. Write one up and include a link to this PR as part of it so the work isn't lost.

adityacs:
@MasslessParticle I have resumed work on this PR. So, re-opened the PR. However, I will keep it in draft mode.

adityacs:
Output will look like following

```json
[
    {
      "metric": {
        "path": "/GET"
      },
      "histograms": [
        [
          1673601440,
          {
            "count": "5",
            "sum": "67",
            "buckets": [
              [
                "2.82842712474619",
                "3.0844216508158815",
                "1"
              ],
              [
                "8.724061861322062",
                "9.513656920021768",
                "1"
              ],
              [
                "14.672064691274738",
                "16",
                "1"
              ],
              [
                "17.448123722644123",
                "19.027313840043536",
                "1"
              ],
              [
                "20.749432874416154",
                "22.62741699796952",
                "1"
              ]
            ]
          }
        ],
        [
          1673601454,
          {
            "count": "5",
            "sum": "67",
            "buckets": [
              [
                "2.82842712474619",
                "3.0844216508158815",
                "1"
              ],
              [
                "8.724061861322062",
                "9.513656920021768",
                "1"
              ],
              [
                "14.672064691274738",
                "16",
                "1"
              ],
              [
                "17.448123722644123",
                "19.027313840043536",
                "1"
              ],
              [
                "20.749432874416154",
                "22.62741699796952",
                "1"
              ]
            ]
          }
        ]
      ]
    }
]

The  `buckets` array values indicates following

```json
[
  "2.82842712474619", -- lower bound
  "3.0844216508158815", -- upper bound
  "1" -- count
],
```

pkg/logql/range_vector.go
```diff
@@ -456,6 +488,51 @@ func quantileOverTime(q float64) func(samples []promql.Point) float64 {
...
+			ZeroThreshold:   floatHistogram.Histogram.GetZeroThreshold(),
+			ZeroCount:       floatHistogram.Histogram.GetZeroCount(),
+			Count:           floatHistogram.Histogram.GetSampleCount(),
+			Sum:             floatHistogram.Histogram.GetSampleSum(),
+			PositiveSpans:   positiveSpans,
+			NegativeSpans:   negativeSpans,
+			PositiveBuckets: floatHistogram.Histogram.GetPositiveDelta(),
+			NegativeBuckets: floatHistogram.Histogram.GetNegativeDelta(),
+		}
+		return h.ToFloat()
```
owen-d, line 532: What's the reason for going through all this transformation work to use a `FloatHistogram` specifically?

pkg/logcli/client/client.go
```diff
@@ -262,6 +262,11 @@ func (c *DefaultClient) doRequest(path, query string, quiet bool, out interface{
 			log.Println("error closing body", err)
 		}
 	}()
+	// err = json.NewDecoder(resp.Body).Decode(out)
```
owen-d, line 265: accident?

pkg/loghttp/query.go
```diff
@@ -188,6 +187,7 @@ func (q *QueryResponseData) UnmarshalJSON(data []byte) error {
 				q.Result = ss
 			case ResultTypeMatrix:
 				var m Matrix
+				fmt.Printf("unmarshal matrix %v\n", string(value))
```
owen-d, line 190: remove

pkg/logproto/logproto.proto
```diff
@@ -162,10 +162,26 @@ message Sample {
...
+	double lower = 1 [(gogoproto.jsontag) = "lower"];
+	double upper = 2 [(gogoproto.jsontag) = "upper"];
+	double count = 3 [(gogoproto.jsontag) = "count"];
+}
+
 // LegacySample exists for backwards compatibility reasons and is deprecated. Do not use.
 message LegacySample {
   double value = 1;
   int64 timestamp_ms = 2;
+  SampleHistogram histogram = 3 [(gogoproto.jsontag) = "histogram"];
```
owen-d, line 184: Why are we adding a histogram to every sample?

pkg/util/marshal/query.go
```diff
@@ -252,6 +264,7 @@ func encodeResult(v parser.Value, s *jsoniter.Stream) error {
 		}
 
 		encodeMatrix(m, s)
+		fmt.Printf("matrix encode %v\n", s)
```
owen-d, line 267: remove

owen-d (commented):
I understand this is a work in progress, so I'll just point out a few things with the approach and skip over tests, comments, etc. It's worth pointing out there are two different use cases (and likely two different implementations) for histogram bucketing here.

1) The use case from the original issue https://github.com/grafana/loki/issues/2953. This asks for specific bucketing (where the buckets are known ahead of time). For instance, this may be buckets which are significant to some external application logic, like `[under_SLO, over_SLO]`.
2) What @cyriltovena suggested, which is the "automatic" bucketing used in prometheus sparse histograms.

For either one, I don't think we'll need to change the data model at all (also as cyril suggested). This should return the same response that Prometheus does -- think `sum by (le) (rate(<metric>_bucket{}[5m]))`.

As for your questions about the performance of the bucketing calculations, I'd love to see some benchmarks but I expect it will be reasonable.

pkg/logql/range_vector.go
adityacs: `promql.Point` only supports `FloatHistogram` https://github.com/prometheus/prometheus/blob/main/promql/value.go#L110

pkg/logcli/client/client.go
adityacs: Yes 😄

pkg/logproto/logproto.proto
adityacs: To match `model.Sample` https://github.com/prometheus/common/blob/main/model/value.go#L36

pkg/logql/range_vector.go
owen-d: Ah, looks like sparse histograms actually changed the data model itself in prometheus responses (see `note`). I don't quite understand it yet.
image

adityacs:
@owen-d I am okay to support both implementations. Native histograms should be sufficient for most of the cases but still, as you pointed out there might be cases where specific bucket boundaries are required.

Regarding data model, we could add values to `le` based on bucket values and it works fine for histograms with pre-defined bucket boundaries. However, for prometheus native histograms, bucket values are not cumulative. So, not sure if we can use `le` as label here. May be we can introduce labels as `lb(lower_bound), up(upper_bound)`? We might lose the information whether it is lower inclusive or upper inclusive.

pharaujo:
@MasslessParticle I'm curious why this PR was closed. I couldn't find any alternative PR for this feature, and the issue which this PR links to remains open.

PR #6273, opened 2022-05-31
make tsdb store instance a singleton

sandeepsukhani:
Currently, if we have multiple schema configs with `tsdb` as index store, we will create multiple store instances.
This would cause a problem since we maintain a state on the local disk and the code is not meant to run multiple instances.
Since we do not do any schema specific operations in `tsdb` and running multiple instances would be complex and wasteful, this PR changes the code to make `tsdb` index store instance a singleton. If we do need to do schema specific handling in future, it would be a good idea to abstract away the handling of different schemas.

I have also done the following changes:
* Refactored the code a bit to avoid initializing tsdb and head manager when running `tsdb` store in read-only mode.
* Make it mandatory to configure `tsdb` with `24h` index periods, same as `boltdb-shipper`, to keep things simple.

### DIFF
--- a/pkg/storage/config/schema_config.go
+++ b/pkg/storage/config/schema_config.go
@@ -42,6 +42,9 @@ const (
 	// BoltDBShipperType holds the index type for using boltdb with shipper which keeps flushing them to a shared storage
 	BoltDBShipperType = "boltdb-shipper"
 	TSDBType          = "tsdb"
+
+	// ObjectStorageIndexRequiredPeriod defines the required index period for object storage based index stores like boltdb-shipper and tsdb
+	ObjectStorageIndexRequiredPeriod = 24 * time.Hour
 )
 
 var (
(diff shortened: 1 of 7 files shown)


pkg/storage/config/schema_config.go
```diff
@@ -282,6 +283,10 @@ func (cfg PeriodConfig) validate() error {
 		return validateError
 	}
 
+	if cfg.IndexType == TSDBType && cfg.IndexTables.Period != 24*time.Hour {
```
chaudum, line 286: Maybe you want to make [`tsdb.indexPeriod`](https://github.com/grafana/loki/pull/6273/files#diff-f74244a4225afefba7e0b4ba3c68b541d2d498a5fc0a0923f0c07a7d2e6e74cbR32) public and use it here:

```suggestion
	if cfg.IndexType == TSDBType && cfg.IndexTables.Period != tsdb.IndexPeriod {
```

chaudum (commented):
Generally lgtm. Have two smaller suggestions.

pkg/storage/config/schema_config.go
sandeepsukhani: tried it already, it causes circular dependency since `tsdb` uses schema config :)

pkg/storage/stores/tsdb/store.go
```diff
@@ -10,16 +13,39 @@ import (
...
+// fetcher.Fetcher instances could be different due to periodic configs having different types of object storage configured
+// for storing chunks.
 func NewStore(indexShipperCfg indexshipper.Config, p config.PeriodConfig, f *fetcher.Fetcher,
 	objectClient client.ObjectClient, limits downloads.Limits, reg prometheus.Registerer) (stores.ChunkWriter, series.IndexStore, error) {
-	var (
-		nodeName = indexShipperCfg.IngesterName
-		dir      = indexShipperCfg.ActiveIndexDirectory
-	)
-	tsdbMetrics := NewMetrics(reg)
+	if storeInstance == nil {
```
owen-d, line 36: I understand but really dislike the side effect of mutating the global variable within the `New` function. Can we handle this in a cleaner fashion by passing the `IndexWriter` as an argument to this function?

owen-d (commented):
Few suggestions, but looks good

owen-d (approved):
Approving this -- @sandeepsukhani is going to open up an additional PR to refactor the dependency chain so we don't have to mutate a global variable during the `NewStore` call

pkg/storage/config/schema_config.go
sandeepsukhani: taken care of this suggestion. Thanks for the review folks!

PR #6275, opened 2022-05-31
ingester: Refactor ingester metrics from globals into the ingesterMetrics struct

simonswine:
This moves global metrics into the dedicated ingester metrics struct


chaudum (approved):
lgtm

PR #6278, opened 2022-06-01
add metrics for tracking index shipper operations

sandeepsukhani:
In PR #6226, we refactored the code to make the `boltdb-shipper` index client use the generic `indexshipper` to manage the index on object storage. This PR takes care of follow up action to add metrics to `indexshipper`.

I have made sure we have the same metrics exposed as before by using `WrapRegistererWithPrefix`, i.e. when using `indexshipper` for `boltdb-shipper`, metrics would be prefixed with `loki_boltdb_shipper_` while for `tsdb`, the prefix would be `loki_tsdb_shipper_`

### DIFF
--- a/pkg/storage/stores/indexshipper/uploads/metrics.go
+++ b/pkg/storage/stores/indexshipper/uploads/metrics.go
@@ -1,0 +1,25 @@
+package uploads
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+const (
+	statusFailure = "failure"
+	statusSuccess = "success"
+)
+
+type metrics struct {
+	tablesUploadOperationTotal *prometheus.CounterVec
+}
+
+func newMetrics(r prometheus.Registerer) *metrics {
+	return &metrics{
+		tablesUploadOperationTotal: promauto.With(r).NewCounterVec(prometheus.CounterOpts{
+			Namespace: "loki_boltdb_shipper",
+			Name:      "tables_upload_operation_total",
+			Help:      "Total number of upload operations done by status",
+		}, []string{"status"}),
+	}
+}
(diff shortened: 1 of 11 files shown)


pkg/storage/stores/indexshipper/uploads/metrics.go
```diff
@@ -0,0 +1,25 @@
+package uploads
```
sandeepsukhani, line 1: old metrics.go for reference https://github.com/grafana/loki/blob/440fb7d74cad481bebb222b21fc063fa56c2f5ac/pkg/storage/stores/shipper/uploads/metrics.go
`openExistingFileFailuresTotal` is only relevant for `boltdb-shipper` so it is already added to it [here](https://github.com/grafana/loki/blob/6ac955cc7d395874ce60516d182678542415bca3/pkg/storage/stores/shipper/index/metrics.go#L14)

pkg/storage/stores/indexshipper/uploads/table_manager.go
```diff
@@ -130,9 +133,11 @@ func (tm *tableManager) uploadTables(ctx context.Context) {
 
 	level.Info(util_log.Logger).Log("msg", "uploading tables")
 
+	status := statusSuccess
 	for _, table := range tm.tables {
 		err := table.Upload(ctx)
 		if err != nil {
+			status = statusFailure
```
kavirajk, line 140: Q: can this be confusing when only few tables failed to upload. But in the metrics, we record as though everything is failed?
`tm.metrics.tablesUploadOperationTotal.WithLabelValues(status).Inc()`

kavirajk (approved):
LGTM overall. Approving it to unblock.

Started one small discussion about partial table upload failures. Can you clarify please?

pkg/storage/stores/indexshipper/uploads/table_manager.go
sandeepsukhani: We track the overall upload operation. We can count individual table upload operations as well but the investigator would anyways have to check the logs to find more details about it.

PR #6282, opened 2022-06-01
Logcli: fix boltdb shipper local query in logcli and support `fake` tenant

slim-bean:
Fixes a bug related to using the logcli local config query option and boltdb shipper
Also defaults the tenant to `fake` if none is provided.


sandeepsukhani (approved):
LGTM

PR #6224, opened 2022-05-23
operator: Add support for grpc over tls for loki components

periklis:
**What this PR does / why we need it**:
The following PR provides support for GRPC over TLS for Loki components.

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] Is this an important fix or new feature? Add an entry in the `CHANGELOG.md`.
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #6288, opened 2022-06-02
operator: Expose only an HTTPS gateway when in openshift-logging mode

aminesnow:
This PR configures LokiStack gateway to exclusively use the HTTPS protocol when using the openshift-logging mode. In addition, traffic to the gateway k8s service is re-encrypted using a reencrypt termination policy while using the openshift-logging mode.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [X] Is this an important fix or new feature? Add an entry in the `CHANGELOG.md`.

### DIFF
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Main
 
 - [6195](https://github.com/grafana/loki/pull/6195) **periklis**: Add ruler config support
+- [6288](https://github.com/grafana/loki/pull/6288) **aminesnow**: Expose only an HTTPS gateway when in openshift mode
 - [6198](https://github.com/grafana/loki/pull/6198) **periklis**: Add support for custom S3 CA
 - [6199](https://github.com/grafana/loki/pull/6199) **Red-GV**: Update GCP secret volume path
 - [6125](https://github.com/grafana/loki/pull/6125) **sasagarw**: Add method to get authenticated from GCP
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Main
 
-- [6195](https://github.com/grafana/loki/pull/6195) **periklis**: Add ruler config support
 - [6288](https://github.com/grafana/loki/pull/6288) **aminesnow**: Expose only an HTTPS gateway when in openshift mode
+- [6195](https://github.com/grafana/loki/pull/6195) **periklis**: Add ruler config support
 - [6198](https://github.com/grafana/loki/pull/6198) **periklis**: Add support for custom S3 CA
 - [6199](https://github.com/grafana/loki/pull/6199) **Red-GV**: Update GCP secret volume path
 - [6125](https://github.com/grafana/loki/pull/6125) **sasagarw**: Add method to get authenticated from GCP
--- a/operator/hack/addons_ocp.yaml
+++ b/operator/hack/addons_ocp.yaml
@@ -37,12 +37,12 @@ spec:
             - name: LOKI_ORG_ID
               value: application
             - name: LOKI_ADDR
-              value: http://lokistack-dev-gateway-http.openshift-logging.svc:8080/api/logs/v1/application
+              value: https://lokistack-dev-gateway-http.openshift-logging.svc:8080/api/logs/v1/application
             - name: LOKI_BEARER_TOKEN_FILE
               value: /var/run/secrets/kubernetes.io/serviceaccount/token
           args:
             - -c
-            - while true; do logcli query '{job="systemd-journal"}'; sleep 30; done
+            - while true; do logcli --ca-cert="/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt" query '{job="systemd-journal"}'; sleep 30; done
       serviceAccountName: lokistack-dev-addons-logcli
 ---
 apiVersion: apps/v1
@@ -118,7 +118,9 @@ metadata:
 data:
   promtail.yaml: |
     clients:
-      - url: http://lokistack-dev-gateway-http.openshift-logging.svc:8080/api/logs/v1/application/loki/api/v1/push
+      - url: https://lokistack-dev-gateway-http.openshift-logging.svc:8080/api/logs/v1/application/loki/api/v1/push
+        tls_config:
+          ca_file: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
         tenant_id: application
         bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
         backoff_config:
--- a/operator/internal/manifests/build_test.go
+++ b/operator/internal/manifests/build_test.go
@@ -263,7 +263,7 @@ func TestBuildAll_WithFeatureFlags_EnableTLSServiceMonitorConfig(t *testing.T) {
 			continue
 		}
 
-		secretName := fmt.Sprintf("%s-http-metrics", name)
+		secretName := fmt.Sprintf("%s-http-tls", name)
 		expVolume := corev1.Volume{
 			Name: secretName,
 			VolumeSource: corev1.VolumeSource{
--- a/operator/internal/manifests/gateway.go
+++ b/operator/internal/manifests/gateway.go
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	tlsMetricsSercetVolume = "tls-metrics-secret"
+	tlsSercetVolume = "tls-secret"
 )
 
 // BuildGateway returns a list of k8s objects for Loki Stack Gateway
@@ -49,7 +49,7 @@ func BuildGateway(opts Options) ([]client.Object, error) {
 
 	if opts.Stack.Tenants != nil {
 		mode := opts.Stack.Tenants.Mode
-		if err := configureDeploymentForMode(dpl, mode, opts.Flags); err != nil {
+		if err := configureDeploymentForMode(dpl, mode, opts.Flags, opts.Name, opts.Namespace); err != nil {
 			return nil, err
 		}
 
@@ -58,6 +58,9 @@ func BuildGateway(opts Options) ([]client.Object, error) {
 		}
 
 		objs = configureGatewayObjsForMode(objs, opts)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	return objs, nil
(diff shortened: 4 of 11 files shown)


operator/bundle/manifests/loki-operator.clusterserviceversion.yaml
```diff
@@ -625,7 +625,7 @@ spec:
                   value: quay.io/observatorium/api:latest
                 - name: RELATED_IMAGE_OPA
                   value: quay.io/observatorium/opa-openshift:latest
-                image: quay.io/openshift-logging/loki-operator:v0.0.1
+                image: quay.io/mbouqsim/loki-operator:v0.0.1-460cfafc1
```
periklis, line 628: Keep an eye not commiting this

operator/bundle/manifests/loki-operator.clusterserviceversion.yaml
```diff
@@ -725,4 +725,4 @@ spec:
     name: gateway
   - image: quay.io/observatorium/opa-openshift:latest
     name: opa
-  version: 0.0.1
+  version: 0.0.1-460cfafc1
```
periklis, line 728: Keep an eye not commiting this

operator/hack/addons_ocp.yaml
```diff
@@ -118,7 +118,7 @@ metadata:
 data:
   promtail.yaml: |
     clients:
-      - url: http://lokistack-dev-gateway-http.openshift-logging.svc:8080/api/logs/v1/application/loki/api/v1/push
+      - url: https://lokistack-dev-gateway-http.openshift-logging.svc:8080/api/logs/v1/application/loki/api/v1/push
```
periklis, line 121: This probably does not work without:
```
# If connecting to a TLS server, configures how the TLS
# authentication handshake will operate.
tls_config:
  # The CA file to use to verify the server
  [ca_file: <string>]
```
See: https://grafana.com/docs/loki/latest/clients/promtail/configuration/#clients

periklis (changes requested):

operator/hack/addons_ocp.yaml
aminesnow: done

operator/internal/manifests/gateway.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	tlsMetricsSercetVolume = "tls-metrics-secret"
+	tlsSercetVolume = "tls-secret"
```
xperimental, line 23:
```suggestion
	tlsSecretVolume = "tls-secret"
```

xperimental (commented):
No proper review, just passing by.

operator/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Main
 
 - [6195](https://github.com/grafana/loki/pull/6195) **periklis**: Add ruler config support
+- [6288](https://github.com/grafana/loki/pull/6288) **aminesnow**: Expose only an HTTPS gateway when in openshift mode
```
periklis, line 4: Reminder before we squash & merge: We generally add for every new PR the changelog entry at the top. I believe yours got pushed down through rebasing onto main.

operator/internal/manifests/build_test.go
```diff
@@ -263,7 +263,7 @@ func TestBuildAll_WithFeatureFlags_EnableTLSServiceMonitorConfig(t *testing.T) {
 			continue
 		}
 
-		secretName := fmt.Sprintf("%s-http-metrics", name)
+		secretName := fmt.Sprintf("%s-http-tls", name)
```
periklis, line 266: Watch out, my PR #6224 also addresses this issue by removing `-metrics` entirely. FYI after re-working my PR a couple of times I err'd on naming the services `lokistack-dev-component-http` and `lokistack-dev-component-grpc`. You have two choices either you undo this renames in your PR or someone of us resolves the merge conflict. ;)

operator/internal/manifests/gateway.go
```diff
@@ -58,6 +58,9 @@ func BuildGateway(opts Options) ([]client.Object, error) {
 		}
 
 		objs = configureGatewayObjsForMode(objs, opts)
+		if err != nil {
+			return nil, err
+		}
```
periklis, line 63: I believe this err check is obsolete now as per `configureGatewayObjsForMode` not returning an error value anymore.

operator/internal/manifests/gateway_tenants.go
```diff
@@ -56,22 +56,29 @@ func ApplyGatewayDefaultOptions(opts *Options) error {
 	return nil
 }
 
-func configureDeploymentForMode(d *appsv1.Deployment, mode lokiv1beta1.ModeType, flags FeatureFlags) error {
+func configureDeploymentForMode(d *appsv1.Deployment, mode lokiv1beta1.ModeType, flags FeatureFlags, name, ns string) error {
```
periklis, line 59: nit. The code base refers to the stack identity fields using variable/argument names as `stackName` and `stackNs`.

operator/internal/manifests/gateway_tenants_test.go
```diff
@@ -184,6 +185,8 @@ func TestConfigureDeploymentForMode(t *testing.T) {
 		dpl   *appsv1.Deployment
 		want  *appsv1.Deployment
 	}
+	name := "test"
+	ns := "test-ns"
```
periklis, line 189: Although I understand that this aims to be a reusability win, I suggest we keep test cases below as independent as possible. Test cases should be isolated from each other even if that means that we have some duplication of inputs. That makes them side-effect free usually and in extend able to execute in parallel.

operator/internal/manifests/openshift/configure.go
```diff
@@ -104,6 +108,89 @@ func ConfigureGatewayDeployment(
...
+		if strings.HasPrefix(arg, "--web.healthchecks.url=") {
+			argI = i
+			break
+		}
+	}
+
+	if argI != -1 {
+		args[argI] = args[len(args)-1]
+		d.Spec.Template.Spec.Containers[gwIndex].Args = args[:len(args)-1]
+	}
```
periklis, line 125: I suggest to simplify this by replacing the value instead of remove/add-anew like this:

```go
for i, arg := range gwArgs {
	if strings.HasPrefix(arg, "--web.healthchecks.url=") {
		gwContainer.Args[i] = "YOUR_NEW_VALUE_HERE"
	}
}
```
WDYT?

periklis (changes requested):
Looks overall good to me :rocket: Only a couple of small improvements and we are set to go! Thank you for your contribution! Keep rocking!

operator/CHANGELOG.md
aminesnow: indeed

operator/internal/manifests/openshift/configure.go
aminesnow: that works too, but I did it this way to keep the mergo.Merge pattern. Either way is fine by me :)

operator/internal/manifests/gateway.go
aminesnow: nice catch!

operator/internal/manifests/gateway_tenants_test.go
aminesnow: makes sense

operator/internal/manifests/build_test.go
aminesnow: I can rollback my changes, but I do believe `tlsMetricsSercetVolume = "tls-metrics-secret"` should be `tlsSercetVolume = "tls-secret"` since we use it here for both metrics and main gateway. What do you think?

periklis: @aminesnow If #6224 is missing to rename the volume, then we should keep the name change to `tls-secret` in this PR.

operator/internal/manifests/openshift/configure.go
periklis: You can still keep the `mergo.Merge` pattern. You first compile the args list and then pass it to the container. Finally you `mergo.Merge`

periklis (commented):
Looks good to me!

operator/internal/manifests/openshift/configure.go
aminesnow: Sure, that's what I do here, but before doing that I delete the `--web.healthchecks.url=` entry from the args list, since doing `mergo.Merge` would just append it and not override it (ending up with two values `--web.healthchecks.url=`). 

Do you suggest that I override this value in the args directly, then compile the rest of the newer args without `--web.healthchecks.url=`? Because that works as well, but it wouldn't follow the `mergo.Merge` since we edit the `--web.healthchecks.url=` value directly instead of doing it via `mergo.Merge`. 
That is the only reason I'm deleting then re inserting the value here, otherwise I can just do it as you suggested.

operator/internal/manifests/build_test.go
periklis: Let's keep it as is the conflict is very limited for both of us.

operator/internal/manifests/openshift/configure.go
xperimental: I have a similar feeling to Peri regarding constructing the args and the multiple use of mergo. I think you can re-use the existing `gw*` variables to your advantage and avoid using multiple mergo-calls. I have [a suggestion](https://github.com/aminesnow/loki/compare/gateway-route-https...xperimental:smaller-configure) how this could look.

xperimental (changes requested):
This achieves the goal of having a TLS gateway when on OpenShift. I would have expected the "TLS-enabled gateway" to be an option that is also available "without OpenShift" and that the OpenShift-config would just set this option enabled.

Should we refactor the PR into that direction or is "TLS only with OpenShift tenancy" the way we want to go?

operator/internal/manifests/openshift/configure.go
```diff
@@ -36,12 +37,15 @@ var (
...
 // With this, the deployment will route authorization request to the OpenShift
 // apiserver through the sidecar.
+// This function also forces the use of a TLS connection for the gateway.
 func ConfigureGatewayDeployment(
 	d *appsv1.Deployment,
 	gwContainerName string,
 	sercretVolumeName, tlsDir, certFile, keyFile string,
 	caDir, caFile string,
 	withTLS, withCertSigningService bool,
+	secretName, serviceName, serverName string,
```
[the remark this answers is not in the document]
aminesnow: 1. The goal is to also use the Route CR and reencrypt to gateway. So it's pretty Openshift specific
2. indeed serviceName is unused! thnx

xperimental (approved):
Looks good. :+1: 

Only remaining question from my point-of-view is whether we need to update the documentation / examples and the provided Grafana configuration in this PR or if this is done later.

periklis (approved):

PR #6289, opened 2022-06-02
Add per-tenant cache metrics

dannykopping:
In order to improve & troubleshoot our cache performance, we need per-tenant metrics. Right now we track cache value size, hit/miss ratio, request latency, and requested key count on a global level. All of these metrics (except for request latency) are too coarse right now to be very useful. Request latency is irrelevant per tenant because we don't have a way to dedicate a caching backend per tenant, so this would be a very noisy per-tenant metric.

#6318

This setting is defaulted to `false` due to potential cardinality explosions.

This PR adds a `tenant` label for the following metrics:

- `loki_cache_value_size_bytes`
- `loki_cache_fetched_keys`
- `loki_cache_hits`

If one has a large number of active tenants, this can result in the creation of **many** more streams.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -1839,6 +1839,10 @@ the index to a backing cache store.
 # CLI flag: -<prefix>.cache.enable-fifocache
 [enable_fifocache: <boolean>]
 
+# Add tenant labels to cache-related metrics.
+# CLI flag: -<prefix>.cache.per-tenant-metrics
+[per_tenant_metrics: <boolean> | default = false]
+
 # The default validity of entries for caches unless overridden.
 # NOTE In Loki versions older than 1.4.0 this was "defaul_validity".
 # CLI flag: -<prefix>.default-validity
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -1839,7 +1839,8 @@ the index to a backing cache store.
 # CLI flag: -<prefix>.cache.enable-fifocache
 [enable_fifocache: <boolean>]
 
-# Add tenant labels to cache-related metrics.
+# Add tenant labels to cache-related metrics. Be mindful that this may create many more metric streams if a large
+# number of active tenants are using the cache.
 # CLI flag: -<prefix>.cache.per-tenant-metrics
 [per_tenant_metrics: <boolean> | default = false]
 
--- a/pkg/storage/chunk/cache/instrumented.go
+++ b/pkg/storage/chunk/cache/instrumented.go
@@ -3,6 +3,7 @@ package cache
 import (
 	"context"
 
+	"github.com/grafana/dskit/tenant"
 	ot "github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
 	otlog "github.com/opentracing/opentracing-go/log"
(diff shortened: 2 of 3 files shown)


docs/sources/configuration/_index.md
```diff
@@ -1833,6 +1833,10 @@ the index to a backing cache store.
 # CLI flag: -<prefix>.cache.enable-fifocache
 [enable_fifocache: <boolean>]
 
+# Add tenant labels to cache-related metrics.
+# CLI flag: -<prefix>.cache.per-tenant-metrics
+[per_tenant_metrics: <boolean> | default = false]
```
trevorwhitney, line 1838: WDYT about just always doing this based on if tenancy is enabled (ie. auth = true)? Curious how much cardinality explosion we're really protecting users from here, and we already have _sooooo_ many configs.

trevorwhitney (commented):
I like where this is headed :)

docs/sources/configuration/_index.md
dannykopping: You know it's rude to review draft PRs :stuck_out_tongue: but thanks for taking a look

Cardinality is something that I plan to address in the description of the PR, but finishing something else up first.

trevorwhitney: haha, it's like we're pairing...but asynchronously! yeah, I didn't take too critical an eye, but couldn't resist getting my opinions in early.

pkg/storage/chunk/cache/instrumented.go
```diff
@@ -116,3 +122,14 @@ func (i *instrumentedCache) Fetch(ctx context.Context, keys []string) ([]string,
...
 }
+
+func (i *instrumentedCache) getTenantID(ctx context.Context) string {
+	// if per-tenant metrics are disabled, set an empty label which will be ignored
+	if !i.perTenant {
+		return ""
+	}
+
+	// best effort: if tenant ID cannot be retrieved, set an empty label which will be ignored
+	userID, _ := tenant.TenantID(ctx)
```
DylanGuedes, line 133: double checking: don't you think it would be worth it to log the error returned from the tenantID() call here?

DylanGuedes (approved):

pkg/storage/chunk/cache/instrumented.go
dannykopping: It's a good question. I ignored the error here because operationally it's not of much use, and if we can't derive the tenant from the context then we'll have quite a few upstream errors to contend with.

DylanGuedes: got it, thanks!

owen-d (commented):
I'm not in favor of adding this due to cardinality concerns on large clusters. It will be helpful on smaller clusters, but at the expense of being a toggle away from a cardinality bomb on large ones :(. Instead, can we include tenant information in logs for later extraction via loki?

dannykopping:
> I'm not in favor of adding this due to cardinality concerns on large clusters. It will be helpful on smaller clusters, but at the expense of being a toggle away from a cardinality bomb on large ones :(. Instead, can we include tenant information in logs for later extraction via loki?

I'm fine with parking this for now :+1:
https://github.com/grafana/loki/pull/6317 gives us most of what we need for now, except for cache value sizes, but we can factor that in somehow into the logs if need to.

PR #6297, opened 2022-06-02
Only consider an interval to have possible expired chunks if it overlaps a delete.

MasslessParticle:
Before, we considered any interval to potentially have an expired chunk if the passed userID matched a user on a DeleteRequest. This PR adds the requirement that an interval must also overlap a delete request to count as having a potential expired chunk.

### DIFF
--- a/pkg/storage/stores/shipper/compactor/deletion/delete_requests_manager.go
+++ b/pkg/storage/stores/shipper/compactor/deletion/delete_requests_manager.go
@@ -203,13 +203,15 @@ func (d *DeleteRequestsManager) MarkPhaseFinished() {
 	}
 }
 
-func (d *DeleteRequestsManager) IntervalMayHaveExpiredChunks(_ model.Interval, userID string) bool {
+func (d *DeleteRequestsManager) IntervalMayHaveExpiredChunks(interval model.Interval, userID string) bool {
 	d.deleteRequestsToProcessMtx.Lock()
 	defer d.deleteRequestsToProcessMtx.Unlock()
 
 	if userID != "" {
 		for _, deleteRequest := range d.deleteRequestsToProcess {
-			if deleteRequest.UserID == userID {
+			if deleteRequest.UserID == userID &&
+				deleteRequest.StartTime <= interval.End &&
+				deleteRequest.EndTime >= interval.Start {
 				return true
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/storage/stores/shipper/compactor/deletion/delete_requests_manager.go
```diff
@@ -203,13 +203,15 @@ func (d *DeleteRequestsManager) MarkPhaseFinished() {
...
-func (d *DeleteRequestsManager) IntervalMayHaveExpiredChunks(_ model.Interval, userID string) bool {
+func (d *DeleteRequestsManager) IntervalMayHaveExpiredChunks(interval model.Interval, userID string) bool {
 	d.deleteRequestsToProcessMtx.Lock()
 	defer d.deleteRequestsToProcessMtx.Unlock()
 
 	if userID != "" {
 		for _, deleteRequest := range d.deleteRequestsToProcess {
-			if deleteRequest.UserID == userID {
+			if deleteRequest.UserID == userID &&
+				deleteRequest.StartTime <= interval.End &&
```
trevorwhitney, line 213: In your comment on the PR you mention if an interval overlaps a delete request, but this logic requires the interval to be fully within the delete request time period (so excluding an overlap at the start or end of the interval). Is that intended? Would there be a situation where we'd want an `||` here instead?

trevorwhitney, line 213: In your comment on the PR you mention if an interval overlaps a delete request, but this logic requires the interval to be fully within the delete request time period (so excluding an overlap at the start or end of the interval). Is that intended? Would there be a situation where we'd want an `||` here instead?

trevorwhitney (commented):
Looks good, just one small question about the "overlap" logic.

pkg/storage/stores/shipper/compactor/deletion/delete_requests_manager.go
MasslessParticle: This actually handles overlaps because any overlapping interval will always start before the end of the delete and end after the start of the delete. Fully overlapping is just a special case. I've  included a picture

Untitled-2022-05-25-1713

trevorwhitney (approved):
That illustration helped a bunch, thanks! LGTM!

slim-bean (approved):
LGTM

PR #6305, opened 2022-06-03
do not initialize tsdb index store when using index gateway client for queries

sandeepsukhani:
When Loki is configured to run the index gateway client for doing `tsdb` queries, we also initialize `tsdb` index store instance just to satisfy the `IndexStore` interface [here](https://github.com/grafana/loki/blob/a2898295040411e8c375be98ed3e4e167724f45e/pkg/storage/stores/series/series_index_gateway_store.go#L34), which is just a fallback for keeping index gateway client and store backwards compatible.
This causes us to un-necessarily run an instance of `indexshipper` which would try to sync the files locally if query readiness is configured.

This PR changes the code to make the `fallbackStore` for `NewIndexGatewayClientStore` optional and keep it `nil` when initializing it for `tsdb` index store.


owen-d (approved):
After the build passes (bit of linting/etc) LGTM

PR #6309, opened 2022-06-03
exposes line length hist on distributors 

owen-d:
As part of the work to better understand line length limits in Loki, this adds a new histogram for them so we can better understand distributions of line lengths coming in.
Also fixes some metric instantiations that were missed in the `init` function by preferring `promauto`.


dannykopping (approved):
LGTM

slim-bean (approved):
LGTM

Issue #6283, opened 2022-06-01
[loki] Further secure TLS communications


saroshali-dbx:
**Is your feature request related to a problem? Please describe.**
Currently when using TLS, Loki will accept requests from any client that has a certificate signed by our Certificate Authority. For enforcing the deny-by-default security requirement, we need to check the common name against the expected value and reject requests that do not match. As such, I'd like to see custom server certificate validation supported as part of Loki.

**Describe the solution you'd like**
I'd like to be able to pass a flag to loki, such as `-cert-allowed-cn`, that can be used to create a custom `VerifyPeerCertificate` function as part of the crypto/tls package. All this function needs to do is verify that the seen common-name is the same as the expected common-name. 

**Describe alternatives you've considered**
Within our organization, we (ab)use custom gRPC certs signed from a custom authoritative root CA. If we wanted this behavior to work without any changes to Loki we'd have to create a separate set of certificates signed by another root CA. Managing and securing CAs is already hard enough, so we'd prefer to not have to add another one. 

**Additional context**
I noticed that the gRPC/HTTP server implementations for loki are lifted fromhttps://github.com/weaveworks/common/blob/master/server/server.go, likely a new parameter would need to tunneled and implemented there. 

I'm willing to attempt an implementation if Loki maintainers think this is a good idea.

PR #6310, opened 2022-06-03
loki-canary: Add support for client-side TLS certs for Loki connection

### LINKED ISSUE #6295: [loki-canary] Add ability to load client certs for reader TLS connection to Loki
When deploying Loki in an environment secured by [mTLS](https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/) it is necessary to have loki-canary support loading client-side certificates. In this case, requests to Loki that do not have an authorized subject name in the client side cert would be dropped (dependent on #6283).

Add a set of flags to loki-canary that will allow specifying filenames for a PEM encoded cert, key, and CA. Then use these to create a transport with `TLSClientConfig` for the `net/http` client to override the default `DefaultTransport` when client certs are provided.

A proxy sidecar could be deployed that would handle the TLS connection to Loki and communicate to loki-canary over http.

**Additional context**
This is part of a larger effort to comply with a Zero Trust security model.

chodges15:
Support for loki-canary in environments using mTLS, in anticipation of deny-by-default Loki server authentication coming for #6283.

Fixes #6295 

**Checklist**

### DIFF
--- a/cmd/loki-canary/main.go
+++ b/cmd/loki-canary/main.go
@@ -3,15 +3,18 @@ package main
 import (
 	"flag"
 	"fmt"
-	"net/http"
 	"os"
-	"os/signal"
 	"strconv"
 	"sync"
 	"syscall"
 	"time"
 
+	"crypto/tls"
+	"net/http"
+	"os/signal"
+
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/common/config"
 	"github.com/prometheus/common/version"
 
 	"github.com/grafana/loki/pkg/canary/comparator"
@@ -36,7 +39,10 @@ func main() {
 	sValue := flag.String("streamvalue", "stdout", "The unique stream value for this instance of loki-canary to use in the log selector")
 	port := flag.Int("port", 3500, "Port which loki-canary should expose metrics")
 	addr := flag.String("addr", "", "The Loki server URL:Port, e.g. loki:3100")
-	tls := flag.Bool("tls", false, "Does the loki connection use TLS?")
+	useTls := flag.Bool("tls", false, "Does the loki connection use TLS?")
+	certFile := flag.String("cert-file", "", "Client PEM encoded X.509 certificate for optional use with TLS connection to Loki")
+	keyFile := flag.String("key-file", "", "Client PEM encoded X.509 key for optional use with TLS connection to Loki")
+	caFile := flag.String("ca-file", "", "Client certificate authority for optional use with TLS connection to Loki")
 	user := flag.String("user", "", "Loki username.")
 	pass := flag.String("pass", "", "Loki password.")
 	tenantID := flag.String("tenant-id", "", "Tenant ID to be set in X-Scope-OrgID header.")
@@ -83,6 +89,22 @@ func main() {
 		os.Exit(1)
 	}
 
+	var tlsConfig *tls.Config
+	tc := config.TLSConfig{}
+	if *certFile != "" || *keyFile != "" || *caFile != "" {
+		tc.CAFile = *caFile
+		tc.CertFile = *certFile
+		tc.KeyFile = *keyFile
+		tc.InsecureSkipVerify = false
+
+		var err error
+		tlsConfig, err = config.NewTLSConfig(&tc)
+		if err != nil {
+			_, _ = fmt.Fprintf(os.Stderr, "TLS configuration error: %s\n", err.Error())
+			os.Exit(1)
+		}
+	}
+
 	sentChan := make(chan time.Time)
 	receivedChan := make(chan time.Time)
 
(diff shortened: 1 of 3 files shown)


DylanGuedes (changes requested):
one more, sorry

pkg/canary/reader/reader.go
```diff
@@ -442,6 +459,35 @@ func (r *Reader) closeAndReconnect() {
 	}
 }
 
+// httpClient uses the config in Reader to return a http client.
+// http.DefaultClient will be returned in the case that the connection to Loki is http or TLS without client certs.
+// For the mTLS case, return a http.Client configured to use the client side certificates.
+func (r *Reader) httpClient() (*http.Client, error) {
+	if r.tlsConfig == nil || r.tls == false {
+		return http.DefaultClient, nil
```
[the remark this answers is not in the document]
chodges15: The most commonly used config when TLS is enabled will not include client certs, updated main to enforce tls is enabled when client TLS config is used. Updated in latest push to make this more clear.

trevorwhitney (approved):
LGTM! definitely something we need!

chodges15:
@DylanGuedes and @trevorwhitney thanks for the quick review, I got the CLA signed  (and updated all commits with same username) and the PR should now be ready to merge.

pkg/canary/reader/reader.go
```diff
@@ -442,6 +459,35 @@ func (r *Reader) closeAndReconnect() {
...
+	}, nil
+}
+
+// webSocketDialer creates a dialer for the web socket connection to Loki
+// websocket.DefaultDialer will be returned in the case that the connection to Loki is http or TLS without client certs.
+// For the mTLS case, return a websocket.Dialer configured to use client side certificates.
+func (r *Reader) webSocketDialer() *websocket.Dialer {
+	dialer := websocket.DefaultDialer
+	if r.clientTLSConfig != nil {
+		dialer.TLSClientConfig = r.clientTLSConfig
```
owen-d, line 486: We probably shouldn't be mutating `websocket.DefaultDialer` in this library package.

Instead, let's create a new one like:
```
var dialer = &Dialer{
	Proxy:            http.ProxyFromEnvironment,
	HandshakeTimeout: 45 * time.Second,
	TLSClientConfig = r.clientTLSConfig,
}
```

owen-d (commented):
LGTM, thanks @chodges15! Just one suggestion.

cmd/loki-canary/main.go
```diff
@@ -83,6 +89,26 @@ func main() {
...
+		tc.InsecureSkipVerify = false
+
+		var err error
+		tlsConfig, err = config.NewTLSConfig(&tc)
+		if err != nil {
+			_, _ = fmt.Fprintf(os.Stderr, "TLS configuration error: %s\n", err.Error())
+			os.Exit(1)
+		}
+	}
+
```
kavirajk, line 111: Correct me if I'm missing anything. We could simplify this initialization code bit.

the gist is 
1. to first use `if useTLS ` to do all this and skip everything completely (even before checking for `*certFile, *keyFile, *caFile`.
2. And trusting `config.NewTLSConfig` for checking if all the files are invalid (even if its empty)

Something simple like
```go
	if useTLS {
		tls, err := config.NewTLSConfig(
			&config.TLSConfig{
				CAFile, *caFile,
				CertFile, *certFile,
				KeyFile, *keyFile,
				InsecureSkipVerify: false,
			},
		)
		if err != nil {
			// print error
			os.Exit(1)
		}
	}
}
```
No need to check for any empty string on those cert files. `NewTLSConfig()` should return error in that case correct?

pkg/canary/reader/reader.go
```diff
@@ -280,7 +292,11 @@ func (r *Reader) Query(start time.Time, end time.Time) ([]time.Time, error) {
 	}
 	req.Header.Set("User-Agent", userAgent)
 
-	resp, err := http.DefaultClient.Do(req)
+	httpClient, err := r.httpClient()
+	if err != nil {
+		return nil, err
```
kavirajk, line 297: I would highly recommend wrapping the error here. (and also on the `QueryCountOverTime`) with some context.

Rationale is I see we use it on both `Query` and `QueryCountOverTime` and it can confuse the consumer of the error, where exactly it is coming from.

cmd/loki-canary/main.go
chodges15: The main reason for doing this was to ensure valid configuration. Setting the client certs requires `-tls` to be set otherwise it is a noop. I actually made this mistake in my own testing and was glad to have the error message. On the other hand, when setting `-tls` it does not require setting client certs.

pkg/canary/reader/reader.go
chodges15: My understanding is that `dialer` is a copy of `websocket.DefaultDialer` and there is no mutation happening outside this package. However, I like your way better as it keeps me from having to do local mutation.

kavirajk:
@chodges15 also addressed your suggestion @owen-d. I think it's good to go. Merging it.

PR #6324, opened 2022-06-07
Remove the stream lag metric.

cstyan:
Removes the stream lag metric since it's not possible to get a correct value and also not cause issues with metrics registration or collection until we have the WAL


clients/cmd/promtail/main.go
```diff
@@ -111,7 +111,11 @@ func main() {
 		}
 	}
 
-	clientMetrics := client.NewMetrics(prometheus.DefaultRegisterer, config.Config.Options.StreamLagLabels)
+	clientMetrics := client.NewMetrics(prometheus.DefaultRegisterer)
+	if config.Options.StreamLagLabels.String() != "" {
+		level.Error(util_log.Logger).Log("msg", "client config stream_lag_labels is deprecated and the associated metric has been removed", "stream_lag_labels", config.Options.StreamLagLabels.String())
```
DylanGuedes, line 116: maybe adding the PR link will be helpful as in the description you included the reasoning why it was removed
```suggestion
		level.Error(util_log.Logger).Log("msg", "client config stream_lag_labels is deprecated and the associated metric has been removed. See https://github.com/grafana/loki/pull/6324.", "stream_lag_labels", config.Options.StreamLagLabels.String())
```

DylanGuedes (approved):
lgtm, just a nit

clients/cmd/promtail/main.go
cstyan: I think this is something we would include in changelogs and release notes rather than an error message

DylanGuedes: good point

MichelHollands (approved):
A rebase is needed but otherwise LGTM

jeschkies:
@cstyan could you merge main so this can land?

PR #6325, opened 2022-06-07
promtail wal

cstyan:
Builds on @cyriltovena and other's work to implement a WAL for promtail. Still WIP.


clients/pkg/promtail/client/batch.go
```diff
@@ -23,37 +36,74 @@ type batch struct {
...
+		err := b.wal.Log(walRecord)
+		if err != nil {
+			fmt.Println("err: ", err)
+		}
+	}()
+
+	var fp uint64
+	lbs := labels.FromMap(util.ModelLabelSetToMap(entry.Labels))
+	sort.Sort(lbs)
+	fp, hashBuffer = lbs.HashWithoutLabels(hashBuffer, []string(nil)...)
```
cyriltovena, line 85: Not sure the hashbuffer can be used safely with globals as multiple clients could collide in usage.

clients/pkg/promtail/client/client.go
```diff
@@ -229,6 +241,66 @@ func NewWithTripperware(metrics *Metrics, cfg Config, streamLagLabels []string,
...
+					entry.Labels = l
+					for _, e := range samples.Entries {
+						entry.Entry = e
+						b.replay(entry)
+					}
+
+				}
+			}
+		}
+		c.sendBatch(tenantID, b)
```
cyriltovena, line 299: Not sure if we need to send, the client would most likely do that, every x or when an entry is pushed although this means you need to keep appending to the old wal.

I'm ok to start like this though.

cyriltovena (commented):
This looks good so far let's finish and polish.

clients/pkg/promtail/client/client.go
cstyan: Not sure I understand your comment.

Right now writing to the WAL is behind en-queuing a log line within the batch. Here I'm using `batch.replay` to add a sample to the batch without logging it to the WAL, since we're reading an existing WAL segment. And then we're sending the batch. This is essentially set up the same way Prometheus remote write is, we replay the existing WAL before accepting new samples to be sent. The thing I am not doing here yet that I should be is checking the batch size vs the max batch size and sending if we would hit that value. In theory that shouldn't happen right now, since when we swap to a new batch in the happy path we get a new WAL file anyways, but we should be safe.

The thing we should do in the future is decouple the scraping of logs and the sending of them over the client connection by having them be written to the WAL in the middle. Clients would then have tailing code to just tail the WAL segment files.

clients/pkg/promtail/client/batch.go
cstyan: We weren't actually using the hash for anything so I've removed it. If we do end up needing it we can use a `sync.Pool`

cstyan:
still plan to work on this soon

chaudum:
This can be closed, because WAL support in Promtail is tracked [here](https://github.com/grafana/loki/issues/8197) and there are different PRs for the implementation.

PR #6329, opened 2022-06-07
Tsdb/index sampling

owen-d:
This PR lays the groundwork for index sampling in TSDB. To do so, it creates a new struct,
```go
type Stats struct {
	Streams uint64
	Chunks  uint64
	Bytes   uint64
	Entries uint64
}
```

A set of `(user, from, through, matchers)` will resolve to the approximation of underlying `streams`, `chunks`, `bytes`, and `entries`.

In the short term, this will enable dynamic sharding in the query planner. In the long term, we can expose this as an endpoint and use it to power a host of new features such as cardinality analysis and informing users how much data a query will touch before running it.

There will be a followup PR which wires this into the index-gateway grpc layer.


slim-bean (approved):
LGTM

PR #6335, opened 2022-06-08
Define a reason why the fifocache has evicted an item

dannykopping:

This PR adds an eviction reason for the fifocache, which will help an operator know if an entry was evicted because the TTL expired vs the cache is full.

N/A

We use a fifocache when running boltdb-shipper for storing ingesters' index caches locally. We do this because we cannot share the cache between ingesters, to prevent missing out chunks that may not exist in other ingesters.
https://github.com/grafana/loki/blob/main/pkg/loki/modules.go#L414-L425

The cache is configured with a fixed size of 200MB, and is currently not configurable; we could consider making this configurable however even in our biggest clusters we're only seeing ~11MB of use currently.

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/storage/chunk/cache/fifo_cache.go
+++ b/pkg/storage/chunk/cache/fifo_cache.go
@@ -86,13 +86,20 @@ type FifoCache struct {
 
 	entriesAdded    prometheus.Counter
 	entriesAddedNew prometheus.Counter
-	entriesEvicted  prometheus.Counter
+	entriesEvicted  *prometheus.CounterVec
 	entriesCurrent  prometheus.Gauge
 	totalGets       prometheus.Counter
 	totalMisses     prometheus.Counter
 	staleGets       prometheus.Counter
 	memoryBytes     prometheus.Gauge
 }
+
+const (
+	expiredReason string = "expired"
+	fullReason           = "full"
+	tooBigReason         = "too big"
+	stoppedReason        = "stopped"
+)
 
 type cacheEntry struct {
 	updated time.Time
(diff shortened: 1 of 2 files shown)


pkg/storage/chunk/cache/fifo_cache.go
```diff
@@ -86,14 +86,21 @@ type FifoCache struct {
...
 	totalMisses     prometheus.Counter
 	staleGets       prometheus.Counter
 	memoryBytes     prometheus.Gauge
 }
 
+const (
+	expiredReason string = "expired" //nolint:staticcheck
+	fullReason           = "full"
+	tooBigReason         = "too big"
+	stoppedReason        = "stopped"
```
kavirajk, line 101:
```suggestion
	expiredReason string = "object-expired" //nolint:staticcheck
	fullReason           = "cache-full"
	tooBigReason         = "object-too-big"
	stoppedReason        = "cache-stopped"
```

kavirajk (commented):
LGTM :+1: 

Minor suggestions on `reason` constants.

pkg/storage/chunk/cache/fifo_cache.go
dannykopping: Thanks for the review :+1: 

I don't think these suggestions make things much more clear TBH.
I agree on this one though `tooBigReason = "object-too-big"`.
Can you expand on why you think the others make things more clear pls?

pkg/storage/chunk/cache/fifo_cache.go
```diff
@@ -86,14 +86,21 @@ type FifoCache struct {
...
 	totalMisses     prometheus.Counter
 	staleGets       prometheus.Counter
 	memoryBytes     prometheus.Gauge
 }
 
+const (
+	expiredReason string = "expired" //nolint:staticcheck
+	fullReason           = "full"
+	tooBigReason         = "object too big"
+	stoppedReason        = "stopped"
```
sandeepsukhani, line 101: `stopped` doesn't give the operator much context, especially the ones who do not go and check the code. I think we should not increment the metric when we are evicting items while stopping. It doesn't make a lot of sense to me.

sandeepsukhani (approved):
LGTM with a minor non-blocking suggestion

pkg/storage/chunk/cache/fifo_cache.go
dannykopping: I thought the same actually, cool. Will remove :+1:

pkg/storage/chunk/cache/fifo_cache.go
kavirajk: if `stopped` is removed. Then other looks ok :)

kavirajk (approved):

pkg/storage/chunk/cache/fifo_cache.go
dannykopping: Discussed offline, Kavi's satisfied with my change

PR #6338, opened 2022-06-08
TSDB/index-sampling-wiring

owen-d:
This PR wires together the new `Stats` index methods which allow us to query an index alone for data statistics by connecting it to our store+grpc patterns.

There will be a followup which uses this for query planning.


sandeepsukhani (approved):
LGTM

PR #6345, opened 2022-06-08
Fix max series test loop that returns early

chroberts:

This PR fixes a test. Before the PR, the `TestEngine_MaxSeries` test `return`s after the first test and prevents the other tests from running.

Fixes #: N/A

Adding `t.Run` allows using the `return`, but I went ahead and also removed the `return` and added an `else` to make it more clear. This would have prevented this and may protect against future moves/refactors.

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/logql/engine_test.go
+++ b/pkg/logql/engine_test.go
@@ -2168,21 +2168,23 @@ func TestEngine_MaxSeries(t *testing.T) {
 		{`count_over_time({app="foo|bar"} |~".+bar" [1m])`, logproto.BACKWARD, true},
 		{`avg(count_over_time({app=~"foo|bar"} |~".+bar" [1m]))`, logproto.FORWARD, false},
 	} {
-		q := eng.Query(LiteralParams{
-			qs:        test.qs,
-			start:     time.Unix(0, 0),
-			end:       time.Unix(100000, 0),
-			step:      60 * time.Second,
-			direction: test.direction,
-			limit:     1000,
+		t.Run(test.qs, func(t *testing.T) {
+			q := eng.Query(LiteralParams{
+				qs:        test.qs,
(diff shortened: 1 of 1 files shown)


pkg/logql/engine_test.go
```diff
@@ -2168,21 +2168,23 @@ func TestEngine_MaxSeries(t *testing.T) {
...
+				require.True(t, errors.Is(err, logqlmodel.ErrLimit))
+			} else {
+				require.Nil(t, err)
+			}
 		})
-		_, err := q.Exec(user.InjectOrgID(context.Background(), "fake"))
-		if test.expectLimitErr {
-			require.NotNil(t, err)
-			require.True(t, errors.Is(err, logqlmodel.ErrLimit))
-			return
```
DylanGuedes, line 2183: hmm if you modify the return to be a `continue` the wrapping inside the t.Run() isn't necessary I think. Do you mind double checking it?

DylanGuedes (changes requested):

pkg/logql/engine_test.go
chroberts: Yes, that is correct. Isn't `t.Run` preferred when making derived tests?

DylanGuedes: it is, you're right. I thought that you wrapped it as a solution to the early return but you actually did it to improve things.

DylanGuedes (approved):
lgtm

pkg/logql/engine_test.go
cstyan: > Isn't t.Run preferred when making derived tests?

generally yes, it makes running one of the specific sub tests possible

cyriltovena (approved):
LGTM

PR #6346, opened 2022-06-08
[logql] don't try to execute a modified function that is not allowed by the AST

cstyan:
The original panic was not useful, the change to the `clone` function and it's usage will ensure that in future cases we at least get an error message along with the panic stack trace. This PR also calls the `validate` function on the modified range query to ensure it's still allowed. In the case of the panic we saw, `unwrap` is not allowed with `count_over_time`, when in the original query we had `rate` with `unwrap` which is allowed.


cyriltovena (approved):
LGTM

PR #6347, opened 2022-06-08
Tsdb/index sampling endpoint

owen-d:
Wires up a new endpoint (``/loki/api/v1/index/stats`) for querying the index with `([]matcher, from, through)`:

```
$ curl -s localhost:3100/loki/api/v1/index/stats?query=%7Bfilename%3D~%22.%2B%22%7D | jq '.'
{
  "streams": 2,
  "chunks": 17,
  "bytes": 0, # guessing this is zero b/c none of my test data rounded up to a KB for inclusion here
  "entries": 794
}
```


pkg/storage/stores/tsdb/multitenant.go
```diff
@@ -29,8 +29,8 @@ func NewMultiTenantIndex(idx Index) *MultiTenantIndex {
 
 func withTenantLabelMatcher(userID string, matchers []*labels.Matcher) []*labels.Matcher {
 	cpy := make([]*labels.Matcher, len(matchers)+1)
-	copy(cpy, matchers)
-	cpy = append(cpy, labels.MustNewMatcher(labels.MatchEqual, TenantLabel, userID))
+	cpy[0] = labels.MustNewMatcher(labels.MatchEqual, TenantLabel, userID)
```
DylanGuedes, line 32: do you mind explaining this change? I see that you're putting the tenantlabelequal matcher before all others but I didn't understand why

DylanGuedes (approved):
lgtm! just a question

pkg/storage/stores/tsdb/multitenant.go
owen-d: This caught me up. Originally it creates `ln + 1` length, meaning the last one is `nil`. Then, I append to it, so it looks like `[a, b, nil, c]`, which causes a panic :(

The rationale for doing it specifically like this though is:
1) create it at the correct length (matchers+1) because we _know_ what the desired length is and we don't want to incur unncessary slice growing allocations (possibly needing to reallocate the entire length) if we had a `len(matchers)` buffer and needed to allocate new memory for the additional slot.
2) set the first position to the tenant label.
3)  We now know the remainder of the slice is exactly the length of `matchers`, so we can copy the rest

DylanGuedes: Got it, thanks! I'll pay attention to that to: prealloc with n+1 in conjunction + appendings is dangerous.

slim-bean (approved):
LGTM very exciting

PR #6349, opened 2022-06-09
Update loki's default port to 3100

simonswine:
I was suprised this was not already the case. I think it would be a good change to make, but I am also contious of the breaking nature of this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,7 @@
 * [5777](https://github.com/grafana/loki/pull/5777) **tatchiuleung**: storage: make Azure blobID chunk delimiter configurable
 * [5650](https://github.com/grafana/loki/pull/5650) **cyriltovena**: Remove more chunkstore and schema version below v9
 * [5643](https://github.com/grafana/loki/pull/5643) **simonswine**: Introduce a ChunkRef type as part of logproto
+* [6349](https://github.com/grafana/loki/pull/6349) **simonswine**: Updates default HTTP listen port from 80 to 3100, make sure to configure the port explicitly if you are using port 80.
 #### Promtail
 
 ##### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,7 +85,7 @@
 * [5777](https://github.com/grafana/loki/pull/5777) **tatchiuleung**: storage: make Azure blobID chunk delimiter configurable
 * [5650](https://github.com/grafana/loki/pull/5650) **cyriltovena**: Remove more chunkstore and schema version below v9
 * [5643](https://github.com/grafana/loki/pull/5643) **simonswine**: Introduce a ChunkRef type as part of logproto
-* [6349](https://github.com/grafana/loki/pull/6349) **simonswine**: Updates default HTTP listen port from 80 to 3100, make sure to configure the port explicitly if you are using port 80.
+* [6349](https://github.com/grafana/loki/pull/6349) **simonswine**: Update the default HTTP listen port from 80 to 3100. Make sure to configure the port explicitly if you are using port 80.
 #### Promtail
 
 ##### Enhancements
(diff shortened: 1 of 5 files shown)


dannykopping (commented):
Not sure how I feel about this.

On the one hand, I like it (except for the fact that [3100 is already registered](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=3100)) and a lot of documentation refers to this port
But on the other hand I don't see much value in changing it at this stage
But on the other other hand we expose port 3100 in our docker image, so who the hell knows :shrug: :smiling_face_with_tear: 

I think this could cause trouble as you say, for relatively meagre gains.

chaudum:
IMO, everything that help simplifying the config is a win, of course mainly for new users (think adoption), since existing users already have the overrides in place.

Regarding the Docker image: Yes, we `EXPOSE 3100`, but it still requires to set the HTTP port to 3100 in the config that needs to be provided. This is already inconsistent!

trevorwhitney (commented):
This is a good config simplification, I'm on board with it. @dannykopping are you thinking it's going to break a lot of existing installations because they're relying on this being port 80? I'm pretty sure we have this set in all our of deployments, as well as the jsonnet and helm libraries with publish.

cyriltovena (approved):
LGTM

dannykopping:
> This is a good config simplification, I'm on board with it. @dannykopping are you thinking it's going to break a lot of existing installations because they're relying on this being port 80? I'm pretty sure we have this set in all our of deployments, as well as the jsonnet and helm libraries with publish.

Not overly worried, let's see what happens and mention it in our changelog/upgrade guide before merging pls @simonswine

dannykopping (approved):

CHANGELOG.md
```diff
@@ -85,6 +85,7 @@
 * [5777](https://github.com/grafana/loki/pull/5777) **tatchiuleung**: storage: make Azure blobID chunk delimiter configurable
 * [5650](https://github.com/grafana/loki/pull/5650) **cyriltovena**: Remove more chunkstore and schema version below v9
 * [5643](https://github.com/grafana/loki/pull/5643) **simonswine**: Introduce a ChunkRef type as part of logproto
+* [6349](https://github.com/grafana/loki/pull/6349) **simonswine**: Updates default HTTP listen port from 80 to 3100, make sure to configure the port explicitly if you are using port 80.
```
dannykopping, line 88: Thanks for adding this.
Can you please add a section in `docs/sources/upgrading/_index.md` about this? This could cause some confusion in the upgrade path

KMiller-Grafana, line 88:
```suggestion
* [6349](https://github.com/grafana/loki/pull/6349) **simonswine**: Update the default HTTP listen port from 80 to 3100. Make sure to configure the port explicitly if you are using port 80.
```

docs/sources/upgrading/_index.md
```diff
@@ -38,6 +38,10 @@ The output is incredibly verbose as it shows the entire internal config struct u
 
 This value now defaults to `loki`, it was previously set to `cortex`. If you are relying on this container name for your chunks or ruler storage, you will have to manually specify `-azure.container-name=cortex` or `-ruler.storage.azure.container-name=cortex` respectively.
 
+#### Default value for `server.http-listen-port` changed
+
+This value now defaults to `3100` so the Loki process doesn't require special privileges anymore (before it had been set to port `80`, which is a privileged port). In case you need Loki to listen on port `80` you can set it back to the previous default using `-server.http-listen-port=80`.
```
KMiller-Grafana, line 43:
```suggestion
This value now defaults to 3100, so the Loki process doesn't require special privileges. Previously, it had been set to port 80, which is a privileged port. If you need Loki to listen on port 80, you can set it back to the previous default using `-server.http-listen-port=80`.
```

KMiller-Grafana (approved):

PR #5013, opened 2021-12-29
[new feature] logql: extrapolate unwrapped rate function

liguozhong:
ref issue:https://github.com/grafana/loki/issues/1136
doc: https://grafana.com/docs/loki/latest/logql/metric_queries/ 
rate function doc:  rate(unwrapped-range): calculates per second rate of all values in the specified interval.

logql: rate({_source="vector-internal-metrics" }|json|name="component_received_event_bytes_total"|tags_component_id="service_metrics"|unwrap counter_value [1m])

image

image
image

qps = 20,000 /s

PR #6361, opened 2022-06-10
Revert implementation of unwrapped rate aggregation (RFC #6351)

### LINKED ISSUE #6351: [RFC]: Sum values in unwrapped rate aggregation instead of treating them as counter

When working on #6344 ([#6348#discussion_r893370459](https://github.com/grafana/loki/pull/6348#discussion_r893370459)) I discovered that the unwrapped rate aggregation (`rate({} | unwrap x [$__interval]`) behaves differently than expected.

The documentation on [unwrapped range aggregations](https://grafana.com/docs/loki/latest/logql/metric_queries/#unwrapped-range-aggregations) states:

> `rate(unwrapped-range)`: calculates per second rate of all values in the specified interval.

This clearly suggests that the individual values extracted from the log lines in a time range are summed up and divided by the range to get the rate per second.

The initial implementation before #5013 also was like that.

https://github.com/grafana/loki/blob/a046d7926122cef1b619d89bfcf15403e14e0de3/pkg/logql/functions.go#L126-L137

For example, executing `rate({} | logfmt | unwrap subqueries [10s])` on following logs lines:

```
ts | line
---|---------------------------------------------------
0  | msg="something regarding subqueries" subqueries=4
1  | msg="something regarding subqueries" subqueries=3
2
3  | msg="something regarding subqueries" subqueries=5
4  | msg="something regarding subqueries" subqueries=2
5  | msg="something regarding subqueries" subqueries=8
6
7
8  | msg="something regarding subqueries" subqueries=2
9  | msg="something regarding subqueries" subqueries=3
```

One would expect to get as result `2.7`, based on `(4+3+5+2+8+2+3)/10`.
However, what you actually get is `1.6`, based on `(3-4+4+5+8)/10`. This is due to the implementation of `rate()` when using `Counter` type metrics, that has been copied from Prometheus.

https://github.com/grafana/loki/blob/caf7dde32be8b085549273fc6c238e3b6b656a3a/pkg/logql/range_vector.go#L231-L256

**However, Loki does not have the notion of "counters".**

The reason why the implementation was changed, was because of a tiny subset of common use of the `rate()` aggregation, where the value in the log lines acts as a "counter" (which can reset to 0), as explained in the PR description of #5013.

I propose to revert the implementation of the rate to the previous behaviour.
Additionally, to cover the special use-case of a counter value, LogQL can provide a new aggregation function `counter_rate()`, which takes an unwrapped range as input and treats it as a counter.

Example:

```
counter_rate({app="foo"} | logfmt | unwrap counter [1m])
```

Leave the implementation as is and change the documentation. This will probably cause a lot of confusion to people how want to rate "normal" values.

chaudum:

This PR implements the RFC described in #6351 

There are two commits. The first one reverts `rate()` to its previous implementation prior to #5013 That means it calculates the per-second rate from the sum of all extracted values. The second commit introduces a new `rate_counter()` aggregation function that treats the extracted values as "counter" and calculates the rate from the average increase of the value.

Fixes #6351

**Checklist**
- [x] Documentation added
- [x] Tests updated


chaudum:
@liguozhong Wanted to give you a heads up because it probably affects you.

liguozhong:
> @liguozhong Wanted to give you a heads up because it probably affects you.



> @liguozhong Wanted to give you a heads up because it probably affects you.

Thanks , I got it

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## Main
 
-* [6099](https://github.com/grafana/loki/pull/6099/files) **cstyan**: Drop lines with malformed JSON in Promtail JSON pipeline stage
+* [6361](https://github.com/grafana/loki/pull/6361) **chaudum**: Add new unwrapped range aggregation `rate_counter()` to LogQL
```
DylanGuedes, line 3: do you mind moving these PR entries to `Unreleased/All Changes/Loki/Enhancements (and Changes)`? At some point we started to add them here in the main section but it is wrong I think.

docs/sources/upgrading/_index.md
```diff
@@ -31,9 +31,15 @@ The output is incredibly verbose as it shows the entire internal config struct u
 
 ## Main / Unreleased
 
-
 ### Loki
 
+#### Implementation of unwrapped `rate` aggregation changed
+
+The implementation of the `rate()` aggregation function changed back to the previous implemention prior to [#5013](https://github.com/grafana/loki/pulls/5013).
```
DylanGuedes, line 38: (previous link not working)
```suggestion
The implementation of the `rate()` aggregation function changed back to the previous implemention prior to [#5013](https://github.com/grafana/loki/pull/5013).
```

DylanGuedes (commented):
Few nits.

Btw, do you mind renaming the PR title to something like what you have in the CHANGELOG? In the changelog it is described as "Sum values in unwrapped rate aggregation instead of treating them as counter".

chaudum:
> Few nits.
> 
> Btw, do you mind renaming the PR title to something like what you have in the CHANGELOG? In the changelog it is described as "Sum values in unwrapped rate aggregation instead of treating them as counter".

Let me make two separate pull requests. Then the PR title also matches the changes.

trevorwhitney (commented):
so, I don't fully understand this PR. for one, why are all the values we assert against in the tests changing? does this actually change computation behavior? was the previous implementation incorrect?

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 * [6105](https://github.com/grafana/loki/pull/6105) **rutgerke** Export metrics for the promtail journal target
 * [6179](https://github.com/grafana/loki/pull/6179) **chaudum**: Add new HTTP endpoint to delete ingester ring token file and shutdown process gracefully
 * [6099](https://github.com/grafana/loki/pull/6099/files) **cstyan**: Drop lines with malformed JSON in Promtail JSON pipeline stage
+* [6361](https://github.com/grafana/loki/pull/6361) **chaudum**: Sum values in unwrapped rate aggregation instead of treating them as counter
+* [6099](https://github.com/grafana/loki/pull/6099) **cstyan**: Drop lines with malformed JSON in Promtail JSON pipeline stage
```
chaudum, line 8: Need to be removed

pkg/logql/engine_test.go
```diff
@@ -49,15 +49,36 @@ func TestEngine_LogsRateUnwrap(t *testing.T) {
...
+			[][]logproto.Series{
+				// 30s range the lower bound of the range is not inclusive only 15 samples will make it 60 included
+				{newSeries(testSize, offset(46, constantValue(1)), `{app="foo"}`)},
+			},
+			[]SelectSampleParams{
+				{&logproto.SampleQueryRequest{Start: time.Unix(30, 0), End: time.Unix(60, 0), Selector: `rate({app="foo"} | unwrap foo[30s])`}},
+			},
+			// SUM(n=47, 61, 1) = 15
+			// 15 / 30 = 0.5
+			promql.Vector{promql.Sample{Point: promql.Point{T: 60 * 1000, V: 0.5}, Metric: labels.Labels{labels.Label{Name: "app", Value: "foo"}}}},
```
ssncferreira, line 65: I'm having some difficulty understanding these tests 😕 
What does the `SUM(n=47, 61, 1)` mean?  I assume this is the result of the `newSeries` method on the `data` property, but don't understand how these values were calculated 🤔

KMiller-Grafana (approved):
Docs in this PR look good to me.

pkg/logql/engine_test.go
chaudum: Regarding the `SUM(n=47, 61, 1)`: it just means the sum of the values from item 47 to item 61 where the value is a constant of 1, like https://www.wolframalpha.com/input?i2d=true&i=Sum%5B1%2C%7Bn%2C47%2C61%7D%5D



Regarding, why 47, 61, and 1, we have to look at the output of the `newSeries()` function: It creates a stream `{app="foo"}` with 300 samples starting at timestamp 46e9 (46s) and ending at timestamp 345e9 (345s). The sample value is a constant 1.

So the query now looks at the time range from ts=30s to ts=60s, where the lower bound is not included. There are 15 items (item 47 to item 61) from the generated series that matches.

Hope this helps.

ssncferreira: Thank you! 🙏 This helps a lot. 
Would it be possible to add this information in the comments to make the tests easier to understand? e.g.

```
// create a stream {app="foo"} with 300 samples starting at 46s and ending at 345s with a constant value of 1
[][]logproto.Series{
  {newSeries(testSize, offset(46, constantValue(1)), `{app="foo"}`)},
},
// query between the time range from ts=30s and ts=60s where the lower bound is not included
[]SelectSampleParams{
  {&logproto.SampleQueryRequest{Start: time.Unix(30, 0), End: time.Unix(60, 0), Selector: `rate({app="foo"} | unwrap foo[30s])`}},
},
// SUM(n=47, 61, 1) = 15 - there are 15 samples (from 47 to 61) matched from the generated series
// 15 / 30 = 0.5
promql.Vector{promql.Sample{Point: promql.Point{T: 60 * 1000, V: 0.5}, Metric: labels.Labels{labels.Label{Name: "app", Value: "foo"}}}},
```

owen-d (approved):

pkg/logql/engine_test.go
chaudum: @ssncferreira Addressed your feedback in https://github.com/grafana/loki/pull/6412/files#diff-6f4083532ac476e6ab63b44775eb6356fba146db52e0a19759eff5045b92de2a

PR #6365, opened 2022-06-10
Add compactor route for the cache gen number api

MasslessParticle:
This route was missing so cache invalidation was always failing.


kavirajk (approved):
LGTM

PR #6368, opened 2022-06-10
Don't interrupt the query path if deletes aren't available

MasslessParticle:
To avoid showing deleted data, the querier originally failed when deletes were unavailable. If the compactor becomes unavailable this can cause loki to stop responding to queries altogether. This PR changes the querier to log an error but continue servicing the request when it fails to get deletes.

It also adds metrics to the `delete_requests_client` so we can alert when there is a chronic failure to get delete requests.

### DIFF
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -177,7 +177,7 @@ func (q *SingleTenantQuerier) SelectSamples(ctx context.Context, params logql.Se
 
 	params.SampleQueryRequest.Deletes, err = q.deletesForUser(ctx, params.Start, params.End)
 	if err != nil {
-		return nil, err
+		level.Error(spanlogger.FromContext(ctx)).Log("msg", "failed loading deletes for user", "err", err)
 	}
 
 	ingesterQueryInterval, storeQueryInterval := q.buildQueryIntervals(params.Start, params.End)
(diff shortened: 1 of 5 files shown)


pkg/querier/querier.go
```diff
@@ -177,7 +177,7 @@ func (q *SingleTenantQuerier) SelectSamples(ctx context.Context, params logql.Se
 
 	params.SampleQueryRequest.Deletes, err = q.deletesForUser(ctx, params.Start, params.End)
 	if err != nil {
-		return nil, err
+		level.Error(spanlogger.FromContext(ctx)).Log("msg", "failed loading deletes for user", "err", err)
```
kavirajk, line 180: nit: since you saying `for user`, nice to include `tenat-id` as well in the error log?

kavirajk (approved):
LGTM. one minor nit.

chaudum (approved):

pkg/querier/querier.go
MasslessParticle: That should be covered by the spanlogger: https://github.com/grafana/dskit/blob/ebb5c6de233d26951aef41eada3cab3295ecb368/spanlogger/spanlogger.go#L99-L111

PR #6371, opened 2022-06-11
feat(promtail): add more complex options of dropping logs at the drop stage

witalisoft:

With the promtail I've encountered limitations with dropping logs ie. setting complex expressions using multiple extracted data fields. Flexible approach for dropping is presented in the Prometheus section [relabel_config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) where you write regular expressions against concatenated labels.

I also noticed that currently, promtail uses options `value`(exact match) and `expression` (regexp match) with configuring the drop stage it can be later limited to just `expression` without causing confused what should be used and simplifying code.

n/a

**Checklist**
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -296,7 +296,7 @@ Here is the list with the changes that were produced since the previous release.
 ##### Fixes
 * [6034](https://github.com/grafana/loki/pull/6034) **DylanGuedes**: Promtail: Fix symlink tailing behavior.
 ##### Changes
-* [6371](https://github.com/grafana/loki/pull/6371) **witalisoft**: Drop stage with a complex match based on multiple extracted data fields, breaking config changes regarding to drop stage section
+* [6371](https://github.com/grafana/loki/pull/6371) **witalisoft**: BREAKING: Support more complex match based on multiple extracted data fields in drop stage
 * [5686](https://github.com/grafana/loki/pull/5686) **ssncferreira**: Move promtail StreamLagLabels config to upper level config.Config
 * [5839](https://github.com/grafana/loki/pull/5839) **marctc**: Add ActiveTargets method to promtail
 * [5661](https://github.com/grafana/loki/pull/5661) **masslessparticle**: Invalidate caches on deletes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -296,6 +296,7 @@ Here is the list with the changes that were produced since the previous release.
 ##### Fixes
 * [6034](https://github.com/grafana/loki/pull/6034) **DylanGuedes**: Promtail: Fix symlink tailing behavior.
 ##### Changes
+* [6371](https://github.com/grafana/loki/pull/6371) **witalisoft**: Drop stage with a complex match based on multiple extracted data fields, breaking config changes regarding to drop stage section
 * [5686](https://github.com/grafana/loki/pull/5686) **ssncferreira**: Move promtail StreamLagLabels config to upper level config.Config
 * [5839](https://github.com/grafana/loki/pull/5839) **marctc**: Add ActiveTargets method to promtail
 * [5661](https://github.com/grafana/loki/pull/5661) **masslessparticle**: Invalidate caches on deletes
--- a/clients/pkg/logentry/stages/drop.go
+++ b/clients/pkg/logentry/stages/drop.go
@@ -17,12 +17,13 @@ import (
 )
 
 const (
-	ErrDropStageEmptyConfig     = "drop stage config must contain at least one of `source`, `expression`, `older_than` or `longer_than`"
-	ErrDropStageInvalidDuration = "drop stage invalid duration, %v cannot be converted to a duration: %v"
-	ErrDropStageInvalidConfig   = "drop stage config error, `value` and `expression` cannot both be defined at the same time."
-	ErrDropStageInvalidRegex    = "drop stage regex compilation error: %v"
-	ErrDropStageInvalidByteSize = "drop stage failed to parse longer_than to bytes: %v"
-	ErrDropStageInvalidSource   = "drop stage source invalid type should be string or list of strings"
+	ErrDropStageEmptyConfig       = "drop stage config must contain at least one of `source`, `expression`, `older_than` or `longer_than`"
+	ErrDropStageInvalidDuration   = "drop stage invalid duration, %v cannot be converted to a duration: %v"
+	ErrDropStageInvalidConfig     = "drop stage config error, `value` and `expression` cannot both be defined at the same time."
+	ErrDropStageInvalidRegex      = "drop stage regex compilation error: %v"
+	ErrDropStageInvalidByteSize   = "drop stage failed to parse longer_than to bytes: %v"
+	ErrDropStageInvalidSource     = "drop stage source invalid type should be string or list of strings"
+	ErrDropStageNoSourceWithValue = "drop stage config must contain `source` if `value` is specified"
 )
 
 var (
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -92,6 +92,7 @@
 ##### Fixes
 * [6034](https://github.com/grafana/loki/pull/6034) **DylanGuedes**: Promtail: Fix symlink tailing behavior.
 ##### Changes
+* [6371](https://github.com/grafana/loki/pull/6371) **witalisoft**: Drop stage with a complex match based on multiple extracted data fields, breaking config changes regarding to drop stage section
```
chaudum, line 95: I think it's enough to indicate the breaking change and omit details about it in the changelog, because the upgrade docs contain that information:

```suggestion
* [6371](https://github.com/grafana/loki/pull/6371) **witalisoft**: BREAKING: Support more complex match based on multiple extracted data fields in drop stage
```

docs/sources/clients/promtail/stages/drop.md
```diff
@@ -16,18 +16,17 @@ There are examples below to help explain.
...
+  [sources: [<string>]]
+
+  # Separator placed between concatenated extracted data names. Default separator `;`.
+  [separator: <string>]
   
-  # RE2 regular expression, if source is provided the regex will attempt to match the source
-  # If no source is provided, then the regex attempts to match the log line
-  # If the provided regex matches the log line or a provided source, the line will be dropped. 
+  # RE2 regular expression, if sources are provided the regex will attempt to match the concatenated sources
+  # If no sources is provided, then the regex attempts to match the log line
```
chaudum, line 26:
```suggestion
  # RE2 regular expression, if sources are provided the regex will attempt to match the concatenated sources.
  # If no sources is provided, then the regex attempts to match the log line.
```

docs/sources/upgrading/_index.md
```diff
@@ -38,6 +38,13 @@ The output is incredibly verbose as it shows the entire internal config struct u
 
 This value now defaults to `loki`, it was previously set to `cortex`. If you are relying on this container name for your chunks or ruler storage, you will have to manually specify `-azure.container-name=cortex` or `-ruler.storage.azure.container-name=cortex` respectively.
 
+### Promtail
+
+#### Drop stages config changed
+
+Config option `source` was renamed to `sources` and now acts as a list of strings. Option `value` was removed and use
```
chaudum, line 45:
```suggestion
The configuration option `source` was renamed to `sources` and now accepts a list of strings. 
```

docs/sources/upgrading/_index.md
```diff
@@ -38,6 +38,13 @@ The output is incredibly verbose as it shows the entire internal config struct u
 
 This value now defaults to `loki`, it was previously set to `cortex`. If you are relying on this container name for your chunks or ruler storage, you will have to manually specify `-azure.container-name=cortex` or `-ruler.storage.azure.container-name=cortex` respectively.
 
+### Promtail
+
+#### Drop stages config changed
+
+Config option `source` was renamed to `sources` and now acts as a list of strings. Option `value` was removed and use 
+option `expression` as regexp instead of.
```
chaudum, line 46:
```suggestion
Also, the configuration option `value` was removed in favor of the option `expression`, which takes a regular expression. 
```

chaudum (commented):
Generally lgtm, only left a few minor suggestions. Also, docs should be approved be @KMiller-Grafana

cstyan:
Alternatively we could also support a relabel config per client (like prometheus does for remote write).

witalisoft:
@KMiller-Grafana @chaudum ping

KMiller-Grafana (changes requested):

KMiller-Grafana (approved):
Thanks for making the changes.  Docs LGTM.

witalisoft:
@chaudum ping

witalisoft:
> Generally lgtm, only left a few minor suggestions. Also, docs should be approved be @KMiller-Grafana
> 
> @slim-bean Is it ok to have such a breaking change in a minor release? I assume it is for Promtail (not for Loki obviously).

@slim-bean 
it's a little bit stale, what are the plans for this change ?

jeschkies (changes requested):
I'm wary of making a baking change here. What do you think admit adding `sources` and validate that only one of `source` and `sources` is set? Alternatively `source` could take a strong or a list. @witalisoft

jeschkies:
> Ok, do you mean having backward compatibility by leaving source field?

Yes.

However, I prefer source to take a string or a list. I think the way is to change the JSON field name back to `source` and have it accept both types. Internally it could always be an array.

clients/pkg/logentry/stages/drop.go
```diff
@@ -21,17 +22,20 @@ const (
...
 var (
 	defaultDropReason = "drop_stage"
+	defaultSeparator  = ";"
 )
 
 // DropConfig contains the configuration for a dropStage
 type DropConfig struct {
-	DropReason *string `mapstructure:"drop_counter_reason"`
-	Source     *string `mapstructure:"source"`
-	Value      *string `mapstructure:"value"`
```
jeschkies, line 34: Removing this is a breaking change. We could keep it and use it for the regexp match if set. If value and regexp are defined the config validation should fail.

jeschkies (changes requested):
Thank you very much. This looks really good. We just need to make sure that `value` is a valid option.

clients/pkg/logentry/stages/drop.go
witalisoft: It's much more complicated to maintain the code handling all the exceptions and expressing them in the documentation for the end-user. Moreover, you have to have a plan to later remove them from code and docs. So I would fail the validation if an obsolete `value` will appear in the configuration.

jeschkies: > So I would fail the validation if an obsolete `value` will appear in the configuration.

If it's still possible to define the `value` it's ok. I've just looked at the [documentation](https://grafana.com/docs/loki/latest/clients/promtail/stages/drop/#drop-stage-schema). I thought `expression` was new but it wasn't. So why not keep `value`? It should still work for concatenated sources.

witalisoft: I was writing about this in a first comment:
> I also noticed that currently, promtail uses options value(exact match) and expression (regexp match) with configuring the drop stage it can be limited to just expression without causing confused what should be used and simplifying code

jeschkies: Sorry for missing that. It seems we missed a transition here. However, we cannot break the configuration until a major version. Unfortunately we need to keep `value` to keep the Promtail configs out there valid. I'm wondering if the `expression` could be set to `mt.Sprintf("^%$", Value)` if `value` is set but `expression` is not. This would have some performance implications, though. What do you think?

witalisoft: Sure I can revert the `value` in the near future. 

> I'm wondering if the expression could be set to mt.Sprintf("^%$", Value) if value is set but expression is not

It should be also combined with [regexp.QuoteMeta](https://pkg.go.dev/regexp?utm_source=godoc#QuoteMeta).

jeschkies: @witalisoft would you like me to take over?

witalisoft: no, I will prepare the changes

witalisoft: @jeschkies ready for the review

jeschkies (approved):
Thanks for your patience 🎉

PR #6372, opened 2022-06-12
JSON label extraction field validation expression overly strict

splitice:

I'm working on replicating a bug. Testing my theory with the CI using this PR.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


splitice:
I'm not certain why the tests I added don't fail. None the less they extend the test suite.

I've found what I beleive to be the issue and have gone about fixing it.

splitice:
Confirmed fix of discovered issue with https://github.com/grafana/loki/pull/6372/commits/6cdbaec3b35fbda6631de8ade8bdcc8d90ea4809

pkg/logql/log/jsonexpr/lexer.go
```diff
@@ -83,21 +83,20 @@ func (sc *Scanner) lex(lval *JSONExprSymType) int {
 	}
 }
 
-func isIdentifier(r rune) bool {
+func isStartIdentifier(r rune) bool {
 	return (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || r == '_'
 }
 
+func isIdentifier(r rune) bool {
+	return (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r == '_'
```
chaudum, line 91: What do you think about re-using the `isStartIndentifier()` function here?

```suggestion
	return isStartIdentifier(r) || (r >= '0' && r <= '9')
```

chaudum (commented):
Thanks for the contribution. Looks good to me! I think this change deserves a changelog entry :)

pkg/logql/log/jsonexpr/lexer.go
splitice: I have no objections.

chaudum:
> I'm not certain why the tests I added don't fail (without the fix). None the less they extend the test suite.

We should investigate this before we merge :)

splitice:
@dannykopping I figured it was something like that. Unfortunately had production to fix and it was easier to just test the known failing query against a build.

I'll integrate feedback and that test case.

splitice:
@dannykopping @chaudum review feedback resosolved merge conflicts

dannykopping (approved):
LGTM

PR #6375, opened 2022-06-13
Naïve fix for optimize bug with `line_format` and `json` expression parser

### LINKED ISSUE #6314: Using json expressions with `line_format` in a metric query leads to unexpected results
**Describe the bug**
Consider this query:

```logql
count_over_time({job="random"} | logfmt | line_format "{{.payload}}" | json features="os"[1m])
```

The `line_format` stage is skipped entirely in this metric query because of an [optimisation](https://github.com/grafana/loki/blob/main/pkg/logql/optimize.go#L30). We can generally ignore the `line_format` stage in a metric query because it has no bearing on the result, **unless** there are subsequent [label or line filter expressions](https://github.com/grafana/loki/blob/main/pkg/logql/optimize.go#L56-L60).

The bug: the JSON expression parser is not recognised as type `*syntax.LabelParserExpr` and therefore the `line_format` stage is skipped.

Thanks to @marctc for the bug report on Slack!

dannykopping:

This PR implements a naïve fix for the bug listed below.

Fixes #6314

The more correct way will be to unify the `JSONExpressionParser` and `LabelParserExpr`.

```golang
func (e *LabelParserExpr) Stage() (log.Stage, error) {
	switch e.Op {
	case OpParserTypeJSON:
		return log.NewJSONParser(), nil
	case OpParserTypeLogfmt:
		return log.NewLogfmtParser(), nil
	case OpParserTypeRegexp:
		return log.NewRegexpParser(e.Param)
	case OpParserTypeUnpack:
		return log.NewUnpackParser(), nil
	case OpParserTypePattern:
		return log.NewPatternParser(e.Param)
	default:
		return nil, fmt.Errorf("unknown parser operator: %s", e.Op)
	}
}
```

Ultimately the `JSONExpressionParser` should be instantiated in here, since this is how all other parsers work. I must've missed this somehow when initially implementing the `JSONExpressionParser`.

It'll take a bit of work because in `pkg/logql/syntax/expr.y`

```yacc
labelParser:
    JSON           { $$ = newLabelParserExpr(OpParserTypeJSON, "") }
  | LOGFMT         { $$ = newLabelParserExpr(OpParserTypeLogfmt, "") }
  | REGEXP STRING  { $$ = newLabelParserExpr(OpParserTypeRegexp, $2) }
  | UNPACK         { $$ = newLabelParserExpr(OpParserTypeUnpack, "") }
  | PATTERN STRING { $$ = newLabelParserExpr(OpParserTypePattern, $2) }
  ;
```

all of the `LabelParserExpr`s are expecting a string argument, while the `JSONExpressionParser` expects a slice of `log.JSONExpression`.

I took a swing at this last week but there's a bit of complexity here which I don't have time right now to address, so in the interim I've implemented this simple fix which will make the problem go away, however hacky it may be.

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,7 +81,7 @@
 * [6358](https://github.com/grafana/loki/pull/6358) **taharah**: Fixes sigv4 authentication for the Ruler's remote write configuration by allowing both a global and per tenant configuration.
 * [6152](https://github.com/grafana/loki/pull/6152) **slim-bean**: Fixes unbounded ingester memory growth when live tailing under specific circumstances.
 * [5685](https://github.com/grafana/loki/pull/5685) **chaudum**: Assert that push values tuples consist of string values
-* [6375](https://github.com/grafana/loki/pull/6375) **dannykopping**: Naïve fix for optimize bug with line_format and json expression parser
+* [6375](https://github.com/grafana/loki/pull/6375) **dannykopping**: Fix bug that prevented users from using the `json` parser after a `line_format` pipeline stage.
 ##### Changes
 * [6042](https://github.com/grafana/loki/pull/6042) **slim-bean**: Add a new configuration to allow fudging of ingested timestamps to guarantee sort order of duplicate timestamps at query time.
 * [6120](https://github.com/grafana/loki/pull/6120) **KMiller-Grafana**: Rename configuration parameter fudge_duplicate_timestamp to be increment_duplicate_timestamp.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -81,6 +81,7 @@
 * [6358](https://github.com/grafana/loki/pull/6358) **taharah**: Fixes sigv4 authentication for the Ruler's remote write configuration by allowing both a global and per tenant configuration.
 * [6152](https://github.com/grafana/loki/pull/6152) **slim-bean**: Fixes unbounded ingester memory growth when live tailing under specific circumstances.
 * [5685](https://github.com/grafana/loki/pull/5685) **chaudum**: Assert that push values tuples consist of string values
+* [6375](https://github.com/grafana/loki/pull/6375) **dannykopping**: Naïve fix for optimize bug with line_format and json expression parser
```
chaudum, line 84: I think from a user's point of view this entry is very cryptic. Wdyt about something like this:

```suggestion
* [6375](https://github.com/grafana/loki/pull/6375) **dannykopping**: Fix bug that prevented users from using the `json` parser after a `line_format` pipeline stage.
```

chaudum (commented):
Think it's ok to do the quick fix now. Have a suggestion for the changelog entry, though.

dannykopping:
> Think it's ok to do the quick fix now. Have a suggestion for the changelog entry, though.

That's indeed more clear, thanks @chaudum :+1: 
Can you approve pls?

chaudum (approved):
:ship:

PR #6376, opened 2022-06-13
Deletion: use correct counter when deleting lines

MichelHollands:

An incorrect counter was used when deleting lines. A test to check the counter is added as well.

**Checklist**
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/storage/stores/shipper/compactor/deletion/delete_requests_manager.go
+++ b/pkg/storage/stores/shipper/compactor/deletion/delete_requests_manager.go
@@ -119,7 +119,7 @@ func (d *DeleteRequestsManager) loadDeleteRequestsToProcess() error {
 		if deleteRequest.CreatedAt.Add(d.deleteRequestCancelPeriod).Add(time.Minute).After(model.Now()) {
 			continue
 		}
-		deleteRequest.deletedLinesTotal = d.metrics.deleteRequestsProcessedTotal.WithLabelValues(deleteRequest.UserID)
+		deleteRequest.deletedLinesTotal = d.metrics.deletedLinesTotal.WithLabelValues(deleteRequest.UserID)
 		d.deleteRequestsToProcess = append(d.deleteRequestsToProcess, deleteRequest)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/storage/stores/shipper/compactor/deletion/delete_requests_manager.go
```diff
@@ -119,7 +119,7 @@ func (d *DeleteRequestsManager) loadDeleteRequestsToProcess() error {
 		if deleteRequest.CreatedAt.Add(d.deleteRequestCancelPeriod).Add(time.Minute).After(model.Now()) {
 			continue
 		}
-		deleteRequest.deletedLinesTotal = d.metrics.deleteRequestsProcessedTotal.WithLabelValues(deleteRequest.UserID)
+		deleteRequest.deletedLinesTotal = d.metrics.deletedLinesTotal.WithLabelValues(deleteRequest.UserID)
```
sandeepsukhani, line 122: The function here is just loading delete requests and is not doing any processing of delete requests. It doesn't look like the right place to track the number of lines a delete request has deleted. I think it would be quite hard to count the number of lines deleted if we are deleting whole chunks.

MichelHollands: The actual counting is done in FilterFunction on the DeleteRequest. This code just stores a reference to the counter on the delete request. It was done this way as there's no easy way to get back to the DeleteRequestsManager (which has the metrics) from the DeleteRequest.

Yes, this metric only works when using `filter-and-delete`. I'll have look in a separate PR to see if it can be incremented when deleting whole chunks. Not sure if we have the number of lines in a chunk.

sandeepsukhani: > The actual counting is done in FilterFunction on the DeleteRequest. This code just stores a reference to the counter on the delete request.
sorry, my bad, I read it wrong

sandeepsukhani (approved):
LGTM

MasslessParticle (approved):

trevorwhitney (approved):
lgtm

PR #6385, opened 2022-06-13
Loki SSD: Add index gateway to the read target

trevorwhitney:

This add the index gateway to the read target, forcing ring mode when run in this way. This allows us to run the index gateway as part of SSD.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -896,6 +896,12 @@ func (t *Loki) initIndexGateway() (services.Service, error) {
 }
 
 func (t *Loki) initIndexGatewayRing() (_ services.Service, err error) {
+	// IndexGateway runs by default on read target, and should always assume
+	// ring mode when run in this way.
+	if t.isModuleActive(Read) {
+		t.Cfg.IndexGateway.Mode = indexgateway.RingMode
+	}
+
 	if t.Cfg.IndexGateway.Mode != indexgateway.RingMode {
 		return
 	}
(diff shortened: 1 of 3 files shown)


pkg/loki/modules.go
```diff
@@ -896,6 +896,12 @@ func (t *Loki) initIndexGateway() (services.Service, error) {
 }
 
 func (t *Loki) initIndexGatewayRing() (_ services.Service, err error) {
+	// IndexGateway runs by default on read target, and should always assume
```
DylanGuedes, line 899: I didn't get this part: why it should always run in ring mode?

trevorwhitney: if you don't run in ring mode, then there's no difference between running the index gateway or not. when there's no index gateway the queriers fetch and cache indexes locally, and since they're not clustered you end up duplicating the indexes on every querier. if you run the index gateway on every read mode without sharding, you end up in the same place with every index being duplicated across every read node. therefore, there doesn't seem to be any good reason to enable the index gateway on every read node unless we also run them in ring mode.

trevorwhitney: That's a good point. What are the problems with the default ring configs, what configs could we check to make sure they're setup correctly?

DylanGuedes: Ex:
* Since the default indexgateway mode is simple, let's imagine I don't configure anything for it
* But let's suppose I define a common instance_addr but no common ring
* And finally, I configure all my other rings to run etcd.

In that scenario, the indexgatewayring will actually run and try to use consul which isn't configured.

Issue #1119, opened 2019-10-04
Configure encoding of scraped logs


aledegano:
As Windows is... special... in the Server2019 edition (at least) the log files are saved with the `UTF-16` encoding, this means that trying to scrape them as-is will result in a mangled output in Grafana.

If it was possible to specify through configuration the encoding of the scraped logs (and then parsing the files accordingly) this problem above could be simply solved.

For the time being I'm re-saving the files with `UTF-8` encoding so that they can be displayed correctly.

jgehrcke:
I assume that the underlying premise here is that Loki as a storage system is entirely bytes-oriented (not text-oriented) at all its interfaces.

That in turn means that these interfaces makes "most sense" (in practice, to humans) when the payload data that loki stores -- byte sequences -- after all represent text stored via the UTF-8 encoding.

Given the byte-oriented interfaces, including the push interface, it's the responsibility of the operators to make sure that the abstract text information (which logs after all are) enters the storage system encoded in a consistent, meaningful, predictable way (using a single, useful text encoding -- say, utf-8). This can only be enforced by convention (not by e.g. the Loki API).

The pushing entities (such as Promtail) can of course help with enforcing that convention, by for example making sure that they only push byte sequences that are valid in the target encoding (say, utf-8).

In addition, when different log sources emit their log data in different text encodings (to the file system) then a pushing entity (Promtail) can help normalize that to a single codec.

The configuration interface in Promtail should then really expose the fact that there is both, an input codec per log file (with which Promtail will decode the byte sequence in the file) as well as an output codec (with which Promtail will encode the log text into a byte sequence before pushing the data to Loki).

snv:
I too am having a problem with Log-Files, which are not UTF8 (and can not that easily be changed).
The above pull-request addresses this issue, but it seems to have been closed, because it has been ignored :(

yosiasz:
https://github.com/grafana/loki/issues/6235

PR #2604, opened 2020-09-08
Promtail: Add text encoding conversion to file targets

vyzigold:
**What this PR does / why we need it**:
This PR adds support for text encoding conversion for file targets. To use this, just add `encoding: <encoding_name>` into the scrapeconfig like this:
```
scrape_configs:
- job_name: system_convert
  encoding: utf-16
  static_configs:
  - targets:
      - localhost
    labels:
      job: varlogs
      __path__: /var/utf16.log
```
The encoding name should be one of the names described by IANA [here](https://www.iana.org/assignments/character-sets/character-sets.xhtml).

**Which issue(s) this PR fixes**:
Fixes #1119 

**Checklist**
- [x] Documentation added
- [x] Tests updated

PR #6395, opened 2022-06-14
Promtail: Add text encoding conversion to file targets

DylanGuedes:
Signed-off-by: Jaromír Wysoglad <jwysogla@redhat.com>

This PR adds support for text encoding conversion for file targets. To use this, just add encoding: <encoding_name> into the scrapeconfig like this:
```yaml
scrape_configs:
- job_name: system_convert
  encoding: utf-16
  static_configs:
  - targets:
      - localhost
    labels:
      job: varlogs
      __path__: /var/utf16.log
```
The encoding name should be one of the names described by IANA [here](https://www.iana.org/assignments/character-sets/character-sets.xhtml).

Fixes https://github.com/grafana/loki/issues/1119

This is a revival of [PR 2604](https://github.com/grafana/loki/pull/2604). The original author [couldn't work on it and suggested we continue his work](https://github.com/grafana/loki/pull/2604#issuecomment-1072468479).

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


clients/pkg/promtail/targets/file/tailer.go
```diff
@@ -217,6 +236,24 @@ func (t *tailer) isRunning() bool {
 	return t.running.Load()
 }
 
+func (t *tailer) convertToUTF8(text string) (string, error) {
+	level.Debug(t.logger).Log("msg", "Converting log encoding", "from", t.encoding, "to", "UTF8")
+
+	encoding, err := ianaindex.IANA.Encoding(t.encoding)
```
cyriltovena, line 242: You should create this only once ?

clients/pkg/promtail/targets/file/tailer.go
```diff
@@ -217,6 +236,24 @@ func (t *tailer) isRunning() bool {
...
+	if err != nil {
+		return "", errors.Wrap(err, "error doing IANA encoding")
+	}
+
+	utf8Reader := transform.NewReader(strings.NewReader(text), encoding.NewDecoder())
```
cyriltovena, line 247: Should you use https://pkg.go.dev/golang.org/x/text/transform#String instead ? Seem more fitted for your usage.

DylanGuedes: good call, thanks

clients/pkg/promtail/targets/file/tailer.go
DylanGuedes: done. I tested with 2 tailers and checked the source code and looks like we won't have concurrency problems reusing the decoder or the IANAEncoding struct, but do you mind checking it too?

clients/pkg/promtail/targets/file/tailer.go
```diff
@@ -149,15 +165,26 @@ func (t *tailer) readLines() {
 			continue
 		}
 
+		var text string
+		if t.decoder != nil {
+			var err error
+			text, err = t.convertToUTF8(line.Text)
```
slim-bean, line 171: what's the value for `text` when there is an error?

clients/pkg/promtail/targets/file/tailer.go
```diff
@@ -149,15 +165,26 @@ func (t *tailer) readLines() {
...
+			if err != nil {
+				level.Error(t.logger).Log("msg", "failed to convert encoding", "error", err)
+				t.metrics.encodingFailures.WithLabelValues(t.path).Inc()
```
slim-bean, line 174: I'm wondering how we should handle this error case.

Thoughts: 
* we could write the error into the log line and send that to Loki?
* we could send the log line as received on to Loki even though it didn't format?

I also wonder if we should log at Error level or Debug level, if this happens a lot it in a way someone can't control but they still want all the lines they can get then this could be very noisy, on the other hand if we only increment the metric people may not notice that and not realize they are missing logs.

DylanGuedes: > we could write the error into the log line and send that to Loki?
I like that tbh. I'll modify this to send it.

> we could send the log line as received on to Loki even though it didn't format?
Loki won't be able to parse it so it won't be that useful. I feel like the best experience for users regarding encoding is to be really verbose, so that they can easily know that something isn't properly configured encoding-wise, but once it isn't verbose anymore they can be sure things are working as expected. Another thing is that this encoding won't do anything by default, so we're only bringing noise to future users that are willing to use this.

> I also wonder if we should log at Error level or Debug level, if this happens a lot it in a way someone can't control but they still want all the lines they can get then this could be very noisy, on the other hand if we only increment the metric people may not notice that and not realize they are missing logs.
Yep it has this trade-off. Personally, I prefer it with error level because this looks important enough to be listed as an error but it is debatable.

Nonetheless, I feel like we should first focus on users that:
- Have their data very consistent towards one encoding format
- Really require us to support encoding
This way, edge cases like users having a mix of log lines encoding etc can be addressed in the future whenever we find there's enough demand for it.

clients/pkg/promtail/targets/file/tailer.go
```diff
@@ -149,15 +166,27 @@ func (t *tailer) readLines() {
 			continue
 		}
 
+		var text string
+		if t.decoder != nil {
+			var err error
+			text, err = t.convertToUTF8(line.Text)
+			if err != nil {
+				level.Error(t.logger).Log("msg", "failed to convert encoding", "error", err)
```
slim-bean, line 174: Since we now send the error message in the log line itself, and we have a metric when this happens, I think this should be debug level, it feels mostly just likely to be noise to me at error level.

PR #6396, opened 2022-06-14
[TSDB] Use index sampling to determine shard factor

owen-d:
This adds some rudimentary request based sharding logic which will choose shard factor based on the amount of bytes in the query and the tenant's max parallelism.

### DIFF
--- a/pkg/querier/queryrange/downstreamer.go
+++ b/pkg/querier/queryrange/downstreamer.go
@@ -61,8 +61,8 @@ func ParamsToLokiRequest(params logql.Params, shards logql.Shards) queryrangebas
 func (h DownstreamHandler) Downstreamer(ctx context.Context) logql.Downstreamer {
 	p := DefaultDownstreamConcurrency
 
-	// In the future when we may increase parallelism,
-	// ensure we don't end up bottlenecking here
+	// We may increase parallelism above the default,
+	// ensure we don't end up bottlenecking here.
 	if user, err := tenant.TenantID(ctx); err == nil {
 		if x := h.limits.MaxQueryParallelism(user); x > 0 {
 			p = x
--- a/pkg/querier/queryrange/downstreamer.go
+++ b/pkg/querier/queryrange/downstreamer.go
@@ -9,6 +9,7 @@ import (
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/promql/parser"
 
+	"github.com/grafana/dskit/tenant"
 	"github.com/grafana/loki/pkg/loghttp"
 	"github.com/grafana/loki/pkg/logql"
 	"github.com/grafana/loki/pkg/logqlmodel"
@@ -21,7 +22,8 @@ const (
 )
 
 type DownstreamHandler struct {
-	next queryrangebase.Handler
+	limits Limits
+	next   queryrangebase.Handler
 }
 
 func ParamsToLokiRequest(params logql.Params, shards logql.Shards) queryrangebase.Request {
@@ -54,8 +56,19 @@ func ParamsToLokiRequest(params logql.Params, shards logql.Shards) queryrangebas
 // from creating an unreasonably large number of goroutines, such as
 // the case of a query like `a / a / a / a / a ..etc`, which could try
 // to shard each leg, quickly dispatching an unreasonable number of goroutines.
-func (h DownstreamHandler) Downstreamer() logql.Downstreamer {
+// In the future, it's probably better to replace this with a channel based API
+// so we don't have to do all this ugly edge case handling/accounting
+func (h DownstreamHandler) Downstreamer(ctx context.Context) logql.Downstreamer {
 	p := DefaultDownstreamConcurrency
+
+	// In the future when we may increase parallelism,
+	// ensure we don't end up bottlenecking here
+	if user, err := tenant.TenantID(ctx); err == nil {
+		if x := h.limits.MaxQueryParallelism(user); x > 0 {
+			p = x
+		}
+	}
+
 	locks := make(chan struct{}, p)
 	for i := 0; i < p; i++ {
 		locks <- struct{}{}
--- a/pkg/querier/queryrange/downstreamer.go
+++ b/pkg/querier/queryrange/downstreamer.go
@@ -5,11 +5,11 @@ import (
 	"fmt"
 
 	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/tenant"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/promql/parser"
 
-	"github.com/grafana/dskit/tenant"
 	"github.com/grafana/loki/pkg/loghttp"
 	"github.com/grafana/loki/pkg/logql"
 	"github.com/grafana/loki/pkg/logqlmodel"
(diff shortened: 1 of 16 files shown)


cyriltovena (approved):
LGTM

pkg/querier/queryrange/shard_resolver.go
```diff
@@ -0,0 +1,153 @@
...
+	start := time.Now()
+	if err := concurrency.ForEachJob(r.ctx, len(grps), r.maxParallelism, func(ctx context.Context, i int) error {
+		matchers := syntax.MatchersString(grps[i].Matchers)
+		lookback := grps[i].Interval
+		if lookback == 0 {
+			lookback = r.defaultLookback
+		}
+		diff := lookback + grps[i].Offset
+		adjustedFrom := r.from.Add(-diff)
+		adjustedThrough := r.through.Add(-diff)
```
cstyan, line 84: excuse my ignorance when it comes to the AST code and structs, here you're just aligning the queries time range to be nice multiples of the interval right? wouldn't we want to align the time range for the last group to be later than `r.through` to cover through the requested time range?

pkg/querier/queryrange/shard_resolver.go
```diff
@@ -0,0 +1,153 @@
+package queryrange
+
+import (
+	"context"
+	"fmt"
+	math "math"
```
cstyan, line 6: nit: did you mean to include the name prefix here?

pkg/querier/queryrange/shard_resolver.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+const (
+	// Just some observed values to get us started on better query planning.
+	p90BytesPerSecond = 300 << 20 // 300MB/s/core
+	// At max, schedule a query for 10s of execution before
+	// splitting it into more requests. This is a lot of guesswork.
+	maxSeconds          = 10
+	maxSchedulableBytes = maxSeconds * p90BytesPerSecond
+)
```
cstyan, line 140: these #'s are the throughput we're trying to target?

pkg/querier/queryrange/shard_resolver.go
```diff
@@ -0,0 +1,153 @@
...
+	if expectedSeconds <= float64(maxParallelism) {
+		power := math.Ceil(math.Log2(expectedSeconds)) // round up to nearest power of 2
+		// Ideally, parallelize down to 1s queries
+		return int(math.Pow(2, power))
+	}
+
+	n := stats.Bytes / maxSchedulableBytes
+	power := math.Ceil(math.Log2(float64(n)))
+	return int(math.Pow(2, power))
+}
```
cstyan, line 153: and here you're just calculating the sharding factor do get some data on what me might want to configure that value to?

pkg/querier/queryrange/downstreamer.go
```diff
@@ -54,8 +56,19 @@ func ParamsToLokiRequest(params logql.Params, shards logql.Shards) queryrangebas
...
-func (h DownstreamHandler) Downstreamer() logql.Downstreamer {
+// In the future, it's probably better to replace this with a channel based API
+// so we don't have to do all this ugly edge case handling/accounting
+func (h DownstreamHandler) Downstreamer(ctx context.Context) logql.Downstreamer {
 	p := DefaultDownstreamConcurrency
+
+	// We may increase parallelism above the default,
+	// ensure we don't end up bottlenecking here.
+	if user, err := tenant.TenantID(ctx); err == nil {
+		if x := h.limits.MaxQueryParallelism(user); x > 0 {
```
jeschkies, line 67: Where was this enforced before?

pkg/querier/queryrange/shard_resolver.go
```diff
@@ -0,0 +1,153 @@
...
+	ctx     context.Context
+	handler queryrangebase.Handler
+	logger  log.Logger
+
+	from, through   model.Time
+	maxParallelism  int
+	defaultLookback time.Duration
+}
+
+// from, through, max concurrency to run
```
jeschkies, line 59: I'm not sure I understand this comment. What do you mean?

pkg/querier/queryrange/shard_resolver.go
```diff
@@ -0,0 +1,153 @@
...
+func (r *dynamicShardResolver) Shards(e syntax.Expr) (int, error) {
+	sp, _ := spanlogger.NewWithLogger(r.ctx, r.logger, "dynamicShardResolver.Shards")
+	defer sp.Finish()
+	// We try to shard subtrees in the AST independently if possible, although
+	// nested binary expressions can make this difficult. In this case,
+	// we query the index stats for all matcher groups then sum the results.
+	grps := syntax.MatcherGroups(e)
```
jeschkies, line 66: I may jump the gun here but what do you think about introducing a query plan that is generated from the AST in the future? This plan could already express if expressions are shardable or not.

pkg/querier/queryrange/shard_resolver.go
```diff
@@ -0,0 +1,153 @@
...
+		lookback := grps[i].Interval
+		if lookback == 0 {
+			lookback = r.defaultLookback
+		}
+		diff := lookback + grps[i].Offset
+		adjustedFrom := r.from.Add(-diff)
+		adjustedThrough := r.through.Add(-diff)
+
+		start := time.Now()
+		resp, err := r.handler.Do(r.ctx, &indexgatewaypb.IndexStatsRequest{
```
jeschkies, line 87: Would it make sense to cache the stats?

jeschkies (commented):
This is pretty exciting! Do we have some data on the improvement for throughput or reduction of shards?

pkg/querier/queryrange/shard_resolver.go
owen-d: Thanks (I did not)

pkg/querier/queryrange/shard_resolver.go
owen-d: Yeah, it's just a rough first pass based on some metrics from clusters we run.

pkg/querier/queryrange/shard_resolver.go
owen-d: Yes, we're trying to guess an optimal shard factor based on the amount of data we see in the index.

pkg/querier/queryrange/shard_resolver.go
owen-d: I'm sure there will be a big query planning rewrite in Loki's future, but let's leave that for a separate task :)

pkg/querier/queryrange/shard_resolver.go
owen-d: Yeah, but I don't think we'll do it in this PR :sweat_smile:

pkg/querier/queryrange/shard_resolver.go
owen-d: Good catch - I shouldn't be subtracting the `defaultLookback` from the `through`.

pkg/querier/queryrange/shard_resolver.go
owen-d: I think I left this in from when I was hacking around :laughing: . Will remove

pkg/querier/queryrange/shard_resolver.go
jeschkies: Of course. I just wanted to verify that this is out there on our road :slightly_smiling_face:

PR #6410, opened 2022-06-16
Compactor: add per tenant compaction delete enabled flag

MichelHollands:

This PR adds a per tenant compaction delete enabled flag called `compactor_deletion_enabled`. A unit test was added for the http middleware that checks access.

This was also manually tested. The following snippet was added in the Loki config file:

``` yaml
limits_config:
  compactor_deletion_enabled: false
```

The runtime config contained the following:

``` yaml
overrides:
  tenant1:
    compactor_deletion_enabled: true
  tenant2:
    compactor_deletion_enabled: false
```

Running curl commands against the API gives the expected results: 

``` sh
% curl  http://localhost:3100/loki/api/v1/delete
no org id
% curl -H "X-Scope-OrgID: tenant1"  http://localhost:3100/loki/api/v1/delete
[]
% curl -H "X-Scope-OrgID: tenant2"  http://localhost:3100/loki/api/v1/delete
access denied
% curl -H "X-Scope-OrgID: tenant3"  http://localhost:3100/loki/api/v1/delete
[]
```

**Checklist**
- [x] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -2373,9 +2373,9 @@ The `limits_config` block configures global and per-tenant limits in Loki.
 # CLI flag: -querier.split-queries-by-interval
 [split_queries_by_interval: <duration> | default = 30m]
 
-# When true, access to the deletion API on the compactor is enabled.
-# CLI flag: -compactor.deletion-enabled
-[compactor_deletion_enabled: <boolean> | default = false]
+# When true, access to the deletion API is enabled.
+# CLI flag: -compactor.allow_deletes
+[allow_deletes: <boolean> | default = false]
 ```
 
 ## sigv4_config
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -2372,6 +2372,10 @@ The `limits_config` block configures global and per-tenant limits in Loki.
 # This also determines how cache keys are chosen when result caching is enabled
 # CLI flag: -querier.split-queries-by-interval
 [split_queries_by_interval: <duration> | default = 30m]
+
+# When true, access to the deletion API on the compactor is enabled.
+# CLI flag: -compactor.deletion-enabled
+[compactor_deletion_enabled: <boolean> | default = false]
 ```
 
 ## sigv4_config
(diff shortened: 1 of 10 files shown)


MasslessParticle:
Would it be cleaner to just make a middleware that we wrap our handlers in? 

If we took that approack, we wouldn't have to implement this same check in every handler.

MichelHollands:
> This looks good! Were we going remove `WholeStreamDelete` in this PR or another one?

Due to the size of this one I was going to do it in a separate PR.

docs/sources/configuration/_index.md
```diff
@@ -2372,6 +2372,10 @@ The `limits_config` block configures global and per-tenant limits in Loki.
 # This also determines how cache keys are chosen when result caching is enabled
 # CLI flag: -querier.split-queries-by-interval
 [split_queries_by_interval: <duration> | default = 30m]
+
+# When true, access to the deletion API on the compactor is enabled.
+# CLI flag: -compactor.deletion-enabled
+[compactor_deletion_enabled: <boolean> | default = false]
```
dannykopping, line 2378: Is this config a bit too specific with implementation details? Could this not be renamed to `allow_deletes`?

pkg/storage/stores/shipper/compactor/deletion/request_handler.go
```diff
@@ -183,3 +206,30 @@ func (dm *DeleteRequestHandler) GetCacheGenerationNumberHandler(w http.ResponseW
...
+		allLimits := dm.limits.AllByUserID()
+		userLimits, ok := allLimits[userID]
+		if ok {
+			if !userLimits.CompactorDeletionEnabled {
+				http.Error(w, "access denied", http.StatusForbidden)
+				return
+			}
+		} else {
+			if !dm.limits.DefaultLimits().CompactorDeletionEnabled {
+				http.Error(w, "access denied", http.StatusForbidden)
```
dannykopping, line 228: Is `access denied` the most accurate description here? That feels more like an authn/z issue, not a "this feature is not available for this tenant" issue

dannykopping (commented):
Added two usability nits, otherwise looks good

pkg/storage/stores/shipper/compactor/deletion/request_handler.go
MichelHollands: Changed to "deletion is not available for this tenant".

dannykopping (approved):
LGTM, thanks @MichelHollands :+1:
The docs changes here are fairly minor @KMiller-Grafana so I'm going to merge, I hope you don't mind?

PR #6411, opened 2022-06-16
operator: Extend schema validation in LokiStack webhook

Red-GV:
This PR moves the validation cases from `schema.go` into the validation webhook. Through this move, the degraded condition and the error path of `schema.BuildSchemaConfig` has been removed.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated


periklis (commented):
Looks good to me

operator/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Main
 
+- [6411](https://github.com/grafana/loki/pull/6411) **Red-GV**: Add validation cases to lokistacks webhook
```
xperimental, line 3: Title needs an update.

operator/api/v1beta1/lokistack_webhook.go
```diff
@@ -23,12 +28,17 @@ var _ webhook.Validator = &LokiStack{}
...
+	return s.validate(&LokiStack{})
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
-func (s *LokiStack) ValidateUpdate(_ runtime.Object) error {
-	return s.validate()
+func (s *LokiStack) ValidateUpdate(old runtime.Object) error {
+	oldStack, ok := old.(*LokiStack)
+	if !ok {
+		t := reflect.TypeOf(old).String()
```
xperimental, line 38: Have we used `reflect` for this purpose before? If not, should we use `fmt.Sprintf("%t", old)` instead? This is just a question of preference, I think I would want to avoid directly referencing `reflect` and have used `Sprintf`/`Errorf` for this purpose in the past. Feel free to ignore :slightly_smiling_face:

operator/api/v1beta1/lokistack_webhook.go
```diff
@@ -25,12 +28,17 @@ var _ webhook.Validator = &LokiStack{}
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (s *LokiStack) ValidateCreate() error {
-	return s.validate()
+	return s.validate(&LokiStack{})
```
periklis, line 31: Why not `s.validate(s)` here?

operator/api/v1beta1/lokistack_webhook.go
Red-GV: We use `reflect` for the same purpose in `mutate.go`. `Errorf` causes a lint error.

operator/api/v1beta1/lokistack_webhook.go
periklis: True, but what type of validations are applied at the end of the day against an empty object?

Red-GV: Point. There are no comparisons to be done during the `create` step, so the empty object is passed. However, passing in the object to compare itself looks more odd than passing in an empty object.

periklis: If there are no comparison-based validation, what is left to validate on the newly-created Lokistack object then?

Red-GV: Hmm, nothing much. Alright, perhaps passing `nil` is the answer here. The `schemas` still need to be validated that there are unique dates and that there is a valid start date.

xperimental (approved):

PR #6413, opened 2022-06-17
[modules] pin version of memberlist

cstyan:
Use a replace directive to ensure we're using the same version of memberlist as mimir/dskit. The memberlist adjacent issues we have seen this week could be a result of using a new version of dskit but without the forked version of memberlist.

cc @JordanRushing @vlad-diachenko @owen-d @slim-bean 


JordanRushing (approved):

slim-bean (approved):
LGTM

PR #6415, opened 2022-06-17
Evenly spread queriers across available nodes

salvacorts:

Currently, queriers run on different nodes. We do this by using the following anti-affinity rule:

```yaml
affinity:
  podAntiAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
    - labelSelector:
        matchLabels:
          name: querier
      topologyKey: kubernetes.io/hostname
```

The original reasoning behind this is to make our systems more resilient to node failures. For example, if we have 5 queriers, all scheduled in the same node, and the node fails, we would not be able to serve queries until the pods get rescheduled and running on another node.

When new queriers are created, they need to be scheduled into different nodes. If there are not enough nodes in the cluster, new nodes need to be provisioned. This has two implications:
- Cost: Each node has an implicit cost on top of the workload we deploy into the node.
- Latency: It takes longer to allocate a new node into the cluster, and then schedule the pod, than to schedule it right away into an already available node.

This PR uses Kubernetes's `TopologySpreadConstraints` to evenly spread queriers across the available nodes, but allows more than one querier to be scheduled into the same node.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/upgrading/_index.md
+++ b/docs/sources/upgrading/_index.md
@@ -36,8 +36,7 @@ The output is incredibly verbose as it shows the entire internal config struct u
 #### Evenly spread queriers across kubernetes nodes
 
 We now evenly spread queriers across the available kubernetes nodes, but allowing more than one querier to be scheduled into the same node.
-If you want to keep running up to one querier per node, you will need to revert the changes for `production/ksonnet/loki/querier.libsonnet`
-made at [6415](https://github.com/grafana/loki/pull/6415).
+If you want to keep running up to one querier per node, set `$._config.querier.use_topology_spread` to false.
 
 #### Implementation of unwrapped `rate` aggregation changed
 
--- a/docs/sources/upgrading/_index.md
+++ b/docs/sources/upgrading/_index.md
@@ -33,6 +33,12 @@ The output is incredibly verbose as it shows the entire internal config struct u
 
 ### Loki
 
+#### Evenly spread queriers across kubernetes nodes
+
+We now evenly spread queriers across the available kubernetes nodes, but allowing more than one querier to be scheduled into the same node.
+If you want to keep running up to one querier per node, you will need to revert the changes for `production/ksonnet/loki/querier.libsonnet`
+made at [6415](https://github.com/grafana/loki/pull/6415).
+
 #### Implementation of unwrapped `rate` aggregation changed
 
 The implementation of the `rate()` aggregation function changed back to the previous implemention prior to [#5013](https://github.com/grafana/loki/pulls/5013).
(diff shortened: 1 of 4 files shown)


salvacorts:
As suggested by @dannykopping in the design doc, I added an entry to the changelog and the upgrade guide.

dannykopping (commented):
This looks good @salvacorts, but I think we should make it configurable.

> If you want to keep running up to one querier per node, you will need to revert the changes for `production/ksonnet/loki/querier.libsonnet`
made at [6415](https://github.com/grafana/loki/pull/6415).

How about we offer both the node affinity & the topology spread via a config value (defaulted to topology spread), and allow configuring the max skew?

This could later be expanded for all components (defaulted to node affinity for now)

salvacorts:
> How about we offer both the node affinity & the topology spread via a config value (defaulted to topology spread), and allow configuring the max skew?

@dannykopping That's a good idea. I just added two new config parameters: `querier.use_topology_spread` and `querier.topology_spread_max_skew`. I also updated the upgrade guide to reflect these changes.

docs/sources/upgrading/_index.md
```diff
@@ -36,8 +36,7 @@ The output is incredibly verbose as it shows the entire internal config struct u
 #### Evenly spread queriers across kubernetes nodes
 
 We now evenly spread queriers across the available kubernetes nodes, but allowing more than one querier to be scheduled into the same node.
-If you want to keep running up to one querier per node, you will need to revert the changes for `production/ksonnet/loki/querier.libsonnet`
-made at [6415](https://github.com/grafana/loki/pull/6415).
+If you want to keep running up to one querier per node, set `$._config.querier.use_topology_spread` to false.
```
dannykopping, line 39:
```suggestion
If you want to run at most a single querier per node, set `$._config.querier.use_topology_spread` to false.
```

production/ksonnet/loki/config.libsonnet
```diff
@@ -47,6 +47,13 @@
       // A higher value will lead to a querier trying to process more requests than there are available
       // cores and will result in scheduling delays.
       concurrency: 4,
+
+      // If use_topology_spread is true, queriers can run on nodes already running queriers but will be
+      // spread through the available nodes using a TopologySpreadConstraints with a max skew
```
dannykopping, line 52: Can we link the docs here pls?

dannykopping (approved):
LGTM! Two minor nits

production/ksonnet/loki/config.libsonnet
salvacorts: Done

docs/sources/upgrading/_index.md
salvacorts: Done

PR #6432, opened 2022-06-20
Fixes bad initialization for uniqueString

cyriltovena:
It was always 0 before and then appending, but in fact we wanted to have a pre-allocated size of `len(s)`.


dannykopping (approved):
LGTM, nice find

PR #6435, opened 2022-06-20
Remove whole stream deletion mode

MichelHollands:

The `whole stream deletion` mode is no longer required as the `filter-and-delete` mode can do the same and more.

**Checklist**
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -574,7 +574,7 @@ func (t *Loki) initQueryFrontendTripperware() (_ services.Service, err error) {
 }
 
 func (t *Loki) cacheGenClient() (generationnumber.CacheGenClient, error) {
-	filteringEnabled, err := deletion.FilteringEnabled(t.Cfg.CompactorConfig.DeletionMode)
+	filteringEnabled, err := deletion.DeleteEnabled(t.Cfg.CompactorConfig.DeletionMode)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


pkg/loki/modules.go
```diff
@@ -881,7 +881,7 @@ func (t *Loki) initCompactor() (services.Service, error) {
 
 	if t.Cfg.CompactorConfig.RetentionEnabled {
```
MasslessParticle, line 882: This could be simplified if we rename the `FilteringEnabled` method to `DeleteEnabled`, remove the switch statement, and changing the condition to `if t.Cfg.CompactorConfig.RetentionEnabled && t.compactor.DeleteE
nabled()`.

pkg/storage/stores/shipper/compactor/compactor.go
```diff
@@ -232,7 +232,7 @@ func (c *Compactor) init(storageConfig storage.Config, schemaConfig config.Schem
 		}
 
 		switch c.deleteMode {
```
MasslessParticle, line 234: This one could also be simplified to just an if statement with `DeleteEnabled`.

pkg/storage/stores/shipper/compactor/deletion/mode_test.go
```diff
@@ -31,3 +27,21 @@ func TestParseMode(t *testing.T) {
 	_, err = ParseMode("something-else")
 	require.ErrorIs(t, errUnknownMode, err)
 }
+
+func TestFilteringEnabled(t *testing.T) {
+	enabled, err := FilteringEnabled("disabled")
```
MasslessParticle, line 32: This is the method we can change because `DeleteEnabled` now implies filtering

MasslessParticle (changes requested):
This looks good. I think things can be simplified by replacing the deletion mode switch statements with if statements.

pkg/storage/stores/shipper/compactor/deletion/mode_test.go
MichelHollands: Changed to use a new method on the mode struct.

MasslessParticle (approved):

MasslessParticle:
LGTM

KMiller-Grafana (commented):
Docs look good to me.

PR #6439, opened 2022-06-20
Consistently name metrics with the same prefix

cstyan:
This was really bothering me while getting things setup for some memberlist migrations. Trying to build a dashboard where various memberlist, ring, kvstore, etc. metrics have different prefixes because of where and how their metrics are registered. I think most of the `cortex_` metrics were a side effect of duplicating code in our effort to remove our cortex dependencies.

This PR should:
1) prefix all metrics we currently have prefixed with `cortex_` to now be prefixed with `loki_`
2) updates dashboards the relied on `cortex_` metrics to now use `loki_`metrics
3) Some other minor changes; we used these metrics as part of some autoscaling calculations in AWS client code, and we had a reference to the ingester flush queue metric in docs referring to it with `cortex_` prefix.

Still need a changelog entry.


production/loki-mixin-compiled/dashboards/loki-reads.json
```diff
@@ -207,7 +207,7 @@
             "repeatIteration": null,
             "repeatRowId": null,
             "showTitle": true,
-            "title": "Frontend (cortex_gw)",
+            "title": "Frontend (loki_gw)",
```
dannykopping, line 210: What's `loki_gw`?

pkg/distributor/distributor_test.go
```diff
@@ -59,22 +59,22 @@ func TestDistributor(t *testing.T) {
...
-			expectedResponse: success,
-			expectedError:    httpgrpc.Errorf(http.StatusBadRequest, validation.LineTooLongErrorMsg, 1, "{foo=\"bar\"}", 10),
-		},
-		{
-			lines:            100,
-			mangleLabels:     true,
-			expectedResponse: success,
-			expectedError:    httpgrpc.Errorf(http.StatusBadRequest, validation.InvalidLabelsErrorMsg, "{ab\"", "1:4: parse error: unterminated quoted string"),
-		},
+		// {
```
dannykopping, line 62: Do we want to keep these commented?

production/loki-mixin/dashboards/dashboard-loki-operational.json
```diff
@@ -713,17 +713,17 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "histogram_quantile(0.99, sum by (le) (job_route:loki_request_duration_seconds_bucket:sum_rate{job=~\"($namespace)/cortex-gw\", route=~\"api_prom_push|loki_api_v1_push\", cluster=~\"$cluster\"})) * 1e3",
+          "expr": "histogram_quantile(0.99, sum by (le) (job_route:loki_request_duration_seconds_bucket:sum_rate{job=~\"($namespace)/loki_-gw\", route=~\"api_prom_push|loki_api_v1_push\", cluster=~\"$cluster\"})) * 1e3",
```
dannykopping, line 716: Replacing `cortex-` here with `loki_` seems wrong

production/loki-mixin/dashboards/dashboard-loki-operational.json
```diff
@@ -5658,17 +5658,17 @@
...
             },
             {
-              "expr": "histogram_quantile(.9, sum(rate(cortex_dynamo_query_pages_count{cluster=\"$cluster\", namespace=\"$namespace\"}[5m])))",
+              "expr": "histogram_quantile(.9, sum(rate(loki__dynamo_query_pages_count{cluster=\"$cluster\", namespace=\"$namespace\"}[5m])))",
               "legendFormat": ".9",
               "refId": "B"
             },
             {
-              "expr": "histogram_quantile(.5, sum(rate(cortex_dynamo_query_pages_count{cluster=\"$cluster\", namespace=\"$namespace\"}[5m])))",
+              "expr": "histogram_quantile(.5, sum(rate(loki__dynamo_query_pages_count{cluster=\"$cluster\", namespace=\"$namespace\"}[5m])))",
```
dannykopping, line 5671: We have `loki__` here which seems wrong

dannykopping (changes requested):
Again, thanks for the massive effort on this! Spotted a couple potential issues which were likely unintended.

I think changing all of these in one fell swoop is going to cause a few issues. I wonder if we could instead pass a registerer that simply publishes the `cortex_` metrics and then replaces the prefix with `loki_` and publishes that too. This would obviously cause the number of metrics to double, but it can be optional. For us, we'd turn this off once we've replaced all the collateral that still uses the `cortex_` prefix, but for our general users the duplication may be preferable to the large effort of renaming metrics.

On top of this, we should announce a deprecation plan for the `cortex_` metrics with the view to (IMHO) deprecating in v3.0 and removing completely in v4.0.

_Aside:_
Can you please avoid force-pushing once a review has begun? If a force-push is done, reviewers can't perform incremental reviews:
https://github.com/grafana/loki/pull/6439/files/de288ccbb82c522c3c463cacc164a10aba690b79..HEAD

image
_Show changes since your last review_ fails because history is rewritten.

PR #6446, opened 2022-06-21
Handle duplicate registration of deletion metrics

MichelHollands:

When running in single binary mode with the ruler enabled a panic occurs because the deletion metrics are registered twice. This fix handles duplicate registration of the metrics in question.

**Checklist**
- [ ] Documentation added
- [X] Tests updated


cstyan (commented):
This is fine but IMO the correct way to solve this is to create the instance of the metrics struct higher up and pass it to `NewDeleteRequestsClient`.

MichelHollands:
> LGTM, my only question is whether there should be a label on the delete metrics for the client that has made the request, which would mean changing the metrics from counters to CounterVec.

I've put that in a separate PR: https://github.com/grafana/loki/pull/6469. Please have a look.

PR #6452, opened 2022-06-21
[promtail/mixin] Remove promtail file lag alert

cstyan:
We're going to remove the associated metric soon anyways and the alert is problematic as well.

### DIFF
--- a/production/promtail-mixin/alerts.libsonnet
+++ b/production/promtail-mixin/alerts.libsonnet
@@ -38,21 +38,6 @@
             },
           },
           {
-            alert: 'PromtailFileLagging',
-            expr: |||
-              promtail_stream_lag_seconds > 10
-            |||,
-            'for': '15m',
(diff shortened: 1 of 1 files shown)


production/promtail-mixin/alerts.libsonnet
```diff
@@ -37,21 +37,6 @@
               |||,
             },
           },
-          {
-            alert: 'PromtailFileLagging',
-            expr: |||
-              promtail_stream_lag_seconds > 10
```
dannykopping, line 43: We had a user find this metric quite useful recently; can you please remind me why we're removing the metric?

dannykopping: OK great :+1: thanks

dannykopping (approved):
LGTM, I'll let you use your newly minted merge privileges here :slightly_smiling_face:

PR #6453, opened 2022-06-21
fix(lambda-promtail): update default lambda_vpc_subnets / lambda_vpc_security_groups values to be appropriate empty list(string) value

### LINKED ISSUE #6451: lambda-promtail: default `lambda_vpc_subnets` and `lambda_vpc_security_groups` set to [""] instead of [] causes constant Terraform resource update-in-place change
These are the default values in `variables.tf` and lead to constant update, presumably because AWS rejects the empty string and sets it to `[]`, so next time Terraform sees that `[]` != `[""]` and identifies that the resource should be updated in-place. A default empty value for `list(string)` should be `[]`

```zsh
# aws_lambda_function.lambda_promtail will be updated in-place
  ~ resource "aws_lambda_function" "lambda_promtail" {
        id                             = "lambda_promtail"
        tags                           = {}
        # (25 unchanged attributes hidden)




      + vpc_config {
          + security_group_ids = [
              + "",
            ]
          + subnet_ids         = [
              + "",
            ]
        }
        # (3 unchanged blocks hidden)
    }
```

conorevans:
Signed-off-by: Conor Evans <coevans@tcd.ie>

Sets appropriate default values

Fixes #6451

### DIFF
--- a/tools/lambda-promtail/variables.tf
+++ b/tools/lambda-promtail/variables.tf
@@ -47,7 +47,7 @@ variable "keep_stream" {
   default     = "false"
 }
 
-variable "extra_labels"{
+variable "extra_labels" {
   type = string
   description = "Comma separated list of extra labels, in the format 'name1,value1,name2,value2,...,nameN,valueN' to add to entries forwarded by lambda-promtail."
   default = ""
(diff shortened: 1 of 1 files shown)


cstyan:
@conorevans Have you tried deploying this? I feel like I used arrays with empty strings for a reason, the `vpc_config` docs state that both fields are required and I don't know terraform well enough to know if empty lists count as fulfilling that required field: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#vpc_config

I can try myself tomorrow but I don't have things easily setup for running lambda-promtail on my end anymore unfortunately.

cstyan:
Sorry for the delay, was out on vacation last week. LGTM

cstyan (approved):

PR #6456, opened 2022-06-22
deb and rpm packaging for all binaries

julienduchesne:
There is currently deb and rpm packages being published for logcli amd64 but they aren't signed

In this PR:
- Add RPM and deb packages for `logcli`, `loki-canary` `loki` and `promtail` for arm, arm64 and amd64
- Sign them with the Grafana GPG key (packages.grafana.com/gpg.key). This will be the same key for all Grafana products
- Add CI steps to test the RPM and deb packages. This launches a docker image with systemd installed, then it installs both loki and promtail and does a `logcli` query on the Loki instance


tools/package-nfpm.jsonnet
```diff
@@ -0,0 +1,41 @@
...
+  maintainer: 'Grafana Labs <support@grafana.com>',
+  description: descriptions[name],
+  vendor: 'Grafana Labs Inc',
+  homepage: 'https://grafana.com/loki',
+  changelog: 'https://github.com/grafana/loki/blob/main/CHANGELOG.md',
+  license: 'AGPL-3.0',
+  contents: [{
+    src: './dist/tmp/packages/%s-linux-%s' % [name, arch],
+    dst: '/usr/local/bin/%s' % name,
+  }],
```
chaudum, line 40: Did I understand it correctly that the package only contain the binary that is then installed to `/usr/local/bin`? This may be useful for LogCLI, but for Promtail and Loki, which need a configuration file, I see more harm than gain. Users would expect that the package would also contain a systemd unit which they can use to start/stop Loki and a config that works out of the box (probably using filesystem storage). Unless the packages do not contain that, I am worried to publish them.

chaudum (changes requested):
Thanks for your efforts to make OS packages happen also for Loki. However I have a big concern which I raised in a comment.

tools/package-nfpm.jsonnet
jeschkies: I agree. Configuration and systemd unit files are a big plus of native packages.

tools/package-nfpm.jsonnet
```diff
@@ -0,0 +1,41 @@
...
+  platform: 'linux',
+  version: '${CIRCLE_TAG}',
+  section: 'default',
+  provides: [name],
+  maintainer: 'Grafana Labs <support@grafana.com>',
+  description: descriptions[name],
+  vendor: 'Grafana Labs Inc',
+  homepage: 'https://grafana.com/loki',
+  changelog: 'https://github.com/grafana/loki/blob/main/CHANGELOG.md',
+  license: 'AGPL-3.0',
```
jeschkies, line 36: Promtail is Apache 2 if I'm not wrong.

jeschkies (commented):
Thanks for getting this started :slightly_smiling_face:

tools/package-nfpm.jsonnet
julienduchesne: 👍 Good point, I'll add those. Adding daemon configs really is the nicest part of OS packaging

KMiller-Grafana (approved):
File contents for tools/packaging/nfpm.jsonnet look good to me.

tools/package-nfpm.jsonnet
julienduchesne: I changed this

julienduchesne:
@jeschkies @chaudum. Re-requesting a review. I added systemd configs for both deb and RPM. I also added a CI test to check that everything keeps working. The CI check uses CentOS and Debian. It installs loki and promtail and then queries with logcli to see that everything works well

.drone/drone.jsonnet
```diff
@@ -537,18 +539,93 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
...
-      ) { when: { event: ['tag'] } },
+      run('write-key',
+          commands=['printf "%s" "$NFPM_SIGNING_KEY" > $NFPM_SIGNING_KEY_FILE'],
+          env={
+            NFPM_SIGNING_KEY: { from_secret: gpg_private_key.name },
+            NFPM_SIGNING_KEY_FILE: '/drone/src/private-key.key',
+          }),
+      run('test packaging',
+          commands=[
+            'go install github.com/google/go-jsonnet/cmd/jsonnet@latest',  // Test, install in build image instead
```
jeschkies, line 591: I'm not sure I understand this step. Why do you install a Go dependency here?

.drone/drone.jsonnet
```diff
@@ -537,18 +539,93 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
...
+            'make BUILD_IN_CONTAINER=false packages',
+          ],
+          env={
+            NFPM_PASSPHRASE: { from_secret: gpg_passphrase.name },
+            NFPM_SIGNING_KEY_FILE: '/drone/src/private-key.key',
+          }),
+      {
+        name: 'test deb package',
+        image: 'docker',
+        commands: ['./tools/packaging/verify-deb-install.sh'],
```
jeschkies, line 601: This is awesome. I'd use some test library but am totally fine with a shell script for now.

tools/packaging/nfpm.sh
```diff
@@ -0,0 +1,24 @@
...
+    exit 1
+fi
+
+rm -rf dist/tmp && mkdir -p dist/tmp/packages
+unzip dist/\*.zip -d dist/tmp/packages
+
+for name in loki loki-canary logcli promtail; do
+    for arch in amd64 arm64 arm; do
+        config_path="dist/tmp/config-${name}-${arch}.json"
+        jsonnet -V "name=${name}" -V "arch=${arch}" "tools/packaging/nfpm.jsonnet" > "${config_path}"
```
jeschkies, line 18: ...ah this is why you need to install Jsonnet :slightly_smiling_face:

tools/packaging/verify-deb-install.sh
```diff
@@ -0,0 +1,34 @@
...
+
+dir="."
+if [ -n "${CI}" ]; then
+    dir="/drone/src"
+fi
+echo "Running on directory: ${dir}"
+
+cat <<EOF | docker exec --interactive "${image}" sh
+    # Install loki and check it's running
+    dpkg -i ${dir}/dist/loki_0.0.0~rc0_amd64.deb
```
jeschkies, line 15: Shouldn't the binary have a version?

tools/packaging/verify-deb-install.sh
```diff
@@ -0,0 +1,34 @@
...
+    echo "blablabla" > /var/log/test.log
+
+    # Install logcli
+    dpkg -i ${dir}/dist/logcli_0.0.0~rc0_amd64.deb
+
+    # Check that there are labels
+    sleep 5
+    labels_found=\$(logcli labels)
+    echo "Found labels: \$labels_found"
+    [ "\$labels_found" != "" ] || (echo "no logs found with logcli" && exit 1)
```
jeschkies, line 33: I'd use a loop but this is a great start. We can always iterate.

jeschkies (approved):
This is really awesome! Just so I understand. The packages are not published yet, right?

julienduchesne:
> This is really awesome! Just so I understand. The packages are not published yet, right?

No, we're working on a global repository (apt.grafana.com and rpm.grafana.com) where we will publish all packages. It'll just be an additional Drone steps. For now, the deb and rpm files will be in the releases. I'm hoping to have the repositories ready before end of July

tools/packaging/verify-deb-install.sh
julienduchesne: oh dang, that'll fail on actual releases. Will fix

.drone/drone.jsonnet
julienduchesne: Installed in the build image here: https://github.com/grafana/loki/pull/6526

rightaway:
@julienduchesne The debian package will install directly on the system not through docker?

Where can I see what the `.service` files will look like on debian for loki and promtail packages?

tools/packaging/nfpm.jsonnet
```diff
@@ -0,0 +1,90 @@
...
+      It is designed to be very cost effective and easy to operate. 
+      It does not index the contents of the logs, but rather a set of labels for each log stream.
+    |||,
+    contents+: [
+      {
+        src: './tools/packaging/loki.service',
+        dst: '/etc/systemd/system/loki.service',
+      },
+      {
+        src: './cmd/loki/loki-local-config.yaml',
```
chaudum, line 26: This config file configures Loki to write to `/tmp/loki`. Not sure if this is the ideal default location, but I think it's good enough for a first iteration.

chaudum (approved):

tools/packaging/promtail.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Promtail service
+After=network.target
+
+[Service]
+Type=simple
+User=promtail
+ExecStart=/usr/bin/promtail -config.file /etc/promtail/config.yml
+
```
MAFLO321, line 9: Wouldn't it be a neat idea if promtail would also be `Restart = on-failure`, just like in `loki.service`?

PR #6459, opened 2022-06-22
Change standard output destination depending on log level

### LINKED ISSUE #5331: Non-error logs are printed to stderr
**Describe the bug**

As in title. This makes alerting configuration based on log severity problematic because everything printed by loki itself seems to be an error. Most logging aggregation tools and platforms treat stderr output as level=error.

For example:

```
2022-02-06 11:42:17 | level=info ts=2022-02-06T10:42:17.760239159Z caller=table.go:362 msg="finished uploading table index_19029
```

^^^ not an error but it's logged to stderr.

**To Reproduce**

1. Start Loki (2.4.2)
2. Wait until it prints its standard log messages (e.g. db, table related info)

**Expected behavior**
Informational logs are printed to stdout, and only errors are printed to stderr.

**Environment:**
 - bare-metal

masanetes:

Since all log levels are output to stderr, logging and its filters are tainted by Loki's logs. To avoid this, we think it is a good idea to change the output destination depending on the level

Fixes #5331

```
./loki -config.file=./cmd/loki/loki-local-config.yaml 2> /dev/null
level=info ts=2022-06-22T09:32:49.577972Z caller=main.go:103 msg="Starting Loki" version="(version=, branch=, revision=)"
level=info ts=2022-06-22T09:32:49.578361Z caller=server.go:260 http=[::]:3100 grpc=[::]:9096 msg="server listening on addresses"
level=info ts=2022-06-22T09:32:49.581057Z caller=table_manager.go:134 msg="uploading tables"
```

over level warning
```
./loki -config.file=./cmd/loki/loki-local-config.yaml 1> /dev/null
level=warn ts=2022-06-22T09:33:21.758191Z caller=experimental.go:20 msg="experimental feature in use" feature="In-memory (FIFO) cache - chunksfifocache"
level=error ts=2022-06-22T09:33:21.762942Z caller=mapper.go:50 msg="unable to read rules directory" path=/tmp/loki/rules-temp err="open /tmp/loki/rules-temp: no such file or directory"
```

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/util/log/log.go
+++ b/pkg/util/log/log.go
@@ -37,20 +37,25 @@ func InitLogger(cfg *server.Config, reg prometheus.Registerer) {
 // prometheusLogger exposes Prometheus counters for each of go-kit's log levels.
 type prometheusLogger struct {
 	logger      log.Logger
+	elogger     log.Logger
 	logMessages *prometheus.CounterVec
 }
 
 // newPrometheusLogger creates a new instance of PrometheusLogger which exposes
 // Prometheus counters for various log levels.
 func newPrometheusLogger(l logging.Level, format logging.Format, reg prometheus.Registerer) log.Logger {
-	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
+	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stdout))
+	elogger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
 	if format.String() == "json" {
-		logger = log.NewJSONLogger(log.NewSyncWriter(os.Stderr))
+		logger = log.NewJSONLogger(log.NewSyncWriter(os.Stdout))
+		elogger = log.NewJSONLogger(log.NewSyncWriter(os.Stderr))
 	}
 	logger = level.NewFilter(logger, levelFilter(l.String()))
+	elogger = level.NewFilter(elogger, levelFilter(l.String()))
 
 	plogger := &prometheusLogger{
-		logger: logger,
+		logger:  logger,
+		elogger: elogger,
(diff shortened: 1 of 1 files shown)


dannykopping (commented):
I appreciate the effort here, and the solution is certainly nice, however I have trouble with this PR for two reasons:

1. the logs can easily be filtered using the `level` field
2. this is a major backwards-incompatibility for all folks who are currently relying on logs being written to `stderr`

It would've been nice if we adopted this strategy from day 0, but here we are and it's tricky to change something so fundamental.

On that basis, I'm inclined to not accept this PR, but I'm interested in your response to the above and the opinions of my fellow maintainers.

MichelHollands:
Yes, I don't think this is good idea either. It will be too easy to inadvertently lose logs if you forget to read either stderr  or stdout. Additionally it's not easy to change. I can imagine a scenario where you want levels Info, Warning and Error to go to stdout and Debug level output to stderr. Better to do this in your logging tool or when you capture these logs.

masanetes:
@dannykopping @MichelHollands 

I appreciate both of your input. I understand that those changes are not easy and can be solved with logging tools and that this modification is not appropriate

muffl0n:
I'm a bit sad that this will not be fixed. Writing errors to stdout seems like a major flaw and should be fixed, maybe in a major release. 
My old CS teacher would have let me fail for something like this 😆

kemics:
@dannykopping Google Cloud counts logs written to stderr as error logs and it makes sense why. This creates false positive alerts in case of Loki and makes it more challenging to monitor. It would be helpful to have this to be configurable

muffl0n:
Having a flag like described in the linked PR sounds awesome as it maintains the mentioned backward compatibility.
Would be awesome to have something similar for loki!

PR #6482, opened 2022-06-23
Add Drone pipeline to test packaging

julienduchesne:
Currently, the packaging process only happens on tags, which could lead to surprises
Especially considering I'm working on deb and RPM packaging (lots more steps than now)

I also moved the publishing from Circle to Drone because:
1. I'll need some new secrets from Vault
2. It should be faster

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -245,9 +245,8 @@ cmd/migrate/migrate:
 #############
 # Releasing #
 #############
-# concurrency is limited to 2 to prevent CircleCI from OOMing. Sorry
-GOX = gox $(GO_FLAGS) -parallel=2 -output="dist/{{.Dir}}-{{.OS}}-{{.Arch}}"
-CGO_GOX = gox $(DYN_GO_FLAGS) -cgo -parallel=2 -output="dist/{{.Dir}}-{{.OS}}-{{.Arch}}"
+GOX = gox $(GO_FLAGS) -parallel=4 -output="dist/{{.Dir}}-{{.OS}}-{{.Arch}}"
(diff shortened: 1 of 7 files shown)


julienduchesne:
Also, additional question. Any reason why all the same pipelines are run for PRs and push. I feel like we could remove the PR event

Makefile
```diff
@@ -245,9 +245,8 @@ cmd/migrate/migrate:
 #############
 # Releasing #
 #############
-# concurrency is limited to 2 to prevent CircleCI from OOMing. Sorry
-GOX = gox $(GO_FLAGS) -parallel=2 -output="dist/{{.Dir}}-{{.OS}}-{{.Arch}}"
-CGO_GOX = gox $(DYN_GO_FLAGS) -cgo -parallel=2 -output="dist/{{.Dir}}-{{.OS}}-{{.Arch}}"
+GOX = gox $(GO_FLAGS) -parallel=4 -output="dist/{{.Dir}}-{{.OS}}-{{.Arch}}"
```
kavirajk, line 248: curious why 4? (I mean can it be higher?)

kavirajk (approved):
thanks @julienduchesne :+1: LGTM. Just few clarifications. No blocker.

Makefile
julienduchesne: I removed it. We don't run the risk of OOMing in Drone, so I'll just let `gox` figure it out

PR #6461, opened 2022-06-22
Loki: Use Shipper as indexClient for read mode

DylanGuedes:
**What this PR does / why we need it**:
This PR forces Loki to not use the IndexGatewayClient when running with target=read. Without this, the read mode ends up using an IndexGatewayClient to resolve index queries; however, as IndexGatewayClient doesn't support resolving these queries, any query ends up hanging forever.

**Which issue(s) this PR fixes**:
N/A

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] Is this an important fix or new feature? Add an entry in the `CHANGELOG.md`.
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #6490, opened 2022-06-24
Disable index gateway client on read target.

jeschkies:
Summary:
PR #6461 introduced a fix for the read target that was supposed to
disable the index gateway clients. However, `fallthrough` does not
behave as expected. Thus we are using a nested if statement here.

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -490,8 +490,13 @@ func (t *Loki) initStore() (_ services.Service, err error) {
 			// Use AsyncStore to query both ingesters local store and chunk store for store queries.
 			// Only queriers should use the AsyncStore, it should never be used in ingesters.
 			asyncStore = true
-			fallthrough
-		case t.Cfg.isModuleEnabled(IndexGateway) || t.Cfg.isModuleEnabled(Read):
+
+			if t.Cfg.isModuleEnabled(Read) {
+				// we want to use the actual storage when running the index-gateway, so we remove the Addr from the config
+				t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Disabled = true
+				t.Cfg.StorageConfig.TSDBShipperConfig.IndexGatewayClientConfig.Disabled = true
+			}
+		case t.Cfg.isModuleEnabled(IndexGateway):
 			// we want to use the actual storage when running the index-gateway, so we remove the Addr from the config
 			t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Disabled = true
 			t.Cfg.StorageConfig.TSDBShipperConfig.IndexGatewayClientConfig.Disabled = true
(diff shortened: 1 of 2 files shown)


dannykopping (approved):
LGTM, thanks

pkg/loki/modules.go
```diff
@@ -490,8 +490,13 @@ func (t *Loki) initStore() (_ services.Service, err error) {
 			// Use AsyncStore to query both ingesters local store and chunk store for store queries.
 			// Only queriers should use the AsyncStore, it should never be used in ingesters.
 			asyncStore = true
-			fallthrough
-		case t.Cfg.isModuleEnabled(IndexGateway) || t.Cfg.isModuleEnabled(Read):
+
+			if t.Cfg.isModuleEnabled(Read) {
+				// we want to use the actual storage when running the index-gateway, so we remove the Addr from the config
+				t.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Disabled = true
```
salvacorts, line 496: nit: maybe we can extract this to a function. If in the future we make modifications on the `t.Cfg.isModuleEnabled(IndexGateway)` case, they should go inside the if statement as well, but we may skip it unintentionally.

pkg/loki/modules_test.go
```diff
@@ -216,7 +216,50 @@ func TestIndexGatewayRingMode_when_TargetIsRead(t *testing.T) {
 		})
 
 	})
+}
+
+func TestIndexGatewayClientConfig_when_TargetIsQuerier(t *testing.T) {
```
salvacorts, line 221: We are testing both the read and the querier target here. So the test name should not be `...TargetIsQuerier`.
Maybe `TestIndexGatewayClientConfig_when_TargetIsQuerierOrRead`?

pkg/loki/modules_test.go
```diff
@@ -216,7 +216,50 @@ func TestIndexGatewayRingMode_when_TargetIsRead(t *testing.T) {
...
+
+		services, err := c.ModuleManager.InitModuleServices(Read)
+		defer func() {
+			for _, service := range services {
+				service.StopAsync()
+			}
+		}()
 
+		require.NoError(t, err)
+		assert.Equal(t, true, c.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Disabled)
```
salvacorts, line 260: nit: `assert.True`

pkg/loki/modules_test.go
```diff
@@ -216,7 +216,50 @@ func TestIndexGatewayRingMode_when_TargetIsRead(t *testing.T) {
...
+
+		services, err := c.ModuleManager.InitModuleServices(Querier)
+		defer func() {
+			for _, service := range services {
+				service.StopAsync()
+			}
+		}()
+
+		require.NoError(t, err)
+		assert.Equal(t, false, c.Cfg.StorageConfig.BoltDBShipperConfig.IndexGatewayClientConfig.Disabled)
```
salvacorts, line 239: nit: `assert.False`

pkg/loki/modules.go
jeschkies: I think that's a next step :slightly_smiling_face:

PR #6511, opened 2022-06-27
Drone: Do not run pipelines twice on PRs

julienduchesne:
Currently, pipelines that should run in PRs are those that are configured to exclude the `main` or tag refs
However, pipelines are configured to run on PRs, tags and pushes. This means that, for branches in the main repo, not the fork, PRs run the same pipelines for the `pull_request` and `push` events
With this PR, the `push` run should no longer trigger on PRs

The change is:
1. Add a global pipeline trigger filter. Only trigger on the main, `k???` branches, `v*` tags or on PRs
2. When pipelines have steps that should only run on either PR or push/tag, they filter on the `event` in the step `when` condition

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -3,29 +3,26 @@ local archs = ['amd64', 'arm64', 'arm'];
 
 local build_image_version = std.extVar('__build-image-version');
 
-local condition(verb) = {
-  tagMain: {
-    ref: {
-      [verb]:
-        [
-          'refs/heads/main',
-          'refs/heads/k???',
-          'refs/tags/v*',
-        ],
-    },
-  },
-  path(path): {
-    paths: {
-      [verb]: [path],
-    },
-  },
+local onPRs = {
+  event: ['pull_request'],
+};
+
+local onTagOrMain = {
+  event: ['push', 'tag'],
+};
+
+local onPath(path) = {
+  paths+: [path],
 };
 
 local pipeline(name) = {
   kind: 'pipeline',
   name: name,
   steps: [],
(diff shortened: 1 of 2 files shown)


dannykopping:
> Currently, pipelines that should run in PRs are those that are configured to exclude the `main` or tag refs However, pipelines are configured to run on PRs, tags and pushes. This means that, for branches in the main repo, not the fork, PRs run the same pipelines for the `pull_request` and `push` events With this PR, the `push` run should no longer trigger on PRs
> 
> The change is:
> 
> 1. Add a global pipeline trigger filter. Only trigger on the main, `k???` branches, `v*` tags or on PRs
> 2. When pipelines have steps that should only run on either PR or push/tag, they filter on the `event` in the step `when` condition

I think we want to run the pipelines both for the PR and when the PR is merged; we've caught a number of problems this way - especially if the PR was not fully up-to-date with `main` when CI was run.

julienduchesne:
> I think we want to run the pipelines both for the PR and when the PR is merged; we've caught a number of problems this way - especially if the PR was not fully up-to-date with `main` when CI was run.

Agreed. This will still run on `main`. Just not on random branches. Right now, all PRs from Grafana employees (who push to the main repo, not a fork) have two Drone jobs running (PR and branch). It's all the same stuff between both jobs
image

trevorwhitney (approved):
Is there anything we're waiting on for this change? This looks reasonable to me, and I think it's easy to change/revert if not behaving as expected.

julienduchesne:
> Is there anything we're waiting on for this change? This looks reasonable to me, and I think it's easy to change/revert if not behaving as expected.

No, it's ready. Thanks for the ✅

PR #6317, opened 2022-06-06
Add cache usage statistics

dannykopping:
**What this PR does / why we need it**:
This PR adds cache statistics for the chunk, index, and result caches. The statistics are collected in the same mechanism that is used currently for ingester/querier stats. The PR also adds some pertinent info to `metrics.go` which shows cache usage on a per-query, per-tenant basis.

This PR should be read with #6289.

Running `logcli query <other args> --stats` will now return extra stats:

```
Cache.Chunk.Requests 		 24
Cache.Chunk.EntriesRequested 	 36
Cache.Chunk.EntriesFound 	 36
Cache.Chunk.EntriesStored 	 0
Cache.Chunk.BytesSent 		 0 B
Cache.Chunk.BytesReceived 	 50 MB
Cache.Index.Requests 		 22
Cache.Index.EntriesRequested 	 22
Cache.Index.EntriesFound 	 22
Cache.Index.EntriesStored 	 0
Cache.Index.BytesSent 		 0 B
Cache.Index.BytesReceived 	 20 kB
Cache.Result.Requests 		 6
Cache.Result.EntriesRequested 	 6
Cache.Result.EntriesFound 	 6
Cache.Result.EntriesStored 	 0
Cache.Result.BytesSent 		 0 B
Cache.Result.BytesReceived 	 23 kB
```

**Which issue(s) this PR fixes**:
#6318

**Special notes for your reviewer**:
Statistics were not being collected correctly in the `queryrange.StatsCollectorMiddleware`. I believe the statistics were being discarded by the `pkg/querier/queryrange/queryrangebase/results_cache.go` middleware, but creating a new stats context in the `StatsCollectorMiddleware` seems to have fixed the issue. This code is pretty hard for me to reason about, so I would appreciate @cyriltovena's review on that since he was the original author, I believe. See https://github.com/grafana/loki/commit/baa239c75bd0776c8180f883fb4794e75774b430.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] Is this an important fix or new feature? Add an entry in the `CHANGELOG.md`.
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #6525, opened 2022-06-28
Remove cache statistics PR from `release-2.6.x`

dannykopping:
We're going to move these stats (introduced in #6317) to a different line in metrics.go to make it more readable, and include more details. We're reverting this so users don't start using the new stats in the combined line when we'll be refactoring this soon


MichelHollands (commented):
This PR only has a change to the Changelog. Is that correct? I was expecting a change in the code as well.

MichelHollands (approved):
LGTM

PR #6537, opened 2022-06-29
generic index compactor

sandeepsukhani:
To add support for tsdb index compaction and custom retention, this PR makes the compactor generic to support various index types. I have added a method to the compactor for registering the index types that just need to implement the specific interfaces to support compaction and custom retention. I have implemented it for `boltdb-shipper` and will open a follow-up PR for `tsdb` to avoid making this PR too big.

I have added some `ToDo`s to the code to take care of in follow up PRs since this PR is already too big.
Other than addressing those ToDos, I would also move the compactor to the generic `indexshipper` package in a separate PR.

**Checklist**
- [x] Tests updated


sandeepsukhani:
converting this to a draft to avoid merging it before I get to test it internally to build some confidence.

sandeepsukhani:
Tested this for a couple of days in our internal dev cluster. Things are looking fine so going to merge this PR.

PR #6540, opened 2022-06-29
Support dashboards for SSD mode

salvacorts:
This PR adds support for SSD mode on the Loki dashboard mixins. It also supports hiding dashboard panels for Grafana Labs' internal components.

We introduce two new configs:
```
// Support Loki SSD mode on dashboards.
ssd: false,

// Enable dashboard and panels for Grafana Labs internal components.
internal_components: false,
```

And a new mixin `mixin-ssd.libsonnet` is added which sets the `ssd` config to `true`.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


production/loki-mixin/dashboards/loki-operational.libsonnet
```diff
@@ -13,11 +13,24 @@ local utils = import 'mixin-utils/utils.libsonnet';
                                showMultiCluster:: true,
                                clusterLabel:: $._config.per_cluster_label,
 
-                               matchers:: {
+                               hiddenRows:: [
+                                 'Cassandra',
```
trevorwhitney, line 17: should we elevate another top level `$._config` value for toggling cassandra? I agree we don't need it for our cloud dashboards or the compiled mixin, but curious if there's no way to turn it on then there's no reason to keep it IMO.

production/loki-mixin/dashboards/loki-reads-resources.libsonnet
```diff
@@ -2,14 +2,21 @@ local grafana = import 'grafonnet/grafana.libsonnet';
 local utils = import 'mixin-utils/utils.libsonnet';
 
 (import 'dashboard-utils.libsonnet') {
+  local index_gateway_pod_matcher = if !$._config.ssd then 'container="index-gateway"' else 'container="loki", pod=~"(enterprise-logs|loki)-read.*"',
```
trevorwhitney, line 5: WDYT about flipping these so we check against `$._config.ssd` instead of it's inverse? I always find checking the inverse to take just a bit more mental effort, so in a case where there's an action for each branch of the conditional I'd prefer not asserting against the inverse.

production/loki-mixin/dashboards/loki-reads.libsonnet
```diff
@@ -3,6 +3,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
 (import 'dashboard-utils.libsonnet') {
   grafanaDashboards+: {
     local dashboards = self,
+    local showBigTable = false,
```
trevorwhitney, line 6: again do we want to expose this to a `$._config` value like Cassandra?

production/loki-mixin/dashboards/loki-reads.libsonnet
```diff
@@ -20,10 +21,10 @@ local utils = import 'mixin-utils/utils.libsonnet';
 
                          matchers:: {
                            cortexgateway: [utils.selector.re('job', '($namespace)/cortex-gw')],
-                           queryFrontend: [utils.selector.re('job', '($namespace)/query-frontend')],
-                           querier: [utils.selector.re('job', '($namespace)/querier')],
-                           ingester: [utils.selector.re('job', '($namespace)/ingester')],
-                           querierOrIndexGateway: [utils.selector.re('job', '($namespace)/(querier|index-gateway)')],
+                           queryFrontend: [utils.selector.re('job', '($namespace)/%s' % (if !$._config.ssd then 'query-frontend' else '(enterprise-logs|loki)-read'))],
```
trevorwhitney, line 24: same feedback here about asserting against the inverse of the condition, I think flipping the if/else would read a bit nicer.

production/loki-mixin/dashboards/loki-writes-resources.libsonnet
```diff
@@ -2,14 +2,18 @@ local grafana = import 'grafonnet/grafana.libsonnet';
 local utils = import 'mixin-utils/utils.libsonnet';
 
 (import 'dashboard-utils.libsonnet') {
+  local ingester_pod_matcher = if !$._config.ssd then 'container="ingester"' else 'container="loki", pod=~"(enterprise-logs|loki)-write.*"',
```
trevorwhitney, line 5: same comment about asserting against the inverse.

production/loki-mixin/dashboards/loki-writes.libsonnet
```diff
@@ -3,6 +3,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
 (import 'dashboard-utils.libsonnet') {
   grafanaDashboards+: {
     local dashboards = self,
+    local showBigTable = false,
```
trevorwhitney, line 6: another value to possible hoist to `$._config`?

trevorwhitney (commented):
This looks great Salva! Just a few small comments and style suggestions.

WDYT of restructuring the `loki-mixin-compiled` folder to contain an SSD and Distributed version of the compiled JSON? I think that it should wait for a subsequent PR to limit the changeset, but I do think we should do it.

production/loki-mixin/dashboards/loki-operational.libsonnet
salvacorts: I think it makes sense to keep it for users that may be running cassandra.
I don't think we should add a new field to `$._config` for toggling this as only this dashboard is affected, and therefore, IMO it makes sense to add the config on the dashboard itself along with the rest of the dashboard-specific configs ([`showAnnotations` for example](https://github.com/grafana/loki/blob/a5ffb57004d827228d8f2a52bf2fd702d6fe976f/production/loki-mixin/dashboards/loki-operational.libsonnet#L11)).

production/loki-mixin/dashboards/loki-reads-resources.libsonnet
salvacorts: I don't have a preference here. I took this approach so the microservices matcher was before the SSD matcher since: 
- We run all our prod and most of our ops/dev cells in microservices mode
- The matcher for microservices is shorter and easier to read/understand

I'll flip them

salvacorts: I just left a couple of exceptions:

- hiddenRows:
```jsonnet
hiddenRows:: [
  'Cassandra',
] + if !$._config.ssd then [] else [
  'Ingester',
],
```

- addRowIf
```jsonnet
.addRowIf(
  !$._config.ssd,
```

salvacorts:
> WDYT of restructuring the loki-mixin-compiled folder to contain an SSD and Distributed version of the compiled JSON? I think that it should wait for a subsequent PR to limit the changeset, but I do think we should do it.

@trevorwhitney I added it. The compiled mixin now lives at the `loki-mixin-compiled-ssd` directory.

production/loki-mixin/dashboards/loki-reads-resources.libsonnet
trevorwhitney: yeah, I agree with the ones you left. thanks!

production/loki-mixin/dashboards/loki-operational.libsonnet
trevorwhitney: sounds good :+1:

trevorwhitney (approved):
lgtm! looks great Salva, thanks!

PR #6550, opened 2022-06-30
Adding "start_delta" entry to metrics.go line

dannykopping:
This will help us determine if a 3h query (derived from `length`) is for some random 3h period in the past, or within the last 3h.
Knowing this is valuable so we can determine if queries are occurring within the `query_ingesters_within` period, and as such the queriers will be fetching data from all ingesters to satisfy a query.

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


pkg/logql/metrics.go
```diff
@@ -116,6 +116,7 @@ func RecordRangeAndInstantQueryMetrics(
 		"query_type", queryType,
 		"range_type", rt,
 		"length", p.End().Sub(p.Start()),
+		"start_delta", time.Since(p.Start()).Round(time.Minute),
```
sandeepsukhani, line 119: should we log `end_delta` if the intention is to easily find whether the query touches the last x hours of data or not?
Or maybe there is a way to add `length` to `start_delta` to find that out?

dannykopping: @sandeepsukhani done

sandeepsukhani (approved):
LGTM

PR #6552, opened 2022-06-30
Do not fail queries if compactor returns unexpected status code

chaudum:

To avoid showing deleted data, the querier originally failed when
deletes were unavailable. If the compactor becomes unavailable this can
cause Loki to stop responding to queries altogether. This PR changes the
querier to log an error but continue servicing the request when it fails
to get deletes.

E.g. when a tenant does not have deletions enabled, the compactor returns 403 (Forbidden) when listing the deletes on `GET /loki/api/v1/deletes`. https://github.com/grafana/loki/blob/k104/pkg/storage/stores/shipper/compactor/deletion/request_handler.go#L223-L233

Related PR #6368

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


kavirajk (approved):
LGTM :+1:

PR #6562, opened 2022-06-30
Canary: add random start time for the periodic metric test query

slim-bean:

When running a lot of canaries it becomes common that they have a highly overlapping window where they all run the metric test query at the same time, this PR introduces some randomness in the first execution window such that all the canaries should approximately distribute over the size of the interval for which they run the metric test query.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/canary/comparator/comparator.go
+++ b/pkg/canary/comparator/comparator.go
@@ -3,6 +3,7 @@ package comparator
 import (
 	"fmt"
 	"io"
+	"math/rand"
 	"sync"
 	"time"
 
@@ -246,7 +247,10 @@ func (c *Comparator) Size() int {
 
 func (c *Comparator) run() {
 	t := time.NewTicker(c.pruneInterval)
-	mt := time.NewTicker(c.metricTestInterval)
+	// Use a random tick up to the interval for the first tick
+	firstMt := true
+	rand.Seed(time.Now().UnixNano())
+	mt := time.NewTicker(time.Duration(rand.Int63n(c.metricTestInterval.Nanoseconds())))
 	sc := time.NewTicker(c.spotCheckQueryRate)
 	defer func() {
 		t.Stop()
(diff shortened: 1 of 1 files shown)


DylanGuedes (approved):
lgtm

pkg/canary/comparator/comparator.go
```diff
@@ -246,7 +247,10 @@ func (c *Comparator) Size() int {
 
 func (c *Comparator) run() {
 	t := time.NewTicker(c.pruneInterval)
-	mt := time.NewTicker(c.metricTestInterval)
+	// Use a random tick up to the interval for the first tick
+	firstMt := true
+	rand.Seed(time.Now().UnixNano())
+	mt := time.NewTicker(time.Duration(rand.Int63n(c.metricTestInterval.Nanoseconds())))
```
trevorwhitney, line 253: I think the default for this is `1hr`, so we delay the first tick between 0-3600s?

trevorwhitney (approved):
lgtm! thanks ed!

pkg/canary/comparator/comparator.go
```diff
@@ -285,6 +289,11 @@ func (c *Comparator) run() {
 				go c.metricTest(time.Now())
 			}
 			c.metTestMtx.Unlock()
+			if firstMt {
+				// After the first tick which is at a random time, resume at the normal periodic interval
+				firstMt = false
+				mt.Reset(c.metricTestInterval)
+			}
```
jeschkies, line 296: I wonder whether a sleep would be easier to understand.

pkg/canary/comparator/comparator.go
slim-bean: assuming I did this correctly that's the idea, the base unit of a Duration is nanoseconds so we just spread the start time of this test for every canary randomly over the interval

pkg/canary/comparator/comparator.go
slim-bean: a sleep would be more complicated to implement without affecting the other tickers in the select. We don't want to interfere with the spot check test for example, and make it wait up to 1hr before running.

jeschkies: Ok. Then let's go with this solution :slightly_smiling_face:

jeschkies (approved):

PR #6514, opened 2022-06-27
operator: Making all pods and containers compliant with restricted policy

Red-GV:
**What this PR does / why we need it**:
With the release of the [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) in 1.24/1.25 Kubernetes, this PR makes all pods and containers created by the operator (and the operator itself) compliant with the [restricted policy](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted). These changes allow for compliance in native Kubernetes and OCP.

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] Is this an important fix or new feature? Add an entry in the `CHANGELOG.md`.
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #6564, opened 2022-06-30
operator: Addons work in restricted policy

Red-GV:
This PR is a flow up to #6514. After making the operator conform to the `restricted` policy, the addons needs to be updated to work under the same conditions. This was partially done in that PR, but it also broke the `promtail` addon. This PR fixes that addon to allow it to send logs without needing to be `privileged`. 

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operator/hack/addons_dev.yaml
+++ b/operator/hack/addons_dev.yaml
@@ -42,7 +42,7 @@ spec:
               value: /var/run/secrets/kubernetes.io/serviceaccount/token
           args:
             - -c
-            - while true; do logcli query '{job="systemd-journal"}'; sleep 30; done
+            - while true; do logcli query '{namespace="default"}'; sleep 30; done
           securityContext:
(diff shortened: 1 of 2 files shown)


periklis (commented):
After reading through the changes, I have a couple of concerns about the direction taken:
1. What is the reason changing the deployment type from `DaemonSet` to `Deployment`? Usually we prefer the former over the latter because of locality to nodes filesystem (e.g. container log files, journal files).
2. Dropping the journal at least for the OCP addon is imho a no-go. I agree that we can use fluentd/vector from ClusterLoggingOperator for this to work, but it is a cool thing to have w/o installing a second operator when testing the Loki Operator.
3. Dropping `/var/lib/docker/containers` but not `/var/logs/pods`, why? Can you elaborate?

periklis (approved):

Issue #4648, opened 2021-11-04
Feedback wanted: deprecate our Loki FluentBit plugin in favor of native FluentBit output.


owen-d:
Hello! A long time ago we wrote a [plugin](https://grafana.com/docs/loki/latest/clients/fluentbit/) for ingesting logs in Loki from `fluentbit`. These days, there's a native output [option](https://docs.fluentbit.io/manual/pipeline/outputs/loki) for Loki available in fluentbit itself courtesy of @edsiper. I'm opening this issue to solicit feedback from Loki users currently using either of these in hopeful preparation for deprecating our plugin in favor of theirs.

cc @cyriltovena

edsiper:
Hi Loki users, 

As part of the Fluent Bit team, we want to bring a first-class citizen experience with Loki and we would like to know what are the specific missing features in our built-in connector:

- https://docs.fluentbit.io/manual/pipeline/outputs/loki

Since the Golang connector will be deprecated, please let us know what is needed to prioritize on our side. 

thanks.

stevehipwell:
@owen-d I swapped over from the native Fluentd implementation to the native Fluent Bit implementation as soon as Loki v2.4 was release and I've been very happy with both parts.

@edsiper I think there are a few outstanding Loki issues on the Fluent Bit repo that need triaging against the latest versions? Off the top of my head the following areas need looking into, but I've not seen any of them since upgrading to the latest versions:

- Missing logs in Loki
- Loki push failures causing cascading errors

cyriltovena:
@edsiper We have out of order now available, so we should be able to change the implementation of fluentbit to send batches in parallel.

patrick-stephens:
I personally agree with deprecation: there is a fair bit of confusion with mismatches in configuration across the two plugins and so people follow a blog post/etc. for the Grafana one but use the Fluent one and then get failures. There is also the duplication of effort required: implement a feature in one then in the other (maybe slightly differently) plus the fragmentation of features. Having a single official plugin is much preferable for support, documentation, development and testing.

patrick-stephens:
@owen-d what was the outcome of this? Just curious if the plan is to deprecate or not - and when if so?

zecrazytux:
First used the fluent-bit native and the had to switch to Loki's one as we had failures to send to loki after a while (maybe after a disconnection / small network interruption)

krafcima:
[...]

PR #6566, opened 2022-07-04
docs: Improve a confusing link to fluent-bit's loki plugin

ksauzz:

This docs explains `grafana-lok`i plugin in this repo, not `loki` plugin in fluent-bit repo. They has completely different configuration items. So the link makes users confused.

None

### DIFF
--- a/docs/sources/clients/fluentbit/_index.md
+++ b/docs/sources/clients/fluentbit/_index.md
@@ -4,7 +4,7 @@ weight: 50
 ---
 # Fluent Bit Loki Output
 
-[Fluent Bit](https://fluentbit.io/) is a fast and lightweight logs and metrics processor and forwarder that can be configured with the Grafana Loki output plugin to ship logs to Loki. You can define which log files you want to collect using the [`Tail`](https://docs.fluentbit.io/manual/pipeline/inputs/tail) or [`Stdin`](https://docs.fluentbit.io/manual/pipeline/inputs/standard-input) data pipeline input. Additionally, Fluent Bit supports multiple `Filter` and `Parser` plugins (`Kubernetes`, `JSON`, etc.) to structure and alter log lines.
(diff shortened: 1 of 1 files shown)


trevorwhitney:
I guess I don't understand the difference either. The configs between [`config.go`](https://github.com/grafana/loki/blob/main/clients/cmd/fluent-bit/config.go) and the [fluent bit loki output](https://docs.fluentbit.io/manual/pipeline/outputs/loki) docs look very similar to me. Can you please elaborate on the differences between the two plugins you mentioned?

ksauzz:
I didn't do comprehensive comparison between loki and grafana-loki. But at least there are the following differences between them, and they are not compatible each others in my understanding.

_ | loki plugin | grafana-loki plugin
-----|----|------
Case Style | snake_case | camel case 
The number of items | 12 | 26
the format of labels | key=value | {"key": "value"}
data type of auto kubernetes labels | On/Off | true/false

## loki plugin's config keys

```
% curl -s https://raw.githubusercontent.com/fluent/fluent-bit/master/plugins/out_loki/loki.c | grep FLB_CONFIG | egrep -o '".*"' | cut -d, -f1 | tr -d '"'
tenant_id
tenant_id_key
labels
auto_kubernetes_labels
drop_single_key
label_keys
remove_keys
line_format
http_user
http_passwd
```

Additionally it seems `host`, `port` is available by fluent-bit core library.

## grafana-loki plugin's config keys

```
% curl -s https://raw.githubusercontent.com/grafana/loki/main/clients/cmd/fluent-bit/config.go | grep -o 'cfg.Get(".*")' | egrep -o '".*"' | tr -d '"'
URL
TenantID
BatchWait
BatchSize
Timeout
MinBackoff
MaxBackoff
MaxRetries
Labels
LogLevel
AutoKubernetesLabels
RemoveKeys
LabelKeys
DropSingleKey
LineFormat
LabelMapPath
Buffer
BufferType
DqueDir
DqueSegmentSize
DqueSync
DqueName
ca_file
cert_file
key_file
insecure_skip_verify
```

trevorwhitney:
I see now, thank you for that. I was honestly mixed up by the fact there where two different plugin maintained by two different teams, which this change will help with! WDYT of adding something to that paragraph that makes it very clear the "Grafana Loki" plugin is the one in this repo, whereas I assume the fluent bit one is maintained by fluent?

ksauzz:
Thank you for the comment. That sounds reasonable to me. I added simple sentence about `loki` plugin. How about this?

ksauzz:
This is a known issue, and grafana team are considering to drop `grafana-loki` plugin since golang support will be deprecated in fluent-bit.
see https://github.com/grafana/loki/issues/4648

PR #6573, opened 2022-07-05
log addtl info about bytes/shard in planning

owen-d:
Extracting metrics for this has proven otherwise difficult because we don't have a `bytes` template and it's a bit inconvenient to add b/c the function returns an error signature.


dannykopping (approved):
LGTM

PR #6579, opened 2022-07-05
loki-canary: Fix resource leak when using client-side certs

chodges15:
When running canary with client side certificates (introduced in #6310), a new `http.Transport` is being created for each request 🤦‍♂️. As a result, connections are left open until they timeout an hour later. This quickly causes resource exhaustion and errors like `dial tcp 1.1.1.1:1111: connect: cannot assign requested address`. 

This change will reuse the same transport for all requests just like `http.DefaultClient` is reused when not using client certs.

Tested by running loki-canary for 5 days and observing with netstat a constant number of TCP connections to the querier.

### DIFF
--- a/pkg/canary/reader/reader.go
+++ b/pkg/canary/reader/reader.go
@@ -51,13 +51,14 @@ type LokiReader interface {
 
 type Reader struct {
 	header          http.Header
-	tls             bool
+	useTLS          bool
 	clientTLSConfig *tls.Config
 	caFile          string
 	addr            string
 	user            string
 	pass            string
 	tenantID        string
+	httpClient      *http.Client
 	queryTimeout    time.Duration
 	sName           string
 	sValue          string
@@ -77,7 +78,7 @@ type Reader struct {
 
 func NewReader(writer io.Writer,
 	receivedChan chan time.Time,
-	tls bool,
+	useTLS bool,
 	tlsConfig *tls.Config,
 	caFile string,
 	address string,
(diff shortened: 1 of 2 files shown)


DylanGuedes (approved):
lgtm

pkg/canary/reader/reader.go
```diff
@@ -92,6 +93,19 @@ func NewReader(writer io.Writer,
...
+
+	// http.DefaultClient will be used in the case that the connection to Loki is http or TLS without client certs.
+	httpClient := http.DefaultClient
+	if tlsConfig != nil {
+		// For the mTLS case, use a http.Client configured with the client side certificates.
+		rt, err := config.NewTLSRoundTripper(tlsConfig, caFile, func(tls *tls.Config) (http.RoundTripper, error) {
+			return &http.Transport{TLSClientConfig: tls}, nil
+		})
+		if err != nil {
+			panic(fmt.Sprintf("Failed to create HTTPS transport with TLS config: %v", err))
```
dannykopping, line 105: I don't think we should be using `panic`. Can we handle this in a less destructive way?

dannykopping (changes requested):
LGTM, thanks for this! Let's please find a way to not use `panic`

pkg/canary/reader/reader.go
chodges15: Good point, there's no reason for a stack trace to come out here. I've kept this as a fatal error that will use `os.Exit` instead.

PR #6585, opened 2022-07-05
Add integration for SSD (read & write targets)

trevorwhitney:

Adds an integration test for Simple Scalable Deployment (SSD), which comprises just read and write targets.

### DIFF
--- a/integration/loki_micro_services_delete_test.go
+++ b/integration/loki_micro_services_delete_test.go
@@ -65,7 +65,7 @@ func TestMicroServicesDeleteRequest(t *testing.T) {
 
 	require.NoError(t, clu.Run())
 
-	tenantID := randStringRunes(12)
+	tenantID := randStringRunes()
 
 	now := time.Now()
 	cliDistributor := client.New(tenantID, "", tDistributor.HTTPURL().String())
--- a/integration/loki_micro_services_test.go
+++ b/integration/loki_micro_services_test.go
@@ -52,7 +52,7 @@ func TestMicroServicesIngestQuery(t *testing.T) {
 
 	require.NoError(t, clu.Run())
 
-	tenantID := randStringRunes(12)
+	tenantID := randStringRunes()
 
 	now := time.Now()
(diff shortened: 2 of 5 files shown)


dannykopping (approved):
:heart: nice!

How comprehensive are we expecting to go with these? The PR description doesn't hint at what we're trying to achieve here. I love that we're starting with this, though.

For example:
This will currently only test in-memory logs in the head block; can you think of a way that we could ensure that the data is persistable to disk?
And should we test other critical features of SSD like ring communication / replication, etc?

trevorwhitney:
@dannykopping great questions. I think we want to be conservative in expanding these tests because of how expensive they are. At this level, this test makes sure that if we make a change to a component, and forget to enable it or propagate that change to SSD mode, we will have an early warning sign. I would hold off expanding the scope of the test until we run into a bug that is related to persisting data to disk, ring communication / replication, etc. before expanding the test.

integration/loki_micro_services_delete_test.go
```diff
@@ -65,7 +65,7 @@ func TestMicroServicesDeleteRequest(t *testing.T) {
 
 	require.NoError(t, clu.Run())
 
-	tenantID := randStringRunes(12)
+	tenantID := randStringRunes()
```
trevorwhitney, line 68: The linter was complaining that this is always called with 12

dannykopping:
> @dannykopping great questions. I think we want to be conservative in expanding these tests because of how expensive they are. At this level, this test makes sure that if we make a change to a component, and forget to enable it or propagate that change to SSD mode, we will have an early warning sign. I would hold off expanding the scope of the test until we run into a bug that is related to persisting data to disk, ring communication / replication, etc. before expanding the test.

Sounds good to me.
What aspect is expensive right now? Are the tests quite slow?

trevorwhitney:
> What aspect is expensive right now? Are the tests quite slow?

They are slower than the unit tests, but I think the real risk is the potential for bloat here. Every additional integration test adds more time than the addition of a unit test.

PR #6589, opened 2022-07-06
Set ingester PDB to have max_unavailable=0

DylanGuedes:
Change our ingesters policy disruption budget to use max_unavailable pods as 0 instead of 1. Without this, in some cases we get more than 1 ingester unavailable, which creates incidents.

N/A

https://spot.io/resources/kubernetes-autoscaling/5-kubernetes-deployment-strategies-roll-out-like-the-pros/

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


dannykopping:
> Without this, in some cases we get more than 1 ingester unavailable, which creates incidents.

How often does this happen? I'm wondering if this change will affect our rollout speed significantly - have you looked into that?

dannykopping:
> > > Without this, in some cases we get more than 1 ingester unavailable, which creates incidents.
> > 
> > 
> > How often does this happen? I'm wondering if this change will affect our rollout speed significantly - have you looked into that?
> 
> Depending on the state, it will happen for every new rollout. For instance, if you have 4/5 ingesters ready and you trigger a new rollout, k8s will actually drop one of the 4 ready pods. And you're right, this will affect the rollout speed of those scenarios where writes would fail, but it is the tradeoff of keeping our writes healthy.

Cool, thanks for that :+1:

dannykopping (approved):
LGTM

vlad-diachenko (approved):
LGTM

PR #6591, opened 2022-07-06
docs: Fixed examples within Loki and Promtail docs

Alex3k:

_Loki Simple Scalable Docs Changes_
The documentation for deploying Loki in Simple Scalable mode via Helm chart suggested to override values using `--set`. This is quite cumbersome when overriding many multiple values, therefore I have included an example to use `--values` also. As part of this, I also pointed to some example values files within the Loki Simple Scalable Helm chart repo

__Promtail Doc Changes__
The Daemonset yaml example had broken yaml and therefore didn't work. I fixed it and provided an example promtail config to get users started instead of them having to go and find another example elsewhere on the website. This now means folks can copy and paste that yaml blob (once they have replaced `{YOUR_LOKI_ENDPOINT}` with a suitable value) and it will work.


docs/sources/installation/simple-scalable-helm.md
```diff
@@ -22,13 +22,13 @@ within a Kubernetes cluster.
     helm repo add grafana https://grafana.github.io/helm-charts
     ```
 
-1. Update the chart repository:
+2. Update the chart repository:
```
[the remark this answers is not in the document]
Alex3k: Very cool! I wasn't aware of that thanks for the guidance :)

Alex3k:
@trevorwhitney I am not sure why docs changes have caused tests to fail :/ Is there a way to re run this as it seems to be grumbling about a port already in use?

docs/sources/installation/simple-scalable-helm.md
```diff
@@ -42,12 +42,20 @@ within a Kubernetes cluster.
...
+        helm upgrade --install loki grafana/loki-simple-scalable --values {PATH_TO_OVERRIDES_VALUES_FILE}
+        ```
+
+    - Deploy with added custom configuration, useful for overriding a small amount of values:
 
         ```bash
         helm upgrade --install loki grafana/loki-simple-scalable --set "key1=val1,key2=val2,..."
         ```
 
+You can find examples of deploying Loki and Grafana Enterprise Logs within the [docs folder of the Simple Scalable Chart](https://github.com/grafana/helm-charts/tree/main/charts/loki-simple-scalable/docs/examples).
```
KMiller-Grafana, line 57:
```suggestion
Find deployment examples at [https://github.com/grafana/helm-charts/tree/main/charts/loki-simple-scalable/docs/examples](https://github.com/grafana/helm-charts/tree/main/charts/loki-simple-scalable/docs/examples).
```

docs/sources/installation/simple-scalable-helm.md
```diff
@@ -112,6 +120,10 @@ spec:
     - {{ .Values.ingress.host }}
 ```
 
+## Run Promtail within Kubernetes as a Daemonset
+
+In order to run Promtail within Kubernetes to gather logs from each nodes it's recommended you start with the example within the [Promtail installation docs](../../clients/promtail/installation/#daemonset-recommended)
```
KMiller-Grafana, line 125:
```suggestion
In order to run Promtail within Kubernetes to gather logs from each node, start with the [Promtail installation](../../clients/promtail/installation/#daemonset-recommended) documentation example.
```

KMiller-Grafana (changes requested):
I've suggested rewordings. These will improve the already-good content, as well as follow style guidelines.

Alex3k:
Thanks @KMiller-Grafana, I have just accepted the rewordings.

KMiller-Grafana (approved):
Thank you for making the changes.

Alex3k:
Hi @KMiller-Grafana, when you get a comment would you be able to merge this please? Thanks in advance :)

KMiller-Grafana:
Hi @Alex3k. I've asked @trevorwhitney to review and merge. Thanks for your patience.

docs/sources/clients/promtail/installation.md
```diff
@@ -133,45 +185,3 @@ roleRef:
     name: promtail-clusterrole
     apiGroup: rbac.authorization.k8s.io
 ```
-
-### Sidecar
```
trevorwhitney, line 137: Why are we completely removing the sidecar example, just because of the added complexity it adds? I do think this is still worth while to document, if we don't want it here could we consider moving it somewhere else?

trevorwhitney (changes requested):
Sorry for the long radio silence on this, it got lost over my vacation in July. I think this looks great, with just the one question about what we should do with the sidecar example? I'm ok if we think it's not suited for a getting started doc, but I do want to make sure we capture it somewhere for people who want to deploy promtail in that way. WDYT?

docs/sources/clients/promtail/installation.md
Alex3k: @trevorwhitney you're right. My thought process was that it didn't help a getting started experience but it is useful to have documented. It should stay in the install doc - it's the best place for it as it's the most relevant place compared to other pages.

Alex3k: @trevorwhitney following that one - the current sidecar example isn't a sidecar - it just deploys promtail as a deployment. I suggest we actually do remove this for now and if folks want to deploy it as a side car they can infer how to do that using the above examples. I feel that is safer than providing config that doesn't actually do what we say it does.

What do you think?

trevorwhitney: Sounds good!

trevorwhitney (approved):
LGTM in light of the sidecar not actually being a sidecar, we can add that documentation correctly at a later time in another place. Thanks!

Alex3k:
@trevorwhitney when you get an opportunity could you click merge please? :)

PR #6611, opened 2022-07-07
Append .x to the versioned directories for the website sync

jdbaldry:
Signed-off-by: Jack Baldry <jack.baldry@grafana.com>

Will manually backport to `release-2.5.x` and `release-2.6.x`

PR #6612, opened 2022-07-07
Append .x to the versioned directories for the website sync (#6611)

jdbaldry:
This ensures documentation backported to this release branch is published to the correct website directory which is "v2.5.x".

Logic in the workflow determines the major and minor versions but we need to append the ".x" part as done for main in #6611.

### DIFF
--- a/.github/workflows/publish-technical-documentation-release.yml
+++ b/.github/workflows/publish-technical-documentation-release.yml
@@ -70,4 +70,4 @@ jobs:
           host: "github.com"
           github_pat: "${{ secrets.GH_BOT_ACCESS_TOKEN }}"
           source_folder: "docs/sources"
-          target_folder: "content/docs/loki/${{ steps.target.outputs.target }}"
+          target_folder: "content/docs/loki/${{ steps.target.outputs.target }}.x"


jdbaldry:
Bump on this

MichelHollands:
I don't understand what this does. Also it might need a rebase: some promtail changes show up in the diff.

MichelHollands:
@jdbaldry  I don't understand what this does. Also it might need a rebase: some promtail changes show up in the diff.

jdbaldry:
I believe the promtail changes are required as part of the fix for CI which is failing on an expired certificate if I remember correctly.

The PR itself makes sure that documentation published from the `release-2.5.x` branch goes to the `content/docs/loki/v2.5.x` directory in the website repository. Without this patch, it will go to the `v2.5` directory.

If you are never intending to release from that branch, we are probably fine to close this and move on. Otherwise we are going to need to look deeper into the CI failures.

jdbaldry:
I've removed the attempts to fix the certificate error in CI so this commit is only updating the website target sync directory.

We will need to override CI to merge this PR but I believe it is worthwhile to avoid publishing confusing content to the website.

@grafana/loki-squad could somebody override CI and merge this?

PR #6616, opened 2022-07-07
Add push route to query-tee

salvacorts:

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


trevorwhitney:
One thing I want to make sure we don't forget to test is what happens when one of the downstream targets in unavailable.

salvacorts:
> One thing I want to make sure we don't forget to test is what happens when one of the downstream targets in unavailable.

The query-tee returns the answer from the preferred backend configured. As far as I've seen, the process of forwarding a request across the backends is done async. So it doesn't block until all backends have answered back to forward new requests or answer back.

PR #6623, opened 2022-07-07
Groupcache performance improvements

MasslessParticle:
We suspected that we were unnecessarily allocating memory for `Fetch` operations. This brings the allocs down a little but it turns out groupcache is pretty optimized as is.

```
Fetch hits
-------------------------------------------------------------------------------------------------------------------------------
Before: 
tpatterson@pop-os cache (groupcache *%) $ go test -bench BenchmarkGroupCacheFetch -run=^$  -benchmem -memprofile memprofile.out
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/storage/chunk/cache
BenchmarkGroupCacheFetch-8       1277026               965.6 ns/op           472 B/op         11 allocs/op

After:
tpatterson@pop-os cache (groupcache *%) $ go test -bench BenchmarkGroupCacheFetch -run=^$  -benchmem -memprofile memprofile.out
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/storage/chunk/cache
BenchmarkGroupCacheFetch-8       1889569               651.0 ns/op           264 B/op          8 allocs/op

Fetch misses
-------------------------------------------------------------------------------------------------------------------------------
tpatterson@pop-os cache (groupcache *%) $ go test -bench BenchmarkGroupCacheFetch -run=^$  -benchmem -memprofile memprofile.out
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/storage/chunk/cache
BenchmarkGroupCacheFetch-8       1000000              1356 ns/op             504 B/op         15 allocs/op
```

There's not much to be done about cache missed because a miss is the longest path the cache can take.

### DIFF
--- a/pkg/storage/chunk/cache/groupcache.go
+++ b/pkg/storage/chunk/cache/groupcache.go
@@ -159,11 +159,12 @@ func (c *GroupCache) Stats() *groupcache.Stats {
 }
 
 type group struct {
-	cache           *groupcache.Group
-	logger          log.Logger
-	wg              *sync.WaitGroup
-	cacheType       stats.CacheType
-	requestDuration *prometheus.HistogramVec
+	cache         *groupcache.Group
+	logger        log.Logger
+	wg            *sync.WaitGroup
+	cacheType     stats.CacheType
+	fetchDuration prometheus.Observer
+	storeDuration prometheus.Observer
 }
 
 func (c *GroupCache) NewGroup(name string, ct stats.CacheType) Cache {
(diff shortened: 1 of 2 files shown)


pkg/storage/chunk/cache/groupcache.go
```diff
@@ -175,18 +176,21 @@ func (c *GroupCache) NewGroup(name string, ct stats.CacheType) Cache {
...
-			Name:        "groupcache_request_duration_seconds",
-			Help:        "Total time spent in seconds doing groupcache requests.",
-			Buckets:     instrument.DefBuckets,
-			ConstLabels: prometheus.Labels{"cache_type": string(ct)},
-		}, []string{"operation"}),
+		cache:         groupcache.NewGroup(name, 1<<30, missGetter),
+		logger:        c.logger,
+		wg:            &c.wg,
+		cacheType:     ct,
+		fetchDuration: requestDuration.WithLabelValues("fetch"),
```
MasslessParticle, line 192: Turns out, each call to `WithLabelValues` incurs an alloc

PR #6626, opened 2022-07-08
Don't add the memberlist label as a selector to services.

cstyan:
Adding this selector can result in outages caused by all write traffic being routed to only new distributors during initial migrations to memberlists, causing the few new distributors at the start of a rollout to be OOM killed.

cc @JordanRushing @vlad-diachenko @slim-bean 

### DIFF
--- a/production/ksonnet/loki/boltdb_shipper.libsonnet
+++ b/production/ksonnet/loki/boltdb_shipper.libsonnet
@@ -76,6 +76,6 @@
   else {},
 
   compactor_service: if $._config.using_boltdb_shipper then
-    k.util.serviceFor($.compactor_statefulset)
+    k.util.serviceFor($.compactor_statefulset, $._config.service_ignored_labels)
   else {},
 }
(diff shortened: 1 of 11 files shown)


chaudum (approved):
Great find :tada: !

kavirajk (approved):
Wow. great find. :rocket: Thanks @cstyan. 

Curious how did you endup finding it?

cstyan:
@kavirajk First I looked at the distributor services via `kubectl describe service` and noticed that our loki deployments had the selector but mimir deployments did not. Next I started looking for what could be different. I knew that we created the service from a deployment or statefulset via kausals `serviceFor` to create the service, and I noticed that it by default took all the labels from a deployment/statefulset and added them as selectors. However, the that `serviceFor` does take the optional `ignoreLabels`, so then I looked in Mimir's jsonnet for where they were adding those labels. 

I missed it earlier when writing our `memberlist.libsonnet` because the definition of which labels to ignore was in a different file.

kavirajk:
Nice thinking. thanks for the explanation @cstyan :)

PR #6627, opened 2022-07-08
build and hook tsdb compaction code to the generic index compactor

sandeepsukhani:
Adds support for tsdb compaction including custom retention.
The way it works is it downloads the index and adds everything(requiring compaction) to `tsdb.Builder`.
While applying custom retention it works on the `tsdb.Builder` and keeps track of what chunks to add/delete to/from the index.
Finally, when the compactor calls for conversation `ToIndexFile`, it does all the add/delete chunk operations and makes `tsdb.Builder` build the index file.

I have not focused too much on performance, we can always benchmark and improve it.
We can use a pool in some places like reading the downloaded index but don't want to complicate the code too much with the initial implementation.

**Checklist**
- [x] Tests updated


owen-d (approved):
LGTM, but I think we can remove `compact.go` now :)

PR #6630, opened 2022-07-08
[TSDB]  fingerprint offsets is now correctly 16 byte aligned

owen-d:
This contains a few fixes I've found over the past few days of debugging TSDB operations (sorry for being all over the place), but the important bits are
1) Correctly calculates `storage.SeriesRef` offsets in the `fingerprintOffsets` table in https://github.com/grafana/loki/commit/7a34b15018d7fd8596f460018ce01cc7b78234d8. This resulted in an incredibly hard to find bug in sharding code because it wasn't:
   * Using section offsets to increase the starting offsets for our series references
   * Using 16 byte offsets of the _file position_ not _series number_

It caused everything but the last shard to be skipped most off the time because the actual references were waaay higher than expected
2) Simplifies the query planning algorithm to consistently prefer smaller units of work.

### DIFF
--- a/pkg/ingester/index/bitprefix.go
+++ b/pkg/ingester/index/bitprefix.go
@@ -107,10 +107,7 @@ func (ii *BitPrefixInvertedIndex) validateShard(shard *astmapper.ShardAnnotation
 		return nil
 	}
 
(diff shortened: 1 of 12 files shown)


cyriltovena (approved):
LGTM

PR #6642, opened 2022-07-08
Loki: Do not store exact log line duplicates (now including unordered inserts)

slim-bean:

The behavior for Loki handling of log lines has been like this:

* Multiple log lines at the exact same timestamp to the ns precision which have _different_ content are accepted and stored
* Multiple log lines at the exact same timestamp to the ns precision which have the _same_ content are deduplicated upon insertion into a chunk.

This behavior was changed when we added support for receiving and storing unordered logs.

This PR partially addresses the issue by making sure that inserts into the unordered head block are deduped according to those rules.

However, I believe it's still possible to insert exact duplicates if it's done so across multiple blocks in a chunk, and this is very difficult to solve for, instead I think we should accept these and de-dupe them at query time.

At query time we de-duplicate exact lines across chunks but not within a single chunk (because the rules were such that we would never have exact dupes in a single chunk), but I think to completely handle this we would need to have exact de-duplication within a chunk at query time also.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/chunkenc/unordered_test.go
+++ b/pkg/chunkenc/unordered_test.go
@@ -144,7 +144,7 @@ func Test_Unordered_InsertRetrieval(t *testing.T) {
 		{
 			desc: "ts remove exact dupe forward",
 			input: []entry{
-				{0, "a"}, {0, "b"}, {0, "b"}, {1, "c"},
+				{0, "a"}, {0, "b"}, {1, "c"}, {0, "b"},
 			},
 			exp: []entry{
 				{0, "a"}, {0, "b"}, {1, "c"},
@@ -154,7 +154,7 @@ func Test_Unordered_InsertRetrieval(t *testing.T) {
 		{
 			desc: "ts remove exact dupe backward",
 			input: []entry{
-				{0, "a"}, {0, "b"}, {0, "b"}, {1, "c"},
+				{0, "a"}, {0, "b"}, {1, "c"}, {0, "b"},
 			},
 			exp: []entry{
 				{1, "c"}, {0, "b"}, {0, "a"},


pkg/chunkenc/unordered_test.go
```diff
@@ -141,6 +141,26 @@ func Test_Unordered_InsertRetrieval(t *testing.T) {
 			},
 			dir: logproto.BACKWARD,
 		},
+		{
+			desc: "ts remove exact dupe forward",
+			input: []entry{
+				{0, "a"}, {0, "b"}, {0, "b"}, {1, "c"},
```
trevorwhitney, line 147: I think I'm missing something here. Should we have a test that has unordered input to make sure the dedupes are still correct as that's the case this PR is addressing?

trevorwhitney (commented):
Are we missing a test for the unordered case, or am I missing something?

kavirajk:
@slim-bean we may not need this workaround if we make /push safe to retry
https://github.com/grafana/loki/pull/6650

owen-d (approved):
I prefer this behavior. It should also handle deduping across blocks as-is because we rebuild out of order chunks before flushing to ensure that once flushed to storage, all chunks maintain internal ordering for performance: https://github.com/grafana/loki/blob/main/pkg/chunkenc/memchunk.go#L914

pkg/chunkenc/unordered.go
```diff
@@ -116,6 +116,15 @@ func (hb *unorderedHeadBlock) Append(ts int64, line string) error {
 	}
 	displaced := hb.rt.Add(e)
 	if displaced[0] != nil {
+		// While we support multiple entries at the same timestamp, we _do_ de-duplicate
+		// entries at the same time with the same content, iterate through any existing
+		// entries and ignore the line if we already have an entry with the same content
+		for _, et := range displaced[0].(*nsEntries).entries {
```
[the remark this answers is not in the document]
slim-bean: it would depend on how many log lines are sent with the exact same timestamp, in most realistic cases this should be ok.

pkg/chunkenc/unordered_test.go
slim-bean: The exact code being tested currently doesn't really care if it's ordered or not, but I did update the tests to be unordered in the spirit of the PR and maybe to better future proof us, good eye though, appreciate you taking the time to review the tests :)

PR #6644, opened 2022-07-08
Introduce Groupcache

MasslessParticle:
This PR introduces an in-process distributed cache to loki as a possible cache. The cache is implemented via the `groupcache` library and configured from the `common` configuration section.

This cache is still considered experimental and testing/tuning is ongoing


dannykopping (approved):
LGTM

PR #6645, opened 2022-07-08
Fix inconsistent coverage in the distributor package

MasslessParticle:
We have a handy code coverage tool that lets us know our change's impacts on code coverage. Something strange that's been happening is the distributor package has been consistently providing non-deterministic results:

coverage

Turns out [this line](https://github.com/grafana/loki/blob/e7fb10f4567f191a8dbc2344b1e62c8072dd695e/pkg/distributor/ingestion_rate_strategy.go#L48) wasn't explicitly tested in it's unit test and was being accidentally tested as part of another test in the same package. As a result, code coverage was inconsistent. 

This PR explicitly tests this code path so the coverage results of our tests are always consistent.


MasslessParticle:
Code coverage in this PR may still be inconsistent because it's still comparing against the flaky version. Hence the `+ 0.3`

MichelHollands (approved):
LGTM

PR #6650, opened 2022-07-09
feat(endpoint): Make `/push` endpoint safe to retry (idempotent).

kavirajk:

This PR make `/push` endpoint idempotent thus safe to retry.

### Problem
We started seeing some duplicate log entries with Ns precision on our cells.

Screenshot_2022-07-11_14-22-55
Screenshot_2022-07-11_14-22-41
Screenshot_2022-07-11_14-22-19

### Cause of the problem
Say our setup runs with replication factor=3.

Single client request recieved on distributors will be sent to three ingesters. And because of quoram, we return success to the client at least 2 of them succeed. Otherwise we ask client to retry.

Consider the case, where it failed in 2 ingesters(A and B) and success in 1 ingester(C), meaning that one ingester already have this log in it's block. Now client retries and if ingester C already got some more requests making it's last line timestamp different, it can accept the old log during retry, making it as duplicate entry.

Also NOTE: this retry can happen for many other reasons, say ingester sent success response but client haven't got it (because of lost in the network timeout between client and distributor)

### Solution. Make `/push` endpoint idempotent and thus safe to retry by client

1. promtail client now sends unique idempotent key per batch push. (single `logproto.PushRequest`)
2. distributor forwards the idempotent key to the all the downstream ingesters (based on replication factor)
3. every ingesters check if log lines were already added to it's block based on idempotent key. if It's already added, then it's a `no-op` for the ingester.

### Rollingout

We cannot discard any client request that doesn't have `idempotent-key` as during promtail rollout, there may be older versions of promtail keep sending the batches. 

So we have taken into account that even batches without idempotent-key would be accepted by ingester during the rollouts. We can make it strict once we have new versions of promtail everywhere.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated


kavirajk:
If we merge this PR, we may not need workaround @slim-bean proposed here.
https://github.com/grafana/loki/pull/6642

jeschkies (commented):
Nice. This is an awesome timing. Would you update the HTTP API as well? Did you consider generating a key by hashing the log event in case it's not supplied?

MasslessParticle:
This looks awesome! 

What do you think about renaming `IdempotentKey` to just `ID` or `RequestID`? `IdempotentKey` is really specific to what we're accomplishing. If it were renamed, we would be communicating what it is rather than what it's for.

chaudum:
> This looks awesome!
> 
> What do you think about renaming `IdempotentKey` to just `ID` or `RequestID`? `IdempotentKey` is really specific to what we're accomplishing. If it were renamed, we would be communicating what it is rather than what it's for.

+1

What about `PushID`, which is a bit less generic than `RequestID`.

kavirajk:
> Nice. This is an awesome timing. Would you update the HTTP API as well? Did you consider generating a key by hashing the log event in case it's not supplied?

> Would you update the HTTP API as well? 

@jeschkies the payload for http push API is `logproto.PushRequest`, I added pushID (idempotentKey) directly there making the API safe to retry. Is there anything else you mean here?

> Did you consider generating a key by hashing the log event in case it's not supplied?

On the promtail side, yes we auto generate UUID. But on the loki side, no. Because considering the roll-out process, we will have two version of promtail (one that pass idempotent Key and one doesn't, older version). But loki has to accept requests from both.

kavirajk:
@chaudum @MasslessParticle thanks for the suggestion. totally make sense. renamed idempotentKey -> pushID.

chaudum (approved):
lgtm!

I would add the pushID to the log line when promtail retries: https://github.com/grafana/loki/blob/main/clients/pkg/promtail/client/client.go#L369

pkg/distributor/distributor.go
```diff
@@ -326,7 +326,7 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
 			if sp := opentracing.SpanFromContext(ctx); sp != nil {
 				localCtx = opentracing.ContextWithSpan(localCtx, sp)
 			}
-			d.sendSamples(localCtx, ingester, samples, &tracker)
+			d.sendSamples(localCtx, req.PushID, ingester, samples, &tracker)
```
cyriltovena, line 329: Are we not going to have a problem if multiple streams are created with the same pushID and end up on the same replica set ?

kavirajk: But this pushID is unique per batch. 

Unless the same exact batch is retried, ingesters will never see the same pushID.

I'm failing to understand how mutiple streams in the batch can cause problem here. Am i missing something?

cstyan: @kavirajk I think I see what Cyril is saying. We could have a lot of series in the push request, and it's likely that some might have overlapping ingester replica sets. Here's how I understand the code you've written at the moment.

 - we start by splitting a push request into individual streams then validating each streams labels and entries (line 232 to 282)
- at the end of the loop for the above we append the key (user id + stream labels) for the stream as well as the stream tracker (which contains just this stream)
- next we loop over the keys and find the ingester replica set for that key, then we append to a slice of samples for each ingester by looping over the set of instances in the replica set (line 300 to 314)
- finally we loop over the list of 'samples for each ingester' and send the samples (stream tracker) to the ingester, using the original push requests push ID as the idempotent identifier (line 321 to 332)

I think the way we can end up rejecting valid Pushes is as follows;
- we have a push request come into the distributor with a request ID of 1234
- this push request has a sample for "series A" and a sample for "series B"
- the replica set for "series A" resolves to ingesters 0-2
- the replica set for "series B" resolves to ingesters 1-3
- the push request generated for "series A" to be sent to ingesters 0-2 has an 'idempotent identifier' of 1234
- the push request generated for "series B" to be sent to ingesters 1-3 also has an 'idempotent identifier' of 1234

So depending on which push request reaches each ingester first they will accept and reject some samples for **different** series from the same original push request.

MasslessParticle:
Seems like we abandoned this. I'm going to close it for now. Please reopen if this is in error!

PR #6655, opened 2022-07-11
bumped version of grafana's memberlist fork

vlad-diachenko:
bumped version of grafana's memberlist fork to have the fix that allows to disable labels check that is required for the migration.
https://github.com/grafana/memberlist/pull/3


dannykopping (approved):
LGTM

PR #6656, opened 2022-07-11
[promtail] Add support for journal matches (only conjuntions)

carlospeon:
Do not process all journal logs. Add an option to add matches to the journal reader.

**Checklist**
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 ##### Enhancements
 * [6395](https://github.com/grafana/loki/pull/6395) **DylanGuedes**: Add encoding support
 * [6828](https://github.com/grafana/loki/pull/6828) **alexandre1984rj** Add the BotScore and BotScoreSrc fields once the Cloudflare API returns those two fields on the list of all available log fields.
+* [6656](https://github.com/grafana/loki/pull/6656) **carlospeon**: Allow promtail to add matches to the journal reader
 
 ##### Fixes
 * [6766](https://github.com/grafana/loki/pull/6766) **kavirajk**: fix(logql): Make `LabelSampleExtractor` ignore processing the line if it doesn't contain that specific label. Fixes unwrap behavior explained in the issue https://github.com/grafana/loki/issues/6713
--- a/clients/pkg/promtail/scrapeconfig/scrapeconfig.go
+++ b/clients/pkg/promtail/scrapeconfig/scrapeconfig.go
@@ -160,6 +160,10 @@ type JournalTargetConfig struct {
 	// Path to a directory to read journal entries from. Defaults to system path
 	// if empty.
 	Path string `yaml:"path"`
+
+	// Journal matches to filter. Character (+) is not supported, only logical AND
+	// matches will be added.
+	Matches string `yaml:"matches"`
 }
 
 // SyslogTargetConfig describes a scrape config that listens for log lines over syslog.
--- a/clients/pkg/promtail/targets/journal/journaltarget.go
+++ b/clients/pkg/promtail/targets/journal/journaltarget.go
@@ -174,12 +174,26 @@ func journalTargetWithReader(
 		return nil, errors.Wrap(err, "parsing journal reader 'max_age' config value")
 	}
 
-	cfg := t.generateJournalConfig(journalConfigBuilder{
+	cb := journalConfigBuilder{
 		JournalPath: targetConfig.Path,
 		Position:    position,
 		MaxAge:      maxAge,
 		EntryFunc:   entryFunc,
-	})
+	}
+
+	matches := strings.Fields(targetConfig.Matches)
+	for _, m := range matches {
+		fv := strings.Split(m, "=")
+		if len(fv) != 2 {
+			return nil, errors.New("Error parsing journal reader 'matches' config value")
+		}
+		cb.Matches = append(cb.Matches, sdjournal.Match{
+			Field: fv[0],
+			Value: fv[1],
+		})
+	}
+
+	cfg := t.generateJournalConfig(cb)
 	t.r, err = readerFunc(cfg)
 	if err != nil {
 		return nil, errors.Wrap(err, "creating journal reader")
@@ -208,6 +222,7 @@ func journalTargetWithReader(
 type journalConfigBuilder struct {
 	JournalPath string
 	Position    string
+	Matches     []sdjournal.Match
 	MaxAge      time.Duration
 	EntryFunc   journalEntryFunc
 }
@@ -221,6 +236,7 @@ func (t *JournalTarget) generateJournalConfig(
 
 	cfg := sdjournal.JournalReaderConfig{
 		Path:      cb.JournalPath,
+		Matches:   cb.Matches,
 		Formatter: t.formatter,
 	}
 
--- a/clients/pkg/promtail/targets/journal/journaltarget_test.go
+++ b/clients/pkg/promtail/targets/journal/journaltarget_test.go
@@ -385,3 +385,38 @@ func Test_MakeJournalFields(t *testing.T) {
 	}
 	assert.Equal(t, expectedFields, receivedFields)
 }
+
+func TestJournalTarget_Matches(t *testing.T) {
+	w := log.NewSyncWriter(os.Stderr)
+	logger := log.NewLogfmtLogger(w)
+
+	testutils.InitRandom()
+	dirName := "/tmp/" + testutils.RandName()
+	positionsFileName := dirName + "/positions.yml"
+
+	// Set the sync period to a really long value, to guarantee the sync timer
+	// never runs, this way we know everything saved was done through channel
+	// notifications when target.stop() was called.
+	ps, err := positions.New(logger, positions.Config{
+		SyncPeriod:    10 * time.Second,
+		PositionsFile: positionsFileName,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	client := fake.New(func() {})
+
+	cfg := scrapeconfig.JournalTargetConfig{
+		Matches: "UNIT=foo.service PRIORITY=1",
+	}
+
+	jt, err := journalTargetWithReader(NewMetrics(prometheus.NewRegistry()), logger, client, ps, "test", nil,
+		&cfg, newMockJournalReader, newMockJournalEntry(nil))
+	require.NoError(t, err)
+
+	r := jt.r.(*mockJournalReader)
+	matches := []sdjournal.Match{{Field: "UNIT", Value: "foo.service"}, {Field: "PRIORITY", Value: "1"}}
+	require.Equal(t, r.config.Matches, matches)
+	client.Stop()
+}
--- a/docs/sources/clients/promtail/scraping.md
+++ b/docs/sources/clients/promtail/scraping.md
@@ -97,6 +97,7 @@ scrape_configs:
       json: false
       max_age: 12h
       path: /var/log/journal
+      matches: _TRANSPORT=kernel
       labels:
         job: systemd-journal
     relabel_configs:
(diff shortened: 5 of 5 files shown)


CHANGELOG.md
```diff
@@ -89,6 +89,7 @@ Here is the list with the changes that were produced since the previous release.
 * [6102](https://github.com/grafana/loki/pull/6102) **timchenko-a**: Add multi-tenancy support to lambda-promtail.
 * [6099](https://github.com/grafana/loki/pull/6099) **cstyan**: Drop lines with malformed JSON in Promtail JSON pipeline stage.
 * [5715](https://github.com/grafana/loki/pull/5715) **chaudum**: Allow promtail to push RFC5424 formatted syslog messages
+* [6656](https://github.com/grafana/loki/pull/6656) **carlospeon**: Allow promtail to add matches to the journal reader
```
DylanGuedes, line 92: this should be on the unreleased section instead

clients/pkg/promtail/targets/journal/journaltarget.go
```diff
@@ -174,12 +174,26 @@ func journalTargetWithReader(
...
-	cfg := t.generateJournalConfig(journalConfigBuilder{
+	cb := journalConfigBuilder{
 		JournalPath: targetConfig.Path,
 		Position:    position,
 		MaxAge:      maxAge,
 		EntryFunc:   entryFunc,
-	})
+	}
+
+	matches := strings.Fields(targetConfig.Matches)
```
DylanGuedes, line 184: hmm just double checking, if matches are empty will still all journal be checked?

DylanGuedes (commented):
hey thanks for your contribution, this is fantastic. Added a nit and a question, lmk if they are too confusing.

clients/pkg/promtail/targets/journal/journaltarget.go
carlospeon: Yes. Before PR sdjournal.JournalReaderConfig is created with no matches at all, and all journal is processed.

dannykopping, line 184: Can you please add some tests for this @carlospeon?

dannykopping (commented):
Thank you for your contribution @carlospeon :+1: 
Added a few nits for clarity and robustness

clients/pkg/promtail/targets/journal/journaltarget.go
carlospeon: Sure, tests added.

docs/sources/clients/promtail/scraping.md
```diff
@@ -109,7 +110,7 @@ here for reference. The `max_age` field ensures that no older entry than the
 time specified will be sent to Loki; this circumvents "entry too old" errors.
 The `path` field tells Promtail where to read journal entries from. The labels
 map defines a constant list of labels to add to every journal entry that Promtail
-reads.
+reads. The `matches` field adds journal filters (only conjuctions are supported).
```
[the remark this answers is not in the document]
carlospeon: Ok, refrased.

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
```diff
@@ -159,6 +159,9 @@ type JournalTargetConfig struct {
 	// Path to a directory to read journal entries from. Defaults to system path
 	// if empty.
 	Path string `yaml:"path"`
+
+	// Journal matches to filter. Disjunctions (+) not supported.
+	Matches string `yaml:"matches"`
```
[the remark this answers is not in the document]
carlospeon: I though about it before submitted this PR, but if "github.com/coreos/go-systemd/sdjournal" support logical AND (some work can be seen in their source code) it looked a bit weird to me to add a single + character as a list item.

So finally I decided to go with a string and the idea to set here the same arguments that you can use with journalctl, for example "_TRANSPORT=kernel + PRIORITY=3", but I don't have a strong opinion which two options is better.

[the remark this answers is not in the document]
dannykopping: Thanks for the detail :+1: 
I think we shouldn't let the implementation detail of a library we use bubble up to the end-user, so perhaps we can specify the matches as a list of conditions (I think this is the most ergonomic / idiomatic "YAML way"), and then provide them to the library in the way it requires.

clients/pkg/promtail/targets/journal/journaltarget_test.go
```diff
@@ -385,3 +385,38 @@ func Test_MakeJournalFields(t *testing.T) {
 	}
 	assert.Equal(t, expectedFields, receivedFields)
 }
+
+func TestJournalTarget_Matches(t *testing.T) {
```
dannykopping, line 389: Would you mind adding another test (or expanding this one) to account for multiple match conditions, please?

docs/sources/clients/promtail/scraping.md
```diff
@@ -109,7 +110,8 @@ here for reference. The `max_age` field ensures that no older entry than the
 time specified will be sent to Loki; this circumvents "entry too old" errors.
 The `path` field tells Promtail where to read journal entries from. The labels
 map defines a constant list of labels to add to every journal entry that Promtail
-reads.
+reads. The `matches` field adds journal filters (the character `+` is not supported,
+only logical `AND` filters are added).
```
dannykopping, line 114: If we make `matches` a list, we can probably change this to:
```suggestion
reads. Multiple `matches` can be defined as a list of journal filters, which must _all_ match in order for the entry to be considered.
```

clients/pkg/promtail/targets/journal/journaltarget.go
dannykopping: Thank you!

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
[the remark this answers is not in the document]
carlospeon: I think the problem is precisely the implementation detail of the library that promtail uses. It uses and array of matches and this decision limits the way you could set AND/OR matches. Maybe that's the cause that this library doesn't implement it at all taking into account that Journal supports AND/OR matches, and the library is just a binding of journald code.

Journal support filters with logical AND and/or logical OR, so I just though how to support that expression of filters in promtail without bringing a no ideal (IMO) decision of the library to promtail.

Anyway, the change proposed is quite simple and could be easily coded although I think it should not be done.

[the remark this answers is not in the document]
dannykopping: OK, i've taken a look at the [manpage](https://man7.org/linux/man-pages/man1/journalctl.1.html) to re-familiarise myself with the specifics of filtering.

Here's one example they cite:

>        To show all fields emitted by a unit and about the unit, option
>        -u/--unit= should be used.  journalctl -u name expands to a
>        complex filter similar to
> 
>            _SYSTEMD_UNIT=name.service
>              + UNIT=name.service _PID=1
>              + OBJECT_SYSTEMD_UNIT=name.service _UID=0
>              + COREDUMP_UNIT=name.service _UID=0 MESSAGE_ID=fc2e22bc6ee647b6b90729ab34a250b1

If a user were to try write a filter expression like the above in a promtail config, might it look like this?

```yaml
  ...
  matches:
    - _SYSTEMD_UNIT=name.service
    - UNIT=name.service _PID=1
    - OBJECT_SYSTEMD_UNIT=name.service _UID=0
    - COREDUMP_UNIT=name.service _UID=0 MESSAGE_ID=fc2e22bc6ee647b6b90729ab34a250b1
  ...
```

In the documentation we could specify that each `matches` item is a disjunction (logical `OR`), and the logical `AND` is combining multiple filters in a single entry separated by space.

It seems the library [supports adding disjunctions](https://github.com/coreos/go-systemd/blob/main/sdjournal/journal.go#L524), but I'm not familiar enough with the code to know how much effort this will be.

I understand that you've put in a lot of effort into this already, so I'm happy to merge this as-is, but in a perfect world I think we should follow the [Principle of Least Surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) here to make the filtering work as a user would expect. If journal filtering allows for disjunctions, then we should support it, too.

If you don't have the time to add disjunction support, that's ok - we can always add it as an enhancement for later; I don't want to block this useful feature.

dannykopping (commented):
@carlospeon feel free to engage me in our [public Slack](https://slack.grafana.com/) if you want to discuss this in a higher-bandwidth way.

I'm happy to proceed with this as-is, but I've shared my thoughts below.

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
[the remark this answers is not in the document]
carlospeon: > In the documentation we could specify that each matches item is a disjunction (logical OR), and the logical AND is combining multiple filters in a single entry separated by space.

 Absolutely. Logical AND have preference and are evaluated first, so IMO that's the way to express the filters. Or going further in the "yaml way" a two levels nested list:

```
  ...
  matches:
    - _SYSTEMD_UNIT=name.service
    - - UNIT=name.service
      - _PID=1
    - - OBJECT_SYSTEMD_UNIT=name.service
      - _UID=0
    - - COREDUMP_UNIT=name.service
      - _UID=0 
      - MESSAGE_ID=fc2e22bc6ee647b6b90729ab34a250b1
  ...
```
In the short time, in order to user current go-systemd functionality there should be only one list item:
```
  ...
  matches:
    - COREDUMP_UNIT=name.service _UID=0 MESSAGE_ID=fc2e22bc6ee647b6b90729ab34a250b1
  ...
```
or
```
  ...
  matches:
    - - COREDUMP_UNIT=name.service
      - _UID=0 
      - MESSAGE_ID=fc2e22bc6ee647b6b90729ab34a250b1
  ...

```
which is a bit weird and we must explain it in the docs. So we have a trade off between being ready to support future functionality and a weird (IMO) configuration or have a more simple configuration and no-compatibility with future functionality.

> It seems the library [supports adding disjunctions](https://github.com/coreos/go-systemd/blob/main/sdjournal/journal.go#L524), but I'm not familiar enough with the code to know how much effort this will be.

Yes, it is ready to use disjuntions, but the [reader](https://github.com/coreos/go-systemd/blob/main/sdjournal/read.go#L67) does not uses them at all. IMO the library should mimic [journalctl add_matches](https://github.com/systemd/systemd/blob/main/src/journal/journalctl.c#L1092). 

My main concern here is that proposing a modification in coreos/go-systemd reader to support disjunctions will need a different JournalReaderConfig struct, maybe replacing the match list with a two levels nested list of matches, and this will break compatibility of anyone using it, like promtail.

I will reach you in your public slack so we can talk about next steps.

clients/pkg/promtail/targets/journal/journaltarget_test.go
carlospeon: Sure, let me wait until we finally solve other conversations.

docs/sources/clients/promtail/scraping.md
carlospeon: I think we need to refrase it again... journald is a bit tricky, [sd_journal_add_match](https://github.com/systemd/systemd/blob/main/src/libsystemd/sd-journal/sd-journal.c#L290) change an AND for an OR when the filter is the same.

It's also documented in the man page:
> If multiple matches are specified
       matching different fields, the log entries are filtered by both,
       i.e. the resulting output will show only entries matching all the
       specified matches of this kind. If two matches apply to the same
       field, then they are automatically matched as alternatives, i.e.
       the resulting output will show entries matching any of the
       specified matches for the same field.

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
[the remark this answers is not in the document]
dannykopping: @carlospeon heads up, I'm on PTO from tomorrow and back on the 22nd Aug.
I'm happy for this to be merged as-is once all the comments are addressed (except for this one)

dannykopping (approved):
LGTM, thank you @carlospeon!
Please fix the conflicts and then we can merge this

clients/pkg/promtail/targets/journal/journaltarget_test.go
carlospeon: Done

kavirajk (approved):
LGTM :+1:

PR #6659, opened 2022-07-11
Running Loki as non-root in Kubernetes

Red-GV:
This PR defines the `USER` in the Dockerfile as `10001`. When running the Loki image as `nonroot` in Kubernetes, the resulting pod(s) will have a `CreateContainerConfigError`. This is because the `USER` in the image is a non-numerical value and thus it cannot resolve if the `USER` is a non-root user by default.

The work around for this in Kubernetes is to define the user to be some non-zero user. However, since the `USER` is already intended to be `10001`, the variable should be actually defined as this numeric value.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/cmd/loki/Dockerfile
+++ b/cmd/loki/Dockerfile
@@ -20,7 +20,7 @@ RUN mkdir -p /loki/rules && \
 # See https://github.com/grafana/loki/issues/1928
 RUN [ ! -e /etc/nsswitch.conf ] && echo 'hosts: files dns' > /etc/nsswitch.conf
 
-USER loki
+USER 10001
 EXPOSE 3100
 ENTRYPOINT [ "/usr/bin/loki" ]
 CMD ["-config.file=/etc/loki/local-config.yaml"]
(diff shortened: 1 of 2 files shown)


DylanGuedes:
Good catch, gj. Just in case, do you think this could be a breaking change for users relying on the former USER?

Red-GV:
Thanks @DylanGuedes. Hmm, that I'm not sure about. I don't think it should, but maybe there are some cases that I'm unaware about in which the change could be breaking.

DylanGuedes:
> Thanks @DylanGuedes. Hmm, that I'm not sure about. I don't think it should, but maybe there are some cases that I'm unaware about in which the change could be breaking.

Got it, thanks. I'll let this open for a few days for others to have a chance to have a look, if no one shows any objection, I'll proceed to merge it.

vlad-diachenko (approved):
as long as `loki` user id is `10001` I think that should not break anything. it's the same user, we just used the id instead of the name

PR #6662, opened 2022-07-12
[Jsonnet] Fix memberlist when using a stateful ruler

Whyeasy:
Signed-off-by: Whyeasy <Whyeasy@users.noreply.github.com>

When using stateful rulers and the memberlist as its ring, generating the resources with jsonnet throws in an error regarding the `ruler_deployment`. The `ruler_statefulset` didn't receive the meberlist labels either.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


Whyeasy:
It looks like there went something wrong with the jsonnet bundler in the pipeline and not the code contribution. Is this correct?

chaudum (approved):
lgtm, maybe another :eyes: from @cstyan ?

Whyeasy:
Still would like to have this fix in 😄 but waiting on another look at it by @cstyan

cstyan:
@Whyeasy  Can you show me the error you get? We're only using statefulsets internally and don't get a jsonnet error about the deployments. But you are right, the label for the statefulset was missing. Nothing would have broken, the label isn't actually used for any logic at the moment.

Whyeasy:
@cstyan, I get the following error:

```Error: got an error while extracting env `environments/dev`: recursion did not resolve in a valid Kubernetes object.  In path `.ruler_deployment.spec.template.metadata.labels` found key `loki_gossip_member` of type `string` instead.```

During reproducing the error I noticed the following change also fixed my issue:

```ruler_deployment+: if !$._config.memberlist_ring_enabled || !$._config.ruler_enabled || $._config.stateful_rulers then {} else gossipLabel,```

But I guess it's better to stay consistent here and at the `ruler_statefulset` as well 😄

cstyan:
Sorry, I missed your reply. I'll check this out more this week.

cstyan:
I'm not able to reproduce the error you get in our environments that only have stateful rulers deployed, but I don't see any issue with merging this since it's just checking the stateful ruler config value for both the deployment and statefulset.

PR #6664, opened 2022-07-12
[TSDB] dont skip writing to TSDB index when chunk exists in cache

owen-d:
Because TSDB indices are accumulated on disks and eventually flushed, we rely on their replication factor, so ensure we don't skip writing index entries when the chunk exists in cache already.


sandeepsukhani (approved):
LGTM

PR #6665, opened 2022-07-12
[TSDB] store uncompressed chunk size in index

owen-d:
Uncompressed size is more intuitive for both understanding throughput on streams via the index sampling endpoint and query planning -- this fixes an earlier oversight which used compressed size.


sandeepsukhani (approved):
LGTM

PR #6668, opened 2022-07-12
fix the issue of overwriting tsdb files during rollouts

sandeepsukhani:
During startup, we load all the index from the `wal` from previous session and build tsdb index out of it and upload the files with timestamp [now](https://github.com/grafana/loki/blob/f18edcf9d8199619a9b63d48a217948078ba6f24/pkg/storage/stores/tsdb/head_manager.go#L196). We also create a new `wal` with timestamp [now](https://github.com/grafana/loki/blob/f18edcf9d8199619a9b63d48a217948078ba6f24/pkg/storage/stores/tsdb/head_manager.go#L206). While rotating the `wal` we build and upload the index out of it with the same timestamp as the `wal` from which it was built [here](https://github.com/grafana/loki/blob/f18edcf9d8199619a9b63d48a217948078ba6f24/pkg/storage/stores/tsdb/head_manager.go#L288). This means the index built from `wal` created at startup would overwrite index build from previous session's `wal`. This PR fixes the issue by building index files with same timestamp as the `wal` from which it was built.

### DIFF
--- a/pkg/storage/stores/tsdb/manager.go
+++ b/pkg/storage/stores/tsdb/manager.go
@@ -168,91 +168,93 @@ func (m *tsdbManager) BuildFromWALs(t time.Time, ids []WALIdentifier) (err error
 	}()
 
 	level.Debug(m.log).Log("msg", "recovering tenant heads")
-	tmp := newTenantHeads(t, defaultHeadManagerStripeSize, m.metrics, m.log)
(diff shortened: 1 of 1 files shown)


owen-d (approved):
This also has the caveat of building a new TSDB for each WAL dir, but given this happens during normal operation anyway (new WAL dirs are created every WAL interval period), I think it's reasonable.

Thanks and nice find!

PR #6673, opened 2022-07-13
Introduce an "enable_groupcache" config to control initialization better

dannykopping:
We initialize groupcache as a module, which happens after the config is parsed. A function named applyFIFOCacheConfig is called to enable fifocache if no cache is configured, and at the time of its calling the modules have not been initialized. This new "enable_groupcache" value is set if groupcache is configured in common, or this value is set manually. With this setting, we can prevent automatic fifocache settings.

### DIFF
--- a/pkg/loki/config_wrapper.go
+++ b/pkg/loki/config_wrapper.go
@@ -85,6 +85,8 @@ func (c *ConfigWrapper) ApplyDynamicConfig() cfg.Source {
 
 		applyInstanceConfigs(r, &defaults)
 
+		applyCommonCacheConfigs(r, &defaults)
+
 		applyCommonReplicationFactor(r, &defaults)
 
 		applyDynamicRingConfigs(r, &defaults)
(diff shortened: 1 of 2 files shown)


pkg/loki/config_wrapper.go
```diff
@@ -166,6 +168,17 @@ func applyInstanceConfigs(r, defaults *ConfigWrapper) {
...
 }
 
+// applyCommonCacheConfigs applies to Loki components the cache-related configurations under the common config section
+// NOTE: only used for GroupCache at the moment
+// TODO: apply to other caches as well
+func applyCommonCacheConfigs(r, _ *ConfigWrapper) {
+	if r.Config.Common.GroupCacheConfig.Enabled {
+		r.Config.ChunkStoreConfig.ChunkCacheConfig.EnableGroupCache = true
+		r.Config.QueryRange.ResultsCacheConfig.CacheConfig.EnableGroupCache = true
+		r.Config.StorageConfig.IndexQueriesCacheConfig.EnableGroupCache = true
```
periklis, line 178: Why is the `write_dedup_cache` not considered here?

periklis: Aware of not using it in combination with boltdb-shipper, but we still support the other index backends from cortex times. OTH enabling this per default will break boltdb-shipper? :confused:

dannykopping: > OTH enabling this per default will break boltdb-shipper? :confused:

Expand on this please?

periklis: I assume that enabling the write-dedup-cache with boltdb-shipper is an inconsistent configuration mix that will Loki crashing or not?

dannykopping: AFAIK it's simply not used by boltdb-shipper
Will look into it :+1:

dannykopping: @periklis I'm going to ignore the write dedupe cache for now.
Groupcache is an optimization, and users can currently use fifocache if they need an in-process cache for index dedupes with non-boltdb setups.

I'm concerned that if we just add support groupcache for write dedupe here that we'll introduce some bugs because boltdb-shipper is our recommended storage configuration and we aren't spending much effort on the other approaches. I don't think that risk is worth it, given that a) there is an in-process cache option already and b) this is just an optimization.

kavirajk (approved):
LGTM. thanks @dannykopping

PR #6678, opened 2022-07-14
Add configurable groupcache capacity

dannykopping:
Adds a configuration option to set the capacity of each groupcache group.

N/A

We'll need to add the ability to configure each cache individually, because the chunk cache generally needs far more capacity than the other two (result, index query), but for now this is suitable.

I've also **decreased** the default capacity to 100MB per group.

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/storage/chunk/cache/groupcache_test.go
+++ b/pkg/storage/chunk/cache/groupcache_test.go
@@ -13,7 +13,7 @@ import (
 )
 
 func TestGroupCache(t *testing.T) {
-	gc, err := setupGroupCache(1)
+	gc, err := setupGroupCache()
 	require.Nil(t, err)
 
 	c := gc.NewGroup("test-group", "test")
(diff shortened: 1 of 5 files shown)


pkg/storage/chunk/cache/groupcache_test.go
```diff
@@ -47,12 +47,10 @@ func TestGroupCache(t *testing.T) {
 }
 
 func setupGroupCache() (*GroupCache, error) {
-	return NewGroupCache(
-		&mockRingManager{},
-		&server.Server{HTTP: mux.NewRouter()},
-		log.NewNopLogger(),
-		nil,
-	)
+	return NewGroupCache(&mockRingManager{}, GroupCacheConfig{
```
MichelHollands, line 50: nit: isn't it cleaner to keep the previous layout with an indented section for the config parameters?

MichelHollands (approved):
One nit but otherwise LGTM

kavirajk (approved):
LGTM!

PR #6681, opened 2022-07-14
Fix TIME_WAIT connection leak from queriers to compactor

MasslessParticle:
http://tleyden.github.io/blog/2016/11/21/tuning-the-go-http-client-library-for-load-testing/

This case is overwhelmingly likely to be from the connection pool as `DefaultMaxIdleConnsPerHost` is only 2. This PR also makes sure to drain the http response body, just in case.

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -1061,7 +1061,10 @@ func (t *Loki) deleteRequestsClient(clientType string, limits retention.Limits) 
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


dannykopping:
Nice! I remember debugging this exact issue with Peter from Mimir last year on the billing forwarder:
https://github.com/grafana/backend-enterprise/pull/1472/files#diff-0e057ea5ec711488359552e50ba11ed62dceb0884abfa6c3c8654a82e7246080R174

PS the backporting seems to have failed?

PR #6696, opened 2022-07-15
Fix per-tenant delete client

MasslessParticle:
- Don't make the request for a disabled tenant even if global deletion is allowed


vlad-diachenko:
I think, there might be an issue. if we have an override for the tenant without `allow_deletes` defined, for example just this: 
```yaml
 retention_period: '184d',
```
then, `ok` will be true here: 
```golang
userLimits, ok := allLimits[userID]; ok
```
then, `userLimits.CompactorDeletionEnabled` will be `false`. 
so, even if the feature is enabled globally, all tenants without explicitly defined `allow_deletes: true` won't be able to use the feature.

vlad-diachenko (approved):
LGTM

PR #6700, opened 2022-07-18
Add function to reset boltDBIndexClientWithShipper singleton

chaudum:

In a downstream project I have a test case that creates multiple stores
with different BoltDB shipper schema configurations. Since the client
for the store is a singleton and is re-used, there needs to be a
function to reset it as well.

**This function must only be used in tests.**

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


MichelHollands (approved):
LGTM

PR #6707, opened 2022-07-18
add jsonnet for configuring memberlist cluster label

sandeepsukhani:
Add changes from PR https://github.com/grafana/loki/pull/6660 which lets us configure memberlist to use cluster labels.

### DIFF
--- a/production/ksonnet/loki/memberlist.libsonnet
+++ b/production/ksonnet/loki/memberlist.libsonnet
@@ -17,11 +17,6 @@
     // but "primary" KV depends on value of multikv_primary.
     memberlist_ring_enabled: false,
 
-    // Configures the memberlist cluster label. When verification is enabled, a memberlist member rejects any packet or stream
-    // with a mismatching cluster label.
-    memberlist_cluster_label: '',
-    memberlist_cluster_label_verification_disabled: false,
-
     // Migrating from consul to memberlist is a multi-step process:
     // 1) Enable multikv_migration_enabled, with primary=consul, secondary=memberlist, and multikv_mirror_enabled=false, restart components.
     // 2) Set multikv_mirror_enabled=true. This doesn't require restart.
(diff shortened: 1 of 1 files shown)


production/ksonnet/loki/memberlist.libsonnet
```diff
@@ -23,6 +17,11 @@
     // but "primary" KV depends on value of multikv_primary.
     memberlist_ring_enabled: false,
 
+    // Configures the memberlist cluster label. When verification is enabled, a memberlist member rejects any packet or stream
+    // with a mismatching cluster label.
+    memberlist_cluster_label: '',
```
slim-bean, line 22: I wonder if we should have a default here? Do we know what the behavior is if the verification is enabled but the label is empty (which I believe is the case the way this PR is currently written)

Or maybe we should have this error if it's not specified??

owen-d: I think we can have the default use some reasonable label, like `"loki"` and set verification to false by default, no?

I'm also not sure why we have `$._config.memberlist_<field>` when we later add
```
        cluster_label: $._config.memberlist_cluster_label,
        cluster_label_verification_disabled: $._config.memberlist_cluster_label_verification_disabled,
```
        
        
  It should work, it's just a bit indirect. I'd prefer we set the field itself though.

sandeepsukhani: > I think we can have the default use some reasonable label, like "loki" and set verification to false by default, no?
We can't do that either because the default is doing the verification and even if we disable it here older instances would still do the verification causing the same problem.

> I'm also not sure why we have $._config.memberlist_<field> when we later add
I just copied over the jsonnet. But yeah, setting it directly would be better. I will change it.

owen-d (approved):

production/ksonnet/loki/memberlist.libsonnet
owen-d: Makes sense, thanks

Issue #6503, opened 2022-06-27
 tail channel closed, stopping tailer


yangvipguang:
**Describe the bug**
Promtail often fails to read logs. log error:
![image](https://user-images.githubusercontent.com/11416887/175890404-cee73ddf-ad1b-4d39-a56d-dfdfb8fe3ed4.png)
Restart promtail and return to normal
![image](https://user-images.githubusercontent.com/11416887/175890608-9c2573ab-12ee-437b-bfc0-75189259f146.png)

Version: 2.5.0
Platform: Centos7

dannykopping:
Hi @yangvipguang. I think this occurring when your file is rotated.

yangvipguang:
@dannykopping  yes， the node  log files  are rotated 。
![image](https://user-images.githubusercontent.com/11416887/176070905-18774884-009f-4a81-b0a1-e76584bb9841.png)
![lQLPJxZxUXJBXsLMh80CZLAluRzDOVBnNwK6ugzgALwA_612_135](https://user-images.githubusercontent.com/11416887/176071398-c93a7539-fd14-4533-8e36-2c5697138ba4.png)

dannykopping:
@yangvipguang are you actually missing logs when this occurs, or are you just asking about the error in the logs?

This looks like good behaviour because when a file it rotated it will be renamed (but still maintain the same `inode`) and then eventually deleted when the contents are compressed. I would expect these log lines to show up in those situations.

yangvipguang:
@dannykopping  yes  actually missing logs when the log is rolated

frittentheke:
It appears Promtail really only tracks the positions based on the filename https://github.com/grafana/loki/blob/fb9a1989d856c78c5aaa43c56ece6f215669e066/clients/pkg/promtail/positions/positions.go#L111
and only recognizes the rotation of a file if it's actively running and watching during the rotation:

This is how this is logged:
```
level=info ts=2022-07-21T21:06:59.436123136Z caller=tailer.go:143 component=tailer msg="tail routine: started" path=/tmp/logs/log1.log
ts=2022-07-21T21:06:59.436160443Z caller=log.go:168 level=info msg="Seeked /tmp/logs/log1.log - &{Offset:12 Whence:0}"
level=info ts=2022-07-21T21:07:29.263830553Z caller=filetargetmanager.go:179 msg="received file watcher event" name=/tmp/logs/log1.log.1 op=CREATE
[...]

PR #6708, opened 2022-07-18
Add compressed files support for Promtail

### LINKED ISSUE #5956: [Promtail] Support reading from compressed log files
Currenly Promtail can only read from files which are not compressed. At the same time applying compression is quite common for logging and holding a few days worth of logs on a machine.

1. While it's quite common to have uncompressed log files for actively written files and then compress them on rotation there are certain loggers or shippers such as _rsyslog_ that can even write a gzip stream right away doing inline compression. This is especially helpful on access logs or event logs which are high volume but also extremely compressible due to their very repetitive patterns or wasteful structure such as JSON. This approach also saves on disk IO as only a tiny fraction of the log is written to the disk.
2. Another aspect is the issue of being unable to read and backfill on somewhat older log data that has already been compressed on the source machine. This could be due to Promtail not running or Loki being unavailable while logroation and compression happens. While there is a feature on logrotate to delay the compression for one iteration (usually used to allow shippers to finish slurping the file), this requires much more disk space on the machine (see reasoning on huge access logs above).

Either an option within the scrape config to allow compressed files to be considered or just the ability for e.g. a gzipped file to be read transparently. It's the same data as an uncompressed file, all the other checks and limits just apply.

There really is no alternative. It's either natively supported by Promtail to read a log stream from compressed files or some manual action is required by operations to either decompress the files again or pipe them to Promtail somehow.

**Additional context**
 - rsyslog writing compressed logs -  https://www.rsyslog.com/doc/v8-stable/configuration/modules/omfile.html#ziplevel
 - syslog-ng piping logs to gzip - `destination d_gzip { program("gzip -c >> logfile.gz"); };`
 - Elastic Filebeat feature request about reading compressed files - https://github.com/elastic/beats/issues/637

 - Logrotate _delaycompress_ option - https://github.com/logrotate/logrotate/blob/master/ChangeLog.md#25---1997-09-01
 - Java _Log4j_ doing compression - https://logging.apache.org/log4j/2.x/manual/appenders.html
 - Golang _logrus_ - https://github.com/sirupsen/logrus/blob/b53d94c8ada4ef48988cb955795bf5c22dcf4b25/README.md#rotation
 - Python logrotation and compression - https://python.readthedocs.io/en/latest/howto/logging-cookbook.html#using-a-rotator-and-namer-to-customize-log-rotation-processing
 - ...

DylanGuedes:
Adds to Promtail the ability to read compressed files. It works by:
1. Infer which compression format to use based on the file extension
2. Uncompress the file with the native `golang/compress` packages
3. Iterate over uncompressed lines and send them to Loki

Its usage is the same as our current file tailing. In the example below all files under `logfiles` folder are parsed/processed, regardless of being compressed or not. `file1` and `file2` are first uncompressed with Gunzip; file3 is parsed as it is.
```yaml
scrape_configs:
  - job_name: simplejob
    static_configs:
      - targets: [localhost]
        labels:
          job: compressedfolder
          __path__: /logfiles/**.*
```
```bash
ls /logfiles
=> file1.tar.gz file2.tar.gz file3.txt
```

Fixes #5956 

Another approach is to infer the compression format based on the header bytes but this approach is much simpler; in the future we can improve it if we think the community needs it.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,285 @@
...
+	go decompresser.updatePosition()
+	metrics.filesActive.Add(1.)
+	return decompresser, nil
+}
+
+// mountReader instantiate a reader ready to be used by the decompresser.
+//
+// The selected reader implementation is based on the extension of the given file name.
+// It'll error if the extension isn't supported.
+func mountReader(f *os.File, logger log.Logger) (reader io.Reader, err error) {
```
dannykopping, line 97: I think `.tar` should be supported here as well - it's a very common format
Speaking of which, if a file is a compressed tarball (.tar,gz) that might need some special handling

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,285 @@
...
+		reader, err = zlib.NewReader(f)
+	} else if ext == ".zip" {
+		decompressLib = "compress/flate"
+		reader = flate.NewReader(f)
+	} else if ext == ".bz2" {
+		decompressLib = "bzip2"
+		reader = bzip2.NewReader(f)
+	}
+
+	level.Info(logger).Log("msg", fmt.Sprintf("using %q to decompress file %q", decompressLib, f.Name()))
```
dannykopping, line 115:
```suggestion
	level.Debug(logger).Log("msg", fmt.Sprintf("using %q to decompress file %q", decompressLib, f.Name()))
```

dannykopping: Things should _just work_ here so this log line is probably only useful in debug mode

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,285 @@
...
+
+	if reader != nil {
+		return reader, nil
+	}
+
+	if err != nil && err != io.EOF {
+		return nil, err
+	}
+
+	return nil, fmt.Errorf("file %q with unsupported extension", f.Name())
```
dannykopping, line 125:
```suggestion
	return nil, fmt.Errorf("file %q has unsupported extension", f.Name())
```

Might be nice to include which extensions are actually supported in this message

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,285 @@
...
+			return
+		}
+	}
+}
+
+// readLines read all existing lines of the given compressed file.
+//
+// It first decompress the file as a whole using a reader and then it will iterate
+// over its chunks, separated by '\n'.
+// During each iteration, the parsed and decoded log line is then sent to the API with the current timestamp.
```
dannykopping, line 154: This is what I was kinda worried about seeing...
If the file is compressed, it could potentially store a huge amount of data which we need to entirely load into memory to iterate over.

Is there a more resource-efficient way we could do this?

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,285 @@
...
+	level.Info(t.logger).Log("msg", "successfully decompressed file", "path", t.path)
+
+	scanner := bufio.NewScanner(r)
+	for i := 0; ; i++ {
+		if !scanner.Scan() {
+			break
+		}
+
+		if scanner.Err() != nil {
+			break
```
dannykopping, line 188: Is this error worth logging?
I guess if it's `EOF` then it's fine but other errors probably not?

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,285 @@
...
+				Line:      finalText,
+			},
+		}
+
+		t.size = int64(unsafe.Sizeof(finalText))
+		t.position++
+
+		if err != nil {
+			break
+		}
```
dannykopping, line 225: Where will this `err` be set?
Line `199` seems unnecessary?

clients/pkg/promtail/targets/file/filetarget.go
```diff
@@ -319,24 +319,56 @@ func (t *FileTarget) startTailing(ps []string) {
...
+		".tar.gz",
+		".z",
+		".bz2",
+	}
+
+	for _, format := range supportedCompressedFormats {
+		if ext == format {
+			return true
+		}
+	}
```
dannykopping, line 359: Micro-optimization: make `supportedCompressedFormats` a map to save a few unnecessary cycles here.

dannykopping (commented):
Overall looks awesome @DylanGuedes!
I'm a bit concerned by the potential here for resource overuse by reading the whole decompressed file into memory, from what I can see.

clients/pkg/promtail/targets/file/decompresser.go
frittentheke: But tar is a container or archive format, but not a compression format. 
The feature of being able to read from compressed files should all about compressed data streams, not handling archive files with more files in them.

clients/pkg/promtail/targets/file/decompresser.go
frittentheke: Totally @dannykopping  ... compression ratios >99% are common when thinking about uniformly structured logs like access or event logs. Reading a whole file into memory would totally not be an option. When suggesting implementing this along the lines of piping data via `zcat` into an existing tool (https://github.com/grafana/loki/issues/5956#issuecomment-1185366990). And isn't that exactly what the mountReader method does? It gets the right type of reader (which is able to read a compressed data stream of various algorithms) and then hands that off to be read from line by line. But there is not slurping into memory, is there?

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	defer f.Close()
+
+	r, err := mountReader(f, t.logger)
+	if err != nil {
+		level.Error(t.logger).Log("msg", "error mounting new reader", "err", err)
+		return
+	}
+
+	level.Info(t.logger).Log("msg", "successfully decompressed file", "path", t.path)
```
frittentheke, line 179: The file was not successfully decompressed at this point. You rather "mounted" a reader onto the file handle.

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,285 @@
...
+	"golang.org/x/text/encoding/ianaindex"
+	"golang.org/x/text/transform"
+
+	"github.com/grafana/loki/pkg/logproto"
+
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+	"github.com/grafana/loki/clients/pkg/promtail/positions"
+)
+
+type decompresser struct {
```
frittentheke, line 33: Maybe rename to _decompressor_ to align this with various Golang libs: https://pkg.go.dev/search?q=decompressor&m=

clients/pkg/promtail/targets/file/decompresser.go
dannykopping: Oh, it seems `.tar.gz` is supported:

```go
if strings.Contains(ext, "gz") { // .gz, .tar.gz
```

Sorry @DylanGuedes, I missed that

DylanGuedes: np and no reason to apologize :smile:

DylanGuedes:
@dannykopping @frittentheke thank you both for the fantastic reviews, I'll be working on those suggestions and other ones that might come along the week and I plan to try testing a few things with this too.

Just one thing, I think even if this doesn't add decent support for compressed log rotations, we should still proceed to at least have this, as a simpler implementation that is to be improved in a follow-up PR (I can start working on it exactly after this gets merged). My argument is that as of now Loki isn't supporting even the most straightforward compression scenarios, so this would at least add some support without a lot of burdens.
For instance, I liked that with this PR you can now tell Promtail to scrape from a folder that has tons of compressed files and that you can keep appending new compressed files to it and all of them will be scraped.

frittentheke:
> For instance, I liked that with this PR you can now tell Promtail to scrape from a folder that has tons of compressed files and that you can keep appending new compressed files to it and all of them will be scraped.

Yes. Question is though, in what order those files will be picked up? Does Promtail sort them by modification date? That would then at least keep as much order as possible.

clients/pkg/promtail/targets/file/decompresser.go
DylanGuedes: Thanks for spotting this, addressed by https://github.com/grafana/loki/pull/6708/commits/58b5634d729b311bdbba63c7000320fcaf3ce272. I also added the extension to the log message.

clients/pkg/promtail/targets/file/decompresser.go
DylanGuedes: Agreed. I included them on https://github.com/grafana/loki/pull/6708/commits/e80a2f87d6da59469d0b84c1588b0d1d7ca685d4.

clients/pkg/promtail/targets/file/decompresser.go
DylanGuedes: actually, I overlooked it, good catch. Yes, in most cases it will be EOF so it shouldn't be logged but in some cases, it might be an actual error. Fixing it...

DylanGuedes: Thanks, addressed by https://github.com/grafana/loki/pull/6708/commits/1168d7a6087fb23f7cfbee08b000138f3f34bbc6.

clients/pkg/promtail/targets/file/decompresser.go
DylanGuedes: Oops, I think I had a brain fog :rofl: 
Fixed by https://github.com/grafana/loki/pull/6708/commits/459d1cc04811058b3cdb47c56af1b068c9040308

clients/pkg/promtail/targets/file/filetarget.go
DylanGuedes: Done by https://github.com/grafana/loki/pull/6708/commits/4bedfb27850286122f667e0fbb1a480aee627223. However, I added it as a function instead, since Go doesn't support constant maps. I'm also reusing it to list the supported extensions.

DylanGuedes:
> Yes. Question is though, in what order those files will be picked up? Does Promtail sort them by modification date? That would then at least keep as much order as possible.

Good call, thanks. I'll add this scenario to the list of things I'll be testing this week.
Of course, even if I implement some kind of sorting there will be still edge cases because the managers will be checking things async but it would for sure be more stable than not having any sorting (which I'm not sure we have as of now).

DylanGuedes:
FYI: I did generate a file with ~400 MB, and compressed it with gunzip which reduced it to ~1 MB, and apparently the gunzip reader didn't really expand the whole file before iterating over it, which is great. I'll try one more time but this time with a bigger file (a few gbs).

DylanGuedes:
> > Yes. Question is though, in what order those files will be picked up? Does Promtail sort them by modification date? That would then at least keep as much order as possible.
> 
> Good call, thanks. I'll add this scenario to the list of things I'll be testing this week. Of course, even if I implement some kind of sorting there will be still edge cases because the managers will be checking things async but it would for sure be more stable than not having any sorting (which I'm not sure we have as of now).

I've tested/investigate it and no, **Promtail doesn't follow any order**. I tried to look for a quick solution to it but tbh it won't be that easy because:
- The discovery is based on the filename/glob
- Once the files to be tailed are discovered, the target manager iterates over them and sends them to a channel
- They're processed as they're received on the channel

We can try sorting them by name before publishing them on the channel, but ordering them by creation date looks to be a challenge.

DylanGuedes:
@dannykopping fyi: I finished the tests I was interested, anything else you'd like me to test/improve?

frittentheke:
> We can try sorting them by name before publishing them on the channel, but ordering them by creation date looks to be a challenge.
> 
> EDIT: nvm, I can actually read the creation/modification date from the name and sort things before hand
> 
> EDIT2: even after sorting the targets, they might be synced out of order because tailing is completely async. I implemented a solution and wrote a few unit tests but it will fail/is failing because tailing itself is async, so even if I start tailing on the right order, they might still process things in a different order.

This seems to be a larger issue of only using file names and not the abstract inodes number and the modification date.
See me comment about this at https://github.com/grafana/loki/issues/6503#issuecomment-1191957798

So this might be a fix / improvement for another day. But since Loki does not really loves older timestamps being ingested Promtail should do what is can to read from files targeting the same labels in proper order ...

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -176,15 +190,19 @@ func (t *decompresser) readLines() {
...
 	scanner := bufio.NewScanner(r)
 	for i := 0; ; i++ {
 		if !scanner.Scan() {
 			break
 		}
 
-		if scanner.Err() != nil {
+		if scannerErr := scanner.Err(); scannerErr != nil {
+			if scannerErr != io.EOF {
+				level.Error(t.logger).Log("msg", "error scanning", "err", scannerErr)
```
dannykopping, line 203: Will this message give us an indication of which target has a problem?

dannykopping:
> FYI: I did generate a file with ~400 MB, and compressed it with gunzip which reduced it to ~1 MB, and apparently the gunzip reader didn't really expand the whole file before iterating over it, which is great. I'll try one more time but this time with a bigger file (a few gbs).

Can we add a benchmark for these so the tests are repeatable, and we can see what the CPU/memory impact is?

clients/pkg/promtail/targets/file/decompresser.go
DylanGuedes: hmmm you mean literally the target (which is the file)? If so, kinda; the component (which is the decompresser) is injected into the logger and by it you can infer that it is the file target.

dannykopping: OK perfect, thanks

DylanGuedes:
> > FYI: I did generate a file with ~400 MB, and compressed it with gunzip which reduced it to ~1 MB, and apparently the gunzip reader didn't really expand the whole file before iterating over it, which is great. I'll try one more time but this time with a bigger file (a few gbs).
> 
> Can we add a benchmark for these so the tests are repeatable, and we can see what the CPU/memory impact is?

I'm looking at it but it is more complicated than I was thinking. Out of curiosity, would you say I should generate a big file, compress it and bench everything programmatically, or were you thinking something different?

dannykopping:
> > > FYI: I did generate a file with ~400 MB, and compressed it with gunzip which reduced it to ~1 MB, and apparently the gunzip reader didn't really expand the whole file before iterating over it, which is great. I'll try one more time but this time with a bigger file (a few gbs).
> > 
> > 
> > Can we add a benchmark for these so the tests are repeatable, and we can see what the CPU/memory impact is?
> 
> I'm looking at it but it is more complicated than I was thinking. Out of curiosity, would you say I should generate a big file, compress it and bench everything programmatically, or were you thinking something different?

> Out of curiosity, would you say I should generate a big file, compress it and bench everything programmatically

Yeah this is what I'd love to see.

As an aside, I went looking in the Go source, and I was surprised to not find benchmarks in the [GZIP pkg](https://cs.opensource.google/go/go/+/master:src/compress/gzip/gunzip_test.go), although there are a couple:

```bash
compress [tags/go1.18.5] $ grep -rn Benchmark
flate/writer_test.go:16:func BenchmarkEncode(b *testing.B) {
flate/reader_test.go:34:func BenchmarkDecode(b *testing.B) {
lzw/writer_test.go:199:func BenchmarkEncoder(b *testing.B) {
lzw/reader_test.go:267:func BenchmarkDecoder(b *testing.B) {
bzip2/bzip2_test.go:238:func BenchmarkDecodeDigits(b *testing.B) { benchmarkDecode(b, digits) }
bzip2/bzip2_test.go:239:func BenchmarkDecodeNewton(b *testing.B) { benchmarkDecode(b, newton) }
bzip2/bzip2_test.go:240:func BenchmarkDecodeRand(b *testing.B)   { benchmarkDecode(b, random) }
```

What I was thinking we could do is:

- add a text file to the test fixtures which is sufficiently big but highly compressible (a couple MB of [server logs](https://www.kaggle.com/datasets/eliasdabbas/web-server-access-logs)?)
- in the benchmark, compress it and save to a file
- start the benchmark timer and begin uncompressing the file
- take note of allocs, MB/s, etc

We may see that one or two compression formats cannot be partially uncompressed and lead to unbounded memory growth, in which case we'll need to find an alternative to the stdlib packages, or put a big warning in the docs for that format.

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,299 @@
...
+
+	r, err := mountReader(f, t.logger)
+	if err != nil {
+		level.Error(t.logger).Log("msg", "error mounting new reader", "err", err)
+		return
+	}
+
+	level.Info(t.logger).Log("msg", "successfully mounted reader", "path", t.path, "ext", filepath.Ext(t.path))
+
+	scanner := bufio.NewScanner(r)
```
jeschkies, line 195: I think the doc string that Danny commented is wrong. Scanner will not decompress the whole file and put it into memory.

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,299 @@
...
+	for i := 0; ; i++ {
```
jeschkies, line 196:
```suggestion
	for line := 0; ; line++ {
```

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	supportedExtsList := strings.Builder{}
+	for ext := range supportedCompressedFormats() {
+		supportedExtsList.WriteString(ext)
+	}
+	return nil, fmt.Errorf("file %q has unsupported extension, it has to be one of %q", f.Name(), supportedExtsList.String())
+}
+
+func (t *decompressor) updatePosition() {
```
jeschkies, line 142: What is the motivation behind this? It's been a while but updating the position didn't have to be async.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * [6358](https://github.com/grafana/loki/pull/6358) **taharah**: Fixes sigv4 authentication for the Ruler's remote write configuration by allowing both a global and per tenant configuration.
 * [6375](https://github.com/grafana/loki/pull/6375) **dannykopping**: Fix bug that prevented users from using the `json` parser after a `line_format` pipeline stage.
 ##### Changes
+* [6415](https://github.com/grafana/loki/pull/6415) **salvacorts**: Evenly spread queriers across kubernetes nodes.
```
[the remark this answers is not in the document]
DylanGuedes: Oops, thanks for catching it. I think those two changelog entries had something wrong but maybe when merging with the main branch something went off. Fixing it :sweat_smile:

clients/pkg/promtail/targets/file/decompresser.go
DylanGuedes: hmmm not sure, but I can confirm I had issues without this when using multiple clients. The other targets are also updating the positions async.

clients/pkg/promtail/targets/file/decompresser.go
DylanGuedes: agreed.

clients/pkg/promtail/targets/file/decompresser.go
DylanGuedes: Done!

clients/pkg/promtail/targets/file/decompresser.go
dannykopping: We tested this together, and found that it is indeed allocating a lot of memory **but** the GC acts as it should, and the memory doesn't grow unbounded. Dylan also verified that the files are decompressed progressively, so at no time do we try to load the whole file into memory at once, so the overall memory usage will not spike very high (we use small buffers when reading the compressed files). We should add a note in the docs to specify that GC frequency will increase dramatically when decompressing large files, and this will cause CPU usage to rise.

clients/pkg/promtail/targets/file/decompresser.go
dannykopping: Confirmed :+1:

clients/pkg/promtail/targets/file/decompresser.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil {
+		level.Error(t.logger).Log("msg", "error mounting new reader", "err", err)
+		return
+	}
+
+	level.Info(t.logger).Log("msg", "successfully mounted reader", "path", t.path, "ext", filepath.Ext(t.path))
+
+	megabyte := 1000000
+	maxLoglineSize := 2 * megabyte
+	buffer := make([]byte, 3*megabyte)
```
dannykopping, line 192: Comments to explain rationale please

dannykopping: I think a 4096 byte buffer would be more performant here; let's start with that, and in a subsequent PR we can try improve performance or make this configurable.

docs/sources/clients/promtail/_index.md
```diff
@@ -36,6 +36,37 @@ Just like Prometheus, `promtail` is configured using a `scrape_configs` stanza.
 drop, and the final metadata to attach to the log line. Refer to the docs for
 [configuring Promtail](configuration/) for more details.
 
+### Support for compressed files
+
+Promtail has native support for ingesting compressed files by a mechanism that
+relies on file extensions. If a discovered file has a expected compression file
```
dannykopping, line 42:
```suggestion
relies on file extensions. If a discovered file has an expected compression file
```

docs/sources/clients/promtail/_index.md
```diff
@@ -36,6 +36,37 @@ Just like Prometheus, `promtail` is configured using a `scrape_configs` stanza.
 drop, and the final metadata to attach to the log line. Refer to the docs for
 [configuring Promtail](configuration/) for more details.
 
+### Support for compressed files
+
+Promtail has native support for ingesting compressed files by a mechanism that
```
dannykopping, line 41:
```suggestion
Promtail now has native support for ingesting compressed files by a mechanism that
```

docs/sources/clients/promtail/_index.md
```diff
@@ -36,6 +36,37 @@ Just like Prometheus, `promtail` is configured using a `scrape_configs` stanza.
...
+  - `.gz`: Data will be decompressed with the native Gunzip Golang pkg (`pkg/compress/gzip`)
+  - `.z`: Data will be decompressed with the native Zlib Golang pkg (`pkg/compress/zlib`)
+  - `.bz2`: Data will be decompressed with the native Bzip2 Golang pkg (`pkg/compress/bzip2`)
+  - `.tar.gz`: Data will be decompressed exactly as the `.gz` extension.
+      However, because `tar` will add its metadata at the beggining of the
+      compressed file, **the first parsed line will contains metadata together with
+      your log line**. It is illustrated at
+      `./clients/pkg/promtail/targets/file/decompresser_test.go`.
+* Positions are supported. That means that, if you interrupt Promtail after
+  parsing and pushing 45% of your compressed file data, you can expect Promtail
```
dannykopping, line 57:
```suggestion
  parsing and pushing (for example) 45% of your compressed file data, you can expect Promtail
```

docs/sources/clients/promtail/_index.md
```diff
@@ -36,6 +36,37 @@ Just like Prometheus, `promtail` is configured using a `scrape_configs` stanza.
...
+  to resume work from the 55% rest.
```
dannykopping, line 58:
```suggestion
  to resume work from the last scraped line and process the rest of the remaining 55%.
```

docs/sources/clients/promtail/_index.md
```diff
@@ -36,6 +36,37 @@ Just like Prometheus, `promtail` is configured using a `scrape_configs` stanza.
...
+* Since decompression and pushing can be very performatic, depending on the size
```
dannykopping, line 59:
```suggestion
* Since decompression and pushing can be very fast, depending on the size
```

docs/sources/clients/promtail/_index.md
```diff
@@ -36,6 +36,37 @@ Just like Prometheus, `promtail` is configured using a `scrape_configs` stanza.
...
+  of your compressed file Loki will rate-limit your ingestion. In that case you
+  might configure Promtail's `limits_config` to slow the pace or increase
```
dannykopping, line 61:
```suggestion
  might configure Promtail's [`limits` stage](https://grafana.com/docs/loki/latest/clients/promtail/stages/limit/) to slow the pace or increase
```

docs/sources/clients/promtail/_index.md
```diff
@@ -36,6 +36,37 @@ Just like Prometheus, `promtail` is configured using a `scrape_configs` stanza.
...
+  ingestion limits on Loki.
```
dannykopping, line 62: Include a link pls

docs/sources/clients/promtail/_index.md
```diff
@@ -36,6 +36,37 @@ Just like Prometheus, `promtail` is configured using a `scrape_configs` stanza.
...
+* Log rotations **aren't supported as of now**, mostly because it requires substantial
+changes on Promtail. If you'd like to see support for it, please create a new
+issue on Github asking for it and explaining your use case.
```
dannykopping, line 65: Why not?

docs/sources/clients/promtail/_index.md
```diff
@@ -36,6 +36,37 @@ Just like Prometheus, `promtail` is configured using a `scrape_configs` stanza.
 drop, and the final metadata to attach to the log line. Refer to the docs for
 [configuring Promtail](configuration/) for more details.
 
+### Support for compressed files
+
+Promtail has native support for ingesting compressed files by a mechanism that
+relies on file extensions. If a discovered file has a expected compression file
+extension, Promtail will lazily decompress the compressed file and push the
```
dannykopping, line 43: Let's add another point around memory consumption & GC / CPU impact

clients/pkg/promtail/targets/file/decompresser_test.go
```diff
@@ -0,0 +1,175 @@
...
+	d.readLines()
+
+	<-d.done
+	time.Sleep(time.Millisecond * 200)
+
+	entries := handler.Received()
+	require.Equal(t, 100000, len(entries))
+}
+
+func TestOnelineFiles(t *testing.T) {
```
dannykopping, line 102: Please add a docstring explaining the value of these tests

dannykopping (approved):
LGTM, some small doc changes and I think we should change that buffer size - but after that, merge at will!

DylanGuedes:
Benchmark results with 3MB buffer size for decompression:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -tags integration,requires_docker -bench ^BenchmarkReadlines$ github.com/grafana/loki/clients/pkg/promtail/targets/file

goos: linux
goarch: amd64
pkg: github.com/grafana/loki/clients/pkg/promtail/targets/file
BenchmarkReadlines/2000_lines_of_log_.tar.gz_compressed-12         	     768	   1559020 ns/op	 3063606 B/op	     239 allocs/op
BenchmarkReadlines/100000_lines_of_log_.gz_compressed-12           	      18	  63992058 ns/op	 5666200 B/op	   21157 allocs/op
PASS
ok  	github.com/grafana/loki/clients/pkg/promtail/targets/file	3.221s
```

DylanGuedes:
Benchmark results with 4096 buffer size:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -tags integration,requires_docker -bench ^BenchmarkReadlines$ github.com/grafana/loki/clients/pkg/promtail/targets/file

goos: linux
goarch: amd64
pkg: github.com/grafana/loki/clients/pkg/promtail/targets/file
BenchmarkReadlines/2000_lines_of_log_.tar.gz_compressed-12         	     933	   1261038 ns/op	   60287 B/op	     236 allocs/op
BenchmarkReadlines/100000_lines_of_log_.gz_compressed-12           	      19	  58614711 ns/op	 2536763 B/op	   20279 allocs/op
PASS
ok  	github.com/grafana/loki/clients/pkg/promtail/targets/file	3.167s

```

DylanGuedes:
@dannykopping Merging as I addressed all your great suggestions. I'll be working on the zip compression on a follow-up PR :smile:

PR #6714, opened 2022-07-19
Revert "Configure ingester PDB to have max_unavailable=0"

sandeepsukhani:
Reverts grafana/loki#6589

As discussed internally, setting `max_unavailable` to `0` would block the nodes from draining. Investigating the root cause behind what caused us to change this value, it looks like the cause was not the rollout or pod disruption; it was a panic in ingester causing multiple ingesters to go down.


owen-d (approved):

vlad-diachenko (approved):
LGTM

PR #6718, opened 2022-07-19
Fix goroutine leak when using http2 with groupcache

MasslessParticle:
If the http2 transport isn't reused, we create a connection on every request that is never terminated. This change just reuses one transport for all groupcache connections


dannykopping (approved):
Thanks @MasslessParticle!

PR #6720, opened 2022-07-20
[Enhancement] Removes memcached failure log

owen-d:
This removes the old memcached warning log when failing to put to memcache. This was confusing for many (looks like a problem when a certain failure rate to the chunk cache is expected). However, I'm also removing this because it happens so often that it shows up as one of the largest querier mutex blocks during profiling, so this brings a performance boost as well.


dannykopping (commented):
Oh hell yes :heart: 
Do we have a metric to track this, though? A metric would be much more resource efficient and we'd still be able to keep an eye on this problem if we ever needed to.

sandeepsukhani (approved):

PR #6722, opened 2022-07-20
[Memcache] use owen-d gomemcache fork for connection sharding

owen-d:
While profiling TSDB which uses many smaller queries, I've noticed that as much as 45% of our mutex lock contention delay is caused by the memcached client, particularly as we push higher query throughputs (more accesses+cache nodes). As an experiment, I forked gomemcached and added a new commit(https://github.com/owen-d/gomemcache/commit/3b901ce25fe5001e8ebe08b519aee72df46d8176) and striped lock accesses, which improved the problem, reducing acquisition delay by ~90% if I recall correctly (I forgot to save the actual profiles to include in this PR).

This PR moves the gomemcached fork we use to my own.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -310,7 +310,7 @@ replace github.com/gocql/gocql => github.com/grafana/gocql v0.0.0-20200605141915
 
 // Same as Cortex
 // Using a 3rd-party branch for custom dialer - see https://github.com/bradfitz/gomemcache/pull/86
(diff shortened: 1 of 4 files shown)


dannykopping (commented):
Nice find!!

I noticed you forked from `bradfitz/gomemcache` not `themihai/gomemcache`. Why is that?

sandeepsukhani (approved):
Nice one! LGTM!

PR #6724, opened 2022-07-20
tsdb store refactoring

sandeepsukhani:
It includes two small changes:
1. We made tsdb store instance public to only allow creating a single instance of it. I have refactored the code not to make it public.
2. Build a tsdb from active wal and upload it to the object storage during ingester shutdown. This is to avoid losing the index if an ingester is being shut down forever.

### DIFF
--- a/pkg/storage/stores/tsdb/store.go
+++ b/pkg/storage/stores/tsdb/store.go
@@ -17,32 +17,59 @@ import (
 	util_log "github.com/grafana/loki/pkg/util/log"
 )
 
-// we do not need to build store for each schema config since we do not do any schema specific handling yet.
-// If we do need to do schema specific handling, it would be a good idea to abstract away the handling since
-// running multiple head managers would be complicated and wasteful.
-var storeInstance *store
-
 type store struct {
 	indexWriter IndexWriter
 	indexStore  series.IndexStore
 }
 
+type newStoreFactoryFunc func(
+	indexShipperCfg indexshipper.Config,
+	p config.PeriodConfig,
+	f *fetcher.Fetcher,
+	objectClient client.ObjectClient,
+	limits downloads.Limits,
+	tableRanges config.TableRanges,
+	reg prometheus.Registerer,
+) (
+	stores.ChunkWriter,
+	series.IndexStore,
+	error,
+)
+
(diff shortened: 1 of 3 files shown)


pkg/storage/stores/tsdb/store.go
```diff
@@ -17,37 +19,69 @@ import (
...
 type store struct {
-	indexWriter IndexWriter
-	indexStore  series.IndexStore
+	indexShipper indexshipper.IndexShipper
+	indexWriter  IndexWriter
+	indexStore   series.IndexStore
+	stopOnce     sync.Once
 }
 
+type newStoreFactoryFunc func(
```
jeschkies, line 29: Is this used anywhere else than for `NewStore`?

pkg/storage/stores/tsdb/store.go
```diff
@@ -17,37 +19,69 @@ import (
...
+			err := storeInstance.init(indexShipperCfg, objectClient, limits, tableRanges, reg)
+			if err != nil {
+				return nil, nil, nil, err
+			}
 		}
-	}
 
-	return NewChunkWriter(f, p, storeInstance.indexWriter), storeInstance.indexStore, nil
-}
+		return NewChunkWriter(f, p, storeInstance.indexWriter), storeInstance.indexStore, storeInstance.Stop, nil
```
jeschkies, line 76: Why not return the `storeInstance`?

pkg/storage/stores/tsdb/store.go
sandeepsukhani: No, its not. It is just for defining function signature.

pkg/storage/stores/tsdb/store.go
sandeepsukhani: For which type? We can't do it for writer atleast as explained here https://github.com/grafana/loki/blob/d6f0455b656d9ad2c99aabcda5ec2dbbbd6135a8/pkg/storage/stores/tsdb/store.go#L45

owen-d: @jeschkies Yeah, we should refactor to make this more intuitive/less ugly.

PR #6727, opened 2022-07-20
Fix log messages for deletion requests

MichelHollands:

Too many log lines were shown when delete requests are processed. In fact it was showing 2 log lines per chunk. This PR removes those. It also adds log lines for when the delete request is accepted and when it starts being processing.

The log lines look like this:

```
level=info ts=2022-07-25T12:11:49.2284138Z caller=request_handler.go:106 msg="delete request for user added" user=1 query="{foo=\"bar\"} |= \"Baz was here\""
...
level=info ts=2022-07-25T12:23:46.6920307Z caller=delete_requests_manager.go:121 msg="Started processing delete request for user" delete_request_id=3144c19a user=1
...
level=info ts=2022-07-25T12:23:46.6961391Z caller=delete_requests_manager.go:218 msg="delete request for user marked as processed" delete_request_id=3144c19a user=1 deleted_lines=0
```

**Checklist**
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/storage/stores/indexshipper/compactor/deletion/metrics.go
+++ b/pkg/storage/stores/indexshipper/compactor/deletion/metrics.go
@@ -50,6 +50,7 @@ type deleteRequestsManagerMetrics struct {
 	loadPendingRequestsAttemptsTotal     *prometheus.CounterVec
 	oldestPendingDeleteRequestAgeSeconds prometheus.Gauge
 	pendingDeleteRequestsCount           prometheus.Gauge
+	deletedLinesTotal                    *prometheus.CounterVec
 }
 
 func newDeleteRequestsManagerMetrics(r prometheus.Registerer) *deleteRequestsManagerMetrics {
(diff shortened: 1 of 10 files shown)


pkg/storage/stores/indexshipper/compactor/deletion/metrics.go
```diff
@@ -81,11 +80,6 @@ func newDeleteRequestsManagerMetrics(r prometheus.Registerer) *deleteRequestsMan
 		Name:      "compactor_pending_delete_requests_count",
 		Help:      "Count of delete requests which are over their cancellation period and have not finished processing yet",
 	})
-	m.deletedLinesTotal = promauto.With(r).NewCounterVec(prometheus.CounterOpts{
```
dannykopping, line 84: You're removing a metric here which I suppose some folks might be using already? Best to include this in the upgrade guide / changelog.

dannykopping: Also, what did you mean by "The metric wasn't always showing the latest value."? Seems like this metric would be pretty useful...

MichelHollands: Good point. I've left the metric. Instead of creating a new metric it references the main metrics struct now.

dannykopping: Thanks @MichelHollands :+1: 
Does that address the "The metric wasn't always showing the latest value" issue mentioned above?

MichelHollands: Yes, it should. I've added additional checks in the unit tests for this as well now.

dannykopping: Magic, thanks

dannykopping (approved):
LGTM

PR #6728, opened 2022-07-20
Remove http2 support for groupcache

MasslessParticle:
http2 stream multiplexing means that requests are handled concurrently and, under high load, can cause the number of goroutines handling requests to grow without bound. The ablity to limit the number of goroutines serving requests is not currently implemented: https://pkg.go.dev/golang.org/x/net@v0.0.0-20220708220712-1185a9018129/http2#Server

```
// MaxHandlers limits the number of http.Handler ServeHTTP goroutines
// which may run at a time over all connections.
// Negative or zero no limit.
// TODO: implement
MaxHandlers [int](https://pkg.go.dev/builtin#int)
```


pkg/storage/chunk/cache/groupcache.go
```diff
@@ -86,11 +75,13 @@ func NewGroupCache(rm ringManager, config GroupCacheConfig, server *server.Serve
 	addr := fmt.Sprintf("http://%s", rm.Addr())
 	level.Info(logger).Log("msg", "groupcache local address set to", "addr", addr)
 
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.MaxIdleConnsPerHost = 100
```
slim-bean, line 79: Not sure if we also need to set any timeouts? I can't remember if the default transport has reasonable timeouts

slim-bean (approved):
One question but otherwise LGTM

kavirajk (approved):
Nice find :+1: 

LGTM. Also agree with Ed's suggestion on setting timeout for transport.

pkg/loki/modules.go
```diff
@@ -140,18 +137,9 @@ func (t *Loki) initServer() (services.Service, error) {
 		})
 	}(t.Server.HTTPServer.Handler)
 
-	// Allow us to receive http/2 cleartext in addition to http/1.1. This needs to be
-	// the outermost handler so it can upgrade requests if necessary
-	t.Server.HTTPServer.Handler = http2CleartextHandler(t.Server.HTTPServer.Handler)
-
```
periklis, line 146: Isn't this piece remove http2 from all our handlers? Do we want this (cc @slim-bean )?

kavirajk: Good catch. :+1:

pkg/storage/chunk/cache/groupcache.go
periklis: Defaults are per se looking like this:
```go
var DefaultTransport [RoundTripper](https://pkg.go.dev/net/http#RoundTripper) = &[Transport](https://pkg.go.dev/net/http#Transport){
	Proxy: [ProxyFromEnvironment](https://pkg.go.dev/net/http#ProxyFromEnvironment),
	DialContext: defaultTransportDialContext(&[net](https://pkg.go.dev/net).[Dialer](https://pkg.go.dev/net#Dialer){
		Timeout:   30 * [time](https://pkg.go.dev/time).[Second](https://pkg.go.dev/time#Second),
		KeepAlive: 30 * [time](https://pkg.go.dev/time).[Second](https://pkg.go.dev/time#Second),
	}),
	ForceAttemptHTTP2:     [true](https://pkg.go.dev/builtin#true),
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * [time](https://pkg.go.dev/time).[Second](https://pkg.go.dev/time#Second),
	TLSHandshakeTimeout:   10 * [time](https://pkg.go.dev/time).[Second](https://pkg.go.dev/time#Second),
	ExpectContinueTimeout: 1 * [time](https://pkg.go.dev/time).[Second](https://pkg.go.dev/time#Second),
}
```

jeschkies (commented):
this may be a naive question but what is the problem with unbounded goroutines? I thought the scheduler can easily handle hundred of thousands of goroutines?

pkg/loki/modules.go
MasslessParticle: http2 was only added for this feature so removing it would just be putting us back where we were

MasslessParticle:
@jeschkies We're also seeing unbounded memory growth because of the memory associated with each request

MasslessParticle:
This isn't necessary after all

PR #6736, opened 2022-07-21
query-tee: support for multi backend response comparison and disabling query proxy for select backends

sandeepsukhani:
It contains the following 2 changes for query-tee:
1. query-tee enables query response comparison between a preferred backend and a secondary backend. However, it did not allow multiple secondary backends to query and compare their responses. This PR updates the code where it would compare responses from all the secondary backends with the primary backend. I have also added a `name` label to the metric, which records the result of query response comparison.
2. Add a flag to disable just the query proxy for select secondary backends. This would allow us to temporarily disable read queries for one of the secondary backends while continuing to keep writing logs to it.

I think we should rename it to `tee` instead of `query-tee` :D

**Checklist**
- [x] Tests updated

### DIFF
--- a/tools/querytee/proxy.go
+++ b/tools/querytee/proxy.go
@@ -28,6 +28,7 @@ type ProxyConfig struct {
 	PreferredBackend               string
 	BackendReadTimeout             time.Duration
 	CompareResponses               bool
+	DisableBackendReadProxy        string
 	ValueComparisonTolerance       float64
 	UseRelativeError               bool
 	PassThroughNonRegisteredRoutes bool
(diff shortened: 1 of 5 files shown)


chaudum (approved):

tools/querytee/proxy.go
```diff
@@ -133,7 +137,7 @@ func NewProxy(cfg ProxyConfig, logger log.Logger, routes []Route, registerer pro
 		}
 	}
 
-	if cfg.CompareResponses && len(p.backends) != 2 {
+	if cfg.CompareResponses && len(p.backends) < 2 {
 		return nil, fmt.Errorf("when enabling comparison of results number of backends should be 2 exactly")
```
salvacorts, line 141:
```suggestion
		return nil, fmt.Errorf("when enabling comparison of results number of backends should be at least 2")
```

salvacorts (approved):
LGTM. Just left a change suggestion.

PR #6752, opened 2022-07-22
Groupcache improvements

MasslessParticle:
This PR consists of 2 changes that improve the performance of groupcache and make it usable:
1. Don't use an distributed cache for the index cache. The index cache is queried hundreds of thousands of times a minute. With a distributed cache, this generates as much as 1,000,000 requests/minute and causes everything to fail
2. Serve groupcache traffic on it's own server. This has the effect of reducing memory consumption on servers by almost 1 GiB.


slim-bean:
> Serve groupcache traffic on it's own server. This has the effect of reducing memory consumption on servers by almost 1 GiB.

Any thoughts on why this is the case?

I wonder if it's related to all the middlewares we add 😐

pkg/storage/chunk/cache/cache.go
```diff
@@ -87,7 +87,7 @@ func IsRedisSet(cfg Config) bool {
 }
 
 func IsGroupCacheSet(cfg Config) bool {
-	return cfg.EnableGroupCache
+	return cfg.GroupCache != nil
```
dannykopping, line 90: I don't think it'll ever be `nil`?
https://github.com/grafana/loki/blob/7e76375199702a29dff3c925f49f4dc311ad61a7/pkg/storage/chunk/cache/cache.go#L43

pkg/loki/modules.go
```diff
@@ -180,7 +168,10 @@ func (t *Loki) initGroupcache() (_ services.Service, err error) {
 
 	t.Cfg.ChunkStoreConfig.ChunkCacheConfig.GroupCache = gc.NewGroup(t.Cfg.ChunkStoreConfig.ChunkCacheConfig.Prefix+"groupcache", stats.ChunkCache)
 	t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.GroupCache = gc.NewGroup(t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.Prefix+"groupcache", stats.ResultCache)
-	t.Cfg.StorageConfig.IndexQueriesCacheConfig.GroupCache = gc.NewGroup(t.Cfg.StorageConfig.IndexQueriesCacheConfig.Prefix+"groupcache", stats.IndexCache)
+
+	// The index cache generates too much traffic to be used. Make it a fifo cache
+	t.Cfg.StorageConfig.IndexQueriesCacheConfig.EnableFifoCache = true
+	t.Cfg.StorageConfig.IndexQueriesCacheConfig.Fifocache.MaxSizeBytes = fmt.Sprint(t.Cfg.Common.GroupCacheConfig.CapacityMB * 1e6)
```
dannykopping, line 174: :heart:

dannykopping (approved):
Awesome work @MasslessParticle :tada:

vendor/modules.txt
```diff
@@ -1217,7 +1217,6 @@ golang.org/x/net/context/ctxhttp
 golang.org/x/net/http/httpguts
 golang.org/x/net/http/httpproxy
 golang.org/x/net/http2
-golang.org/x/net/http2/h2c
```
dannykopping, line 1220: Hhmm so are we not using h2c anymore?

MasslessParticle: We're not. It was being used to upgrade `http/1.1` connections to `http/2`. Now, we're just serving `http/2` here https://github.com/grafana/loki/blob/27b9fe50f0fe85607706a0ac6aa45ad032d0bf0a/pkg/storage/chunk/cache/groupcache.go#L133-L157

PR #6755, opened 2022-07-23
Lambda-Promtail: add a random shard env variable to lambda promtail

slim-bean:

For high volume log groups there isn't an easy way to split up the volume and often you can hit the Loki per_stream_rate_limit.

This PR let's you assign some random label sharding in lambda_promtail to allow for "hopefully" a decent random distribution of log lines over the configured `RANDOM_SHARD_SIZE`

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/tools/lambda-promtail/lambda-promtail/main.go
+++ b/tools/lambda-promtail/lambda-promtail/main.go
@@ -5,10 +5,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math/rand"
 	"net/url"
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/prometheus/common/model"
 
@@ -31,6 +33,7 @@ var (
 	username, password, extraLabelsRaw, tenantID string
 	keepStream                                   bool
 	batchSize                                    int
+	randomShardSize                              int
 	s3Clients                                    map[string]*s3.Client
 	extraLabels                                  model.LabelSet
 )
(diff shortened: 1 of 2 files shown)


tools/lambda-promtail/lambda-promtail/main.go
```diff
@@ -77,6 +80,12 @@ func setupArguments() {
 		batchSize, _ = strconv.Atoi(batch)
 	}
 
+	randomShard := os.Getenv("RANDOM_SHARD_SIZE")
```
chaudum, line 83: I find the name `RANDOM_SHARD_SIZE` rather misleading. The shard size is not random, but a random shard ID between 0 and the provided value is assigned. Therefore I'd rather name it `SHARD_SIZE` or `RANDOM_SHARD_ID`, where the latter somehow indicates a boolean.

chaudum (commented):
A changelog entry is missing.

owen-d (approved):

tools/lambda-promtail/lambda-promtail/main.go
cstyan: I agree. It could just be `NUM_SHARDS` or `NUM_STREAM_SHARDS`?

slim-bean: I don't disagree however this config was already supplied to a user and I don't want to change the name and break them in the future. Specifically because we will no longer need this when Loki can auto shard high volume streams automatically.

This is why I'm not really going to expose this in the docs.

slim-bean:
Trying to decide if this should be merged or not. we are working on having auto shard high volume streams right now and that would eliminate the need for this option.

MasslessParticle:
Automatic Stream Sharding Means we don't need this anymore. Closing

PR #6762, opened 2022-07-25
Stream responses from Querier to Query Frontend.

jeschkies:
The is the result of a time boxed experiment. I wanted to see if we can stream results from the querier to the query frontend without changing the scheduler API too much.

It seems possible. However, right now the benefit is minimal. The JSON ecnoder, jsoniter, does not support stream encoding (https://github.com/json-iterator/go/issues/488). We also buffer before serialization.

Querier to query frontend streaming would enable
* Partial result streaming.
* Remote read API without buffering.
* Less buffering, i.e. fewer allocations and less memory consumption.


pkg/querier/worker/scheduler_processor.go
```diff
@@ -140,13 +141,165 @@ func (sp *schedulerProcessor) querierLoop(c schedulerpb.SchedulerForQuerier_Quer
...
+		one := buf[0:512]
+		buf = buf[512:]
+
+		if !r.sentFirst {
+			r.recorder.Write(one)
+		} else {
+			r.body.Write(one)
+		}
+		r.Flush()
+	}
```
jeschkies, line 215: This is just chunking is enforced. Ideally we use a buffered writer.

PR #6763, opened 2022-07-25
Rephrase error message (indexgateway->groupcache)

DylanGuedes:
Change error message text from indexgateway to groupcache.


dannykopping (approved):
Whoops, thanks!

Issue #6713, opened 2022-07-19
Issue using label filter prior to unwrapping the same field


owen-d:
```
# This works
sum(sum_over_time({job="loki-ops/query-frontend"}
  |= "metrics.go"
  | logfmt
  | subqueries!=0
  | unwrap subqueries
  [5m]
))

# but this does not parse correctly
sum(sum_over_time({job="loki-ops/query-frontend"}
  |= "metrics.go"
  | logfmt
  | subqueries!="0"
  | unwrap subqueries
  [5m]
))
```

Note the string `!="0"` does not work, but the numeric version `!=0` does. I suspect we're doing some incorrect type conversion/etc somewhere.

kavirajk:
quickly checked. Looks like this issue is only when using "!=" operation in label filters. works fine for "=" with both string and numeric values.

And using other binary operations like "<=" or ">=" returns decent error message.
```
syntax error: unexpected STRING, expecting BYTES or NUMBER or DURATION (<nil>)
```
Still investigating why is this case for "!=".

cyriltovena:
I think the query should be 

```
sum(sum_over_time({job="loki-ops/query-frontend"}
  |= "metrics.go"
  | logfmt
  | subqueries!=""
  | unwrap subqueries
  [5m]
))
```

but I'm not sure why 

```
sum(sum_over_time({job="loki-ops/query-frontend"}
  |= "metrics.go"
  | logfmt
  | subqueries!=0
  | unwrap subqueries
  [5m]
))
```
work you might have been lucky ? or we do have an issue because empty label should error when converting to a number.

owen-d:
> or we do have an issue because empty label should error when converting to a number.
This is what I suspect is the problem too

kavirajk:
> work you might have been lucky ? 

The problem is 100% reproducible btw.

The core of the problem is when trying to extract the metric sample from a logline via unwrap, and if that label doesn't exist on the logline, the this error is set.

it comes from here. https://github.com/grafana/loki/blob/kavirajk%2Flogql-label-check-bug/pkg/logql/log/metrics_extraction.go#L181-L182

Two options here.
1. We can fix this only on during metrics extraction (if label doesn't exist, assume to have value of 0 as sample value)
2. Or we ignore log lines that doesn't contain label during != and !~ like in this PR #6766 

**Before rejecting the option (2). Please have a look at the below screen shots. Is it ok to have two different responses for these queries.**

Q1
```
[...]

PR #6766, opened 2022-07-25
fix(logql): Make `LabelSampleExtractor` work as expected when label doesn't exist in logline

kavirajk:
Consider Log lines

```
msg=hello subqueries=5
msg=hello subqueries=0
msg=hello
```

If you run following logql
```
{foo="bar"}|logfmt|subqueries!=0
```
it returns
```
msg=hello subqueries=5
```

But if you run following logql
```
{foo="bar"}|logfmt|subqueries!="0" # NOTE: "0" as string.
```
it returns
```
msg=hello subqueries=5
msg=hello
```

NOTE: it also returns log lines that doesn't contain `subqueries` label in the first
place.

This cause subtle error if used in metric queries using that `label` in unwrap.

e.g: following logql on above log lines returns
```
sum_over_time({foo="bar"}|logfmt|subqueries!="0"|unwrap subqueries[1m])
```

Returns `pipeline error: 'SampleExtractionErr'`. Because, the lines without `subqueries`
labels are not skipped during `ProcessLine` and during metric extraction, our extractor
sets this `SampleExtractionerror`.

**EDIT**: After some discussions with @cyriltovena and @owen-d. We have decided fix only the sample extractor without changing any behavior of label filter.

Fixes # https://github.com/grafana/loki/issues/6713

**Checklist**
- [ ] Documentation added
- [x] Tests updated


pkg/logql/log/label_filter.go
```diff
@@ -342,7 +342,10 @@ func (s *StringLabelFilter) Process(_ int64, line []byte, lbs *LabelsBuilder) ([
 	if s.Name == logqlmodel.ErrorLabel {
 		return line, s.Matches(lbs.GetErr())
 	}
-	v, _ := lbs.Get(s.Name)
+	v, ok := lbs.Get(s.Name)
```
kavirajk, line 345: this is the real fix.

pkg/logql/log/label_filter_test.go
```diff
@@ -350,3 +351,32 @@ func TestReduceAndLabelFilter(t *testing.T) {
...
+		filter      *StringLabelFilter
+		labels      labels.Labels
+		shouldMatch bool
+	}{
+		{
+			name:        "without-label",
+			filter:      NewStringLabelFilter(labels.MustNewMatcher(labels.MatchNotEqual, "subqueries", "0")),
+			labels:      labels.Labels{{Name: "msg", Value: "hello"}}, // no label `subqueries`
+			shouldMatch: false,
+		},
```
ssncferreira, line 367: It would also be nice to add a test where `subqueries` is "0" and another where it is the empty string:
```
labels.Labels{{Name: "msg", Value: "hello"}, {Name: "subqueries", Value: "0"}},
labels.Labels{{Name: "msg", Value: "hello"}, {Name: "subqueries", Value: ""}},
```

pkg/logql/log/metrics_extraction_test.go
```diff
@@ -152,6 +153,69 @@ func Test_Extract_ExpectedLabels(t *testing.T) {
...
+		name       string
+		extractor  SampleExtractor
+		checkLines []checkLine
+		shouldFail bool
+	}{
+		{
+			name: "with just logfmt and stringlabelfilter",
+			// {foo="bar"} | logfmt | subqueries != "0" (note: "0", a stringlabelfilter)
+			extractor: mustSampleExtractor(
+				LabelExtractorWithStages("subqueries", ConvertFloat, []string{"foo"}, false, false, []Stage{NewLogfmtParser(), NewStringLabelFilter(labels.MustNewMatcher(labels.MatchNotEqual, "subqueries", "0"))}, NoopStage),
```
ssncferreira, line 178: Why does it have the `ConvertFloat` stage? 🤔 
If the query is in string `subqueries != "0"` does it still attempts to convert to float?

ssncferreira (commented):
I'm a bit confused here because this seems related to the value type that we use in the query, i.e. either string `!= "0"` or integer `!= 0`.
If a log line does not contain a label, e.g.
```
msg=hello
```
I would always assume the value of the label `subqueries` would be the empty string `""`, and, therefore, it would be different than both string `"0"` and integer `0`. 
With the changes from this PR it seems this assumption is wrong, so it should be clearly documented

pkg/logql/log/metrics_extraction_test.go
kavirajk: NOTE this is the `metrics_extraction` (used in metrics query not log query). 

1. ConvertFloat is to extract the value from the logline (subqueries) and convert it into sample value (other conversation methods like ConvertBytes, ConvertDuration, will treat that value differently)
2. `subqueries != "0"` is just to match the logs as a labelmatcher after `logfmt`

Both are different. Does that answer your question @ssncferreira ?

pkg/logql/log/label_filter_test.go
kavirajk: good point and nice catch. I considered `subqueries !="0"` response should be exactly same as `subqueries != 0`. But I failed to consider the case what happens if `subquerie =""`. You are right it should match lines that doesn't have label for queries like `subqueries=""` and `subqueries~=""`

Fixed it in https://github.com/grafana/loki/pull/6766/commits/bab6ca046c0382702aa0992577a8d9db5dc01342

does that sounds better?

ssncferreira: Looks good, thanks 👍

ssncferreira (approved):
LGTM! Thanks 🙏 
Do you think it would also make sense to update the documentation with some examples? (can be part of a separate PR)

pkg/logql/log/label_filter.go
```diff
@@ -342,7 +342,24 @@ func (s *StringLabelFilter) Process(_ int64, line []byte, lbs *LabelsBuilder) ([
...
+	// NOTE(kavi): whether the `label` exists in logline matters when doing `!=` or `!~`.
+	// Because, consider following long lines
+	// ```
+	// 1.msg="hello" subqueries=5
+	// 2.msg="hello" subqueries=0
+	// 3.msg="hello"
+	// 4.msg="hello" subqueries=""
+	//```
+
+	// Now when quering like `{...}|logfmt|subqueries=""` should consider lines 3 and 4.
```
cyriltovena, line 356: Not sure about this, it doesn't really make it easier to understand for the end user as this make it more obscure:

- when using equality we consider non existent label empty 
- but for inequality we don't.

Honestly in https://github.com/grafana/loki/issues/6713  I think the expectations where wrong.

```
sum(sum_over_time({job="loki-ops/query-frontend"}
  |= "metrics.go"
  | logfmt
  | subqueries!="0"
  | unwrap subqueries
  [5m]
))
```

When subqueries doesn't exist then subqueries!="0" is true.

What is weird was that empty labels does transform into a 0 integer.

kavirajk: > When subqueries doesn't exist then subqueries!="0" is true.

So you saying `subqueries!=0` and `subqueries!="0"` ok to return different log lines?. But that's unexpected behaviour to end-user IMHO. Both should return exactly same results IMO.

> What is weird was that empty labels does transform into a 0 integer.

That's because, we don't mutate the log line if label is not present. So treat it as if label="" during label filter, But still log line doesn't contain actual label, which breaks when doing unwrap on same label during metrics extraction.

owen-d: > What is weird was that empty labels does transform into a 0 integer.

I agree with @cyriltovena. This is the part I think we need to address.

kavirajk: > May be we discard empty values ? If that is possible.

Yes that's possible. We can skip empty values during metrics extraction when using unwrap without changing any of the existing label filter behavior.

Also the more I think about it, subqueries!="0" and subqueries!=0 are two different label filters and should be treated as such. And former can indeed match log lines that doesn't have subqueries label at all (we are assuming it has default empty string substring=" when doesn't exist).

I have change the PR like it's mentioned here.

cyriltovena (approved):
LGTM

I think the changelog needs a better documentation of the change.

PR #6777, opened 2022-07-26
Promtail: Add new target for receiving GCP PubSub Push messages

thepalbi:
Promtail currently supports scraping GCP logs using a [pull model](https://grafana.com/docs/loki/next/clients/promtail/scraping/#gcplog-scraping), in which promtail subscribes to a GCP PubSub topic, to where logs are routed. There's another mechanism possible for receiving messages from PubSub topics (for example logs, but could potentially be other kind of messages) which is by using a [push subscription](https://cloud.google.com/pubsub/docs/push).

This PR adds a new Promtail target which is able to receive messages from a GCP PubSub push subscription. It has been mainly thought for logs, but since it's generic it's able to support whatever message the GCP topic sends, since they all use a similar envelope. The target exposes and HTTP server (with just one `POST /gcp/api/v1/push` endpoint), which is then configured as target of this Push subscriptions.

Just to give an example, the envelope that GCP uses for this messages has the following shape:
```json
{
	"message": {
		"attributes": {
			"logging.googleapis.com/timestamp": "2022-07-25T22:19:09.903683708Z"
		},
		"data": "ewogICJzZXZlcml0eSI6ICJERUJVRyIsCiAgInRleHRQYXlsb2FkIjogIkZ1bmN0aW9uIGV4ZWN1dGlvbiB0b29rIDExOTggbXMuIEZpbmlzaGVkIHdpdGggc3RhdHVzIGNvZGU6IDIwMCIsCiAgInRyYWNlIjogInByb2plY3RzL3dpcmVkLWhlaWdodC90cmFjZXMvNTRhYThhMjA4NzUyNTNjNmI0N2NhYThiZDI4YWQ2NTIiCn0=",
		"messageId": "5187581549398349",
		"message_id": "5187581549398349",
		"publishTime": "2022-07-25T22:19:15.56Z",
		"publish_time": "2022-07-25T22:19:15.56Z"
	},
	"subscription": "projects/wired-height-350515/subscriptions/test"
}
```
This message has been extracted from a push subscription, listening for logs from a cloud-function.

The target can be configured as follows:
```
  gcp_push:
    server:
      http_listen_address: 0.0.0.0
      http_listen_port: 8080
      log_level: debug
      register_instrumentation: true
    labels: 
      job: gcp_push_logs_test
    use_incoming_timestamp: true
```

This target exposes some of the received "label" as "internal labels" to be re-labeled by the user using the scrape config (much like the syslog target does), and dropped if not used. These are:

- `__gcp_message_id`: message id
- `__gcp_attributes_*`: All attributes read from `.message.attributes` are added as a temporal label, conveniently adapting the label name since it might contain non-loki-supported characters. For example for `logging.googleapis.com/timestamp` it's converted to `__gcp_attributes_logging_googleapis_com_timestamp`.

**Some things to keep in mind**:
- GCP offers some nice delivery garantuees for not overloading the receiving sender like a [sliding window algo with slow start](https://cloud.google.com/pubsub/docs/push#delivery_rate).
- It also offers acks and nacks, and retries, but because of promtail async processing, this are lost when the entry enters the underlying channel towards loki's client. The idea is once WAL-support is added, this will improve vastly.

n/a

Documentation to be added.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


clients/pkg/promtail/targets/serverutils/config.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"flag"
+
+	"github.com/imdario/mergo"
+	"github.com/weaveworks/common/server"
+)
+
+// MergeWithDefaults applies server.Config defaults to a given and different server.Config.
+func MergeWithDefaults(config server.Config) (server.Config, error) {
```
thepalbi, line 11: This functionality was shared across all promtails target spinning up an http server (loki push, heroku and this new gcp push one), so extracted it in a `serverutils` package. Happy to rename it or move it elsewhere.

clients/pkg/promtail/targets/gcppush/target_test.go
```diff
@@ -0,0 +1,300 @@
...
+			"logging.googleapis.com/timestamp": "2022-07-25T22:19:09.903683708Z"
+		},
+		"data": "ewogICJzZXZlcml0eSI6ICJERUJVRyIsCiAgInRleHRQYXlsb2FkIjogIkZ1bmN0aW9uIGV4ZWN1dGlvbiB0b29rIDExOTggbXMuIEZpbmlzaGVkIHdpdGggc3RhdHVzIGNvZGU6IDIwMCIsCiAgInRyYWNlIjogInByb2plY3RzL3dpcmVkLWhlaWdodC90cmFjZXMvNTRhYThhMjA4NzUyNTNjNmI0N2NhYThiZDI4YWQ2NTIiCn0=",
+		"messageId": "5187581549398349",
+		"message_id": "5187581549398349",
+		"publishTime": "2022-07-25T22:19:15.56Z",
+		"publish_time": "2022-07-25T22:19:15.56Z"
+	},
+	"subscription": "projects/wired-height-350515/subscriptions/test"
+}`
```
thepalbi, line 45: Sample payload obtained by setting up a dummy push subscription receiver that logs everything. The `message.data` was simplified, which is just decoded and forwarded to loki, and it can vary depending on what is forwarded from the subscribed topic.

clients/pkg/promtail/targets/gcppush/translation.go
```diff
@@ -0,0 +1,117 @@
...
+		if err != nil {
+			return api.Entry{}, fmt.Errorf("invalid publish timestamp format: %w", err)
+		}
+	}
+
+	// If the incoming request carries the tenant id, inject it as the reserved label, so it's used by the
+	// remote write client.
+	if xScopeOrgID != "" {
+		labels[lokiClient.ReservedLabelTenantID] = model.LabelValue(xScopeOrgID)
+	}
```
thepalbi, line 99: This functionality mimics what was done in https://github.com/grafana/loki/commit/d4cdc376820867cd88248a9b6874e18b1cdac6e6. The idea is to make this target multitenant-supporting, so one can inject the tenantID from upstream using this reserved header (which is well-kown in loki world) and let the loki-client forward it when doing the remote write.

clients/pkg/promtail/targets/gcppush/translation.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+	}, nil
+}
+
+// convertToLokiCompatibleLabel converts an incoming GCP Push message label to a loki compatible format. There are lables
+// such as `logging.googleapis.com/timestamp`, which contain non-loki-compatible characters, which is just alphanumeric
+// and _. The approach taken is to translate every non-alphanumeric separator character to an underscore.
+func convertToLokiCompatibleLabel(label string) string {
+	// TODO: Since this is running for every incoming message, maybe it's more performant to do something
+	// like a loop over characters, and checking if it's not loki compatible, instead of a regexp.
```
thepalbi, line 115: For reviewer, could use some input in here

clients/pkg/promtail/targets/gcppush/translation.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/grafana/loki/pkg/logproto"
+	"github.com/grafana/loki/pkg/util"
+)
+
+// Configured as a global in this file to avoid recompiling this regex everywhere.
+var labelToLokiCompatible *regexp.Regexp
+
+func init() {
+	// TODO: Maybe use a regexp negative filter and grab everything non-alphanumeric non-underscore?
+	labelToLokiCompatible = regexp.MustCompile("[.-/]")
```
thepalbi, line 26: For reviewer, could use some input in here

kavirajk:
@thepalbi thanks a ton for adding support for GCP pubsub `push` subscription type. Skimmed through the PR and overall looks great :+1: 

Just wanted to clarify one thing. Can we still keep single target type for GCP(`gcplog`) and just add extra config `subscription_type` to differentiate `push` and `pull`? (default being `pull`)? Rather than adding completely new target type.
I see we use more or less similar config for both except the `server` block for `push` subscription type.

We can keep everything more or less same, except if `subscription_type: push` we create newGCPpushTarget.

wdyt?

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
```diff
@@ -37,7 +37,7 @@ type Config struct {
 	PipelineStages    stages.PipelineStages      `yaml:"pipeline_stages,omitempty"`
 	JournalConfig     *JournalTargetConfig       `yaml:"journal,omitempty"`
 	SyslogConfig      *SyslogTargetConfig        `yaml:"syslog,omitempty"`
-	GcplogConfig      *GcplogTargetConfig        `yaml:"gcplog,omitempty"`
+	GcplogConfig      *GCPLogTargetConfig        `yaml:"gcplog,omitempty"`
```
thepalbi, line 40: drive by: renamed the main target to use a better case

clients/pkg/promtail/targets/gcplog/pull_target.go
```diff
@@ -0,0 +1,164 @@
...
+
+	go func() {
+		_ = target.run()
+	}()
+
+	return target, nil
+}
+
+// nolint:revive
+func newGcplogTarget(
```
thepalbi, line 70: nit: maybe this one should also be renamed?

clients/pkg/promtail/targets/gcplog/push_translation.go
```diff
@@ -0,0 +1,117 @@
...
+		if err != nil {
+			return api.Entry{}, fmt.Errorf("invalid publish timestamp format: %w", err)
+		}
+	}
+
+	// If the incoming request carries the tenant id, inject it as the reserved label, so it's used by the
+	// remote write client.
+	if xScopeOrgID != "" {
+		labels[lokiClient.ReservedLabelTenantID] = model.LabelValue(xScopeOrgID)
+	}
```
thepalbi, line 99: This functionality mimics what was done in https://github.com/grafana/loki/commit/d4cdc376820867cd88248a9b6874e18b1cdac6e6. The idea is to make this target multitenant-supporting, so one can inject the tenantID from upstream using this reserved header (which is well-kown in loki world) and let the loki-client forward it when doing the remote write.

thepalbi:
> @thepalbi thanks a ton for adding support for GCP pubsub `push` subscription type. Skimmed through the PR and overall looks great 👍
> 
> Just wanted to clarify one thing. Can we still keep single target type for GCP(`gcplog`) and just add extra config `subscription_type` to differentiate `push` and `pull`? (default being `pull`)? Rather than adding completely new target type. I see we use more or less similar config for both except the `server` block for `push` subscription type.
> 
> We can keep everything more or less same, except if `subscription_type: push` we create newGCPpushTarget.
> 
> wdyt?

@kavirajk completely agree. Just merged both targets under the same package. Refactored (mainly renames) a little bit the Pull one so that both of them fit better. Let me know what you think 😀

clients/pkg/promtail/targets/gcplog/push_translation.go
```diff
@@ -0,0 +1,117 @@
...
+	}, nil
+}
+
+// convertToLokiCompatibleLabel converts an incoming GCP Push message label to a loki compatible format. There are lables
+// such as `logging.googleapis.com/timestamp`, which contain non-loki-compatible characters, which is just alphanumeric
+// and _. The approach taken is to translate every non-alphanumeric separator character to an underscore.
+func convertToLokiCompatibleLabel(label string) string {
+	// TODO: Since this is running for every incoming message, maybe it's more performant to do something
+	// like a loop over characters, and checking if it's not loki compatible, instead of a regexp.
+	return util.SnakeCase(labelToLokiCompatible.ReplaceAllString(label, "_"))
```
kavirajk, line 116: Good observation about using loop instead of regexp. But let's may be keep that for later, if that end up as bottleneck.

clients/pkg/promtail/targets/gcplog/target.go
```diff
@@ -1,165 +1,38 @@
...
-		return nil, err
+	config *scrapeconfig.GCPLogTargetConfig,
+) (Target, error) {
+	switch config.SubscriptionType {
+	case "pull", "":
+		return newPullTarget(metrics, logger, handler, relabel, jobName, config)
+	case "push":
+		return newPushTarget(metrics, logger, handler, jobName, config, relabel)
+	default:
+		return nil, fmt.Errorf("invalid subscription type: %s", config.SubscriptionType)
```
kavirajk, line 36:
```suggestion
		return nil, fmt.Errorf("invalid subscription type: %s. valid options are 'push' and 'pull'", config.SubscriptionType)
```

clients/pkg/promtail/targets/target/target.go
```diff
@@ -44,6 +44,9 @@ const (
 
 	// HerokuDrainTargetType is a Heroku Logs target
 	HerokuDrainTargetType = TargetType("HerokuDrain")
+
+	// GCPPushTargetType is a GCP PubSub Push target
+	GCPPushTargetType = TargetType("GCPPush")
```
kavirajk, line 49: We don't need this anymore right?

clients/pkg/promtail/targets/gcplog/pull_target.go
```diff
@@ -0,0 +1,164 @@
...
+
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+	"github.com/grafana/loki/clients/pkg/promtail/scrapeconfig"
+	"github.com/grafana/loki/clients/pkg/promtail/targets/target"
+)
+
+// PullTarget represents the target specific to GCP project, with a pull subscription type.
+// It collects logs from GCP and push it to Loki.
+// nolint:revive
+type PullTarget struct {
```
kavirajk, line 21: Now `PullTarget` doesn't have to be public correct?.

clients/pkg/promtail/targets/gcplog/pull_target.go
kavirajk: this is just returning the struct. why don't we make it as part of `newPullTarget` itself?. Any reason to have it as separate function?

kavirajk (commented):
LGTM :tada: thanks @thepalbi for addressing the comments.

Left few minor nits and suggestion. We are good to merge after that.

kavirajk:
@thepalbi also can you also please update the docs accordingly?
https://grafana.com/docs/loki/latest/clients/promtail/gcplog-cloud/

clients/pkg/promtail/targets/gcplog/push_translation.go
thepalbi: For sure, will leave the comment in case we need to re-evaluate

clients/pkg/promtail/targets/target/target.go
thepalbi: Right!

clients/pkg/promtail/targets/gcplog/pull_target.go
thepalbi: Yup, since the actual type returned by the public `NewGCPLogTarget` func is `gcplog.Target`, we can keep the actual target types package private.

thepalbi:
> @thepalbi also can you also please update the docs accordingly? https://grafana.com/docs/loki/latest/clients/promtail/gcplog-cloud/

Yup, wanted to have a final 👍 on the implementation before doing this. Will add the docs so the PR is sounds!

clients/pkg/promtail/targets/target/target.go
```diff
@@ -18,8 +18,8 @@ const (
 	// SyslogTargetType is a syslog target
 	SyslogTargetType = TargetType("Syslog")
 
-	// GcplogTargetType is a target where log entries are pulled from pubsub topic.
-	GcplogTargetType = TargetType("Gcplog")
+	// GCPLogTargetType is a target where log entries are pulled from pubsub topic.
+	GCPLogTargetType = TargetType("GCPLog")
```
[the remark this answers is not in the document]
thepalbi: Done, reverted all those renames to make everything as backwards compatible as possible.

thepalbi:
@kavirajk added the docs on `configuration` and `scraping` sections as discussed. Also resolved all open comments. Can you take another pass + tag the correct docs person?

cstyan:
Do we have any concerns about this target potentially allowing users to easily hit per stream rate limit issues? It looks to me like it would be very easy to configure many things to all write to one pubsub and have it all pushed to promtail as a single stream. 

Should we consider enforcing some amount of the attributes being used as labels?

kavirajk:
@cstyan that's the perfectly valid concern. 

Even now user can take advantage of relabeling those gcp atrributes.
https://github.com/grafana/loki/blob/pablo%2Fgcp-push-target/clients/pkg/promtail/targets/gcplog/formatter.go#L60-L72

And you are right, user may not be aware about the impact before sending too much messages into single stream.

IMO, good place to add this note is in gcp-cloud setup doc https://grafana.com/docs/loki/latest/clients/promtail/gcplog-cloud/

I think @thepalbi decided to update the cloud setup doc in the follow up PR. Can you take care of that when creating follow up PR @thepalbi ?

kavirajk (approved):
LGTM :+1: 

@KMiller-Grafana can you please approve the doc changes?

thepalbi:
@cstyan sorry I missed you comment. That's a valid point you have. In the trails I ran with actual GCP subscription, I couldn't find any common attribute that might be interested to add to each stream. I know parsing the actual log line (which is always json) can lead to several useful ones. More than happy to:
1. Document this as @kavirajk mentioned
2. If we find some common pattern like a pipeline opening the json and grabbing some field, or unmarshalling directly in the target address it in a follow up

docs/sources/clients/promtail/configuration.md
```diff
@@ -330,6 +330,9 @@ job_name: <string>
 # Describes how to scrape logs from the Windows event logs.
 [windows_events: <windows_events_config>]
 
+# Configuration describing how to pull/receive GCP Logs.
+[gcplog: <gcplog_config>]
```
KMiller-Grafana, line 334: Maybe it's obvious, but I'd like to spell out what GCP means.
```suggestion
# Configuration describing how to pull/receive Google Cloud Platform (GCP) logs.
[gcplog: <gcplog_config>]
```

docs/sources/clients/promtail/configuration.md
```diff
@@ -930,6 +933,55 @@ labels:
...
 ```
 
+### GCP Log
+
+The `gcplog` block configures Promtail receive GCP logs. It has two possible strategies for doing so, depending on how the 
+`subscription_type` attribute is configured:
+1. **Pull**: Using GCP PubSub [pull subscriptions](https://cloud.google.com/pubsub/docs/pull). Promtail will consume log messages directly from the configured GCP PubSub topic.
+2. **Push**: Using GCP PubSub [push subscriptions](https://cloud.google.com/pubsub/docs/push). Promtail will expose an HTTP server, to which GCP will deliver logs.
+
+When picking the `push` subscription type, keep in mind the following:
```
KMiller-Grafana, line 943: Thanks for accepting all these small changes! I'm also copying spelling of "Pub/Sub" from Google Cloud documentation.
```suggestion
The `gcplog` block configures how Promtail receives GCP logs. There are two strategies, based on the configuration of `subscription_type`:
- **Pull**: Using GCP Pub/Sub [pull subscriptions](https://cloud.google.com/pubsub/docs/pull). Promtail will consume log messages directly from the configured GCP Pub/Sub topic.
- **Push**: Using GCP Pub/Sub [push subscriptions](https://cloud.google.com/pubsub/docs/push). Promtail will expose an HTTP server, and GCP will deliver logs to that server.

When using the `push` subscription type, keep in mind:
```

docs/sources/clients/promtail/configuration.md
```diff
@@ -930,6 +933,55 @@ labels:
...
+- The `server` configuration is the same as [server](#server), since Promtail exposes an HTTP server for target that requires so.
+- An endpoint at `POST /gcp/api/v1/push`, which expects requests from GCP PubSub message delivery system.
```
KMiller-Grafana, line 945:
```suggestion
- The `server` configuration is the same as [server](#server), since Promtail exposes an HTTP server for the target that requires it to be the same.
- An endpoint at `POST /gcp/api/v1/push`, which expects requests from a GCP Pub/Sub message delivery system.
```

docs/sources/clients/promtail/configuration.md
```diff
@@ -930,6 +933,55 @@ labels:
...
+[subscription_type: <string> | default = "pull"]
+
+# [pull] GCP project ID
+[project_id: <string>]
+
+# [pull] GCP PubSub subscription from where Promtail will pull logs from
+[subscription: <string>]
+
+# [push] Server configuration options
+[server: <server_config>]
```
KMiller-Grafana, line 958: Since we're using the square brackets ([]) to indicate optional configuration parameters, I'd like to eliminate them from the description prose.  My preference is to state the subscription_type in the sentence, as in this suggestion:
```suggestion
# If the subscription_type is pull,  the GCP project ID
[project_id: <string>]

# If the subscription_type is pull, GCP PubSub subscription from where Promtail will pull logs from
[subscription: <string>]

# If the subscription_type is push, the server configuration options
[server: <server_config>]
```

docs/sources/clients/promtail/configuration.md
```diff
@@ -930,6 +933,55 @@ labels:
...
+# Label map to add to every log message.
+labels:
+  [ <labelname>: <labelvalue> ... ]
+```
+
+### Available Labels
+
+When Promtail receives GCP logs, various internal labels are made available for [relabeling](#relabeling). This depend on the subscription type chosen.
+
+**Pull**
```
KMiller-Grafana, line 974: How about if we make this easier to skim for our readers?
```suggestion
**Internal labels available for pull**
```

docs/sources/clients/promtail/configuration.md
```diff
@@ -930,6 +933,55 @@ labels:
...
+
+- `__gcp_logname`
+- `__gcp_resource_type`
+- `__gcp_resource_labels_<NAME>`
+
+**Push**
```
KMiller-Grafana, line 980: And, same here as a suggestion for push:
```suggestion
**Internal labels available for push**
```

docs/sources/clients/promtail/configuration.md
```diff
@@ -930,6 +933,55 @@ labels:
...
+
+- `__gcp_message_id`
+- `__gcp_attributes_*`: All attributes read from `.message.attributes` in the incoming push message. Each attribute key is conveniently renamed since it might contain unsupported characters. For example for `logging.googleapis.com/timestamp` it's converted to `__gcp_attributes_logging_googleapis_com_timestamp`.
```
KMiller-Grafana, line 983: This might be easier to parse and read:
```suggestion
- `__gcp_attributes_*`: All attributes read from `.message.attributes` in the incoming push message. Each attribute key is conveniently renamed, since it might contain unsupported characters. For example, `logging.googleapis.com/timestamp` is converted to `__gcp_attributes_logging_googleapis_com_timestamp`.
```

docs/sources/clients/promtail/scraping.md
```diff
@@ -190,13 +190,19 @@ resuming the target without skipping logs.
 
 see the [configuration](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#windows_events) section for more information.
 
-## Gcplog scraping
-Promtail supports scraping cloud resource logs(say GCS bucket logs, Load Balancer logs, Kubernetes Cluster logs) from GCP.
-Configs are set in `gcplog` section in `scrape_config`
+## GCP Log scraping
+
+Promtail supports scraping cloud resource logs (say GCS bucket logs, Load Balancer logs, Kubernetes Cluster logs) from GCP.
+Configs are set in `gcplog` section in `scrape_config`.
```
KMiller-Grafana, line 196: Small suggestions here:
```suggestion
Promtail supports scraping cloud resource logs such as GCS bucket logs, load balancer logs, and Kubernetes cluster logs from GCP.
Configuration is specified in the `gcplog` section, within `scrape_config`.
```

docs/sources/clients/promtail/scraping.md
```diff
@@ -223,6 +229,38 @@ When Promtail receives GCP logs, various internal labels are made available for
...
+      - source_labels: ['__gcp_message_id']
+        target_label: 'message_id'
+      - source_labels: ['__gcp_attributes_logging_googleapis_com_timestamp']
+        target_label: 'incoming_ts'
+```
+
+When configuring the GCP Log push target, Promtail will start a HTTP server listening in port `8080` as configured in the `server`
+section. This server will expose just one endpoint `POST /gcp/api/v1/push` responsible for receiving logs from GCP.
+
+It also supports `relabeling` and `pipeline` stages just like other targets.
```
KMiller-Grafana, line 254: Some small grammar improvements here.
```suggestion
When configuring the GCP Log push target, Promtail will start an HTTP server listening on port `8080`, as configured in the `server`
section. This server exposes the single endpoint `POST /gcp/api/v1/push`, responsible for receiving logs from GCP.

It also supports `relabeling` and `pipeline` stages.
```

KMiller-Grafana (changes requested):
Thanks for these helpful docs.

docs/sources/clients/promtail/configuration.md
thepalbi: Sure!

thepalbi:
@KMiller-Grafana addressed all comments, and left a new suggestion in one.

docs/sources/clients/promtail/configuration.md
```diff
@@ -930,6 +933,54 @@ labels:
...
+[use_incoming_timestamp: <boolean> | default = false]
+
+# Label map to add to every log message.
+labels:
+  [ <labelname>: <labelvalue> ... ]
+```
+
+### Available Labels
+
+When Promtail receives GCP logs, various internal labels are made available for [relabeling](#relabeling). This depend on the subscription type chosen.
```
KMiller-Grafana, line 971: I'm approving the docs in this PR even without this change.  Please do make this small change, as it fixes a typo.
```suggestion
When Promtail receives GCP logs, various internal labels are made available for [relabeling](#relabeling). This depends on the subscription type chosen.
```

KMiller-Grafana (approved):
Thank you for revising the docs.  The docs in this PR look good to me.  I put in a suggestion for a typo fix, but I'm approving now so you won't need a re-review from me.

clients/pkg/promtail/targets/gcplog/push_target.go
```diff
@@ -0,0 +1,158 @@
...
+		if err != nil {
+			level.Error(h.logger).Log("msg", "gcp push target shutdown with error", "err", err)
+		}
+	}()
+
+	return nil
+}
+
+func (h *pushTarget) push(w http.ResponseWriter, r *http.Request) {
+	entries := h.handler.Chan()
```
jeschkies, line 99: I'm not sure the channel would change often. Why not `h.entires`?

clients/pkg/promtail/targets/gcplog/push_target_test.go
```diff
@@ -0,0 +1,304 @@
...
+			defer eh.Stop()
+
+			serverConfig, port, err := getServerConfigWithAvailablePort()
+			require.NoError(t, err, "error generating server config or finding open port")
+			config := &scrapeconfig.GcplogTargetConfig{
+				Server:               serverConfig,
+				Labels:               tc.args.Labels,
+				UseIncomingTimestamp: false,
+				SubscriptionType:     "push",
+			}
```
jeschkies, line 138: Running the server is quite a handful. What do you think about testing the handler function `push` with a [ResponseRecorder](https://pkg.go.dev/net/http/httptest#ResponseRecorder) instead?

clients/pkg/promtail/targets/gcplog/push_translation.go
jeschkies: Sorry. I must disagree. This is a rather low hanging optimization that we will probably not attend afterwards. `util.SnakeCase` is already looping through the string. You could a simple method that also escapes characters.

jeschkies (changes requested):
Nice work!

thepalbi:
Thanks for the look @jeschkies , I'll address the comments in a follow up 😀

clients/pkg/promtail/targets/gcplog/push_target.go
thepalbi: You mean saving `h.handler.Chan()` as a struct field, and directly accessing it?

jeschkies: Sorry. I was out. Yes.

Issue #6667, opened 2022-07-12
panic: exemplar labels have 85 runes, exceeding the limit of 64


liguozhong:
**Describe the bug**

	/Users/fuling/go/pkg/mod/github.com/liguozhong/loki@v1.6.2-0.20220712032519-992e483b6271/pkg/storage/chunk/cache/instrumented.go:95 +0x1af

here

![image](https://user-images.githubusercontent.com/9583245/178498747-51f5d11e-f3d9-4301-ba79-640d817efef7.png)

![image](https://user-images.githubusercontent.com/9583245/178527150-1f9e285c-d70e-49e4-819f-a5634c79f626.png)

https://github.com/prometheus/client_golang/blob/main/prometheus/value.go
![image](https://user-images.githubusercontent.com/9583245/178498657-59f0a3fa-06fa-44b2-9a9a-51a309859cee.png)

**To Reproduce**
1. Started Loki (SHA or version)
2. Started Promtail (SHA or version) to tail '...'
3. Query: `{} term`

**Expected behavior**

**Environment:**
 - Infrastructure: [e.g., Kubernetes, bare-metal, laptop]
 - Deployment tool: [e.g., helm, jsonnet]

**Screenshots, Promtail config, or terminal output**
If applicable, add any output to help explain your problem.

bboreham:
Should be addressed by https://github.com/weaveworks/common/pull/247.
Also note https://github.com/prometheus/client_golang/pull/1091

John-Lin:
Same issue here when we upgrade to distributed Loki v2.6.1. distributors have same error message. 
We disable the [tracing feature](https://grafana.com/docs/loki/latest/configuration/#tracing) for workaround.

liguozhong:
this pr has fix this issue.
https://github.com/grafana/loki/pull/6859

periklis:
Isn't this address by #6957 already?

liguozhong:
6859>6957 , so I don't think 6957 has fixed this yet

periklis:
The issue is imho fixed by bumping the prometheus/client_golang to 1.13 with #6957. The changelog of client_golang (https://github.com/prometheus/client_golang/releases/tag/v1.13.0) mentions:
```
[BUGFIX] Raise exemplar labels limit from 64 to 128 bytes as specified in OpenMetrics spec. https://github.com/prometheus/client_golang/pull/1091
```

bboreham:
[...]

PR #6780, opened 2022-07-27
loki: Attach the panic recovery handler on all HTTP handlers

periklis:
Attaches the panic recovery handler on all HTTP handlers. 

Mitigates panics for  #6667 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -129,16 +129,19 @@ func (t *Loki) initServer() (services.Service, error) {
 	s := NewServerService(t.Server, servicesToWaitFor)
 
 	// Best effort to propagate the org ID from the start.
-	t.Server.HTTPServer.Handler = func(next http.Handler) http.Handler {
+	h := func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			if !t.Cfg.AuthEnabled {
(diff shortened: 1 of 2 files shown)


pkg/loki/modules.go
```diff
@@ -126,18 +126,21 @@ func (t *Loki) initServer() (services.Service, error) {
...
 				next.ServeHTTP(w, r.WithContext(user.InjectOrgID(r.Context(), "fake")))
 				return
 			}
+
 			_, ctx, _ := user.ExtractOrgIDFromHTTPRequest(r)
 			next.ServeHTTP(w, r.WithContext(ctx))
 		})
 	}(t.Server.HTTPServer.Handler)
 
+	// Best effort to propagate the org ID from the start.
```
kavirajk, line 141: Doesn't this comment (`Best effort to propagate the org ID from the start.`) belongs to handler `h`?

kavirajk (commented):
thanks @periklis . LGTM. with one minor clarification.

pkg/loki/modules.go
periklis: @kavirajk Move the comment above to `h` handler, should be fine now.

kavirajk (approved):
Thanks @periklis LGTM

PR #6784, opened 2022-07-27
chore(groupcache): disable hotcache for chunks

kavirajk:

chore(groupcache): disable hotcache for chunks

Fixes #<issue number>

Hot cache(most accessed keys:values) in groupcache are replicated in all the nodes. Doing so for chunks cache (biggest of all) can cause huge memory consumption on all the peers.

So disable it on chunk cache. It's fine for results and index cache.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/storage/chunk/cache/groupcache.go
+++ b/pkg/storage/chunk/cache/groupcache.go
@@ -169,15 +169,16 @@ func (c *GroupCache) Stats() *groupcache.Stats {
 }
 
 type group struct {
-	cache         *groupcache.Group
-	logger        log.Logger
-	wg            *sync.WaitGroup
-	cacheType     stats.CacheType
-	fetchDuration prometheus.Observer
-	storeDuration prometheus.Observer
+	cache           *groupcache.Group
+	logger          log.Logger
+	wg              *sync.WaitGroup
+	cacheType       stats.CacheType
+	fetchDuration   prometheus.Observer
+	storeDuration   prometheus.Observer
+	hotCacheEnabled bool
 }
 
-func (c *GroupCache) NewGroup(name string, ct stats.CacheType) Cache {
+func (c *GroupCache) NewGroup(name string, ct stats.CacheType, hotCacheEnabled bool) Cache {
 	// Return a known error on miss to track which keys need to be inserted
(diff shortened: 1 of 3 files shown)


pkg/storage/chunk/cache/groupcache.go
```diff
@@ -243,7 +245,7 @@ func (c *group) Store(ctx context.Context, keys []string, values [][]byte) error
 
 	var lastErr error
 	for i, key := range keys {
-		if err := c.cache.Set(ctx, key, values[i], time.Time{}, false); err != nil {
+		if err := c.cache.Set(ctx, key, values[i], time.Time{}, c.hotCacheEnabled); err != nil {
```
dannykopping, line 248: I think the simpler way here will be to check if `c.GetCacheType() != stats.ChunkCache`

kavirajk: thought about that but. But would be easier if we want to experiment it with other types of cache as well. Also to benchmark same cache with and without hotcache for example.

dannykopping: Let's start simple, and make it configurable if we need to.

The config option `hotCacheEnabled` is not really accurate because hot cache is still enabled either way, this setting here just controls if we write directly into hot cache

kavirajk: Sure. updated.

pkg/storage/chunk/cache/groupcache.go
```diff
@@ -243,7 +243,7 @@ func (c *group) Store(ctx context.Context, keys []string, values [][]byte) error
 
 	var lastErr error
 	for i, key := range keys {
-		if err := c.cache.Set(ctx, key, values[i], time.Time{}, false); err != nil {
+		if err := c.cache.Set(ctx, key, values[i], time.Time{}, c.cacheType != stats.ChunkCache); err != nil {
```
dannykopping, line 246: I think it'd be safer to use the `GetCacheType()` getter here since we wrap cache implementations

kavirajk: updated.

dannykopping (approved):
LGTM

Issue #6786, opened 2022-07-27
Make it possible to do Per-cache configuration for group cache


09jvilla:
The user should be able to configure different sizes for chunks cache and results cache when selecting the in-process cache (I'm intentionally not using the word `groupcache`; see https://github.com/grafana/loki/issues/6785) as their caching technology.

kavirajk:
https://github.com/grafana/loki/pull/6793

PR #6793, opened 2022-07-28
chore(groupcache): Support per cache group capacity

kavirajk:

Support per cache group capacity. 

With this individual group in groupcache can have own memory threshold.

Fixes # https://github.com/grafana/loki/issues/6786

1. Will add per group ballast config in separate PR after testing
2. Will replace `/groupcache/` with generic name also in different PR.

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -173,9 +173,11 @@ func (t *Loki) initGroupcache() (_ services.Service, err error) {
 	)
 	t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.GroupCache = gc.NewGroup(
 		t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.Prefix+"groupcache",
-		&t.Cfg.ChunkStoreConfig.ChunkCacheConfig.Groupcache,
+		&t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.Groupcache,
 		stats.ResultCache,
 	)
+
+	// The index cache generates too much traffic to be used. Make it a fifo cache
 	t.Cfg.StorageConfig.IndexQueriesCacheConfig.EnableFifoCache = true
 	t.Cfg.StorageConfig.IndexQueriesCacheConfig.Fifocache.MaxSizeBytes = fmt.Sprint(t.Cfg.Common.GroupCacheConfig.CapacityMB * 1e6)
 
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -168,12 +168,12 @@ func (t *Loki) initGroupcache() (_ services.Service, err error) {
 
 	t.Cfg.ChunkStoreConfig.ChunkCacheConfig.GroupCache = gc.NewGroup(
 		t.Cfg.ChunkStoreConfig.ChunkCacheConfig.Prefix+"groupcache",
-		&t.Cfg.ChunkStoreConfig.ChunkCacheConfig.Groupcache,
+		&t.Cfg.ChunkStoreConfig.ChunkCacheConfig.GroupCacheConfig,
 		stats.ChunkCache,
 	)
 	t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.GroupCache = gc.NewGroup(
 		t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.Prefix+"groupcache",
-		&t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.Groupcache,
+		&t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.GroupCacheConfig,
 		stats.ResultCache,
 	)
 
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -166,10 +166,16 @@ func (t *Loki) initGroupcache() (_ services.Service, err error) {
 		return nil, err
 	}
 
-	t.Cfg.ChunkStoreConfig.ChunkCacheConfig.GroupCache = gc.NewGroup(t.Cfg.ChunkStoreConfig.ChunkCacheConfig.Prefix+"groupcache", stats.ChunkCache)
-	t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.GroupCache = gc.NewGroup(t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.Prefix+"groupcache", stats.ResultCache)
-
-	// The index cache generates too much traffic to be used. Make it a fifo cache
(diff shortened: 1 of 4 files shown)


pkg/loki/modules.go
```diff
@@ -166,10 +166,16 @@ func (t *Loki) initGroupcache() (_ services.Service, err error) {
...
-
-	// The index cache generates too much traffic to be used. Make it a fifo cache
+	t.Cfg.ChunkStoreConfig.ChunkCacheConfig.GroupCache = gc.NewGroup(
+		t.Cfg.ChunkStoreConfig.ChunkCacheConfig.Prefix+"groupcache",
+		&t.Cfg.ChunkStoreConfig.ChunkCacheConfig.Groupcache,
+		stats.ChunkCache,
+	)
+	t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.GroupCache = gc.NewGroup(
+		t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.Prefix+"groupcache",
+		&t.Cfg.ChunkStoreConfig.ChunkCacheConfig.Groupcache,
```
dannykopping, line 176: Why the chunk cache config for the results cache?

pkg/storage/chunk/cache/cache.go
```diff
@@ -35,6 +35,7 @@ type Config struct {
 	MemcacheClient MemcachedClientConfig `yaml:"memcached_client"`
 	Redis          RedisConfig           `yaml:"redis"`
 	Fifocache      FifoCacheConfig       `yaml:"fifocache"`
+	Groupcache     GroupConfig           `yaml:"groupcache"` // TODO(kavi): consider resolving conflict with `GroupCache` field in same struct.
```
dannykopping, line 38: Is this TODO still relevant?

pkg/storage/chunk/cache/groupcache.go
```diff
@@ -69,6 +69,11 @@ type GroupCacheConfig struct {
 	Cache Cache `yaml:"-"`
 }
 
+// GroupGconfig represents config per Group.
```
dannykopping, line 72:
```suggestion
// GroupConfig represents config per Group.
```

pkg/storage/chunk/cache/groupcache.go
```diff
@@ -198,15 +203,19 @@ func (c *GroupCache) Stats() *groupcache.Stats {
...
-	logger        log.Logger
-	wg            *sync.WaitGroup
-	cacheType     stats.CacheType
+	cache     *groupcache.Group
+	logger    log.Logger
+	wg        *sync.WaitGroup
+	cacheType stats.CacheType
+
+	// cacheBytes represents maxSize (in bytes) this group can grow.
+	cacheBytes int64 // TODO(kavi): expose it as _info metrics later?
```
dannykopping, line 212: I think this metric would belong with the other groupcache metrics, but yes i love the idea of making this a metric

pkg/loki/modules.go
```diff
@@ -166,10 +166,16 @@ func (t *Loki) initGroupcache() (_ services.Service, err error) {
 		return nil, err
 	}
 
-	t.Cfg.ChunkStoreConfig.ChunkCacheConfig.GroupCache = gc.NewGroup(t.Cfg.ChunkStoreConfig.ChunkCacheConfig.Prefix+"groupcache", stats.ChunkCache)
-	t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.GroupCache = gc.NewGroup(t.Cfg.QueryRange.ResultsCacheConfig.CacheConfig.Prefix+"groupcache", stats.ResultCache)
-
-	// The index cache generates too much traffic to be used. Make it a fifo cache
```
dannykopping, line 172: Let's not lose this comment

pkg/storage/chunk/cache/groupcache.go
```diff
@@ -215,6 +224,11 @@ func (c *GroupCache) NewGroup(name string, ct stats.CacheType) Cache {
 	c.wg.Add(1)
 	c.startWaitingForClose()
 
+	cap := c.capacity
+	if cfg.CapacityMB != 0 {
+		cap = cfg.CapacityMB * 1e6 // MB into bytes
+	}
```
dannykopping, line 230: Hhmm, do we need to keep the `capacity` field here?

kavirajk: I would think so. `c.capacity` comes from global groupcache config (when we create single groupcache instance from `common.Groupcache` config).

`Groupcache.capacity` -> global capacity
`group.cacheBytes` -> local to particular group.

We need both.

does that answers your question? or is it something else?. 

nit: I can make both consistent name as `cacheBytes` (in sync with how it's used in original google/mailgun Groupcache package).

pkg/loki/modules.go
kavirajk: good catch. Thanks :+1: . Fixed. (was bit haste while resolving merge conflict with Travis PR)

pkg/storage/chunk/cache/cache.go
kavirajk: Yes. we have two fields on this struct.

`Groupcache` -> GroupConfig (named it this way to be consistent with fifocache)
`GroupCache` -> Instance of groupcache we populate in `init modules.go`. (not exposed `yaml:"-"`)

Any suggestions welcome!

pkg/storage/chunk/cache/groupcache.go
kavirajk: Fixed :+1:

pkg/storage/chunk/cache/cache.go
dannykopping: Thanks for the explanation. I think we should resolve this somehow before merging

pkg/storage/chunk/cache/groupcache.go
dannykopping: Ah gotcha :+1: 
Thanks

pkg/storage/chunk/cache/cache.go
kavirajk: Fixed it in https://github.com/grafana/loki/pull/6793/commits/d7df061d90ae4ee5cb61963d3c781e152a1e7ebd

dannykopping (approved):
LGTM with a non-blocking nit

kavirajk:
> <nit>Can we set config.CapacityMB * 1e6 in a single place?</nit>

I'm afraid we can't. Also both are in-fact different fields.  One is for global groupcache and other is for Local Group, both getting initialized at different places.

PR #6796, opened 2022-07-28
Fix Promtail dashboard

DylanGuedes:
Fix Promtail production dashboard/panels. Error message: `bad_data: 1:40: parse error: unexpected character inside braces: '~'`.


salvacorts (approved):
Thank you!

PR #6801, opened 2022-07-29
Add operational guide for autoscaling

salvacorts:
This PR adds an operational guide for autoscaling.
Right now it only contains a section for the queriers but the idea is to add more sections as we autoscale other Loki components. 

Fixes #<issue number>

I tested that the new docs look fine by running `make docs`.

**Checklist**
- [X] Documentation added
- [ ] Tests updated


docs/sources/operations/autoscaling.md
```diff
@@ -0,0 +1,135 @@
...
+- The scheduler queue size.
+- The queries running in the queriers.
+
+### Scaling metric
+
+The Query Scheduler exposes the `cortex_query_scheduler_inflight_requests` metric.
+It tracks the number of queued queries plus the number of queries currently running in the querier workers.
+The following query is useful to scale queriers based on the inflight requests.
+
+```
```
salvacorts, line 33: Don't we have syntax highlighting for PromQL queries?

dannykopping, line 33:
```suggestion
```promql
```

dannykopping: I think this may work

dannykopping (approved):
Super!

docs/sources/operations/autoscaling.md
salvacorts: Unfortunately, that doesn't seem to work 😞 
Neither I could see any example with syntax highlighting.

dannykopping: How are you testing it?

salvacorts: Move into the `docs` dir and run `make docs`. It creates a local server with the docs. Then go to http://localhost:3002/docs/loki/latest/operations/autoscaling/

dannykopping: Hhmm I wonder if that exactly matches the website's Hugo settings; I doubt it?
In any case, let's land this and see what happens. I'm reasonably confident we support promql because of Mimir's docs.

salvacorts: Done.

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,139 @@
...
+number of inflight requests 75% of the time in the last seven days, targeting a 75% utilization of our queriers.
+So if we use 6 workers per querier, we will use the following query:
+
+```promql
+clamp_min(ceil(
+    avg(
+        avg_over_time(cortex_query_scheduler_inflight_requests{namespace="loki-cluster", quantile="0.75"}[7d])
+    ) / scalar(floor(vector(6 * 0.75)))
+), 1)
+```
```
Raskyld, line 70: If I understand correctly, you want that 75% of the time, the average number of request in the system (`L` in queuing theory) is equal to 75% the number of workers per querier, hence, there would be no query in the queue (they would be addressed immediately).
By evaluating the ratio between the **current** `L` (thanks to the inflight metric) and our **target** `L = 0.75 * nb_worker` you know how much times we need to scale the **current** number of querier to reach our **target**.

Side note, this is only meaningful if `lambda` (the request arrival rate) converge over time which is, I think, why you recommend using the average on a whole week.

So the result of this PromQL query isn't actually the minimum number of queriers necessary but the scale factor that we need to apply to the **current** number of queriers in the cluster when we evaluated the inflight metric.

**tl;dr** If I understood your methodology correctly, you still need to add `* nb_querier` for this query to give the minimal number of querier required for the cluster to instantly handle a query in 75% of the time.
**Or** you can remove the sentence "running **_at least one_** queriers" for "running only one querier" as it means `nb_querier` would be 1, but it's not practical as most users will already have a running cluster with a specific number of querier to maintain the cluster in an equilibrium state.

Please, let met know if I am wrong in my understanding of your approach and thanks a lot for your work, it saved me a bunch of time ! :rocket:

KMiller-Grafana (changes requested):
A clear and nice writeup of the autoscaling.  Thank you.  The quantity of suggestions looks like a lot, but I've made the suggestions to fix typos and better follow our documentation style guide.

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,139 @@
...
+we have designed a set of resources to help you autoscale your Loki queriers.
+
+## Prerequisites
+
+You need to run Loki in Kubernetes as a set of microservices.
+
+We recommend using [KEDA (*Kubernetes-based Event Driven Autoscaler*)](https://keda.sh/) to configure autoscaling
+based on Prometheus metrics. Refer to the [KEDA documentation](https://keda.sh/docs/latest/deploy) to learn more
+about setting up KEDA in your Kubernetes cluster.
+
```
Raskyld, line 20: I think you forgot to mention in the prerequisites that users need to run the v2 of query frontend (as the metrics used for the scaling are only exposed from the query-scheduler)

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,139 @@
...
+## Scaling metric
+
+Queriers pull queries from the query scheduler queue and process them on the querier workers. Therefore, it makes sense to scale based on:
+
+- The scheduler queue size.
+- The queries running in the queriers.
+
+The Query Scheduler exposes the `cortex_query_scheduler_inflight_requests` metric.
+It tracks the number of queued queries plus the number of queries currently running in the querier workers.
+The following query is useful to scale queriers based on the inflight requests.
```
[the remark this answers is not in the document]
salvacorts: > I'm suggesting that we use "query-scheduler," as most of the Mimir docs use a hyphenated form

That makes perfect sense.

>  Is it tracking the sum of the number of queued queries and the number of queries? Or is it tracking 2 distinct items?

It tracks the sum.

> "It tracks the sum of queued queries plus the number of queries currently running in the querier workers."

Sounds good to me!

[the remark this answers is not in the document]
salvacorts: Wrt to the suggested change (line 20)
```
Refer to the [Deploying KEDA] to learn more.
```

Should be 
```
Refer to the [Deploying KEDA section] to learn more.
```

wdyt?

docs/sources/operations/autoscaling_queriers.md
salvacorts: > [...] you want that 75% of the time, the average number of request in the system (L in queuing theory) is equal to 75% the number of workers per querier

Not quite. Using the 75 percentile (75% of the time) is to avoid taking short-lasting spikes into account.

> If I understood your methodology correctly, you still need to add * nb_querier for this query

I don't think so. We divide by the number of workers we aim to use so we have extra room for spikes.
E.g. if the avg is 10 and we aim to use 2 workers out of the 4 workers available per querier, this query would return 5 (the number of queriers). Because we'd be using 50% of the workers (2 out of 4), we have room for twice the avg workload.

docs/sources/operations/autoscaling_queriers.md
salvacorts: > the v2 of query frontend

@Raskyld Can you elaborate on what you mean by "v2" here?

@KMiller-Grafana should we add versioning requisites in these docs? IIUC, docs are coupled with the version (e.g. [docs for 2.4.x][1] and [docs for latest][2]).

[1]: https://grafana.com/docs/loki/v2.4.x/
[2]: https://grafana.com/docs/loki/latest/

docs/sources/operations/autoscaling_queriers.md
Raskyld: > @Raskyld Thanks for your comment! Sorry for the delay, I took some vacations.

No problem! Hope you enjoyed it :)

> Not quite. Using the 75 percentile (75% of the time) is to avoid taking short-lasting spikes into account.

Gotcha!

> I don't think so. We divide by the number of workers we aim to use so we have extra room for spikes. E.g. if the avg is 10 and we aim to use 2 workers out of the 4 workers available per querier, this query would return 5 (the number of queriers). Because we'd be using 50% of the workers (2 out of 4), we have room for twice the avg workload.

After two weeks and some sleep I realized my misunderstanding come from the metrics used.
Just for the record, and to be sure I understand correctly :

`cortex_query_scheduler_inflight_requests` observes (every `250ms`) the number of queries being processed by a querier worker or queued.
So you want the minimal number of worker = the average number of inflight requests 75% of the time ; then you divide the average by the number of active worker you want (e.g. 75% of all the workers) to know how much querier you need.

Am I correct? :pray: 

I have another concern : It is possible to replicate the `query-scheduler` as long as there is less or as much as instance of scheduler than instance of querier worker (i.e. `-querier.max-concurrent`) ; this means, you could have multiple series of this metric right? If so, shouldn't you sum the averages before dividing?

docs/sources/operations/autoscaling_queriers.md
Raskyld: Of course : there is two implementations of the `query-frontend` ([namely v1, v2](https://github.com/grafana/loki/tree/main/pkg/lokifrontend/frontend)) the first one has a builtin query queue and the second one use `scheduler_worker` (see [here](https://github.com/grafana/loki/blob/main/pkg/lokifrontend/frontend/v2/frontend_scheduler_worker.go)) to forward queries to the schedulers. 

*tl; dr* If you use the query schedulers you're using query-frontend v2 otherwise you fallback to the v1.

The v1 typically doesn't have the metric you're using for the auto-scaling, hence, it would be wise to add the query scheduler setup in the requirements.

If you need more information do not hesitate !

docs/sources/operations/autoscaling_queriers.md
salvacorts: > [...] the average number of inflight requests 75% of the time ; then you divide the average by the number of active worker you want (e.g. 75% of all the workers) to know how much querier you need.
> Am I correct? 

Yes! 

> you could have multiple series of this metric right? If so, shouldn't you sum the averages before dividing?

The `avg` is an aggregation function that will average all the series. To have different averages for each series you would do `avg by (instance)`.

There is an example [in the docs](https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators) that shows this:
> If the metric http_requests_total had time series that fan out by application, instance, and group labels, we could calculate the total number of seen HTTP requests per application and group over all instances via:
> `sum without (instance) (http_requests_total)`
> Which is equivalent to:
> `sum by (application, group) (http_requests_total)`
> If we are just interested in the total of HTTP requests we have seen in all applications, we could simply write:
> `sum(http_requests_total)`

docs/sources/operations/autoscaling_queriers.md
salvacorts: > there is two implementations of the query-frontend ([namely v1, v2](https://github.com/grafana/loki/tree/main/pkg/lokifrontend/frontend))

TIL. I'll add a note to the prerequisites. Thank you! 🚀

Raskyld: You're very welcome! Thanks you :D

docs/sources/operations/autoscaling_queriers.md
Raskyld: Guess it's my turn to say TIL!
`avg_over_time` uses an aggregation operator but does **not** aggregate while `avg` alone does!

Raskyld: Thanks for all those explanations :ok_hand:

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
+---
+title: Autoscaling Loki queriers
+menuTitle: Autoscaling queriers
+description: Kubernetes deployments of a microservices mode Loki cluster can use KEDA to autoscale the quantity of queriers.
+weight: 30
+---
+
+# Autoscaling Loki queriers
+
+A microservices deployment of a Loki cluster that runs on Kubernetes will typically handle a
```
chri2547, line 10:
```suggestion
A microservices deployment of a Loki cluster that runs on Kubernetes typically handles a
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+workload that varies throughout the day.
+To make Loki easier to operate and optimize the cost of running Loki at scale,
+we have designed a set of resources to help you autoscale your Loki queriers.
+
+## Prerequisites
+
+You need to run Loki in Kubernetes as a set of microservices. You need to use the query-scheduler.
+
+We recommend using [Kubernetes Event-Driven Autoscaling (KEDA)](https://keda.sh/) to configure autoscaling
+based on Prometheus metrics. Refer to the [Deploying KEDA](https://keda.sh/docs/latest/deploy) section to learn more
```
chri2547, line 20:
```suggestion
based on Prometheus metrics. Refer to [Deploying KEDA](https://keda.sh/docs/latest/deploy) to learn more
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+about setting up KEDA in your Kubernetes cluster.
+
+## Scaling metric
+
+Queriers pull queries from the query-scheduler queue and process them on the querier workers. Therefore, it makes sense to scale based on:
```
chri2547, line 25:
```suggestion
Because queriers pull queries from the query-scheduler queue and process them on the querier workers, you should scale metrics based on:
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+
+```promql
+sum(
+  max_over_time(
+    cortex_query_scheduler_inflight_requests{namespace="loki-cluster", quantile="<Q>"}[<R>]
+  )
+)
+```
+
+We can use the quantile (Q) and the range (R) parameters to fine-tune the metric.
```
chri2547, line 42:
```suggestion
Use the quantile (Q) and the range (R) parameters to fine-tune the metric.
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+The higher Q is, the more sensitive the metric is to short-lasting spikes.
+As R increases, we reduce the variation over time in the metric.
```
chri2547, line 44: I'm confused about 'we' in this sentence. Does the system automatically reduce variation?

docs/sources/operations/autoscaling_queriers.md
chri2547, line 42: My suggestion below assumes that the user fine-tunes the metric. I'm not sure if my assumption is correct.

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+A higher value for R helps avoid the autoscaler from modifying the number of replicas too frequently.
```
chri2547, line 45:
```suggestion
A higher R-value helps avoid the autoscaler from modifying the number of replicas too frequently.
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+
+In our experience, we have found that a Q of 0.75 and an R of 2 minutes work well.
+Our recommendation is to adjust these values according to your workload.
```
chri2547, line 48:
```suggestion
You can adjust these values according to your workload.
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Cluster capacity planning
+
+When configuring KEDA to scale your Loki queriers, we need to configure:
```
chri2547, line 52: Confused by we...is it truly a 'we' as in Grafana Labs is configuring KEDA? Or is this something the user does? I'll be able to add a suggestion after I get clarity.

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+
+- The threshold for scaling up and down
+- The scale down stabilization period
+- The minimum and the maximum number of queriers
+
+Querier workers process queries from the queue. Each of our queriers can be configured to run several workers.
```
chri2547, line 58:
```suggestion
Querier workers process queries from the queue. You can configure each Loki querier to run several workers.
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+Our recommendation is to aim for using 75% of those workers, to leave some workforce headroom for workload spikes.
```
chri2547, line 59:
```suggestion
To reserve workforce headroom to address workload spikes, do not use more than 75% of the workers. 
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+Therefore, if we configure our queriers to run 6 workers, we will set a threshold of `floor(0.75 * 6) = 4`.
```
chri2547, line 60:
```suggestion
For example, if you configure the Loki queriers to run 6 workers, set a threshold of `floor(0.75 * 6) = 4`.
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+
+Our recommendation for the minimum number of queriers is to run at least one querier, and to look at the average
+number of inflight requests 75% of the time in the last seven days, targeting a 75% utilization of the queriers.
```
chri2547, line 63: I'm not sure what these lines mean. Is the user attempting to determine the minimum number of queriers to run?

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+clamp_min(ceil(
+    avg(
+        avg_over_time(cortex_query_scheduler_inflight_requests{namespace="loki-cluster", quantile="0.75"}[7d])
+    ) / scalar(floor(vector(6 * 0.75)))
+), 1)
+```
+
+Our recommendation for the maximum number of queriers is running the number of queriers that would be required
+to process all the inflight requests you had 50% of the time during the last seven days.
+As for the previous example, if each querier runs 6 workers, we divide the inflight requests by 6.
```
chri2547, line 76:
```suggestion
As for the previous example, if each querier runs 6 workers, divide the inflight requests by 6.
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+
+```promql
+clamp_min(ceil(
+    avg(
+        avg_over_time(cortex_query_scheduler_inflight_requests{namespace="loki-cluster", quantile="0.75"}[7d])
+    ) / scalar(floor(vector(6 * 0.75)))
+), 1)
+```
+
+Our recommendation for the maximum number of queriers is running the number of queriers that would be required
```
chri2547, line 74: Similar question here, is the user trying to determine the max number of queriers?

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+
+```promql
+ceil(
+    max(
+        max_over_time(cortex_query_scheduler_inflight_requests{namespace="loki-cluster", quantile="0.5"}[7d])
+    ) / 6
+)
+```
+
+Finally, to minimize the scenario where we scale up shortly after scaling down, we recommend setting
```
chri2547, line 87:
```suggestion
To minimize the scenario where Loki scales up shortly after scaling down, we recommend setting
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+  advanced:
+    horizontalPodAutoscalerConfig:
+      behavior:
+        scaleDown:
+          stabilizationWindowSeconds: 1800
+```
+
+## Prometheus alerting when at capacity
+
+Because the configured maximum may not be enough, a Prometheus alert can identify
```
chri2547, line 127:
```suggestion
Because the configured maximum might not be sufficient, a Prometheus alert can identify
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+when the quantity of queriers has been at its configured maximum for an extended time. This example specifies 3 hours as the extended time:
```
chri2547, line 128:
```suggestion
when the quantity of queriers has been at its configured maximum for an extended time. The following example specifies three hours (`3h`) as the extended time:
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+Finally, to minimize the scenario where we scale up shortly after scaling down, we recommend setting
+a stabilization window for scaling down.
+
+
+### KEDA configuration
+
+This [KEDA ScaledObject](https://keda.sh/docs/latest/concepts/scaling-deployments/) example configures autoscaling
+for the querier deployment in the `loki-cluster` namespace.
+We set the minimum number of replicas to 10 and the maximum number of replicas to 50.
+Since each querier runs 6 workers, aiming to use 75% of those workers, we set the threshold to 4.
```
chri2547, line 96:
```suggestion
Because each querier runs 6 workers, aiming to use 75% of those workers, the threshold is set to 4.
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+
+Finally, to minimize the scenario where we scale up shortly after scaling down, we recommend setting
+a stabilization window for scaling down.
+
+
+### KEDA configuration
+
+This [KEDA ScaledObject](https://keda.sh/docs/latest/concepts/scaling-deployments/) example configures autoscaling
+for the querier deployment in the `loki-cluster` namespace.
+We set the minimum number of replicas to 10 and the maximum number of replicas to 50.
```
chri2547, line 95:
```suggestion
The example shows the minimum number of replicas set to 10 and the maximum number of replicas set to 50.
```

chri2547 (changes requested):
Suggestions and few questions. Thanks!

docs/sources/operations/autoscaling_queriers.md
salvacorts: That's correct.

docs/sources/operations/autoscaling_queriers.md
salvacorts: No. The user can modify Q and R to fine-tune the metric.

docs/sources/operations/autoscaling_queriers.md
salvacorts: "Our recommendation is to..." wording comes from feedback from Karen. I'm ok with reverting that, but just a note.

docs/sources/operations/autoscaling_queriers.md
salvacorts: It's something the user does.

docs/sources/operations/autoscaling_queriers.md
salvacorts: I would rather write:
```
To reserve workforce headroom to address workload spikes, our recommendation is to not use more than 75% of the workers.
```

docs/sources/operations/autoscaling_queriers.md
salvacorts: Correct. These are recommendations about how to estimate the minimum number of queriers to run.

docs/sources/operations/autoscaling_queriers.md
salvacorts: Yes

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+sum(
+  max_over_time(
+    cortex_query_scheduler_inflight_requests{namespace="loki-cluster", quantile="<Q>"}[<R>]
+  )
+)
+```
+
+Use the quantile (Q) and the range (R) parameters to fine-tune the metric.
+The higher Q is, the more sensitive the metric is to short-lasting spikes.
+As R increases, we reduce the variation over time in the metric.
```
chri2547, line 44:
```suggestion
As R increases, you can reduce the variation over time in the metric.
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+A higher R-value helps avoid the autoscaler from modifying the number of replicas too frequently.
+
+In our experience, we have found that a Q of 0.75 and an R of 2 minutes work well.
+You can adjust these values according to your workload.
+
+## Cluster capacity planning
+
+When configuring KEDA to scale your Loki queriers, we need to configure:
```
chri2547, line 52:
```suggestion
To scale the Loki queries, you configure the following settings:
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+- The threshold for scaling up and down
+- The scale down stabilization period
+- The minimum and the maximum number of queriers
+
+Querier workers process queries from the queue. You can configure each Loki querier to run several workers.
+To reserve workforce headroom to address workload spikes, our recommendation is not to use more than 75% of the workers.
+For example, if you configure the Loki queriers to run 6 workers, set a threshold of `floor(0.75 * 6) = 4`.
+
+Our recommendation for the minimum number of queriers is to run at least one querier, and to look at the average
+number of inflight requests 75% of the time in the last seven days, targeting a 75% utilization of the queriers.
```
chri2547, line 63:
```suggestion
To determine the minimum number of queries that you should run, run at least one querier for seven days and determine the average
number of inflight requests the system processes 75% of the time. The target utilization of the queries is 75%.
```

docs/sources/operations/autoscaling_queriers.md
```diff
@@ -0,0 +1,141 @@
...
+```promql
+clamp_min(ceil(
+    avg(
+        avg_over_time(cortex_query_scheduler_inflight_requests{namespace="loki-cluster", quantile="0.75"}[7d])
+    ) / scalar(floor(vector(6 * 0.75)))
+), 1)
+```
+
+Our recommendation for the maximum number of queriers is running the number of queriers that would be required
+to process all the inflight requests you had 50% of the time during the last seven days.
```
chri2547, line 75: The maximum number of queriers to run is equal to the number of queriers required to process 50% of the inflight requests during a seven-day timespan.

chri2547 (approved):
A few more suggestions. LGTM!

docs/sources/operations/autoscaling_queriers.md
salvacorts: > run at least one querier for seven days

This doesn't sound correct to me. Our recommendation is to run at least one querier (hence the `clamp_min`). but running one querier for seven days.

docs/sources/operations/autoscaling_queriers.md
salvacorts: This is not correct: "50% of the inflight requests" is not the same as "all the inflight requests you had 50% of the time". The latter refers to a quantile; the 50 percentile is the median but not half.

docs/sources/operations/autoscaling_queriers.md
chri2547: @salvacorts , updated the suggestion. Does that wording work better?

docs/sources/operations/autoscaling_queriers.md
chri2547: Gotcha, ok what about "The maximum number of queriers to run is equal to the number of queriers required to process all inflight requests 50% of the time during a seven-day timespan."

does that work?

docs/sources/operations/autoscaling_queriers.md
salvacorts: Works perfectly! Thank you :)

docs/sources/operations/autoscaling_queriers.md
salvacorts: That sounds perfect. Committed the changes. Thanks!

PR #6802, opened 2022-07-29
Run `loki -version` without needing a config defined

### LINKED ISSUE #6770: loki -version should work without config file
**Describe the bug**
`loki -version` currently requires a configuration file to be passed or found at default location. This is unexpected, since it does not use the config file and causes troubles when deployments first check version to then install proper configuration file

**To Reproduce**
1. Make sure no config file is at the default location.
2. `loki -version`

**Expected behavior**
Print version information.

```
# /usr/local/bin/loki  -version
failed parsing config: config.yaml,config/config.yaml does not exist, set config.file for custom config path
```

gyanesh-m:
It doesn't throw error during version check if config file isn't present.

Fixes #6770 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/cmd/loki/main.go
+++ b/cmd/loki/main.go
@@ -24,13 +24,13 @@ import (
 func main() {
 	var config loki.ConfigWrapper
 
+	if config.PrintVersion || len(os.Args[1:]) == 1 && os.Args[1] == "version" {
+		fmt.Println(version.Print("loki"))
+		os.Exit(0)
+	}
 	if err := cfg.DynamicUnmarshal(&config, os.Args[1:], flag.CommandLine); err != nil {
 		fmt.Fprintf(os.Stderr, "failed parsing config: %v\n", err)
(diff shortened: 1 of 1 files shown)


cmd/loki/main.go
```diff
@@ -24,14 +24,14 @@ import (
 func main() {
 	var config loki.ConfigWrapper
 
+	if config.PrintVersion || len(os.Args[1:]) == 1 && os.Args[1] == "version" {
```
dannykopping, line 27:
```suggestion
	if len(os.Args) >= 1 && os.Args[1] == "-version" {
```

`config.PrintVersion` will never be set, and the arg should be `-version`

dannykopping (changes requested):

dannykopping (approved):
Thanks for your contribution @gyanesh-m :+1:

Issue #6775, opened 2022-07-26
Compactor falls hopelessly behind on busy cluster


afayngelerindbx:
**Describe the bug**
We ran into an issue where our compactor has fallen behind so badly that it cannot catch up.
<img width="836" alt="Screen Shot 2022-07-26 at 3 25 51 PM" src="https://user-images.githubusercontent.com/38990103/181094936-a5fd9c5e-56af-4bd3-8931-fe6bb93949ac.png">

Looking at the code, I noticed that if a compactor cannot complete a successful compaction, it must start over at the beginning. Moreover,  it will attempt to compact the same files without giving preference to newer data. Since newer data accumulates while the older is being cleared, compaction falls behind irreparably.

**To Reproduce**
I suspect the only thing needed to reproduce this is a large enough cluster and a compactor that doesn't run for a day or so. This can also occur organically if there are S3 connectivity issues etc. 

**Expected behavior**
It would be good if compactors had a way to recover from this behavior. I hacked up a version that I am currently testing out in our environment that implements the following mitigation mechanisms:

- The compactor is forced to stop compacting roughly at the end of the compaction interval and save partially compacted state to object store
- Each compaction prioritizes newer tables so if there's a backlog of uncompacted index files in previous days' tables, this backlog does not prevent the compactor from making progress on new indexes.
- I've also added flags to limit which tables are worked on by a particular compactor. This will allow me to deploy multiple compactor rings: one to work on recent data with a short interval setting and another to work on historic back logs with a larger time out.

If there are better mitigations that can be implemented, please, let me know.

**Environment:**
 - Infrastructure: baremetal
 - Deployment tool: custom
 
**Screenshots, Promtail config, or terminal output**
If applicable, add any output to help explain your problem.

JStickler:
Closing as stale. This issue was opened against a version of Loki that is no longer supported. If this issues still exists in more recent versions of Loki, please leave a comment and we can reopen the issue.

PR #6804, opened 2022-07-29
compactor: prioritize newer tables when compacting. add operator flags to compac…

afayngelerindbx:
…t time ranges of tables

Currently, the order in which the compactor picks tables for compactions is non-deterministic. Usually, the latest table is not picked to be compacted first. This means that when a backlog of compactions is generated, the most recent table is not compacted and a backlog in the most recent table is formed, perpetuating the problem. This change makes it more likely that newly generated index files are compacted first, curbing backlog growth.

Helped in resolving #6775 

**Checklist**
I'd like some feedback here about which of the following steps are necessary for this diff e.g. do the new flags need docs or changelog entries:

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/indexshipper/compactor/compactor.go
+++ b/pkg/storage/stores/indexshipper/compactor/compactor.go
@@ -85,9 +85,9 @@ type Config struct {
 	DeleteRequestCancelPeriod time.Duration   `yaml:"delete_request_cancel_period"`
 	MaxCompactionParallelism  int             `yaml:"max_compaction_parallelism"`
 	CompactorRing             util.RingConfig `yaml:"compactor_ring,omitempty"`
-	RunOnce                   bool            `yaml:"-"`
-	TablesToCompact           int             `yaml:"-"`
-	YoungestTableToCompact    int             `yaml:"-"`
+	RunOnce                   bool            `yaml:"_"`
+	TablesToCompact           int             `yaml:"tables_to_compact"`
+	SkipLatestNTables         int             `yaml:"skip_latest_n_tables"`
 
 	// Deprecated
 	DeletionMode string `yaml:"deletion_mode"`
@@ -114,7 +114,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 
 	cfg.CompactorRing.RegisterFlagsWithPrefix("boltdb.shipper.compactor.", "collectors/", f)
 	f.IntVar(&cfg.TablesToCompact, "boltdb.shipper.compactor.tables-to-compact", 0, "The number of most recent tables to compact in a single run. Default: all")
-	f.IntVar(&cfg.YoungestTableToCompact, "boltdb.shipper.compactor.youngest-table", 0, "Compact only tables older than the n-th youngest.")
+	f.IntVar(&cfg.SkipLatestNTables, "boltdb.shipper.compactor.skip-latest-n-tables", 0, "Skip compacting latest N tables")
 
 }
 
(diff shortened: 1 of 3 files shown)


pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -98,6 +101,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.DeletionMode, "boltdb.shipper.compactor.deletion-mode", "disabled", fmt.Sprintf("Deletion mode. Can be one of %v", strings.Join(deletion.AllModes(), "|")))
 	cfg.CompactorRing.RegisterFlagsWithPrefix("boltdb.shipper.compactor.", "collectors/", f)
 	f.BoolVar(&cfg.RunOnce, "boltdb.shipper.compactor.run-once", false, "Run the compactor one time to cleanup and compact index files only (no retention applied)")
+	f.IntVar(&cfg.TablesToCompact, "boltdb.shipper.compactor.tables-to-compact", 0, "The number of most recent tables to compact in a given run. Default: all")
+	f.IntVar(&cfg.YoungestTableToCompact, "boltdb.shipper.compactor.youngest-table", 0, "Compact only tables older than the n-th youngest.")
```
sandeepsukhani, line 105: What do you think about naming it like this? Makes it clearer. Any other suggestions are welcome.
```suggestion
	f.IntVar(&cfg.SkipLatestNTables, "boltdb.shipper.compactor.skip-latest-n-tables", 0, "Skip compacting latest N tables")
```

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -98,6 +101,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.DeletionMode, "boltdb.shipper.compactor.deletion-mode", "disabled", fmt.Sprintf("Deletion mode. Can be one of %v", strings.Join(deletion.AllModes(), "|")))
 	cfg.CompactorRing.RegisterFlagsWithPrefix("boltdb.shipper.compactor.", "collectors/", f)
 	f.BoolVar(&cfg.RunOnce, "boltdb.shipper.compactor.run-once", false, "Run the compactor one time to cleanup and compact index files only (no retention applied)")
+	f.IntVar(&cfg.TablesToCompact, "boltdb.shipper.compactor.tables-to-compact", 0, "The number of most recent tables to compact in a given run. Default: all")
```
sandeepsukhani, line 104:
```suggestion
	f.IntVar(&cfg.TablesToCompact, "boltdb.shipper.compactor.tables-to-compact", 0, "The number of most recent tables to compact in a single run. Default: all")
```

sandeepsukhani (commented):
Left some very minor nits otherwise, it LGTM!
The new flags need to be documented [here](https://github.com/grafana/loki/blob/965650740cb54d6a2cc5ee2f931ea01af44ac48e/docs/sources/configuration/_index.md?plain=1#L2065).

pkg/storage/stores/indexshipper/compactor/compactor.go
afayngelerindbx: Thought a bit about how to phrase this. I think your suggestion is better than anything I can come up with.

sandeepsukhani (approved):
LGTM! Thanks for taking care of the feedback!

PR #6806, opened 2022-07-29
Allow fifocache to be configured with groupcache

MasslessParticle:
We stopped the panic, but it made it so a user can't configure both fifocache and groupcache together. This solves the issue differentiating if groupcache is set for a particular cache based on `cache != nil` and whether it's globally configured via `cfg.GroupCacheEnabled`


kavirajk (approved):
Nice. Make sense! thanks @MasslessParticle

PR #6808, opened 2022-07-29
Add groupcache timeouts

MasslessParticle:
p97 latency on groupcache requests is ~30ms but it seems like there's a 100s timeout when things get stuck. This should aggressively kill connections when they get stuck so calls to the cache don't block queries.

`ReadIdleTimeout` is a heartbeat that must respond in `PingTimeout`. `WriteByteTimeout` is what it sounds like: a timeout on writes.


dannykopping (commented):
If we start seeing latency across the board, will we not start failing all cache requests with timeouts this short?

Side-note: we should include this in https://github.com/grafana/loki/issues/6786 as a line-item for making this configurable; I don't think these values will work for everyone.

MasslessParticle:
It's reasonable to make them configurable. In the case of broad latency, I think failing all cache calls is the intended solution because fetching from the cache blocks queries which causes all sorts of issues.

pkg/storage/chunk/cache/groupcache.go
```diff
@@ -92,6 +93,10 @@ func NewGroupCache(rm ringManager, config GroupCacheConfig, logger log.Logger, r
 	addr := fmt.Sprintf("http://%s", rm.Addr())
 	level.Info(logger).Log("msg", "groupcache local address set to", "addr", addr)
 
+	http2Transport.ReadIdleTimeout = config.HeartbeatInterval
+	http2Transport.PingTimeout = config.HeartbeatTimeout
+	http2Transport.WriteByteTimeout = config.WriteByteTimeout
```
dannykopping, line 98: Which of these would affect the read timeout?

kavirajk (approved):
LGTM :+1:

dannykopping (commented):
I think we should expose a metric (if we don't already) so we can know when these requests fail so we can alert on it when it reaches pathological levels.

If a call fails to the result cache, the [error is ignored](https://github.com/grafana/loki/blob/main/pkg/querier/queryrange/queryrangebase/results_cache.go#L552-L556) (which is probably ok, but bad for debugging) and then the result will have to be recomputed. Similarly, the [error is ignored](https://github.com/grafana/loki/blob/main/pkg/querier/queryrange/queryrangebase/results_cache.go#L594) when storing the computed result.

MasslessParticle:
I'm concerned there's not a way to catch a timeout. I'd expect an error to be of the form `unable to connect to: context cancelled` and logged. In that way this is consistent with our other cache implementations.

If we want to understand cache misses due to timeouts, I'd think that's another group of work across all the caches.

pkg/storage/chunk/cache/groupcache.go
MasslessParticle: `ReadIdleTimeout` and `PingTimout` combined form a sort of "read timeout".

`ReadIdleTimeout` sends a heartbeat when the connection goes idle. This ensures that we test the connection at least once every `ReadIdleTimeout`. 

`PingTimeout` is our tolerance for heartbeat responses.

Lets assume our connection has hung, `ReadIdleTimeout` is 1s, and `PingTimout` is 100ms:
1. No traffic occurs in 1s so we send a HeartBeat
2. No `ACK` is received in 100ms, terminate the connection

In the case where traffic is flowing normally, the connection doesn't send heartbeats because `ReadIdleTimeout` is never reached.

dannykopping (approved):
Let's try this out and revisit if we need to add more observability

PR #6809, opened 2022-07-29
[jsonnet] only set the consul kvstore configs if memberlist is not being used

cstyan:
This PR removes the configuration of consul/kvstore configs if memberlist is being used, which it is by default.

I can understand if we don't want to change this configuration/code to avoid potential configuration issues or confusion for OSS users, however for internal use we've run into some confusion as to why (for example) in newly created Loki clusters is there consul configuration if we're only actually using memberlist. 

Note that if OSS users are doing migrations from consul or another kv store to memberlist the memberlist_migration_teardown jsonnet config option will preserve the config of running components at the end of the migration until their rollout is completed.

### DIFF
--- a/production/ksonnet/loki/config.libsonnet
+++ b/production/ksonnet/loki/config.libsonnet
@@ -229,7 +229,7 @@
           ring: {
             heartbeat_timeout: '1m',
             replication_factor: $._config.replication_factor,
-            kvstore: {
+            kvstore: if $._config.memberlist_ring_enabled then {} else {
               store: 'consul',
               consul: {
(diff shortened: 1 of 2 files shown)


production/ksonnet/loki/config.libsonnet
```diff
@@ -229,7 +229,7 @@
           ring: {
             heartbeat_timeout: '1m',
             replication_factor: $._config.replication_factor,
-            kvstore: {
+            kvstore: if $._config.memberlist_ring_enabled then {} else {
```
DylanGuedes, line 232: shouldn't this check be inside the consul key (line ~214)?

dannykopping (approved):
LGTM, thanks!

PR #6810, opened 2022-07-29
Simplify shard membership calculations

owen-d:
This calculates shard membership in an equivalent but simpler way.


kavirajk (approved):
LGTM :+1: 

Assuming we have enough test cases for `Match` method. :)

PR #6817, opened 2022-08-01
compactor: adds tunable to control upload parallelism

afayngelerindbx:
This can help with compaction latency. I patched the compactor to set this to 100.

Helps with #6775

**Checklist**
- [x] Documentation added


pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -95,6 +96,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.RetentionDeleteWorkCount, "boltdb.shipper.compactor.retention-delete-worker-count", 150, "The total amount of worker to use to delete chunks.")
 	f.DurationVar(&cfg.DeleteRequestCancelPeriod, "boltdb.shipper.compactor.delete-request-cancel-period", 24*time.Hour, "Allow cancellation of delete request until duration after they are created. Data would be deleted only after delete requests have been older than this duration. Ideally this should be set to at least 24h.")
 	f.IntVar(&cfg.MaxCompactionParallelism, "boltdb.shipper.compactor.max-compaction-parallelism", 1, "Maximum number of tables to compact in parallel. While increasing this value, please make sure compactor has enough disk space allocated to be able to store and compact as many tables.")
+	f.IntVar(&cfg.UploadParallelism, "boltdb.shipper.compactor.upload-parallelism", uploadIndexSetsConcurrency, "Number of upload/remove operations to execute in parallel when finalizing a compaction. Parallel compactions will each issue this many uploads concurrently.")
```
sandeepsukhani, line 99: I think we should drop the constant and hardcode it here, which is easy to reference.

We can also drop `Parallel compactions will each issue this many uploads concurrently.` from the help text to keep it concise.

sandeepsukhani (commented):
Left just a minor nit otherwise, it LGTM.

I didn't make it tunable in the first place since it also concurrently compresses the files, so if someone increases this value too much it could result in worst performance due to cotention instead of improving it.

We see it working fine with the hardcoded value of 10, even in large clusters with > 5k tenants. Are you seeing any problems with that in your case, and do you see any improvements increasing the concurrency?

afayngelerindbx:
> We see it working fine with the hardcoded value of 10, even in large clusters with > 5k tenants. Are you seeing any problems with that in your case, and do you see any improvements increasing the concurrency?

We were seeing 1-2 minutes of runtime in our cluster where uploads were happening. I (somewhat recklessly) tuned this up to 100 and the time went down to ~30s(I don't have access to exact graphs any more, sorry).

sandeepsukhani:
> We were seeing 1-2 minutes of runtime in our cluster where uploads were happening. I (somewhat recklessly) tuned this up to 100 and the time went down to ~15-30s(I don't have access to exact graphs any more, sorry). I suspect we have somewhat different perf characteristics since we have lots of idle CPU but we're not co-located with object storage.

Nice!

> One thing that maybe should be considered is making this tunable global(e.g. having a per-run semaphore rather than a per table setting). Lmk if you think we should abandon this PR in favor of that approach.

Making it per-run would be a bit more complex to implement, and I do not see a strong reason to do it and complicate the code.

sandeepsukhani (approved):
LGTM! Thanks for the PR!
Can you please rebase it so that I can go ahead and merge it?

sandeepsukhani (approved):
LGTM

PR #6818, opened 2022-08-01
Observe how long we block writing our own logs

dannykopping:
In our Grafana Cloud service, we're seeing some pods logging thousands of log lines a second. Each call to `Log()` incurs a `write` syscall, which is quite expensive at that scale.

Under benchmark conditions, each of these syscalls take ~1000ns to run - which is not much, but 30k of those per second translates to 30ms of unnecessary latency, not to mention the horror that that many context-switches incurs in the kernel & Go scheduler. _30k logs per second is not hyperbole - that is what we're seeing in our production environments on occasion!_

These measurements are extremely small so normal CPU profiling doesn't observe them completely; the pprof CPU profiler runs at [100Hz by default](https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/runtime/pprof/pprof.go;l=772), which is 100x per second or once every ~10ms. These operations are happening on a ~microsecond scale so at best the CPU profile is incomplete (_10ms = 10,000μs_).

This PR adds a timing metric around all calls to `Log()`, which measures the total amount of time we've spent **blocking** while writing log lines to stderr.

We're using `go-kit/log` under the hood, and it doesn't perform any buffering of the lines it writes. I've forked this lib and implemented a [line-buffered logger](https://github.com/go-kit/log/compare/main...dannykopping:linebuffer) which we will test once this PR rolls out so we can measure the impact.

Once we roll out this line-buffered logger, we should see a significant drop in the amount of time blocking - and we should see a concomitant decrease in context-switches 

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/util/log/log.go
+++ b/pkg/util/log/log.go
@@ -3,6 +3,7 @@ package log
 import (
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -38,6 +39,7 @@ func InitLogger(cfg *server.Config, reg prometheus.Registerer) {
 type prometheusLogger struct {
 	logger      log.Logger
 	logMessages *prometheus.CounterVec
+	timeSpent   prometheus.Counter
 }
 
 // newPrometheusLogger creates a new instance of PrometheusLogger which exposes
@@ -56,6 +58,11 @@ func newPrometheusLogger(l logging.Level, format logging.Format, reg prometheus.
 			Name:      "log_messages_total",
 			Help:      "Total number of log messages.",
 		}, []string{"level"}),
+		timeSpent: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Namespace: "loki",
+			Name:      "log_messages_time_spent_total",
+			Help:      "Total time blocking while writing logs to output file.",
+		}),
 	}
 	// Initialise counters for all supported levels:
 	supportedLevels := []level.Value{
@@ -74,7 +81,10 @@ func newPrometheusLogger(l logging.Level, format logging.Format, reg prometheus.
 
 // Log increments the appropriate Prometheus counter depending on the log level.
 func (pl *prometheusLogger) Log(kv ...interface{}) error {
+	start := time.Now()
 	pl.logger.Log(kv...)
+	pl.timeSpent.Add(time.Since(start).Seconds())
+
 	l := "unknown"
 	for i := 1; i < len(kv); i += 2 {
 		if v, ok := kv[i].(level.Value); ok {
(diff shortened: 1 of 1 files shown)


pkg/util/log/log.go
```diff
@@ -56,6 +58,21 @@ func newPrometheusLogger(l logging.Level, format logging.Format, reg prometheus.
 			Name:      "log_messages_total",
 			Help:      "Total number of log messages.",
 		}, []string{"level"}),
+		timeSpent: promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
+			Namespace: "loki",
+			Name:      "log_messages_time_spent",
+			Help:      "Total time blocking while writing logs to output file.",
+			Buckets: []float64{
```
kavirajk, line 65: could be `prometheus.ExponentialBuckets(100 * 1e9, 10, 8) // 1e9 - nanoseconds`. ?

start with 100 nanoseconds, with factor of 10, upto 8 times. Meaning 100 nanoseconds to 1 second.

pkg/util/log/log.go
```diff
@@ -74,7 +91,10 @@ func newPrometheusLogger(l logging.Level, format logging.Format, reg prometheus.
 
 // Log increments the appropriate Prometheus counter depending on the log level.
 func (pl *prometheusLogger) Log(kv ...interface{}) error {
+	start := time.Now()
 	pl.logger.Log(kv...)
+	pl.timeSpent.Observe(time.Since(start).Seconds())
```
kavirajk, line 96: Not sure how much overhead this `Observe` is going to take on the hotpath logging. Hope it's worth it.

kavirajk (approved):
LGTM :). 

Few minor nits.

pkg/util/log/log.go
dannykopping: Sure, but I prefer to be explicit

slim-bean (approved):
LGTM!

pkg/util/log/log.go
dannykopping: That's a good point Kavi, I'm not actually sure.
This won't be around permanently, it's just designed so that we can measure the problem so when we implement the fix we can see if it helps or not.

In fact, I'll convert this to a draft so we can refer to it and cherry-pick it into a build, but I won't merge it

MichelHollands: If it is on the hot path, https://github.com/iovisor/bcc/blob/ee433d4e22650bc2ceca2119a727707f493269b3/tools/funclatency.py from eBPF is another option.

MichelHollands (approved):
LGTM. Really cool!

pkg/util/log/log.go
dannykopping: Ooh that's interesting, something to try for sure if this doesn't work

dannykopping:
Closing this, since https://github.com/grafana/loki/pull/6954 is merged.

It's incredibly difficult to observe the performance impact of thousands of short events per second, but the line-buffered logger at least is more mechanically sympathetic to the way the Go scheduler works. We will test it out for a couple weeks and see if there is any measurable impact.

PR #6819, opened 2022-08-01
Log stream flushes to see how many chunks we flush at a time

dannykopping:
Our chunk flushing behaviour is a bit opaque at the moment; this logging adds more detail and clarifies what has actually happened a bit more.

N/A

I replaced `failed to flush user` which really made no sense to `failed to flush chunks` in an effort to try highlight what has really failed.

An example:

```
level=error ts=2022-08-01T14:42:41.909348925Z caller=flush.go:144 org_id=docker msg="failed to flush" err="failed to flush chunks: store put chunk: stat /tmp/loki/chunks/docker/df7eb4c863ffd707: permission denied, num_chunks: 2, labels: {filename=\"/tmp/random\", http_method=\"GET\", job=\"generated-logs\"}"
```

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


docs/sources/upgrading/_index.md
```diff
@@ -133,7 +133,7 @@ Meanwhile, the legacy format is a string in the following format:
...
 * `max_concurrent` under the `querier` config now defaults to `10` instead of `20`.
 * `match_max_concurrent` under the `frontend_worker` config now defaults to true, this supersedes the `parallelism` setting which can now be removed from your config. Controlling query parallelism of a single process can now be done with the `querier` `max_concurrent` setting.
-* `flush_op_timeout` under the `ingester` configuration block now defaults to `10m`, increased from `10s`. This can help when replaying a large WAL on Loki startup, and avoid `msg="failed to flush user" ... context deadline exceeded` errors.
+* `flush_op_timeout` under the `ingester` configuration block now defaults to `10m`, increased from `10s`. This can help when replaying a large WAL on Loki startup, and avoid `msg="failed to flush chunks" ... context deadline exceeded` errors.
```
slim-bean, line 136: See comment below

pkg/ingester/flush.go
```diff
@@ -139,11 +139,9 @@ func (i *Ingester) flushLoop(j int) {
 		}
 		op := o.(*flushOp)
 
-		level.Debug(util_log.Logger).Log("msg", "flushing stream", "userid", op.userID, "fp", op.fp, "immediate", op.immediate)
-
 		err := i.flushUserSeries(op.userID, op.fp, op.immediate)
 		if err != nil {
-			level.Error(util_log.WithUserID(op.userID, util_log.Logger)).Log("msg", "failed to flush user", "err", err)
+			level.Error(util_log.WithUserID(op.userID, util_log.Logger)).Log("msg", "failed to flush", "err", err, "user", op.userID)
```
slim-bean, line 144: I think I would leave this alone for now, this is a pretty well established error message.

Also, the logger includes the user ID so we shouldn't have to add it at the end.

slim-bean (approved):
Requesting we don't change the one log message, otherwise LGTM!

pkg/ingester/flush.go
dannykopping: Good point, done!

docs/sources/upgrading/_index.md
slim-bean, line 136:
```suggestion
* `flush_op_timeout` under the `ingester` configuration block now defaults to `10m`, increased from `10s`. This can help when replaying a large WAL on Loki startup, and avoid `msg="failed to flush" ... context deadline exceeded` errors.
```

I think this represents what the new error message will be, but also is less likely to confuse if others find this while running an older version and looking at the docs? WDYT?

pkg/ingester/flush.go
slim-bean: Also after some thought I'm fine with removing `user` from that error message.

slim-bean: > Good point, done!

Not sure if the push didn't make it or if github is playing tricks on me, but I still see `user` in the list of args?

docs/sources/upgrading/_index.md
dannykopping: Good idea

pkg/ingester/flush.go
dannykopping: Nope, I'd just missed the one case you actually pointed out but did the other others :upside_down_face:

slim-bean (approved):
LGTM!

PR #6823, opened 2022-08-01
Add ability for Ruler to remote write to multiple clients

### LINKED ISSUE #4368: Add ability to for Ruler to remote write to multiple (clustered) Prometheus servers
We run multiple Prometheus servers for HA, we need to be able to remote-write to all servers so that if one server crashes we don't loose the metrics from loki. 

Ability to point at more than 1 remote Prometheus, possible allow Prometheus to scrape loki, and if possible allow discovery of Prometheus servers using Service Discovery of K8s. 

I have tried to mirror the traffic from loki to prometheus but this is unreliable and hard to setup with current tools

**Additional context**
#4241

aminesnow:
This PR adds the ability for the ruler to write to more than one remote client. In order to preserve backward compatibility, a new field `Clients` is added to the `remote_write` config in addition to the previous `Client` field. The latter is only taken into account when `Clients` is empty.

Resolves #4368 

Should `Client` be deprecated for the upcoming major release?

**Checklist**
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/integration/cluster/cluster.go
+++ b/integration/cluster/cluster.go
@@ -8,6 +8,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"os"
+	"strings"
 	"sync"
 	"text/template"
 	"time"
@@ -74,7 +75,44 @@ frontend_worker:
 
 frontend:
   scheduler_address: localhost:{{.schedulerPort}}
+
+{{if .remoteWriteUrls}}
+ruler:
+  wal:
+    dir: {{.rulerWALPath}}
+  storage:
+    type: local
+    local:
+      directory: {{.rulesPath}}
+  rule_path: {{.sharedDataPath}}/rule
+  enable_api: true
+  ring:
+    kvstore:
+      store: inmemory
+  remote_write:
+    enabled: true
+    clients:
+      remote_client1:
+        url: {{index .remoteWriteUrls 0}}/api/v1/write
+      remote_client2:
+        url: {{index .remoteWriteUrls 1}}/api/v1/write
+{{end}}
 `))
+
+	rulesConfig = `
+groups:
+- name: always-firing
+  interval: 1s
+  rules:
+  - alert: fire
+    expr: |
+      1 > 0
+    for: 0m
+    labels:
+      severity: warning
+    annotations:
+      summary: test
+`
 )
 
 func wrapRegistry() {
@@ -197,8 +235,13 @@ type Component struct {
 	httpPort int
 	grpcPort int
 
-	configFile string
-	dataPath   string
+	configFile   string
+	dataPath     string
+	rulerWALPath string
+	rulesPath    string
+	RulesTenant  string
+
+	RemoteWriteUrls []string
 }
 
 func (c *Component) HTTPURL() *url.URL {
@@ -228,12 +271,46 @@ func (c *Component) writeConfig() error {
 		return fmt.Errorf("error creating data path: %w", err)
 	}
 
+	if len(c.RemoteWriteUrls) > 0 {
+		c.rulesPath, err = os.MkdirTemp(c.cluster.sharedPath, "rules")
+		if err != nil {
+			return fmt.Errorf("error creating rules path: %w", err)
+		}
+
+		fakeDir, err := os.MkdirTemp(c.rulesPath, "fake")
+		if err != nil {
+			return fmt.Errorf("error creating rules/fake path: %w", err)
+		}
+
+		s := strings.Split(fakeDir, "/")
+		c.RulesTenant = s[len(s)-1]
+
+		c.rulerWALPath, err = os.MkdirTemp(c.cluster.sharedPath, "ruler-wal")
+		if err != nil {
+			return fmt.Errorf("error creating ruler-wal path: %w", err)
+		}
+
+		rulesConfigFile, err := os.CreateTemp(fakeDir, "rules*.yaml")
+		if err != nil {
+			return fmt.Errorf("error creating rules config file: %w", err)
+		}
+
+		if _, err = rulesConfigFile.Write([]byte(rulesConfig)); err != nil {
+			return fmt.Errorf("error writing to rules config file: %w", err)
+		}
+
+		rulesConfigFile.Close()
+	}
+
 	if err := configTemplate.Execute(configFile, map[string]interface{}{
-		"dataPath":       c.dataPath,
-		"sharedDataPath": c.cluster.sharedPath,
-		"grpcPort":       c.grpcPort,
-		"httpPort":       c.httpPort,
-		"schedulerPort":  c.grpcPort,
+		"dataPath":        c.dataPath,
+		"sharedDataPath":  c.cluster.sharedPath,
+		"grpcPort":        c.grpcPort,
+		"httpPort":        c.httpPort,
+		"schedulerPort":   c.grpcPort,
+		"remoteWriteUrls": c.RemoteWriteUrls,
+		"rulesPath":       c.rulesPath,
+		"rulerWALPath":    c.rulerWALPath,
 	}); err != nil {
 		return fmt.Errorf("error writing config file: %w", err)
 	}
--- a/integration/cluster/cluster.go
+++ b/integration/cluster/cluster.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"net"
+	"net/http"
 	"net/http/httptest"
 	"net/url"
 	"os"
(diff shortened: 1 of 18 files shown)


pkg/ruler/config.go
```diff
@@ -48,14 +48,25 @@ func (c *Config) Validate() error {
...
+		}
+
+		if len(c.Clients) > 0 {
+			for _, clt := range c.Clients {
+				if clt.URL == nil {
+					return fmt.Errorf("remote-write enabled but client '%s' URL is not configured", clt.Name)
+				}
+			}
+		}
 	}
```
periklis, line 70: Suggest to flip the check to `!c.Enabled` and early return `nil`. This will save you one level of indentation and make the happy code path easier to read.

pkg/ruler/config.go
```diff
@@ -48,14 +48,25 @@ func (c *Config) Validate() error {
 }
 
 type RemoteWriteConfig struct {
-	Client              config.RemoteWriteConfig `yaml:"client"`
-	Enabled             bool                     `yaml:"enabled"`
-	ConfigRefreshPeriod time.Duration            `yaml:"config_refresh_period"`
+	Client              *config.RemoteWriteConfig  `yaml:"client,omitempty"`
+	Clients             []config.RemoteWriteConfig `yaml:"clients"`
```
periklis, line 52: This looks like adding a slice of remote write configs for a single tenant. Should we actually have a separate config per tenant? Shouldn't this be a `map[string]config.RemoteWriteConfig`. I suggest you take as good example of how to introduce a config per tenant the following in per-tenant-ruler-configs `ruler_remote_write_*` in the `limits_config` section:  https://grafana.com/docs/loki/latest/configuration/#limits_config

periklis (changes requested):
In general a good start, I suggest to revise the config placement to `limits_config` to enable per-tenant remote write configuration.

pkg/ruler/config.go
aminesnow: it's tricky because, if I understand correctly, multi tenancy for the ruler is handled by `limit_config` overrides. By using a map, are you suggesting that we handle multi tenancy for ruler directly in the configs instead of overrides?

pkg/ruler/ruler.go
```diff
@@ -10,6 +10,12 @@ import (
 )
 
 func NewRuler(cfg Config, engine *logql.Engine, reg prometheus.Registerer, logger log.Logger, ruleStore rulestore.RuleStore, limits RulesLimits) (*ruler.Ruler, error) {
+	// For backward compatibility, client and clients are defined in the remote_write config.
+	// When both are present, Clients config overrides client, otherwise one or the other is used.
+	if len(cfg.RemoteWrite.Clients) == 0 && cfg.RemoteWrite.Client != nil {
```
dannykopping, line 15: What do you think about erroring in this case? Nobody should ever do this

pkg/util/cfg/dynamic.go
```diff
@@ -22,14 +22,15 @@ func DynamicUnmarshal(dst DynamicCloneable, args []string, fs *flag.FlagSet) err
...
+		ConfigFileLoader(args, "config.file", true),
 		// Apply any dynamic logic to set other defaults in the config. This function is called after parsing the
 		// config files so that values from a common, or shared, section can be used in
 		// the dynamic evaluation
 		dst.ApplyDynamicConfig(),
 		// Load configs from the config file a second time, this will supersede anything set by the common
 		// config with values specified in the config file.
-		ConfigFileLoader(args, "config.file"),
+		// Second marshaling is not strict because the config is already populated and that causes errors with maps.
+		ConfigFileLoader(args, "config.file", false),
```
dannykopping, line 33: Can you please expand a bit on why this was necessary to add?
This is such a fundamental behaviour in Loki, and we should be careful about changing it

"causes errors with maps" - what did you see here?

integration/cluster/cluster.go
```diff
@@ -228,12 +271,46 @@ func (c *Component) writeConfig() error {
 		return fmt.Errorf("error creating data path: %w", err)
 	}
 
+	if len(c.RemoteWriteUrls) > 0 {
+		c.rulesPath, err = os.MkdirTemp(c.cluster.sharedPath, "rules")
```
dannykopping, line 275: Please clean up these temp files after the run

dannykopping (commented):
This is amazing work @aminesnow :tada: 
I've left a few small notes, but overall I'm very happy with this! Thank you for your efforts here.

pkg/ruler/ruler.go
aminesnow: Sure, this would be a very confusing usage of the config indeed!

pkg/util/cfg/dynamic.go
aminesnow: By loading the config file twice and unmarshaling it into the same object, using strict yaml unmarshal causes a `already set in map` error with the `Clients` config, because it's a map that already has the keys we are trying to unmarshal into it.

integration/cluster/cluster.go
aminesnow: done

dannykopping: Can I ask you to please leave comments to the author to resolve? Otherwise it becomes difficult to track.
Where is the cleanup happening?

pkg/util/cfg/dynamic.go
```diff
@@ -22,14 +22,18 @@ func DynamicUnmarshal(dst DynamicCloneable, args []string, fs *flag.FlagSet) err
...
 		// the dynamic evaluation
 		dst.ApplyDynamicConfig(),
 		// Load configs from the config file a second time, this will supersede anything set by the common
 		// config with values specified in the config file.
-		ConfigFileLoader(args, "config.file"),
+		// By loading the config file twice and unmarshaling it into the same object,
+		// using strict yaml unmarshal causes an `already set in map` error with the `Clients` config,
+		// because it's a map that already has the keys we are trying to unmarshal into it.
+		// That is why we don't use strict for the second marshaling.
+		ConfigFileLoader(args, "config.file", false),
```
dannykopping, line 36: Thanks for the explanation but I'd love it if we could figure out how to solve this without changing this.
I'm concerned making this unstrict will introduce some subtle bugs.
I don't know this part of the code well, which is contributing to my skepticism.

pkg/util/cfg/dynamic.go
dannykopping: Thanks for the explanation but I'd love it if we could figure out how to solve this without changing this.
I'm concerned making this unstrict will introduce some subtle bugs.
I don't know this part of the code well, which is contributing to my skepticism.

aminesnow: It's not actually an issue since we already marshal the file with strict the first time, so any problems with the yaml would result in an error the first time we do the marshaling, so the second one we can confidently marshal without being strict

integration/cluster/cluster.go
aminesnow: oh sorry my bad!

pkg/util/cfg/dynamic.go
dannykopping: I'm sure there was a reason for the strict parsing the second time around, though; @slim-bean I see you initially wrote this; could you weigh in pls?

integration/cluster/cluster.go
dannykopping: @aminesnow are we cleaning up the temp files/directories somewhere?

pkg/ruler/ruler.go
```diff
@@ -11,7 +12,11 @@ import (
 
 func NewRuler(cfg Config, engine *logql.Engine, reg prometheus.Registerer, logger log.Logger, ruleStore rulestore.RuleStore, limits RulesLimits) (*ruler.Ruler, error) {
 	// For backward compatibility, client and clients are defined in the remote_write config.
-	// When both are present, Clients config overrides client, otherwise one or the other is used.
+	// When both are present, an error is thrown.
+	if len(cfg.RemoteWrite.Clients) > 0 && cfg.RemoteWrite.Client != nil {
+		return nil, errors.New("Both Client and Clients are defined! Please use one or the other")
```
dannykopping, line 17: Please be specific about where this error is coming from:
```suggestion
		return nil, errors.New("ruler remote write config: both 'client' and 'clients' options are defined; 'client' is deprecated, please only use 'clients'")
```

dannykopping (commented):
I'm back from PTO and gave it another review. This is getting very close to mergeable, thanks @aminesnow!

I'm still concerned about the YAML config change. I'll ping @slim-bean for his opinion on it.

integration/cluster/cluster.go
aminesnow: @dannykopping hey, just back from PTO.
yes we do, in the `cleanup` function line 398

periklis (commented):
Looks good to me!

integration/cluster/cluster.go
dannykopping: Ah, I had to checkout the PR to understand the flow. Got it, thanks :+1:

dannykopping (approved):
LGTM, thanks for the excellent work @aminesnow!

periklis (approved):

PR #6824, opened 2022-08-01
compactor: observe the number of index files that need to be compacted

afayngelerindbx:
When debugging the compactor, I was often looking for the "listed" log
line as well as the "deleting" log lines to determine whether
compactor is making materialized progress.

This change adds a guage so that the number of files that need compaction
can be tracked in a metrics store.

This would've helped with #6775

### DIFF
--- a/pkg/storage/stores/indexshipper/compactor/metrics.go
+++ b/pkg/storage/stores/indexshipper/compactor/metrics.go
@@ -16,7 +16,7 @@ type metrics struct {
 	compactTablesOperationLastSuccess     prometheus.Gauge
 	applyRetentionLastSuccess             prometheus.Gauge
 	compactorRunning                      prometheus.Gauge
-	indexFilesToCompactForLastCompaction  *prometheus.GaugeVec
+	uncompactedIndexFilesDiscovered       *prometheus.GaugeVec
 }
 
 func newMetrics(r prometheus.Registerer) *metrics {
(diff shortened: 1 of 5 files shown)


pkg/storage/stores/indexshipper/compactor/metrics.go
```diff
@@ -45,6 +46,11 @@ func newMetrics(r prometheus.Registerer) *metrics {
 			Name:      "compactor_running",
 			Help:      "Value will be 1 if compactor is currently running on this instance",
 		}),
+		indexFilesToCompactForLastCompaction: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: "loki_boltdb_shipper",
+			Name:      "compact_tables_operation_index_files_in_last_compaction",
+			Help:      "The number of index files observed by last compaction",
```
kavirajk, line 52: The name(`indexFilesToCompact.*`) and help(`index files observed by last compaction`) seems to contradict?. It's hard to understand whether
1. it tracks index files that was part of last run compaction?
2. Or tracks the index files that will be compacted in next compaction?

Also your PR description seems to accomplish (2).

Can you clarify please?

afayngelerindbx: Thanks for reviewing. 

Looking at this after a bit of break, it is definitely more complicated than it needs to be. For what I was doing, it is sufficient to gauge the number of files discovered. I've reworked the diff to do so.

I think what I was trying to get with this diff is an estimate of how many un-compacted files are in storage. I was also hoping to use this metric to measure compaction throughput. However, I don't think I'm doing a good job of that here.

pkg/storage/stores/indexshipper/compactor/table.go
```diff
@@ -127,7 +131,9 @@ func (t *table) compact(applyRetention bool) error {
 	t.usersWithPerUserIndex = usersWithPerUserIndex
 
 	level.Info(t.logger).Log("msg", "listed files", "count", len(indexFiles))
```
MasslessParticle, line 133: There's a case above where `len(indexFiles) == 0`, we probably want to collect that metric too?

pkg/storage/stores/indexshipper/compactor/metrics.go
MasslessParticle: What about a name of something like `indexFilesToCompact`, `compact_tables_operation_index_files_count`, etc to decouple the naming from when things were compacted? That way this metric will tell you how many tables were to be compacted at any point in time.

afayngelerindbx:
Haven't had time to follow up on this. Not sure it is still necessary.

PR #6827, opened 2022-08-02
Treat version flag at any position

simonswine:
The behaviour was changed as part of #6802. This should restore the previous behaviour and allow the version to be displayed without config file.


dannykopping (approved):
LGTM :+1:

PR #6829, opened 2022-08-02
Bring helm chart back into Loki repo

### LINKED ISSUE #6873: [Helm] Move the loki helm chart back into the loki repo
Merge `grafana/loki-simple-scalable` and `grafana/loki` into one chart that supports both modes, and move that chart back into the loki repo. Currently there are ~7 loki/GEL helm charts, making it very confusing which one to use. This also creates a large surface area for issues/PRs making it difficult for the community to keep these charts up to date.

The end goal here is to publish 1 helm chart under `grafana/loki`, replacing the current chart published under that name. Since that current chart is for single binary mode, we want to continue to support that mode of operation in the new chart, but would like to change the default to a simple scalable architecture due to it's ease of scalability.

trevorwhitney:

This is the first step of the process to bring the helm chart back into the Loki repo. We currently have ~7 public helm charts. This creates a lot of confusion for the community on which ones to start with, and is also a lot of additional surface area that is hard for Loki contributors to keep track of. By bringing the helm chart that we use ourselves back into the Loki repo, it will be a more clear starting point for helm users. It will also increase visibility for Loki contributors into helm PRs and issues. This will _not_ affect where the helm chart is published. It will still be published to the `grafana/helm-charts` following the same pattern as the Mimir helm chart.

The helm chart that we are currently using and most actively supporting is the `loki-simple-scalable` chart. Part of this effort is to rename that chart to just `loki`, which means it needs to be expanded to support the single binary deployment topology currently deployed by the existing `loki` helm chart. This PR is the first attempt to merge those two charts.

This chart currently works for a basic install of each deployment topology. There is more work to be done before we publish it. This work will include:

  * Making sure there is a migration path from the existing `loki` chart to this new one
  * Making sure there is a migration path from the existing `loki-simple-scalable` chart to this new one
  * Making sure the single binary deployment works in both `filesystem` and minio-backed multi-replica mode
  * Migrating CI from `grafana/helm-charts` for testing helm chart changes
  * Integrating with the github action in the `grafana/helm-charts` repo (this will publish it, so we will tackle this last)

A question I have for reviewers is what we should name each deployment mode (single binary vs. simple scalable)? We have received feedback from the community that some did not think "simple scalable" was suitable for production loads since it had "simple" in the name. This might be an opportunity to adjust that perception. Some options I've thought of include:

  * `single-binary` and `simple-scalable` (currently how it's coded)
  * `basic` and `scalable` (leaving open the option for `advanced` if we ever choose to support micro-services topology in this chart as well)
  * `ha`, `redundant` or similar vs. not

Fixes #6873


production/helm/loki/CHANGELOG.md
```diff
@@ -0,0 +1,47 @@
...
+- [BUGFIX] Fix a bug in how object storage client was defined for the ruler
+
+## 1.7.2
+
+- [ENHANCEMENT] Memcache can now easily be configured as an external caching option.
+=======
+## 1.6.3
+
+- [CHANGE] The `prometheusRule` value had been removed. If you were using this to add custom recording rules, move `prometheusRule.groups` to `monitoring.rules.additionalGroups`, which will merge your custom groups with the ones that now ship out of the box with this chart.
+>>>>>>> 41f3c7fc (Include required CRDs, condense rules to 1 file)
```
DylanGuedes, line 33: hmm some rebase trails?

production/helm/ct.yaml
```diff
@@ -1,6 +1,11 @@
+---
+remote: origin
+target-branch: main
 chart-repos:
   - elastic=https://helm.elastic.co
   - grafana=https://grafana.github.io/helm-charts
   - prometheus=https://prometheus-community.github.io/helm-charts
+  - minio=https://helm.min.io
+helm-extra-args: --timeout 600s
```
DylanGuedes, line 9: hmm maybe this --timeout flag can be used as a new default instead?

production/helm/loki/CHANGELOG.md
```diff
@@ -0,0 +1,47 @@
...
+
+Entries should be ordered as follows:
+
+- [CHANGE]
+- [FEATURE]
+- [ENHANCEMENT]
+- [BUGFIX]
+
+Entries should include a reference to the pull request that introduced the change.
+
```
DylanGuedes, line 13: since the chart is in a different location now, I'm not sure if the versioning should be kept. Anyway, just a thought: maybe we should try using a versioning like `x-y`, were x is the related loki version and y is the chart version. So this chart would be `2.6.0-1` or something. I feel like semver doesn't suit it very well idk

production/helm/loki/Chart.yaml
```diff
@@ -1,13 +1,25 @@
...
-version: 2.1.1
-appVersion: v2.0.0
-kubeVersion: "^1.10.0-0"
-description: "DEPRECATED Loki: like Prometheus, but for logs."
-home: https://grafana.com/loki
-icon: https://raw.githubusercontent.com/grafana/loki/master/docs/sources/logo.png
+description: Helm chart for Grafana Loki in simple, scalable mode
+type: application
+appVersion: 2.6.1
+version: 3.0.0
```
DylanGuedes, line 6: is this version the chart version?

DylanGuedes (commented):
since I'll be OOO next week, I could only review part of your PR. Most of these are questions anyway :smile:

production/helm/loki/CHANGELOG.md
trevorwhitney: oh gross, thanks for catching that!

production/helm/loki/Chart.yaml
trevorwhitney: correct. this chart will be published under `grafana/loki`, which is currently [this](https://github.com/grafana/helm-charts/tree/main/charts/loki) chart (version 2.13.3 currently)

production/helm/ct.yaml
trevorwhitney: I'm not sure how to do that? I'm pretty sure these are passed as extra args to `helm init` when using the ct tool for CI jobs, maybe there's a config for it in the CI jobs?

production/helm/loki/CHANGELOG.md
trevorwhitney: agreed, nuked it to just have an entry for `3.0.0` mentioning the deprecation with links back to the original locations

production/helm/ct.yaml
DylanGuedes: ohh got it, I thought it was a Loki timeout. nvm

production/helm/loki/Chart.yaml
DylanGuedes: got it

production/helm/loki/Chart.yaml
```diff
@@ -1,13 +1,25 @@
...
-engine: gotpl
+  - https://grafana.com/oss/loki/
+  - https://grafana.com/docs/loki/latest/
+icon: https://grafana.com/docs/loki/latest/logo_and_name.png
+dependencies:
+  - name: minio
+    alias: minio
+    version: 8.0.9
+    repository: https://helm.min.io/
+    condition: minio.enabled
```
simonswine, line 18: Minio seems to have a different repo now: https://github.com/minio/minio/tree/master/helm/minio

simonswine (commented):
This is some massive piece of work, I will give a more thorough review tomorrow

production/helm/loki/Chart.yaml
trevorwhitney: Will address that in a later issue: https://github.com/grafana/loki/issues/6883

production/helm/loki/CHANGELOG.md
```diff
@@ -0,0 +1,18 @@
...
+## 3.0.0
+
+- [CHANGE] Loki helm chart was moved to this location in the Loki repo. The chart now supports both
+[single binary](https://github.com/grafana/helm-charts/tree/main/charts/loki) and [simple scalable](https://github.com/grafana/helm-charts/tree/main/charts/loki-simple-scalable) deployment modes. For changes prior to version 3.0.0, please
+looki in the respective deprectated [single binary](https://github.com/grafana/helm-charts/tree/main/charts/loki) and [simple scalable](https://github.com/grafana/helm-charts/blob/main/charts/loki-simple-scalable/CHANGELOG.md) charts.
```
jeschkies, line 18:
```suggestion
looki in the respective deprectated [single binary](https://github.com/grafana/helm-charts/tree/main/charts/loki) and [simple scalable](https://github.com/grafana/helm-charts/blob/main/charts/loki-simple-scalable/CHANGELOG.md) charts.
```
```suggestion
look in the respective deprectated [single binary](https://github.com/grafana/helm-charts/tree/main/charts/loki) and [simple scalable](https://github.com/grafana/helm-charts/blob/main/charts/loki-simple-scalable/CHANGELOG.md) charts.
```

production/helm/loki/Chart.lock
```diff
@@ -0,0 +1,9 @@
+dependencies:
+- name: minio
+  repository: https://helm.min.io/
+  version: 8.0.9
+- name: grafana-agent-operator
```
jeschkies, line 5: Out of curiosity will this collide if I have the Agent installed already or is Helm supposed to resolve all dependencies?

production/helm/loki/README.md
```diff
@@ -1,69 +1,406 @@
-# ⚠️  DEPRECATED - Loki Helm Chart
+# loki-simple-scalable
```
jeschkies, line 1:
```suggestion
# Loki Simple Scalable
```

production/helm/loki/README.md
```diff
@@ -1,69 +1,406 @@
...
-You can update the chart repository by running:
+## Requirements
 
-```bash
-$ helm repo update
-```
+| Repository | Name | Version |
+|------------|------|---------|
+| https://grafana.github.io/helm-charts | grafana-agent-operator(grafana-agent-operator) | 0.2.2 |
+| https://helm.min.io/ | minio(minio) | 8.0.9 |
```
jeschkies, line 18: Part of me wants to generate this entry :joy:

production/helm/loki/README.md
```diff
@@ -1,69 +1,406 @@
...
+Add the following repo to use the chart:
 
-```bash
-$ helm upgrade --install loki loki/loki
+```sh
+helm repo add grafana https://grafana.github.io/helm-charts
 ```
 
-## Run Loki behind https ingress
+## Values
```
jeschkies, line 28: I find these long READMEs often overwhelming. What do you think about linking to the documentation instead?

production/helm/loki/src/rules.yaml
```diff
@@ -0,0 +1,49 @@
+groups:
```
jeschkies, line 1: Where are these rules used?

jeschkies (commented):
I'm new to Helm and the Loki Helm charts. What differences should I take a look at?

production/helm/loki/Chart.lock
jeschkies: I've read the docs. Should the agent be installed into its own namespace? Maybe we should make this dependency configurable.

production/helm/loki/src/dashboards/loki-chunks.json
```diff
@@ -0,0 +1,1336 @@
+{
```
jeschkies, line 1: I'm wondering. Shouldn't we use the dashboards from the mixin?

production/helm/loki/templates/monitoring/pod-logs.yaml
```diff
@@ -0,0 +1,49 @@
+---
```
jeschkies, line 1: How does this relate to the logs-instance?

production/helm/loki/Chart.lock
trevorwhitney: I think exposing a separate namespace for the agent is a good idea. In the case of someone already having the agent installed, I would suggest they disable this by setting `monitoring.selfMonitoring.grafanaAgent.installOperator` to `false`.

production/helm/loki/README.md
trevorwhitney: good catch, this file should be auto-generated. For the purpose of this PR I'm going to bring in the README from the simple scalable chart and we can fix it up later to include the single binary steps as well.

production/helm/loki/README.md
trevorwhitney: yes, it should be. Added the `README.md.gotmpl`

production/helm/loki/README.md
trevorwhitney: I think we need to keep the values because of the auto-generation, but I'd be open to moving them to the bottom of the file.

production/helm/loki/src/rules.yaml
trevorwhitney: recording rules used in our dashboards, but they're not actually wired up to work yet

production/helm/loki/src/dashboards/loki-chunks.json
trevorwhitney: Are you suggesting we instead compile them as part of the build process? I think that would be a great addition, but that we should tackle that in a later PR.

For now these are the pre-compiled version, pulled directly from: https://github.com/grafana/loki/tree/main/production/loki-mixin-compiled-ssd

production/helm/loki/templates/monitoring/pod-logs.yaml
trevorwhitney: The logs instance tells grafana agent where to send the logs, whereas pod logs is telling the agent how to scrape for logs. PodLogs can be thought of as the source, ie. defining scrape configs, whereas LogsInstance is the sink, ie. the actual Loki database.

jeschkies (approved):
Let's land it and iterate.

PR #6834, opened 2022-08-03
Improve error message when entry is too far behind

simonswine:
This improves on the error if the ingested timestamp is significantly behind the highest timestamp of a stream (bigger than MaxChunkAge)

before: 

```
entry with timestamp 2022-08-02 12:32:37 +0000 UTC ignored, reason: 'entry too far behind' for stream: {foo="bar2"}
```

after:

```
entry with timestamp 2022-08-02 12:32:37 +0000 UTC ignored, reason: 'entry too far behind, oldest acceptable timestamp is: 2022-08-03T15:25:57Z' for stream: {foo="bar2"}
```

I think we can probably do even better (cc @KMiller-Grafana). But I guess that gets us a step closer.

### DIFF
--- a/pkg/chunkenc/interface.go
+++ b/pkg/chunkenc/interface.go
@@ -18,14 +18,30 @@ import (
 var (
 	ErrChunkFull       = errors.New("chunk full")
 	ErrOutOfOrder      = errors.New("entry out of order")
-	ErrTooFarBehind    = errors.New("entry too far behind")
 	ErrInvalidSize     = errors.New("invalid size")
 	ErrInvalidFlag     = errors.New("invalid flag")
 	ErrInvalidChecksum = errors.New("invalid chunk checksum")
 )
 
+type errTooFarBehind struct {
+	cutoff time.Time
+}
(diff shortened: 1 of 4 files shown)


pkg/chunkenc/interface.go
```diff
@@ -16,16 +16,33 @@ import (
 
 // Errors returned by the chunk interface.
 var (
-	ErrChunkFull       = errors.New("chunk full")
-	ErrOutOfOrder      = errors.New("entry out of order")
-	ErrTooFarBehind    = errors.New("entry too far behind")
+	ErrChunkFull  = errors.New("chunk full")
+	ErrOutOfOrder = errors.New("entry out of order")
+	//	ErrTooFarBehind    = errors.New("entry too far behind")
```
dannykopping, line 21:
```suggestion
```

dannykopping (approved):
:heart: always love these types of PRs

JordanRushing (approved):
Looks good; thanks!

PR #6835, opened 2022-08-03
Adapt queryTimeout to be a per-tenant configuration

DylanGuedes:
- Adds context timeout to metadata calls based on the new per-tenant
  query timeout
- Add span logs to calls missing it

The motivation to change this configuration to be per-tenant instead of global is that it makes sense to alleviate the timeout for particular tenants or make it more strict for others. Especially useful for scenarios where the Loki client doesn't handle context cancellations correctly; in such scenarios, since Loki would still process expensive queries that were canceled, this allows one to have small timeouts for most tenants, which will help mitigate unnecessary work without having a timeout that is too short for important tenants.

N/A


DylanGuedes:
Still missing updating the docs and testing a few things, but I'm moving it to ready in case someone is willing to review it/help me finding problems with my approach here :smile:

pkg/logcli/client/file.go
```diff
@@ -183,13 +183,18 @@ func (f *FileClient) GetOrgID() string {
 }
 
 type limiter struct {
-	n int
+	n       int
+	timeout time.Duration
```
kavirajk, line 187: Q: I don't see it's getting set anywhere in the file client. Could lead to unexpected behavior when using Logcli with file client I think. Can you also fix it where `limiter` is getting instantiated? (and with some sane default if value is not provided?)

pkg/logql/engine.go
```diff
@@ -226,7 +219,14 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
 }
 
 func (q *query) Eval(ctx context.Context) (promql_parser.Value, error) {
-	ctx, cancel := context.WithTimeout(ctx, q.timeout)
+	queryTimeout := time.Minute * 2
```
kavirajk, line 222: I think better we move this to package level const? (e.g: DefaultQueryTimeout). And Q: what's the rationale behind 2 minutes as default? (Previously query engine was using 5 * minute as default timeout. Is it to standardize it? then we may need to add it in upgrade guide as well I think)

pkg/querier/http.go
```diff
@@ -376,14 +408,23 @@ func (q *QuerierAPI) TailHandler(w http.ResponseWriter, r *http.Request) {
 // SeriesHandler returns the list of time series that match a certain label set.
 // See https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers
 func (q *QuerierAPI) SeriesHandler(w http.ResponseWriter, r *http.Request) {
+	log, ctx := spanlogger.New(r.Context(), "query.Series")
```
kavirajk, line 411: Seeing it is being repeated for almost all the handler, should we make this as middleware?

pkg/logql/engine.go
DylanGuedes: hmm where did you find that the engine had 5 minutes default timeout?

pkg/logcli/client/file.go
DylanGuedes: good catch, ty. I modified it to always use the 5minutes timeout, wdyt?

pkg/querier/http.go
DylanGuedes: hmm I think it shouldn't, as some handlers won't want this (ex: the tail handler)

pkg/logql/engine.go
```diff
@@ -226,7 +216,14 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
...
```
dannykopping, line 219: We're changing the default timeout to 2m here?
This should definitely be documented.

pkg/logql/engine.go
```diff
@@ -226,7 +216,14 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
...
+	userID, err := tenant.TenantID(ctx)
+	if err != nil {
+		level.Warn(q.logger).Log("msg", "couldn't fetch tenantID to evaluate query timeout, using default value of 2m", "err", err)
```
dannykopping, line 222: Rather use the timeout itself:
```suggestion
		level.Warn(q.logger).Log("msg", fmt.Sprintf("couldn't fetch tenantID to evaluate query timeout, using default value of %s", queryTimeout), "err", err)
```

Disclaimer: I'm not sure if `%s` works here, but you get the idea
If the timeout is changed above, we definitely don't want a stale log message stating the previous default of 2m

pkg/querier/http.go
```diff
@@ -64,8 +64,16 @@ func NewQuerierAPI(cfg Config, querier Querier, limits *validation.Overrides, lo
 
 // RangeQueryHandler is a http.HandlerFunc for range queries.
 func (q *QuerierAPI) RangeQueryHandler(w http.ResponseWriter, r *http.Request) {
+	log, ctx := spanlogger.New(r.Context(), "query.RangeQuery")
+
+	userID, err := tenant.TenantID(ctx)
+	if err != nil {
+		level.Error(log).Log("msg", "couldn't fetch tenantID", "err", err)
+	}
```
dannykopping, line 72: I don't think you want to continue if a tenant ID is missing.

The suggestion below copies the error handling from `TailHandler` in the same file

```suggestion
	userID, err := tenant.TenantID(ctx)
	if err != nil {
		level.Warn(logger).Log("msg", "error getting tenant id", "err", err)
		serverutil.WriteError(httpgrpc.Errorf(http.StatusBadRequest, err.Error()), w)
		return
	}
```

pkg/querier/http.go
```diff
@@ -103,8 +111,16 @@ func (q *QuerierAPI) RangeQueryHandler(w http.ResponseWriter, r *http.Request) {
 
 // InstantQueryHandler is a http.HandlerFunc for instant queries.
 func (q *QuerierAPI) InstantQueryHandler(w http.ResponseWriter, r *http.Request) {
+	log, ctx := spanlogger.New(r.Context(), "query.InstantQuery")
+
+	userID, err := tenant.TenantID(ctx)
+	if err != nil {
+		level.Error(log).Log("msg", "couldn't fetch tenantID", "err", err)
```
dannykopping, line 118: Ditto for the above and other handlers

pkg/querier/http.go
dannykopping: I definitely agree about the middleware; we can apply the middleware selectively - which is already done in `pkg/loki/modules.go`:


```go
	queryHandlers := map[string]http.Handler{
		"/loki/api/v1/query_range":         httpMiddleware.Wrap(http.HandlerFunc(t.querierAPI.RangeQueryHandler)),
		"/loki/api/v1/query":               httpMiddleware.Wrap(http.HandlerFunc(t.querierAPI.InstantQueryHandler)),
		"/loki/api/v1/label":               http.HandlerFunc(t.querierAPI.LabelHandler),
		"/loki/api/v1/labels":              http.HandlerFunc(t.querierAPI.LabelHandler),
		"/loki/api/v1/label/{name}/values": http.HandlerFunc(t.querierAPI.LabelHandler),
		"/loki/api/v1/series":              http.HandlerFunc(t.querierAPI.SeriesHandler),
		"/loki/api/v1/index/stats":         http.HandlerFunc(t.querierAPI.IndexStatsHandler),
```

pkg/querier/querier.go
```diff
@@ -60,7 +59,6 @@ type Config struct {
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	cfg.Engine.RegisterFlagsWithPrefix("querier", f)
 	f.DurationVar(&cfg.TailMaxDuration, "querier.tail-max-duration", 1*time.Hour, "Limit the duration for which live tailing request would be served")
-	f.DurationVar(&cfg.QueryTimeout, "querier.query-timeout", 1*time.Minute, "Timeout when querying backends (ingesters or storage) during the execution of a query request")
```
dannykopping, line 63: We should aim to deprecate options first, then remove.
A missing flag will cause the application to not start AFAIK.
All flag changes should be accompanied by a changelog/upgrade guide entry

pkg/logql/engine.go
DylanGuedes: The timeout is only 2 minutes for when we can't fetch the tenant. But regardless of it, no reason to change this to 2 minutes so I'm returning it to 5 minutes.

pkg/querier/querier.go
DylanGuedes: I added a deprecated notice; WDYT?

pkg/querier/http.go
DylanGuedes: Done, thanks.

pkg/querier/http.go
DylanGuedes: Done, thanks.

docs/sources/configuration/_index.md
```diff
@@ -321,7 +321,8 @@ The `querier` block configures the Loki Querier.
 
 # Configuration options for the LogQL engine.
 engine:
-  # Timeout for query execution
+  # Timeout for query execution.
+  # This configuration is deprecated in favor of querier.query-timeout.
```
dannykopping, line 325: Just to stick with the loose pattern already in the file

```suggestion
  # Deprecated: use querier.query-timeout instead.
```

pkg/querier/querier.go
dannykopping: Nice, thanks!

dannykopping (approved):
LGTM!

PR #6836, opened 2022-08-03
[jsonnet] don't deploy consul if we're using memberlist

cstyan:
This PR changes our jsonnet code so that we don't deploy consul if we're using memberlist and not actively migrating from consul to memberlist.


trevorwhitney (approved):
lgtm

PR #6837, opened 2022-08-04
Delete requests processing improvements

sandeepsukhani:
It includes a couple of improvements for handling a large number of delete requests:
* Group delete requests by users to quickly check if a user has any delete requests or not.
* Track the oldest start and latest end time of delete requests from a user to quickly check if a chunk overlaps that interval before checking it against all the delete requests.

I have also removed the overlap check for user index(i.e userID != "") in `IntervalMayHaveExpiredChunks`.
This is to ensure we update the index for chunks that overlap multiple tables and are requested for deletion while the delete requests do not cover all the table intervals.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
+++ b/pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
@@ -263,7 +263,7 @@ func (d *DeleteRequestsManager) IntervalMayHaveExpiredChunks(_ model.Interval, u
 
 	// We can't do the overlap check between the passed interval and delete requests interval from a user because
 	// if a request is issued just for today and there are chunks spanning today and yesterday then
-	// the overlap check would skip processing yesterdays index which would result in index pointing to deleted chunks.
+	// the overlap check would skip processing yesterdays index which would result in the index pointing to deleted chunks.
 	if userID != "" {
 		return d.deleteRequestsToProcess[userID] != nil
 	}
--- a/pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
+++ b/pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
@@ -18,11 +18,17 @@ const (
 	statusFail    = "fail"
 )
 
+type userDeleteRequests struct {
+	requests []DeleteRequest
+	// requestsInterval holds the earliest start time and latest end time considering all the delete requests
+	requestsInterval model.Interval
+}
+
 type DeleteRequestsManager struct {
 	deleteRequestsStore       DeleteRequestsStore
 	deleteRequestCancelPeriod time.Duration
 
-	deleteRequestsToProcess map[string][]DeleteRequest
+	deleteRequestsToProcess map[string]*userDeleteRequests
 	chunkIntervalsToRetain  []retention.IntervalFilter
 	// WARN: If by any chance we change deleteRequestsToProcessMtx to sync.RWMutex to be able to check multiple chunks at a time,
 	// please take care of chunkIntervalsToRetain which should be unique per chunk.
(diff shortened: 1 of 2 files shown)


pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
```diff
@@ -199,52 +227,47 @@ func (d *DeleteRequestsManager) MarkPhaseFailed() {
...
 }
 
-func (d *DeleteRequestsManager) IntervalMayHaveExpiredChunks(interval model.Interval, userID string) bool {
+func (d *DeleteRequestsManager) IntervalMayHaveExpiredChunks(_ model.Interval, userID string) bool {
 	d.deleteRequestsToProcessMtx.Lock()
 	defer d.deleteRequestsToProcessMtx.Unlock()
 
+	// We can't do the overlap check between the passed interval and delete requests interval from a user because
+	// if a request is issued just for today and there are chunks spanning today and yesterday then
+	// the overlap check would skip processing yesterdays index which would result in index pointing to deleted chunks.
```
MichelHollands, line 266:
```suggestion
	// the overlap check would skip processing yesterdays index which would result in the index pointing to deleted chunks.
```

MichelHollands (approved):
LGTM

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
```diff
@@ -199,52 +227,47 @@ func (d *DeleteRequestsManager) MarkPhaseFailed() {
...
-			)
-		} else {
-			level.Info(util_log.Logger).Log(
-				"msg", "delete request for user marked as processed",
-				"delete_request_id", deleteRequest.RequestID,
-				"user", deleteRequest.UserID,
-				"deleted_lines", deleteRequest.DeletedLines,
-			)
+	for _, userDeleteRequests := range d.deleteRequestsToProcess {
+		for _, deleteRequest := range userDeleteRequests.requests {
```
dannykopping, line 238: I think `userDeleteRequests` could be nil here; we should add a guard to prevent a panic

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
```diff
@@ -199,52 +227,47 @@ func (d *DeleteRequestsManager) MarkPhaseFailed() {
...
+				)
+			}
+			d.metrics.deleteRequestsProcessedTotal.WithLabelValues(deleteRequest.UserID).Inc()
 		}
-		d.metrics.deleteRequestsProcessedTotal.WithLabelValues(deleteRequest.UserID).Inc()
 	}
 }
 
-func (d *DeleteRequestsManager) IntervalMayHaveExpiredChunks(interval model.Interval, userID string) bool {
+func (d *DeleteRequestsManager) IntervalMayHaveExpiredChunks(_ model.Interval, userID string) bool {
```
dannykopping, line 260: If we don't need this arg, why are we passing it?

dannykopping: In general I think this method needs to be renamed since it seems it no longer deals with an interval

dannykopping (commented):
This is great Sandeep, thank you
Left a couple notes

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
sandeepsukhani: we always add a non-nil entry, did I miss anything?

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
sandeepsukhani: The function satisfies an interface which is shared with the retention code.

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
dannykopping: Sure, but later when we refactor this code and someone sets a nil value on purpose or by accident, this will panic. Better to be defensive IMHO

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
dannykopping: Gotcha!

dannykopping:
@sandeepsukhani do you have plans to finalise this during this week?

MasslessParticle:
This LGTM. I'm going to merge it. I'll add the requested `nil` in a follow PR I'm basing on this.

PR #6841, opened 2022-08-04
[TSDB] Indexgateway Protobuf refactor 

owen-d:
This PR is a no-op. It lays the ground work for future efforts to introduce TSDB index sampling on the ingesters. To do this, a few things were required to avoid circular dependencies. 

* Move a few bits from `queryrangebase` to `queryrangebase/definitions`, including both interfaces (`Request`, `Response`, `Codec`, etc) and some building block protobuf types (`Prometheus{Request,Response}Header`, `CachingOptions`). This allowed us to then remove the existing `indexgatewaypb` package and move it's code to `logproto` instead.

- [x] Tested the rollout in a cluster to ensure there aren't compatibility issues.


slim-bean (approved):
LGTM

PR #6846, opened 2022-08-05
Existing limits were low for large envs

trevorwhitney:

Existing limits were low for a large environment. In an attempt to standardize across our deployments it seemed to make most sense to change the defaults here and override when we want less, rather than having to override when we want more.


vlad-diachenko:
hey @trevorwhitney . should we also adjust limits for other components like ingester, querier, etc... ??

owen-d (approved):
LGTM. This is a followup to https://github.com/grafana/deployment_tools/pull/38494

P.S. We'll also want to remove any existing distributor sizing overrides in other cells, assuming it won't cause scale downs or we have room for such.

trevorwhitney:
@vlad-diachenko that might be a good idea, but lets table that for another PR. this is in direct response to our own cloud envs having different requests/limits, so attempting to standarize

PR #6851, opened 2022-08-08
Only process deletes in batches

MasslessParticle:
This is a setting to protect the compactor when there are many delete requests. Only process `delete_batch_size` per compaction.

The default of 70 was chosen based on a recent incident where we saw 16-hour delete requests run at a rate of ~28/minute. At that rate, 70 should take ~2.5 minutes. Even if the requests cover 4x the data, this should still keep the deletion phase of compaction to ~10 minutes.

### DIFF
--- a/pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
+++ b/pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
@@ -2,6 +2,7 @@ package deletion
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
```diff
@@ -149,6 +152,11 @@ func (d *DeleteRequestsManager) loadDeleteRequestsToProcess() error {
 		if deleteRequest.EndTime > ur.requestsInterval.End {
 			ur.requestsInterval.End = deleteRequest.EndTime
 		}
+
+		deleteCount++
+		if deleteCount >= d.batchSize {
+			break
```
trevorwhitney, line 158: so when we go over batch size, we just bail and go on to the next request? what happens to the requests that didn't get processed? are they just dropped, queued up for the next delete?

In any case I think a log message might be helpful here, to let the user know maybe how many were dropped, or just that the limit was hit, but there are still requests in the queue being skipped?

trevorwhitney (approved):
LGTM, just one small question.

owen-d (approved):

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
MasslessParticle: The requests are pulled out by status `Received`. When they're completed, the status is updated to processed. This change means that the remaining deletes wait to be processed (but filtering continues).

PR #6852, opened 2022-08-08
[TSDB] Ingester Index Sampling

owen-d:
This PR does two things:
 1) Fixes a proto package path incompatibility. Older versions (prior to https://github.com/grafana/loki/pull/6841) used the `indexgatewaypb` package. While https://github.com/grafana/loki/pull/6841 ensured the payloads were compatible, the grpc service description generated didn't match the historical `indexgatewaypb.IndexGateway`, which caused errors when trying to query the old index gateways with new queriers. This PR moves the `IndexGateway` service into it's own proto package with the name `indexgatewaypb` and generates code in the `logproto` golang package, which fixes errors during rollouts as the service descriptions now match.
 2) Wires up the ingester & ingester-querier to support querying TSDB for index statistics so we can more effectively shard data still inside the ingesters which haven't made it to storage yet.


slim-bean (approved):
LGTM

PR #6854, opened 2022-08-08
Add ability to limit the impact of pathological delete requests.

MasslessParticle:
This PR addresses the situation where a delete request has been created that it keeps compaction from ever completing:

The following changes are made
- Allow configuration of deletion modes per user rather than binary enabled/disabled
- At compaction time, filter delete requests based on tenant mode. Filtered delete requests won't count against the deletion batch so the compactor can keep making progress.
- Allow removal of pending delete requests at any time. This allows us to delete a poorly-performing delete request even after it's supposed to have been started
- Remove global deletion configuration. All configuration is now exposed through limits. By default, the limit is set to `filter-and-delete`


pkg/validation/limits.go
```diff
@@ -234,7 +235,16 @@ func (l *Limits) Validate() error {
 			l.StreamRetention[i].Matchers = matchers
 		}
 	}
-	return nil
+
+	err := fmt.Errorf("delete format must be one of 'disabled', 'filter-only', or 'filter-and-delete'")
```
MasslessParticle, line 239: Ideally we don't want this to be just strings, but importing `deletion` creates an import loop that's not trivial to unravel. If we want it, I'm happy to make a follow in PR

pkg/storage/stores/indexshipper/compactor/deletion/request_handler_test.go
```diff
@@ -16,49 +16,6 @@ import (
 	"github.com/grafana/loki/pkg/validation"
 )
 
-type retentionLimit struct {
```
MasslessParticle, line 19: Just moved these so the test's the first thing you see when entering the file

pkg/storage/stores/indexshipper/compactor/deletion/util.go
```diff
@@ -0,0 +1,40 @@
...
+func parseDeletionQuery(query string) (syntax.LogSelectorExpr, error) {
+	logSelectorExpr, err := syntax.ParseLogSelector(query, false)
+	if err != nil {
+		return nil, errInvalidQuery
+	}
+
+	return logSelectorExpr, nil
+}
+
+func validDeletionLimit(l retention.Limits, userID string) (bool, error) {
```
MasslessParticle, line 25: We ask about deletion mode wrt tenants a lot now so I dumped things into here

pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
```diff
@@ -21,19 +20,17 @@ const deletionNotAvailableMsg = "deletion is not available for this tenant"
 
 // DeleteRequestHandler provides handlers for delete requests
 type DeleteRequestHandler struct {
-	deleteRequestsStore       DeleteRequestsStore
-	metrics                   *deleteRequestHandlerMetrics
-	limits                    retention.Limits
-	deleteRequestCancelPeriod time.Duration
+	deleteRequestsStore DeleteRequestsStore
```
MasslessParticle, line 23: Remove the cancellation period from the API. This allows us to delete any pending request at any time.

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -240,16 +227,8 @@ func (c *Compactor) init(objectClient client.ObjectClient, schemaConfig config.S
 			return err
 		}
 
-		if c.deleteMode.DeleteEnabled() {
```
MasslessParticle, line 243: Deletes are always initialized, now. It's up to overrides to gate access

pkg/loki/modules.go
```diff
@@ -621,15 +621,6 @@ func (t *Loki) initQueryFrontendTripperware() (_ services.Service, err error) {
 }
 
 func (t *Loki) cacheGenClient() (generationnumber.CacheGenClient, error) {
```
MasslessParticle, line 623: The cache gen client won't surface errors to the user if deletion is disabled. The behavior will be unchanged from what happens now when deletion is globally enabled but disabled for a user.

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager_test.go
```diff
@@ -339,12 +339,6 @@ func TestDeleteRequestsManager_Expired(t *testing.T) {
 				isExpired:           false,
 				nonDeletedIntervals: nil,
 			},
-			expectedDeletionRangeByUser: map[string]model.Interval{
```
MasslessParticle, line 342: This is the interesting change in behavior: before, we always created one overarching request of intervals that might be affected but filtering deletes at the load stage means that we know there aren't any potential `Expired` chunks for this user

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -97,7 +96,6 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.DeleteBatchSize, "boltdb.shipper.compactor.delete-batch-size", 70, "The max number of delete requests to run per compaction cycle.")
 	f.DurationVar(&cfg.DeleteRequestCancelPeriod, "boltdb.shipper.compactor.delete-request-cancel-period", 24*time.Hour, "Allow cancellation of delete request until duration after they are created. Data would be deleted only after delete requests have been older than this duration. Ideally this should be set to at least 24h.")
 	f.IntVar(&cfg.MaxCompactionParallelism, "boltdb.shipper.compactor.max-compaction-parallelism", 1, "Maximum number of tables to compact in parallel. While increasing this value, please make sure compactor has enough disk space allocated to be able to store and compact as many tables.")
-	f.StringVar(&cfg.DeletionMode, "boltdb.shipper.compactor.deletion-mode", "disabled", fmt.Sprintf("Deletion mode. Can be one of %v", strings.Join(deletion.AllModes(), "|")))
```
slim-bean, line 100: We will need some notes for this in the upgrade guide, also for removing it from yaml above.

pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
slim-bean: I think this comment is attached to the wrong place?

pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
```diff
@@ -174,11 +171,6 @@ func (dm *DeleteRequestHandler) cancelDeleteRequestHandler(w http.ResponseWriter
 		return
 	}
 
-	if deleteRequest.CreatedAt.Add(dm.deleteRequestCancelPeriod).Before(model.Now()) {
```
slim-bean, line 177: I think the comment about 'Remove the cancellation period from the API' was probably originally here?

pkg/validation/limits.go
```diff
@@ -196,7 +197,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	_ = l.QuerySplitDuration.Set("30m")
 	f.Var(&l.QuerySplitDuration, "querier.split-queries-by-interval", "Split queries by an interval and execute in parallel, 0 disables it. This also determines how cache keys are chosen when result caching is enabled")
 
-	f.BoolVar(&l.CompactorDeletionEnabled, "compactor.allow-deletes", false, "Enable access to the deletion API.")
+	f.StringVar(&l.DeletionMode, "compactor.deletion-mode", "filter-and-delete", "Set the deletion mode for the user.")
```
slim-bean, line 200: The previous value for this when it was a compactor config was "disabled". I believe we still mark this functionality as experimental so I wonder if we should keep the same `disabled` value until we are ready to move this feature to a non-experimental place?

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
```diff
@@ -130,6 +130,13 @@ func (d *DeleteRequestsManager) loadDeleteRequestsToProcess() error {
 			continue
 		}
 
+		if ok, err := d.shouldProcessRequest(deleteRequest); !ok {
+			if err != nil {
+				return err
+			}
+			continue
+		}
```
MichelHollands, line 138:
```suggestion
		ok, err := d.shouldProcessRequest(deleteRequest);
		if err != nil {
			return err
		}
		if !ok {
			continue
		}
```

This looks cleaner.

pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
MasslessParticle: It was just deleting the property on the type. This is where the check gets removed

pkg/storage/stores/indexshipper/compactor/compactor.go
MasslessParticle: addressed this in the latest commits

integration/loki_micro_services_delete_test.go
```diff
@@ -23,10 +23,9 @@ func TestMicroServicesDeleteRequest(t *testing.T) {
 			"-target=compactor",
 			"-boltdb.shipper.compactor.compaction-interval=1s",
 			"-boltdb.shipper.compactor.retention-delete-delay=1s",
-			"-boltdb.shipper.compactor.deletion-mode=filter-and-delete",
-			// By default a minute is added to the delete request start time. This compensates for that.
+			// By default, a minute is added to the delete request start time. This compensates for that.
 			"-boltdb.shipper.compactor.delete-request-cancel-period=-60s",
-			"-compactor.allow-deletes=true",
+			"-boltdb.shipper.compactor.deletion-mode=filter-and-delete",
```
owen-d, line 28: Can we remove the prefix `boltdb.shipper`? Presumably it will need to work with TSDB indices in the future and I think this will confuse users.

pkg/validation/limits.go
owen-d: I'm afraid of how easy this is to break when we add new deletion modes but forget to update it here. It should be fairly easy to add a new deletionmode package that just has the `DeletionMode` types we need and can be imported.

owen-d (commented):
A few suggestions, but looking good!

owen-d (approved):

PR #6856, opened 2022-08-09
chore(makefile): Remove `NETGO` check after building loki binaries.

kavirajk:

Encountered after upgrade to go1.19 in local breaks this assumption. Meaning, even if your loki binary built with `-tags netgo` that file `cgo_stub.go` doesn't exist in the binary. And any `make <binary>` keep failing.

```
 \nYour go standard library was built without the 'netgo' build tag.
To fix that, run
    sudo go clean -i net
    sudo go install -tags netgo std
make: *** [Makefile:149: cmd/loki/loki] Error 1
```

But even those commands not helping. Looks like this check was added by Tom at 2018.
tbh, checking for filename inside the binary is too much internal details that is subjected to change (which is what happened here I think).

NOTE: Had some internal discussions and confirmed that we don't need this check in recent Go versions anymore.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


MichelHollands (approved):
LGTM

PR #6860, opened 2022-08-09
Add delete api validations

MasslessParticle:
To help protect against pathological delete scenarios, this PR adds the following validation to the Delete API:
- Ensure a start date is present
- Ensure the timestamp format is either 10-digit Unix seconds or RFC3339

Additionally, this PR moves the delete middleware to it's own file. This is just so validation testing could be done without setting up limits.

### DIFF
--- a/pkg/storage/stores/indexshipper/compactor/deletion/tenant_request_handler.go
+++ b/pkg/storage/stores/indexshipper/compactor/deletion/tenant_request_handler.go
@@ -17,18 +17,15 @@ func TenantMiddleware(limits retention.Limits, next http.Handler) http.Handler {
 			return
 		}
 
-		allLimits := limits.AllByUserID()
-		userLimits, ok := allLimits[userID]
-		if ok {
-			if !userLimits.CompactorDeletionEnabled {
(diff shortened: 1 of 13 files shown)


pkg/storage/stores/indexshipper/compactor/deletion/tenant_request_handler.go
```diff
@@ -0,0 +1,36 @@
...
+func TenantMiddleware(limits retention.Limits, next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		userID, err := tenant.TenantID(ctx)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		allLimits := limits.AllByUserID()
```
owen-d, line 20: nit: we have a convention of defining methods on limit structs themselves. This is easier to consume as it doesn't require much boilerplate when we check a limit and lets us define smaller interfaces over just the methods we need.

owen-d (approved):
One suggestion, but LGTM

PR #6869, opened 2022-08-10
update deletion docs

owen-d:
Followup to https://github.com/grafana/loki/pull/6854 which fixes documentation references.

### DIFF
--- a/docs/sources/operations/storage/logs-deletion.md
+++ b/docs/sources/operations/storage/logs-deletion.md
@@ -28,4 +28,4 @@ With `filter-and-delete`, log lines matching the query in the delete request are
 
 A delete request may be canceled within a configurable cancellation period. Set the `delete_request_cancel_period` in the compactor's YAML configuration or on the command line when invoking Loki. Its default value is 24h.
 
-Access to the deletion API can be enabled per tenant via the `compactor.retention_enabled` setting, which must be `true`.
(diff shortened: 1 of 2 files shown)


docs/sources/operations/storage/logs-deletion.md
```diff
@@ -28,4 +28,4 @@ With `filter-and-delete`, log lines matching the query in the delete request are
 
 A delete request may be canceled within a configurable cancellation period. Set the `delete_request_cancel_period` in the compactor's YAML configuration or on the command line when invoking Loki. Its default value is 24h.
 
-Access to the deletion API can be enabled per tenant via the `allow_deletes` setting.
+Access to the deletion API can be enabled per tenant via the `compactor.retention_enabled` setting, which must be `true`.
```
MasslessParticle, line 31: We should probably say something like:

> Access to the deletion API can be enabled per tenant via the `compactor.deletion_mode` runtime configuration. 

`retention_enabled` is covered further up in the doc so I think we're covered ther

owen-d, line 31:
```suggestion
As long as the `compactor.retention_enabled` setting is `true`, the API endpoints will be available. Afterwards, access to the deletion API can be enabled per tenant via the `deletion_mode` tenant override.
```

MasslessParticle (approved):

PR #6870, opened 2022-08-10
operator: Configure gateway to honor the global TLS security profile

aminesnow:
This PR enables the gateway to be configured with the minimum TLS version and ciphers defined in the global apiservers configuration on the OpenShift Platform.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -240,7 +241,19 @@ func CreateOrUpdateLokiStack(
...
+			return &status.DegradedError{
+				Message: "Missing apiserver object",
+				Reason:  lokiv1.ReasonMissingOpenShiftAPIServer,
+				Requeue: false,
+			}
+		}
+		return kverrors.Wrap(err, "failed to lookup apiServer")
+	}
+
+	objects, err := manifests.BuildAll(opts, apiServer.Spec.TLSSecurityProfile)
```
Red-GV, line 256: I wonder if the better option here is to expand the `Options` structure to also have an optional TLS configuration pieces from the `apiServer` object. 

Also, returning an error here causes an issue in base K8s, no? We should only return a degraded error if we are on an OpenShift cluster.

operator/internal/manifests/gateway_tenants_test.go
```diff
@@ -259,6 +259,7 @@ func TestConfigureDeploymentForMode(t *testing.T) {
 										fmt.Sprintf("--web.healthchecks.url=https://localhost:%d", gatewayHTTPPort),
 										"--tls.client-auth-type=NoClientCert",
 										"--tls.min-version=VersionTLS12",
+										"--tls.cipher-suites=TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
```
Red-GV, line 262: Maybe move this to some helper function to reference rather than put the entire list here? Perhaps there's a way to grab all the cipher-suites and filter them by tls version allowed?

Red-GV (commented):
Looks good so far!

A few comments on this one:

operator/internal/handlers/lokistack_create_or_update.go
xperimental: I'd agree that adding it to `Options` is probably a better deal. We can also convert the `TLSSecurityProfile` to the information we actually need (minimum TLS version and cipher list) before passing it into the manifest generation.

The request to get the APIServer resource can only be done on OpenShift and will fail on other clusters. But even on OpenShift it can not be assumed that the TLSSecurityProfile is configured. This leads to the question what a sensible fallback is, I'd say either we go with the intermediate profile like the PR implementation already does or we skip configuring the TLS version and ciphers like we did before.

operator/internal/manifests/gateway_tenants_test.go
xperimental: An alternative would be to configure some "dummy values" just for testing.

periklis (changes requested):
This is a great start into supporting a TLS profile for our managed components in Loki. This work is really awesome :rocket: 

In general the code looks clear and concise for me. However, we need to polish out this feature for k8s first and make the automation piggypacking on `openshiftv1.ApiServers` CRD only optional. Thus I suggest to address this like this:
- First let's add a configuration option in the operator configuration for selecting one of the three named profiles. All three are general names from Mozilla's work on groups TLS versions and cipher suites.
- Add a feature flag for OpenShift that overwrites per default the operator config for TLSProfile with the values from the APIServer CRD.
- In general pass anything from the CRD or the operator configuration to the custom `manifest.Options` with new fields and minimize the exposure to optional CRDs.

WDYT? (cc @xperimental @Red-GV )

periklis (changes requested):
This looks way better and more decoupled than the first version. However I suggest to decouple from using enumerations from the openshift/api package to remove the burden of maintaining version compatibility in general (as per both our code base and openshift/api depend on k8s). 

I suggest simply to create an enumeration for three TLS Profiles (https://wiki.mozilla.org/Security/Server_Side_TLS) and use it across our code base. In addition I suggest to replace `DisableAPIServerTLSProfile` with a simple `TLSProfile` field so non-openshift users can declare what profile they want to choose directly.

aminesnow:
@periklis Thanks for the feedback! 
There is actually a 4th enum value `Custom`, and it allows the user to define a custom profile. I would like to keep supporting it in addition to the 3 main ones.
Also, `DisableAPIServerTLSProfile` is defined in the openshift options because by default, when using openshift, we get the profile from the api server, and this allows the user to disable it so that they can define their own setting. So I'm confused by your comment regarding this..

periklis:
> @periklis Thanks for the feedback! There is actually a 4th enum value `Custom`, and it allows the user to define a custom profile. I would like to keep supporting it in addition to the 3 main ones. 

Supporting `Custom` would require to add a set of two fields one for TLSMinVersion and one for TLSCipherSuites in the operator project config. I would like to skip this extra options if nobody asks for them. 

> Also, `DisableAPIServerTLSProfile` is defined in the openshift options because by default, when using openshift, we get the profile from the api server, and this allows the user to disable it so that they can define their own setting. So I'm confused by your comment regarding this..

An Options with `DisableAPIServer...` as a prefix is quite confusing for normal k8s users. I even doubt if this feature is well known and understood in the OpenShift ecosystem. Thus I would like to lower the confusion by making this field the input enumeration to select one of the three public listed profiles (See link to Mozilla). Our project config for OpenShift users will let this field empty and use the `ApiServer` CRD to fetch the profile.

aminesnow:
@periklis understood, thanks for the explanation!

operator/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - [6923](https://github.com/grafana/loki/pull/6923) **xperimental**: Reconcile owner reference for existing objects
 - [6907](https://github.com/grafana/loki/pull/6907) **Red-GV**: Adding valid subscription annotation to operator metadata
+- [6870](https://github.com/grafana/loki/pull/6870) **aminesnow**: Configure gateway to honor the global tlsSecurityProfile on Openshift
```
periklis, line 5:
```suggestion
- [6870](https://github.com/grafana/loki/pull/6870) **aminesnow**: Configure gateway to honor the global tlsSecurityProfile on Openshift
- [6923](https://github.com/grafana/loki/pull/6923) **xperimental**: Reconcile owner reference for existing objects
- [6907](https://github.com/grafana/loki/pull/6907) **Red-GV**: Adding valid subscription annotation to operator metadata
```

operator/go.mod
```diff
@@ -25,6 +25,7 @@ require github.com/ViaQ/logerr/v2 v2.0.0
 require (
 	github.com/google/go-cmp v0.5.7
 	github.com/grafana/loki v1.6.2-0.20220708124813-b92f113cb096
+	github.com/openshift/library-go v0.0.0-20220810125113-79fc5d549223
```
periklis, line 28: I suggest to pin this to `release-4.11` as the rest of our openshift dependencies.

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -240,6 +243,35 @@ func CreateOrUpdateLokiStack(
...
+		if apiServer.Spec.TLSSecurityProfile != nil {
+			tlsProfile = *apiServer.Spec.TLSSecurityProfile
+		}
+	}
+
+	tlsMinVersion, ciphers := getSecurityProfileInfo(&tlsProfile)
+	opts.TLSProfileSpec = manifests.TLSProfileSpec{
+		MinTLSVersion: tlsMinVersion,
+		Ciphers:       ciphers,
+	}
```
periklis, line 273: Can we extract this in a separate file under `operator/internal/handlers/internal/tlsprofile` like we do for rules and other companion objects?

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -310,3 +342,29 @@ func isNamespaceScoped(obj client.Object) bool {
 		return true
 	}
 }
+
+func getSecurityProfileInfo(profile *openshiftv1.TLSSecurityProfile) (string, []string) {
```
periklis, line 346: Same as above.

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -221,6 +228,10 @@ func TestCreateOrUpdateLokiStack_SetsNamespaceOnAllObjects(t *testing.T) {
 			k.SetClientObject(out, &defaultGatewaySecret)
 			return nil
 		}
+		if apiServer.Name == name.Name {
+			k.SetClientObject(out, &apiServer)
+			return nil
+		}
```
periklis, line 234: Why do we need for the general usecase the `openshiftv1.APIServer` CR here? This should work without it.

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -307,6 +318,10 @@ func TestCreateOrUpdateLokiStack_SetsOwnerRefOnAllObjects(t *testing.T) {
 			k.SetClientObject(object, &defaultGatewaySecret)
 			return nil
 		}
+		if apiServer.Name == name.Name {
+			k.SetClientObject(object, &apiServer)
+			return nil
+		}
```
periklis, line 324: Same here.

operator/internal/manifests/gateway.go
```diff
@@ -50,7 +50,8 @@ func BuildGateway(opts Options) ([]client.Object, error) {
 
 	if opts.Stack.Tenants != nil {
 		mode := opts.Stack.Tenants.Mode
-		if err := configureGatewayDeploymentForMode(dpl, mode, opts.Gates, opts.Name, opts.Namespace); err != nil {
+		if err := configureGatewayDeploymentForMode(dpl, mode, opts.Gates, opts.Name, opts.Namespace,
+			opts.TLSProfileSpec.MinTLSVersion, opts.TLSProfileSpec.Ciphers); err != nil {
```
periklis, line 54:
```suggestion
		minTLSVersion := opts.TLSProfileSpec.MinTLSVersion
		ciphers := opts.TLSProfileSpec.Ciphers
		if err := configureGatewayDeploymentForMode(dpl, mode, opts.Gates, opts.Name, opts.Namespace, minTLSVersion, ciphers); err != nil {
```

operator/internal/manifests/options.go
```diff
@@ -34,6 +37,32 @@ type Options struct {
...
+	// TLSProfileOldType is a TLS security profile based on:
+	// https://wiki.mozilla.org/Security/Server_Side_TLS#Old_backward_compatibility
+	TLSProfileOldType TLSProfileType = "Old"
+	// TLSProfileIntermediateType is a TLS security profile based on:
+	// https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28default.29
+	TLSProfileIntermediateType TLSProfileType = "Intermediate"
+	// TLSProfileModernType is a TLS security profile based on:
+	// https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility
+	TLSProfileModernType TLSProfileType = "Modern"
+)
```
periklis, line 54: I believe this enumeration is better put in the `operator/apis/config/v1/projectconfig_types.go`

periklis (changes requested):

operator/internal/handlers/internal/tlsprofile/tlsprofile.go
```diff
@@ -0,0 +1,71 @@
+package tlsprofile
+
+import (
+	"context"
+
+	"github.com/ViaQ/logerr/v2/kverrors"
+	projectconfig_types "github.com/grafana/loki/operator/apis/config/v1"
```
periklis, line 7: Same here why not simply `projectconfigv1`?

operator/internal/handlers/internal/tlsprofile/tlsprofile.go
```diff
@@ -0,0 +1,71 @@
...
+		tlsProfile = openshiftv1.TLSSecurityProfile{
+			Type: openshiftv1.TLSProfileType(tlsProfileType),
+		}
+	} else {
+		tlsProfile = openshiftv1.TLSSecurityProfile{
+			Type: openshiftv1.TLSProfileIntermediateType,
+		}
+
+		var apiServer openshiftv1.APIServer
+		if err := k.Get(ctx, client.ObjectKey{Name: "cluster"}, &apiServer); err != nil {
```
periklis, line 29: Let's make the name a constant to make this visible that we support a single apiserver.

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -7,12 +7,14 @@ import (
 	"time"
 
 	configv1 "github.com/grafana/loki/operator/apis/config/v1"
+	projectconfig_types "github.com/grafana/loki/operator/apis/config/v1"
```
periklis, line 10: Same here.

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -85,6 +86,12 @@ var (
 		},
 		Data: map[string]string{},
 	}
+
+	apiServer = openshiftv1.APIServer{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cluster",
+		},
+	}
```
periklis, line 94: This seems not used anymore and we can delete it here

operator/internal/manifests/options.go
```diff
@@ -2,6 +2,7 @@ package manifests
 
 import (
 	configv1 "github.com/grafana/loki/operator/apis/config/v1"
+	projectconfig_types "github.com/grafana/loki/operator/apis/config/v1"
```
periklis, line 5: Same here.

operator/internal/manifests/gateway_tenants.go
```diff
@@ -55,7 +55,11 @@ func ApplyGatewayDefaultOptions(opts *Options) error {
 	return nil
 }
 
-func configureGatewayDeploymentForMode(d *appsv1.Deployment, mode lokiv1.ModeType, fg configv1.FeatureGates, stackName, stackNs string) error {
```
periklis, line 58: AFAICS we configure the TLS minVersion and ciphers only for OpenShift here. However we support per definition of the feature gate this feature for non-openshift tenants too. Thus this feature gate impact on the lokistack-gateway and opa-openshift container does not belong here. IMHO this belongs in the `manifests/gateway.go` eg as we configure the HTTP PKI for example.

operator/internal/handlers/internal/tlsprofile/tlsprofile.go
```diff
@@ -0,0 +1,71 @@
...
+	projectconfig_types "github.com/grafana/loki/operator/apis/config/v1"
+	"github.com/grafana/loki/operator/internal/external/k8s"
+	openshiftv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/crypto"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetSecurityProfileInfo gets the tls profile info to apply.
+func GetSecurityProfileInfo(ctx context.Context, k k8s.Client, tlsProfileType projectconfig_types.TLSProfileType) (projectconfig_types.TLSProfileSpec, error) {
```
periklis, line 16: Please add some tests for this functions to safe guard that we extract the spec correctly for the supported cases.

periklis (changes requested):

operator/cmd/loki-broker/main.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/ViaQ/logerr/v2/log"
 	"github.com/go-logr/logr"
 	configv1 "github.com/grafana/loki/operator/apis/config/v1"
+	projectconfig_types "github.com/grafana/loki/operator/apis/config/v1"
```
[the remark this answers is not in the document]
aminesnow: I agree

operator/internal/handlers/internal/tlsprofile/tlsprofile.go
aminesnow: was just working on that

operator/internal/manifests/gateway_tenants.go
aminesnow: I see. I must have misunderstood what the object was.

operator/internal/manifests/gateway_test.go
```diff
@@ -279,3 +279,44 @@ func TestBuildGateway_WithExtraObjectsForTenantMode_ReplacesIngressWithRoute(t *
...
+		},
+		Stack: lokiv1.LokiStackSpec{
+			Template: &lokiv1.LokiTemplateSpec{
+				Gateway: &lokiv1.LokiComponentSpec{
+					Replicas: rand.Int31(),
+				},
+			},
+			Tenants: &lokiv1.TenantsSpec{
+				Mode: lokiv1.OpenshiftLogging,
+			},
```
periklis, line 304: IMHO using per default `Tenants.Mode: OpenShiftLogging` is misleading for our code base. Because it emits both required containers it does not make it a good test case. I suggest to make a test table with tenant modes and drop the following assert for this test as we have an other test case namely `TestBuildGateway_HasConfigForTenantMode`:
```go
require.Len(t, d.Spec.Template.Spec.Containers, 2)
```

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -85,6 +87,12 @@ var (
 		},
 		Data: map[string]string{},
 	}
+
+	apiServer = openshiftv1.APIServer{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: tlsprofile.APIServerName,
+		},
+	}
```
periklis, line 95: Do we still need this? Looks unused.

operator/internal/handlers/internal/tlsprofile/tlsprofile_test.go
```diff
@@ -0,0 +1,120 @@
...
+			k.SetClientObject(object, &apiServer)
+			return nil
+		}
+		return apierrors.NewNotFound(schema.GroupResource{}, "something wasn't found")
+	}
+
+	k.StatusStub = func() client.StatusWriter { return sw }
+
+	return k
+}
```
periklis, line 120: I suggest not to create such helpers as the test cases should IMHO code wide isolated from each other. If you want to write the client once the pattern is to write a test table instead of three test functions.

periklis (commented):
Minor suggestions and we are good to go

operator/internal/handlers/lokistack_create_or_update_test.go
aminesnow: forgot to delete it

periklis (approved):

PR #6872, opened 2022-08-10
Remove unused code

MasslessParticle:
An interface for deletion limits was added that means we don't need to use `retention.Limits` here anymore. It also means we don't have to convert between limit types in the test anymore


MichelHollands (approved):
LGTM

PR #6879, opened 2022-08-10
Add timeouts to deletions 

MasslessParticle:
To help protect the system from a situation where deletes take too long and block compaction, this PR introduces timeouts to the deletion process.

As part of retention, all chunks for an affected timerange are scanned and any active deletion filters are applied. This change adds a context to that loop so that it can time out. 

The timeout may not occur exactly when the context expires. Instead, this change prioritizes changing entire chunks so that progress is made even if the timeout expires.

This change adds a `MarkPhaseTimedOut` method to `Expiration Checkers` so they can take the appropriate actions. In this case, the list of deletes to be processed are cleared which prevents `MarkPhaseFinished` from marking them as `Processed`. This ensures deletes will be retried the next time compaction happens.

There is the chance that a set of deletes will never complete so this change adds a `loki_compactor_delete_processing_fails_total` metric that records how many failures occurred and a `cause` that is either `error` or `timeout`

### DIFF
--- a/pkg/storage/stores/indexshipper/compactor/retention/retention.go
+++ b/pkg/storage/stores/indexshipper/compactor/retention/retention.go
@@ -85,14 +85,14 @@ type Marker struct {
 	markTimeout      time.Duration
 }
 
-func NewMarker(workingDirectory string, expiration ExpirationChecker, chunkClient client.Client, r prometheus.Registerer) (*Marker, error) {
+func NewMarker(workingDirectory string, expiration ExpirationChecker, markTimeout time.Duration, chunkClient client.Client, r prometheus.Registerer) (*Marker, error) {
 	metrics := newMarkerMetrics(r)
 	return &Marker{
 		workingDirectory: workingDirectory,
 		expiration:       expiration,
 		markerMetrics:    metrics,
 		chunkClient:      chunkClient,
-		markTimeout:      15 * time.Minute,
+		markTimeout:      markTimeout,
 	}, nil
 }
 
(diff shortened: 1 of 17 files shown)


pkg/storage/stores/indexshipper/compactor/retention/util_test.go
```diff
@@ -121,10 +121,11 @@ type table struct {
 	chunks map[string][]chunk.Chunk
 }
 
-func (t *table) ForEachChunk(callback ChunkEntryCallback) error {
+func (t *table) ForEachChunk(ctx context.Context, callback ChunkEntryCallback) error {
 	for userID, chks := range t.chunks {
 		i := 0
-		for _, chk := range chks {
+		for j := 0; j < len(chks) && ctx.Err() == nil; j++ {
```
MasslessParticle, line 127: It's not ideal to change this in the test. The actual implementations include tests and this ensures that we can test the behavior around the timeout in `retention.go`

sandeepsukhani (commented):
Overall the changes look fine to me. The problem though is that the timeout is being applied on retention as well, but the flag only mentions deletion. We should either change it or update the flag and docs to include both deletion and custom retention honouring the timeout?

Also, the timeout is per table and not across all the tables, which should be called out too.

pkg/storage/stores/indexshipper/compactor/retention/retention.go
```diff
@@ -204,7 +218,13 @@ func markforDelete(ctx context.Context, tableName string, marker MarkerStorageWr
 		return false, nil
 	})
 	if err != nil {
-		return false, false, err
+		if errors.Is(err, context.DeadlineExceeded) && errors.Is(iterCtx.Err(), context.DeadlineExceeded) {
+			// Deletes timed out. Don't return an error so compaction can continue and deletes can be retried
+			level.Warn(util_log.Logger).Log("msg", "Timed out while running delete")
```
MichelHollands, line 223: Does this already have the userID and or delete request ID (perhaps using a custom logger)? If not please add it.

MichelHollands (approved):
LGTM

PR #6880, opened 2022-08-11
Store request latency metrics

sandeepsukhani:
We currently only have index request latency metrics on individual store types i.e boltdb-shipper, tsdb, bigtable etc. This makes it hard to instrument dashboards when we change the store type. This PR adds a metric to measure the overall index request latency irrespective of the underlying store type.

* The metric tracks the latency across all the stores, so if a query spans multiple stores, it measures the total latency from all the stores. A single query would rarely span multiple stores after splitting, so it should not be a big deal.
* The metric only measures index query latency and not the write latency. For measuring just the write latency, I need to refactor some code so left a todo for it and will do a follow up PR.

### DIFF
--- a/pkg/storage/stores/composite_store.go
+++ b/pkg/storage/stores/composite_store.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"sort"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
+	"github.com/weaveworks/common/instrument"
 
 	"github.com/grafana/loki/pkg/storage/chunk"
 	"github.com/grafana/loki/pkg/storage/chunk/fetcher"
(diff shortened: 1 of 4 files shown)


pkg/storage/stores/composite_store.go
```diff
@@ -93,6 +96,19 @@ func (c compositeStore) SetChunkFilterer(chunkFilter chunk.RequestChunkFilterer)
 }
 
 func (c compositeStore) GetSeries(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]labels.Labels, error) {
+	var lbls []labels.Labels
+	if err := instrument.CollectedRequest(ctx, "series", instrument.NewHistogramCollector(c.metrics.indexQueryLatency), instrument.ErrorCode, func(ctx context.Context) error {
```
MichelHollands, line 100: TIL instrument.CollectedRequest: useful to know

MichelHollands (approved):
LGTM

PR #6882, opened 2022-08-11
[bugfix] fixes issues sharding series by storageref vs fingerprint in…

owen-d:
… stripeSeries

I noticed a `concurrent map read/write` panic coming from our `stripeSeries` code at
```
github.com/grafana/loki/pkg/storage/stores/tsdb.(*stripeSeries).getByID(0xc0f7f52a00, 0x2?)
```

Reading over the code, I realized we were using the same lock index during `Append` calls, which is incorrect. To fix this problem and make the code a bit easier to work with, I embedded `RWLock`s in both the `seriesHashmap` and the `memSeriesMap`. This way, we don't reuse the same locks for both purposes, as the `seriesHashmap`s are sharded by `fingerprint` whereas the `memSeriesMap`s are sharded by `StorageRef`.


sandeepsukhani (approved):
LGTM

PR #6886, opened 2022-08-12
Add error detail to error message if available

simonswine:
I am think it might be helpful making the error detail a bit more promiment, but the first step is #6890.


MasslessParticle:
Closing due to inactivity

PR #6903, opened 2022-08-12
Shard deletes requests based on time ranges

MasslessParticle:
### Overview
In order to protect the compactor for long-running deletes, this PR introduces the ability to break delete requests into multiple requests of a predetermined time range. The time range can be configured globally or on a per-request basis with a `max_query_range` parameter. The API always allows you to interact with the requests entered even if they're sharded as an implementation detail.

This change also modifies the way delete requests are processed. Before they were processed in the order they were returned from the underlying store. Now they are first sorted by `start time`. Batches of shorter requests, grouped by time range, should mean that any given batch touches as few chunks as possible.

### Delete Store Changes:
- This change introduces the concept of a `delete request group`: a group of delete requests that all share the same request id. This allows for the one-to-many mapping required to make this feature work. All get/remove operations are now in terms of `DeleteRequestGroups`
- This change introduces a none-to-many mapping of delete request details to delete request. This is achieved by appending a `Sequence Number` to [hash key](https://github.com/grafana/loki/blob/a02ce69a98e7ffe498078324df891affdfacaafe/pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_store.go#L102) of delete requests.
- To maintain backwards comparability with delete-requests stored before this change, the sequence number is ignored when it is 0. This means that all single-request entries made before this change will be looked up by the same key they were entered with.
- Deletes added after this change won't be compatible with the old delete store but all sharded deletes are available via the API so they can be reconstructed if a downgrade is necessary.


docs/sources/configuration/_index.md
```diff
@@ -2105,6 +2105,14 @@ compacts index shards to more performant forms.
 # CLI flag: -boltdb.shipper.compactor.delete-request-cancel-period
 [delete_request_cancel_period: <duration> | default = 24h]
 
+# Constrain the size of any one delete request. When a delete request > delete_max_query_range
+# is input, the request is sharded into smaller requests of no more than delete_max_query_range
+# 0 means no max_query_period.
+#
+# Default is 0
```
DylanGuedes, line 2112: nit: I think you can remove it since it is shown in line 2114 already
```suggestion
```

docs/sources/configuration/_index.md
```diff
@@ -2105,6 +2105,14 @@ compacts index shards to more performant forms.
 # CLI flag: -boltdb.shipper.compactor.delete-request-cancel-period
 [delete_request_cancel_period: <duration> | default = 24h]
 
+# Constrain the size of any one delete request. When a delete request > delete_max_query_range
+# is input, the request is sharded into smaller requests of no more than delete_max_query_range
```
DylanGuedes, line 2109:
```suggestion
# is input, the request is sharded into smaller requests of no more than delete_max_query_range.
```

docs/sources/configuration/_index.md
```diff
@@ -2105,6 +2105,14 @@ compacts index shards to more performant forms.
 # CLI flag: -boltdb.shipper.compactor.delete-request-cancel-period
 [delete_request_cancel_period: <duration> | default = 24h]
 
+# Constrain the size of any one delete request. When a delete request > delete_max_query_range
```
DylanGuedes, line 2108: nit (obs: my english is bad so if that's correct just ignore this :rofl: )
```suggestion
# Constrain the size of a delete request. When a delete request > delete_max_query_range
```

pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
```diff
@@ -81,6 +90,63 @@ func (dm *DeleteRequestHandler) AddDeleteRequestHandler(w http.ResponseWriter, r
 	w.WriteHeader(http.StatusNoContent)
 }
 
+func shardDeleteRequestsByQueryRange(startTime, endTime model.Time, query, userID string, queryRange time.Duration) []DeleteRequest {
+	var deleteRequests []DeleteRequest
```
DylanGuedes, line 94: nit: small optimization
```suggestion
	deleteRequests := make([]DeleteRequest, 0, (endTime.Sub(startTime))/queryRange))
```

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_store.go
```diff
@@ -63,60 +64,100 @@ func NewDeleteStore(workingDirectory string, indexStorageClient storage.Client)
...
 func (ds *deleteRequestsStore) Stop() {
 	ds.indexClient.Stop()
 }
 
-// AddDeleteRequest creates entries for a new delete request.
-func (ds *deleteRequestsStore) AddDeleteRequest(ctx context.Context, userID string, startTime, endTime model.Time, query string) error {
-	_, err := ds.addDeleteRequest(ctx, userID, model.Now(), startTime, endTime, query)
-	return err
-}
+// AddDeleteRequestGroup creates entries for new delete requests. All passed delete requests will be associeted to
```
DylanGuedes, line 77:
```suggestion
// AddDeleteRequestGroup creates entries for new delete requests. All passed delete requests will be associated to
```

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_store.go
```diff
@@ -63,60 +64,100 @@ func NewDeleteStore(workingDirectory string, indexStorageClient storage.Client)
...
+	return fmt.Sprintf("%s:%s:%d", userID, requestID, sequenceNumber)
+}
+
+func (ds *deleteRequestsStore) generateID(ctx context.Context, req DeleteRequest) ([]byte, error) {
+	requestID := generateUniqueID(req.UserID, req.Query)
+
+	for {
+		_, err := ds.GetDeleteRequestGroup(ctx, req.UserID, string(requestID))
+		if err != nil {
+			if err == ErrDeleteRequestNotFound {
```
DylanGuedes, line 152: hmmm why is `ErrDeleteRequestNotFound` handled differently here?

DylanGuedes (approved):
lgtm, just a few nits

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_store.go
MasslessParticle: The job of this is to ensure that there are no duplicates. `ErrDeleteRequestNotFound` means we're good to proceed while any other error means that something broke.

docs/sources/configuration/_index.md
MasslessParticle: Yours is way clearer, thanks!

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_store.go
MasslessParticle: I've restructured it a little bit in the hope that it is a little clearer.

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_store.go
```diff
@@ -63,63 +64,100 @@ func NewDeleteStore(workingDirectory string, indexStorageClient storage.Client)
...
 
 	// create a gen number for this result
-	writeBatch.Add(DeleteRequestsTableName, fmt.Sprintf("%s:%s", cacheGenNum, userID), []byte{}, generateCacheGenNumber())
+	writeBatch.Add(DeleteRequestsTableName, fmt.Sprintf("%s:%s", cacheGenNum, req.UserID), []byte{}, generateCacheGenNumber())
+}
 
-	err := ds.indexClient.BatchWrite(ctx, writeBatch)
-	if err != nil {
-		return nil, err
+func deleteRequestHash(userID, requestID string, sequenceNumber int64) string {
```
sandeepsukhani, line 136: The first request in the new sharded requests won't have a sequence number. I think we should either make it uniform or make it explicitly clear here that it is intentional.

sandeepsukhani (commented):
I have left some minor suggestions, but overall, the code looks great to me. Good job on the tests!

pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
```diff
@@ -81,6 +90,59 @@ func (dm *DeleteRequestHandler) AddDeleteRequestHandler(w http.ResponseWriter, r
...
+	if queryRange > dm.maxQueryRange && dm.maxQueryRange != 0 {
+		dur, err := time.ParseDuration(dm.maxQueryRange.String())
+		if err != nil {
+			level.Error(util_log.Logger).Log("msg", "error parsing query range", "err", err)
+			return 0, err
+		}
+		return 0, fmt.Errorf("query range can't be greater than %s", dur.String())
+	}
+
+	return queryRange, nil
```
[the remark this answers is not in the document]
ashwanthgoli: I guess it doesn't really matter looking at the sharding logic. But the behavior within this method would be consistent with the return value of `maxQueryRange` :)

simonswine (approved):
Thanks for working on this. I think this is mostly LGTM, apart from the mentioned point raised by others.

pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
```diff
@@ -81,6 +90,59 @@ func (dm *DeleteRequestHandler) AddDeleteRequestHandler(w http.ResponseWriter, r
...
+	if err != nil || queryRange < time.Second {
+		return 0, errors.New("invalid query range: valid time units are 's', 'm', 'h'")
+	}
+
+	if queryRange > dm.maxQueryRange && dm.maxQueryRange != 0 {
+		dur, err := time.ParseDuration(dm.maxQueryRange.String())
+		if err != nil {
+			level.Error(util_log.Logger).Log("msg", "error parsing query range", "err", err)
+			return 0, err
+		}
```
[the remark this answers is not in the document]
MasslessParticle: Good catch. Line 128 is supposed to be `model.ParseDuration` instead of `time.ParseDuration` to take advantage of the better formatting for the error message

pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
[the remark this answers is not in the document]
MasslessParticle: :+1: I've made the case where `endTime.Sub(startTime) > maxQueryRange` be a validation error.

pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
```diff
@@ -114,25 +222,25 @@ func (dm *DeleteRequestHandler) CancelDeleteRequestHandler(w http.ResponseWriter
...
-
-	if deleteRequest.Status != StatusReceived {
+	toDelete := filterProcessed(deleteRequests)
+	if len(toDelete) == 0 {
 		http.Error(w, "deletion of request which is in process or already processed is not allowed", http.StatusBadRequest)
 		return
 	}
 
-	if err := dm.deleteRequestsStore.RemoveDeleteRequest(ctx, userID, requestID, deleteRequest.CreatedAt, deleteRequest.StartTime, deleteRequest.EndTime); err != nil {
+	if err := dm.deleteRequestsStore.RemoveDeleteRequests(ctx, toDelete); err != nil {
```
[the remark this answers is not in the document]
MasslessParticle: Because of the requirement to recover from a delete request that cannot complete, I think we need the ability to cancel requests at any time. I've added `% Complete` to requests that are partially processed to provide some more visibility to the user. The result is:

```
[
  {
    "request_id": "626dd400",
    "start_time": 1660653094,
    "end_time": 1660663894,
    "query": "{job=\"loki-ops/ingester\"} |= \"warn\"",
    "status": "66% Complete",
    "created_at": 1660770050.104
  }
]
```

What do you think?

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_store.go
```diff
@@ -216,63 +254,83 @@ func (ds *deleteRequestsStore) queryDeleteRequests(ctx context.Context, deleteQu
...
+	err := ds.indexClient.QueryPages(ctx, deleteRequestQuery, func(query index.Query, batch index.ReadBatchResult) (shouldContinue bool) {
+		itr := batch.Iterator()
+		itr.Next()
+
+		requestWithDetails, parseError = parseDeleteRequestTimestamps(itr.RangeValue(), deleteRequest)
+		if parseError != nil {
+			return false
 		}
 
+		parseError = requestWithDetails.SetQuery(string(itr.Value()))
```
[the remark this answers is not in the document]
MasslessParticle: I've done some renaming in here to help clarify what's going on. PTAL

pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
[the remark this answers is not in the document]
sandeepsukhani: Yeah, showing `%` in status looks like a good idea. The problem with partial cancellation, though, is that there is a chance that it already would have marked chunks from some of the tables for deletion, so it won't be a clean cancellation. Also, since the compactor already picked up the request for processing, you would have to increase the cancellation period config and restart the compactor.

So I wonder if we should add a query param to the cancellation request to explicitly ask the compactor to force cancel the request and document that it won't be a clean cancellation and some data might already have been marked for deletion. However, it would still require the compactor to restart to not continue processing the delete request since we can't interrupt the ongoing compaction, which I think should be okay for now. Other than documentation, it would also save us from doing the config change step.

What do you think?

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_store.go
```diff
@@ -133,10 +133,11 @@ func (ds *deleteRequestsStore) writeDeleteRequest(req DeleteRequest, writeBatch
...
 }
 
-func deleteRequestHash(userID, requestID string, sequenceNumber int64) string {
-	// Any requests made before the addition of sequence numbers won't have one
-	// Ensure backward compatibility by treating the 0th sequence number as the
-	// old format
+// backwardCompatibleDeleteRequestHash generates the hash key for a delete request.
+// Sequence numbers were added after deletion was in production so any requests made
+// before then won't have one. Ensure backward compatibility by treating the 0th
+// sequence number as the old format without any number
```
sandeepsukhani, line 139: Let us also add a note saying the sequence number of the first shard(i.e seq no 0) in the new requests would also be empty.

pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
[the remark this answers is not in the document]
MasslessParticle: :+1: I think that works. The latest commit disallows cancellation of partially-completed deletes unless a `force=true` query parameter is passed.

Issue #6894, opened 2022-08-12
Promtail fails to fetch logs from Docker container from remote Docker daemon docker_sd_config with https URL


longsleep:
**Describe the bug**
Promtail `scrape_configs` contains this section:
```yaml
  - job_name: docker_https
    docker_sd_configs:
      - host: https://test2.lxd:2375
        refresh_interval: 5s
        tls_config:
          ca_file: /srv/tls/local-ca_bundle.crt
          cert_file: /srv/tls/certs/docker-client.crt
          key_file: /srv/tls/certs/docker-client.key
    relabel_configs:
      - source_labels: ['__meta_docker_container_name']
        regex: '/(.*)'
        target_label: 'container'
```

This seems to fetch the containers just fine, but then promtail fails to actually load the log files for the containers.

```
Aug 12 15:39:17 munich promtail[2367033]: level=error ts=2022-08-12T13:39:17.138495308Z caller=target.go:101 target=docker/3a7283bdee7c31eb890bb2cf9599378cdf4ee9cee2eb609f272c0825ad3274e7 msg="could not fetch logs for container" container=3a7283bdee7c31eb890bb2cf9599378cdf4ee9cee2eb609f272c0825ad3274e7 err="Error response from daemon: Client sent an HTTP request to an HTTPS server."
```

**To Reproduce**
1. Configure Promptail to scape using `docker_sd_configs` with a `https` host (remote Docker daemon).
2. I use the Promtail binary build
```
promtail, version 2.6.1 (branch: HEAD, revision: 6bd05c9a4)
  build user:       root@af90ed01061f
  build date:       2022-07-18T08:41:09Z
  go version:       go1.17.6
  platform:         linux/amd64
```
3. After startup, promtail directly starts logging errors like in the example above.

**Expected behavior**
Promtail should load the logs from remote https Docker daemon just fine.

**Environment:**
 - Infrastructure: bare-metal

jeschkies:
Good catch! I think we should set the scheme [here](https://github.com/grafana/loki/blob/main/clients/pkg/promtail/targets/docker/target_group.go#L66). We should also follow Prometheus Docker discovery https://github.com/prometheus/prometheus/blob/main/discovery/moby/docker.go#L145

PR #6909, opened 2022-08-15
Fix reading logs from remote Docker daemon with TLS

longsleep:
The Docker daemon can listen with TLS encryption and thus the
Docker client needs to make HTTPS requests when the configured
Docker host has `https` scheme configured. Furthermore other
HTTP client related settings when the Docker daemon is using
`http` or `https` protocol. This change initializes the Docker
client HTTP support from the scraping configuration to let
TLS connections succeed.

Fixes: https://github.com/grafana/loki/issues/6894

This is more or less a trivial fix, as most of the details are already implemented by Prometheus. Though the details required extending the `targetGroup` struct as did not pass through the required configuration. With some refactoring the change could probably be made more elegant. This is a quick fix which works for me

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


clients/pkg/promtail/targets/docker/target_group.go
```diff
@@ -60,11 +66,36 @@ func (tg *targetGroup) sync(groups []*targetgroup.Group) {
...
+		// options to the Docker client makes those non-HTTP requests fail.
+		if hostURL.Scheme == "http" || hostURL.Scheme == "https" {
+			rt, err := config.NewRoundTripperFromConfig(tg.httpClientConfig, "docker_sd")
+			if err != nil {
+				return err
+			}
+			opts = append(opts,
+				client.WithHTTPClient(&http.Client{
+					Transport: rt,
+					Timeout:   0, // Use 0 timeout, means it it reads until all has been read.
```
salvacorts, line 90: I think this should be configurable within the target group config.

See Prometheus implementation as an example:
https://github.com/prometheus/prometheus/blob/main/discovery/moby/docker.go#L153

clients/pkg/promtail/targets/docker/targetmanager.go
```diff
@@ -73,14 +73,15 @@ func NewTargetManager(
...
-						host:          sdConfig.Host,
+						metrics:          metrics,
+						logger:           logger,
+						positions:        positions,
+						targets:          make(map[string]*Target),
+						entryHandler:     pipeline.Wrap(pushClient),
+						defaultLabels:    model.LabelSet{},
+						relabelConfig:    cfg.RelabelConfigs,
+						host:             sdConfig.Host,
+						httpClientConfig: sdConfig.HTTPClientConfig,
```
salvacorts, line 84: Maybe a bit out of the scope of this PR but, looking at line 72:
```
syncerKey := fmt.Sprintf("%s/%s:%d", cfg.JobName, sdConfig.Host, sdConfig.Port)
```

Looks like we have a `sdConfig.Port` we don't use anywhere else, and we actually expect the user to set any port number in the host itself.

It would be good to use the configured `sdConfig.Port` unless the port is specified in the host itself.

salvacorts (changes requested):
Awesome work! Thanks for putting this together.

I've tested it locally and works great :)

I just left a couple of comments.

longsleep:
Thanks for the reply, will pick this up again asap.

jeschkies:
@longsleep should we take over? 😊

longsleep:
> @longsleep should we take over?

@jeschkies if you can please do. I can't tell when I will have the time to come back to this yet (but it is not forgotten).

jeschkies:
@longsleep could you give me permissions to push to you branch?

longsleep:
> @longsleep could you give me permissions to push to you branch?

@jeschkies sure I invited you to my fork, let me know if that is not enough. Thanks!

clients/pkg/promtail/targets/docker/targetmanager.go
jeschkies: I need to look where and how the port should be used. I think this should be done in another PR.

jeschkies (approved):

jeschkies:
Thanks, @longsleep. Could you also sign the CLA?

longsleep:
> Thanks, @longsleep. Could you also sign the CLA?

Sure, done!

clients/pkg/promtail/targets/docker/targetmanager.go
salvacorts: Doing it on a separate PR sounds good to me as long as we leave a TODO somewhere.

salvacorts (approved):

Issue #6438, opened 2022-06-20
Return a different error when query is out of retention.


jeschkies:
**Is your feature request related to a problem? Please describe.**
When I issue a query that is older than the retention period I receive an error message such as
```
invalid query, through < from (2021-10-27 00:00:00 +0000 UTC < 2022-05-20 15:42:06.265393559 +0000 UTC m=-2394895.003142755) 
```

That message is not too enlightening.

**Describe the solution you'd like**
Ideally the message would tell me that the query is outside of the retention period.

stazdx:
This data is no longer available. Check the documentation if you want to extend the retention storage.

The idea is that 'retention storage' redirects to the url https://grafana.com/docs/loki/latest/operations/storage/retention/

PR #6913, opened 2022-08-15
Improve error message

MichelHollands:

A customer mentioned that this error message was not too clear. It is rewritten now.

Fixes https://github.com/grafana/loki/issues/6438.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -638,7 +638,8 @@ type timeRangeLimits interface {
 func validateQueryTimeRangeLimits(ctx context.Context, userID string, limits timeRangeLimits, from, through time.Time) (time.Time, time.Time, error) {
 	now := nowFunc()
 	// Clamp the time range based on the max query lookback.
-	if maxQueryLookback := limits.MaxQueryLookback(userID); maxQueryLookback > 0 && from.Before(now.Add(-maxQueryLookback)) {
+	var maxQueryLookback time.Duration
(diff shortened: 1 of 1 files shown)


pkg/querier/querier.go
```diff
@@ -652,7 +653,7 @@ func validateQueryTimeRangeLimits(ctx context.Context, userID string, limits tim
 		return time.Time{}, time.Time{}, httpgrpc.Errorf(http.StatusBadRequest, util_validation.ErrQueryTooLong, (through).Sub(from), maxQueryLength)
 	}
 	if through.Before(from) {
-		return time.Time{}, time.Time{}, httpgrpc.Errorf(http.StatusBadRequest, "invalid query, through < from (%s < %s)", through, from)
+		return time.Time{}, time.Time{}, httpgrpc.Errorf(http.StatusBadRequest, "this data is no longer available, it is past now - max_query_lookback (%s)", maxQueryLookback)
```
DylanGuedes, line 656: WDYT of somehow merging both messages? this way users filtering logs based on the older message won't experience this as a breaking change

MichelHollands: I'd rather not because the old message is just wrong/deceptive. It changes the from parameter and prints the new value.

DylanGuedes: ohhh you right. I think this looks good so I'll resolve the thread, but I'll let to others to approve just because, even though the previous message is misleading, it is still a change in an HTTP request return, which I'm not sure how impactful it can be

chaudum (approved):

PR #6917, opened 2022-08-15
reintroduces now deprecated+noop deletion configs for compatibility

owen-d:
This reintroduces a few recently deleted configurations to retain backwards compatibility for old fields, even though their behavior is now a no-op:

 * `boltdb.shipper.compactor.deletion-mode` (compactor config)
 *  `compactor.allow-deletes` (limits config)


ssncferreira (approved):
LGTM 💯

PR #6888, opened 2022-08-12
memberlist: allow consul deployment

irizzant:
Signed-off-by: irizzant <i.rizzante@gmail.com>

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #6887

**Special notes for your reviewer**:

**Checklist**
- [X] Documentation added
- [X] Tests updated
- [X] Is this an important fix or new feature? Add an entry in the `CHANGELOG.md`.
- [X] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #6918, opened 2022-08-15
[jsonnet] Fix conditional for deploying consul when not using memberlist

cstyan:
PR #6888 broke some additional jsonnet we use internally to extend the `consul_deployment`, although the changes in that PR were necessary for OSS users who do want to use consul. 

What we need to do is use {} instead of null since `{} + null` is invalid and will result in an error. Then we can modify our internal extensions to the same conditional we have in this `memberlist.jsonnet` file to only add on to the consul deployment if we are not using memberlist.


vlad-diachenko (approved):
LGTM

PR #6925, opened 2022-08-16
Disable distributed cache for the default loki local config

MasslessParticle:
It breaks if there's not a ring setup


kavirajk (approved):
LGTM. thanks @MasslessParticle.

PR #6934, opened 2022-08-18
log query which causes panic in logql.clone

owen-d:
Found a panic in one of our clusters and wasn't sure the query which caused it. This PR tries to log the offending query during mapping prior to the panic.


DylanGuedes (approved):
lgtm

PR #6936, opened 2022-08-19
Better error message for empty selector

simonswine:


MichelHollands (approved):
LGTM

jeschkies (commented):
I'm not sure this is the same error, though.

> In theory I think we should remove the restriction on querying all log streams

I've just had a call with a user that would like to just search for a string so basically have `"foobar"` be the equivalent of `{} |= "foobar"`. With your remove it seems we are almost there. However, I thought we cannot search all streams...

alex88:
Is there an issue tracking the removal of the restriction on querying all log streams?
Just wanted to add my support on that, we run loki as a log backend and have each device sending logs to its own tenant, to have the user go in the log page and just see all the latest logs I have to manually add a stream default value and query for that.

jeschkies (approved):

PR #6937, opened 2022-08-19
Fix topk and bottomk operations with int <= 0

ssncferreira:

**How the problem was identified**
Range mapper query fails for topk/bottomk expressions with parameter 0. 
When parsing an expression the current flow is:
1) `ParseSampleExpr` from query string: https://github.com/grafana/loki/blob/main/pkg/logql/rangemapper.go#L80
```
	origExpr, err := syntax.ParseSampleExpr(query)
```
this operation succeeds with no error
2) `ParseSampleExpr` from `expr.String()`: https://github.com/grafana/loki/blob/main/pkg/logql/rangemapper.go#L451
```
	e, err := syntax.ParseSampleExpr(expr.String())
```
`expr.String()` does not consider the topk parameter 0 because of this condition: https://github.com/grafana/loki/blob/main/pkg/logql/syntax/ast.go#L927-L931
```
	if e.Params != 0 {
		params = []string{fmt.Sprintf("%d", e.Params), e.Left.String()}
	} else {
		params = []string{e.Left.String()}
	}
```
incorrectly converting the expression to string `topk(sum by(geoip_country_code)(count_over_time({app="foo"} | json | ( geoip_country_code!="" , __error__="" )[15m])))` which then panic at https://github.com/grafana/loki/blob/main/pkg/logql/syntax/ast.go#L869-L871 because param is nil
```
		if params == nil {
			panic(logqlmodel.NewParseError(fmt.Sprintf("parameter required for operation %s", operation), 0, 0))
		}
```

Fixes #<issue number>

The changes in this PR:
1) panic if the param is smaller or equal to 0, returning an error immediately on step 1.
2) fix `expr.String()` to consider the first parameter of topk/bottomk even if 0

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * [5406](https://github.com/grafana/loki/pull/5406) **ctovena**: Revise the configuration parameters that configure the usage report to grafana.com.
 
 ##### Fixes
+* [6937](https://github.com/grafana/loki/pull/6937) **ssncferreira**: Fix topk and bottomk expressions with parameter <= 0.
 * [6358](https://github.com/grafana/loki/pull/6358) **taharah**: Fixes sigv4 authentication for the Ruler's remote write configuration by allowing both a global and per tenant configuration.
 * [6375](https://github.com/grafana/loki/pull/6375) **dannykopping**: Fix bug that prevented users from using the `json` parser after a `line_format` pipeline stage.
 ##### Changes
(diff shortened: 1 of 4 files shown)


MichelHollands (approved):
Very nice, thanks for fixing this.

owen-d (approved):
LGTM, nice explanation & fix

PR #6942, opened 2022-08-19
Modify sharding to not duplicate all entries for all shards

DylanGuedes:
Modify sharding implementation to only use a subset of entries for each shard.


MasslessParticle (approved):
lgtm

PR #6950, opened 2022-08-22
Make TSDB store singleton resettable

chaudum:
This changes allows downstream projects to reset the TSDB store singleton. This must only be called in test cases where a new store instances cannot be explicitly created.

Also add a `nil` check to the reset function for the BoltDB shipper singleton instance.


MichelHollands (approved):
LGTM

slim-bean (approved):
LGTM

PR #6951, opened 2022-08-22
operator: Operational Alerts

### LINKED ISSUE #6373: operator: Additional Prometheus Alerts
This is a request to add additional Prometheus alerts to the operator. The proposal is based on the definitions provided here: https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/loki-observability.md

# LokiRequestLatency

To amend the [proposed query](https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/loki-observability.md#LokiRequestLatency), this metric could be split into a `Read` and `Write` scenario and focused to mainly look on the `querier`, `query-frontend`, `ingester` for `read` and `distributor` for `write`.

However, `Read` itself might be tricky because certain queries may problems.

# LokiTenantRateLimit

Instead of the [proposed query](https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/loki-observability.md#LokiTenantRateLimit), perhaps the number of responses which are returned with a 429 error code should be used. While the `reason` might be not as obvious, we might be able to supplement this by suggesting to query for `loki_discarded_samples_total` to discover this. The value of trying to go above 100 seems slightly arbitrary.

query:

```PromQL
for: 15m
expr: |
    sum(
        rate(
            loki_request_duration_seconds_count{status_code="429"}[1m]
        )
    ) by (namespace, job, route)
    /
    sum(
        rate(
            loki_request_duration_seconds_count[1m]
        )
    ) by (namespace, job, route)
    * 100
    > 10
```

# LokiStorageSlowWrite

Same as [proposed query](https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/loki-observability.md#lokistorageslowwrite).

# LokiStorageSlowRead

Same as [proposed query](https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/loki-observability.md#lokistorageslowread). The value should be updated from `1` to `10` or `15` based on this [feedback](https://github.com/grafana/loki/pull/5345#discussion_r803586300).

# LokiWritePathHighLoad

Since the ingesters use WAL, watching when the storage is being flushed because of backpressuring might be a more generic indicator for detection high load instead of the [proposed query](https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/loki-observability.md#lokiwritepathhighload)

query:

```PromQL
for: 15m
expr: |
    min_over_time(
        loki_ingester_wal_replay_flushing[5m]
    )
    > 0
```

# LokiReadPathHighLoad

Same as [proposed query](https://github.com/grafana/loki/pull/5345/commits/d720bde5ecdfa9b5bb1192dd3ef03e61872e5e7f#r810598836).

query:

```PromQL
for: 15m
expr: |
    histogram_quantile(0.90,
       sum(
          rate(
            loki_logql_querystats_latency_seconds_bucket[5m]
          )
       ) by (le, namespace, job)
    )
    > 10
```

Red-GV:
This PR adds several alerts which provide information about the operational status of the Lokistack.

Fixes #6373

**Checklist**
- [x] Documentation added
- [x] Tests updated


operator/docs/lokistack/sop.md
```diff
@@ -136,3 +136,161 @@ A service(s) has crashed.
...
+- Console access to the cluster
+- Edit access to the deployed operator and Loki namespace:
+  - OpenShift
+    - `openshift-logging` (LokiStack)
+    - `openshift-operators-redhat` (Loki Operator)
+
+### Steps
+
+- Examine the metrics for the reason and tenant that is being limited: `loki_discarded_samples_total{namespace="<namespace>"}`
+- Change the ingestion limits for the affected tenant or decrease the rate of logs entering the system
```
periklis, line 192: Links to the CRD limits spec would be useful and maybe a small table between the reasons and the limits config knobs. For example:
- For reason `rate_limited` you can play with `IngestionRate` and/or `IngestionBurstSize`
- For reason `stream_limit` you can play with `MaxGlobalStreamsPerTenant`
- For query limits you can play with `MaxEntriesLimitPerQuery` or `MaxChunksPerQuery` or `MaxQuerySeries`

operator/docs/lokistack/sop.md
```diff
@@ -136,3 +136,161 @@ A service(s) has crashed.
...
+- Edit access to the deployed operator and Loki namespace:
+  - OpenShift
+    - `openshift-logging` (LokiStack)
+    - `openshift-operators-redhat` (Loki Operator)
+
+### Steps
+
+- Adjust the ingestion limits for the affected tenant or increase the number of ingesters
+
+## Loki Write Path High Load
```
periklis, line 272:
```suggestion
## Loki Read Path High Load
```

periklis (commented):
A first pass:

operator/docs/lokistack/sop.md
Red-GV: I've added a table for the `ingester`. However, for the `querier`, there doesn't seem to be any `reason` for those in the metrics. I'd imagine this is probably because it's meant to be expressed via the response maybe?

operator/internal/manifests/internal/alerts/prometheus-alerts.yaml
```diff
@@ -82,7 +70,114 @@ groups:
...
+    annotations:
+      message: |-
+        The write path is experiencing high load.
+      summary: "The write path is experiencing high load, causing backpressure storage flushing."
+      runbook_url: "[[ .RunbookURL ]]#Loki-Write-Path-High-Load"
+    expr: |
+      sum(
+        max_over_time(
+          loki_ingester_wal_replay_flushing[1m]
+        )
```
xperimental, line 159: I don't think we need the `max_over_time` here. Just using the raw number should be fine. The "over time" part should be handled by the `for: 15m` clause of the alert.

operator/internal/manifests/internal/alerts/prometheus-alerts.yaml
```diff
@@ -82,7 +70,114 @@ groups:
...
+    expr: |
+      histogram_quantile(0.99,
+        sum(
+          rate(
+            loki_logql_querystats_latency_seconds_bucket[5m]
+          )
+        ) by (le, job, namespace)
+      )
+      * 100
+      > 30
```
xperimental, line 180: Is this maybe a copy-paste error from the percent-based alerts? The unit of the `histogram_quantile` result should be "seconds".

operator/internal/manifests/internal/alerts/prometheus-alerts.yaml
```diff
@@ -82,7 +70,114 @@ groups:
         increase(
           loki_panic_total[10m]
         )
-      ) by (namespace, job)
+      ) by (job, namespace)
```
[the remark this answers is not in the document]
Red-GV: The order change was mainly consistency with the order to the new alerts.

operator/internal/manifests/internal/alerts/prometheus-alerts.yaml
Red-GV: From the tests, the `LokiRequestLatency` alert, and looking at the values, I think the value does need to be multiplied by 100 to actually produce results.

operator/docs/lokistack/sop.md
```diff
@@ -136,3 +136,170 @@ A service(s) has crashed.
...
+### Steps
+
+- Check the logs of all the services
+- Check to ensure that the Loki components can reach the storage
+
+## Loki Tenant Rate Limit
+
+### Impact
+
+A tenant is being rate limited, resulting in potential loss of data.
```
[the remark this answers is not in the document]
Red-GV: It _might_ be fine as is. The reason I say that is because I'm thinking of what happens if the client is constantly getting battered with 429s because the configuration needs to change. What I'm "worried" about is someone reading that that the data loss is not immediate and delay fixing the issue.

periklis (approved):

PR #6952, opened 2022-08-22
Loki: Implement stream sharding

DylanGuedes:
Implement a new feature named `stream sharding`. Distributors with stream sharding enabled will reduce big streams before pushing them to the ingesters by adding a new sharding label. By having a new label, a bigger stream is hence reduced.
This shouldn't impact Loki in any form since it is disabled by default.

This initial implementation contains:
- A sharder interface to be used internally regardless of the implementation
- An initial sharder implementation that supports sharding streams
- A mechanism to recognize per-stream limits and trigger sharding based on it
- Related configurations that dictate if sharding should be enabled or not. We're not documenting these configurations yet as we don't want users to use by mistake since this is far from ready

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -51,6 +51,14 @@ var (
 	rfStats           = usagestats.NewInt("distributor_replication_factor")
 )
 
+type ShardStreamMode string
+
+const (
+	NeverShardMode       ShardStreamMode = "never"
+	AlwaysShardMode                      = "always"
+	DynamicallyShardMode                 = "dynamic"
+)
+
 // TODO: document shard streams configurations/modes.
 type ShardStreamsConfig struct {
 	// TODO: create enum for sharding modes.
@@ -59,7 +67,7 @@ type ShardStreamsConfig struct {
 	//   'never': never shards any stream. Default mode.
 	//   'dynamic': shard only streams that did hit the per-stream rate limit. First requests are expected to be slower
 	// but once the correct amount of sharding is calculated, no latency increase is expected.
-	Mode string `yaml:"mode"`
+	Mode ShardStreamMode `yaml:"mode"`
 
 	// Debug defines if debugging log lines should be enabled or not.
 	//
@@ -83,7 +91,7 @@ type Config struct {
 func (cfg *Config) RegisterFlags(fs *flag.FlagSet) {
 	cfg.DistributorRing.RegisterFlags(fs)
 
-	cfg.ShardStreams.Mode = "never"
+	cfg.ShardStreams.Mode = NeverShardMode
 
 	// TODO: add sharding flags.
 }
@@ -329,7 +337,7 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
 		}
 		stream.Entries = stream.Entries[:n]
 
-		if d.cfg.ShardStreams.Mode != "never" {
+		if d.cfg.ShardStreams.Mode != NeverShardMode {
 			derivedKeys, derivedStreams := shardStream(stream, d.cfg, d.streamSharder, userID)
 			keys = append(keys, derivedKeys...)
 			streams = append(streams, derivedStreams...)
@@ -535,7 +543,7 @@ func (d *Distributor) sendSamplesErr(ctx context.Context, ingester ring.Instance
 	d.ingesterAppendFailures.WithLabelValues(ingester.Addr).Inc()
 
 	// TODO: look for a better way to recognize a per-stream rate limit.
-	if d.cfg.ShardStreams.Mode == "never" {
+	if d.cfg.ShardStreams.Mode == NeverShardMode {
 		return err
 	}
 
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -55,6 +55,16 @@ func (cfg *Config) RegisterFlags(fs *flag.FlagSet) {
 	cfg.DistributorRing.RegisterFlags(fs)
 }
 
+type ShardIter interface {
+	NumShards() int
+	NextShardID() int
+}
+
+type StreamSharder interface {
+	ShardsFor(stream string) ShardIter
+	IncreaseShardsFor(stream string)
+}
+
 // Distributor coordinates replicates and distribution of log streams.
 type Distributor struct {
 	services.Service
@@ -66,6 +76,7 @@ type Distributor struct {
 	ingestersRing    ring.ReadRing
 	validator        *Validator
 	pool             *ring_client.Pool
+	streamSharder    StreamSharder
 
 	// The global rate limiter requires a distributors ring to count
 	// the number of healthy instances.
@@ -75,11 +86,9 @@ type Distributor struct {
 
 	subservices        *services.Manager
 	subservicesWatcher *services.FailureWatcher
-
 	// Per-user rate limiter.
 	ingestionRateLimiter *limiter.RateLimiter
 	labelCache           *lru.Cache
-
 	// metrics
 	ingesterAppends        *prometheus.CounterVec
 	ingesterAppendFailures *prometheus.CounterVec
@@ -87,7 +96,15 @@ type Distributor struct {
 }
 
 // New a distributor creates.
-func New(cfg Config, clientCfg client.Config, configs *runtime.TenantConfigs, ingestersRing ring.ReadRing, overrides *validation.Overrides, registerer prometheus.Registerer) (*Distributor, error) {
+func New(
+	cfg Config,
+	clientCfg client.Config,
+	configs *runtime.TenantConfigs,
+	ingestersRing ring.ReadRing,
+	overrides *validation.Overrides,
+	sharder StreamSharder,
+	registerer prometheus.Registerer,
+) (*Distributor, error) {
 	factory := cfg.factory
 	if factory == nil {
 		factory = func(addr string) (ring_client.PoolClient, error) {
@@ -135,6 +152,7 @@ func New(cfg Config, clientCfg client.Config, configs *runtime.TenantConfigs, in
 		ingestersRing:          ingestersRing,
 		distributorsLifecycler: distributorsLifecycler,
 		validator:              validator,
+		streamSharder:          sharder,
 		pool:                   clientpool.NewPool(clientCfg.PoolConfig, ingestersRing, factory, util_log.Logger),
 		ingestionRateLimiter:   limiter.NewRateLimiter(ingestionRateStrategy, 10*time.Second),
 		labelCache:             labelCache,
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -439,7 +440,8 @@ func shardStream(stream logproto.Stream, cfg Config, streamSharder StreamSharder
 			continue
 		}
 
-		lbs = append(lbs, labels.Label{Name: ShardLbName, Value: fmt.Sprintf("%d", i)})
+		idx := strconv.Itoa(i)
+		lbs = append(lbs, labels.Label{Name: ShardLbName, Value: idx})
 		streamCopy.Labels = lbs.String()
 		streamCopy.Hash = lbs.Hash()
 
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -433,6 +433,13 @@ func shardStream(stream logproto.Stream, cfg Config, streamSharder StreamSharder
 
 	entriesPerWindow := float64(len(stream.Entries)) / float64(shards) // divide and keep decimal value.
 	for i := 0; i < shards; i++ {
+		fIdx := float64(i)
+		lowerBound := int(fIdx * entriesPerWindow)
+		upperBound := min(int(entriesPerWindow*(1+fIdx)), len(stream.Entries))
+		if lowerBound > upperBound {
+			continue
+		}
+
 		streamCopy := logproto.Stream{}
 		lbs, err := syntax.ParseLabels(stream.Labels)
 		if err != nil {
@@ -444,10 +451,6 @@ func shardStream(stream logproto.Stream, cfg Config, streamSharder StreamSharder
 		lbs = append(lbs, labels.Label{Name: ShardLbName, Value: idx})
 		streamCopy.Labels = lbs.String()
 		streamCopy.Hash = lbs.Hash()
-
-		fIdx := float64(i)
-		lowerBound := int(fIdx * entriesPerWindow)
-		upperBound := min(int(entriesPerWindow*(1+fIdx)), len(stream.Entries))
 
 		streamCopy.Entries = stream.Entries[lowerBound:upperBound]
 		if cfg.ShardStreams.Debug {
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -567,7 +567,7 @@ func shardLimitedStream(perStreamErrorMsg string, sharder StreamSharder) {
 		return
 	}
 
-	labels := fmt.Sprintf("{%s}", matches[0][1]) // add curly braces
+	labels := "{" + matches[0][1] + "}"
 
 	lbs, err := syntax.ParseLabels(labels)
 	if err != nil {
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -3,8 +3,12 @@ package distributor
 import (
 	"context"
 	"flag"
+	"fmt"
 	"net/http"
 	"time"
+
+	"github.com/go-kit/log/level"
+	"github.com/prometheus/prometheus/model/labels"
 
 	"github.com/grafana/dskit/limiter"
 	"github.com/grafana/dskit/ring"
@@ -33,6 +37,10 @@ import (
 )
 
 const (
+	// ShardLbName is the internal label to be used by Loki when dividing a stream into smaller pieces.
+	// Possible values are only increasing integers starting from 0.
+	ShardLbName = "__stream_shard__"
+
 	ringKey = "distributor"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/distributor/distributor.go
```diff
@@ -154,6 +200,11 @@ func New(cfg Config, clientCfg client.Config, configs *runtime.TenantConfigs, in
 			Name:      "distributor_replication_factor",
 			Help:      "The configured replication factor.",
 		}),
+		streamSharding: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
+			Namespace: "loki",
+			Name:      "distributor_stream_sharding",
```
dannykopping, line 213: For counters, I'd recommend following the [convention](https://prometheus.io/docs/practices/naming/#metric-names) and adding a `_total` suffix.

```suggestion
			Name:      "distributor_streams_sharded_total",
```

pkg/distributor/distributor.go
```diff
@@ -276,8 +329,14 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
...
-		keys = append(keys, util.TokenFor(userID, stream.Labels))
-		streams = append(streams, streamTracker{stream: stream})
+		if d.cfg.ShardStreams.Mode != "never" {
+			derivedKeys, derivedStreams := shardStream(stream, d.cfg, d.streamSharder, userID)
+			keys = append(keys, derivedKeys...)
+			streams = append(streams, derivedStreams...)
+		} else {
+			keys = append(keys, util.TokenFor(userID, stream.Labels))
+			streams = append(streams, streamTracker{stream: stream})
+		}
```
dannykopping: Very minor non-blocking nit which speaks to style:

Try avoid double-negatives, as it helps when reading the code (especially when it's 2AM and you've been paged and you don't know the code).

```suggestion
		if d.cfg.ShardStreams.Mode == "never" {
			keys = append(keys, util.TokenFor(userID, stream.Labels))
			streams = append(streams, streamTracker{stream: stream})
		} else {
			derivedKeys, derivedStreams := shardStream(stream, d.cfg, d.streamSharder, userID)
			keys = append(keys, derivedKeys...)
			streams = append(streams, derivedStreams...)
		}
```

pkg/distributor/distributor.go
```diff
@@ -339,6 +398,59 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
 	}
 }
 
+func min(x1, x2 int) int {
```
dannykopping, line 409: Out of interest, do we have this because `math.Min` only accepts `float64`?

pkg/distributor/distributor.go
```diff
@@ -415,12 +527,46 @@ func (d *Distributor) sendSamplesErr(ctx context.Context, ingester ring.Instance
...
-	if err != nil {
-		d.ingesterAppendFailures.WithLabelValues(ingester.Addr).Inc()
+
+	if err == nil {
+		return nil
+	}
+
+	d.ingesterAppendFailures.WithLabelValues(ingester.Addr).Inc()
+
+	// TODO: look for a better way to recognize a per-stream rate limit.
```
dannykopping, line 545: I think this comment belongs more to the branch below, right? This confused me momentarily

pkg/distributor/distributor.go
```diff
@@ -339,6 +398,59 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
...
+	entriesPerWindow := float64(len(stream.Entries)) / float64(shards) // divide and keep decimal value.
+	for i := 0; i < shards; i++ {
+		streamCopy := logproto.Stream{}
+		lbs, err := syntax.ParseLabels(stream.Labels)
+		if err != nil {
+			level.Error(logger).Log("msg", "couldn't extract labels from stream", "stream", streamCopy.Labels)
+			continue
+		}
+
+		lbs = append(lbs, labels.Label{Name: ShardLbName, Value: fmt.Sprintf("%d", i)})
```
dannykopping, line 442: The `fmt.Sprintf` here may be quite expensive; I would suggest benchmarking this function as it's in the critical path.

pkg/distributor/distributor.go
```diff
@@ -339,6 +398,59 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
...
+		streamCopy.Labels = lbs.String()
+		streamCopy.Hash = lbs.Hash()
+
+		fIdx := float64(i)
+		lowerBound := int(fIdx * entriesPerWindow)
+		upperBound := min(int(entriesPerWindow*(1+fIdx)), len(stream.Entries))
+
+		streamCopy.Entries = stream.Entries[lowerBound:upperBound]
```
dannykopping, line 450: I think we should validate here that:

1. `lowerBound >= 0` (will panic otherwise)
2. `lowerBound < upperBound` (otherwise it'll panic if greater, or return no entries if equal)

pkg/distributor/distributor.go
```diff
@@ -415,12 +535,46 @@ func (d *Distributor) sendSamplesErr(ctx context.Context, ingester ring.Instance
...
 
+// shardLimitedStream shards limited stream based on the given per-stream rate limited message.
+func shardLimitedStream(perStreamErrorMsg string, sharder StreamSharder) {
+	re := regexp.MustCompile("\\{(.*?)\\}")
+	matches := re.FindAllStringSubmatch(perStreamErrorMsg, 1)
+	if len(matches) < 1 {
+		return
+	}
+
+	labels := fmt.Sprintf("{%s}", matches[0][1]) // add curly braces
```
dannykopping, line 565: If we can avoid `fmt.Sprintf` I think we can save a bit of CPU (`Sprintf` uses reflection) and possibly an alloc; the equivalent will work:
```suggestion
	labels := "{" + matches[0][1] + "}"
```

pkg/distributor/streamsharder.go
```diff
@@ -0,0 +1,45 @@
...
+	return &streamSharder{
+		streams: make(map[string]int),
+	}
+}
+
+func (s *streamSharder) ShardsFor(stream logproto.Stream) (int, bool) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	shards := s.streams[stream.Labels]
```
dannykopping, line 24: Since you're just reading here, a `RWMutex` would work and maybe reduce some contention

dannykopping (commented):
Looks great! Left a few perf/safety-related comments, and the rest are stylistic and are non-blocking

pkg/distributor/distributor.go
DylanGuedes: yeah. I believe in the near future we might migrate to a generic implementation though (because of... generics)

pkg/distributor/distributor.go
DylanGuedes: I pushed a benchmark and a change to not use `fmt.Sprintf` and you are right, not using `fmt.Sprintf` did help. There's other optimizations that we can do too, such as not parsing all labels just to append the shard label, but I prefer doing them in a follow-up PR to make reviewing easier.

pkg/distributor/distributor.go
DylanGuedes: Agreed on the first, pushed a change that guarantees we'll only calculate all of this if shards  > 0, which guards against it.

For second, I wrote a guard-clause for it and added a few tests covering it (number of shards > len(entries) is the only scenario where lowerbound>upperbound would be possible).

pkg/distributor/distributor.go
DylanGuedes: Agreed, ty for catching it.

pkg/distributor/distributor.go
DylanGuedes: Oops, agreed. Moved it to the right place, thanks for catching it.

pkg/distributor/distributor.go
```diff
@@ -432,21 +432,25 @@ func shardStream(stream logproto.Stream, cfg Config, streamSharder StreamSharder
 
 	entriesPerWindow := float64(len(stream.Entries)) / float64(shards) // divide and keep decimal value.
 	for i := 0; i < shards; i++ {
+		fIdx := float64(i)
+		lowerBound := int(fIdx * entriesPerWindow)
+		upperBound := min(int(entriesPerWindow*(1+fIdx)), len(stream.Entries))
+		if lowerBound > upperBound {
+			continue
```
dannykopping, line 439: It _may_ be worth logging a debug entry here, or incrementing a metric - otherwise your shard splitting may be silently failing with no real indication why

pkg/distributor/distributor.go
dannykopping: Works for me :+1: thanks!

pkg/distributor/distributor.go
```diff
@@ -415,12 +539,46 @@ func (d *Distributor) sendSamplesErr(ctx context.Context, ingester ring.Instance
...
+	if errMsg := err.Error(); strings.Contains(errMsg, "Per stream rate limit exceeded (limit") {
+		d.streamSharding.Inc()
+		shardLimitedStream(errMsg, d.streamSharder)
 	}
 	return err
 }
 
+// shardLimitedStream shards limited stream based on the given per-stream rate limited message.
+func shardLimitedStream(perStreamErrorMsg string, sharder StreamSharder) {
+	re := regexp.MustCompile("\\{(.*?)\\}")
```
MichelHollands, line 563: Can you move this outside of the loop? It gets compile here every time.

pkg/distributor/distributor_test.go
```diff
@@ -373,6 +375,313 @@ func Test_TruncateLogLines(t *testing.T) {
...
+
+	cfg := Config{ShardStreams: ShardStreamsConfig{
+		Debug: false,
+	}}
+	b.Run("high number of entries, low number of shards", func(b *testing.B) {
+		streamSharder := NewStreamSharderMock(1)
+		stream.Entries = allEntries
+		for n := 0; n < b.N; n++ {
+			shardStream(stream, cfg, streamSharder, "fake")
+		}
```
MichelHollands, line 657:
```suggestion
		streamSharder := NewStreamSharderMock(1)
		stream.Entries = allEntries
		b.ResetTimer()
		for n := 0; n < b.N; n++ {
			shardStream(stream, cfg, streamSharder, "fake")
		}
```
If the `NewStreamSharderMock()` takes time you want to exclude it from the benchmark.

pkg/distributor/distributor.go
```diff
@@ -33,6 +39,10 @@ import (
 )
 
 const (
+	// ShardLbName is the internal label to be used by Loki when dividing a stream into smaller pieces.
+	// Possible values are only increasing integers starting from 0.
+	ShardLbName = "__stream_shard__"
```
jeschkies, line 44: Aren't we also using `__shard` for splitting queries? I wonder if this could be confusing.

pkg/distributor/distributor.go
```diff
@@ -415,12 +539,46 @@ func (d *Distributor) sendSamplesErr(ctx context.Context, ingester ring.Instance
...
+	if d.cfg.ShardStreams.Mode == NeverShardMode {
+		return err
+	}
+
+	// TODO: look for a better way to recognize a per-stream rate limit.
+	if errMsg := err.Error(); strings.Contains(errMsg, "Per stream rate limit exceeded (limit") {
+		d.streamSharding.Inc()
+		shardLimitedStream(errMsg, d.streamSharder)
 	}
 	return err
```
jeschkies, line 558: Should there be a retry?

pkg/distributor/streamsharder.go
```diff
@@ -0,0 +1,45 @@
...
+
+	return 0, false
+}
+
+func (s *streamSharder) IncreaseShardsFor(stream logproto.Stream) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	shards := s.streams[stream.Labels]
+	s.streams[stream.Labels] = max(shards*2, 2)
```
jeschkies, line 37: I think this needs some explanation.

pkg/distributor/distributor.go
DylanGuedes: it has the `__stream` prefix so IMO it is distinguishing well. Do you have a suggestion? We could try not using the `shard` word but I think it would be more confusing than helpful, as we're clearly sharding things.

pkg/distributor/distributor.go
```diff
@@ -41,18 +51,55 @@ var (
...
 
+type ShardStreamMode string
+
+const (
+	NeverShardMode       ShardStreamMode = "never"
+	AlwaysShardMode                      = "always"
+	DynamicallyShardMode                 = "dynamic"
+)
+
+// TODO: document shard streams configurations/modes.
```
jeschkies, line 62: Just making sure that these are addressed before the merge :slightly_smiling_face:

jeschkies (changes requested):
Please address the TODOs before landing :slightly_smiling_face:

pkg/distributor/distributor.go
DylanGuedes: In the PR description there's an explanation of why we're not documenting this yet.

pkg/distributor/streamsharder.go
DylanGuedes: Right, we can add a docstring for that. But explaining here anyway: this is to make sure the first sharding will result in at least 2 shards. Since it is the first sharding, shards would be 0, so shards * 2 = 0 which wouldn't work. This is basically to avoid doing:
```
if (shards) == 0 {
  shards = 1
}
s.streams[stream.Labels] = shards*2
```

pkg/distributor/distributor.go
DylanGuedes: agreed, good catch. I was taking a look at it and apparently it won't be too easy; I'll dig into it a little bit more today. Depending on the complexity I'd prefer doing it in a follow-up PR to make reviewing this one easier, but let's see!

pkg/distributor/distributor.go
DylanGuedes: Done. I added it as an attribute in the Distributor struct that is reused here; WDYT?

pkg/distributor/distributor_test.go
DylanGuedes: the allocation is outside of the for loop so it isn't really affected. WDYT?

pkg/distributor/streamsharder.go
DylanGuedes: I pushed a commit handling it, lmk what you think.

pkg/distributor/distributor.go
DylanGuedes: for this first iteration I think it makes sense to have this kind of debugging, so I added a log line here reporting the scenario. WDYT?

dannykopping: :+1: perfect, thanks

dannykopping (approved):
LGTM, nice work! I'd be interested to see the benchmarks

DylanGuedes:
> I'd be interested to see the benchmarks

@dannykopping I'm pasting the results below. High number of shards = 64 which is insane, high number of entries = 25k which is also quite high. It is slower with more shards because we have to allocate N streams for every call where N is the number of shards AND the way we're appending the shard label is bad but we already have it mapped to address in the future.
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -tags integration,requires_docker -bench ^BenchmarkShardStream$ github.com/grafana/loki/pkg/distributor

goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/distributor
BenchmarkShardStream/high_number_of_entries,_low_number_of_shards-12         	11121926	       115.2 ns/op	      84 B/op	       2 allocs/op
BenchmarkShardStream/low_number_of_entries,_low_number_of_shards-12          	 8990886	       111.6 ns/op	      84 B/op	       2 allocs/op
BenchmarkShardStream/high_number_of_entries,_high_number_of_shards-12        	   10000	    112689 ns/op	   43058 B/op	     898 allocs/op
BenchmarkShardStream/low_number_of_entries,_high_number_of_shards-12         	   10000	    118374 ns/op	   43078 B/op	     898 allocs/op
PASS
ok  	github.com/grafana/loki/pkg/distributor	4.900s
```

dannykopping:
> > I'd be interested to see the benchmarks
> 
> @dannykopping I'm pasting the results below. High number of shards = 64 which is insane, high number of entries = 25k which is also quite high. It is slower with more shards because we have to allocate N streams for every call where N is the number of shards AND the way we're appending the shard label is bad but we already have it on our list to be optimzed in the future.
> 
> ```
> Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -tags integration,requires_docker -bench ^BenchmarkShardStream$ github.com/grafana/loki/pkg/distributor
> 
> goos: linux
> goarch: amd64
> pkg: github.com/grafana/loki/pkg/distributor
> cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
> BenchmarkShardStream/high_number_of_entries,_low_number_of_shards-12         	11121926	       115.2 ns/op	      84 B/op	       2 allocs/op
> BenchmarkShardStream/low_number_of_entries,_low_number_of_shards-12          	 8990886	       111.6 ns/op	      84 B/op	       2 allocs/op
> BenchmarkShardStream/high_number_of_entries,_high_number_of_shards-12        	   10000	    112689 ns/op	   43058 B/op	     898 allocs/op
> BenchmarkShardStream/low_number_of_entries,_high_number_of_shards-12         	   10000	    118374 ns/op	   43078 B/op	     898 allocs/op
> PASS
> ok  	github.com/grafana/loki/pkg/distributor	4.900s
> ```

OK cool, thanks a lot. Yeah 900 allocs/op will really hurt performance in the critical write path, so before we put this in production with high numbers of shards we should address this.

MasslessParticle:
Benchmarks for the latest commit:
```
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/distributor
BenchmarkShardStream/high_number_of_entries,_low_number_of_shards-8               495884              2407 ns/op            1089 B/op         19 allocs/op
BenchmarkShardStream/low_number_of_entries,_low_number_of_shards-8                497318              2666 ns/op            1089 B/op         19 allocs/op
BenchmarkShardStream/high_number_of_entries,_high_number_of_shards-8               56112             21486 ns/op           14516 B/op        143 allocs/op
BenchmarkShardStream/low_number_of_entries,_high_number_of_shards-8                48384             22079 ns/op           14522 B/op        143 allocs/op
```

pkg/distributor/distributor.go
MasslessParticle: We removed the trigger that to split the query from this PR so it can be its own

pkg/distributor/distributor.go
MasslessParticle: :+1:  done today

MasslessParticle:
The latest commit also changes the benchmarks to make the `low_number_of_shards` = 2. When it was 1 the code short-circuited so we weren't really testing anything.

pkg/distributor/streamsharder.go
jeschkies: Thanks. That clears it up.

pkg/distributor/distributor.go
```diff
@@ -41,18 +51,34 @@ var (
...
 func (cfg *Config) RegisterFlags(fs *flag.FlagSet) {
 	cfg.DistributorRing.RegisterFlags(fs)
+	fs.BoolVar(&cfg.ShardStreams.Enabled, "distributor.stream-sharding.enabled", false, "Automatically shard streams to keep them under the per-stream rate limit")
+	fs.BoolVar(&cfg.ShardStreams.DebugLoggingEnabled, "distributor.stream-sharding.debug-logging-enabled", false, "Enable debug logging when sharding streams")
+}
+
+// StreamSharder manages the state necessary to shard streams.
+type StreamSharder interface {
+	ShardsFor(stream logproto.Stream) (int, bool)
+	IncreaseShardsFor(stream logproto.Stream)
```
jeschkies, line 81: Is this called in production? I cannot find it.

pkg/distributor/distributor.go
```diff
@@ -339,6 +379,98 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
...
+func (d *Distributor) createShard(stream logproto.Stream, lbls labels.Labels, streamPattern string, totalShards, shardNumber int) (logproto.Stream, bool) {
+	lowerBound, upperBound, ok := d.boundsFor(stream, totalShards, shardNumber)
+	if !ok {
+		return logproto.Stream{}, false
+	}
+
+	shardLabel := strconv.Itoa(shardNumber)
+	lbls[len(lbls)-1] = labels.Label{Name: ShardLbName, Value: shardLabel}
+	return logproto.Stream{
+		Labels:  strings.Replace(streamPattern, ShardLbPlaceholder, shardLabel, 1),
```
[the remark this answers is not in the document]
MasslessParticle: It is! `labels.String()` was responsible for a surprisingly large proportion of our allocs when profiled

pkg/distributor/distributor.go
MasslessParticle: Not yet. `IncreaseShards` will be called as part of a follow-on PR that triggers stream sharding based on errors from the ingester

jeschkies (approved):

PR #6954, opened 2022-08-23
Use a line-buffered logger to deamplify write syscalls

dannykopping:
We initialise a global logger in `pkg/util/log/log.go` and use it extensively throughout the Loki codebase. Every time we write a log message, a `write` syscall is invoked. Syscalls are problematic because they transition the process from userspace to kernelspace, which means:

- a context-switch is incurred, which is inherently expensive ([1-2 microseconds](https://eli.thegreenplace.net/2018/measuring-context-switching-and-memory-overheads-for-linux-threads/))
- the goroutine executing the code is **blocked**
- the underlying OS thread (_M_ in the go scheduler model) is **also blocked**
- the goroutine has to be rescheduled once the syscall exits
- the go scheduler may need to spawn additional OS threads if all are blocked in syscalls - which can also be expensive

This change introduces the use of a line-buffered logger. It has a buffer of [256](https://gist.github.com/dannykopping/0704db32c0b08751d1d2494efaa734c2) entries, and once that buffer is filled it will flush to disk. However, a situation will arise in which that buffer remains somewhat empty for a period of time, so there is a periodic flush mechanism, configured to flush every 100ms. There is also a preallocated bytes slice of 10MB which is reused, to avoid excessive slice resizing & garbage collection.

This does mean that we could lose up to 256 log messages in case of an ungraceful termination of the process, but this would need to be precisely timed within the 100ms flushes - in other words, the likelihood is low, and generally we shouldn't `kill -9` any Loki process. 

N/A

This PR uses my private fork of `go-kit/log`; once we've validated this to work in our environment I will attempt to upstream the change.

[Line-buffering](https://www.gnu.org/software/libc/manual/html_node/Buffering-Concepts.html) is common, and in Linux the `stdout` stream is [line-buffered](https://man7.org/linux/man-pages/man3/stdin.3.html#NOTES) when attached to a terminal.

We also currently make use of the `log.NewSyncWriter` writer, which is totally unnecessary since `write` syscalls are threadsafe by nature, and we also don't need to worry about out-of-order logs anymore. Removing this should also eliminate some contention, and there's a config option to control this separately while we test this out.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/util/log/log.go
+++ b/pkg/util/log/log.go
@@ -47,8 +47,11 @@ type prometheusLogger struct {
 // Prometheus counters for various log levels.
 func newPrometheusLogger(l logging.Level, format logging.Format, reg prometheus.Registerer) log.Logger {
 
-	// buffer up to 64 log lines in memory before flushing to a write(2) syscall
-	var logBufferSize uint32 = 64
+	var (
+		logEntries    uint32 = 256                    // buffer up to 256 log lines in memory before flushing to a write(2) syscall
+		logBufferSize uint32 = 10e6                   // 10MB
+		flushTimeout         = 100 * time.Millisecond // flush the buffer after 100ms regardless of how full it is, to prevent losing many logs in case of ungraceful termination
+	)
 
 	logMessages := promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 		Namespace: "loki",
@@ -59,19 +62,22 @@ func newPrometheusLogger(l logging.Level, format logging.Format, reg prometheus.
 		Namespace: "loki",
 		Name:      "log_flushes",
 		Help:      "Histogram of log flushes using the line-buffered logger.",
-		Buckets:   prometheus.ExponentialBuckets(1, 2, int(math.Log2(float64(logBufferSize)))+1),
+		Buckets:   prometheus.ExponentialBuckets(1, 2, int(math.Log2(float64(logEntries)))+1),
 	})
 
 	// TODO: it's technically possible here to lose logs between the 100ms flush and the process being killed
 	// 	=> call buf.Flush() in a signal handler if this is a concern, but this is unlikely to be a problem
-	buf := log.NewLineBufferedLogger(os.Stderr, logBufferSize, 100*time.Millisecond)
-	buf.OnFlush(func(bufLen uint32) {
-		logFlushes.Observe(float64(bufLen))
-	})
+	buf := log.NewLineBufferedLogger(os.Stderr, logEntries,
+		log.WithFlushPeriod(flushTimeout),
+		log.WithPrellocatedBuffer(logBufferSize),
+		log.WithFlushCallback(func(entries uint32) {
+			logFlushes.Observe(float64(entries))
+		}),
+	)
 
-	logger := log.NewLogfmtLogger(log.NewSyncWriter(buf))
+	logger := log.NewLogfmtLogger(buf)
 	if format.String() == "json" {
-		logger = log.NewJSONLogger(log.NewSyncWriter(buf))
+		logger = log.NewJSONLogger(buf)
 	}
 	logger = level.NewFilter(logger, levelFilter(l.String()))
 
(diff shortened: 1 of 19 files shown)


pkg/util/log/log.go
```diff
@@ -38,24 +41,64 @@ func InitLogger(cfg *server.Config, reg prometheus.Registerer) {
...
 
 // newPrometheusLogger creates a new instance of PrometheusLogger which exposes
 // Prometheus counters for various log levels.
-func newPrometheusLogger(l logging.Level, format logging.Format, reg prometheus.Registerer) log.Logger {
-	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
+func newPrometheusLogger(l logging.Level, format logging.Format, reg prometheus.Registerer, buffered bool, sync bool) log.Logger {
+
+	// buffered logger settings
+	var (
+		logEntries    uint32 = 256                    // buffer up to 256 log lines in memory before flushing to a write(2) syscall
```
salvacorts, line 56: Would be nice if these could be configured.

dannykopping:
> > we shouldn't kill -9 any Loki process.
> 
> I'm more concerned about OOMed processes. If a process runs out of memory, it'll be killed and we might lose important logs to assess what made the memory of the process increase.

That's true, but generally logs are not useful in diagnosing problems of this nature - continuous profiling would be the solution here.

pkg/util/log/log.go
dannykopping: Maybe in a subsequent PR we can make it configurable, but I don't think it's worth it personally.
It's probably already tuned ~optimally, so I don't think a user would need to configure it.

salvacorts (approved):

pkg/util/log/log.go
```diff
@@ -38,24 +41,64 @@ func InitLogger(cfg *server.Config, reg prometheus.Registerer) {
...
+		logBufferSize uint32 = 10e6                   // 10MB
```
trevorwhitney, line 57: what happens if 256 entries goes over 10MB? Will the buffer also flush on size?

pkg/util/log/log.go
```diff
@@ -38,24 +41,64 @@ func InitLogger(cfg *server.Config, reg prometheus.Registerer) {
...
+		Namespace: "loki",
+		Name:      "log_flushes",
+		Help:      "Histogram of log flushes using the line-buffered logger.",
+		Buckets:   prometheus.ExponentialBuckets(1, 2, int(math.Log2(float64(logEntries)))+1),
+	})
+
+	var writer io.Writer
+	if buffered {
+		// TODO: it's technically possible here to lose logs between the 100ms flush and the process being killed
+		// 	=> call buf.Flush() in a signal handler if this is a concern, but this is unlikely to be a problem
```
trevorwhitney, line 76: is there a cost to also calling Flush in a signal handler? not necessary for this PR, but feels like a pretty cheap and easy way to hedge against this edge case.

trevorwhitney (approved):
This looks great, a welcome addition.

One other thought. Do we want to be able to flush this buffer when the `/flush` endpoint on an ingester is called (or any other stateful component we do manual scale downs of)?

pkg/util/log/log.go
dannykopping: The preallocated buffer is just to prevent slice resizing, and because it's reused it prevents unnecessary GC.
If 10MB is exceeded then the buffer's underlying slice will grow, so nothing will really happen - it won't flush on size.

pkg/util/log/log.go
dannykopping: No real cost, except last time I looked the signal handler is buried inside the weaveworks/common lib and it'll be a bit of a chore to work around that. Maybe I'm wrong though?

dannykopping (commented):
Thanks for the review!

dannykopping:
> One other thought. Do we want to be able to flush this buffer when the `/flush` endpoint on an ingester is called (or any other stateful component we do manual scale downs of)?

Calling the `/flush` endpoint will trigger a bunch of actions which will produce logs themselves - ultimately we just need a shutdown handler here to cover all cases I believe.

PR #6955, opened 2022-08-23
Helm: Publish Helm charts to grafana/helm-charts.

### LINKED ISSUE #6877: [Helm] Publish helm chart to grafana/helm-charts
While we are moving the helm chart source back to this repo, we don't want to change where the chart is hosted. As a result we need to incorporate the CI/GitHub Actions from https://github.com/grafana/helm-charts into this repo to publish the new chart to the same place.

* Migrate CI from grafana/helm-charts for testing helm chart changes
* Integrate the github action in the grafana/helm-charts repo (this will publish it, so we will tackle this last)

jeschkies:
This change pulls in the GitHub workflow for publishing Loki Helm charts to `grafana/helm-charts`.

Fixes #6877


.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,20 @@
...
+  call-update-helm-repo:
+    uses: grafana/helm-charts/.github/workflows/update-helm-repo.yaml@main
+    with:
+      charts_dir: production/helm
+      cr_configfile: production/helm/cr.yaml
+      ct_configfile: production/helm/ct.yaml
+    secrets:
+      helm_repo_token: ${{ secrets.GH_BOT_ACCESS_TOKEN }}
+      # See https://github.com/grafana/helm-charts/blob/main/INTERNAL.md about this key
+      gpg_key_base64: ${{ secrets.HELM_SIGN_KEY_BASE64 }}
```
jeschkies, line 20: @trevorwhitney do you know which key we used to use? There's a public key checked into `production/helm`. Should we generate a new key?

.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,20 @@
...
+    branches:
+      - main
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+
+jobs:
+  call-update-helm-repo:
+    uses: grafana/helm-charts/.github/workflows/update-helm-repo.yaml@main
+    with:
+      charts_dir: production/helm
```
jeschkies, line 14: I wonder whether this will also release `promtail` and `fluent-bit`.

.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,20 @@
+name: helm-release
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
```
jeschkies, line 8: Ideally we infer the Loki version. What do you think?

.github/workflows/helm-release.yaml
trevorwhitney: yikes, no, I have no idea. is there a way we can reverse engineer what key was being used in the `grafana/helm-charts` repo? maybe we could ask @krajorama and @Logiraptor what they did when the moved the mimir helm chart?

.github/workflows/helm-release.yaml
trevorwhitney: yes, my guess is it will. I think we could either a) remove everything but the README from those charts, and include the SHA they were deleted at (in a separate PR that we merge before this one), or b) can we sepecify a `charts_dir` of `production/helm/loki`?

I actually prefer option `a` as I think it cleans up the code base a bit.

.github/workflows/helm-release.yaml
trevorwhitney: What do you mean by that, not sure I'm following?

trevorwhitney (changes requested):
This looks great, thanks for adding this. I'm going to put a request changes on this just as a precaution as once this is merged the chart will publish, so I want to make sure we're good to go before that.

.github/workflows/helm-release.yaml
trevorwhitney: https://github.com/grafana/loki/pull/6978 <- PR removing the deprecated helm charts

.github/workflows/helm-release.yaml
Logiraptor: I'll have to defer to @krajorama on what was done. I can see a signing key exists in the settings of the mimir repo, but I can't view the secret contents.

.github/workflows/helm-release.yaml
jeschkies: I meant we infer the Loki version by the tag for this flow. This is not really connected to this line though.

trevorwhitney (approved):
LGTM

.github/workflows/helm-release.yaml
jeschkies: We are still missing the key.

.github/workflows/helm-release.yaml
slim-bean: I don't think we want this workflow to run on tags, we typically push the tag for the release in order to build the binaries and images, and _then_ we update the helm charts/jsonnet in a separate step only after the release artifacts exist.

trevorwhitney: I agree, let's remove the `tags` section alltogehter, which also more closely matches what mimir does:
```yaml  
push:
    branches:
      - main
      - "release-[0-9]+.[0-9]+"
```

Note: we don't wan the `release-` bit either. They've done some work to add releasing the helm chart to their release workflow, but we're not there yet.

PR #6956, opened 2022-08-23
tsdb: build rotated heads asynchronously

ashwanthgoli:
Currently `Append` call to the head has to wait for block creation to complete when rotating head.
And since we rotate head every 15m (default period), this would result in latency spikes at these times.
This PR moves the TSDB build step to an async routine.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/storage/stores/tsdb/head.go
+++ b/pkg/storage/stores/tsdb/head.go
@@ -61,6 +61,8 @@ type Metrics struct {
 	tsdbCreationFailures          prometheus.Counter
 	tsdbManagerUpdatesTotal       prometheus.Counter
 	tsdbManagerUpdatesFailedTotal prometheus.Counter
+	tsdbHeadRotationsTotal        prometheus.Counter
+	tsdbHeadRotationsFailedTotal  prometheus.Counter
 }
 
 func NewMetrics(r prometheus.Registerer) *Metrics {
@@ -84,6 +86,14 @@ func NewMetrics(r prometheus.Registerer) *Metrics {
 		tsdbManagerUpdatesFailedTotal: promauto.With(r).NewCounter(prometheus.CounterOpts{
 			Name: "loki_tsdb_manager_updates_failed_total",
 			Help: "Total number of tsdb manager update failures (loading/rotating tsdbs in mem)",
+		}),
+		tsdbHeadRotationsTotal: promauto.With(r).NewCounter(prometheus.CounterOpts{
+			Name: "loki_tsdb_head_rotations_total",
+			Help: "Total number of tsdb head rotations",
+		}),
+		tsdbHeadRotationsFailedTotal: promauto.With(r).NewCounter(prometheus.CounterOpts{
+			Name: "loki_tsdb_head_rotations_failed_total",
+			Help: "Total number of tsdb head rotations failed",
 		}),
 	}
 }
--- a/pkg/storage/stores/tsdb/head_manager.go
+++ b/pkg/storage/stores/tsdb/head_manager.go
@@ -187,11 +187,8 @@ func (m *HeadManager) loop() {
 						"msg", "failed rotating tsdb head",
 						"err", err,
 					)
-					m.metrics.tsdbHeadRotationsFailedTotal.Inc()
 					continue
 				}
-
-				m.metrics.tsdbHeadRotationsTotal.Inc()
 			}
 
 			// build tsdb from rotated-out period
@@ -314,7 +311,14 @@ func managerPerTenantDir(parent string) string {
 	return filepath.Join(parent, "per_tenant")
 }
 
-func (m *HeadManager) Rotate(t time.Time) error {
+func (m *HeadManager) Rotate(t time.Time) (err error) {
+	defer func() {
+		m.metrics.tsdbHeadRotationsTotal.Inc()
+		if err != nil {
+			m.metrics.tsdbHeadRotationsFailedTotal.Inc()
+		}
+	}()
+
 	// create new wal
 	nextWALPath := walPath(m.dir, t)
 	nextWAL, err := newHeadWAL(m.log, nextWALPath, t)
--- a/pkg/storage/stores/tsdb/head_manager.go
+++ b/pkg/storage/stores/tsdb/head_manager.go
@@ -152,11 +152,6 @@ func (m *HeadManager) loop() {
 		}
 
 		if err := m.buildTSDBFromWAL(m.prev.initialized); err != nil {
-			level.Error(m.log).Log(
-				"msg", "failed building tsdb head",
-				"period", m.period.PeriodFor(m.prev.initialized),
-				"err", err,
-			)
 			return errors.Wrap(err, "building tsdb head")
 		}
 
@@ -177,6 +172,10 @@ func (m *HeadManager) loop() {
 		case <-ticker.C:
 			// retry tsdb build failures from previous run
 			if err := buildPrev(); err != nil {
+				level.Error(m.log).Log(
+					"msg", "failed building tsdb head",
+					"err", err,
+				)
 				// rotating head without building prev would result in loss of index for that period (until restart)
 				continue
 			}
@@ -193,7 +192,12 @@ func (m *HeadManager) loop() {
 			}
 
 			// build tsdb from rotated-out period
-			buildPrev()
+			if err := buildPrev(); err != nil {
+				level.Error(m.log).Log(
+					"msg", "failed building tsdb head",
+					"err", err,
+				)
+			}
 		case <-m.cancel:
 			return
 		}
--- a/pkg/storage/stores/tsdb/head_manager.go
+++ b/pkg/storage/stores/tsdb/head_manager.go
@@ -146,26 +146,26 @@ func NewHeadManager(logger log.Logger, dir string, metrics *Metrics, tsdbManager
 func (m *HeadManager) loop() {
 	defer m.wg.Done()
 
+	buildPrev := func() error {
+		if err := m.buildTSDBFromWAL(m.prev.initialized); err != nil {
+			return errors.Wrap(err, "building tsdb head")
+		}
+
+		// Now that the tsdbManager has the updated TSDBs, we can remove our references
+		m.mtx.Lock()
+		defer m.mtx.Unlock()
+		m.prevHeads = nil
+		m.prev = nil
+
+		return nil
+	}
+
 	ticker := time.NewTicker(defaultRotationCheckPeriod)
 	defer ticker.Stop()
 
 	for {
 		select {
 		case <-ticker.C:
-			buildPrev := func() error {
-				if err := m.buildTSDBFromWAL(m.prev.initialized); err != nil {
-					return errors.Wrap(err, "building tsdb head")
-				}
-
-				// Now that the tsdbManager has the updated TSDBs, we can remove our references
-				m.mtx.Lock()
-				defer m.mtx.Unlock()
-				m.prevHeads = nil
-				m.prev = nil
-
-				return nil
-			}
-
 			// retry tsdb build failures from previous run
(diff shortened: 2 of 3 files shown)


pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -131,10 +140,47 @@ func NewHeadManager(logger log.Logger, dir string, metrics *Metrics, tsdbManager
...
+			m.mtx.Lock()
+			if m.prev != nil {
+				if err := m.buildTSDBFromWAL(m.prev.initialized); err != nil {
+					level.Error(m.log).Log(
+						"msg", "failed building tsdb from rotated out period",
+						"period", m.period.PeriodFor(m.prev.initialized),
+						"err", err,
+					)
+					// keeps retrying on tick until it gets rotated out
+					// should we be worried about this?
```
chaudum, line 167: This would at least result in a different behaviour than there is now. I guess if building the TSDB once, it would very likely fail with the same error again in subsequent iterations.

ashwanthgoli: agree, currently this would result in `Append` failure
Not sure how likely this is to fail, should I add a metric instead for some visibility?

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -131,10 +140,47 @@ func NewHeadManager(logger log.Logger, dir string, metrics *Metrics, tsdbManager
 
 	})
 
+	go m.loop()
+
 	return m
 }
 
+func (m *HeadManager) loop() {
+	m.loopWg.Add(1)
```
kavirajk, line 149: Wondering why `wg.Add()` inside the spinning go routine itself (go `loop()`)?. I think `Add()` should be called before actually spinning the go routine.

https://pkg.go.dev/sync#WaitGroup.Add
```
Typically this means the calls to Add should execute before the statement creating the goroutine or other event to be waited for. If a WaitGroup is reused to wait for several independent sets of events, new Add calls must happen after all previous Wait calls have returned. See the WaitGroup example.
```

pkg/storage/stores/tsdb/head_manager.go
kavirajk: curious isn't `m.prev != nil` takes care of `Append()` failure? before building TSDB from WAL?

pkg/storage/stores/tsdb/head_manager.go
ashwanthgoli: agree, will move it out

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -101,6 +105,9 @@ type HeadManager struct {
 	activeHeads, prevHeads *tenantHeads
 
 	Index
+
+	loopWg   sync.WaitGroup
```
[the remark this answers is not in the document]
ashwanthgoli: I went with the naming convention from ingester.go. `cancel` on it's own would suggest that it's cancelling the HeadManager but it's only handling the loop part atm.
I don't have any strong opinion on this though, can rename these to a single token which might be more inline with go naming conventions :)

pkg/storage/stores/tsdb/head_manager.go
ashwanthgoli: not sure if I understood the question correctly :/

actually the next `Append` won't even call the `Rotate` method since the new head covers that period.
So without this PR. we would only try to build prev head once (as part of `Append`), if that fails we leave it for recovery from WAL during start-up

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -272,17 +318,6 @@ func (m *HeadManager) Rotate(t time.Time) error {
 	m.activeHeads = nextHeads
```
sandeepsukhani, line 318: Let us move all the logic of rotating the head to the ticker loop above to keep it in one place. This approach would also help us avoid the possibility of losing the index when the loop is failing to build the index, but the `Append` keeps rotating the index every 15m without knowing about it.

Here is how I imaging loop would work:
* if `prev` is not nil: `// This is for handling the case where the loop rotated the head but failed to build the index.`
   * Build TSDB from the WAL.
   * If the TSDB got built successfully, rotate the head and let the next tick take care of building the index.
* else:
   * Rotate the head and build the index from `prev`.
   * Set `prev` to nil if we built the index successfully.

This change would mean that `Append` would always keep writing to the active head. It would be the loop taking care of rotating the head and building TSDB out of it. If the loop fails to build the TSDB, it will keep retrying it during each tick, while the `Append` could keep writing to the active head.

Here we should make sure we take the exclusive lock on the mutex for as little as possible, e.g. while changing the references of `m.prev*` and `m.active*` or any other operation which can impact concurrent operations. 

We should also add a metric to track the failures and successes in building the index so we can alert on it.

Does it make sense?

sandeepsukhani (commented):
Left some suggestions, but other than that, it LGTM! Congratulations on the first PR 🎉

pkg/storage/stores/tsdb/head_manager.go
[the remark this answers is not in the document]
kavirajk: it's just a `nit` feel free to ignore :)

pkg/storage/stores/tsdb/head_manager.go
owen-d: This sounds good to me.

I think we can merge once the following are addressed:
> Let us move all the logic of rotating the head to the ticker loop above to keep it in one place. 

> if prev is not nil: `// This is for handling the case where the loop rotated the head but failed to build the index.`

pkg/storage/stores/tsdb/head.go
```diff
@@ -85,6 +87,14 @@ func NewMetrics(r prometheus.Registerer) *Metrics {
...
 			Help: "Total number of tsdb manager update failures (loading/rotating tsdbs in mem)",
 		}),
+		tsdbHeadRotationsTotal: promauto.With(r).NewCounter(prometheus.CounterOpts{
+			Name: "loki_tsdb_head_rotations_total",
+			Help: "Total number of tsdb head rotations",
+		}),
+		tsdbHeadRotationsFailedTotal: promauto.With(r).NewCounter(prometheus.CounterOpts{
+			Name: "loki_tsdb_head_rotations_failed_total",
+			Help: "Total number of tsdb head rotations failed",
+		}),
```
sandeepsukhani, line 97: you can have just a single metric with `status` label which would have values `success` / `fail`.

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -134,7 +143,78 @@ func NewHeadManager(logger log.Logger, dir string, metrics *Metrics, tsdbManager
...
+				}
+
+				// Now that the tsdbManager has the updated TSDBs, we can remove our references
+				m.mtx.Lock()
+				defer m.mtx.Unlock()
+				m.prevHeads = nil
+				m.prev = nil
+
+				return nil
+			}
```
sandeepsukhani, line 167: we can move this outside of the for loop, maybe just before it is defined.

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -134,7 +143,78 @@ func NewHeadManager(logger log.Logger, dir string, metrics *Metrics, tsdbManager
...
+	defer m.wg.Done()
+
+	ticker := time.NewTicker(defaultRotationCheckPeriod)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			buildPrev := func() error {
+				if err := m.buildTSDBFromWAL(m.prev.initialized); err != nil {
```
sandeepsukhani, line 156: I think we should track building tsdb from wal in a metric too. I feel it also has higher chances on failure due to its complexity compared to rotation of heads.

sandeepsukhani (commented):
Thanks for taking care of the feedback! It looks good now.
Left a very few minor nits, but other than that it LGTM!

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -134,7 +143,78 @@ func NewHeadManager(logger log.Logger, dir string, metrics *Metrics, tsdbManager
...
+				if err := m.Rotate(now); err != nil {
+					level.Error(m.log).Log(
+						"msg", "failed rotating tsdb head",
+						"err", err,
+					)
+					m.metrics.tsdbHeadRotationsFailedTotal.Inc()
+					continue
+				}
+
+				m.metrics.tsdbHeadRotationsTotal.Inc()
```
owen-d, line 194: This metric isn't being incremented if there's an error, which would be a mistake. Taking @sandeepsukhani 's suggestion on metric refactoring would fix this problem.

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -134,7 +143,78 @@ func NewHeadManager(logger log.Logger, dir string, metrics *Metrics, tsdbManager
...
+
+	ticker := time.NewTicker(defaultRotationCheckPeriod)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			// retry tsdb build failures from previous run
+			if m.prev != nil {
+				err := buildPrev()
```
owen-d, line 171: We have metrics for rotations, but not for building TSDBs. I think we should either track both operations or combine them into the same metric.

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -134,7 +143,75 @@ func NewHeadManager(logger log.Logger, dir string, metrics *Metrics, tsdbManager
...
+				if err != nil {
+					level.Error(m.log).Log(
+						"msg", "failed building tsdb head",
+						"period", m.period.PeriodFor(m.prev.initialized),
+						"err", err,
+					)
+				}
+			}
```
owen-d, line 204: It looks like this all of this is duplicated a few lines up. I think we can put it all in the helper function.

pkg/storage/stores/tsdb/head_manager.go
owen-d: We still need to add metrics for when `buildTSDBFromWAL` fails.

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -134,7 +143,71 @@ func NewHeadManager(logger log.Logger, dir string, metrics *Metrics, tsdbManager
...
+	ticker := time.NewTicker(defaultRotationCheckPeriod)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			// retry tsdb build failures from previous run
+			if err := buildPrev(); err != nil {
+				level.Error(m.log).Log(
+					"msg", "failed building tsdb head",
```
owen-d, line 176:
```suggestion
					"msg", "failed building tsdb head",
					"period", m.period.PeriodFor(m.prev.initialized),
```

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -134,7 +143,71 @@ func NewHeadManager(logger log.Logger, dir string, metrics *Metrics, tsdbManager
...
+				)
+				// rotating head without building prev would result in loss of index for that period (until restart)
+				continue
+			}
+
+			now := time.Now()
+			if m.period.PeriodFor(now) > m.period.PeriodFor(m.activeHeads.start) {
+				if err := m.Rotate(now); err != nil {
+					level.Error(m.log).Log(
+						"msg", "failed rotating tsdb head",
```
owen-d, line 187:
```suggestion
						"msg", "failed rotating tsdb head",
						"period", m.period.PeriodFor(m.prev.initialized),
```

owen-d (approved):

PR #6403, opened 2022-06-15
Bump prometheus dependency

jesusvazquez:
This PR updates prometheus dependency to its latest commit. It comes with quite a few changes so I've made a list of commits that make it easy to review them.

PR #6957, opened 2022-08-23
Upgrade prometheus/client_golang and reconfigure to restore `go_sched.*` metrics

dannykopping:
These metrics were removed unintentionally in #6403.

The `go_sched_latencies` histogram is very important for determining pathological scheduler behaviour (I need it to verify the impact of https://github.com/grafana/loki/pull/6954).

N/A

Don't be afraid of the PR size; it's mainly just revendored code.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


pkg/loki/modules.go
```diff
@@ -96,8 +97,23 @@ const (
 	UsageReport              string = "usage-report"
 )
 
+func (t *Loki) initRegisterer() prometheus.Registerer {
```
MichelHollands, line 100: nit: The name of this function makes it look like there is a module called Registerer.

MichelHollands (approved):
LGTM

dannykopping:
@DylanGuedes raised a point in Slack: will the upgrade of `prometheus/common` from `v0.35.0` to `v0.37.0` present any challenges?

https://github.com/prometheus/common/compare/v0.35.0...v0.37.0

The diff doesn't seem to show anything too scary, and unfortunately there is no changelog in `prometheus/common`, so we'll just have to see...

pkg/loki/modules.go
dannykopping: That's true, I think I'll leave this for now and refactor it when I get rid of the reliance on the `prometheus.DefaultXXX` references

periklis:
@dannykopping Can we consider backporting this to 2.6? My key pain point is: 
```
[BUGFIX] Raise exemplar labels limit from 64 to 128 bytes as specified in OpenMetrics spec. https://github.com/prometheus/client_golang/pull/1091
```

dannykopping:
Sure :+1:

PR #6958, opened 2022-08-23
Support helm migration from current grafana/loki chart

trevorwhitney:

These are changes to the helm chart and documentation need to migration from the current version of the `grafana/loki` helm chart to the new one in this repo.

The current plan (as outlined in the README) is to support 3 upgrade paths:

1. From the existing `grafana/loki` chart using filesystem storage. This path will result in a single binary deployment using local filesystem storage with the new chart. This PRs includes changes to make sure this path works.
2. ~From the existing `grafana/loki` chart using object storage. This path will result in an SSD deployment with 1 read and 1 write), and has yet to be tested/documented.~ (Edit: I no longer think we need to support this use case)
3. From the existing `grafana/loki-simple-scalable` chart using object storage. This path will result in an SSD reployment with 3 read and 3 write replicas, and has yet to be tested/documented.

By clarifying the three modes, and that single binary will only support a single, filesystem backed instance, it made possible some refactors. For example, the user no longer needs to specify a `deploymentMode`, and we were able to reorganize the templates a bit into a flatter folder structure.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


production/helm/loki/README.md
```diff
@@ -25,6 +25,119 @@ Add the following repo to use the chart:
...
+
+  1. Upgrade from `grafana/loki` using local `filesystem` storage
+  1. Upgrade from `grafana/loki` using a cloud based object storage such as S3 or GCS, or an api compatible equivilent like MinIO.
+  1. Upgrade from `grafana/loki-simple-scalable` using a cloud based object storage such as S3 or GCS, or an api compatible equivilent like MinIO.
+
+### Upgrading from `grafana/loki`
+
+#### Upgrading from `filesystem` storage
+
+When Loki is backed by `filesystem` storage, we assume a single instance of Loki that is not HA. As a result, this upgrade method will involve downtime. The upgrade will involve deleting the previously deployed loki stateful set, the running the `helm upgrade` which will create the new one with the same name, which should attach to the existing PVC or ephemeral storage, thus preserving you data. Will still highly recommend backing up all data before conducting the upgrade.
```
jeschkies, line 44:
```suggestion
When Loki is backed by `filesystem` storage, we assume a single instance of Loki that is not HA. As a result, this upgrade method will involve downtime. The upgrade will involve deleting the previously deployed Loki stateful set, running the `helm upgrade` which will create the new one with the same name, which should attach to the existing PVC or ephemeral storage, thus preserving you data. Will still recommend backing up all data before conducting the upgrade.
```

production/helm/loki/README.md
```diff
@@ -25,6 +25,119 @@ Add the following repo to use the chart:
...
+    type: 'filesystem'
+```
+
+Assuming you installed `grafana/loki` as `loki` in namespace `loki`, your upgrade command might look like:
+
+```console
+helm upgrade loki ~/workspace/grafana/loki/production/helm/loki --values environments/loki-migration-test/values-new.yaml --namespace loki
+```
+
+You will need to manually delete the existing stateful set for the above command to work.
```
jeschkies, line 63: Could you give the `kubectl` command for this?

production/helm/loki/README.md
```diff
@@ -25,6 +25,119 @@ Add the following repo to use the chart:
...
+
+### Upgrading from `grafana/loki-simple-scalable`
+
+## Configuration
+
+By default, this chart configures Loki to run `read` and `write` targets in a scalable architecture
+designed to work with object storage. This configuration has been tested to work with [boltdb-shipper](https://grafana.com/docs/loki/latest/operations/storage/boltdb-shipper/)
+and [memberlist](https://grafana.com/docs/loki/latest/configuration/#memberlist_config) while other storage and discovery options should work as well.
```
jeschkies, line 71: Not sure this is worth mentioning here.

production/helm/loki/README.md
```diff
@@ -25,6 +25,119 @@ Add the following repo to use the chart:
...
+
+If this chart is configured to run with a filesystem backend, it will assume you only want a single instance of Loki deployed
```
jeschkies, line 73:
```suggestion
If this chart is configured to run with a `filesystem` backend, it will assume you only want a single instance of Loki deployed
```

production/helm/loki/README.md
```diff
@@ -25,6 +25,119 @@ Add the following repo to use the chart:
...
+in "Single Binary" mode.
```
jeschkies, line 74:
```suggestion
in "Single Binary" mode. See "Upgrading from `filesystem` storage" instead.
```

production/helm/loki/README.md
```diff
@@ -25,6 +25,119 @@ Add the following repo to use the chart:
...
+
+However, the chart does not support setting up Consul or Etcd for discovery,
+and it is not intended to support these going forward.
+They would have to be set up separately.
+Instead, memberlist can be used which does not require a separate key/value store.
+The chart creates a headless service for the memberlist which read and write nodes are part of.
```
jeschkies, line 80: Not sure how this helps me upgrading.

production/helm/loki/templates/_helpers.tpl
```diff
@@ -6,12 +6,63 @@ Expand the name of the chart.
 {{- coalesce .Values.nameOverride $default | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/*
+singleBinary fullname
+*/}}
```
jeschkies, line 11: What do these helpers do?

production/helm/loki/templates/service-headless.yaml
```diff
@@ -0,0 +1,22 @@
...
+    - name: http-metrics
+      port: 3100
+      targetPort: http-metrics
+      protocol: TCP
+  selector:
+    {{- if eq (include "loki.requiresMultiStageUpgrade" . ) "true" }}
+    {{- include "loki.legacy.singleBinarySelectorLabels" . | nindent 4 }}
+    {{- else -}}
+    {{- include "loki.selectorLabels" . | nindent 4 }}
+    {{- end }}
```
jeschkies, line 22: Why is this required?

jeschkies (commented):
I have a few questions below. Also, how did you run the upgrade test?

production/helm/loki/templates/_helpers.tpl
trevorwhitney: the way the `loki.name` and `loki.fullname` work between the existing `grafana/loki` and `grafana/loki-simple-scalable` charts differs a bit. To enable the migration path where the new chart will replace the stateful set created by `grafana/loki` we needed to bring in this function from `grafana/loki` to make sure they had the same name.

The helpers below `isSingleBinary` and `isScalable` determine which set of templates to deploy based on whether or not object storage was requested.

production/helm/loki/README.md.gotmpl
```diff
@@ -167,8 +101,37 @@ This chart includes dashboards for monitoring Loki. These are a work in progress
...
+        rules:
+          - record: job:loki_request_duration_seconds_bucket:sum_rate
+            expr: sum(rate(loki_request_duration_seconds_bucket[1m])) by (le, job)
+          - record: job_route:loki_request_duration_seconds_bucket:sum_rate
+            expr: sum(rate(loki_request_duration_seconds_bucket[1m])) by (le, job, route)
+          - record: node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate
+            expr: sum(rate(container_cpu_usage_seconds_total[1m])) by (node, namespace, pod, container)
+```
+
+{{ template "chart.valuesSection" . }}
```
trevorwhitney, line 137: values section moved to the bottom of README

production/helm/loki/README.md.gotmpl
```diff
@@ -16,140 +16,74 @@ Add the following repo to use the chart:
...
-By default and inspired by Grafana's [Tanka setup](https://github.com/grafana/loki/tree/master/production/ksonnet/loki), the chart installs the gateway component which is an NGINX that exposes Loki's API
-and automatically proxies requests to the correct Loki components (read or write).
+By default and inspired by Grafana's [Tanka setup](https://github.com/grafana/loki/tree/master/production/ksonnet/loki), the chart
+installs the gateway component which is an NGINX that exposes Loki's API and automatically proxies requests to the correct
+Loki components (read or write, or single instance in the case of filesystem storage).
 The gateway must be enabled if an Ingress is required, since the Ingress exposes the gateway only.
 If the gateway is enabled, Grafana and log shipping agents, such as Promtail, should be configured to use the gateway.
 If NetworkPolicies are enabled, they are more restrictive if the gateway is enabled.
 
-## Metrics
```
trevorwhitney, line 128: Metrics section refactored into the monitoring section

production/helm/loki/README.md.gotmpl
```diff
@@ -16,140 +16,74 @@ Add the following repo to use the chart:
...
-The chart supports the components shown in the following table.
-Ingester, distributor, querier, and query-frontend are always installed.
-The other components are optional.
-
-| Component | Optional | Enabled by default |
-| --- | --- | --- |
-| gateway |  ✅ |  ✅ |
-| write |  ❎ | n/a |
-| read |  ❎ | n/a |
+## Upgrading from v2.x
```
trevorwhitney, line 19: Added upgrade/migration documentation. Currently only the first approach is spelled out as that's the only one tested in this PR.

production/helm/loki/templates/gateway/configmap-gateway.yaml
```diff
@@ -1,4 +1,4 @@
-{{- $isSimpleScalable := eq (include "loki.deployment.isSimpleScalable" .) "true" -}}
+{{- $isSimpleScalable := eq (include "loki.deployment.isScalable" .) "true" -}}
```
trevorwhitney, line 1: using this opportunity to get away from the terms "Simple Scalable". This has confused community members who figured "Simple" meant not production ready.

production/helm/loki/templates/_helpers.tpl
```diff
@@ -240,17 +270,13 @@ Create the service endpoint including port for MinIO.
...
 {{- end -}}
 
-{{/*
-Return if deployment mode is single binary
-*/}}
-{{- define "loki.deployment.isSingleBinary" -}}
-  {{- eq .Values.loki.deploymentMode "single-binary" }}
+{{/* Configure the correct name for the memberlist service */}}
+{{- define "loki.memberlist" -}}
+{{ include "loki.name" . }}-memberlist
```
trevorwhitney, line 280: This is a future proofing. By moving this into a function, it will make us easier to support migrations where we honor an existing memberlist DNS to enable the new deployment to join the same ring as an existing deployment. Worst case we don't need that and this remains a one line function, which I think is worth it.

production/helm/loki/README.md
trevorwhitney: the assumption is you are using `helm`, thus there is no `kubectl` command. If a user want to run `helm template` to `kubectl` the raw yaml themselves I would consider that an advanced technique we don't need to document.

trevorwhitney:
@jeschkies for the upgrade test, I first do `helm install loki grafana/loki --values values-old.yaml --namespace loki`, then after it's installed and stabalized, I run `helm upgrade loki ${PATH_TO_LOKI_REPO}/production/helm/loki --values values-new.yaml --namespace loki`.

`values-old.yaml`:
```yaml
---
auth_enabled: false
persistence:
  enabled: true
```

`values-new.yaml`:
```yaml
---
loki:
  commonConfig:
    replication_factor: 1
  storage:
    type: "filesystem"
```

production/helm/loki/templates/service-headless.yaml
trevorwhitney: That's a good question and I've decided it's not, see my refactor.

I do think we'll need something like this when we figure out how to do a zero downtime migration, but I only think we need that for people migrating from distributed or enterprise SSD charts. Since this PR only address the migration from filesystem backed single binary, I think it's safe to assume HA is not required.

The idea here is that if we want to be able to spin up the new cluster along-side an existing cluster to do a multi-stage, zero downtime deployment, we'll need the new deployment to include the legacy label set, and we'll need to new services to use the legacy label selector so it captures both new and old pods. However, like I mentioned, not needed yet.

production/helm/loki/README.md
trevorwhitney: hmm, yeah, let's just nuke it

production/helm/loki/templates/service-headless.yaml
jeschkies: Thanks for explaining!

jeschkies (commented):
I'd approve but the PR states that the migration is tested. Are you planning to add tests or will this be done in a follow up?

jeschkies (approved):

production/helm/loki/README.md
trevorwhitney: ahh, yes, haha, and I suppose hit `<ctrl-d>` in `k9s` isn't appropriate, even though it's what I always use now so I've forgotten the `kubectl` command

production/helm/loki/README.md
trevorwhitney: I added this, just in a slightly different way

PR #6964, opened 2022-08-24
Evenly spread distributors across available nodes

salvacorts:

This is equivalent to https://github.com/grafana/loki/pull/6415 but for distributors.
Why only distributors?
- Ingesters are critical and stateful, so we cannot risk losing multiple ingesters on a node failure.
- Other workloads do not typically have a high number of replicas, therefore, the risk of losing multiple replicas on a node failure outweighs the benefits.

**Checklist**
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/upgrading/_index.md
+++ b/docs/sources/upgrading/_index.md
@@ -44,10 +44,18 @@ Currently `embedded-cache` with `distributed: true` can be enabled only for resu
 We now evenly spread distributors across the available kubernetes nodes, but allowing more than one distributors to be scheduled into the same node.
 If you want to run at most a single distributors per node, set `$._config.distributors.use_topology_spread` to false.
 
+While we attempt to schedule at most 1 distributor per Kubernetes node with the `topology_spread_max_skew: 1` field,
+if no more nodes are available then multiple distributors will be scheduled on the same node.
+This can potentially impact your service's reliability so consider tuning these values according to your risk tolerance.
+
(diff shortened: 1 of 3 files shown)


docs/sources/upgrading/_index.md
```diff
@@ -39,6 +39,11 @@ We introduced a new cache called `embedded-cache` which is an in-process cache s
 
 Currently `embedded-cache` with `distributed: true` can be enabled only for results cache.
 
+#### Evenly spread distributors across kubernetes nodes
+
+We now evenly spread distributors across the available kubernetes nodes, but allowing more than one distributors to be scheduled into the same node.
+If you want to run at most a single distributors per node, set `$._config.distributors.use_topology_spread` to false.
```
dannykopping, line 45: I wonder if it's not worth adding to this and the queriers note below:

> While we attempt to schedule at most 1 pod per Kubernetes node with the `topology_spread_max_skew: 1` field, if no more nodes are available then multiple pods will be scheduled on the same node. This can potentially impact your service's reliability so consider tuning these values according to your risk tolerance.

dannykopping (approved):
LGTM, with a note

docs/sources/upgrading/_index.md
salvacorts: Sounds good. Added it to both the distributors and queriers upgrading notes. Replaced "pod" with "querier" and "distributor" respectively

PR #6972, opened 2022-08-25
Return Querier/QueryTimeout YAML configuration

DylanGuedes:
On PR #6835 I mistakenly removed the `querier:query_timeout` YAML configuration. This PR returns it and uses it if it is still configured, while logging a warning message to alert users of the upcoming changes.


dannykopping (approved):
LGTM, thanks

PR #6977, opened 2022-08-25
Loki: Modify ingesters to return rate-limited streams to distributors

DylanGuedes:
Modify ingesters to return a list of rate-limited streams on its reply to `(*distributor).Push`. This is done by using the gRPC details feature. I've updated the tests and apparently, the returned error message is exactly the same so this shouldn't be a breaking change.

Also modifies `(*stream).Push` to return a list of all entries with errors instead of returning the final gRPC error. This will allow us in the future to easily return to distributors the indices of all reject entries, so that distributors won't have to always retry all entries.

This work is necessary for the new stream sharding feature.


pkg/ingester/instance.go
```diff
@@ -211,6 +221,73 @@ func (i *instance) Push(ctx context.Context, req *logproto.PushRequest) error {
...
+
+	fmt.Fprintf(&buf, "total ignored: %d out of %d", len(failedEntriesWithError), totalEntries)
+
+	var details []*types.Any
+
+	// Populate error details with the current stream.
+	if hadPerStreamError {
+		rls := logproto.RateLimitedStream{Labels: streamName}
+		marshalledStream, err := types.MarshalAny(&rls)
+		if err == nil {
```
MasslessParticle, line 279: We probably want to log if `MarshalAny` has an error because it means something very strange is happening

pkg/ingester/recovery.go
```diff
@@ -165,10 +165,13 @@ func (r *ingesterRecoverer) Push(userID string, entries RefEntries) error {
...
 
 		// ignore out of order errors here (it's possible for a checkpoint to already have data from the wal segments)
-		bytesAdded, err := s.(*stream).Push(context.Background(), entries.Entries, nil, entries.Counter, true)
+		bytesAdded, entriesWithErrors := s.(*stream).Push(context.Background(), entries.Entries, nil, entries.Counter, true)
 		r.ing.replayController.Add(int64(bytesAdded))
-		if err != nil && err == ErrEntriesExist {
-			r.ing.metrics.duplicateEntriesTotal.Add(float64(len(entries.Entries)))
+		if len(entriesWithErrors) > 0 {
+			lastEntryWithError := entriesWithErrors[len(entriesWithErrors)-1]
+			if lastEntryWithError.e == ErrEntriesExist {
```
MasslessParticle, line 172: Using `errors.Is` here will help us not be tied to this _exact_ error

pkg/ingester/instance.go
```diff
@@ -211,6 +221,73 @@ func (i *instance) Push(ctx context.Context, req *logproto.PushRequest) error {
 	return appendErr
 }
 
+// mountGRPCError mounts an error to be returned in a GRPC call.
+//
+// The returned error is enriched with the gRPC error status and with a list of details.
+// As of now, the list of details is a list of all streams that were rate-limited. This list can be used
+//   by the distributor to fine tune streams that were limited.
+func mountGRPCError(s *stream, failedEntriesWithError []entryWithError, totalEntries int) error {
```
MasslessParticle, line 229: What do you think of calling this something like `errorForFailedEntries`?

pkg/ingester/instance.go
```diff
@@ -211,6 +221,73 @@ func (i *instance) Push(ctx context.Context, req *logproto.PushRequest) error {
...
+	streamName := s.labelsString
+
+	limitedFailedEntries := failedEntriesWithError
+	if maxIgnore := s.cfg.MaxReturnedErrors; maxIgnore > 0 && len(limitedFailedEntries) > maxIgnore {
+		limitedFailedEntries = limitedFailedEntries[:maxIgnore]
+	}
+
+	hadPerStreamError := false
+	for _, entryWithError := range limitedFailedEntries {
+		if !hadPerStreamError {
```
MasslessParticle, line 261: We don't need to check for `perStreamError` here because we already have the status code `http.StatusTooManyRequests`

pkg/ingester/instance.go
```diff
@@ -211,6 +221,73 @@ func (i *instance) Push(ctx context.Context, req *logproto.PushRequest) error {
...
+			"entry with timestamp %s ignored, reason: '%s' for stream: %s,\n",
+			entryWithError.entry.Timestamp.String(), entryWithError.e.Error(), streamName)
+	}
+
+	fmt.Fprintf(&buf, "total ignored: %d out of %d", len(failedEntriesWithError), totalEntries)
+
+	var details []*types.Any
+
+	// Populate error details with the current stream.
+	if hadPerStreamError {
```
MasslessParticle, line 276: Just need to check if `statusCode == http.StatusTooManyRequests` here

pkg/ingester/instance.go
```diff
@@ -211,6 +221,73 @@ func (i *instance) Push(ctx context.Context, req *logproto.PushRequest) error {
...
+		fmt.Fprintf(&buf,
+			"entry with timestamp %s ignored, reason: '%s' for stream: %s,\n",
+			entryWithError.entry.Timestamp.String(), entryWithError.e.Error(), streamName)
+	}
+
+	fmt.Fprintf(&buf, "total ignored: %d out of %d", len(failedEntriesWithError), totalEntries)
+
+	var details []*types.Any
+
+	// Populate error details with the current stream.
```
MasslessParticle, line 275: Consider making this chunk of code a function called something like `errorDetails` rather than using a comment.

MasslessParticle (changes requested):
This is an awesome start. There are some things we should do before merging and some stylistic suggestions.

pkg/ingester/instance.go
DylanGuedes: But we also use `http.StatusTooManyRequests` for the total ingestion rate limit (which is different from the per-stream rate limit). If we mark a stream as rate-limited when it actually did hit the total ingestion rate, we might end up wrongly sharding a stream that shouldn't.

WDYT?

pkg/ingester/instance.go
DylanGuedes: ditto

pkg/ingester/instance.go
DylanGuedes: Good call. I pushed a new one but not sure if the naming is good, WDYT?

pkg/ingester/instance.go
MasslessParticle: The errors are checked in different places. The distributor returns 429 for global rate limiting so the only 429s we'll ever get from the ingester are for per stream rate limiting.

MasslessParticle (approved):

PR #6978, opened 2022-08-25
Remove deprecated helm chart source code

trevorwhitney:

All of these helm charts have been moved out of this repo to `grafana/helm-charts`. As we are now moving `grafana/loki` back into this repo, this deprecated charts are confusing. I have opted to include `README`s with the deprecation notice, as well as a link to the current sha of `main` so that people can easily find the old source code.

This looks like a much bigger PR than it is. The vast majority of the 44 files touched are just deletes of the whole file.


slim-bean (approved):
makes sense to me to remove these, should have done it long ago

PR #6983, opened 2022-08-26
Loki: add `__line__` and `__timestamp__` to label_format

### LINKED ISSUE #6402: __line__ function not defined in label_format()
**Describe the bug**
The [documentation for template functions](https://grafana.com/docs/loki/latest/logql/template_functions/) mentions that line values can be accessed with `__line__`. In #4879 (Loki 2.5.0) it does seem that this was added to `line_format()` but it does not seem to be implemented for `label_format()`

**To Reproduce**
1. Started Loki (2.5.0)
2. Started Promtail to tail systemd journal
3. Query: `{unit="cron.service"} | label_format line=`{{ __line__ }}``

**Expected behavior**
Log output with `line` label set to value of line.

**Actual behavior**
Query error: `rpc error: code = Unknown desc = parse error : stage '| label_format line="{{ __line__ }}"' : invalid template for label 'line': template: label:1: function "__line__" not defined`

image

slim-bean:

We have the `__line__` and `__timestamp__` functions currently available in the `line_format` stage but not in the `label_format` stage. This PR fixes that.

Specifically `__timestamp__` can be valuable in a label because you can do things like

```
label_format hour_of_day=`{{__timestamp__ | date "15"}}`
```

To be able to conditionaly select log lines based on the hour of the day they were created, this can be useful for conditional alerting as an example.

Fixes #6402

**Checklist**
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/logql/log/fmt.go
+++ b/pkg/logql/log/fmt.go
@@ -90,6 +90,20 @@ var (
 	}
 )
 
+func addLineAndTimestampFunctions(currLine func() string, currTimestamp func() int64) map[string]interface{} {
+	functions := make(map[string]interface{}, len(functionMap)+2)
+	for k, v := range functionMap {
+		functions[k] = v
+	}
+	functions[functionLineName] = func() string {
+		return currLine()
+	}
+	functions[functionTimestampName] = func() time.Time {
+		return time.Unix(0, currTimestamp())
+	}
+	return functions
+}
+
 func init() {
 	sprigFuncMap := sprig.GenericFuncMap()
 	for _, v := range templateFunctions {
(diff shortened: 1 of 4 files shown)


pkg/logql/log/fmt.go
```diff
@@ -90,6 +90,20 @@ var (
...
 
+func addLineAndTimestampFunctions(currLine func() string, currTimestamp func() int64) map[string]interface{} {
+	functions := make(map[string]interface{}, len(functionMap)+2)
+	for k, v := range functionMap {
+		functions[k] = v
+	}
+	functions[functionLineName] = func() string {
+		return currLine()
+	}
+	functions[functionTimestampName] = func() time.Time {
```
DylanGuedes, line 101: newb question: shouldn't you have a separate function for timestamp and another one for the unix timestamp?

pkg/logql/log/fmt.go
```diff
@@ -246,21 +260,29 @@ func NewLabelsFormatter(fmts []LabelFmt) (*LabelsFormatter, error) {
 	}
 	formats := make([]labelFormatter, 0, len(fmts))
```
DylanGuedes, line 261: nit: it can be len(fmts)+2 now?

pkg/logql/log/fmt.go
slim-bean: this "function" creates the timestamp internally as a `date` object which can be combined with other functions like `unixEpoch` and `toDate` which let you format however you would like.

I'm gonna have an additional PR to handle formatting the unix timestamp as millis or nanos

pkg/logql/log/fmt.go
slim-bean: I don't think this changes? I'm adding to the function map but not the formats slice. 

The functions list does get incremented by 2 to add the 2 new functions

https://github.com/grafana/loki/blob/286dbfbde38a035a4f23f2b41b1e5b658a0650e0/pkg/logql/log/fmt.go#L94

pkg/logql/log/fmt.go
DylanGuedes: got it, ty

DylanGuedes (approved):

PR #6988, opened 2022-08-26
Helm: Add helm ci

trevorwhitney:

Tests Helm chart when changed as part of PR CI process.


trevorwhitney:
CI is trying to run over all the old deprecated charts. I think we should merge https://github.com/grafana/loki/pull/6978 to solve that.

jeschkies (approved):

PR #6990, opened 2022-08-26
querier deployment rollingupdate works as a percentage to allow faste…

owen-d:
…r rollouts in large environments


slim-bean (approved):
LGTM

PR #6999, opened 2022-08-29
operator: Adding Lokistack Gateway Request Errors Alert

Red-GV:
When a Loki component is removed (via moving the operator into the unmanaged state), it will not be able to produce a metric showing that it is unable to respond to the request. Therefore, an additional metric for the metrics emitted by the Lokistack Gateway should be used to capture these failed attempts and create alerts notifying the user(s) that something is wrong.

**Checklist**
- [x] Documentation added
- [x] Tests updated


operator/docs/lokistack/sop.md
```diff
@@ -46,6 +46,36 @@ A service(s) is failing to process at least 10% of all incoming requests.
...
+
+`Critical`
+
+### Access Required
+
+- Console access to the cluster
+- Edit access to the deployed operator and Loki namespace:
+  - OpenShift
+    - `openshift-logging`
+    - `openshift-operators-redhat`
```
periklis, line 69: We should place a hint which namespace hosts which component, e.g.:
```
openshift-logging (LokiStack)
openshift-operators-redhat (Loki Operator)
```

WDYT?

operator/docs/lokistack/sop.md
```diff
@@ -46,6 +46,36 @@ A service(s) is failing to process at least 10% of all incoming requests.
...
+
+### Steps
+
+- Ensure that the deployed Loki components are ready and available
+- Ensure that store services (`ingester`, `querier`, `index-gateway`, `compactor`) can communicate with backend storage
+- Examine metrics for signs of failure
+  - WAL Complications
+    - `loki_ingester_wal_disk_full_failures_total`
+    - `loki_ingester_wal_corruptions_total`
```
periklis, line 78: We need a step for the user to check that the gateway containers are ready and available, e.g. maybe opa-openshift just fails to talk to the apiserver.

operator/internal/manifests/internal/alerts/build.go
```diff
@@ -13,7 +13,7 @@ import (
 
 const (
 	// RunbookDefaultURL is the default url for the documentation of the Prometheus alerts
-	RunbookDefaultURL = "https://github.com/grafana/loki/tree/main/operator/docs/alerts.md"
+	RunbookDefaultURL = "https://github.com/grafana/loki/blob/main/operator/docs/lokistack/sop.md"
```
periklis, line 16: Good catch :rocket:

operator/internal/manifests/internal/alerts/prometheus-alerts.yaml
```diff
@@ -10,21 +10,44 @@ groups:
...
+      sum(
+        rate(
+          http_request_duration_seconds_count{group="logsv1"}[1m]
+        )
+      ) by (namespace, job, method, tenant)
+      * 100
+      > 10
+    for: 15m
+    labels:
+      severity: critical
```
periklis, line 50: I suggest to add the `handler` label to the `by` directive to indicate which path is hit (e.g. ingestion `handler="push"` or query path `handler=~"query|query_range|label|labels|label_values"`).

periklis (commented):
In general looks good to me, I would like to discussion if we want to split this into two alerts, one per path (ingestion vs. query path). This would require to use the `handler` label selector:
- ingestion `handler="push"` 
- query path `handler=~"query|query_range|label|labels|label_values"`

Red-GV:
I think splitting into two alerts is a good idea, @periklis. Both are major problems, but I would consider the errors on the write path to be more dire than on the read path.

periklis (commented):
Looks good to me, leaving the option for @xperimental to take a look before I approve

operator/docs/lokistack/sop.md
```diff
@@ -34,8 +34,8 @@ A service(s) is failing to process at least 10% of all incoming requests.
 - Console access to the cluster
 - Edit access to the deployed operator and Loki namespace:
   - OpenShift
-    - `openshift-logging`
-    - `openshift-operators-redhat`
+    - `openshift-logging` (Lokistack)
```
xperimental, line 37: I think we should choose one notation for `LokiStack` that we use throughout the docs. Currently it seems `LokiStack` is used more than `Lokistack`.

```suggestion
    - `openshift-logging` (LokiStack)
```

operator/internal/manifests/internal/alerts/prometheus-alerts.yaml
```diff
@@ -10,21 +10,67 @@ groups:
...
+    expr: |
+      sum(
+        rate(
+          http_request_duration_seconds_count{code=~"5..", group="logsv1", handler="push"}[1m]
+        )
+      ) by (namespace, job, tenant)
+      /
+      sum(
+        rate(
+          http_request_duration_seconds_count{group="logsv1", handler="push"}[1m]
```
xperimental, line 43: I think I'd rather use `http_requests_total` instead of `http_request_duration_seconds_count`. They _should_ have the same values, but the `http_requests_total` metric is the metric usually used in documentation for these "number of failed requests" counts.

operator/internal/manifests/internal/alerts/prometheus-alerts.yaml
```diff
@@ -10,21 +10,67 @@ groups:
...
+    expr: |
+      sum(
+        rate(
+          http_request_duration_seconds_count{code=~"5..", group="logsv1", handler=~"query|query_range|label|labels|label_values"}[1m]
+        )
+      ) by (namespace, job, tenant)
+      /
+      sum(
+        rate(
+          http_request_duration_seconds_count{group="logsv1", handler=~"query|query_range|label|labels|label_values"}[1m]
```
xperimental, line 66: See comment regarding `http_requests_total`.

xperimental (commented):
Looks good to me, just two "cosmetic" comments.

One idea I had was if we should add a recording rule for the two "complex" queries used to measure the failed requests, so that it can be re-used in dashboards without recalculation.

xperimental (approved):

periklis (approved):

PR #7000, opened 2022-08-29
operator: Configure default node affinity for all pods

xperimental:

Currently we only support running the operator on linux-based nodes, so it makes sense to extend this to the operand containers as well to avoid confusion in multi-os clusters.

One open question is whether this should be done always or only if the operator is running in OpenShift as the support situation might be different outside of OpenShift.

Fixes [LOG-2895](https://issues.redhat.com/browse/LOG-2895)

**Checklist**

- [x] Tests updated

### DIFF
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Main
 
+- [7000](https://github.com/grafana/loki/pull/7000) **xperimental**: Configure default node affinity for all pods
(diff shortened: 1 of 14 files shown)


xperimental:
The idea to add a feature gate for this came a bit too late. Do not merge yet :slightly_smiling_face:

Red-GV (commented):
lgtm

periklis (approved):

PR #7004, opened 2022-08-30
[TSDB] 600MB max per shard in query planning

owen-d:
This PR gives a little better understanding into shard query planning and targets a range of `(300MB,600MB]` per shard. Doubling this also allows us to halve query parallelism configs and still schedule the same amount of work in parallel, which should make for easier schema upgrades as the parallelism differences between TSDB and prior period configs will be less.

### DIFF
--- a/pkg/querier/queryrange/shard_resolver.go
+++ b/pkg/querier/queryrange/shard_resolver.go
@@ -142,14 +142,14 @@ func (r *dynamicShardResolver) Shards(e syntax.Expr) (int, error) {
 
 const (
 	// Just some observed values to get us started on better query planning.
-	p90BytesPerSecond = 300 << 20 // 300MB/s/core
+	maxBytesPerShard = 600 << 20 // 600MB/s/core
 )
 
 func guessShardFactor(stats stats.Stats) int {
-	expectedSeconds := float64(stats.Bytes) / float64(p90BytesPerSecond)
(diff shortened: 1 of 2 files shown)


pkg/querier/queryrange/shard_resolver.go
```diff
@@ -142,14 +142,20 @@ func (r *dynamicShardResolver) Shards(e syntax.Expr) (int, error) {
...
 	// Just some observed values to get us started on better query planning.
-	p90BytesPerSecond = 300 << 20 // 300MB/s/core
+	maxBytesPerShard = 600 << 20
 )
 
+// Since we shard by powers of two and we increase shard factor
+// once each shard surpasses maxBytesPerShard, if the shard factor
+// is at least two, the range of data per shard is (maxBytesPerShard/2, maxBytesPerShard]
+// For instance, for a maxBytesPerShard of 500MB and a query touching 1000MB, we split into two shards off 500MB.
+// If there are 10004MB, we split into four shards of 251MB
```
DylanGuedes, line 152:
```suggestion
// For instance, for a maxBytesPerShard of 500MB and a query touching 1000MB, we split into two shards of 500MB.
// If there are 1004MB, we split into four shards of 251MB.
```

DylanGuedes (approved):
lgtm, just left a nit but shouldn't block you

PR #7005, opened 2022-08-30
Hide duplicates caused by stream sharding

MasslessParticle:
As streams are sharded and resharded across different ingesters, the likelihood of duplicate logs being ingested increases dramatically. This removes duplicates at query time by dynamically removing the `__shard_stream__` label  for both log and sample queries so the `stream hash` used by the `MergeEntryIterator` will match when doing comparisons for duplicates.

The `stream hash` is created in either the `store` or `ingester instance` at query time so the labels are removed there. Additionally, the `MergeEntryIterator` has been modified to remove duplicates in a stream rather than just across streams. 

For completeness, `__shard_stream__` has also been removed from Label and Series Queries.


pkg/iter/entry_iterator.go
```diff
@@ -193,18 +193,6 @@ func (i *mergeEntryIterator) Next() bool {
 		return false
 	}
 
-	// shortcut for the last iterator.
```
MasslessParticle, line 196: Without tracking previous, this trick doesn't work for a single stream anymore

pkg/iter/entry_iterator.go
```diff
@@ -246,7 +233,7 @@ Outer:
 				!entry.Timestamp.Equal(i.buffer[0].Entry.Timestamp) {
 				break
 			}
-			for _, t := range previous {
+			for _, t := range i.buffer {
```
MasslessParticle, line 236: Entries may have been added to the buffer since `previous` was created so we need to iterate the actual buffer on each subsequent entry.

pkg/iter/entry_iterator.go
MasslessParticle, line 196: Without tracking previous, this trick doesn't work for a single stream anymore

pkg/storage/batch.go
```diff
@@ -568,6 +570,13 @@ func (it *sampleBatchIterator) buildIterators(chks map[model.Fingerprint][][]*La
 	return result, nil
 }
 
+func filteredLabels(ls labels.Labels) labels.Labels {
+	return labels.NewBuilder(ls).
+		Del(labels.MetricName).
+		Del(distributor.ShardLbName).
```
DylanGuedes, line 576: Question because I never used `Del` myself: if `stream_shard` isn't enabled the label won't be present anyway; will that be a problem with Del? Or it just ignores it?

pkg/storage/util_test.go
```diff
@@ -29,8 +31,8 @@ import (
 )
 
 var (
-	fooLabelsWithName = labels.Labels{{Name: "foo", Value: "bar"}, {Name: "__name__", Value: "logs"}}
-	fooLabels         = labels.Labels{{Name: "foo", Value: "bar"}}
+	fooLabelsWithNameAndShard = labels.Labels{{Name: "foo", Value: "bar"}, {Name: "__name__", Value: "logs"}, {Name: distributor.ShardLbName, Value: "3"}}
```
DylanGuedes, line 34: concern: it is great that you could cover the behavior when `__stream_shard` is present but by changing this, you actually did change all other scenarios so that we're not covering when **stream_shard isn't present**. Just a thought.

DylanGuedes (approved):
looks fantastic. No big concerns from my side so I'm approving it anyway to not stuck it, but I raised a few questions/nits.

pkg/ingester/instance_test.go
```diff
@@ -331,7 +334,7 @@ func Test_SeriesQuery(t *testing.T) {
 			},
 			[]logproto.SeriesIdentifier{
 				// Separated by shard number
-				{Labels: map[string]string{"app": "test2", "job": "varlogs"}},
+				{Labels: map[string]string{"app": "test", "job": "varlogs"}},
```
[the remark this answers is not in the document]
MasslessParticle: The new labels cause the index to put the streams in different shards so we get this one back, now

pkg/iter/entry_iterator_test.go
```diff
@@ -898,3 +923,29 @@ func TestDedupeMergeEntryIterator(t *testing.T) {
 	require.Equal(t, "3", it.Entry().Line)
 	require.Equal(t, time.Unix(2, 0), it.Entry().Timestamp)
 }
+
+func TestSingleStreamDedupeMergeEntryIterator(t *testing.T) {
+	it := NewMergeEntryIterator(context.Background(),
+		[]EntryIterator{
+			NewStreamIterator(logproto.Stream{
+				Labels: ``,
```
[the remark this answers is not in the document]
MasslessParticle: As long as the `LabelHash` matches, it doesn't matter what the labels are. That's why the labels needed to be filtered when the iterators are first created.

pkg/storage/batch.go
MasslessParticle: Del doesn't do anything if the label doesn't exist.

pkg/storage/util_test.go
MasslessParticle: My thought was that `__stream_shard__` should be removed the same way `__name__` is because it's not ever relevant to a user. I'm not sure, though and happy to hear from anyone else.

MasslessParticle:
This change had a minimal affect on benchmarks:
```
benchmark                               old ns/op     new ns/op     delta
BenchmarkSortIterator/merge_sort-8      4186835       4243637       +1.36%
BenchmarkSortIterator/sort-8            1338503       1330357       -0.61%
BenchmarkSortSampleIterator/merge-8     1655672       1619700       -2.17%
BenchmarkSortSampleIterator/sort-8      287801        288213        +0.14%

benchmark                               old allocs     new allocs     delta
BenchmarkSortIterator/merge_sort-8      6              6              +0.00%
BenchmarkSortIterator/sort-8            104            104            +0.00%
BenchmarkSortSampleIterator/merge-8     6              6              +0.00%
BenchmarkSortSampleIterator/sort-8      3              3              +0.00%

benchmark                               old bytes     new bytes     delta
BenchmarkSortIterator/merge_sort-8      11104         11104         +0.00%
BenchmarkSortIterator/sort-8            318808        318809        +0.00%
BenchmarkSortSampleIterator/merge-8     9416          9416          +0.00%
BenchmarkSortSampleIterator/sort-8      1928          1928          +0.00%
```

pkg/storage/util_test.go
DylanGuedes: hmmm fair point, `__stream_shard` behavior is definitely very similar to `__name__`. That said, I'm approving it.

pkg/ingester/ingester.go
```diff
@@ -700,10 +700,12 @@ func (i *Ingester) QuerySample(req *logproto.SampleQueryRequest, queryServer log
 	if err != nil {
 		return err
 	}
+	var iters []iter.SampleIterator
```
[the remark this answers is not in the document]
MasslessParticle: Results from these iterators are sent to the querier in batches [here](https://github.com/grafana/loki/blob/c23e541fe7d6aa4e35e720df4e191f034325d24f/pkg/ingester/instance.go#L873). Those batches may end up in different instances of iterators in the querier and result in duplicates even when though we use a `MergeSampleIterator` [here](https://github.com/grafana/loki/blob/c23e541fe7d6aa4e35e720df4e191f034325d24f/pkg/querier/querier.go#L214). 

This change ensures that an instance always uses a `MergeSampleIterator` to avoid sending duplicates in a single stream to the querier.

pkg/ingester/instance.go
```diff
@@ -721,6 +729,10 @@ outer:
 		if chunkFilter != nil && chunkFilter.ShouldFilter(stream.labels) {
 			continue
 		}
+
+		// To Enable downstream deduplication of sharded streams, remove the shard label
+		stream.labels = labels.NewBuilder(stream.labels).Del(distributor.ShardLbName).Labels()
```
owen-d, line 734: I'm worried that we're doing this on an ancillary function right here. This is mutating the underlying stream, which is confusing (it only mutates the stream the first time it's passed over by this function) and likely causes bugs with WAL checkpoints.

I don't think we should mutate the labels for a stream, but instead just filter it out when serving queries.

owen-d (commented):
I'll try to finish giving this a review, but left some intermediate feedback.

pkg/querier/querier.go
```diff
@@ -161,9 +161,7 @@ func (q *SingleTenantQuerier) SelectLogs(ctx context.Context, params logql.Selec
 
 		iters = append(iters, storeIter)
 	}
-	if len(iters) == 1 {
```
owen-d, line 164: Why is this no longer valid?

pkg/iter/entry_iterator.go
owen-d: Can you elaborate a bit?

pkg/querier/querier.go
MasslessParticle: `MergedEntryIter` is the only way we dedupe entries. Now that duplicates are being considered within a single iterator, it needs to be enforced that the returned iter is a `MergedEntryIterator`

To illustrate: [this test](https://github.com/grafana/loki/blob/25cc9402b7a076be2d3d33a9277233f233d376b0/pkg/querier/querier_test.go#L1115-L1152) will fail when that code is left in place

pkg/iter/entry_iterator.go
MasslessParticle: Sure! In the case there is only one iterator, this shortcut doesn't store previous entries or check current entries against previous entries in the buffer so it allows any duplicates through. Making the code do that starts to look a lot like what's going on below.

pkg/querier/querier.go
owen-d: I think we should keep this code as is and instead the iterators coming from the ingester|store themselves should require that they dedupe their own contents. The test you mentioned is introduced in this PR and would fail because it's introducing duplicates coming from a single store iterator. Instead, I believe it should test deduping across iterators.

pkg/iter/entry_iterator.go
owen-d: I think this is only a problem if there are duplicates within a single iterator. If we avoid that happening, these optimizations still work.

owen-d (commented):
I think it would make more sense to standardize on the following behavior:

* Single iterators return deduped contents
* Merge iterators dedupe across iterators

Can we instead have the `ForStream` implementations strip out unwanted labelsets?
```
// Pipeline can create pipelines for each log stream.
type Pipeline interface {
	ForStream(labels labels.Labels) StreamPipeline
}
```

It's possible our dedupe code should _just work_ if we make that change

PR #7007, opened 2022-08-31
[new feature] syntax: support "vector" grama

### LINKED ISSUE #6946: Implement vector() function
I want to turn null (no data) to zero in query result in easy and reliable way.

There are other Loki users who want the same.
https://community.grafana.com/t/how-can-i-turn-no-data-to-zero-in-loki/40694

Implement a function like [vector()](https://prometheus.io/docs/prometheus/latest/querying/functions/#vector) in PromQL.
> vector(s scalar) returns the scalar s as a vector with no labels.
Then users can join their query with vector() to turn NULL to zero, as below.
```
vector1
or
vector(0)
```

(1) Grafana's transform feature.
However, this feature only works for a dashboard panel with single Loki query. It doesn't work for multiple queries for a panel. 

(2) Tried to generate zeros reliably, but not perfect
```
sum(count_over_time({namespace="traefik"}|="foo"[1m])) 
or
sum(count_over_time({namespace="traefik"}[1m])) < bool 0
```
The 2nd vector can still return NULL if there is no log data matching labels sets. Not useful for log streams which has fluctuating log ingestion. 

(3) Tried absent_over_time() as 2nd vector instead, it didn't work for me.
- Bug report: [Using absent_over_time with logical operator returns "Unimplemented" error. #6943](https://github.com/grafana/loki/issues/6943)

(4) First, stored the log-metrics in Prometheus. Second, use vector() in Prometheus. 
This works of course. But I wish I can do the same using LokiQL which is upstream.

**Additional context**

liguozhong:
Implement vector() function.

The "vector function" feature is divided into multiple PRs to avoid excessively large PRs resulting in too long review time.

Fixes #6946

The test for syntax has passed
image

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


pkg/logql/syntax/ast.go
```diff
@@ -1562,3 +1565,38 @@ func MatcherGroups(expr Expr) []MatcherRange {
...
+	Val float64
+	implicit
+}
+
+func NewVectorExpr(scalar string) *VectorExpr {
+	n, err := strconv.ParseFloat(scalar, 64)
+	if err != nil {
+		panic(logqlmodel.NewParseError(fmt.Sprintf("unable to parse vectorExpr as a float: %s", err.Error()), 0, 0))
+	}
+	//fmt.Println("vector new scalar:", scalar)
```
dannykopping, line 1579: Remove this please

pkg/logql/syntax/ast.go
```diff
@@ -1562,3 +1565,38 @@ func MatcherGroups(expr Expr) []MatcherRange {
...
+
+type VectorExpr struct {
+	Val float64
+	implicit
+}
+
+func NewVectorExpr(scalar string) *VectorExpr {
+	n, err := strconv.ParseFloat(scalar, 64)
+	if err != nil {
+		panic(logqlmodel.NewParseError(fmt.Sprintf("unable to parse vectorExpr as a float: %s", err.Error()), 0, 0))
```
dannykopping, line 1577: Do we have to panic here?

dannykopping (commented):
I'm excited about this one, thanks @liguozhong :+1:

pkg/logql/syntax/ast.go
liguozhong: done，thanks

pkg/logql/syntax/ast.go
liguozhong: Other Exprs are panic too.

image

dannykopping: Hhhmm, that's not great.
I think that's a problem, and we should fix it, so let's not make the problem worse here IMHO.
I think we should remove the panic. Maybe we can find a way to make this fail gracefully without panicking the whole querier? Perhaps setting the vector value as some crazy number like `MaxFloat64` if the parsing fails, and if this value is present when querying, return an error that bubbles up to the user to give a helpful error message.

liguozhong: https://pkg.go.dev/bufio#Scanner.Scan

Does VectorExpr refer to Scanner to provide an Err() function?

image

liguozhong: Probably changed to look like this.
image

jeschkies:
Merging recent main should fix the failing `helm-ci` job.

liguozhong:
> Merging recent main should fix the failing `helm-ci` job.

done，thanks

dannykopping (approved):
LGTM, thanks @liguozhong.
I'll keep an eye out for the next PR which will make this work with the query engine.

liguozhong:
> LGTM, thanks @liguozhong. I'll keep an eye out for the next PR which will make this work with the query engine.

thanks，vector(0) is really needed in the oncall alarm .

PR #7010, opened 2022-08-31
Abide to Mixin linting rules.

jeschkies:
Cloud integrations need to pass linting rules such as proper dashboard names. It should be `Data Source` not `datasource`. See also grafana/cloud-onboarding#2228


dannykopping (commented):
Please add a description - some context here would be helpful

cstyan:
Have we tried this out? I'm not sure if this label field is used in the copying of template variables when using the `Loki Dashboards` to navigate to other dashboards. If it is, this might break that functionality.

jeschkies:
@cstyan great catch. I did not try the mixin.

jeschkies:
@cstyan just found that the rule dashboard does not even have the navigation at the top right.

cstyan:
I just checked, at least for `Loki / Recording Rules` the navigation works, are you saying it doesn't work for the compiled dashboards?

In any case we're only changing the label here, not the name of the template variable, so we should be good to move forward with this.

jeschkies:
@cstyan if tested the navigation it works. There was none for `Recording Rules` before.

production/loki-mixin/dashboards/recording-rules.libsonnet
```diff
@@ -4,6 +4,8 @@ local utils = import 'mixin-utils/utils.libsonnet';
 (import 'dashboard-utils.libsonnet') {
   grafanaDashboards+:
     {
-      'loki-mixin-recording-rules.json': raw,
+      'loki-mixin-recording-rules.json': raw +
+                      $.dashboard('Loki / Recording Rules', uid='recording-rules')
```
jeschkies, line 8: Technically this is a breaking change because we rename the dashboard.

cstyan (approved):

PR #7012, opened 2022-08-31
Generate Helm chart README.

jeschkies:
The Helm CI would not pass the linting stage.

Summary
* updates the README
* introduces a filter that will only run the Helm CI on changes to the
  Helm chart directory
* removes trailing whitespace
* corrects the Minio chart repository


trevorwhitney (approved):
LGTM!

PR #7013, opened 2022-08-31
[bugfix] fix query_frontend init fail error

liguozhong:
My loki cluster is distributed without a compactor, but compactorAddress must require me to fill in a compactor url.
Using the latest version of the master branch, the query frontend of loki cluster cannot be started and will encounter this error log.
```go
query filtering for deletes requires 'compactor_address' to be configured
```

Fixes #<issue number>

my loki cluster Topology.
image

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


jeschkies:
Please merge main to fix the failing `helm-ci` job.

liguozhong:
> Please merge main to fix the failing `helm-ci` job.

done，thanks

ashwanthgoli:
Returning the address of localhost is only valid for single binary or ssd deployments. For multi-store index (as it does not support delete requests), query-frontend need not make calls to compactor. So a better fix for this issue would be to pass down a nil `cacheGenNumLoader` to `NewTripperware` func for indexes that don't support deletions
https://github.com/grafana/loki/blob/e39ea21e3cacda92661d1d77770c7d1486ec4124/pkg/querier/queryrange/roundtrip.go#L41-L46

It would be nice if you could also refactor `cacheGenClient()` to return `*GenNumberLoader` instead. It can include the logic to return nil based on the index type (for TSDB and multi-store)
https://github.com/grafana/loki/blob/e39ea21e3cacda92661d1d77770c7d1486ec4124/pkg/loki/modules.go#L1067-L1074

Thanks for taking time out and submitting the PR :)

liguozhong:
> Returning the address of localhost is only valid for single binary or ssd deployments. For multi-store index (as it does not support delete requests), query-frontend need not make calls to compactor. So a better fix for this issue would be to pass down a nil `cacheGenNumLoader` to `NewTripperware` func for indexes that don't support deletions
> 
> https://github.com/grafana/loki/blob/e39ea21e3cacda92661d1d77770c7d1486ec4124/pkg/querier/queryrange/roundtrip.go#L41-L46
> 
> It would be nice if you could also refactor `cacheGenClient()` to return `*GenNumberLoader` instead. It can include the logic to return nil based on the index type (for TSDB and multi-store)
> 
> https://github.com/grafana/loki/blob/e39ea21e3cacda92661d1d77770c7d1486ec4124/pkg/loki/modules.go#L1067-L1074
> 
> Thanks for taking time out and submitting the PR :)

hi, thank you for such a detailed review suggestion.
I resubmitted the code once as you suggested, if I understand correctly

pkg/loki/modules.go
```diff
@@ -654,13 +653,26 @@ func (t *Loki) initQueryFrontendTripperware() (_ services.Service, err error) {
 	return services.NewIdleService(nil, nil), nil
 }
 
-func (t *Loki) cacheGenClient() (generationnumber.CacheGenClient, error) {
+func (t *Loki) cacheGenClient() (*generationnumber.GenNumberLoader, error) {
+	if config.UsingTSDB(t.Cfg.SchemaConfig.Configs) {
+		return nil, nil
```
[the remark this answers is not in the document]
liguozhong: hi , according to your suggestion, I submitted another commit, canceling the code snippets such as ' return nil,nil ' Fragments like this are really bad. 
I operator the loki cluster full-time, this is my job, your suggestions are very helpful to me, thank you for your patience to review and guide

pkg/loki/modules.go
```diff
@@ -654,13 +653,26 @@ func (t *Loki) initQueryFrontendTripperware() (_ services.Service, err error) {
 	return services.NewIdleService(nil, nil), nil
 }
 
-func (t *Loki) cacheGenClient() (generationnumber.CacheGenClient, error) {
+func (t *Loki) cacheGenClient() (*generationnumber.GenNumberLoader, error) {
```
MasslessParticle: This should probably just return the interface type `queryrangebase.CacheGenNumberLoader`

MasslessParticle: Let's also rename the function to `cacheGenNumberLoader`

MasslessParticle (approved):

PR #7017, opened 2022-08-31
Loki: Return all rate-limited streams to distributors

DylanGuedes:
Modify ingesters to return all rate-limited streams to distributors instead of only the last one seen. The list of streams is injected in the error details, supported by our gRPC client. The error message text is the same so this shouldn't break anything relying on the existing error messages.
This is necessary for the new stream sharding work.


pkg/ingester/instance.go
```diff
@@ -218,27 +233,40 @@ func (i *instance) Push(ctx context.Context, req *logproto.PushRequest) error {
...
+	if len(failedEntriesWithError) == 0 {
+		return 0, nil
+	}
+
 	lastEntryWithErr := failedEntriesWithError[len(failedEntriesWithError)-1]
 	_, ok := lastEntryWithErr.e.(*validation.ErrStreamRateLimit)
 	outOfOrder := chunkenc.IsOutOfOrderErr(lastEntryWithErr.e)
 	if !outOfOrder && !ok {
-		return lastEntryWithErr.e
+		return 0, lastEntryWithErr.e
```
MasslessParticle, line 266: If we're using gRPC codes, this means `OK`. Looking at the `httpgrpc.Errorf` and `httpgrpc.HTTPResponseFromError` code, they aren't doing any special code conversions so we can just stick with http codes and use `http.BadRequest`

MasslessParticle (changes requested):
Nice work so far!

I think we should introduce another type to collect the errors as we process the streams and collect a list of those then figure out the error based on the list like this:

1. If all errors in the list are neither 400 or 429, return 500 with the message of the last error (the current behavior)
2. If any of the errors are 429, return 429 with a list of rate-limited streams
3. Otherwise return 400 with a list of bad-request streams and the error message of the last error (preserves the current behavior, but includes bad-request streams)

The error types could look something like 
```go
type pushErr struct {
    code int32
    err error
    details details
}
```

This approach means that we can deal with errors similarly to what we do now in the loop and we can figure everything out at the end.

pkg/ingester/instance.go
DylanGuedes: Oh shit, good catch. `0` here is definitely a mistake: I think I was previously returning as the first argument the number of entries that were ingested or something, and probably overlooked this part of the code :smile:

DylanGuedes:
> Nice work so far!
> 
> I think we should introduce another type to collect the errors as we process the streams and collect a list of those then figure out the error based on the list like this:
> 
> 1. If all errors in the list are neither 400 or 429, return 500 with the message of the last error (the current behavior)
> 2. If any of the errors are 429, return 429 with a list of rate-limited streams
> 3. Otherwise return 400 with a list of bad-request streams and the error message of the last error (preserves the current behavior, but includes bad-request streams)
> 
> The error types could look something like
> 
> ```go
> type pushErr struct {
>     code int32
>     err error
>     details details
> }
> ```
> 
> This approach means that we can deal with errors similarly to what we do now in the loop and we can figure everything out at the end.

Agreed. I've pushed a commit changing it to be similar to what you suggested, but as it is not 100% equal lmk what you think of it. There are a few things for me to improve (ex: pipeline will probably break, the proto names etc), but let me know first if you think this is going in a good direction before I put effort into making it pretty.

DylanGuedes:
Closing this as we'll designed a better approach to the stream sharding work.

PR #7020, opened 2022-08-31
Add nix configuration

trevorwhitney:

This PR adds nix configuration to the Loki repo to make it easier for people who use nix to integrate building Loki into their existing workflows. For example, I have a `direnv` hook that uses nix shell to make sure I have the correct version of all dependent binaries (ie. `go`, `golangci-lint`, etc) on my `$PATH` whenever I change into the Loki directory. I find this especially useful for building with `CGO` enabled as it requires a C compiler with systemd headers to be available on your path.

Furthermore, by having a `flake.nix` in the root of this repo, it will make it easier for users of the nix package manager to grab a specific version of Loki for their own systems / projects.

This PR does _not_ attempt to replace any existing build processes, this is merely supplemental. All these nix configurations wrap existing `Makefile` targets, and the addition of these files will have no effect on anyone not using nix.

Please see the README.md included in this PR for more information.

**Checklist**
- [X] Documentation added


nix/build-vars.nix
```diff
@@ -0,0 +1,10 @@
+# DO NOT EDIT, this file is auto-generated by
+# generate-build-vars.sh for purposes of getting around
+# the dependency our build has on the .git folder, which is
+# not copied to the nix store when building. See the README for
+# more details.
+{
+  imageTag = "main-5a6d09f";
+  gitRevision = "5a6d09f83";
+  gitBranch = "main";
```
trevorwhitney, line 9: in general there probably isn't much need to commit changes to this file, so I decided to have it track `main`. The `gitRevision` is only used on dirty git trees, and there's a shell hook that will automatically update these values for your local branch when you run `nix develop`

owen-d (approved):

nix/build-vars.nix
simonswine: What is the reasoning to have it in git?

I suspect to still be able to build the derivation when the flake is imported. I would prefer to have a `unknown` revision instead. As otherwise it is just implying the wrong thing.

nix/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+To build the repo (including running tests), from the root of the repo you can run `nix build`.
+
+## Installing Nix
+
+Nix is supported on Linux, MacOS, and Windows (WSL2). Check [here](https://nixos.org/download.html#download-nix) for installation instructions for your specific platform.
+
+You will also need to enable the Flakes feature to use Nix with this repo. See this [wiki](https://nixos.wiki/wiki/Flakes) for instructions on enabling Flakes.
+
+## Dealing with .git
```
simonswine, line 32: Have you tried the approach lined out in: https://stackoverflow.com/questions/67025103/how-inject-git-revision-into-nix-derivation

flake.nix
```diff
@@ -0,0 +1,70 @@
+{
+  description = "Grafana Enterprise Logs";
```
simonswine, line 2:
```suggestion
  description = "Grafana Loki";
```

simonswine (approved):
I welcome this initiative. It is quite useful to have this as standardized environments and maybe in the future we could expand this to also form the basis of our CI/release (if helpful)

I think before we merge we should have some CI around that at least enforced the formatting of nix files and run a build.

For some reason in my local environment (on nixos) I have trouble building the loki derivation and I need that patch:

```
failed to initialize build cache at /homeless-shelter/.cache/go-build: mkdir /homeless-shelter: permission denied
[...]
failed to initialize build cache at /homeless-shelter/.cache/golangci-lint: mkdir /homeless-shelter: permission denied
```

```diff
--- a/nix/loki.nix
+++ b/nix/loki.nix
@@ -30,11 +30,13 @@ pkgs.stdenv.mkDerivation {
   '';

   buildPhase = ''
+    export GOCACHE=$TMPDIR/go-cache
     make clean loki logcli loki-canary promtail
   '';

   doCheck = true;
   checkPhase = ''
+    export GOCACHE=$TMPDIR/go-cache
     make lint test
   '';
```

nix/README.md
trevorwhitney: I think I am using the approach lined out in that link, which if I read correctly is to use `self.rev` for flakes? Maybe I'm misunderstanding if there's another workaround in that article?

trevorwhitney: Ahh, I see, the suggestion is to use `(builtins.fetchGit args).rev`. That will not work, as the url here would be `./..` or `self`, which in either case points to the source in the nix store which has already been stripped of git info, so `fetchGit` will complain that this is not a git repo.

trevorwhitney:
> What is the reasoning to have it in git?
>
> I suspect to still be able to build the derivation when the flake is imported. I would prefer to have a unknown revision > > instead. As otherwise it is just implying the wrong thing.

So, if you import the flake it will not use the value of `gitRevision` from this file, it will instead use `self.rev`, which will be correct. However, maybe we could put only `gitBranch` in this file, and use either `self.rev` or `unknown` or `dirty` as the sha? That would make it so we should never have to commit changes to this file.

trevorwhitney:
Build is currently broken, waiting on: https://github.com/grafana/loki/pull/7029

PR #7036, opened 2022-09-02
tsdb: build tsdb from tenantheads

ashwanthgoli:
refactors tsdb manager code to expose a new method `BuildFromHead` which comes in handy for building tsdb from tenant heads after rotation.

Currently we build it by recovering head from WAL even when we have access to the in-mem tenant heads.

using `loki_tsdb_head_rotations_*` metric to track loop failures which could be caused either by rotation failures or if rotation is blocked by prev head failing to build

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/tsdb/head_manager.go
+++ b/pkg/storage/stores/tsdb/head_manager.go
@@ -152,11 +152,6 @@ func (m *HeadManager) loop() {
 		}
 
 		if err := m.buildTSDBFromHead(m.prevHeads); err != nil {
-			level.Error(m.log).Log(
-				"msg", "failed building tsdb head",
-				"period", m.period.PeriodFor(m.prev.initialized),
-				"err", err,
-			)
 			return err
 		}
 
@@ -175,16 +170,24 @@ func (m *HeadManager) loop() {
 	for {
 		select {
 		case <-ticker.C:
+			m.metrics.tsdbHeadRotationsTotal.Inc()
+
 			// retry tsdb build failures from previous run
 			if err := buildPrev(); err != nil {
+				m.metrics.tsdbHeadRotationsFailedTotal.Inc()
+				level.Error(m.log).Log(
+					"msg", "failed building tsdb head",
+					"period", m.period.PeriodFor(m.prev.initialized),
+					"err", err,
+				)
 				// rotating head without building prev would result in loss of index for that period (until restart)
 				continue
 			}
 
 			now := time.Now()
-			curPeriod := m.period.PeriodFor(m.activeHeads.start)
-			if m.period.PeriodFor(now) > curPeriod {
+			if curPeriod := m.period.PeriodFor(m.activeHeads.start); m.period.PeriodFor(now) > curPeriod {
 				if err := m.Rotate(now); err != nil {
+					m.metrics.tsdbHeadRotationsFailedTotal.Inc()
 					level.Error(m.log).Log(
 						"msg", "failed rotating tsdb head",
 						"period", curPeriod,
@@ -195,7 +198,13 @@ func (m *HeadManager) loop() {
 			}
 
 			// build tsdb from rotated-out period
-			buildPrev()
+			if err := buildPrev(); err != nil {
+				level.Error(m.log).Log(
+					"msg", "failed building tsdb head",
+					"period", m.period.PeriodFor(m.prev.initialized),
+					"err", err,
+				)
+			}
 		case <-m.cancel:
 			return
 		}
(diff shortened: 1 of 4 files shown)


pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -151,8 +151,8 @@ func (m *HeadManager) loop() {
 			return nil
 		}
 
-		if err := m.buildTSDBFromWAL(m.prev.initialized); err != nil {
-			return errors.Wrap(err, "building tsdb head")
+		if err := m.buildTSDBFromHead(m.prevHeads); err != nil {
+			return err
```
sandeepsukhani, line 155: Let us retain the error wrapping since it helps add more context and debug the issue easily.
```suggestion
			return errors.Wrap(err, "building tsdb from head")
```

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -170,8 +170,11 @@ func (m *HeadManager) loop() {
 	for {
 		select {
 		case <-ticker.C:
+			m.metrics.tsdbHeadRotationsTotal.Inc()
```
sandeepsukhani, line 173: This one is always incremented even if it fails below. I think it should only be incremented when we succeed.

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -270,7 +286,9 @@ func (m *HeadManager) Start() error {
 		return errors.Wrap(err, "building tsdb")
 	}
 
+	m.metrics.tsdbWALTruncationsTotal.Inc()
```
sandeepsukhani, line 289: same here, increment only if we succeed.

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -349,27 +360,40 @@ func (m *HeadManager) Rotate(t time.Time) (err error) {
...
+			"period", period,
+			"err", err,
+		)
+	}
+
+	return nil
+}
+
+func (m *HeadManager) truncateWALForPeriod(period int) (err error) {
+	m.metrics.tsdbWALTruncationsTotal.Inc()
```
sandeepsukhani, line 382: same

pkg/storage/stores/tsdb/manager.go
```diff
@@ -156,105 +158,123 @@ func (m *tsdbManager) Start() (err error) {
...
 
-func (m *tsdbManager) BuildFromWALs(t time.Time, ids []WALIdentifier) (err error) {
-	level.Debug(m.log).Log("msg", "building WALs", "n", len(ids), "ts", t)
-	// get relevant wals
-	// iterate them, build tsdb in scratch dir
-	defer func() {
-		m.metrics.tsdbCreationsTotal.Inc()
-		if err != nil {
-			m.metrics.tsdbCreationFailures.Inc()
+func (m *tsdbManager) build(heads *tenantHeads) (err error) {
```
sandeepsukhani, line 161: Just `build` does not give much context.
```suggestion
func (m *tsdbManager) buildFromHead(heads *tenantHeads) (err error) {
```

pkg/storage/stores/tsdb/manager.go
```diff
@@ -156,105 +158,123 @@ func (m *tsdbManager) Start() (err error) {
...
-				return err
-			}
+	return m.build(heads)
+}
 
-			level.Debug(m.log).Log("msg", "finished building tsdb for period", "pd", p, "dst", dst.Path(), "duration", time.Since(start))
+func (m *tsdbManager) BuildFromWALs(t time.Time, ids []WALIdentifier) (err error) {
+	level.Debug(m.log).Log("msg", "building WALs", "n", len(ids), "ts", t)
+	defer func() {
+		m.metrics.tsdbCreationsTotal.WithLabelValues("wal").Inc()
```
sandeepsukhani, line 262: same feedback here regarding incrementing it only during a success.

sandeepsukhani (commented):
Overall the code looks good! I have left just minor feedback. You can also do a follow-up PR to take care of feedback on metrics at once with the other changes we discussed.

pkg/storage/stores/tsdb/head_manager.go
ashwanthgoli: yeah, currently all these are also tracking attempts. I can move these around in the refactoring work that we discussed

sandeepsukhani (approved):
LGTM! As discussed, let us do a follow-up PR refactoring the metrics.

PR #7040, opened 2022-09-02
Remove duplicated `loki_boltdb_shipper` prefix from `tables_upload_operation_total` metric

bakunowski:
This PR removes the duplicated `loki_boltdb_shipper` prefix from the `tables_upload_operation_total` metric. 

In https://github.com/grafana/loki/pull/6278 metrics for tracking `indexshipper` operations were added. `WrapRegistererWithPrefix` was used to make sure the same metrics as before were exposed (see [here](https://github.com/grafana/loki/blob/e39ea21e3cacda92661d1d77770c7d1486ec4124/pkg/storage/stores/shipper/shipper_index_client.go#L88)).

While the `downloads` metrics had the `Namespace` field removed (see [here](https://github.com/grafana/loki/blob/e39ea21e3cacda92661d1d77770c7d1486ec4124/pkg/storage/stores/indexshipper/downloads/metrics.go#L22)), the `uploads` metric did not (see [here](https://github.com/grafana/loki/blob/e39ea21e3cacda92661d1d77770c7d1486ec4124/pkg/storage/stores/indexshipper/uploads/metrics.go#L20)). This results in the `tables_upload_operation_total` being exposed as `loki_boltdb_shipper_loki_boltdb_shipper_tables_upload_operation_total` instead of `loki_boltdb_shipper_tables_upload_operation_total`.

This PR serves both as an issue and a fix.

This is my best interpretation why the exposed name of the `uploads` metric has changed. If this is incorrect, or the change was intentional - I'm sorry for the noise.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


kavirajk (commented):
LGTM. Thanks for the PR @bakunowski. Sounds totally reasonable. Also double checked with metrics coming from internal Loki clusters.

Since this is the user visible change, can you please add it in CHANGELOG as well please? I will merge after that.

bakunowski:
Thanks for the review @kavirajk, glad to hear I wasn't completely off the mark. Updated CHANGELOG in the `Fixes` section.

sandeepsukhani (approved):
LGTM! Thanks for fixing it!

PR #7049, opened 2022-09-05
Helm: Fix Prometheus rule LokiTooManyCompactorsRunning

calvinbui:
Signed-off-by: Calvin Bui <3604363+calvinbui@users.noreply.github.com>

The Helm chart deploys two services, `loki` and `loki-headless`. 

The `loki_boltdb_shipper_compactor_running` metric will show all loki pods attached to both services.

The Prometheus rule in the chart, `LokiTooManyCompactorsRunning`,  does an aggregation over an entire namespace, which will always double the amount of compactors it detects.

image

Fixes N/A

**Checklist**
- [x] Documentation added
- [ ] Tests updated


calvinbui:
CLA signed and chart version bumped

production/helm/loki/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
+## 3.0.1
+
+- [BUGFIX] Fix Prometheus rule `LokiTooManyCompactorsRunning` aggregating multiple services together.
```
kavirajk, line 16:
```suggestion
- [BUGFIX] Fix Prometheus rule `LokiTooManyCompactorsRunning` aggregating multiple services together in Loki helm chart.
```

Just to be more clear about, this is not changing the alerts for other environments like `tanka` and only for helm charts where we have `loki-headless`.

kavirajk (commented):
LGTM. Thanks @calvinbui for the PR. Sounds reasonable to me. Leave it to @trevorwhitney  for final approval.

Added one minor suggestion to the changelog entry.

calvinbui:
On second thought, I'm not sure if this is the best approach as they'll be two alerts if it does go >1

trevorwhitney (changes requested):
> On second thought, I'm not sure if this is the best approach as there'll be two alerts if it does go >1

Agreed. I think instead we should only scrape 1 of those services. My inclination would be to drop `loki-headless`, WDYT?

calvinbui:
> > On second thought, I'm not sure if this is the best approach as there'll be two alerts if it does go >1
> 
> Agreed. I think instead we should only scrape 1 of those services. My inclination would be to drop `loki-headless`, WDYT?

what about the simple-scalable services? they use -read and -write suffixed services

varac:
Hej, looking forward for this fix, what's the current state ?

trevorwhitney:
> what about the simple-scalable services? they use -read and -write suffixed services

yeah, good point, looks like maybe we want to always deploy the `loki` service (for both single binary and scalable topologies), and then we can just scrape this service and drop metrics from headless. WDYT?

zarikx:
If I'm not mistaken, `loki-headless` just needs `prometheus.io/service-monitor: "false"` label (similar to `service-read-headless` and `service-write-headless` services) to prevent discovering single-binary StatefulSet twice.

calvinbui:
> If I'm not mistaken, `loki-headless` just needs `prometheus.io/service-monitor: "false"` label (similar to `service-read-headless` and `service-write-headless` services) to prevent discovering single-binary StatefulSet twice.

agreed, the best solution. 

PR updated

trevorwhitney (approved):
lgtm! thanks!

trevorwhitney:
@calvinbui looks like this just needs a `helm-docs` update and it's good to go.

trevorwhitney:
@calvinbui looks like a markdown linting error?

trevorwhitney:
@calvinbui looks like all we need is one more rebase, a lot moving on the helm chart these past few days :smile: Thanks!

calvinbui:
😦😩

PR #7055, opened 2022-09-05
[Helm] Make compactor and limits_config configurable

D3luxee:
This PR adds the option to configure the limits_config and compactor in order to apply retentions based on the compactor.
Before the helm charts were migrated from the grafana/helm-charts repo there were some open issues about this for the loki-simple-scalable chart. (https://github.com/grafana/helm-charts/issues/1693)

For us the configuration of retentions via compactor is a important feature and i think the changes i applied here are minor.

Fixes partially #7068

### DIFF
--- a/production/helm/loki/values.yaml
+++ b/production/helm/loki/values.yaml
@@ -65,29 +65,28 @@ loki:
     {{- else }}
     auth_enabled: {{ .Values.loki.auth_enabled }}
     {{- end }}
-
     server:
       http_listen_port: 3100
       grpc_listen_port: 9095
-
     memberlist:
       join_members:
         - {{ include "loki.memberlist" . }}
-
     {{- if .Values.loki.commonConfig}}
     common:
     {{- toYaml .Values.loki.commonConfig | nindent 2}}
       storage:
       {{- include "loki.commonStorageConfig" . | nindent 4}}
     {{- end}}
-
     limits_config:
+    {{- if .Values.loki.limitsConfig}}
+    {{- toYaml .Values.loki.limitsConfig | nindent 2}}
+    {{- else }}
       enforce_metric_name: false
       reject_old_samples: true
       reject_old_samples_max_age: 168h
       max_cache_freshness_per_query: 10m
       split_queries_by_interval: 15m
-
+    {{- end}}
     {{- with .Values.loki.memcached.chunk_cache }}
     {{- if and .enabled .host }}
     chunk_store_config:
@@ -100,7 +99,6 @@ loki:
           service: {{ .service }}
     {{- end }}
     {{- end }}
-
     {{- if .Values.loki.schemaConfig}}
     schema_config:
     {{- toYaml .Values.loki.schemaConfig | nindent 2}}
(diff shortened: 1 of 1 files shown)


production/helm/loki/values.yaml
```diff
@@ -201,6 +209,12 @@ loki:
 
   # -- Check https://grafana.com/docs/loki/latest/configuration/#schema_config for more info on how to configure schemas
   schemaConfig: {}
+  
+  # -- Check https://grafana.com/docs/loki/latest/configuration/#limits_config for more info on how to configure limits
+  limitsConfig: {}
+  
+  # -- Check https://grafana.com/docs/loki/latest/configuration/#compactor for more info on how to configure compactor
+  compactorConfig: {}
```
jeschkies, line 217: @trevorwhitney do we want to move all configuration into values or rather move all but the most important into `config`?

trevorwhitney:
@D3luxee what are the changes you need to make to the compactor config, and why do you need these values instead of the defaults? 

Loki is hard to configure, and we're hoping to use the helm values as an abstraction layer that reduces the configuration surface area. What I'd ideally like is for the Loki team to classify basic/advanced configs, and only expose basic configs via value structs in the helm chart, and require setting the full config string whenever doing something advanced.

I agree with exposing the `limits_config`, but I think it may be getting into a slippery slope to expose every possible section via a value struct. However, I could be persuaded if there's good reasons why those sections would often need to be tweaked beyond the defaults.

D3luxee:
Hi,
i want to be able to configure a storage retention using the compactor as described here: https://grafana.com/docs/loki/latest/operations/storage/retention/#retention-configuration

This requires changing the compactor config but also adding a limits config.
The limits config needs to be even more configurable because it can contain retentions per stream.

We use the compactor instead of the table manager because it allows more granular configuration and the docs state:
```
The Compactor retention will become the default and have long term support.
```

towolf:
@trevorwhitney 

> Loki is hard to configure, and we're hoping to use the helm values as an abstraction layer that reduces the configuration surface area. What I'd ideally like is for the Loki team to classify basic/advanced configs, and only expose basic configs via value structs in the helm chart, and require setting the full config string whenever doing something advanced

I think this approach is tricky because you need to cover a lot of ground with the merged Helm chart, while support all the various config knobs with your configuration abstraction.

We had to almost immediately override the config text template, because the necessary knobs were not exposed. We want to persist to Google GCS but keep the Loki Ruler rules in a configmap (supported in the old `loki` chart but not present in loki-simple-scalable.

Working with the templated string is pretty awkward compared to a native YAML map, which translates to loki config.yaml directly. This can be overridden in a straight-forward manner.

Now the general documentation found for Loki is not always valid for the Helm chart because of the indirection.

jeschkies:
> I think this approach is tricky because you need to cover a lot of ground with the merged Helm chart, while supporting all the various config knobs with your configuration abstraction.

Yes. But users must be able to define the bare minimum wihtout setting everything else to `null`. If I'm not mistaken it breaks for nested items. E.g. I must be able to just change `max_queriers_per_tenant` in `limits_config` without any other changes.

D3luxee:
Feel free to merge your PR and close mine.
For me its about the feature not the implementation.

What about the use case that my colleague @towolf mentioned with the ruler?
Should we open an issue for this, maybe we can share how we resolved it ourself with the current chart.

jeschkies:
> Should we open an issue for this, maybe we can share how we resolved it ourself with the current chart.

That would be very helpful :slightly_smiling_face:

trevorwhitney:
closing in favor of #7108

PR #7057, opened 2022-09-05
Improvements to Promtail GCP push target

thepalbi:

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/clients/pkg/promtail/targets/gcplog/push_target.go
+++ b/clients/pkg/promtail/targets/gcplog/push_target.go
@@ -22,13 +22,14 @@ import (
 )
 
 type pushTarget struct {
-	logger         log.Logger
-	handler        api.EntryHandler
 	config         *scrapeconfig.GcplogTargetConfig
+	entries        chan<- api.Entry
+	handler        api.EntryHandler
 	jobName        string
-	server         *server.Server
+	logger         log.Logger
 	metrics        *Metrics
 	relabelConfigs []*relabel.Config
+	server         *server.Server
 }
 
 // newPushTarget creates a brand new GCP Push target, capable of receiving message from a GCP PubSub push subscription.
@@ -36,11 +37,12 @@ func newPushTarget(metrics *Metrics, logger log.Logger, handler api.EntryHandler
 	wrappedLogger := log.With(logger, "component", "gcp_push")
 
 	ht := &pushTarget{
-		metrics:        metrics,
-		logger:         wrappedLogger,
+		config:         config,
+		entries:        handler.Chan(),
 		handler:        handler,
 		jobName:        jobName,
-		config:         config,
+		logger:         wrappedLogger,
+		metrics:        metrics,
 		relabelConfigs: relabel,
 	}
 
@@ -96,7 +98,6 @@ func (h *pushTarget) run() error {
 }
 
 func (h *pushTarget) push(w http.ResponseWriter, r *http.Request) {
-	entries := h.handler.Chan()
 	defer r.Body.Close()
 
 	pushMessage := PushMessage{}
@@ -125,7 +126,7 @@ func (h *pushTarget) push(w http.ResponseWriter, r *http.Request) {
 
 	level.Debug(h.logger).Log("msg", fmt.Sprintf("Received line: %s", entry.Line))
 
-	entries <- entry
+	h.entries <- entry
 	h.metrics.gcpPushEntries.WithLabelValues().Inc()
 	w.WriteHeader(http.StatusNoContent)
 }
--- a/clients/pkg/promtail/targets/gcplog/push_translation.go
+++ b/clients/pkg/promtail/targets/gcplog/push_translation.go
@@ -1,15 +1,15 @@
 package gcplog
 
 import (
+	"bytes"
 	"encoding/base64"
 	"fmt"
-	"regexp"
-	"strings"
-	"time"
-
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/model/relabel"
+	"regexp"
+	"strings"
+	"time"
 
 	"github.com/grafana/loki/clients/pkg/promtail/api"
 	lokiClient "github.com/grafana/loki/clients/pkg/promtail/client"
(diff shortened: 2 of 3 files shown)


clients/pkg/promtail/targets/gcplog/push_translation.go
```diff
@@ -37,29 +28,39 @@ type PushMessage struct {
...
 	lbs := labels.NewBuilder(nil)
 	lbs.Set("__gcp_message_id", m.Message.ID)
-
-	// labels from gcp log entry. Add it as internal labels
+	lbs.Set("__gcp_subscription_name", m.Subscription)
 	for k, v := range m.Message.Attributes {
 		lbs.Set(fmt.Sprintf("__gcp_attributes_%s", convertToLokiCompatibleLabel(k)), v)
 	}
 
+	// Apply relabeling
```
kgeckhart, line 55: Do you think there would be any harm in removing the `Apply relabeling` and `Drop internal or invalid labels` logic to let the `parseGCPLogsEntry` call further down take care of it? At a glance it looks like the logic is the same.

thepalbi: Because of how `parseGCPLogsEntry` uses the push-specific labels we pass down, they wouldn't be affected by the relabeling + drop internal it does. [This](https://github.com/thepalbi/loki/blob/1da24900d668fa7afa33e4f00e8c92d9f2e4078c/clients/pkg/promtail/targets/gcplog/formatter.go#L87) is the point they are used, which is after relabel + drop internal.
We could could instead pass them like:
```
func parseGCPLogsEntry(
	data []byte,
	other model.LabelSet,
	useIncomingTimestamp bool,
	relabelConfig []*relabel.Config,
        additionalInternalLabels ...model.LabelSet
) (api.Entry, error)
```
That way we differentiate them from the ones that must be merged after like the fixed labels, or the `__tenant_id__`. wdyt? It would just be to avoid repeating code, since the sum of iterations between both loops is the same

clients/pkg/promtail/targets/gcplog/push_translation.go
```diff
@@ -73,45 +74,46 @@ func translate(m PushMessage, other model.LabelSet, useIncomingTimestamp bool, r
...
 
-	ts := time.Now()
-
 	decodedData, err := base64.StdEncoding.DecodeString(m.Message.Data)
 	if err != nil {
 		return api.Entry{}, fmt.Errorf("failed to decode data: %w", err)
 	}
-	line := string(decodedData)
-
-	if useIncomingTimestamp {
```
jeschkies, line 87: Where did this go?

clients/pkg/promtail/targets/gcplog/push_translation.go
```diff
@@ -73,45 +74,46 @@ func translate(m PushMessage, other model.LabelSet, useIncomingTimestamp bool, r
...
 // such as `logging.googleapis.com/timestamp`, which contain non-loki-compatible characters, which is just alphanumeric
 // and _. The approach taken is to translate every non-alphanumeric separator character to an underscore.
 func convertToLokiCompatibleLabel(label string) string {
-	// TODO: Since this is running for every incoming message, maybe it's more performant to do something
-	// like a loop over characters, and checking if it's not loki compatible, instead of a regexp.
-	return util.SnakeCase(labelToLokiCompatible.ReplaceAllString(label, "_"))
+	return util.SnakeCase(convertSeparatorCharacterToUnderscore(label))
+}
+
+func convertSeparatorCharacterToUnderscore(s string) string {
```
jeschkies, line 108: Hm, there must be something like this in Loki already...but I couldn't find it myself.

In any case `fmt.Fprintf` might be slow. What about `strings.Replacer`? https://go.dev/play/p/9W5a0KvB5M0

clients/pkg/promtail/targets/gcplog/push_translation.go
```diff
@@ -73,45 +74,46 @@ func translate(m PushMessage, other model.LabelSet, useIncomingTimestamp bool, r
...
-		Entry: logproto.Entry{
-			Timestamp: ts,
-			Line:      line,
-		},
-	}, nil
+	// The labelset built by merging fixed target-level labels, push-specific labels and tenant ID is passed as to be merged
+	// to the resulting one of processing the actual GCP Logs entry.
+	entry, err := parseGCPLogsEntry(decodedData, labels, useIncomingTimestamp, relabelConfigs)
+	if err != nil {
+		return api.Entry{}, fmt.Errorf("failed to process logs entry: %w", err)
```
kavirajk, line 95:
```suggestion
		return api.Entry{}, fmt.Errorf("failed to parse logs entry: %w", err)
```

clients/pkg/promtail/targets/gcplog/push_target.go
```diff
@@ -96,36 +98,41 @@ func (h *pushTarget) run() error {
...
 	err = json.Unmarshal(bs, &pushMessage)
 	if err != nil {
-		h.metrics.gcpPushErrors.WithLabelValues().Inc()
+		h.metrics.gcpPushErrors.WithLabelValues("format").Inc()
 		level.Warn(h.logger).Log("msg", "failed to unmarshall gcp push request", "err", err.Error())
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
+	if err = pushMessage.Validate(); err != nil {
+		h.metrics.gcpPushErrors.WithLabelValues("invalid").Inc()
```
kavirajk, line 119:
```suggestion
		h.metrics.gcpPushErrors.WithLabelValues("invalid_message").Inc()
```

kavirajk: may be `invalid_message`?

clients/pkg/promtail/targets/gcplog/push_translation.go
kavirajk, line 108: Thanks for this change. :+1: 
should we lock this with some test cases? (if there isn't any?).

clients/pkg/promtail/targets/gcplog/push_translation.go
kavirajk, line 87: Q: given we have `useIncomingTimestamp` config is `false` by default. Removing this would break the existing scrape config? No?

https://grafana.com/docs/loki/next/clients/promtail/scraping/#push

clients/pkg/promtail/targets/gcplog/push_translation.go
kavirajk: If it's not too much of a work, I would also prefer to remove the repeating code from `translate` and let `parseGCPLogsEntry` handles it. 

Or we can move relabel and merge to another helper and just call it in two places? will that work?

kavirajk (commented):
Thanks @thepalbi for this follow up PR. Added few suggestions and minor nits but overall looks great :+1:

clients/pkg/promtail/targets/gcplog/push_translation.go
thepalbi: `parseGCPLogsEntry` has a similar logic that uses fields that that make more sense to use as timestamp, instead of the push one.

thepalbi: See [this answer](https://github.com/grafana/loki/pull/7057#discussion_r965006269).

docs/sources/clients/promtail/configuration.md
```diff
@@ -957,7 +957,7 @@ When using the `push` subscription type, keep in mind:
 [server: <server_config>]
 
 # Whether Promtail should pass on the timestamp from the incoming GCP Log message.
-# When false, or if no timestamp is present in the syslog message, Promtail will assign the current
+# When false, or if no timestamp is present in the GCP Log message, Promtail will assign the current
```
thepalbi, line 960: drive by

kavirajk (approved):
LGTM 👍 thanks @thepalbi

clients/pkg/promtail/targets/gcplog/push_translation.go
```diff
@@ -37,81 +27,61 @@ type PushMessage struct {
...
+var separatorCharacterReplacer *strings.Replacer
 
-	return api.Entry{
-		Labels: labels,
-		Entry: logproto.Entry{
-			Timestamp: ts,
-			Line:      line,
-		},
-	}, nil
+func init() {
```
kavirajk, line 78: Can you please move it to global const? (I guess no reason to use `init()` here?)

PR #7063, opened 2022-09-06
feat(loki-canary): Add support to push logs directly to Loki.

kavirajk:
Add additional `push` mode to loki-canary, which pushes the logs directly to given Loki URL as it generates logs.

The real function of Loki Canary is to act like a tenant and help us know the whether Loki installation is working as perceived by a real tenant. 

Main rationale for this additional push mode is to make canary more standalone without needing for `promtail` (or `grafana-agent`) to scrape it's logs and send to loki, with this change, Loki canary happily tests Loki behavior without needing any other dependencies.

**NOTE**:
If you run Loki behind any proxy that has different authorization policies to READ or WRITE to Loki, then  important change that canary operator need to be aware of it, now the user credentials they pass via `-user` and `-pass` to access loki endpoints need to have both `read' and `write` permissions (previously canary just query the logs where as promtail is the one pushes the logs, so just READ permissions was sufficient).

Fixes #<issue number>

There will be follow up PR(s) to cleanup, particularly `reader` and `comparitor` component in terms of logging with proper logger. Rationale is to keep the changes small per PR to make it easy to review.

This changes were tested it in one of the internal Loki dev cell.

**This PR is a no-op if this new `push` flag is disabled (it's disabled by default)**

**Checklist**
- [x] Documentation added
- [x] Tests updated


MasslessParticle (approved):

cmd/loki-canary/main.go
```diff
@@ -112,15 +116,45 @@ func main() {
...
+				config.DefaultHTTPClientConfig,
+				*lName, *lVal,
+				*sName, *sValue,
+				tlsConfig,
+				*caFile,
+				*user, *pass,
+				log.NewLogfmtLogger(os.Stdout),
+			)
+			if err != nil {
+				_, _ = fmt.Fprintf(os.Stderr, "Unable to create writer for Loki, check config: %s", err)
```
dannykopping, line 150: Why not use the logger here?

cmd/loki-canary/main.go
```diff
@@ -148,7 +182,7 @@ func main() {
 	}()
 
 	terminate := make(chan os.Signal, 1)
-	signal.Notify(terminate, syscall.SIGTERM, os.Interrupt)
+	signal.Notify(terminate, syscall.SIGTERM, syscall.SIGINT, os.Interrupt)
```
dannykopping, line 185: Why was this added?

pkg/canary/writer/push.go
```diff
@@ -0,0 +1,208 @@
...
+	defaultMaxReponseBufferLen = 1024
+
+	pushEndpoint = "/loki/api/v1/push"
+)
+
+var (
+	defaultUserAgent = fmt.Sprintf("canary-push/%s", build.GetVersion().Version)
+)
+
+// Push is a io.Writer, that writes given long entries by pushing
```
dannykopping, line 37:
```suggestion
// Push is a io.Writer, that writes given log entries by pushing
```

pkg/canary/writer/push.go
```diff
@@ -0,0 +1,208 @@
...
+// directly to the given loki server URL. Each `Push` instance handles for a single tenant.
+// No batching of log lines happens when sending to Loki.
```
dannykopping, line 39: Why no batching?

pkg/canary/writer/push.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to push payload: %w", err)
+	}
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			level.Error(p.logger).Log("msg", "failed to close response body", "error", err)
+		}
+	}()
+
+	if resp.StatusCode/100 != 2 {
```
dannykopping, line 198: We should make it very explicit that this push mode does not retry when rate-limited, or when receiving 5xx responses.

I'm skeptical about this, though, because it means that the canary may show that a Loki cell is unhealthy even when things are fine, because of a transient error or a rate-limit.

pkg/canary/writer/push_test.go
```diff
@@ -0,0 +1,173 @@
...
+	testUsername = "canary"
+	testPassword = "secret"
+)
+
+func Test_Push(t *testing.T) {
+	// create dummy loki server
+	responses := make(chan response, 1) // buffered not to block the response handler
+
+	// mock loki server
+	mock := httptest.NewServer(createServerHandler(responses))
```
dannykopping, line 34: Nice test!

dannykopping (changes requested):
Raised a couple points here, let's address before we merge because I see you already have one approval.

pkg/canary/writer/push.go
kavirajk: The main rationale is two folded

1. Handling batch means, we need to take care of the lifecycle of the pusher during shutdown, say flush the batch before stopping. Kinda complicates the existing canary model a bit.
2.Given we push 1 or 2 logs (with just tiny payload) per sec. Didn't wanted to add these complexity now.

cmd/loki-canary/main.go
kavirajk: Oops. `os.Interrupt` is same as `syscall.SIGINT`. Overlooked. will remove

cmd/loki-canary/main.go
kavirajk: Currently, before this PR, canary was never using proper logger and were writing directly to os.Stderr and os.Stdout (both in `main.go` and other components).

I just wanted to keep it consistent for this PR (focussed on only changes related to `push`) and do cleanup in separate PR as mentioned in the reviewer note
```
There will be follow up PR(s) to cleanup, particularly reader and comparitor component in terms of logging with proper logger. Rationale is to keep the changes small per PR to make it easy to review.
```

pkg/canary/writer/push.go
kavirajk: Yea right. there is no backoff or retry mechanism in this push mode currently. that was intentional for the same purpose you mentioned. It's just a normal http client with timeout. curious what made you to suspect this would retry?

or do you want to add some comment for clarification?

kavirajk: >I guess because any transient failure in the Loki cell which self-heals will look like a failure to the canary, and give the operator an incorrect impression

So I assume you want to have retry mechanism similar to promtail. sounds reasonable. will add it

dannykopping: Yeah maybe an exponential backoff with a configurable number of retries
Doesn't have to be too fancy

kavirajk: Done :+1:

pkg/canary/writer/push.go
```diff
@@ -185,24 +191,38 @@ func (p *Push) send(ctx context.Context, payload []byte) error {
...
+		}
+		status := resp.StatusCode
+
+		if status/100 != 2 {
+			scanner := bufio.NewScanner(io.LimitReader(resp.Body, defaultMaxReponseBufferLen))
+			line := ""
+			if scanner.Scan() {
+				line = scanner.Text()
+			}
+			err = fmt.Errorf("server returned HTTP status %s (%d): %s", resp.Status, status, line)
```
dannykopping, line 211: Is this err logged anywhere?

dannykopping (commented):
Thanks @kavirajk! One comment about logging, then I think we're good to go

pkg/canary/writer/push.go
kavirajk: It is returned from this func at the end. And will be logged by the client calling `.Write()`.
https://github.com/grafana/loki/blob/kavirajk%2Fsupport-push-mode-canary/pkg/canary/writer/writer.go#L100

dannykopping: Ah, gotcha. Thanks.

dannykopping (approved):
LGTM!

PR #7067, opened 2022-09-06
Missing conditional on ServiceMonitor

### LINKED ISSUE #7066: Helm ServiceMonitoring cannot be disabled
**Describe the bug**

In the `3.0.0` helm chart, the monitoring resource `ServiceMonitor` is created by default. Not all Kubernetes clusters use the `monitoring.coreos.com/v1` API and so the resource fails to install. 

The helm `values.yaml` does contain `monitoring.serviceMonitor.enabled` so I assume this conditional was missed off within the ServiceMonitoring resource. 

**To Reproduce**
Install the helm chart with the default values to a cluster without `monitoring.coreos.com/v1` API

**Expected behavior**
You should be able to disable the need to install `ServiceMonitor` when it's not applicable 

PR Incoming shortly

adamstrawson:

Affects the Loki Helm chart

The ServiceMonitor resource was missing a conditional for the `enabled` flag, as per the other monitoring resources.

Fixes #7066

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/helm/loki/templates/monitoring/servicemonitor.yaml
+++ b/production/helm/loki/templates/monitoring/servicemonitor.yaml
@@ -1,5 +1,6 @@
----
+{{- if .Values.monitoring.serviceMonitor.enabled }}
 {{- with .Values.monitoring.serviceMonitor }}
+---
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
@@ -53,4 +54,5 @@ spec:
       tlsConfig:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+{{- end }}
 {{- end }}
--- a/production/helm/loki/templates/monitoring/servicemonitor.yaml
+++ b/production/helm/loki/templates/monitoring/servicemonitor.yaml
@@ -1,5 +1,5 @@
-{{- if .Values.monitoring.serviceMonitor.enabled }}
 {{- with .Values.monitoring.serviceMonitor }}
+{{- if .enabled }}
 ---
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor


production/helm/loki/templates/monitoring/servicemonitor.yaml
```diff
@@ -1,5 +1,6 @@
----
+{{- if .Values.monitoring.serviceMonitor.enabled }}
```
trevorwhitney, line 1: nit: if we move that down a line, we can just say `{{ if .enabled }}` which might read a tad nicer, wdyt?

trevorwhitney (approved):
one small nit and please bump the chart version as well, thanks for the fix!

production/helm/loki/templates/monitoring/servicemonitor.yaml
adamstrawson: Happy to make the change, although it wouldn't be consistent with the other monitoring resources? Thoughts?

https://github.com/grafana/loki/blob/75648288ac1dbf2b39a81f47dac0def381399b33/production/helm/loki/templates/monitoring/grafana-agent.yaml#L1
https://github.com/grafana/loki/blob/75648288ac1dbf2b39a81f47dac0def381399b33/production/helm/loki/templates/monitoring/prometheus-rules.yaml#L1

trevorwhitney: fixed: https://github.com/grafana/loki/pull/7096

adamstrawson: Nice one, I was happy to include those changes but wanted the go ahead first, so thanks for that! 😅  

I've updated the conditional to match your PR, once yours is merged I'll bump the chart versions here too.

PR #7069, opened 2022-09-06
Add support for custom internal server

periklis:
The following PR adds support for a custom internal server listener dedicated for the start to serve the ready handler on a separate port and optionally TLS configuration. The key purpose for this one is to serve the Loki HTTP API via mTLS while serving the readiness endpoints without (i.e. kubernetes readiness/liveness probes do not support authentication)

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -369,9 +369,8 @@ func (t *Loki) Run(opts RunOpts) error {
 	// before starting servers, register /ready handler. It should reflect entire Loki.
 	if t.Cfg.InternalServer.Enable {
 		t.InternalServer.HTTP.Path("/ready").Methods("GET").Handler(t.readyHandler(sm))
-	} else {
-		t.Server.HTTP.Path("/ready").Methods("GET").Handler(t.readyHandler(sm))
 	}
+	t.Server.HTTP.Path("/ready").Methods("GET").Handler(t.readyHandler(sm))
 
 	grpc_health_v1.RegisterHealthServer(t.Server.GRPC, grpcutil.NewHealthCheck(sm))
 
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -40,6 +40,7 @@ import (
 	"github.com/grafana/loki/pkg/ruler/rulestore"
 	"github.com/grafana/loki/pkg/runtime"
 	"github.com/grafana/loki/pkg/scheduler"
+	internalserver "github.com/grafana/loki/pkg/server"
 	"github.com/grafana/loki/pkg/storage"
 	"github.com/grafana/loki/pkg/storage/chunk/cache"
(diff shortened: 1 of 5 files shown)


pkg/loki/loki.go
```diff
@@ -131,6 +133,7 @@ func (c *Config) registerServerFlagsWithChangedDefaultValues(fs *flag.FlagSet) {
 	// Register to throwaway flags first. Default values are remembered during registration and cannot be changed,
 	// but we can take values from throwaway flag set and reregister into supplied flags with new default values.
 	c.Server.RegisterFlags(throwaway)
+	c.InternalServer.RegisterFlags(throwaway)
```
simonswine, line 136: Don't think that is necessary

pkg/server/internal_server.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"flag"
+	"time"
+
+	serverww "github.com/weaveworks/common/server"
+)
+
+// Config extends weaveworks server config
+type Config struct {
+	serverww.Config `yaml:",inline"`
```
simonswine, line 12: That also includes a lot of struct fields which are unused, I would prefer to shrink those down a bit more.

I don't know if it even needs to be supporting TLS, but I guess the list of the registered flags would be a good start.

simonswine (approved):
LGTM. I think this is a worthwhile improvement.

I am also wondering if it is worth including other endpoints (like `/metrics`) to that internal listener? (Not necessarily in this PR)

pkg/server/internal_server.go
periklis: In the scope of exposing only the ready handler it makes sense to shrink this more. However, as you mention above it makes sense to expose the metrics from this handler too and thus I would like to keep the inline struct as such for start but limit the inputs via the shrinked registered flags. As per exposing metrics allowing TLS from start is imho a good thing to do. For example on OpenShift we allow only scraping over HTTPS.

pkg/loki/loki.go
periklis: Actually it is need to register the flags to main flagset of Loki or else we can't configure the internal server.

pkg/server/internal_server.go
```diff
@@ -0,0 +1,31 @@
...
+	Enable          bool `yaml:"enable"`
+}
+
+// RegisterFlags add internal server flags to flagset
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.StringVar(&cfg.Config.HTTPListenAddress, "internal-server.http-listen-address", "", "HTTP internal server listen address.")
```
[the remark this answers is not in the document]
periklis: I took the defaults from weaveworks verbatim, but for an internal server that is used for readiness and metrics it makes sense to default listen on localhost.

PR #7070, opened 2022-09-06
operator: Add support for certificates to allow external S3 storage connections

shwetaap:
Signed-off-by: Shweta Padubidri <spadubid@redhat.com>

This PR supports bringing up Loki in a cluster that runs behind a proxy
It also adds support to provide proxy certificates to support secure traffic behind a proxy in a cluster. 

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operator/internal/handlers/lokistack_create_or_update.go
+++ b/operator/internal/handlers/lokistack_create_or_update.go
@@ -110,15 +110,15 @@ func CreateOrUpdateLokiStack(
 			return kverrors.Wrap(err, "failed to lookup lokistack object storage CA config map", "name", key)
 		}
 
-		if !storage.IsValidCAConfigMap(&cm) {
+		if !storage.IsValidCAConfigMap(&cm, stack.Spec.Storage.TLS.Key) {
 			return &status.DegradedError{
-				Message: "Invalid object storage CA configmap contents: missing key `service-ca.crt` or no contents",
+				Message: "Invalid object storage CA configmap contents: missing key or no contents",
 				Reason:  lokiv1.ReasonInvalidObjectStorageCAConfigMap,
 				Requeue: false,
 			}
 		}
 
-		objStore.TLS = &storageoptions.TLSConfig{CA: cm.Name}
+		objStore.TLS = &storageoptions.TLSConfig{CA: cm.Name, Key: stack.Spec.Storage.TLS.Key}
 	}
 
 	var (
--- a/operator/apis/loki/v1/lokistack_types.go
+++ b/operator/apis/loki/v1/lokistack_types.go
@@ -323,6 +323,14 @@ type LokiTemplateSpec struct {
 
 // ObjectStorageTLSSpec is the TLS configuration for reaching the object storage endpoint.
 type ObjectStorageTLSSpec struct {
+	// Key is the data key of a ConfigMap containing a CA certificate.
+	// It needs to be in the same namespace as the LokiStack custom resource.
+	//
+	// +optional
+	// +kubebuilder:validation:optional
+	// +kubebuilder:default:=service-ca.crt
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:io.kubernetes:ConfigMap",displayName="CA ConfigMap Key"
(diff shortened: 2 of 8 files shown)


operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -204,6 +204,20 @@ func CreateOrUpdateLokiStack(
 		}
 	}
 
+	proxyCfg, err := manifests.GetProxyConfig(k.Client)
```
periklis, line 207: Requests for active resources on the k8s apiserver should be place in the `manifests` package. The `manifests` package is dedicated for creating resource manifests (e.g. deployments, services, configmaps, etc.) from static inputs passed via `manifests.Options`. Consider moving the code using the k8s client to an internal package dedicated for the `handlers` packages, e.g. rules, gateway secrets in https://github.com/grafana/loki/tree/main/operator/internal/handlers/internal

From the resulting configmap we extract what we need and pass it as raw inputs into a new `manifests.Options` field. This inputs can be used to create the trusted ca bundle configmap statically.

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -204,6 +204,20 @@ func CreateOrUpdateLokiStack(
...
+	}
+
+	// Create cluster proxy trusted CA bundle.
+	var lokiTrustBundle *corev1.ConfigMap
+	if proxyCfg != nil {
+		lokiTrustBundle, err = manifests.createOrGetTrustedCABundleConfigMap(manifests.LokiTrustedCAName, ctx, k, stack)
+		if err != nil {
+			return err
+		}
+	}
```
periklis, line 219: As said above the Get part of this methods belongs to a dedicated handlers internal package. In addition we should not mix `create` with `get` in a single function here (as we do in other operators unfortunately) for a couple of reasons:
1. We don't rely in this operator on custom `createOrUpdate` functions as we use the generic one from the controller-runtime. This keep some sanity and uniformity in equality checks and create/update logic.
2. We keep `get` calls isolated in the handlers package and pass them down to the `manifests` builders via `manifests.Options`. This ensures that are our builders incl. conditional logic (e.g. openshift, feature flags) do not require active communication to the APIserver and remain easy to test via unit tests.

Besides the above the creation of this TrustedCABundle ConfigMap incl. the Volumes/VolumeMounts additions for the Loki Components is a conditional feature. This means we need an amendment in the LokiStack API for users to input/select the configMap that includes their TrustedCABundle. This enables making this feature k8s-native first. For OpenShift users can rely that we auto-populate the configmap via the dedicated inject-ca-bundle annotations. Latter should be dedicated patching code isolated in the `manifests.openshift` package.

operator/internal/manifests/ingester.go
```diff
@@ -121,6 +122,10 @@ func NewIngesterStatefulSet(opts Options) *appsv1.StatefulSet {
 		SecurityContext: podSecurityContext(opts.Gates.RuntimeSeccompProfile),
 	}
 
+	for i, container := range podSpec.Containers {
+		podSpec.Containers[i].Env = append(container.Env, proxy.ReadProxyVarsFromEnv()...)
+	}
```
periklis, line 127: Do we always want to do this?

periklis (changes requested):
This draft shows a clear path forward on how to expand LokiStack's capabilities to talk to dependencies (i.e. S3 storage) when a proxy is used and/or a TrustedCABundle is required for successful data-in-transit encryption.

Having said this, as commented below I suggest to improve this PR in two key directions to make this useful for non-OpenShift users, too:
1. The TrustedCABundle ConfigMap belongs to the LokiStack API to allow users define this if needed. For OpenShift users we can automate the user choice.
2. Populating the Loki container env vars always with the Proxy env vars is something we need to discuss if this is should be a user choice or a solely operator choice. I like to err towards "let the users make a choice", but I am open for arguments here.

Besides the above key points, please consider the code placement comments below to keep our `manifests` package dedicated for building static manifests and separating openshift-related code in the dedicated package for this purpose.

periklis (approved):

PR #7073, opened 2022-09-06
Add prefix for helm-release tags

trevorwhitney:

The helm release job creates a tag for each new release. The tags was previously `<chart name>-<chart version>`, which in the case of the loki chart in this repo was `loki-3.0.0`. This is confusing, so we are adding the prefix `helm-` to these  tags.


DylanGuedes:
double checking, don't we have to update this on any docs?

trevorwhitney:
nope, the tag is not mentioned anywhere, but the old one did cause confusion in the community

MasslessParticle (approved):

kavirajk (approved):

PR #7087, opened 2022-09-07
Add sequence num to log output for delete requests

MichelHollands:

Print the sequence number of the delete request so the messages are not all the same.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
+++ b/pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
@@ -270,6 +270,7 @@ func (d *DeleteRequestsManager) Expired(ref retention.ChunkEntry, _ model.Time) 
 			level.Info(util_log.Logger).Log(
 				"msg", "no chunks to retain: the whole chunk is deleted",
 				"delete_request_id", deleteRequest.RequestID,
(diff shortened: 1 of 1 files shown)


pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
```diff
@@ -270,6 +270,7 @@ func (d *DeleteRequestsManager) Expired(ref retention.ChunkEntry, _ model.Time)
 			level.Info(util_log.Logger).Log(
 				"msg", "no chunks to retain: the whole chunk is deleted",
 				"delete_request_id", deleteRequest.RequestID,
+				"sequence num", deleteRequest.SequenceNum,
```
dannykopping, line 273: I'd suggest making these snake_case to match the rest

```suggestion
				"sequence_num", deleteRequest.SequenceNum,
```

dannykopping (approved):
LGTM with nit

pkg/storage/stores/indexshipper/compactor/deletion/delete_requests_manager.go
MichelHollands: Good catch.

PR #7093, opened 2022-09-07
Only release helm for helm paths

trevorwhitney:

Only need to run the helm release action when helm paths were changed


cstyan (approved):
:+1: assuming the path regex is correct

PR #7094, opened 2022-09-07
Implement all-or-nothing stream ingestion

MasslessParticle:
Per-stream sharding will introduce automatic stream sharding on rate-limited streams. The ingester's current line-by-line behavior may result in duplicates when a partially-accepted stream is resharded and sent to other ingesters. This PR makes the ingester either rate limit the entire stream before attempting to write lines to a chunk. The PR

1. Separates validation, storage, and wal replay into their own functions
2. Implements a flag to rate limit the whole stream or not
3. Checks rate limits for the whole stream rather than line-by-line
4. Returns an error with information on the whole stream when rate limited

### DIFF
--- a/pkg/ingester/instance.go
+++ b/pkg/ingester/instance.go
@@ -262,7 +262,8 @@ func errorForFailedEntries(s *stream, failedEntriesWithError []entryWithError, t
 			entryWithError.entry.Timestamp.String(), entryWithError.e.Error(), streamName)
 	}
 
-	fmt.Fprintf(&buf, "total ignored: %d out of %d", len(failedEntriesWithError), totalEntries)
+	// If it's a 400 or 429, the whole stream was rejected
+	fmt.Fprintf(&buf, "total ignored: %d out of %d", totalEntries, totalEntries)
 
 	var details []*types.Any
 
--- a/pkg/ingester/instance.go
+++ b/pkg/ingester/instance.go
@@ -262,8 +262,7 @@ func errorForFailedEntries(s *stream, failedEntriesWithError []entryWithError, t
 			entryWithError.entry.Timestamp.String(), entryWithError.e.Error(), streamName)
 	}
 
-	// If it's a 400 or 429, the whole stream was rejected
-	fmt.Fprintf(&buf, "total ignored: %d out of %d", totalEntries, totalEntries)
+	fmt.Fprintf(&buf, "total ignored: %d out of %d", len(failedEntriesWithError), totalEntries)
 
 	var details []*types.Any
 
--- a/pkg/ingester/stream.go
+++ b/pkg/ingester/stream.go
@@ -177,14 +177,14 @@ func (s *stream) Push(
 		s.metrics.chunkCreatedStats.Inc(1)
 	}
 
-	bytesAdded, storedEntries := s.storeEntries(ctx, toStore)
+	bytesAdded, storedEntries, entriesWithErr := s.storeEntries(ctx, toStore)
 	s.recordAndSendToTailers(record, storedEntries)
 
 	if len(s.chunks) != prevNumChunks {
 		s.metrics.memoryChunks.Add(float64(len(s.chunks) - prevNumChunks))
 	}
 
-	return bytesAdded, invalid
+	return bytesAdded, entriesWithErr
 }
 
 func (s *stream) recordAndSendToTailers(record *WALRecord, entries []logproto.Entry) {
@@ -231,9 +231,10 @@ func (s *stream) recordAndSendToTailers(record *WALRecord, entries []logproto.En
 	}
 }
 
-func (s *stream) storeEntries(ctx context.Context, entries []logproto.Entry) (int, []logproto.Entry) {
+func (s *stream) storeEntries(ctx context.Context, entries []logproto.Entry) (int, []logproto.Entry, []entryWithError) {
 	var bytesAdded, outOfOrderSamples, outOfOrderBytes int
 
+	var invalid []entryWithError
 	storedEntries := make([]logproto.Entry, 0, len(entries))
 	for i := 0; i < len(entries); i++ {
 		chunk := &s.chunks[len(s.chunks)-1]
@@ -242,6 +243,7 @@ func (s *stream) storeEntries(ctx context.Context, entries []logproto.Entry) (in
 		}
 
 		if err := chunk.chunk.Append(&entries[i]); err != nil {
+			invalid = append(invalid, entryWithError{&entries[i], err})
 			if chunkenc.IsOutOfOrderErr(err) {
 				outOfOrderSamples++
 				outOfOrderBytes += len(entries[i].Line)
@@ -257,7 +259,7 @@ func (s *stream) storeEntries(ctx context.Context, entries []logproto.Entry) (in
 	}
 
 	s.reportMetrics(outOfOrderSamples, outOfOrderBytes, 0, 0)
-	return bytesAdded, storedEntries
+	return bytesAdded, storedEntries, invalid
 }
 
 func (s *stream) validateEntries(entries []logproto.Entry, isReplay bool) ([]logproto.Entry, []entryWithError) {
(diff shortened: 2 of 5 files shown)


MasslessParticle:
Benchmarks:
```
Original:
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/ingester
Benchmark_PushStream-8              9436            135428 ns/op          167872 B/op       1606 allocs/op
PASS
ok      github.com/grafana/loki/pkg/ingester    1.326s

New:
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/ingester
Benchmark_PushStream-8             10000            120120 ns/op          165196 B/op       1600 allocs/op
PASS
ok      github.com/grafana/loki/pkg/ingester    1.248s



Comparison
benchmark                  old ns/op     new ns/op     delta
Benchmark_PushStream-8     135428        120120        -11.30%

benchmark                  old allocs     new allocs     delta
Benchmark_PushStream-8     1606           1600           -0.37%

benchmark                  old bytes     new bytes     delta
Benchmark_PushStream-8     167872        165196        -1.59%
```

DylanGuedes (changes requested):
Looks superb, gj. I added a few suggestions but they're not really important (except maybe for the one regarding using `latestTs` variable).

But I'm requesting changes to block the PR for the following:
- WDYT of having two implementations: a `(*instance).PushV1` (which would be the old implementation) and a `(*instance).PushV2` (your new Push implementation) and to call each of them depending on `stream_sharding` being enabled or not?
Reason is: even if stream sharding is disabled, the moment we merge this, we'd been rejecting all streams that at least one entry is rate-limited; I fear we might affect the experience of users to the worse undesirely.

pkg/ingester/stream.go
```diff
@@ -173,132 +177,171 @@ func (s *stream) Push(
...
+
+		toStore = append(toStore, entries[i])
 	}
 
-	if len(s.chunks) != prevNumChunks {
-		s.metrics.memoryChunks.Add(float64(len(s.chunks) - prevNumChunks))
+	// Each successful call to 'AllowN' advances the limiter. With all-or-nothing
+	// ingestion, the limiter should only be advanced when the whole stream can be
+	// sent
+	now := time.Now()
```
DylanGuedes, line 319:
```suggestion
	now := time.Now()
	s.highestTs = highestTs
```

pkg/ingester/instance.go
```diff
@@ -262,7 +262,8 @@ func errorForFailedEntries(s *stream, failedEntriesWithError []entryWithError, t
 			entryWithError.entry.Timestamp.String(), entryWithError.e.Error(), streamName)
 	}
 
-	fmt.Fprintf(&buf, "total ignored: %d out of %d", len(failedEntriesWithError), totalEntries)
+	// If it's a 400 or 429, the whole stream was rejected
```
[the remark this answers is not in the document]
MasslessParticle: This is a good call. There's a chance that we won't reject all entries when we `ErrOutOfOrder` from `chunk.Append`.

pkg/ingester/stream.go
```diff
@@ -173,132 +177,171 @@ func (s *stream) Push(
...
+	if !s.limiter.AllowN(now, totalBytes) {
+		rateLimitedSamples = len(toStore)
+		rateLimitedBytes = totalBytes
+		for i := 0; i < len(toStore); i++ {
+			failedEntriesWithError = append(failedEntriesWithError, entryWithError{toStore[i], &validation.ErrStreamRateLimit{RateLimit: flagext.ByteSize(limit), Labels: s.labelsString, Bytes: flagext.ByteSize(len(toStore[i].Line))}})
```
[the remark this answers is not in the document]
MasslessParticle: Ahhh, this is the reason `entryWithError` used a pointer. That way this slice only ever stores pointers to memory that already exists even though it may have to alloc to pass things around.

pkg/ingester/stream.go
```diff
@@ -173,132 +177,171 @@ func (s *stream) Push(
 		s.metrics.chunkCreatedStats.Inc(1)
 	}
 
-	var storedEntries []logproto.Entry
-	failedEntriesWithError := []entryWithError{}
+	bytesAdded, storedEntries, entriesWithErr := s.storeEntries(ctx, toStore)
+	s.recordAndSendToTailers(record, storedEntries)
+
+	if len(s.chunks) != prevNumChunks {
```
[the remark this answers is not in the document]
MasslessParticle: This should be ok. I don't think the number of chunks can ever go backward.

pkg/ingester/stream.go
```diff
@@ -209,96 +298,69 @@ func (s *stream) Push(
...
-						continue
-					}
-					tailer.send(stream, s.labels)
-				}
-				s.tailerMtx.RUnlock()
-
-				if len(closedTailers) != 0 {
-					s.tailerMtx.Lock()
-					defer s.tailerMtx.Unlock()
+	s.reportMetrics(outOfOrderSamples, outOfOrderBytes, rateLimitedSamples, rateLimitedBytes)
```
DylanGuedes, line 347: Hmm double checking: I believe you are discarding out of order-related metrics two times: here and in the storeEntries func.

pkg/ingester/stream_test.go
```diff
@@ -427,3 +463,36 @@ func Benchmark_PushStream(b *testing.B) {
...
+	limiter := NewLimiter(limits, NilMetrics, &ringCountMock{count: 1}, 1)
+
+	s := newStream(&Config{MaxChunkAge: 24 * time.Hour, RateLimitWholeStream: true}, limiter, "fake", model.Fingerprint(0), ls, true, NilMetrics)
+	t, err := newTailer("foo", `{namespace="loki-dev"}`, &fakeTailServer{}, 10)
+	require.NoError(b, err)
+
+	go t.loop()
+	defer t.close()
+
+	s.tailers[1] = t
```
DylanGuedes, line 486: Q: by default, there are already two pre-created tailers?

DylanGuedes (approved):
lgtm! I added one observation regarding potentially reporting metrics twice but shouldn't block your work. Also, let's wait for one more pair of eyes :smile:

pkg/ingester/stream.go
MasslessParticle: Things won't be counted twice because we won't try to store anything filtered by `validation`.

pkg/ingester/stream.go
```diff
@@ -163,7 +163,11 @@ func (s *stream) Push(
 		return 0, []entryWithError{{entry: &logproto.Entry{}, e: ErrEntriesExist}}
 	}
 
-	var bytesAdded int
+	toStore, invalid := s.validateEntries(entries, isReplay)
+	if s.cfg.RateLimitWholeStream && hasRateLimitErr(invalid) {
```
owen-d, line 167: I think we can do this more efficiently. Currently, we're
1) Iterating through the entries for validation, calculating total bytes
2) iterating through all entries again to attach errors to them if they trip the rate limit
3) iterating them a third time here to see if one of them is rate limited

I think we can do this more efficiently, but happy to let it be in a separate followup PR.

owen-d (approved):
I think this works. Left some nits for discussion, but will approve it.

pkg/ingester/stream.go
```diff
@@ -209,96 +298,69 @@ func (s *stream) Push(
...
 	}
 
-	if len(storedEntries) != 0 {
-		// record will be nil when replaying the wal (we don't want to rewrite wal entries as we replay them).
-		if record != nil {
-			record.AddEntries(uint64(s.fp), s.entryCt, storedEntries...)
-		} else {
-			// If record is nil, this is a WAL recovery.
-			s.metrics.recoveredEntriesTotal.Add(float64(len(storedEntries)))
+	// Each successful call to 'AllowN' advances the limiter. With all-or-nothing
```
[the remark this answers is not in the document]
MasslessParticle: It's a little misleading. The number of available tokens on the limit is only updated when the check succeeds https://cs.opensource.google/go/x/time/+/e5dcc9cf:rate/rate.go;l=346-368;drc=583f2d630306214ee49ea373317b53b90026aab7

PR #7095, opened 2022-09-07
Add chart-tools and chart-releaser for helm

trevorwhitney:

Adds the `chart-tools` and `chart-releaser` binaries to the nix shell for doing helm releases/testing.


simonswine (approved):
LGTM

PR #7109, opened 2022-09-08
Fix pod affinity selector for single binary

### LINKED ISSUE #7105: [Helm 3.0.1] loki.singleBinarySelectorLabels in values.yaml
Hello,

in the new helm chart, in the values.yaml, line 733, I think there is a mistake :
https://github.com/grafana/loki/blob/main/production/helm/loki/values.yaml#L733
It should be 
```yaml
              {{- include "loki.singleBinarySelectorLabels" . | nindent 10 }}
```
Could you check that ?

Rgds.

trevorwhitney:

Fix incorrect selectors for anti-affinity rules for loki single binary.

Fixes #7105

### DIFF
--- a/production/helm/loki/values.yaml
+++ b/production/helm/loki/values.yaml
@@ -730,7 +730,7 @@ singleBinary:
       requiredDuringSchedulingIgnoredDuringExecution:
         - labelSelector:
             matchLabels:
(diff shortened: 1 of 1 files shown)


DylanGuedes (approved):
lgtm

PR #7125, opened 2022-09-09
Fix errors loading deletes after force-removing partially-completed delete

MasslessParticle:
When sharded deletes that are partially completed are force-removed, everything is removed except the shards that are currently being deleted. The shards being deleted aren't necessarily in numerical order so there's no way to know assume which shards have and haven't been deleted.

The Delete store assumed it could look up requests in sequential order and returned empty requests where there were gaps in the stored data. This change takes the `sequence number` from the stored key of the delete requests it finds and uses that to look up details.


MichelHollands (approved):
LGTM

PR #5051, opened 2022-01-05
[new] promtail pipeline:  Promtail Rate Limit stage #5048

liguozhong:
Promtail Rate Limit stage #5048
image

PR #7127, opened 2022-09-10
[new feature] promtail: Add a Promtail stage for probabilistic sampling

### LINKED ISSUE #6654: Add a Promtail stage for probabilistic sampling
Sometimes we don't have direct access to what or how often an application logs, especially for third-party dependencies. Furthermore, I (personally) will primarily dig into logs when there’s something going wrong. 😅 

For this reason, users may find it useful to be able to limit the ingestion rate of certain types of logs, both for lower costs and signal-to-noise reduction. Like we do for traces, probabilistic sampling is a good option here, and Loki’s label-based approach is also a natural fit. For example, users could sample to keep 10% of ‘trace’ level logs, or 20% of logs coming from an 200 - OK response, but always keep logs containing bad words like ‘fatal’ or ‘error’.

One thing to note here, is that ideally we'd try to preserve this information (using a new label?), so that the user can understand that it’s not that "200 - OK" responses have dropped by 80%, it’s just that they’re only logging a subset of them.

Add a `sampling` stage to Promtail. Similar to the `drop` stage, it would accept a regex, a source, and a float between [0, 1].

Then, it would check the log line contents or the given source and probabilistically drop or keep the log line.

Other than instrumenting the application emitting the logs itself, a similar effect could be obtained by a complex pipeline containing the timestamp and a regex on it (eg. to only ingest lines whose for timestamps where 'second' is between [0, 10]).

A similar effect is also obtained by using the `limit` stage, although it is a hardcoded limit, and not a probabilistic one.

**Additional context**
A [similar proposal](https://github.com/grafana/loki/issues/166) was opened a long time ago, but failed to get any traction.

liguozhong:

The sampling stage can be directly sampled.
The implementation of sampling is to use the algorithm in jaeger go client
```
pipeline_stages:
- sampling:
     rate: 0.1
```
or it can be used with match for precise sampling.
```
pipeline_stages:
- json:
    expressions:
      app:
- match:
    pipeline_name: "app2"
    selector: "{app=\"poki\"}"
    stages:
    - sampling:
        rate: 0.1
```

Fixes #6654

The promtail 'rate' stage is also used with the 'match' stage for log filtering.This design makes the code very clean.
Other log agents vector have log filtering built into the sampling operator, which I think is too complicated
https://vector.dev/docs/reference/configuration/transforms/sample/
```
 transforms:
  my_transform_id:
    type: sample
    inputs:
      - my-source-or-transform-id
    exclude: null
    rate: 10
```

'rate' stage review suggestions .
https://github.com/grafana/loki/pull/5051
image

**Checklist**
- [x] Documentation added
- [x] Tests updated


clients/pkg/logentry/stages/sampling.go
```diff
@@ -0,0 +1,121 @@
...
+		samplingBoundary: samplingBoundary,
+		pool: sync.Pool{
+			New: func() interface{} {
+				return rand.NewSource(seedGenerator.Int63())
+			},
+		},
+	}, nil
+}
+
+// dropStage applies Label matchers to determine if the include stages should be run
```
ashwanthgoli, line 74: this needs correction :)

clients/pkg/logentry/stages/sampling.go
```diff
@@ -0,0 +1,120 @@
...
+
+	samplingRate := math.Max(0.0, math.Min(cfg.SamplingRate, 1.0))
+	samplingBoundary := uint64(float64(maxRandomNumber) * samplingRate)
+	seedGenerator := utils.NewRand(time.Now().UnixNano())
+	return &samplingStage{
+		logger:           log.With(logger, "component", "stage", "type", "sampling"),
+		cfg:              cfg,
+		dropCount:        getDropCountMetric(registerer),
+		samplingBoundary: samplingBoundary,
+		pool: sync.Pool{
```
ashwanthgoli, line 66: I am assuming the reference implementation is using a `sync.Pool` for storing rand generators since the generator created by `NewSource` is not safe for concurrent use. And as an added benefit generating random number from this pool would make use of generators created with different seeds.

But in this stage, we have a single go routine that's generating the random numbers which means we don't necessarily need a `sync.Pool` maybe we should drop this

jeschkies:
Nice. Could you update the documentation?

liguozhong:
OK

liguozhong:
done.

docs/sources/clients/promtail/stages/sampling.md
```diff
@@ -0,0 +1,34 @@
...
+title: sampling
+description: sampling stage
+---
+# sampling
+
+The `sampling` stage is a stage that sampling logs. 
+
+## Sampling stage schema
+
+The `sampling` stage is used to sampling the logs.rate: 0.1 means that 10% of the logs can be pushed to the loki server.
```
JStickler, line 11:
```suggestion
The `sampling` stage is used to sampling the logs.  Configuring the value `rate: 0.1` means that 10% of the logs will be pushed to the Loki server.
```

docs/sources/clients/promtail/stages/sampling.md
```diff
@@ -0,0 +1,34 @@
...
+
+```yaml
+sampling:
+  # The rate sampling in lines per second that Promtail will push to Loki.The value is between 0 and 1.
```
JStickler, line 15:
```suggestion
  # The rate sampling in lines per second that Promtail will push to Loki.The value is between 0 and 1, where a value of 0 means no logs are sampled and a value of 1 means 100% of logs are sampled.
```

docs/sources/clients/promtail/stages/sampling.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Examples
+
+The following are examples showing the use of the `sampling` stage.
+
+### sampling
+
+Simple `sampling` stage configurations.
+
+#### Match a line and sampling
```
JStickler, line 27: The following example only includes a sampling rate.  Was there supposed to be some sort of line or string matching?

JStickler (commented):
[Docs squad} Couple of suggestions and questions for the documentation.

liguozhong:
thanks，I modified the document to 2 examples

jeschkies (approved):

jeschkies:
I think @JStickler needs to approve :slightly_smiling_face:

JStickler (approved):
[Docs squad]

liguozhong:
thanks

MasslessParticle:
Awesome, thanks for the contribution!

PrayagS:
> PR was merged 4m ago, any plans for merging it to the release branch as well?

+1. Looking forward to utilizing this feature. With the `limit` stage, it's more of a hit-and-miss thing to get the percentage right.

PR #7129, opened 2022-09-10
Helm chart: Add enableServiceLinks

TAYTS:
This is to allow disable `enableServiceLinks` so that the service environments are not injected into the pods where in my scenario Loki is deployed in the same namespace as the Knative services resulted the Loki pod unable to start up due to too many service environment being injected.

Reference issue for pod unable to startup due to too many services: https://github.com/kubernetes/kubernetes/issues/84539#issue-514261281

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/helm/loki/Chart.yaml
+++ b/production/helm/loki/Chart.yaml
@@ -4,7 +4,7 @@ name: loki
 description: Helm chart for Grafana Loki in simple, scalable mode
 type: application
 appVersion: 2.7.0
-version: 4.1.0
+version: 4.2.0
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -654,7 +654,7 @@ http {
 
 {{/* Configure enableServiceLinks in pod */}}
 {{- define "loki.enableServiceLinks" -}}
-{{- if semverCompare ">=1.13-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if semverCompare ">=1.13-0" .Capabilities.KubeVersion.Version -}}
 {{- if or (.Values.loki.enableServiceLinks) (ne .Values.loki.enableServiceLinks false) -}}
 enableServiceLinks: true
 {{- else -}}
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -651,3 +651,14 @@ http {
   }
 }
 {{- end }}
+
+{{/* Configure enableServiceLinks in pod */}}
+{{- define "loki.enableServiceLinks" -}}
+{{- if semverCompare ">=1.13-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if or (.Values.loki.enableServiceLinks) (eq (.Values.loki.enableServiceLinks | toString) "<nil>") -}}
+enableServiceLinks: true
+{{- else -}}
+enableServiceLinks: false
+{{- end -}}
+{{- end -}}
+{{- end -}}
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -655,7 +655,7 @@ http {
 {{/* Configure enableServiceLinks in pod */}}
 {{- define "loki.enableServiceLinks" -}}
 {{- if semverCompare ">=1.13-0" .Capabilities.KubeVersion.GitVersion -}}
-{{- if or (.Values.loki.enableServiceLinks) (eq (.Values.loki.enableServiceLinks | toString) "<nil>") -}}
+{{- if or (.Values.loki.enableServiceLinks) (ne .Values.loki.enableServiceLinks false) -}}
 enableServiceLinks: true
 {{- else -}}
 enableServiceLinks: false
--- a/production/helm/loki/templates/gateway/deployment-gateway.yaml
+++ b/production/helm/loki/templates/gateway/deployment-gateway.yaml
@@ -37,14 +37,8 @@ spec:
         {{- end }}
         {{- include "loki.gatewaySelectorLabels" . | nindent 8 }}
     spec:
-      serviceAccountName: {{ include "loki.serviceAccountName" . -}}
-      {{- if semverCompare ">=1.13-0" .Capabilities.KubeVersion.GitVersion }}
-      {{- if or (.Values.loki.enableServiceLinks) (eq (.Values.loki.enableServiceLinks | toString) "<nil>") }}
-      enableServiceLinks: true
-      {{- else }}
-      enableServiceLinks: false
-      {{- end }}
-      {{- end }}
+      serviceAccountName: {{ include "loki.serviceAccountName" . }}
+      {{ include "loki.enableServiceLinks" . }}
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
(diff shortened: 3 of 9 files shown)


jeschkies:
Could you raise the Helm Chart version so it's released?

TAYTS:
@jeschkies I have updated the Helm chart version, please help to review again, thank you

production/helm/loki/templates/gateway/deployment-gateway.yaml
```diff
@@ -32,6 +32,13 @@ spec:
         {{- include "loki.gatewaySelectorLabels" . | nindent 8 }}
     spec:
       serviceAccountName: {{ include "loki.serviceAccountName" . -}}
+      {{- if semverCompare ">=1.13-0" .Capabilities.KubeVersion.GitVersion }}
+      {{- if or (.Values.loki.enableServiceLinks) (eq (.Values.loki.enableServiceLinks | toString) "<nil>") }}
```
trevorwhitney, line 36: why doe we need the `toString` compared to `"<nil>"`? shouldn't helm treat a nil value as falsey? Also, would you consider moving this to a single `if` statement, potentially even extracting a template function if it's messy?

trevorwhitney (commented):
Can you describe the problem a bit more? Is there some technical limit on the number of service environment variables that can be injected into a POD? What was the technical issue this created for you?

production/helm/loki/templates/gateway/deployment-gateway.yaml
TAYTS: The `toString` is to set the `enableServiceLinks` to true (default value) when `enableServiceLinks` is not provided in the values.yaml.

We will need  to have 2 if because in k8s version < v1.13.0, there is no `enableServiceLinks` and I have moved this chunk into template to make it dry.

TAYTS: We were deploying the loki-stack in the same namespace as knative and as the service grow, we were not able to start the pod due to too many services environment variables being injected. We are moving loki-stack to different namespace at the moment to fix the issue, but thought this feature could help those that has constraint with the namespace(ie. only allow to deploy to certain namespace).

You can refer to the link I have attached about the pod unable to startup due to too many service environments injected into the pods.
Here is the link for an interesting experiment did for the knative: https://github.com/knative/serving/issues/8498

TAYTS:
@trevorwhitney Is it possible to help review this MR? Thank you so much

production/helm/loki/templates/_helpers.tpl
```diff
@@ -359,3 +359,14 @@ Create the service endpoint including port for MinIO.
 {{- define "enterprise-logs.canarySecret" }}
 {{- .Values.enterprise.canarySecret | default (printf "%s-canary-secret" (include "loki.name" . )) -}}
 {{- end -}}
+
+{{/* Configure enableServiceLinks in pod */}}
+{{- define "loki.enableServiceLinks" -}}
+{{- if semverCompare ">=1.13-0" .Capabilities.KubeVersion.GitVersion -}}
```
trevorwhitney, line 365: why `KubeVersion.GitVersion` instead of `KubeVersion.Version`. We are currently using the latter, and I'd like to keep in consistent if possible.

production/helm/loki/templates/_helpers.tpl
```diff
@@ -359,3 +359,14 @@ Create the service endpoint including port for MinIO.
...
+{{- if or (.Values.loki.enableServiceLinks) (eq (.Values.loki.enableServiceLinks | toString) "<nil>") -}}
+enableServiceLinks: true
+{{- else -}}
+enableServiceLinks: false
+{{- end -}}
```
trevorwhitney, line 370: I still don't love the to string and comparison to the string "<nil>", as I feel like that's too tightly coupled to go template implementation details. How would you feel about switching the logic to avoid that?

```suggestion
{{- if not .Values.loki.enableServiceLinks -}}
enableServiceLinks: false
{{- else -}}
enableServiceLinks: true
{{- end -}}
```

trevorwhitney: this result in a `true` if it's explicitly `true` or omitted

trevorwhitney (commented):
After reading a bit more into those Knative issue you linked (thank you) I'm open to this change. I have one small tweak to your template function I'd like to see, then LGTM!

production/helm/loki/templates/_helpers.tpl
TAYTS: This won't work because if the value is omitted, it is treated as false

production/helm/loki/templates/_helpers.tpl
TAYTS: That was helm v2, let me change to `KubeVersion.Version` instead

production/helm/loki/templates/_helpers.tpl
TAYTS: I made the change to `{{- if or (.Values.loki.enableServiceLinks) (ne .Values.loki.enableServiceLinks false) -}}` so instead of using `toString` will check if it is false, this way will decouple with go templating implementation, hope this is ok for you

trevorwhitney: I'm not following. My understanding was we want `enableServiceLinks` to be `true` if it's either explicitly set to `true` or if it's omitted. The above should work since we set the default to `true`, so if omitted it should use the default value. I checked out your branch, and made only the change above, and these are my results.

`helm template . --set loki.enableServiceLinks=false --show-only templates/gateway/deployment-gateway.yaml`:
```
---
# Source: loki/templates/gateway/deployment-gateway.yaml
...
    spec:
      serviceAccountName: loki
      enableServiceLinks: false
...
```

`helm template . --set loki.enableServiceLinks=true --show-only templates/gateway/deployment-gateway.yaml`:
```
---
# Source: loki/templates/gateway/deployment-gateway.yaml
...
    spec:
      serviceAccountName: loki
      enableServiceLinks: true
...
```

`helm template . --show-only templates/gateway/deployment-gateway.yaml`
```
---
# Source: loki/templates/gateway/deployment-gateway.yaml
...
    spec:
      serviceAccountName: loki
      enableServiceLinks: true
...
```

Sorry for the confusion, but is this not the correct behavior?

production/helm/loki/Chart.yaml
```diff
@@ -4,7 +4,7 @@ name: loki
 description: Helm chart for Grafana Loki in simple, scalable mode
 type: application
 appVersion: 2.6.1
-version: 3.2.2
+version: 3.2.3
```
trevorwhitney, line 7: sorry, just slipped in `3.3.0` under this, need to bump this to `3.3.1` now

production/helm/loki/templates/_helpers.tpl
TAYTS: This second condition is to cover the case where the "enableServiceLinks" is missing in the `values.yaml`

production/helm/loki/Chart.yaml
TAYTS: Updated the version

production/helm/loki/templates/_helpers.tpl
trevorwhitney: If it's missing in `values.yaml` I don't understand why it wouldn't get the default value?

jeschkies (approved):
Thanks. Could you generate the documentation reference?

jeschkies:
@TAYTS could you merge the current main branch? It will fix the build error.

PR #7130, opened 2022-09-12
tsdb: boltdb shipper backup store tsdb

sandeepsukhani:
Since we are planning to start deploying `tsdb` to critical cells, we want to have a disaster recovery without any data loss.
This PR adds support to `tsdb` store itself to configure `boltdb-shipper` as a backup chunk writer.
When the flag is enabled, `tsdb` store would also index chunks in `boltdb-shipper` index store with table names set to `<configured-tsdb-table-prefix>`+`backup_`+`<table-number>`.

This is the simplest option that I can come up with without maintaining a lot of code and we can simply drop the support before marking `tsdb` as production ready.

The rollback/recovery plan would just be to change the schema config to use `boltdb-shipper` as index store with `<configured-tsdb-table-prefix>`+`backup_` as prefix like below:

**tsdb enabled:**
```
schema_config:
  configs:
    - from: 2022-01-01
      store: boltdb-shipper
      index:
        prefix: index_
    - from: 2022-09-10
      store: tsdb
      index:
        prefix: index_
```

**rollback:**
```
schema_config:
  configs:
    - from: 2022-01-01
      store: boltdb-shipper
      index:
        prefix: index_
    - from: 2022-09-10
      store: boltdb-shipper
      index:
        prefix: index_backup_
```

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/tsdb/chunkwriter.go
+++ b/pkg/storage/stores/tsdb/chunkwriter.go
@@ -12,6 +12,7 @@ import (
 	"github.com/grafana/loki/pkg/storage/chunk"
 	"github.com/grafana/loki/pkg/storage/chunk/fetcher"
 	"github.com/grafana/loki/pkg/storage/config"
+	"github.com/grafana/loki/pkg/storage/stores"
 	"github.com/grafana/loki/pkg/storage/stores/tsdb/index"
 	"github.com/grafana/loki/pkg/util/spanlogger"
 )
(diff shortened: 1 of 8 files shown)


pkg/storage/stores/tsdb/chunkwriter.go
```diff
@@ -45,6 +52,9 @@ func (w *ChunkWriter) Put(ctx context.Context, chunks []chunk.Chunk) error {
 		if err := w.PutOne(ctx, chunk.From, chunk.Through, chunk); err != nil {
 			return err
 		}
+		if err := w.backupChunkWriter.PutOne(ctx, chunk.From, chunk.Through, chunk); err != nil {
```
ashwanthgoli, line 55: not sure if I missing something here. Do we need to call `PutOne` on backupChunkWriter here given that it's already being called within `PutOne` method?

sandeepsukhani: Ahh, right. Not sure how I missed that 🤦

ashwanthgoli:
just for my understanding, do we also intend to backup chunks? If no, does the fetcher take care of deduplicating chunk writes from these two writers?
https://github.com/grafana/loki/blob/e39ea21e3cacda92661d1d77770c7d1486ec4124/pkg/storage/stores/series/series_store_write.go#L85

sandeepsukhani:
> just for my understanding, do we also intend to backup chunks? If no, does the fetcher take care of deduplicating chunk writes from these two writers?
> 
> https://github.com/grafana/loki/blob/e39ea21e3cacda92661d1d77770c7d1486ec4124/pkg/storage/stores/series/series_store_write.go#L85

@ashwanthgoli yes, fetcher should take care of deduping the chunks using the chunks cache. If the chunk is there in the cache, it means it has been flushed already so it skips uploading it again to the object storage.

owen-d (approved):

PR #7146, opened 2022-09-13
Add local k3d environment for enterprise-logs

trevorwhitney:

This adds a local k3d based development cluster for testing out the enterprise logs side of the helm chart, which can be enabled by setting:

```
enterprise:
  enabled: true
```

It requires both the `gex-plugins` and `enterprise-logs` repos to be checked out locally. In the future I would like to remedy this by publishing an open source provisioner and providing an alternative way to install the GEL plugin, but neither of those are possible at this moment, and we still need to be able to run tests against the enterprise configuration for the helm chart.

Instructions for use:
* make sure `gex-plugins` and `enterprise-logs` are sibling directories to wherever `loki` is checked out, or set the necessary environment variables (check the `Makefile` at `tools/dev/k3d/Makefile`)
* cd into `tools/dev/k3d` in this repo
* run `make enterprise-logs`


tools/dev/k3d/Makefile
```diff
@@ -28,21 +38,24 @@ create-registry:
...
-			--wait; \
-	else \
-		echo "Cluster already exists"; \
-	fi
+secrets: secrets/grafana.jwt secrets/gel.jwt
 
-prepare: create-registry create-cluster update-repos build-latest-image
+secrets/grafana.jwt:
+	mkdir -p secrets/
+	op document get "loki/grafana.jwt" > $(CURDIR)/secrets/grafana.jwt
```
MichelHollands, line 45: Can you add a note or a check somewhere that this is the command line version of 1Password? There is a op tool in Linux (https://linux.die.net/man/1/op) as well.

MichelHollands (approved):
LGTM

tools/dev/k3d/Makefile
trevorwhitney: Ahh, I didn't know about that. Yes, good call, will do! And for anyone reading this, those secrets are in the shared grafana 1password.

PR #7147, opened 2022-09-13
Fix minio values, reduce replicas, and fix tokengen job

trevorwhitney:

This PR fixes a few small things with the helm chart:

* After upgrading to the new minio chart, the subvalues for that chart were not correctly changed to the new values for `rootUser` and `rootPassword`
* The default replica count for the new minio chart is 16, which seems a little much for our needs.
* The `grafana-token` created by the tokengen job was meant to be used when configuring the GEL plugin, but it was not working, so I removed it.


MichelHollands (approved):
LGTM

PR #7149, opened 2022-09-14
Store level latency metric

sandeepsukhani:
In Reads/Writes dashboard, we have panels for plotting index QPS and latency, but they are specific to `boltdb-shipper` and `bigtable`. Each index store exposes its metric, but we can't have all the supported store types in those dashboards.
We have added experimental support for `tsdb` which is going to be the one we will recommend when it becomes production ready so it is going to be yet another store which a lot of people would be interested in monitoring.

To avoid having latency and qps panel for each index type, I added `loki_index_request_duration_seconds` metric in PR #6880 but it was only measuring index latencies. In this PR I am renaming it to `loki_store_request_duration_seconds` to measure overall store request latencies including `PUT` chunks call.

I have added this metric in a separate panel to Reads / Writes dashboards. We should drop store-specific panels in the next major release. Renaming this metric should not be a problem since we have not done a release since I added it.

### DIFF
--- a/pkg/storage/stores/composite_store.go
+++ b/pkg/storage/stores/composite_store.go
@@ -73,9 +73,7 @@ func (c *CompositeStore) Stores() []Store {
 
 func (c compositeStore) Put(ctx context.Context, chunks []chunk.Chunk) error {
 	for _, chunk := range chunks {
-		err := c.forStores(ctx, chunk.From, chunk.Through, func(innerCtx context.Context, from, through model.Time, store Store) error {
-			return store.PutOne(innerCtx, from, through, chunk)
-		})
+		err := c.PutOne(ctx, chunk.From, chunk.Through, chunk)
 		if err != nil {
 			return err
 		}
@@ -85,7 +83,9 @@ func (c compositeStore) Put(ctx context.Context, chunks []chunk.Chunk) error {
 
 func (c compositeStore) PutOne(ctx context.Context, from, through model.Time, chunk chunk.Chunk) error {
 	return c.forStores(ctx, from, through, func(innerCtx context.Context, from, through model.Time, store Store) error {
-		return store.PutOne(innerCtx, from, through, chunk)
+		return instrument.CollectedRequest(ctx, "put_chunk", instrument.NewHistogramCollector(c.metrics.storeRequestLatency), instrument.ErrorCode, func(ctx context.Context) error {
+			return store.PutOne(innerCtx, from, through, chunk)
+		})
 	})
 }
 
@@ -97,7 +97,7 @@ func (c compositeStore) SetChunkFilterer(chunkFilter chunk.RequestChunkFilterer)
 
 func (c compositeStore) GetSeries(ctx context.Context, userID string, from, through model.Time, matchers ...*labels.Matcher) ([]labels.Labels, error) {
 	var lbls []labels.Labels
-	if err := instrument.CollectedRequest(ctx, "series", instrument.NewHistogramCollector(c.metrics.indexQueryLatency), instrument.ErrorCode, func(ctx context.Context) error {
+	if err := instrument.CollectedRequest(ctx, "series", instrument.NewHistogramCollector(c.metrics.storeRequestLatency), instrument.ErrorCode, func(ctx context.Context) error {
 		var err error
 		lbls, err = c.getSeries(ctx, userID, from, through, matchers...)
 		return err
@@ -133,7 +133,7 @@ func (c compositeStore) getSeries(ctx context.Context, userID string, from, thro
 // LabelValuesForMetricName retrieves all label values for a single label name and metric name.
 func (c compositeStore) LabelValuesForMetricName(ctx context.Context, userID string, from, through model.Time, metricName string, labelName string, matchers ...*labels.Matcher) ([]string, error) {
 	var values []string
-	if err := instrument.CollectedRequest(ctx, "label_values", instrument.NewHistogramCollector(c.metrics.indexQueryLatency), instrument.ErrorCode, func(ctx context.Context) error {
+	if err := instrument.CollectedRequest(ctx, "label_values", instrument.NewHistogramCollector(c.metrics.storeRequestLatency), instrument.ErrorCode, func(ctx context.Context) error {
 		var err error
 		values, err = c.labelValuesForMetricName(ctx, userID, from, through, metricName, labelName, matchers...)
 		return err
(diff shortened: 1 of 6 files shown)


pkg/storage/stores/composite_store.go
```diff
@@ -73,9 +73,7 @@ func (c *CompositeStore) Stores() []Store {
 
 func (c compositeStore) Put(ctx context.Context, chunks []chunk.Chunk) error {
 	for _, chunk := range chunks {
-		err := c.forStores(ctx, chunk.From, chunk.Through, func(innerCtx context.Context, from, through model.Time, store Store) error {
-			return store.PutOne(innerCtx, from, through, chunk)
-		})
+		err := c.PutOne(ctx, chunk.From, chunk.Through, chunk)
```
dannykopping, line 76: This seems unrelated to the purpose of the PR; some drive-by refactoring?
Wondering why we looped over all stores before.

pkg/storage/stores/metrics.go
```diff
@@ -6,16 +6,15 @@ import (
...
 }
 
 func newMetrics(reg prometheus.Registerer) *metrics {
 	return &metrics{
-		indexQueryLatency: promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+		storeRequestLatency: promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
 			Namespace: "loki",
-			Name:      "index_request_duration_seconds",
-			Help:      "Time (in seconds) spent in serving index query requests",
+			Name:      "store_request_duration_seconds",
```
dannykopping, line 16: The changes from #6880 never made it into a release, correct?

dannykopping: Is this metric only for index stores?

production/loki-mixin-compiled/dashboards/loki-reads.json
```diff
@@ -704,6 +704,194 @@
...
+                  "points": false,
+                  "renderer": "flot",
+                  "seriesOverrides": [ ],
+                  "spaceLength": 10,
+                  "span": 6,
+                  "stack": true,
+                  "steppedLine": false,
+                  "targets": [
+                     {
+                        "expr": "sum by (status) (\n  label_replace(label_replace(rate(loki_boltdb_shipper_request_duration_seconds_count{cluster=~\"$cluster\",job=~\"($namespace)/(querier|index-gateway)\", operation=\"Shipper.Query\"}[$__rate_interval]),\n  \"status\", \"${1}xx\", \"status_code\", \"([0-9])..\"),\n  \"status\", \"${1}\", \"status_code\", \"([a-z]+)\"))\n",
```
dannykopping, line 821: Should we still be using `loki_boltdb_shipper_request_duration_seconds_count` here?

production/loki-mixin-compiled/dashboards/loki-writes.json
```diff
@@ -516,6 +516,194 @@
...
+                  "points": false,
+                  "renderer": "flot",
+                  "seriesOverrides": [ ],
+                  "spaceLength": 10,
+                  "span": 6,
+                  "stack": true,
+                  "steppedLine": false,
+                  "targets": [
+                     {
+                        "expr": "sum by (status) (\n  label_replace(label_replace(rate(loki_boltdb_shipper_request_duration_seconds_count{cluster=~\"$cluster\",job=~\"($namespace)/ingester\", operation=\"WRITE\"}[$__rate_interval]),\n  \"status\", \"${1}xx\", \"status_code\", \"([0-9])..\"),\n  \"status\", \"${1}\", \"status_code\", \"([a-z]+)\"))\n",
```
dannykopping, line 633: Ditto

dannykopping (commented):
LGTM, but if this metric is just for _index_ stores then I'd suggest we change the name to not confuse users (like myself initially) that may think this metric measures _all_ store requests, including chunk stores.

pkg/storage/stores/composite_store.go
sandeepsukhani: The [PutOne](https://github.com/grafana/loki/pull/7149/files/1cffed0ce83de39b74ab5d3873ae1031d3d98d9e#diff-099c12d62a2a013f805bf56cc21ce6af89456588cb905552b018def958512cb0R84) call below loops through all the stores. I just tried to reuse the code.

pkg/storage/stores/metrics.go
sandeepsukhani: > The changes from #6880 never made it into a release, correct?
yes, latest release `2.6.1` doesn't have it.

> Is this metric only for index stores?
All the read requests here are purely index queries, while the write request i.e `Put` and `PutOne`, uploads the chunk and indexes it as well. It would be hard to purely track just the index write request without refactoring the code.

production/loki-mixin-compiled/dashboards/loki-reads.json
sandeepsukhani: It is the existing `BoltDB Shipper` panel that I have not touched. Since the new metric won't be available to most users until we do a release or two, I think we should keep it as is.

sandeepsukhani:
> LGTM, but if this metric is just for _index_ stores then I'd suggest we change the name to not confuse users (like myself initially) that may think this metric measures _all_ store requests, including chunk stores.

I can't think of a better name. We will have to refactor the code to measure just the index latency on the chunk flush operations, which would let us rename the metric to `loki_index_request_duration_seconds` latencies. I think it might be tricky to do it, so I can give it a try in a separate PR.

pkg/storage/stores/metrics.go
dannykopping: Yeah I think having a metric named `loki_store_request_duration_seconds` that doesn't include chunk store reads but only includes writes will be quite confusing.

How hard do you anticipate the refactoring would be? I want to try optimise for clarity as much as possible given that our system is already very complex.

pkg/storage/stores/composite_store.go
dannykopping: Ah! I see, thanks.

pkg/storage/stores/metrics.go
sandeepsukhani: I looked at the code, and it seems quite complex and would take time. The other option would be to keep the old metric `loki_index_request_duration_seconds` which tracks just the index requests for now, and update just the reads dashboards until we refactor the code. I will keep the refactoring work on my to-do list until then.

dannykopping: >  The other option would be to keep the old metric loki_index_request_duration_seconds which tracks just the index requests for now, and update just the reads dashboards until we refactor the code. I will keep the refactoring work on my to-do list until then.

Thanks Sandeep, I think that might work out better for our users - if you don't mind?

sandeepsukhani:
Closing this in favour of #7163

PR #7162, opened 2022-09-14
[helm] Storage: Read support for Azure Object Storage

### LINKED ISSUE #7090: [Helm] 3.0.0 Chart does not work when using Azure as storage backend
**Describe the bug**

I used to deploy Loki using the `grafana/loki-simple-scalable` Helm chart. I learned that it is deprecated, and the functionality now exists on the `grafana/loki` Helm chart. The latter chart works if you use AWS S3 (or MinIO) or Google GCS, but it doesn't when using Azure.

**To Reproduce**

Modify the values file to use Azure as a Block Storage backend. As the Helm chart itself doesn't correctly support Azure, here is how I do it:

```yaml
---
gateway:
  enabled: false

monitoring:
  selfMonitoring:
    enabled: false
    grafanaAgent:
      installOperator: false
  serviceMonitor:
    labels:
      release: monitor
  rules:
    labels:
      release: monitor
  alerts:
    labels:
      release: monitor
  dashboards:
    labels:
      release: monitor
      grafana_dashboard: "1"

write:
  replicas: 2
  extraArgs:
  - -config.expand-env=true
  extraEnv:
  - name: AZ_STORAGE_ACCOUNT_NAME
    valueFrom:
      secretKeyRef:
        name: azure-storage
        key: AZ_STORAGE_ACCOUNT_NAME
  - name: AZ_STORAGE_ACCOUNT_KEY
    valueFrom:
      secretKeyRef:
        name: azure-storage
        key: AZ_STORAGE_ACCOUNT_KEY

read:
  replicas: 2
  extraArgs:
  - -config.expand-env=true
  extraEnv:
  - name: AZ_STORAGE_ACCOUNT_NAME
    valueFrom:
      secretKeyRef:
        name: azure-storage
        key: AZ_STORAGE_ACCOUNT_NAME
  - name: AZ_STORAGE_ACCOUNT_KEY
    valueFrom:
      secretKeyRef:
        name: azure-storage
        key: AZ_STORAGE_ACCOUNT_KEY

loki:
  auth_enabled: false
  storage:
    type: azure
  commonConfig: null
  structuredConfig:
    ruler:
      alertmanager_url: http://monitor-alertmanager:9093
      enable_sharding: true
      storage:
        type: azure
        azure:
          container_name: loki-ruler
    common:
      path_prefix: /var/loki
      replication_factor: 1
      storage:
        azure:
          account_name: ${AZ_STORAGE_ACCOUNT_NAME}
          account_key: ${AZ_STORAGE_ACCOUNT_KEY}
          container_name: loki-data
          request_timeout: 0
  limits_config:
    ingestion_rate_mb: 10
  storage_config:
    boltdb_shipper:
      active_index_directory: /var/loki/boltdb-shipper-active
      cache_location: /var/loki/boltdb-shipper-cache
      cache_ttl: 24h
      shared_store: azure
  schemaConfig:
    configs:
    - from: '2022-08-01'
      store: boltdb-shipper
      object_store: azure
      schema: v12
      index:
        period: 24h
        prefix: loki_index_
```

The above works perfectly fine with `grafana/loki-simple-scalable` but not with `grafana/loki`, and I believe it is because of the following:

https://github.com/grafana/loki/blob/main/production/helm/loki/templates/_helpers.tpl#L274-L276

Which affects:

https://github.com/grafana/loki/blob/main/production/helm/loki/templates/_helpers.tpl#L28-L30

Therefore, it doesn't create the read and write StatefulSets.

**Expected behavior**

Despite still being experimental, I've been using the Azure backend, which seems to be working fine. According to the documentation, it is a supported backend, so I would expect the Helm chart to allow it.

**Environment:**

 - Infrastructure: AKS
 - Deployment tool: Helm

N/A

roelofb:
Using Azure Object Storage worked before Helm chart version 3.0.

Fixes #7090

### DIFF
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -299,7 +299,7 @@ Create the service endpoint including port for MinIO.
 
 {{/* Determine if deployment is using object storage */}}
 {{- define "loki.isUsingObjectStorage" -}}
-{{- or (eq .Values.loki.storage.type "gcs") (eq .Values.loki.storage.type "s3") -}}
+{{- or (eq .Values.loki.storage.type "gcs") (eq .Values.loki.storage.type "s3") (eq .Values.loki.storage.type "azure") -}}
 {{- end -}}
 
 {{/* Configure the correct name for the memberlist service */}}


trevorwhitney (approved):
LGTM, thanks for the contribution!

luddskunk:
@trevorwhitney The build seems to fail with the 3.0.7 version. Could you help? I'm not sure how I bump it correctly. Do we re-run the job?

luddskunk:
https://github.com/grafana/loki/actions/runs/3093882585

trevorwhitney:
correct, that version already exists so it would have required a version bump. Since we always release off master, however, this change has likely been included in a newer release since then.

PR #7172, opened 2022-09-15
ingester_querier: early exit queries on reaching quorum

ashwanthgoli:
Currently `forGivenIngesters` and its derivatives don't allow passing down context to the work function as a result of which we are not making use of the tracking functionality provided by `replicationSet.Do` to exit early and relieve pressure on the ingesters.

This change ensures that Unary rpc calls receive ctx populated by the replication set tracker to exit early on reaching quorum or max failures. 

The same cannot be applied to stream rpc calls given how the code is structured at the moment where we are processing the streams outside of the work func

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


ashwanthgoli:
> 

Thanks danny! will do

cyriltovena (approved):
LGTM

dannykopping (approved):
This is really excellent work, thanks @ashwanthgoli!

PR #7173, opened 2022-09-15
[helm] Add Loki Canary to Helm Chart

### LINKED ISSUE #7019: Add loki canary to helm chart
When I install Loki via the helm chart, how do I know it's working?

Include the loki-canary helm chart with the loki helm chart.

Document installing both charts.

**Additional context**
Adding the canary allows us to add a test that uses it as well.

trevorwhitney:

This PR adds the Loki canary as well as the GEL provisioner to the Loki helm chart. The Loki canary expands the self-monitoring functionality of the Helm chart, giving operators insight beyond whether Loki is just running, but also if it's functioning properly.

The GEL provisioner was necessary for the canary to work when the enterprise option (`enterprise.enabled = true`) is enabled. Since enabling enterprise in the helm chart also enables auth/multi-tenancy, the addition of the provisioner allows the helm chart to automatically create a tenant for the canary via the GEL admin API, and to create applicable access policies and tokens. Furthermore, additional tenants can also be specified, which will be provisioned, creating a read and write token for each that will be stored in k8s secrets.

Fixes #7019 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [X] Is this an important fix or new feature? Add an entry in the `CHANGELOG.md`.

### DIFF
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -333,7 +333,7 @@ Create the service endpoint including port for MinIO.
 {{- define "loki.host" -}}
 {{- $isSingleBinary := eq (include "loki.deployment.isSingleBinary" .) "true" -}}
 {{- $url := printf "%s.%s.svc.%s" (include "loki.gatewayFullname" .) .Release.Namespace .Values.global.clusterDomain }}
-{{- if $isSingleBinary  }}
+{{- if and $isSingleBinary (not .Values.gateway.enabled)  }}
   {{- $url = printf "%s.%s.svc.%s:3100" (include "loki.singleBinaryFullname" .) .Release.Namespace .Values.global.clusterDomain }}
 {{- end }}
 {{- printf "%s" $url -}}
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -128,7 +128,6 @@ Docker image name for enterprise logs
 {{- define "loki.enterpriseImage" -}}
 {{- $dict := dict "service" .Values.enterprise.image "global" .Values.global.image "defaultVersion" .Values.enterprise.version -}}
 {{- include "loki.baseImage" $dict -}}
-{{/* {{- printf "foo" -}} */}}
 {{- end -}}
 
 {{/*
@@ -136,6 +135,14 @@ Docker image name
 */}}
 {{- define "loki.image" -}}
 {{- if .Values.enterprise.enabled -}}{{- include "loki.enterpriseImage" . -}}{{- else -}}{{- include "loki.lokiImage" . -}}{{- end -}}
+{{- end -}}
+
+{{/*
+Docker image name for kubectl container
+*/}}
+{{- define "loki.kubectlImage" -}}
+{{- $dict := dict "service" .Values.kubectlImage "global" .Values.global.image "defaultVersion" "latest" -}}
+{{- include "loki.baseImage" $dict -}}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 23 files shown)


production/helm/loki/templates/monitoring/_helpers-monitoring.tpl
```diff
@@ -11,7 +11,19 @@ Client definition for LogsInstance
 {{- end }}
 - url: {{ $url }}
   externalLabels:
-    cluster: {{ include "loki.fullname" . -}}
+    cluster: {{ include "loki.fullname" . }}
+  {{- if or .Values.loki.auth_enabled .Values.enterprise.enabled }}
+  tenantId: {{ .Values.monitoring.selfMonitoring.tenant }}
```
DylanGuedes, line 16: is this identation right?

DylanGuedes (approved):
few nits

production/helm/loki/templates/monitoring/_helpers-monitoring.tpl
jeschkies: Why is a tenant ID required for enterprise?

jeschkies (commented):
I need to give it one more pass.

production/helm/loki/templates/_helpers.tpl
```diff
@@ -322,3 +329,33 @@ Create the service endpoint including port for MinIO.
 {{ include "loki.name" . }}-memberlist
 {{- end -}}
 
+{{/* Determine the public host for the Loki cluster */}}
+{{- define "loki.host" -}}
+{{- $isSingleBinary := eq (include "loki.deployment.isSingleBinary" .) "true" -}}
+{{- $url := printf "%s.%s.svc.%s" (include "loki.gatewayFullname" .) .Release.Namespace .Values.global.clusterDomain }}
+{{- if $isSingleBinary  }}
+  {{- $url = printf "%s.%s.svc.%s:3100" (include "loki.singleBinaryFullname" .) .Release.Namespace .Values.global.clusterDomain }}
```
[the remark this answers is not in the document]
trevorwhitney: so I think the reasoning here is why would someone be using a gateway when running monolithic mode, since monolithic assumes a single instance (filesystem storage cannot support > 1 replica)? Maybe we should be more explicit about that requirement though, both in the check here and a validation file?

production/helm/loki/templates/monitoring/_helpers-monitoring.tpl
trevorwhitney: probably not required for enterprise since with enterprise we're providing the username below.

trevorwhitney: changed the logic a bit, and yes double checked indenation, following with with `auth_enabled = true` but enterprise off:
```
spec:
  clients:
    
    - url: http://loki-gateway.default.svc.cluster.local/loki/api/v1/push
      externalLabels:
        cluster: release-name-loki
      tenantId: self-monitoring
```

production/helm/loki/templates/_helpers.tpl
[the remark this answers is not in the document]
DylanGuedes: hmm we could be more explicit definetely, but what you said is fair enough. Resolved!

towolf:
Hi @trevorwhitney, you you explain, why you added the loki-canary ServiceAccount as a Helm `post-install` hook? This fails to work with Helm via Terraform (https://github.com/hashicorp/terraform-provider-helm/issues/683), and I see no reason for this.

https://github.com/grafana/loki/blob/main/production/helm/loki/templates/loki-canary/serviceaccount.yaml#L10

towolf:
And another question @trevorwhitney, why do you default to `latest` and not to `.Chart.Appversion` here?

https://github.com/grafana/loki/blob/main/production/helm/loki/templates/loki-canary/_helpers.tpl#L28

trevorwhitney:
@towolf the reason for the `post-install` hook was probably because I copied the service account template from either the tokengen or provisioner jobs, and my guess for those being `post-install` hooks are to delay the running of the job. But I agree, we probably don't need to delay anything for the canary so we can remove that.

Also, we should definitely use `.Char.Appversion` here, agreed!

trevorwhitney:
@towolf here's a PR to fix, thanks for pointing these out!
https://github.com/grafana/loki/pull/7236

PR #7176, opened 2022-09-15
Implement ingester API to return stream rates

MasslessParticle:
This PR introduces an API on the Ingester that reports all current streams in memory and their current rate in bytes/second. The reponse from the API contains a slice of `StreamRate` structs:

```go
type StreamRate struct {
	StreamHash        uint64 `protobuf:"varint,1,opt,name=streamHash,proto3" json:"streamHash,omitempty"`
	StreamHashNoShard uint64 `protobuf:"varint,2,opt,name=streamHashNoShard,proto3" json:"streamHashNoShard,omitempty"`
	Rate              int64  `protobuf:"varint,3,opt,name=rate,proto3" json:"rate,omitempty"`
}
```

In the interest of space on the caller and in transmission, only hashes and rates are sent.
- `StreamHash` will be used to deduplicate replicas
- `StreamHashNoShard` will be used to aggregate a sharded stream's total rate.

Rate calculation will be in a follow on PR


pkg/ingester/stream.go
```diff
@@ -387,6 +397,7 @@ func (s *stream) validateEntries(entries []logproto.Entry, isReplay bool) ([]log
 		}
 	}
 
+	s.rateCalculator.Record(int64(totalBytes))
```
jeschkies, line 400: Why is this not using `validBytes`?

pkg/ingester/rate_calculator.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return c
+}
+
+func (c *RateCalculator) recordSample() {
+	t := time.NewTicker(1000 / bucketCount * time.Millisecond)
+	for range t.C {
+		rate := c.sample.Swap(0)
+		c.storeNewRate(rate)
+	}
```
jeschkies, line 36: Is this never stopped?

jeschkies (commented):
Interesting. I've always wondered how rate limiting was implemented. I wish the algorithm would expose the actual rate :/

pkg/ingester/instance_test.go
```diff
@@ -135,6 +137,87 @@ func TestConcurrentPushes(t *testing.T) {
...
+		return true
+	}, time.Second, time.Millisecond)
+
+	for _, r := range rates.StreamRates {
+		hashNoShard, ok := hashes[r.StreamHash]
+		require.True(t, ok)
+		require.Equal(t, hashNoShard, r.StreamHashNoShard)
+	}
+
+	fmt.Println(rates.StreamRates)
```
DylanGuedes, line 218: I think you forgot to remove this :smile:

pkg/ingester/instance_test.go
```diff
@@ -135,6 +137,87 @@ func TestConcurrentPushes(t *testing.T) {
 	// test passes if no goroutine reports error
 }
 
+func TestStreamRates(t *testing.T) {
+	limits, err := validation.NewOverrides(defaultLimitsTestConfig(), nil)
+	require.NoError(t, err)
+	limiter := NewLimiter(limits, NilMetrics, &ringCountMock{count: 1}, 1)
+
+	inst, err := newInstance(defaultConfig(), defaultPeriodConfigs, "test", limiter, loki_runtime.DefaultTenantConfigs(), noopWAL{}, NilMetrics, &OnceSwitch{}, nil)
+	require.Nil(t, err)
```
DylanGuedes, line 146: nit: i don't have a preference for `require.Nil` or `require.NoError`, but since you used NoError on line 142 maybe you want to use it here too

pkg/ingester/instance_test.go
```diff
@@ -135,6 +137,87 @@ func TestConcurrentPushes(t *testing.T) {
...
+	)
+
+	uniqueLabels := map[string]bool{}
+	startChannel := make(chan struct{})
+	hashes := map[uint64]uint64{}
+
+	var buf []byte
+	wg := sync.WaitGroup{}
+	for i := 0; i < concurrent; i++ {
+		l := makeRandomLabels()
```
DylanGuedes, line 161: could you explain how this block works? (lines 161~165)

pkg/ingester/instance_test.go
```diff
@@ -135,6 +137,87 @@ func TestConcurrentPushes(t *testing.T) {
...
+				}})
+
+				require.NoError(t, err)
+
+				tt = tt.Add(entriesPerIteration * time.Nanosecond)
+			}
+		}(l.String())
+	}
+
+	time.Sleep(100 * time.Millisecond)
```
DylanGuedes, line 190: just to see if I understand this: this sleep holds all iteration go routines to make sure they start at a similar time. Is that it?

pkg/ingester/instance_test.go
```diff
@@ -135,6 +137,87 @@ func TestConcurrentPushes(t *testing.T) {
...
+	require.Eventually(t, func() bool {
+		rates, err = inst.StreamRates(context.Background(), &logproto.StreamRatesRequest{})
+		require.NoError(t, err)
+
+		if len(rates.StreamRates) != 10 {
+			return false
+		}
+
+		for _, r := range rates.StreamRates {
+			if r.Rate != 79000 {
```
DylanGuedes, line 205: how was 79k calculated? maybe you can change this to use the variables you used to calculate it

pkg/ingester/stream.go
MasslessParticle: This preserves the current behavior where stream rate limits are calculated against all the bytes in the stream, not just the valid ones. I assume we do it this way because the rate limit is to protect the ingester and we process every entry, valid or not.

pkg/ingester/rate_calculator.go
```diff
@@ -0,0 +1,61 @@
...
+
+	go c.recordSample()
+
+	return c
+}
+
+func (c *RateCalculator) recordSample() {
+	t := time.NewTicker(1000 / bucketCount * time.Millisecond)
+	defer t.Stop()
```
MasslessParticle, line 33: This isn't going to do anything, unfortunately. We need to add a `Stop` method to the rate limiter than can be called when the stream is removed otherwise this goroutine will hang around forever.

pkg/ingester/instance_test.go
MasslessParticle: 1. We want `concurrent` number of streams happening all at the same time 
2. Each stream needs some unique labels but our uniqueness algorithm isn't great so we throw everything into a map generate another one until it's unique

pkg/ingester/instance_test.go
MasslessParticle: That makes this goroutine sleep until the others are ready It's not strictly necessary

pkg/ingester/instance_test.go
MasslessParticle: I added a comment for this

pkg/ingester/ingester.go
```diff
@@ -620,6 +622,23 @@ func (i *Ingester) Push(ctx context.Context, req *logproto.PushRequest) (*logpro
 	return &logproto.PushResponse{}, err
 }
 
+// GetStreamRates returns a response containing all streams and their current rate
+func (i *Ingester) GetStreamRates(ctx context.Context, req *logproto.StreamRatesRequest) (*logproto.StreamRatesResponse, error) {
```
DylanGuedes, line 626: WDYT of adding a TODO for us to, in the future, add a `sorted=true/false` flag to this call? Viewing the results sorted by streams that have the highest ratings have the potential to help a lot with debugging. But since it isn't the focus of this PR, I vote for it to just be a TODO.

DylanGuedes (approved):
lgtm

just a small observation that shouldn't change anything

pkg/ingester/ingester.go
MasslessParticle: I don't know how useful this is for humans because we're only returning hashes rather than labels but I'm happy to add a TODO to remind us

DylanGuedes: oh I forgot about the hash thing :facepalm: resolving this

PR #7177, opened 2022-09-16
allow backup store metrics registration by including backend type in labels

owen-d:
The refactor [here](https://github.com/grafana/loki/blob/main/pkg/storage/store.go#L237) introduced in https://github.com/grafana/loki/pull/7154 will cause a panic due to duplicate registration via the same store prefix when using a backup index store (for use during tsdb migrations).

This PR adds more granularity to the label value to differentiate these and avoid the problem.


MichelHollands (approved):
LGTM

PR #7187, opened 2022-09-16
TSDB indexBuckets helper logs instead of errors

owen-d:
indexBuckets will return the list of overlapping TSDB buckets for a time range. Historically, this errored when there was no overlap, which is generally impossible. While mutating period configurations as a test (which should never be done in production), I realized this would preemptively error us when there were old TSDB WALs from a bucket that no longer belonged to a TSDB period (i.e. by replacing a `tsdb` index with a `boltdb-shipper` one).

This PR changes this to a log line instead as we don't need to error and stop computation; it's better to return an empty overlap instead.

### DIFF
--- a/pkg/storage/stores/tsdb/manager.go
+++ b/pkg/storage/stores/tsdb/manager.go
@@ -19,6 +19,7 @@ import (
 	"github.com/grafana/loki/pkg/storage/config"
 	"github.com/grafana/loki/pkg/storage/stores/indexshipper"
 	"github.com/grafana/loki/pkg/storage/stores/tsdb/index"
+	util_log "github.com/grafana/loki/pkg/util/log"
 )
 
 // nolint:revive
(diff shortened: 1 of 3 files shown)


pkg/storage/stores/tsdb/manager.go
```diff
@@ -296,7 +294,7 @@ func indexBuckets(from, through model.Time, tableRanges config.TableRanges) (res
 		}
 	}
 	if len(res) == 0 {
-		return nil, fmt.Errorf("could not find config for table(s) fom: %d, through %d", start, end)
+		level.Warn(util_log.Logger).Log("err", "could not find config for table(s) fom: %d, through %d", start, end)
```
DylanGuedes, line 297: nit
```suggestion
		level.Warn(util_log.Logger).Log("err", "could not find config for table(s) from: %d, through %d", start, end)
```

DylanGuedes (approved):
lgtm but a nit there

kavirajk (approved):
LGTM

PR #7196, opened 2022-09-19
Helm: add support for configuring ruler

agologan:

Enables configuring various `ruler` options like alertmanager url.
`rulerConfig` is added second in the template so users can override the `storage` config if necessary.
Have seen questions in community Slack asking how to get filesystem support and alertmanager back when migrating from the old loki helm which should now be possible.

**Which issue(s) this PR fixes**: N/A

**Special notes for your reviewer**: N/A

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/values.yaml
+++ b/production/helm/loki/values.yaml
@@ -121,11 +121,14 @@ loki:
             period: 24h
     {{- end }}
 
-    {{- if or .Values.minio.enabled (eq .Values.loki.storage.type "s3") (eq .Values.loki.storage.type "gcs") }}
+    {{- if .Values.loki.rulerConfig}}
     ruler:
       storage:
+      {{- if or .Values.minio.enabled (eq .Values.loki.storage.type "s3") (eq .Values.loki.storage.type "gcs") }}
       {{- include "loki.rulerStorageConfig" . | nindent 4}}
-    {{- end -}}
+      {{- end }}
+    {{- toYaml .Values.loki.rulerConfig | nindent 2}}
+    {{- end }}
 
     {{- with .Values.loki.memcached.results_cache }}
     query_range:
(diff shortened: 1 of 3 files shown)


production/helm/loki/values.yaml
```diff
@@ -121,11 +121,14 @@ loki:
...
-    {{- if or .Values.minio.enabled (eq .Values.loki.storage.type "s3") (eq .Values.loki.storage.type "gcs") }}
+    {{- if .Values.loki.rulerConfig}}
     ruler:
       storage:
+      {{- if or .Values.minio.enabled (eq .Values.loki.storage.type "s3") (eq .Values.loki.storage.type "gcs") }}
       {{- include "loki.rulerStorageConfig" . | nindent 4}}
-    {{- end -}}
+      {{- end }}
+    {{- toYaml .Values.loki.rulerConfig | nindent 2}}
+    {{- end }}
```
trevorwhitney, line 131: What would you think about something like this, so we'd always include that ruler storage section, and then by default we'd have the `rulerConfig` be an empty object? That would preserve parity with what we currently have. Might need to double check the indention is right in the suggestion.

```suggestion
    ruler:
      storage:
      {{- if or .Values.minio.enabled (eq .Values.loki.storage.type "s3") (eq .Values.loki.storage.type "gcs") }}
      {{- include "loki.rulerStorageConfig" . | nindent 4}}
      {{- end }}
    {{- with .Values.loki.rulerConfig}}
    {{- toYaml . | nindent 2}}
    {{- end }}
```

production/helm/loki/values.yaml
```diff
@@ -221,6 +224,12 @@ loki:
   # -- Check https://grafana.com/docs/loki/latest/configuration/#schema_config for more info on how to configure schemas
   schemaConfig: {}
 
+  # -- Check https://grafana.com/docs/loki/latest/configuration/#ruler for more info on configuring ruler
+  rulerConfig:
+    alertmanager_url: ""
+    enable_alertmanager_v2: false
+    enable_api: true
```
trevorwhitney, line 231:
```suggestion
  rulerConfig: {}
```

trevorwhitney (changes requested):
thanks, great addition, something I think we need given the questions you've seen on slack. what would you think about defaulting the ruler config to an empty object, but then including it after the storage section if provided? I think this might maintain closer parity to what we have now.

production/helm/loki/values.yaml
agologan: Good suggestion, was thinking there may be a better way to approach it myself. 👍

agologan:
Thanks for the suggestion. Wasn't sure of the proposal, but this is better.

agologan:
Same as my other PR updated **README.md.gotmpl** b/c the README was edited manually and will fail docs lint step - not really related to this change.

trevorwhitney (approved):
lgtm, thanks for making that change :smile:

PR #7199, opened 2022-09-20
Loki: Shard streams based on desired rate

DylanGuedes:
Modify sharding to calculate number of shards based on a desired rate and the current ingestion rate. The current ingestion rate is using a stub interface, but shortly, it will use the real ingestion rate fetched from ingesters.

N/A

* This is part of the stream sharding work

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


MasslessParticle (commented):
I think a better approach would be to have `streamsharder` return the current rate from the interface and move the math that's in `ShardCountFor` to a function in `stream.go`

This'll allow us to easily test sharding by mocking the `streamsharder` and means that we won't have to touch anything when we go actually implement the cache.

`streamsharder` should probably also be renamed because it's doing something altogether different now. What do you think of `shardSizeCache`?

pkg/distributor/distributor.go
```diff
@@ -75,10 +79,10 @@ func (cfg *Config) RegisterFlags(fs *flag.FlagSet) {
...
 }
 
-// StreamSharder manages the state necessary to shard streams.
-type StreamSharder interface {
-	ShardCountFor(stream logproto.Stream) (int, bool)
-	IncreaseShardsFor(stream logproto.Stream)
+// RateStore manages the ingestion rate of streams, populated by data fetched from ingesters.
+type RateStore interface {
+	RateFor(stream logproto.Stream) (int, error)
+	StoreRate(stream logproto.Stream, rate int) error
```
MasslessParticle, line 85: We won't need this on the interface. The rate store will take total care of how sharding should work

pkg/distributor/distributor.go
```diff
@@ -92,7 +96,9 @@ type Distributor struct {
 	ingestersRing    ring.ReadRing
 	validator        *Validator
 	pool             *ring_client.Pool
-	streamSharder    StreamSharder
+
+	rateStore RateStore
+	sharder   func(stream logproto.Stream, desiredRate int, rateStore RateStore) (int, error)
```
MasslessParticle, line 101: Would it make more sense to have the `rateStore` return a known rate and just make sure we get the right number of shards out of `shardStream`? That would be a bit simpler and cover stream sharding in one logical test

pkg/distributor/ratestore.go
```diff
@@ -0,0 +1,46 @@
+package distributor
```
MasslessParticle, line 1: This bit of functionality doesn't need any implementation for the rate store. If we want a no op, it could just return the per-stream-rate-limit. We're only using it when stream sharding is enabled so we should probably just have it return 0

pkg/distributor/distributor.go
DylanGuedes: agreed, good catch.

pkg/distributor/distributor.go
DylanGuedes: hmmm but isn't `rateStore` returning a known rate for a given stream? or do you mean like moving from an interface to a func?

>  just make sure we get the right number of shards out of shardStream

the `shardStream` is already tested (implemented by an older PR). So what I only felt the necessity to add tests for the new `calculateShards`, as it relies on things that weren't existent when the `streamStream` tests were written.
I'm planning in adding tests for `shardCountFor` though, WDYT?

pkg/distributor/distributor.go
```diff
@@ -310,11 +317,24 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
...
+			derivedKeys, derivedStreams, err := d.shardStream(stream, userID)
+			if err != nil {
+				level.Error(util_log.Logger).Log("msg", "error sharding stream", "err", err)
+				shardErr = err
+			} else {
+				keys = append(keys, derivedKeys...)
+				streams = append(streams, derivedStreams...)
+			}
+
+			if unshardableErr, ok := err.(*unshardableStreamErr); ok {
```
MasslessParticle, line 331: I was considering this over the weekend. Returning an error is basically the same thing as a per-stream-rate limit error. What do you think about returning just increment a metric that can cause an alert?

MasslessParticle:
This looks great! The last thing we should do is:
- Document `desired_rate` and make sure we've registered a flag for it
- We should probably make it human readable, like the other rates: https://github.com/grafana/loki/blob/74c8cf03ba4fb2abd979a9af05bb945813a4505c/docs/sources/configuration/_index.md?plain=1#L2316-L2319

MasslessParticle (approved):

PR #7203, opened 2022-09-20
highly sharded subqueries are ellided in expr.String

owen-d:
This is mostly a quality of life improvement. Highly sharded subqueries can be burdensome to view and they're logged on every request. This PR will only display the first four subqueries in a `Concat{Log,Sample}Expr`, which is easier on the eyes when viewing logs and is more efficient.


MichelHollands (approved):
LGTM

PR #7209, opened 2022-09-21
avoid using bloomfilters for chunks in stats calls by avoiding duplicates

sandeepsukhani:
Avoid using bloomfilters for chunks deduplication in tsdb `Stats` calls by avoiding fetching duplicate entries.
The idea is to split and align queries by [ObjectStorageIndexRequiredPeriod](https://github.com/grafana/loki/blob/61794710a720da135d8479b7c2f723c740e86404/pkg/storage/config/schema_config.go#L47) and make each split process chunks with a start time >= start time of the table interval.
In other terms, table interval that contains start time of the chunk, owns it.
For e.g. if the table interval is 10s, and we have chunks 5-7, 8-12, 11-13.
Query with range 6-15 would be split into 6-10, 10-15.
query1 would process chunks 5-7, 8-12 and query2 would process chunks 11-13.

This check is not applied for the first split so that we do not eliminate any chunks that overlaps the original query intervals but starts at the previous table.
For e.g. if the table interval is 10s, and we have chunks 5-7, 8-13, 14-13.
Query with range 11-12 should process chunk 8-13 even though its start time <= start time of table we will query for index.

The caveat here is that we will overestimate the data we will be processing if the index is not compacted yet since it could have duplicate chunks when RF > 1. I think it is okay since the Stats call is just an estimation and need not be accurate.

Removing all the extra processing saves us quite a bit of CPU and memory, as seen from the benchmark comparison between the two implementations:
```
name                  old time/op    new time/op    delta
IndexClient_Stats-10     187µs ± 0%      34µs ± 1%  -82.00%  (p=0.008 n=5+5)

name                  old alloc/op   new alloc/op   delta
IndexClient_Stats-10    61.5kB ± 4%    12.5kB ± 2%  -79.69%  (p=0.008 n=5+5)

name                  old allocs/op  new allocs/op  delta
IndexClient_Stats-10     1.46k ± 0%     0.48k ± 0%  -67.28%  (p=0.008 n=5+5)
```

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/index/stats/stats_test.go
+++ b/pkg/storage/stores/index/stats/stats_test.go
@@ -6,8 +6,6 @@ import (
 
 	"github.com/prometheus/common/model"
 	"github.com/stretchr/testify/require"
-
-	"github.com/grafana/loki/pkg/storage/stores/tsdb/index"
 )
 
 func TestStatsBloom_Stream(t *testing.T) {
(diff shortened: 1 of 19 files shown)


pkg/storage/stores/index/stats/stats_test.go
```diff
@@ -24,24 +22,3 @@ func TestStatsBloom_Stream(t *testing.T) {
 
 	require.Equal(t, uint64(2), sb.stats.Streams)
 }
-
-func TestStatsBloom_Chunks(t *testing.T) {
```
sandeepsukhani, line 28: had to drop this test since it is not relevant anymore.

owen-d (approved):
Great. LGTM

PR #7210, opened 2022-09-21
Helm-Chart: Make PVCs in Statefulsets optional

### LINKED ISSUE #7068: [helm/loki-simple-scalable] Are PVCs really required? / Retention not working?
## My Goal
- Run Loki only with Object Storage from MinIO, if it is possible.
- I want to run Loki scalable and easy as possible, that's why I decided to use [loki-simple-scalable](https://github.com/grafana/helm-charts/tree/main/charts/loki-simple-scalable) resp. the new location .

## My Question
- Is it possible to run loki-simple-scalable without PVCs (only S3 from MinIO)? 
  - If yes, how does the configuration look like?
  - If no, why are PVCs required?, Which contents are saved in my PVCs?, Why are my PVCs filling up? 
- Are the retentions configured in the `compactor` and `limits_config` working with S3 or do I need to configure the retention on my MinIO Bucket? As I understand the release notes of [Loki 2.3](https://grafana.com/docs/loki/latest/release-notes/v2-3/) there is no need for external (not loki) retentions (_Loki now has the ability to apply custom retention based on stream selectors! This will allow much finer control over log retention all of which is now handled by Loki, **no longer requiring the use of object store configs for retention**._) -> If yes, how does the configuration look like?
- Just for my interest: What's the difference between configure retention in the `compactor` component and configure retention in `table manager`? In the documentation the following written: _Retention through the Table Manager is achieved by relying on the object store TTL feature, and will work for both boltdb-shipper store and chunk/index store. However retention through the Compactor is supported only with the boltdb-shipper store._ But in the section of the Table Manager it contradicts itself: _When using S3 or GCS, the bucket storing the chunks needs to have the expiry policy set correctly. For more details check S3’s documentation or GCS’s documentation._

## My current configuration
```yaml
gateway:
  replicas: 2
  autoscaling:
    enabled: true
    minReplicas: 2
    maxReplicas: 3
    targetCPUUtilizationPercentage: 60
    targetMemoryUtilizationPercentage: 80
  ingress:
    enabled: true
  resources:
    limits:
      #cpu: 50m
      memory: 30Mi
    requests:
      cpu: 5m
      memory: 15Mi

loki:
  # The value below "config" must be a string instead of a map
  config: |
    auth_enabled: false
    common:
      path_prefix: /var/loki
      storage:
        filesystem: null
        s3:
          insecure: true
          s3: http://USERNAME:PASSWORD@minio.minio.svc.cluster.local:9000/loki
          s3forcepathstyle: true
    storage_config:
      boltdb_shipper:
        active_index_directory: /var/loki/index
        cache_location: /var/loki/cache
        cache_ttl: 24h         # Can be increased for faster performance over longer query periods, uses more disk space
        shared_store: s3
    schema_config:
      configs:
        - from: "2020-09-07"
          store: boltdb-shipper
          object_store: s3
          schema: v12
          index:
            period: 24h
            prefix: index_
    limits_config:
      retention_period: 1h
      enforce_metric_name: false
      reject_old_samples: true
      split_queries_by_interval: 15m
    compactor:
      compaction_interval: 10m
      retention_enabled: true
      retention_delete_delay: 2h
    table_manager:
      retention_deletes_enabled: true
      retention_period: 1h
    chunk_store_config:
      max_look_back_period: 1h
    query_scheduler:
      max_outstanding_requests_per_tenant: 1000
    memberlist:
      join_members:
      - loki-memberlist
    server:
      http_listen_port: 3100
      grpc_listen_port: 9095

write:
  replicas: 2
  autoscaling:
    enabled: true
    minReplicas: 2
    maxReplicas: 3
    targetCPUUtilizationPercentage: 60
    targetMemoryUtilizationPercentage: 80
  persistence:
    size: 5Gi
    storageClass: longhorn
  resources:
    limits:
      #cpu: 400m
      memory: 1200Mi # Don't decrease this limit. Loki uses enough memory to prevent dataloss (when storage backend unavailable)
    requests:
      cpu: 35m
      memory: 300Mi # Don't decrease this limit. Loki can fastly grow up in memory and this can cause OOM-Killing of other pods

read:
  replicas: 2
  autoscaling:
    enabled: true
    minReplicas: 2
    maxReplicas: 3
    targetCPUUtilizationPercentage: 60
    targetMemoryUtilizationPercentage: 80
  persistence:
    size: 1Gi
    storageClass: longhorn
  resources:
    limits:
      #cpu: 400m
      memory: 150Mi
    requests:
      cpu: 50m
      memory: 75Mi
    
serviceMonitor:
  enabled: true


monitoring:
  selfMonitoring:
    enabled: false
    grafanaAgent:
      installOperator: false
```

+ Environment specific Helm values:
```yaml
gateway:
  ingress:
    hosts:
    - host: loki.apps-test-1....
      paths: 
      - path: /
        pathType: ImplementationSpecific
    tls:
    - hosts: 
      - loki.apps-test-1....
      secretName: ingress-wildcard-cert
```

_-> I already opened this issue in the previous [location](https://github.com/grafana/helm-charts/issues/1693) of the Loki-Simple-Scalable Helm Chart, but because the Helm chart moved (thanks to @zanhsieh), I opened a new issue here._

atze234:
Signed-off-by: atze234 <schorsus@gmx.de>

Make PVCs of Statefulsets optional as of the old loki Chart

Fixes #7068 

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/CHANGELOG.md
+++ b/production/helm/loki/CHANGELOG.md
@@ -11,6 +11,9 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
+## 3.0.9
+- [ENHANCEMENT] Make PVC for Statefulsets optional but enabled as default
+  
 ## 3.0.4
 
 - [CHANGE] Default minio replicas to 1 node with 2 drives. The old config used the default, which was 16 nodes with 1 drive each.
--- a/production/helm/loki/Chart.yaml
+++ b/production/helm/loki/Chart.yaml
@@ -3,7 +3,7 @@ name: loki
 description: Helm chart for Grafana Loki in simple, scalable mode
 type: application
 appVersion: 2.6.1
-version: 3.0.8
+version: 3.0.9
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
--- a/production/helm/loki/README.md
+++ b/production/helm/loki/README.md
@@ -330,8 +330,15 @@ monitoring:
 | read.image.repository | string | `nil` | Docker image repository for the read image. Overrides `loki.image.repository` |
 | read.image.tag | string | `nil` | Docker image tag for the read image. Overrides `loki.image.tag` |
 | read.nodeSelector | object | `{}` | Node selector for read pods |
+| read.persistence.enabled | bool | `true` | Enable Persistence for Read Statefulset |
 | read.persistence.size | string | `"10Gi"` | Size of persistent disk |
 | read.persistence.storageClass | string | `nil` | Storage class to be used. If defined, storageClassName: <storageClass>. If set to "-", storageClassName: "", which disables dynamic provisioning. If empty or set to null, no storageClassName spec is set, choosing the default provisioner (gp2 on AWS, standard on GKE, AWS, and OpenStack). |
+| read.persistence.accessModes | list | `[ReadWriteOnce]` | Persistence AccessModes |
+| read.persistence.annotations | object | `{}` | Persistence Annotations |
+| read.persistence.labels | object | `{}` | Persistence Labels |
+| read.persistence.existingClaim | string | `nil` | Name of existing Claim |
+| read.persistence.selector | object | `{}` | PVC Selector |
+| read.persistence.persistentVolumeClaimRetentionPolicy | object | `{}` | Retention Policies for PVC |
 | read.podAnnotations | object | `{}` | Annotations for read pods |
 | read.priorityClassName | string | `nil` | The name of the PriorityClass for read pods |
 | read.replicas | int | `3` | Number of replicas for the read |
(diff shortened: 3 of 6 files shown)


trevorwhitney (changes requested):
Thanks for the addition, though I have a few question before moving forward with this approach.

How are you handling the case of the WAL and the Index Gateway in this configuration? They need local disk. With the Index Gateway, it's just a performance concern, so I can see that being ok if scalability is more important than having to occasionally re-download the indexes. However, we have plans to add a 3rd target to SSD mode that will make the read target a deployment consisting of just Query Frontend and Querier, which I think will solve the scalability concerns.

I'm more concerned with the WAL. Without persistent WAL, how do you handle data recovery in case of an ingester failure/restart?

Maybe we need to turn the WAL off when persistence is disabled for write pods? Though, that too is a bit concerning to me as the helm chart is designed to be the easiest and most opinionated way to run Loki. We publish jsonnet for those wishing to craft more specialized deployments, and I'm not sure I want to expose disabling of the WAL as a side effect of disabling persistence in the "easy" mode.

atze234:
Hi @trevorwhitney i put the persistence.enabled default to true because of the wal and the possibility of data loss. 
When persistence is disabled the wal will be deleted on every restart, but thats not really a Problem for a test cluster where i just want to have an easy, low-cost loki running.
So i ported this non-pvc-mode over from old chart.

Looking at that this way i‘d say an Installation without PVC would be easier/lightweight as well, because in an easy mode installation i dont want to have pvc‘s. Also i try to avoid them as much as possible.

I agree that for the „read“ a deployment would fit better then, but it makes the installation a bit more complicated or confusing and the chart as well.

trevorwhitney:
Can you describe your "test cluster" a little better? The main goal of this chart is an easy, opinionated, and production ready installation. There is already a lot of feedback that Loki has too many configuration options, and can be difficult to get started with. Thus, it is the goal of this helm chart to reduce configuration as much as possible.

We publish jsonnet for people needing highly configurable deployments, and I wonder if that would be a better candidate for your "test cluster"? Or perhaps you could use helm [post rendering](https://helm.sh/docs/topics/advanced/#post-rendering) with something like jsonnet or kustomize to achieve this?

I'm hesitant to expand the config surface area of the chart for a test use case that most people will not use, and that we would never recommend for production.

atze234:
Ok i understand that. I just wanted to make the pvc configureable for any environment where wal dataloss would be ok. for example a cluster for staging area for applications, where i just want to test applications. in such environments i generally try to avoid pvc to save some money and/or resources.
As the old chart supported to disable the PVC i thought it would be ok here too, but i see that the persistent WAL is part of loki and its mandatory to have it persistent. From this point of view it was a mistake that the PVC could be disabled in the old chart and so its ok to not make it possible to opt-out.

trevorwhitney:
@atze234 thanks for your understanding. One thing we're trying to do with moving the chart into the Loki repo is to take a more involved (from the Loki team) and more opinionated approach, to help people get up and running with production ready clusters faster, so we may need to make a few departures from the old chart in support of that goal.

Issue #7142, opened 2022-09-13
Bug(loki-canary): Race between updating sent `totalEntries` and `entryReceived`


kavirajk:
**Describe the bug**
It all started when using new `push` mode in loki-canary (which directly pushes logs to Loki, instead of writing to local file and let promtail/grafana-agent to scrape and push it to Loki)

Noticed that some logs are missing by websocket tailing. More details on how it works [here](https://grafana.com/docs/loki/latest/operations/loki-canary/)

![Screenshot_2022-09-13_11-48-52](https://user-images.githubusercontent.com/3735252/189870445-da7346af-70f2-4d3a-b7bd-f3076578086e.png)

which is calculated from following promql.
```
(sum(increase(loki_canary_websocket_missing_entries_total{cluster=~"<cluster>",namespace=~"<loki-namespace>"}[$__range]))/sum(increase(loki_canary_entries_total{cluster=~"<cluster>",namespace=~"<loki-namespace>"}[$__range])))*100
```

### Problem
Loki canary opens websocket at the background to listen for log messages that are previously sent. It keeps track of already sent entries in `totalEntries` slice. it waits for `1m` (by default) to make sure if all the messages that are sent before, is received via websocket. 

If it doesn't it increment the metric `websocket_missing_entries`. And finally it makes direct query to loki to check if those entries are really missing, if not found it increments `missing_entries` counter. **NOTE: even if it receives the previously missing entry via direct query, it doesn't update the previous `websocket_missing_entries`**.

So I see following error describing websocket is missing some log entries.
```
websocket failed to receive entry 1663058352152669300 within 60.000000 seconds
```
but then, if you go back in time, that entry is indeed received, but canary unable to recognize it, because it doesn't have that entries in sent `totalEntries` yet.
```
received an unexpected entry with ts 1663058352152669300
```

Also you see that exact entry is received when queried directly.
```
missing websocket entry 1663058352152669300 was found 65.00036012 seconds after it was originally sent
```

The problem looks like caused by race between `sent` and `recv` channel in comparator `run()` [loop](https://github.com/grafana/loki/blob/main/pkg/canary/comparator/comparator.go#L264-L267).

So when loop is busy with `c.entryReceived` and meanwhile entry is sent and didn't get a chance to update it's sent list ([because receive loop also holding mutex to that list](https://github.com/grafana/loki/blob/main/pkg/canary/comparator/comparator.go#L205-L206))

[...]

PR #7211, opened 2022-09-21
fix(loki-canary): Send to Loki after updating `totalEntries`.

kavirajk:
Fixes: #7142 (take a look for more details about the problem there)

After trying few things to avoid this race and discussin it with Ed, This small changes seems to be more accurate and seems to fix the issue.

This is unblocking rolling out new canary in our internal cells

Update: For some reason I thought the fix didn't fix the original issue when I tested it with internal loki dev cell, almost a month ago. But after testing it for more than 2hr on same cell. I realized it is working fine (will leave it for a day like this just to confirm)

Verified following:
```
{.... container="loki-canary"} |= "websocket missing"
```
and noticed it's happening no longer after the fix.

Also verified the metrics
```
sum(increase(loki_canary_websocket_missing_entries_total{...}[$__range]))
```
It dropped to zero.

Screenshot_2022-10-30_21-08-43
Screenshot_2022-10-30_21-08-29

Update 2: After running it for whole day, looks there was two log entries missed on the websocket. My guess is since websocket is long-live connection and something can interrupt the connection loosing this message.

Screenshot_2022-10-31_10-32-20

I also quickly checked how often this happens on other bigger envs (say ops) and looks like it's not that uncommon.

Screenshot_2022-10-31_10-30-56

Fixes https://github.com/grafana/loki/issues/7142

Tested it on internal loki dev cells

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/canary/writer/writer.go
+++ b/pkg/canary/writer/writer.go
@@ -94,12 +94,11 @@ func (w *Writer) run() {
 				w.pad = str.String()
 				w.prevTsLen = tsLen
 			}
-
+			w.sent <- t
 			_, err := fmt.Fprintf(w.w, LogEntry, ts, w.pad)
 			if err != nil {
 				level.Error(w.logger).Log("msg", "failed to write log entry", "error", err)
 			}
-			w.sent <- t
 		case <-w.quit:
 			return
 		}


dannykopping (approved):
LGTM

PR #7216, opened 2022-09-21
Fix promtail `metric` stage causing failure when getting extracted metrics

thepalbi:
Promtail allows ingested log entries to be modified before pushing them to Loki with the use of pipelines. One of the available pipeline "stages" is called `metrics`, and allows extracting metrics from incoming logs. These metrics are labeled with the labels available in the processed `Entry`. There's some edge cases in which if either a entry label is invalid, or it starts with the prometheus internal label prefix, `__`, calling the promtail `/metrics` fails.

One way to easily reproduce this is by having the following configuration:
```yaml
pipeline_stages:
- static_labels:
    good_label: "1"
- tenant:
    value: "2"
- metrics:
    loki_count:
      type: Counter
      description: "should count all entries"
      config:
        match_all: true
        action: inc
```
Which exposes a `promtail_custom_loki_count` metrics that counts processed log entries. In this case, the step before the `metrics` one injects an internal label `__tenant_id__` to be used by the remote write client. Since this label starts with the prometheus reserved prefix,  this will make calling the `metrics` endpoint fail with an error like the following:
```
An error has occurred while serving metrics:

1 error(s) occurred:
* "__tenant_id__" is not a valid label name for metric "promtail_custom_loki_count"
```

This PR attempts to fix that by, before creating the corresponding metrics, dropping invalid or internal labels. That way promtail will just use labels that are valid.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/clients/pkg/logentry/metric/metricvec.go
+++ b/clients/pkg/logentry/metric/metricvec.go
@@ -1,6 +1,7 @@
 package metric
 
 import (
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


owen-d (approved):
LGTM, thanks!

PR #5455, opened 2022-02-23
[performance] querier : parallel  execute logql pipeline  in  go routine

liguozhong:
**What this PR does / why we need it**:
When logql is LogSelector type, parallel execute the pipeline in in the go routine that fetch the s3 chunk data to improve the performance。

This PR has been tested in our online loki cluster and is very effective for logql query performance. The logql query time of 6h has been reduced from 50s to 14s.

with this pr ： 6h query =》 14s
image
without this pr ： 6h query =》 50s
image

**Which issue(s) this PR fixes**:
Fixes #5405

**Special notes for your reviewer**:
This PR is just a draft demo, I'm testing it online loki cluster.

Turn on this PR and configure 96 logql processing goroutine。
```yaml
querier:
  post_filter_chunk: true
  post_filter_max_parallel: 96
```

key changes:
pkg/storage/batch.go:684
 fetchLazyChunks()
image
tracing
image
parallel  pipeline
image

This Pr is really danger. it is powerful but OOM warning.
and metrics query diff:  199s-> 26s
without this pr
image
with this pr
image
resouce diff: 
 CPU :3 core -> 22core
 Memory: 7Gb ->27Gb
image

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] Add an entry in the `CHANGELOG.md` about the changes.

PR #7230, opened 2022-09-22
Loki: Improve query timeouts behavior

DylanGuedes:
Improves the different query-related timeouts behavior by fixing a bug and having better faults.
The changes are:
- Fix the timeout middleware by creating a new request that will use the new WithDeadline context
- For some cases the `engine:timeout` is ignored. This PR improves such scenario by using **the query timeout as a fallback** instead of the default `5min` for all cases

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


liguozhong:
great!👍
Because the default timeout of 5m makes our original timeout of 30m invalid, this PR is very compatible with the previous timeout configuration

DylanGuedes:
> great!+1 Because the default timeout of 5m makes our original timeout of 30m invalid, this PR is very compatible with the previous timeout configuration.
> 
> Sorry for leaving a review tip, but I'm not loki team,I am very concerned about the code of this module, but because of the default change of 5m, our loki timeout has changed from 30m to 5m.
> 
> This is a break change.I mean engine.go is really the key code

Sorry for the issues, and you're absolutely correct. I was the one who implemented the change and, although I paid a lot of attention to not making this a breaking change, in the end it broke things anyway :cry: 

> As a user, I need to modify the timeout configuration to make the behavior consistent
> 
> old conf
> 
> ```
> querier:
>   query_timeout: 30m
> ```
> 
> new conf
> 
> ```
> limits_config:
>   query_timeout: 30m
> ```
> 
> One day after I upgraded the code from the latest loki master branch, our loki users started reporting timeouts. After a period of investigation, I found that the timeout time changed from 30m to 5m. I'm not sure if this part of the code is the cause, because the error log is too vague, but after I modified the configuration, this error disappeared. image

Nice to know that at least you could fix it/return to the previous behavior.
Could you do me a favor and test that returning to `querier:query_timeout: 30m` only is working with this branch? I tested it locally and even wrote unitary tests but I'd be even more confident if it also works for you.

And by the way, you should receive the following error message:
```
Request timed out, decrease the duration of the request or add more label matchers (prefer exact match over regex match) to reduce the amount of data processed.
```
Weird that you received just `context deadline exceeded`.
image

liguozhong:
Your PR looks like it will make Loki's timeout behavior better.
https://github.com/grafana/loki/pull/5455
This is our online branch, but the latest code is synced from master branch every week, and currently I can't find the code branch that was wrong before. But I can wait after this PR(https://github.com/grafana/loki/pull/7230)  is merged into master before I merge the code from master to our branch. Then do the experiment you mentioned.

one more thing:
Because I need to achieve similar performance to ELK in the {app="buy"} |="traceid" log search scenario, without the feature of https://github.com/grafana/loki/pull/5455, I will not be able to replace ELK by loki. But https://github.com/grafana/loki/pull/5455 is not the goal of loki,  so I keep this branch in sync with master branch every week.

pkg/logql/engine.go
```diff
@@ -222,13 +227,21 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
...
-	userID, err := tenant.TenantID(ctx)
-	if err != nil {
-		level.Warn(q.logger).Log("msg", fmt.Sprintf("couldn't fetch tenantID to evaluate query timeout, using default value of %s", queryTimeout), "err", err)
-	} else {
-		queryTimeout = q.limits.QueryTimeout(userID)
+	queryTimeout := q.timeout
+	if q.timeout == 0 {
+		// Engine didn't set query timeout or user already migrated to new limits:query_timeout configuration.
+		// In this case, we can safely use the new limits timeout.
+		// If the limits timeout isn't configured either, use former 5 min timeout.
```
sandeepsukhani, line 234: I don't see this happening here since we are overwriting `queryTimeout` with `q.limits.QueryTimeout(userID)` and it would return `1m`, i.e. the default value for that user limit.

pkg/querier/http.go
```diff
@@ -502,10 +502,11 @@ func WrapQuerySpanAndTimeout(call string, q *QuerierAPI) middleware.Interface {
 				queryTimeout = q.cfg.QueryTimeout
 			}
 
-			_, cancel := context.WithDeadline(ctx, time.Now().Add(queryTimeout))
+			newCtx, cancel := context.WithTimeout(ctx, queryTimeout)
```
sandeepsukhani, line 505: This code looks correct to me. Just curious, was it an oversight or intentional to ignore the returned context here in the previous PR?

sandeepsukhani: Also, why do we need to do this if the `Engine` can enforce the timeout? As we can see, multiple places to enforce a timeout is causing a lot of trouble.

pkg/querier/http_test.go
```diff
@@ -34,3 +38,150 @@ func TestTailHandler(t *testing.T) {
...
+		defaultLimits.QueryTimeout = model.Duration(shortestTimeout)
+		limits, err := validation.NewOverrides(defaultLimits, nil)
+		require.NoError(t, err)
+		api := NewQuerierAPI(mockQuerierConfig(), nil, limits, log.NewNopLogger())
+
+		// configure old querier:query_timeout parameter.
+		// although it is longer than the limits timeout, it should supersede it.
+		api.cfg.QueryTimeout = time.Millisecond * 100
+
+		// querier:query_timeout is 5ms but it sleeps for 100ms, so timeout injected in the request is expected.
```
sandeepsukhani, line 118: `querier:query_timeout` and `sleepFor` both are `100ms`, no? Am I missing something?

DylanGuedes: sorry, the behavior/test is correct but the comment is wrong. my intention with this test was:
- the new query_timeout is 5ms, but it should be ignored/superseded by the old query_timeout (100ms in this case). This is ensured by `time.After(5ms)` happening before `ctx.Done()`, hence breaking the `select` loop
- Hmm now I'm thinking in adding an additional behavior/mechanism to `connSimulator`: a threshold where it should check that it won't timeout before that threshold. For instance: for this test, the conn simulator will first sleep for 5ms and ensure that it didn't timeout. Then, sleep for 95ms and ensure that it did timeout. WDYT?
-

pkg/querier/http.go
DylanGuedes: > This code looks correct to me. Just curious, was it an oversight or intentional to ignore the returned context here in the previous PR?

It was for sure an oversight on my side, sorry for that. 

> Also, why do we need to do this if the Engine can enforce the timeout? As we can see, multiple places to enforce a timeout is causing a lot of trouble.

Engine is used for queries and subqueries, while this part of the code is only used by queries.

I agree with you that this mechanism is error prone, but I feel like we can't yet jump to the final solution of getting rid of all timeouts and only relying on an engine timeout.

pkg/logql/engine.go
DylanGuedes: agreed, it should be rare, but since `tenant.TenantID` might error, I think we're in better shape with that conditional clause. WDYT?

pkg/querier/http.go
sandeepsukhani: No need to say sorry, it happens and I do mistakes too :)

> I agree with you that this mechanism is error prone, but I feel like we can't yet jump to the final solution of getting rid of all timeouts and only relying on an engine timeout.

Curious, what is stopping us from relying just on the engine? Is it backwards compatibility or confidence in it or something else?

pkg/logql/engine.go
sandeepsukhani: We are expected to have the tenantID set in the context, which is also required by downstream code. I think we can return an error if we don't find it.

pkg/querier/http.go
sandeepsukhani: I think we can make the engine backwards compatible if we do [this](https://github.com/grafana/loki/pull/7230/files/7ec1b5eaf0597e51d89409253e530283fff82aea#r980698619) unless I am missing something?

pkg/logql/engine.go
DylanGuedes: we can, but wouldn't it be a breaking change?

pkg/querier/http.go
DylanGuedes: > I think we can make the engine backwards compatible if we do this unless I am missing something?

agreed, I think we can make it backwards compatible this way.
However, be aware that triggering the timeout here instead of in the engine allow us to also timeout if the request struggle into something before reaching the engine.
On a side note, I would prefer having this PR focused on fixing the current behavior and then in a follow-up PR to centralize timeouts behavior in a single place. Mainly because changing the overall behavior will need some discussion between more maintainers, which will take more time.

sandeepsukhani: > However, be aware that triggering the timeout here instead of in the engine allow us to also timeout if the request struggle into something before reaching the engine.

I think timeout would be enforced in the engine to make it work with ruler as well? Anyways, I think the timer should start at the earliest point possible.

> On a side note, I would prefer having this PR focused on fixing the current behavior and then in a follow-up PR to centralize timeouts behavior in a single place. Mainly because changing the overall behavior will need some discussion between more maintainers, which will take more time.

I totally agree. Let us do a separate PR for it.

pkg/logql/engine.go
sandeepsukhani: I don't think so since all of our store code needs a tenantID in the context to know which user it belongs to. Infact we would have already looked for it even before reaching here, for e.g while enforcing max query parallelism in QF and max query length in Queriers.

pkg/querier/http_test.go
sandeepsukhani: I think you can add just a negative testcase with `dealine` > `sleepFor` and look for `didTimeout` being `false`? It will have the same effect and keep the code simple?

pkg/querier/http.go
DylanGuedes: > I think timeout would be enforced in the engine to make it work with ruler as well?

Indeed. 

> Anyways, I think the timer should start at the earliest point possible.

Same! So, I believe this is the main challenge with unifying the timeouts: multiple parts of the code query things in different forms, but we don't want to start counting the deadline too late.

pkg/querier/http_test.go
DylanGuedes: good call, thank you. I have though to check for `didTimeout` being truth, right?

pkg/logql/engine.go
DylanGuedes: right, but other things like the logcli also use this code. I don't have a strong opinion against removing this conditional but I feel safer keeping it.

That said if you think removing it is a good call just lmk and I'll remove it. Just wanted to double check things to make sure I'm not missing something :smile:

sandeepsukhani: My main concern here is that this comment is not matching with what code is doing here:
`If the limits timeout isn't configured either, use former 5 min timeout.`
If engine timeout is not set, we are using limits timeout, which defaults to `1m`.
We should either fix the comment or change the code to match the comment.

pkg/querier/http.go
sandeepsukhani: Let us give it a try in a separate PR. Going to resolve this for now.

sandeepsukhani (commented):
There is just one unresolved comment. The code LGTM. We can merge it once that comment gets resolved.

pkg/logql/engine.go
DylanGuedes: damnnnnnn got it, thanks for the patience :smile: 

yep, we're definitely defaulting to 1m instead of 5m in such cases. I reverted to not fallback to the 5min then and rewrote that comment, WDYT?

sandeepsukhani (approved):

PR #7248, opened 2022-09-26
Ingester stream api stop

DylanGuedes:
Stop the rateCalculator timer when stream is about to be removed.

N/A

### DIFF
--- a/pkg/ingester/instance.go
+++ b/pkg/ingester/instance.go
@@ -310,6 +310,7 @@ func (i *instance) getOrCreateStream(pushReqStream logproto.Stream, record *WALR
 
 // removeStream removes a stream from the instance.
 func (i *instance) removeStream(s *stream) {
+	s.Stop()
 	if i.streams.Delete(s) {
 		i.index.Delete(s.labels, s.fp)
 		i.streamsRemovedTotal.Inc()
--- a/pkg/ingester/instance.go
+++ b/pkg/ingester/instance.go
@@ -310,7 +310,6 @@ func (i *instance) getOrCreateStream(pushReqStream logproto.Stream, record *WALR
 
 // removeStream removes a stream from the instance.
 func (i *instance) removeStream(s *stream) {
-	s.Stop()
 	if i.streams.Delete(s) {
 		i.index.Delete(s.labels, s.fp)
 		i.streamsRemovedTotal.Inc()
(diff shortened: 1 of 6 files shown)


pkg/ingester/instance.go
```diff
@@ -310,6 +310,7 @@ func (i *instance) getOrCreateStream(pushReqStream logproto.Stream, record *WALR
 
 // removeStream removes a stream from the instance.
 func (i *instance) removeStream(s *stream) {
+	s.Stop()
```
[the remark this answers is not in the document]
DylanGuedes: good catch.

pkg/ingester/rate_calculator.go
```diff
@@ -34,6 +37,10 @@ func (c *RateCalculator) recordSample() {
 	for range t.C {
```
MasslessParticle, line 37: This works but a more idiomatic golang pattern is to use a stop chan of type `struct{}`

```go
for {
    select {
    case <-c.stopChan:
        return
    case <-t.C:
        rate := c.sample.Swap(0)
        c.storeNewRate(rate)		
    }
}
```

Then `Stop` can just
```go
close(c.stopChan)
```

DylanGuedes: I pushed a commit using a channel for it instead, but since I'm a newb with channels, I'm not sure the way I did it is idiomatic. WDYT?

pkg/ingester/streams_map.go
```diff
@@ -45,6 +45,7 @@ func (m *streamsMap) StoreByFP(fp model.Fingerprint, s *stream) {
 
 // Delete must be called inside WithLock
 func (m *streamsMap) Delete(s *stream) bool {
+	s.Stop()
```
MasslessParticle, line 48: We only want to do this if `loaded == true`

DylanGuedes: hmm are you sure? I considered it but since we'll always have the rateLimiter operating for a stream (even if it was not loaded), I felt this was necessary for all cases. I'll investigate what happens in such cases.

MasslessParticle: That's a fair point. `LoadAndDelete` returning false means the stream wasn't in the map. Is it even possible to delete a stream that's not in the map?

pkg/ingester/rate_calculator.go
MasslessParticle: I'd put the `t.C` case on the top because it's the thing I want someone to see first when they read this part, but that's stylistic. `select` statements are not evaluated in order so there's no behavioral difference.

lgtm

MasslessParticle (approved):

PR #7250, opened 2022-09-26
Expose tenant ID through relabel in push promtail targets

thepalbi:

Some Promtail targets that support pushing logs through HTTP, allow using a specific header for writing logs for multiple tenants. This log is propagated through a Promtail internal label, that is then consumed by the remote write client.

If Promtail is used in a multi-tenant manner, it'd be nice to somehow be able to track how much is each tenant's share from the logs received. This PR exposes the received tenant id value through relabeling. 

This allows for example, keeping it as a label and exposing a metric though the `metrics` pipeline stage, which can be segregated by tenant.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/clients/pkg/promtail/targets/gcplog/push_target_test.go
+++ b/clients/pkg/promtail/targets/gcplog/push_target_test.go
@@ -275,7 +275,15 @@ func TestPushTarget_UseTenantIDHeaderIfPresent(t *testing.T) {
 
 	prometheus.DefaultRegisterer = prometheus.NewRegistry()
 	metrics := gcplog.NewMetrics(prometheus.DefaultRegisterer)
-	pt, err := gcplog.NewGCPLogTarget(metrics, logger, eh, nil, "test_job", config)
+	pt, err := gcplog.NewGCPLogTarget(metrics, logger, eh, []*relabel.Config{
+		{
+			SourceLabels: model.LabelNames{"__tenant_id__"},
+			Regex:        relabel.MustNewRegexp("(.*)"),
+			Replacement:  "$1",
+			TargetLabel:  "tenant_id",
+			Action:       relabel.Replace,
+		},
+	}, "test_job", config)
 	require.NoError(t, err)
 	defer func() {
 		_ = pt.Stop()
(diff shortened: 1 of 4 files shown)


clients/pkg/promtail/targets/gcplog/push_target_test.go
```diff
@@ -275,7 +275,15 @@ func TestPushTarget_UseTenantIDHeaderIfPresent(t *testing.T) {
 
 	prometheus.DefaultRegisterer = prometheus.NewRegistry()
 	metrics := gcplog.NewMetrics(prometheus.DefaultRegisterer)
-	pt, err := gcplog.NewGCPLogTarget(metrics, logger, eh, nil, "test_job", config)
+	pt, err := gcplog.NewGCPLogTarget(metrics, logger, eh, []*relabel.Config{
+		{
```
trevorwhitney, line 279: nit: could you pull the `[]*relableConfig` into a variable to improve readability of the `NewGCPLogTarget` call?

clients/pkg/promtail/targets/heroku/target_test.go
```diff
@@ -297,7 +297,15 @@ func TestHerokuDrainTarget_UseTenantIDHeaderIfPresent(t *testing.T) {
 
 	prometheus.DefaultRegisterer = prometheus.NewRegistry()
 	metrics := NewMetrics(prometheus.DefaultRegisterer)
-	pt, err := NewTarget(metrics, logger, eh, "test_job", config, nil)
+	pt, err := NewTarget(metrics, logger, eh, "test_job", config, []*relabel.Config{
```
trevorwhitney, line 300: some nit about extracting a variable

trevorwhitney (approved):
Seems reasonable to me. Any performance impact you can think of by adding this additional internal label to every log received when the tenant ID header value is being used?

thepalbi:
@trevorwhitney thanks for reviewing. So it won't impact log lines unless the label is kept by a relabel config. Otherwise, all `__` prefixed labels are dropped in the targets.
Having said that, my main motivator is exposing this to the metrics stage, so that I can track per-tenant metrics. Keeping the tenant_id in the log entries per se I think it's not that useful since it will create a really high number of streams.

PR #7268, opened 2022-09-27
Promtail: Change max support compressed line size to be 2MB

DylanGuedes:
Change Promtail compression support max line size to 2MB instead of 4096 bytes (~4KB).
This however doesn't change the default buffer size.
4096 bytes means we can't ever have a buffer size bigger than ~4KB but since the default mem page size is 4096 ([source](https://unix.stackexchange.com/questions/128213/how-is-page-size-determined-in-virtual-address-space)), we end up ignoring all lines.


dannykopping (approved):
LGTM. How did this slip through the last PR?
This may indicate the absence of a required test

DylanGuedes:
> LGTM. How did this slip through the last PR? This may indicate the absence of a required test

I messed up: I pushed a commit changing it to 4096 everywhere, the CI failed but I concluded it failed for a test that was already flake and didn't have anything to do with this changes.

PR #7272, opened 2022-09-27
operator: Use cluster monitoring alertmanager by default on openshift clusters

aminesnow:
This PR enables the use of cluster monitoring alertmanager by default on Openshift clusters when applicable. The configuration can however be overridden by the user config.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -250,6 +250,20 @@ func CreateOrUpdateLokiStack(
...
+
+		err = k.Get(ctx, key, &svc)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return kverrors.Wrap(err, "failed to lookup alertmanager service", "name", key)
+		}
+
+		if err == nil {
+			opts.Ruler.OCPAlertManagerEnabled = true
+		}
+	}
```
periklis, line 265: I suggest to push this code block in a separate package internal to handlers package. This handler is already too big and needs some uncluttering, let's start with new additions like this one.

operator/internal/manifests/config.go
```diff
@@ -54,9 +60,33 @@ func ConfigOptions(opt Options) config.Options {
...
+					RefreshInterval: "1m",
+				},
+			}
+
+			if err := mergo.Merge(opt.Ruler.Spec.AlertManagerSpec, ams); err != nil {
+				return config.Options{}, kverrors.Wrap(err, "failed merging RulerSpec options")
+			}
+		}
+
+	}
```
periklis, line 87: The general pattern to apply openshift partial patches for automation looks like this:
1. Create a configure method in the openshift package
2. Apply the configure after `ConfigOptions()` returns and handle any error

See ConfigureDeploymentForTenantMode.

operator/internal/manifests/var.go
```diff
@@ -42,6 +42,14 @@ const (
 	dataDirectory         = "/tmp/loki"
 	rulesStorageDirectory = "/tmp/rules"
 
+	rulerContainerName = "loki-ruler"
+
+	monitoringSVCMain = "alertmanager-main"
+	// MonitoringNS is the namespace containing cluster monitoring objects such as alertmanager.
+	MonitoringNS = "openshift-monitoring"
+	// MonitoringSVCOperated is the name of the alertmanager service used for alerts.
+	MonitoringSVCOperated = "alertmanager-operated"
```
periklis, line 51: All these belongs to the openshift package.

operator/internal/manifests/config_test.go
```diff
@@ -282,8 +284,141 @@ func TestConfigOptions_RetentionConfig(t *testing.T) {
...
+				Stack: lokiv1.LokiStackSpec{
+					Tenants: &lokiv1.TenantsSpec{
+						Mode: lokiv1.Dynamic,
+					},
+				},
+			},
+			wantOptions: nil,
+		},
+		{
+			desc: "openshift-logging mode",
```
periklis, line 384: We have a fourth mode to test nowadays `openshift-network`

operator/internal/manifests/config_test.go
```diff
@@ -282,8 +284,141 @@ func TestConfigOptions_RetentionConfig(t *testing.T) {
...
```
periklis, line 323: We have a fourth mode to test nowadays `openshift-network`

periklis (changes requested):
In general it looks good to me. Only a couple of code separation improvements suggested to keep openshift things in a single package.

operator/internal/manifests/config.go
```diff
@@ -22,6 +21,12 @@ func LokiConfigMap(opt Options) (*corev1.ConfigMap, string, error) {
 		return nil, "", err
 	}
 
+	if opt.OpenShiftOptions.BuildOpts.OCPAlertManagerEnabled {
+		if err = ConfigureOptionsForMode(&cfg, opt.Stack.Tenants.Mode); err != nil {
+			return nil, "", err
+		}
+	}
```
periklis, line 28: Can you flip the outer if block inside `openshift.ConfigureOptions(cfg)`. This way we limit exposing OpenShiftOptions outside of the openshift package.

operator/internal/handlers/internal/rules/rules.go
```diff
@@ -118,3 +120,20 @@ func selectRecordingRules(ctx context.Context, k k8s.Client, rs *lokiv1.RulesSpe
...
+		err := k.Get(ctx, key, &svc)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return false, kverrors.Wrap(err, "failed to lookup alertmanager service", "name", key)
+		}
+
+		return err == nil, nil
+	}
+
+	return false, nil
+}
```
periklis, line 140: I suggest we create a separate `openshift` package and place the code in a file `alertmanager.go`, naming the function as `openshift.alertManagerSvcExists`

operator/internal/manifests/config.go
```diff
@@ -261,3 +240,15 @@ func retentionConfig(ls *lokiv1.LokiStackSpec) config.RetentionOptions {
...
+func ConfigureOptionsForMode(cfg *config.Options, mode lokiv1.ModeType) error {
+	switch mode {
+	case lokiv1.Static, lokiv1.Dynamic:
+		return nil // nothing to configure
+	case lokiv1.OpenshiftLogging, lokiv1.OpenshiftNetwork:
+		return openshift.ConfigureOptions(cfg)
+	}
+
+	return nil
+}
```
periklis, line 254: Please move this to the `gateway_tenants.go` file where all the other configure function live.

operator/internal/manifests/openshift/options.go
```diff
@@ -36,13 +36,14 @@ type AuthorizationSpec struct {
...
-	RulerName            string
-	Labels               map[string]string
+	LokiStackName          string
+	LokiStackNamespace     string
+	GatewayName            string
+	GatewaySvcName         string
+	GatewaySvcTargetPort   string
+	RulerName              string
+	Labels                 map[string]string
+	OCPAlertManagerEnabled bool
```
periklis, line 46: You can skip the `OCP` prefix here, as the field is accessible via `openshift.BuiltOptions.AlertManagerEnabled`

periklis (changes requested):

operator/internal/manifests/config.go
aminesnow: why gateway_tenants? it has nothing to do with the gateway..

periklis: Actually the filename is wrong nowadays it should be more `tenants.go`

aminesnow: ok I see, I will rename it then to avoid  confusion

operator/internal/handlers/internal/openshift/alertmanager.go
```diff
@@ -0,0 +1,31 @@
...
+
+		err := k.Get(ctx, key, &svc)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return false, kverrors.Wrap(err, "failed to lookup alertmanager service", "name", key)
+		}
+
+		return err == nil, nil
+	}
+
+	return false, nil
```
periklis, line 30:
```suggestion
	if opts.Stack.Tenants == nil || opts.Stack.Tenants.Mode != lokiv1.OpenshiftLogging {
		return false, nil
	}
	
	var svc corev1.Service
	key := client.ObjectKey{Name: openshift.MonitoringSVCOperated, Namespace: openshift.MonitoringNS}

	err := k.Get(ctx, key, &svc)
	if err != nil && !apierrors.IsNotFound(err) {
		return false, kverrors.Wrap(err, "failed to lookup alertmanager service", "name", key)
	}
	
	return err == nil, nil
```

periklis (commented):
Great work!

operator/internal/manifests/config.go
periklis: Let's rename in a separate PR, because this is a crucial file that will break many PRs right now.

periklis (commented):
Two minor bits and we are set to go.

periklis (approved):

PR #7273, opened 2022-09-27
Tests: Reduce flakiness of GCPLog and integration tests

DylanGuedes:
- Reduce flakiness of GCPLog tests by reusing context and stopping GCP topic
- Reduce flakiness of integration tests by:
  - using port as they're assigned
  - reusing context
  - adding timeout to calls (hence, if something get stuck, tests fail earlier)

### DIFF
--- a/clients/pkg/promtail/targets/gcplog/pull_target_test.go
+++ b/clients/pkg/promtail/targets/gcplog/pull_target_test.go
@@ -25,31 +25,32 @@ import (
 
 func TestPullTarget_Run(t *testing.T) {
 	// Goal: Check message written to pubsub topic is received by the target.
-	tt, apiclient, pubsubClient, teardown := testPullTarget(t)
+	ctx := context.Background()
+	tt, apiclient, pubsubClient, teardown := testPullTarget(ctx, t)
 	defer teardown()
 
 	// seed pubsub
-	ctx := context.Background()
 	tp, err := pubsubClient.CreateTopic(ctx, topic)
 	require.NoError(t, err)
+	defer tp.Stop()
+
 	_, err = pubsubClient.CreateSubscription(ctx, subscription, pubsub.SubscriptionConfig{
 		Topic: tp,
 	})
 	require.NoError(t, err)
 
 	var wg sync.WaitGroup
-
 	wg.Add(1)
+
 	go func() {
 		defer wg.Done()
-		_ = tt.run()
+		tt.run()
 	}()
 
-	publishMessage(t, tp)
-
+	publishMessage(ctx, t, tp)
 	// Wait till message is received by the run loop.
 	// NOTE(kavi): sleep is not ideal. but not other way to confirm if api.Handler received messages
-	time.Sleep(1 * time.Second)
+	time.Sleep(100 * time.Millisecond)
 
 	err = tt.Stop()
 	require.NoError(t, err)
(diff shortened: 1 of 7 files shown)


clients/pkg/promtail/targets/gcplog/pull_target_test.go
```diff
@@ -25,38 +25,41 @@ import (
...
+		tt.run() //nolint:errcheck
 	}()
 
-	publishMessage(t, tp)
-
+	publishMessage(ctx, t, tp)
 	// Wait till message is received by the run loop.
 	// NOTE(kavi): sleep is not ideal. but not other way to confirm if api.Handler received messages
-	time.Sleep(1 * time.Second)
+	time.Sleep(100 * time.Millisecond)
```
dannykopping, line 53: Do we need to change the sleep here? Shaving off 0.9s is probably not worth it if it makes the tests fail

integration/client/client.go
```diff
@@ -377,8 +380,11 @@ type Rules struct {
 }
 
 // RunRangeQuery runs a query and returns an error if anything went wrong
-func (c *Client) RunRangeQuery(query string) (*Response, error) {
-	buf, statusCode, err := c.run(c.rangeQueryURL(query))
+func (c *Client) RunRangeQuery(ctx context.Context, query string) (*Response, error) {
+	ctx, cancelFunc := context.WithTimeout(ctx, 30*time.Second)
```
dannykopping, line 384: Since we're defining `30*time.Second` in a few places, maybe it should be a central config of some kind?

dannykopping (commented):
Really appreciate the initiative here @DylanGuedes!
Left a couple small notes but overall LGTM

clients/pkg/promtail/targets/gcplog/pull_target_test.go
DylanGuedes: No, it is not necessary. But since this sleep was here to build enough time for `.Received()` to return the correct results, I split the sleep into two calls. I changed it now so that the sum of the two sleeps is equal to 1 second, WDYT?

integration/client/client.go
DylanGuedes: I made it a const instead then to make it less error prone. WDYT?

integration/client/client.go
```diff
@@ -472,7 +474,7 @@ func (c *Client) rangeQueryURL(query string) string {
 }
 
 func (c *Client) LabelNames(ctx context.Context) ([]string, error) {
-	ctx, cancelFunc := context.WithTimeout(ctx, time.Second)
+	ctx, cancelFunc := context.WithTimeout(ctx, requestTimeout)
```
dannykopping, line 477: These are changing from 1s to 30s now - is that intended?

clients/pkg/promtail/targets/gcplog/pull_target_test.go
dannykopping: Sounds good :+1: thanks

dannykopping (approved):
LGTM, with a question

integration/client/client.go
```diff
@@ -472,7 +472,7 @@ func (c *Client) rangeQueryURL(query string) string {
 }
 
 func (c *Client) LabelNames(ctx context.Context) ([]string, error) {
-	ctx, cancelFunc := context.WithTimeout(ctx, time.Second)
+	ctx, cancelFunc := context.WithTimeout(ctx, 30*time.Second)
```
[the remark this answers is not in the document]
DylanGuedes: I changed it for consistency (because I used 30s in other places on that same code). But the overall thinking here is: if we mess up with something, those calls hang forever (it happened to me). This timeout is just to not make our pipeline takes forever is a similar thing happen.

PR #7283, opened 2022-09-28
Tests: Improve several tests behavior

DylanGuedes:
Follow-up of https://github.com/grafana/loki/pull/7273
* Tune a filetarget test to remove created files to guarantee that all tailers are interrupted/fred
* Call pubsub close for our GCP `pull target`
* Append testName to the job names used in GCP tests. This will help us to identify which tests are struggling.


DylanGuedes:
Obs: The last 7 runs passed so this definitely make things better (although I don't think this will fix the GCPlogs flakiness).

chaudum (approved):
lgtm

PR #7292, opened 2022-09-29
Fix logentry drop to more intuitively match numeric fields (fixes #7011)

### LINKED ISSUE #7011: Promtail `drop` stage won't match a numeric json field using `value`
**Describe the bug**

Take the following example promtail config
```
          scrape_configs:
            - job_name: testing
              pipeline_stages:
              - json:
                expressions:
                  level:
              - drop:
                source: "level"
                value: "50"
```               

This will not match the following json even though intuitively you would expect that it would.

```
{
  "level": 50
}
```

This is because the `extracted map` in promtail only holds strings and we do an exact match to the json value:

https://github.com/grafana/loki/blob/cff8324f241cdc13318ff4b834be750a15ce05fc/clients/pkg/logentry/stages/drop.go#L162

**Expected behavior**
Intuitively you would expect this to match

**Workaround**

The regex matcher converts the json value to a string before doing the comparison, so this config will work:

```
          scrape_configs:
            - job_name: testing
              pipeline_stages:
              - json:
                expressions:
                  level:
              - drop:
                source: "level"
                expression: "50"
```               

I think a likely easy solution to this would be to convert the value to string in the exact match case like we do for regex: https://github.com/grafana/loki/blob/cff8324f241cdc13318ff4b834be750a15ce05fc/clients/pkg/logentry/stages/drop.go#L187

jmherbst:

1. Makes the `drop` stage matching for numeric fields more intuitive. 
2. Adds tests scenarios to confirm fix

Fixes #7011

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


clients/pkg/logentry/stages/drop.go
```diff
@@ -159,15 +159,22 @@ func (m *dropStage) shouldDrop(e Entry) bool {
 					level.Debug(m.logger).Log("msg", "line met drop criteria for finding source key in extracted map")
 				}
 			} else {
-				if *m.cfg.Value == v {
+				s, err := getString(v)
+				if err != nil {
+					if Debug {
+						level.Debug(m.logger).Log("msg", "Failed to convert extracted map value to string, cannot test value line will not be dropped.", "err", err, "type", reflect.TypeOf(v))
```
DylanGuedes, line 165:
```suggestion
						level.Debug(m.logger).Log("msg", "line will not be dropped, failed to convert extracted map value to string", "err", err, "type", reflect.TypeOf(v))
```

clients/pkg/logentry/stages/drop.go
```diff
@@ -159,15 +159,22 @@ func (m *dropStage) shouldDrop(e Entry) bool {
 					level.Debug(m.logger).Log("msg", "line met drop criteria for finding source key in extracted map")
 				}
 			} else {
-				if *m.cfg.Value == v {
+				s, err := getString(v)
```
DylanGuedes, line 162: Although I think this is an improvement, it means we won't drop lines in scenarios were the parsing didn't work, which is different from the current scenario. Because of it, I think this should be documented in our CHANGELOG.

DylanGuedes (approved):
Thanks for your contribution, that is nice since debugging Promtail stages is sometimes not easy and this helps with it.

I added a nit/suggestion but LGTM.

CHANGELOG.md
```diff
@@ -36,13 +36,14 @@
...
 * [7212](https://github.com/grafana/loki/pull/7212) **Juneezee**: Replaces deprecated `io/ioutil` with `io` and `os`.
+* [7292](https://github.com/grafana/loki/pull/7292) **jmherbst**: Add string conversion to value based drops to more intuitively match numeric fields. String conversion failure will result no lines being dropped.
 
 #### Promtail
 
 ##### Enhancements
 
 * [6708](https://github.com/grafana/loki/pull/6708) **DylanGuedes**: Add compressed files support to Promtail.
-* [5977](https://github.com/grafana/loki/pull/5977) **juissi-t** lambda-promtail: Add support for Kinesis data stream events
+* [5977P](https://github.com/grafana/loki/pull/5977) **juissi-t** lambda-promtail: Add support for Kinesis data stream events
```
DylanGuedes, line 46:
```suggestion
* [5977](https://github.com/grafana/loki/pull/5977) **juissi-t** lambda-promtail: Add support for Kinesis data stream events
```

CHANGELOG.md
```diff
@@ -36,13 +36,14 @@
 * [6349](https://github.com/grafana/loki/pull/6349) **simonswine**: Update the default HTTP listen port from 80 to 3100. Make sure to configure the port explicitly if you are using port 80.
 * [6835](https://github.com/grafana/loki/pull/6835) **DylanGuedes**: Add new per-tenant query timeout configuration and remove engine query timeout.
 * [7212](https://github.com/grafana/loki/pull/7212) **Juneezee**: Replaces deprecated `io/ioutil` with `io` and `os`.
+* [7292](https://github.com/grafana/loki/pull/7292) **jmherbst**: Add string conversion to value based drops to more intuitively match numeric fields. String conversion failure will result no lines being dropped.
```
DylanGuedes, line 39:
```suggestion
* [7292](https://github.com/grafana/loki/pull/7292) **jmherbst**: Add string conversion to value based drops to more intuitively match numeric fields. String conversion failure will result in no lines being dropped.
```

DylanGuedes (changes requested):

CHANGELOG.md
jmherbst: Corrected in https://github.com/grafana/loki/pull/7292/commits/67524b25f9f7437d41b289b90646b004668ace2d - Thanks @DylanGuedes

CHANGELOG.md
jmherbst: Corrected in https://github.com/grafana/loki/pull/7292/commits/67524b25f9f7437d41b289b90646b004668ace2d - Thanks @DylanGuedes

MichelHollands (approved):
LGTM

DylanGuedes (approved):

PR #7294, opened 2022-09-29
AST rangemapper support binary operation with a splittable operand

ssncferreira:
The previous implementation of AST rangemapper only supported mapping a instant query with a binary expression if both operands were splittable, or if 1 operand was splittable and the other was a literal expression (e.g., `5`). 
This PR updates the rangemapper to always map a binary expression if at least 1 operand is splittable.

In case sharding is enabled, the mapped query will be input to the shardmapper, and since this AST mapper supports all expression types, the non-splittable operand can be passed as input to be sharded.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Loki
 
 ##### Enhancements
+* [7294](https://github.com/grafana/loki/pull/7294) **ssncferreira**: Add support for AST `rangemapper` to split binary expressions if at least one operand is splittable.
 * [7227](https://github.com/grafana/loki/pull/7227) **Red-GV**: Add ability to configure tls minimum version and cipher suites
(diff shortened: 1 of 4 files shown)


ssncferreira:
Closing due to a known error verified for instance in the [unit test](https://github.com/grafana/loki/blob/2fb1b0e3ee5ac22626d0a72c3f660de512a69f74/pkg/logql/downstream_test.go#L294) where a [errorQuerier](https://github.com/grafana/loki/blob/main/pkg/logql/downstream.go#L223) is set as the default evaluator querier, resulting in an `unimplemented` error from the [SelectSamples](https://github.com/grafana/loki/blob/main/pkg/logql/downstream.go#L229-L231) method.

Created issue https://github.com/grafana/loki/issues/7303 to further investigate.

PR #7299, opened 2022-09-29
operator: Ensuring amd64 binaries are built; Correcting distroless image name

Red-GV:

This PR ensures that an amd64 binary is built. On M1 Macs, the operator image cannot be utilized because the image crashes with a `exec format error` error. This also adjusts the name of the distroless image used to what is depicted in the table here: https://github.com/GoogleContainerTools/distroless#what-images-are-available.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/operator/Dockerfile
+++ b/operator/Dockerfile
@@ -16,11 +16,11 @@ COPY controllers/ controllers/
 COPY internal/ internal/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build -mod=readonly -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=readonly -a -o manager main.go
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/static-debian11:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
 USER 65532:65532


operator/Dockerfile
```diff
@@ -16,11 +16,11 @@ COPY controllers/ controllers/
 COPY internal/ internal/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build -mod=readonly -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=readonly -a -o manager main.go
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/static-debian11:nonroot
```
periklis, line 23: Why is this change needed? The `-debian11` images are only for debugging. However we use this dockerfile to build production images on [docker.io/grafana/loki-operator](https://hub.docker.com/r/grafana/loki-operator)

operator/Dockerfile
```diff
@@ -16,11 +16,11 @@ COPY controllers/ controllers/
 COPY internal/ internal/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build -mod=readonly -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=readonly -a -o manager main.go
```
periklis, line 19: We use this Dockerfile to build multi-arch production images on [docker.io/grafana/loki-operator](https://hub.docker.com/r/grafana/loki-operator). So we cannot simply pin to `amd64`

periklis (changes requested):

operator/Dockerfile
Red-GV: Gotcha. I think then I just misunderstood the context behind the names described in the GitHub.

operator/Dockerfile
Red-GV: Gotcha. Hmm, maybe there should be a note in the developer notes for this then. As is, the `Dockerfile` does not work on an M1 Mac.

PR #7300, opened 2022-09-29
Bugfix include cache generations on querier response

owen-d:
We check the cache number (incremented by deletes) after executing a query with the value returned by the queriers which executed it. If they’re different, we know we can’t safely cache. However, we’re never actually returning a number from the queriers. This means that as soon as this cache number is incremented by a delete, it’ll disable the results cache for that tenant completely.

I suspect this has been a latent bug since we forked cortex (goes back to missing some code from this [PR](https://github.com/cortexproject/cortex/pull/2279) into cortex from @sandeepsukhani  back in the day!). It just showed itself once we started using deletes.

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -40,7 +40,6 @@ import (
 	"github.com/grafana/loki/pkg/lokifrontend/frontend/v2/frontendv2pb"
 	"github.com/grafana/loki/pkg/querier"
 	"github.com/grafana/loki/pkg/querier/queryrange"
-	"github.com/grafana/loki/pkg/querier/queryrange/queryrangebase"
 	"github.com/grafana/loki/pkg/ruler"
 	base_ruler "github.com/grafana/loki/pkg/ruler/base"
 	"github.com/grafana/loki/pkg/runtime"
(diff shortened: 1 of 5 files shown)


pkg/loki/modules.go
```diff
@@ -364,6 +364,7 @@ func (t *Loki) initQuerier() (services.Service, error) {
 
 	httpMiddleware := middleware.Merge(
 		httpreq.ExtractQueryMetricsMiddleware(),
+		serverutil.CacheGenNumberHeaderSetterMiddleware(t.cacheGenerationLoader),
```
sandeepsukhani, line 367: we can add a check for `supportIndexDeleteRequest` to see if we need to add this middleware or not?

sandeepsukhani (approved):
Nice catch! LGTM!

chaudum (commented):
Great find! Can we verify in a test that the query response contains the `ResultsCacheGenNumberHeaderName` header value if index deletes are supported?

owen-d:
> Great find! Can we verify in a test that the query response contains the ResultsCacheGenNumberHeaderName header value if index deletes are supported?

Yes, I think we can (perhaps in a larger integration test), but I'll leave it as a followup opportunity.

PR #7304, opened 2022-09-30
Import Canary dashboard into Loki mixin.

jeschkies:
Summary:
This imports our canary into the Loki mixin so it can be used by the Helm chart. The dashboard is disabled by default. It can be enabled by setting

```jsonnet
{
  _config+:: {
    canary+: {
      enabled: true,
    },
  },
}
```

### DIFF
--- a/production/loki-mixin/dashboards/loki-canary-dashboard.libsonnet
+++ b/production/loki-mixin/dashboards/loki-canary-dashboard.libsonnet
@@ -8,8 +8,6 @@ local grafana = import 'grafonnet/grafana.libsonnet';
     canary+: {
       enabled: false,
     },
-    tags: ['loki'],
-    dashboard_prefix: 'Loki / ',
   },
   grafanaDashboards+: if !$._config.canary.enabled then {} else {
     local dashboard = (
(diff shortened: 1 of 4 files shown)


production/loki-mixin/dashboards/loki-canary-dashboard.libsonnet
```diff
@@ -0,0 +1,153 @@
...
+local g = import 'grafana-builder/grafana.libsonnet';
+local grafana = import 'grafonnet/grafana.libsonnet';
+
+{
+  _config+:: {
+    canary+: {
+      enabled: false,
+    },
+    tags: ['loki'],
+    dashboard_prefix: 'Loki / ',
```
vlad-diachenko, line 12: *  `tags: ['loki']` - already defined by default in https://github.com/grafana/loki/blob/main/production/loki-mixin/config.libsonnet#L4 , so we do not need to override it here ,
*  `dashboard_prefix: 'Loki / ',` is not used. on line 19 we set it again.

```suggestion
```

vlad-diachenko (approved):
LGTM. just a nit

PR #7305, opened 2022-09-30
update regex for matching nlb logs while using lambda promtail

sandeepsukhani:
Regex in lambda promtail looks for specific file patterns for reading and pushing logs from it.
It currently does not read logs generated by nlb i.e having `_net` in the name. This PR updates the regex.
Cross checked it with was docs here https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-access-logs.html


MichelHollands (approved):
LGTM

PR #7311, opened 2022-09-30
Loki: Per-tenant stream sharding

DylanGuedes:
- Move stream sharding configuration to its own package to avoid cyclic imports
- Change stream sharding to be a per-tenant configuration
- Change ingesters to reject whole streams due to rate-limit based on per-tenant stream sharding 
- Change stream sharding flags prefix from `distributor.shard-stream` to `shard-stream`

This is part of the stream sharding work


pkg/ingester/stream.go
```diff
@@ -346,7 +348,7 @@ func (s *stream) validateEntries(entries []logproto.Entry, isReplay bool) ([]log
 		}
 
 		now := time.Now()
-		if !s.cfg.RateLimitWholeStream && !s.limiter.AllowN(now, len(entries[i].Line)) {
+		if rateLimitWholeStream && !s.limiter.AllowN(now, len(entries[i].Line)) {
```
MasslessParticle, line 351: This ought to be `!rateLimitWholeStream`. 

I see why our test missed this: because whole-stream rate limiting happens at the end, the result looks the same. It's important that we not call `AllowN` when `rateLimitWholeStream==true` because a successful call changes the limiter's state and affects the rate limit. It there a way we can tell how the limiter was called in the test to help us in the future?

pkg/ingester/stream_test.go
```diff
@@ -334,21 +338,24 @@ func TestPushRateLimit(t *testing.T) {
...
+	_, err = s.Push(context.Background(), entries, recordPool.GetRecord(), 0, true, true)
+	require.Error(t, err)
 	require.Contains(t, err.Error(), (&validation.ErrStreamRateLimit{RateLimit: l.PerStreamRateLimit, Labels: s.labelsString, Bytes: flagext.ByteSize(len(entries[1].Line))}).Error())
 }
 
 func TestPushRateLimitAllOrNothing(t *testing.T) {
 	l := validation.Limits{
 		PerStreamRateLimit:      10,
 		PerStreamRateLimitBurst: 10,
+		ShardStreams: &shardstreams.Config{
```
MasslessParticle, line 350: I don't think we need this setup because the limits for this are used in `instance.go` to determine `rateLimitWholeStream`

We should probably have a test in `instance.go` to make sure this stays wired together the same way in the future.

pkg/distributor/distributor_test.go
```diff
@@ -671,11 +671,22 @@ func TestStreamShard(t *testing.T) {
...
+			flagext.DefaultValues(distributorLimits)
+			distributorLimits.ShardStreams.DesiredRate = desiredRate
+
+			overrides, err := validation.NewOverrides(*distributorLimits, nil)
+			require.NoError(t, err)
+
+			validator, err := NewValidator(overrides)
+			require.NoError(t, err)
+
+			// tenantCfgs := runtime.DefaultTenantConfigs()
```
MasslessParticle, line 684: nit: should delete this

MasslessParticle (changes requested):
Looks good, overall. We've got to fix a bug and improve the tests

pkg/ingester/stream.go
DylanGuedes: great catch, thank you. I'll investigate how to do that.

pkg/distributor/distributor_test.go
DylanGuedes: oops :smile:

pkg/ingester/stream.go
DylanGuedes: Pushed a fix and a test covering it, wdyt?

pkg/ingester/stream_test.go
DylanGuedes: Good call. I added a test covering it, WDYT?

pkg/ingester/stream_test.go
```diff
@@ -312,6 +313,9 @@ func TestPushRateLimit(t *testing.T) {
...
```
MasslessParticle, line 316: nit: Because whole stream rate limiting is controlled by the flag we pass to `Push`, there's no need to set this up here, either.

MasslessParticle (approved):
Left a nit but lgtm

PR #7313, opened 2022-09-30
[dashboards] update dashboards for zone aware ingesters

cstyan:
Mostly related to the job selector; I deployed these against a local grafana instance along the way and I think I got everything correct but we might still need to update some of these again.

### DIFF
--- a/production/loki-mixin-compiled-ssd/dashboards/loki-writes-resources.json
+++ b/production/loki-mixin-compiled-ssd/dashboards/loki-writes-resources.json
@@ -65,6 +65,14 @@
                         "legendFormat": "{{pod}}",
                         "legendLink": null,
                         "step": 10
+                     },
+                     {
+                        "expr": "sum by(pod) (loki_ingester_memory_streams{cluster=~\"$cluster\", job=~\"($namespace)/(loki|enterprise-logs)-write\"})",
+                        "format": "time_series",
+                        "intervalFactor": 2,
+                        "legendFormat": "{{pod}}",
+                        "legendLink": null,
+                        "step": 10
                      }
                   ],
                   "thresholds": [ ],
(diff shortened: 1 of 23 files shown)


MasslessParticle (approved):

production/loki-mixin-compiled-ssd/dashboards/loki-writes-resources.json
```diff
@@ -323,7 +331,7 @@
                   "thresholds": [ ],
                   "timeFrom": null,
                   "timeShift": null,
-                  "title": "Memory (go heap inuse)",
+                  "title": "Memory (go heap inuse), ingester",
```
trevorwhitney, line 334: I don't think this is correct. `go_memstats_heap_inuse_bytes` will be for the whole process, which in the case of `loki-write` is more than just the ingester.

production/loki-mixin/dashboards/dashboard-loki-operational.json
```diff
@@ -25,61 +25,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 1,
-  "id": 68,
   "iteration": 1588704280892,
-  "links": [
```
trevorwhitney, line 30: why are we removing the links?

trevorwhitney (approved):
Approving to unblock, but 1 issue and 1 question. Thanks!

production/loki-mixin/dashboards/dashboard-loki-operational.json
cstyan: they should be populated via the common jsonnet code; technically I didn't need to remove this since we override the contents of `links` completely when calling either of these functions: https://github.com/grafana/loki/blob/main/production/loki-mixin/dashboards/dashboard-utils.libsonnet#L42-L74

this was just meant as cleanup, it can be confusing to try and wade through dashboard json when everything other than the actual queries are going to be overwritten by jsonnet code

production/loki-mixin-compiled-ssd/dashboards/loki-writes-resources.json
cstyan: yep the changes, other than the single `ingester` -> `ingester.*` at the top of writes-resources.libsonnet which generates this file were unnecessary so I've backed them out

Issue #1866, opened 2020-03-30
Multi cluster support?


omerlh:
Is there an official recommendation regarding running Loki on multiple clusters? Currently, each time we’re deploying a cluster, we need to set up a new data source in Grafana manually (we have single Grafana showing data from all clusters). It’s a bit hard to scale, maintain and use (cannot loon for data from all clusters). 
Are there any existing solutions? It’s not the same issue as tenancy, more like the issue that Thanos/federation solved for Prometheus - one instance that can be used for querying all clusters.

cyriltovena:
Right now the only scale path we offer is tanka in this repository `production/ksonnet/`, it's a distributed cluster where you can send all your logs from any other kubernetes cluster. (That version scales easily to 150MB/sec)

In the future we're looking at improving helm for that same use case and we're also thinking about loki to loki remote writes.

Regarding the datasource you can provision it via the Grafana API, but that doesn't solve looking at data from all cluster.

Federation is a good idea, but I'm not sure the current helm chart can handle multiple cluster load.

omerlh:
We’re using jsonnet to install Loki today, so if federation is possible I’ll be happy to test it.
Where I can read more about the distributed cluster architecture? Does it mean that one cluster is the master, and the rest just write to it?

cyriltovena:
No you just run a single cluster and it can eat all you're sending. It's not federation.

I assume you were using helm, if you already use jsonnet, then as a workaround for now you could just have a single big cluster and ingest all your k8s cluster with it.

omerlh:
So promtail will send all the logs to a centralized loki instance?

cyriltovena:
Yes that’s how we run it.

slim-bean:
Closing this issue as I _think_ this is resolved, please comment if that's not the case. Thanks!

tabishis:
In my case, I've multiple Kubernetes clusters which host apps which specific teams support. In each cluster, I've Loki and Promtail deployed. I've a centralized Grafana from which I create a data source to each cluster's Loki endpoint. It's too many Lokis to manage. What I'd like to do is only have promtail deployed to each Kubernetes cluster which will push data to a centralized Loki instance. Problem with this approach is the centralized Loki instance will contain data for all Kubernetes clusters, which in turn makes all the data available from all K8S clusters to all teams. I'd rather not filter data in Grafana dashboards by namespace for Loki logs to display to each team. Is there a way to control this at Loki level?

migueleliasweb:
[...]

PR #7324, opened 2022-10-03
[new feature] add Remote read API

### LINKED ISSUE #7306: Remote read API
Remote read API is already available in Prometheus, Thanos, Cortex and Mimir. When it is added to Loki it will enable sending queries to multiple Loki clusters via Loki instance with remote read configs. This will require just one data source in Grafana pointing to the Loki instance that pulls data from multiple upstream Loki clusters. Currently there have to be a separate data source for every managed Loki cluster. Using remote read API this way would solve some of the use cases posted in https://github.com/grafana/loki/issues/1866 .

Remote read API available on Loki queriers or query frontends under /api/v1/read like it was done in [Mimir](https://grafana.com/docs/mimir/latest/operators-guide/reference-http-api/#remote-read).

The alternative to that would be [cluster query federation](https://grafana.com/docs/enterprise-metrics/latest/operations/cluster-query-federation/), but it is also not (yet) available in Loki.

liguozhong:

image

[new feature] add Remote read API.

remote read use /Users/fuling/go/src/github.com/grafana/loki/pkg/logcli/client/client.go, loki server do not add new http handler.

```go
type Client interface {
	Query(queryStr string, limit int, time time.Time, direction logproto.Direction, quiet bool) (*loghttp.QueryResponse, error)
	QueryRange(queryStr string, limit int, start, end time.Time, direction logproto.Direction, step, interval time.Duration, quiet bool) (*loghttp.QueryResponse, error)
	ListLabelNames(quiet bool, start, end time.Time) (*loghttp.LabelResponse, error)
	ListLabelValues(name string, quiet bool, start, end time.Time) (*loghttp.LabelResponse, error)
	Series(matchers []string, start, end time.Time, quiet bool) (*loghttp.SeriesResponse, error)
	LiveTailQueryConn(queryStr string, delayFor time.Duration, limit int, start time.Time, quiet bool) (*websocket.Conn, error)
	GetOrgID() string
}
```

loki.yaml config demo
```
remote_read:
  - url: http://loki_us.svc:3100
    name: loki_us
    orgID: buy
  - url: http://loki_eu.svc:3100
    name: loki_eu
    orgID: carts
```

prometheus remote read doc：https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_read
Fixes #7306 And #https://github.com/grafana/loki/issues/1866

This PR is already quite large, this PR is only implementation. The basic framework of remote read and the configuration of remote read and an interface read.go SelectLogs test.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


jeschkies (changes requested):
We were planning a remote read API as well. I'll reject it just so we have time reviewing it not because I think our solution would be better.

I'm super there's interest in it.

jeschkies:
Is there an issue or something similar where this is discussed in more detail?

pkg/querier/remote/read.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	if response.Status != loghttp.QueryStatusSuccess {
+		return nil, errors.Errorf("remote read Querier selectLogs fail,response.Status %v", response.Status)
+	}
+
+	streams, ok := response.Data.Result.(loghttp.Streams)
+	if !ok {
+		return nil, errors.New("remote read Querier selectLogs fail,value cast (loghttp.Streams) fail")
+	}
+	return iter.NewStreamsIterator(streams.ToProto(), params.Direction), nil
```
jeschkies, line 74: I think we should inject the remote id like we did with the tenant ID for [multi tenant queries](https://github.com/grafana/loki/blob/8d9f162460f307d25fb4305eeccdcb0f24fbcbbb/pkg/querier/multi_tenant_querier.go#L294).

jeschkies (changes requested):
Nice work. I meant to introduce a new remote read API that would stream the results so we would only transfer the data we need.

liguozhong:
> Is there an issue or something similar where this is discussed in more detail?

hi @jeschkies ,As far as I know, there is no detailed discussion about remote read, but I have some experience with the remote read module in the prometheus project before. 

This PR is only to provide a template for the community to discuss, we can discuss in this PR.

I'm very sorry to submit this PR directly without discussing it in detail, trust me, I have no ill intentions. I just want to make loki a more mature and complete logging solution faster.

pkg/querier/remote/read.go
liguozhong: agree , you are right

liguozhong: done.

jeschkies:
> I'm very sorry to submit this PR directly without discussing it in detail, trust me, I have no ill intentions. I just want to make loki a more mature and complete logging solution faster.

@liguozhong no worries. I didn't mean to criticize. I was out for some time and thought I missed some info.

I'm super happy you took this initiative. I have a few ideas that I'll add later here.

liguozhong:
When I was operating prometheus before, I had some experience with remote read.

```
case 1: 
promql -> prometheus federation -> remote read adapter -> prometheus node

case 2: 
promql -> prometheus federation -> nginx（SLB or any load balance）-> remote read adapter -> prometheus node
```

The remote read of prometheus is implemented through the http protocol, so it is easy to use nginx to form a more complex global network topology.

But.
Previously, we tried to achieve a global read capability similar to globalview through the grpc protocol, but load balancing such as nginx that could not work for the grpc protocol.

so, we should pay attention to a question, whether we are http or grpc, if we use stream api, we should be careful about grpc.

liguozhong:
> @liguozhong @jeschkies whats the current status here I only see last progress in october? We're also interested in getting this feature.
> 
> We're bascially are willing to use loki similar to thanos.

I am waiting for review

pkg/querier/remote/read.go
```diff
@@ -0,0 +1,157 @@
...
+func (q Querier) SelectSamples(ctx context.Context, params logql.SelectSampleParams) (iter.SampleIterator, error) {
+	response, err := q.client.QueryRange(ctx, params.Selector, 1, params.Start, params.End, logproto.FORWARD, 0, 0, false)
+	if err != nil {
+		return nil, err
+	}
+	if response.Status != loghttp.QueryStatusSuccess {
+		return nil, errors.Errorf("remote read Querier selectSamples fail,response.Status %v", response.Status)
+	}
+
+	matrix, ok := response.Data.Result.(loghttp.Matrix)
```
owen-d, line 86: nit: need to fix error message mentioning `logthttp.Streams`. Also we may need to handle `loghttp.Vector` as well as `loghttp.Matrix`.

pkg/querier/fanout/fanout.go
```diff
@@ -0,0 +1,127 @@
...
+	if len(q.secondaries) == 0 {
+		return q.Querier.SelectLogs(ctx, params)
+	}
+	iters := make([]iter.EntryIterator, 0)
+	localIter, err := q.Querier.SelectLogs(ctx, params)
+	if err != nil {
+		return nil, err
+	}
+	iters = append(iters, localIter)
+	for _, remoteStore := range q.secondaries {
```
owen-d, line 33: We probably want to do these concurrently. Check out `github.com/grafana/dskit/concurrency` and the `ForEachJob` function specifically.

owen-d (commented):
Skimmed this, looks like a reasonable first approach to me. I expect it'll have some cost/performance struggles at high scale since we aren't pushing aggregations to the other loki installations, but we can always handle that separately when it becomes an issue. I'll give this a more thorough look soon.

Thanks and sorry for the wait!

pkg/querier/remote/read.go
liguozhong: done. thanks

pkg/querier/fanout/fanout.go
liguozhong: done.thanks.

liguozhong:
ready for reveiw .

JStickler (dismissed):
[docs squad]  Documentation LGTM

pkg/logcli/client/client.go
```diff
@@ -283,6 +285,16 @@ func (c *DefaultClient) getHTTPRequestHeader() (http.Header, error) {
 	if c.OrgID != "" {
 		h.Set("X-Scope-OrgID", c.OrgID)
 	}
+	//do not check error here
+	tenantIDs, _ := tenant.TenantIDs(ctx)
+	fakeTenant := len(tenantIDs) == 1 && tenantIDs[0] == "fake"
```
jeschkies, line 290: I think this should be documented better. Why should the error not be checked?

pkg/querier/fanout/fanout.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"github.com/grafana/loki/pkg/iter"
+	"github.com/grafana/loki/pkg/loghttp"
+	"github.com/grafana/loki/pkg/logproto"
+	"github.com/grafana/loki/pkg/logql"
+	"github.com/grafana/loki/pkg/querier"
+	"github.com/grafana/loki/pkg/storage/stores/index/stats"
+)
+
+const readConcurrency = 5
```
jeschkies, line 16: This should be configurable :slightly_smiling_face:

pkg/querier/remote/read.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, err
+	}
+	if response.Status != loghttp.QueryStatusSuccess {
+		return nil, errors.Errorf("remote read Querier selectSamples fail,response.Status %v", response.Status)
+	}
+
+	value := response.Data.Result
+	switch value.Type() {
+	case logqlmodel.ValueTypeStreams:
+		return nil, fmt.Errorf("remote read Querier selectLogs fail,Unable to parse unsupported type: %s ", value.Type())
```
jeschkies, line 91:
```suggestion
		return nil, fmt.Errorf("Unable to parse unsupported type: %s ", value.Type())
```

unifies the errors a little.

pkg/querier/remote/read.go
jeschkies: Sorry. I cannot find it :see_no_evil:

pkg/querier/remote/read.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+type Querier struct {
+	client client.Client
+	codec  queryrangebase.Codec
+	name   string
+}
+
+func (q Querier) SelectLogs(ctx context.Context, params logql.SelectLogParams) (iter.EntryIterator, error) {
+	response, err := q.client.QueryRange(ctx, params.Selector, int(params.Limit), params.Start, params.End, params.Direction, 0, 0, false)
```
jeschkies, line 64: So we can mimic streaming a little. The logcli client already uses batches to fetch results. We could change it so that it returns an iterator over the entries. This way we won't fetch too much all at once.

pkg/logcli/client/client.go
liguozhong: done. thanks
```
// ExtractOrgID gets the org ID from the context.
func ExtractOrgID(ctx context.Context) (string, error) {
	orgID, ok := ctx.Value(orgIDContextKey).(string)
	if !ok {
		return "", ErrNoOrgID
	}
	return orgID, nil
}
```

pkg/querier/fanout/fanout.go
liguozhong: thanks,done ,add queirer.MaxRemoteReadConcurrent config .

pkg/querier/remote/read.go
liguozhong: return iter.NewStreamsIterator(streams.ToProto(), params.Direction), nil

pkg/iter/entry_iterator.go:452

// NewStreamsIterator returns an iterator over logproto.Stream
func NewStreamsIterator(streams []logproto.Stream, direction logproto.Direction) EntryIterator {
	is := make([]EntryIterator, 0, len(streams))
	for i := range streams {
		is = append(is, NewStreamIterator(streams[i]))
	}
	return NewSortEntryIterator(is, direction)
}

liguozhong: > So we can mimic streaming a little. The logcli client already uses batches to fetch results. We could change it so that it returns an iterator over the entries. This way we won't fetch too much all at once.

`			resultLength, lastEntry = q.printResult(resp.Data.Result, out, lastEntry)
`

Thanks a lot, this streaming log note is really great.👍👍👍

I took a look at the implementation of logcli. The advantage of streaming is that it can continuously output the results to the cli command line.

But our remote read this time is to aggregate the results of multiple loki clusters, and then summarize the results uniformly, and return them to grafana at one time. At present, the http request of grafana log explore cannot implement streaming to print log plaintext.

Therefore, the implementation of remote read in this PR can not introduce streaming read for the time being. This will add a lot of code complexity, and grafana log explore cannot enjoy the same experience improvement as logcli.

Maybe we can complete this streaming read in the next PR, I am willing to go to grafana/grafana's loki datasource code to complete streaming printing from the ui source to the loki server end to end.  

In fact, this loki log explore supports streaming is also the result that our team has always wanted. The log explore of newRelic has a similar feature.

jeschkies: I didn't mean to implement streaming. I meant to fetch the results from the remote in batches like `DoQuery` does and return and iterator instead. My assumption was that we are not aggregating everything because of the limit. I need to take a closer look if we could do batching for metrics as well.

liguozhong: 👍 OK，I got it, I will try to implement SelectLogs+batch next week.

liguozhong:
I tried to implement the code for batch pulling remote storage. 
          If batchSize==0, it will still pull all at once
2 mode provide options for loki's operator to choose.

But this batch pulling code makes the code look more difficult to read, please help review

owen-d (changes requested):
I'm worried about performance of this at scale, which often coincides with users that want this feature. Right now, _no_ aggregation is done remotely, meaning we're going to ship potentially _a lot_ of data back to the host cluster on each query. This can quickly become prohibitively expensive (data egress costs).

Some other thoughts

* (mentioned above) How do we minimize bandwidth between clusters? Ideally we'd need to do more sophisticated query planning (aggregation push-down).
* What happens when the ideal shard factor should differ on different clusters during query parallelization?

I'm inclined to say we don't want this feature because at scale, this will be both costly and slow :( .

liguozhong:
> I'm worried about performance of this at scale, which often coincides with users that want this feature. Right now, _no_ aggregation is done remotely, meaning we're going to ship potentially _a lot_ of data back to the host cluster on each query. This can quickly become prohibitively expensive (data egress costs).
> 
> Some other thoughts
> 
> * (mentioned above) How do we minimize bandwidth between clusters? Ideally we'd need to do more sophisticated query planning (aggregation push-down).
> * What happens when the ideal shard factor should differ on different clusters during query parallelization?
> 
> I'm inclined to say we don't want this feature because at scale, this will be both costly and slow :( .

Agreed, introducing this feature will cause a lot of complaints

SuperQ:
@owen-d / @liguozhong This kind of remote-read federation is very much needed.

For example, if have multiple compute/cloud service providers in multiple geo regions. I would rather ship data per query over, than ship _all_ logs over our regional links. Most log output goes un-queried, so keeping the logs inside a regional or provider boundary is desired.

We would also rather have the ability to circuit-breaker pattern such that if a single geo region/provider is unavailable we can continue to have access to the regions/providers.

Without remote read / distributed query, Loki would a SPoF in our observability architecture.

Having a remote read / federated model is why we went with Thanos over Mimir.

liguozhong:
> @owen-d / @liguozhong This kind of remote-read federation is very much needed.
> 
> For example, if have multiple compute/cloud service providers in multiple geo regions. I would rather ship data per query over, than ship _all_ logs over our regional links. Most log output goes un-queried, so keeping the logs inside a regional or provider boundary is desired.
> 
> We would also rather have the ability to circuit-breaker pattern such that if a single geo region/provider is unavailable we can continue to have access to the regions/providers.
> 
> Without remote read / distributed query, Loki would a SPoF in our observability architecture.
> 
> Having a remote read / federated model is why we went with Thanos over Mimir.

Hi, I understand your situation very well. At present, I also operate and maintain 22 loki clusters. 
When I need to query a traceID log, I need to click 22 times each time. This is very uncomfortable. 
I am willing to complete this PR to free up this unproductive duplication of work , 

needs to be discussed by more people.

SuperQ:
Having this as an option, and then improving on things like query pushdown is probably the best incremental approach. Documenting the down-sides like fan-out and lack of push-down is fine IMO.

For example in Thanos, we had lots of issues with fanout (we have 2000+ Prometheus instances), so we introduced a label enforcement proxy to make sure users selected which promehteus instances they needed per query. This is slightly less intuitive, but a lot less cumbersome to maintaining 2000+ datasources. :)

jeschkies:
What do you think about blocking queries that do not perform well without remote aggregation?

liguozhong:
> What do you think about blocking queries that do not perform well without remote aggregation?

sorry, I missed your message.
Today our dev team asked me for this feature again. I found your message in the PR today.

`sum(count_over_time{app="buy2"})`

As far as I know, when the querier queries the ingester, it will execute this statement（`count_over_time{app="buy2"}`） in the ingester instead of `pulling all the original data back to the querier` for execution. What the ingester returns to the queier is a time series result . And this amount of data is very small, there will be no performance problems.

The Remote read API is the same, the performance problem is mainly concentrated in s3 file download, | json calculation. instead of happening in `sum`

Maybe my understanding is wrong.

jeschkies:
> What the ingester returns to the querier is a time series result . And this amount of data is very small, there will be no performance problems.

@liguozhong, that's because the ingester covers a small time range. If you query over days it's an issue. If I'm not mistaken we want to use something like the [shard_summer](https://github.com/grafana/loki/blob/main/pkg/querier/astmapper/shard_summer.go) here that fans out to each remote. Say you have one thousand pods and use their ids a labels. You'll have over a thousand streams returning. However, if you sum by say the namespace, the streams are reduced.

liguozhong:
> > What the ingester returns to the querier is a time series result . And this amount of data is very small, there will be no performance problems.
> 
> @liguozhong, that's because the ingester covers a small time range. If you query over days it's an issue. If I'm not mistaken we want to use something like the [shard_summer](https://github.com/grafana/loki/blob/main/pkg/querier/astmapper/shard_summer.go) here that fans out to each remote. Say you have one thousand pods and use their ids a labels. You'll have over a thousand streams returning. However, if you sum by say the namespace, the streams are reduced.

sorry ,I miss this review tips . I will try to make it right. thanks ,"over a thousand streams returning" this is really danger .

JStickler:
@liguozhong   This PR is over a year old.  Can we close it?  Or are you still hoping to get this merged?

paulojmdias:
Any updates on this PR?

nexuhan:
What would be needed to finalize this PR?

paulojmdias:
I believe this is a feature that Loki users desire and need.

With this in mind, I decided to build [Lokxy](https://github.com/paulojmdias/lokxy), a log aggregator for Loki, designed to collect and unify log streams from multiple sources into a single, queryable endpoint. This allows the users to aggregate data from multiple Loki deployments into a single Grafana data source for example.

I'd love for you to check it out, contribute, or share your thoughts. Every feedback helps move this project forward. Let's make it better together! 💪

JStickler:
@liguozhong This PR is over two years old. Can we close it? Or are you still hoping to get this merged?

chaudum:
There are no plans from our side to implement this as is. There is also a good alternative (Lokxy) that was [mentioned above](https://github.com/grafana/loki/pull/7324#issuecomment-2413533184). So closing this PR.

PR #7338, opened 2022-10-05
mmap tsdb index file while opening it for querying

sandeepsukhani:
In PR #7321, we re-introduced mmap for reading tsdb index, but it does not cover all the paths where we open the index. This PR takes care of it.

I have also removed the code for decompressing the index file since we always get a decompressed file from the `indexshipper`.


owen-d (approved):
Nice catch, thanks!

PR #7301, opened 2022-09-29
Service selector for single binary deployment fix.

besteban1989:
Single binary deployment is not working when going through default service. The issue is with the labels that currently includes the canary pods.

Fixes #7260

PR #7345, opened 2022-10-05
[loki helm chart] fixed to pass CI

vlad-diachenko:
Bumped version of the chart and added changelog. 
These changes were missed in https://github.com/grafana/loki/pull/7301


MichelHollands (approved):
LGTM

PR #7349, opened 2022-10-05
Implement Distributor Rate Store

MasslessParticle:
The `RateStore` concurrently queries all ingester `GetStreamRate` APIs and stores the results. `RateStore` deduplicates replicas and combines sharded streams so a caller of `RateFor` gets an unsharded view of a stream's rate.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -58,16 +58,19 @@ type Config struct {
 
 	// For testing.
 	factory ring_client.PoolFactory `yaml:"-"`
+
+	RateStore RateStoreConfig `yaml:"rate_store"`
 }
 
 // RegisterFlags registers distributor-related flags.
 func (cfg *Config) RegisterFlags(fs *flag.FlagSet) {
 	cfg.DistributorRing.RegisterFlags(fs)
+	cfg.RateStore.RegisterFlagsWithPrefix("distributor", fs)
 }
 
 // RateStore manages the ingestion rate of streams, populated by data fetched from ingesters.
 type RateStore interface {
-	RateFor(stream *logproto.Stream) (int, error)
+	RateFor(streamHash uint64) int64
 }
 
 // Distributor coordinates replicates and distribution of log streams.
@@ -150,6 +153,7 @@ func New(
 	if err != nil {
 		return nil, err
 	}
+
 	d := Distributor{
 		cfg:                    cfg,
 		clientCfg:              clientCfg,
@@ -188,7 +192,10 @@ func New(
 	d.replicationFactor.Set(float64(ingestersRing.ReplicationFactor()))
 	rfStats.Set(int64(ingestersRing.ReplicationFactor()))
 
-	servs = append(servs, d.pool)
+	rs := NewRateStore(d.cfg.RateStore, ingestersRing, d.pool, registerer)
+	d.rateStore = rs
+
+	servs = append(servs, d.pool, rs)
 	d.subservices, err = services.NewManager(servs...)
 	if err != nil {
 		return nil, errors.Wrap(err, "services manager")
@@ -196,8 +203,6 @@ func New(
 	d.subservicesWatcher = services.NewFailureWatcher()
 	d.subservicesWatcher.WatchManager(d.subservices)
 	d.Service = services.NewBasicService(d.starting, d.running, d.stopping)
-
-	d.rateStore = &noopRateStore{}
 
 	return &d, nil
 }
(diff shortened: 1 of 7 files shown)


pkg/distributor/ratestore.go
```diff
@@ -1,25 +1,200 @@
...
 
-func (n *noopRateStore) RateFor(stream *logproto.Stream) (int, error) {
-	return n.rate, nil
+func (s *rateStore) getRatesFromIngesters(ctx context.Context, clients chan logproto.StreamDataClient, responses chan *logproto.StreamRatesResponse) {
+	for c := range clients {
+		ctx, cancel := context.WithTimeout(ctx, s.ingesterTimeout)
+
+		resp, err := c.GetStreamRates(ctx, &logproto.StreamRatesRequest{})
+		if err != nil {
+			level.Error(util_log.Logger).Log("msg", "unable to get stream rates", "err", err)
```
DylanGuedes, line 167: maybe add a label with which ingester failed it

pkg/distributor/ratestore.go
```diff
@@ -1,25 +1,200 @@
...
+			s.rateRefreshFailures.WithLabelValues("ingester_error").Inc()
+		}
+
+		responses <- resp
```
DylanGuedes, line 171: you send resp to `responses` even if an error happened. asking to double check that if it is fine

pkg/distributor/ratestore.go
```diff
@@ -1,25 +1,200 @@
...
+			streamRates[rate.StreamHash] = rate
+		}
+	}
+
+	return streamRates
 }
 
-func (n *noopRateStore) RateFor(stream *logproto.Stream) (int, error) {
-	return n.rate, nil
+func (s *rateStore) getRatesFromIngesters(ctx context.Context, clients chan logproto.StreamDataClient, responses chan *logproto.StreamRatesResponse) {
```
DylanGuedes, line 161: thats extremely nitpicking but do you mind having this function before `ratesPerStream`? to align with the order of invocation

pkg/distributor/ratestore.go
```diff
@@ -1,25 +1,200 @@
...
+
+func ratesPerStream(responses chan *logproto.StreamRatesResponse, totalResponses int) map[uint64]*logproto.StreamRate {
+	streamRates := make(map[uint64]*logproto.StreamRate)
+	for i := 0; i < totalResponses; i++ {
+		resp := <-responses
+		if resp == nil {
+			continue
+		}
+
+		for i := 0; i < len(resp.StreamRates); i++ {
```
DylanGuedes, line 144: shouldn't you use a different variable here to not conflict with `i` from line 138?

pkg/distributor/ratestore.go
```diff
@@ -1,25 +1,200 @@
...
+	return nil
 }
 
-func (u *unshardableStreamErr) Error() string {
-	return fmt.Sprintf("couldn't shard stream %s. number of shards (%d) is higher than number of entries (%d)", u.labels, u.shardNum, u.entriesNum)
+func (s *rateStore) aggregateByShard(streamRates map[uint64]*logproto.StreamRate) map[uint64]int64 {
+	rates := make(map[uint64]int64)
+	for _, sr := range streamRates {
+		if _, ok := rates[sr.StreamHashNoShard]; ok {
+			rates[sr.StreamHashNoShard] += sr.Rate
```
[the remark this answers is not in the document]
DylanGuedes: Gotcha, that's fair. I'll resolve it to not block the work as I'm satisfied with the answer so it's now on your side if you think it is worth benchmarking it or not.

pkg/distributor/distributor.go
```diff
@@ -188,7 +198,20 @@ func New(
 	d.replicationFactor.Set(float64(ingestersRing.ReplicationFactor()))
 	rfStats.Set(int64(ingestersRing.ReplicationFactor()))
 
-	servs = append(servs, d.pool)
+	rs := NewRateStore(
```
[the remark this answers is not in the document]
DylanGuedes: It depends. The good part of it is that we can merge this work with more confidence/earlier, as we'd just have to have sharding disabled by default. The bad part is that it requires substantial changes and, once the work is validated, it becomes completely obsolete. That said, I'm happy with either.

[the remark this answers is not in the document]
MasslessParticle: That's a good point. WDYT of the latest commit? If no one has sharding enabled, we don't make the network call. That means this component runs but it doesn't do anything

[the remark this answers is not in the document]
DylanGuedes: it does it, thank you.

pkg/distributor/ratestore.go
```diff
@@ -1,25 +1,234 @@
...
+	rates := s.aggregateByShard(streamRates)
+
+	s.rateLock.Lock()
+	defer s.rateLock.Unlock()
+	s.rates = rates
+
+	return nil
+}
+
+func (s *rateStore) anySharingEnabled() bool {
```
DylanGuedes, line 126: nit (sharing -> sharding)

in case it was unwanted make sure to also update it at line 101 (mentioning it as just applying this suggestion will result in broken code)
```suggestion
func (s *rateStore) anyShardingEnabled() bool {
```

MasslessParticle: Thank you! 🤦‍♂️

DylanGuedes (approved):
lgtm!

Issue #7279, opened 2022-09-28
Helm chart - S3 credentials via role-based access and k8s annotations not honored


ralbon:
**Describe the bug**
Loki deployed in AWS EKS (1.23) doesn't assume AWS IAM role to manage s3 object storage configured.

**To Reproduce**
1. Deploy an AWS Role with correct policy to be assumed by Kubernetes serviceaccount
2. Use chart from `helm-loki-3.2.0` with values bellow:
```
...
  loki:
    auth_enabled: false
    storage:
      type: s3
      bucketNames:
        chunks: <mybucket>
        ruler: <mybucket>
      s3:
        endpoint: "https://s3.<region>.amazonaws.com"
        region: "<region>"
        insecure: false

  serviceAccount:
    annotations:
      eks.amazonaws.com/role-arn:  "<myarn>"
...
```

**terminal output**

However, read and write pod start with errors saying it cannot access to storage.
```
rpc error: code = Unknown desc = WebIdentityErr: failed to retrieve credentials
caused by: SerializationError: failed to unmarshal error message
        status code: 405, request id:
caused by: UnmarshalError: failed to unmarshal error message
        00000000  3c 3f 78 6d 6c 20 76 65  72 73 69 6f 6e 3d 22 31  |<?xml version="1|
00000010  2e 30 22 20 65 6e 63 6f  64 69 6e 67 3d 22 55 54  |.0" encoding="UT|
00000020  46 2d 38 22 3f 3e 0a 3c  45 72 72 6f 72 3e 3c 43  |F-8"?>.<Error><C|
00000030  6f 64 65 3e 4d 65 74 68  6f 64 4e 6f 74 41 6c 6c  |ode>MethodNotAll|
00000040  6f 77 65 64 3c 2f 43 6f  64 65 3e 3c 4d 65 73 73  |owed</Code><Mess|
00000050  61 67 65 3e 54 68 65 20  73 70 65 63 69 66 69 65  |age>The specifie|
00000060  64 20 6d 65 74 68 6f 64  20 69 73 20 6e 6f 74 20  |d method is not |
00000070  61 6c 6c 6f 77 65 64 20  61 67 61 69 6e 73 74 20  |allowed against |
00000080  74 68 69 73 20 72 65 73  6f 75 72 63 65 2e 3c 2f  |this resource.</|
00000090  4d 65 73 73 61 67 65 3e  3c 4d 65 74 68 6f 64 3e  |Message><Method>|
[...]

PR #7358, opened 2022-10-06
Helm: Loki config resource use secret instead of configmap

### LINKED ISSUE #7357: [Helm] Create loki config as Secret instead of ConfigMap
In some environments we are required to utilize S3 Programmatic Credentials and when these are created in cluster via the Helm chart these credentials get created within a ConfigMap, since there are credentials that should be obfuscated, the config should be within a Secret resource similar to the [`loki` helm-chart setup](https://github.com/grafana/helm-charts/blob/loki-2.14.0/charts/loki/templates/secret.yaml#L10).

Loki configuration is placed within a Kubernetes Secret resource instead of ConfigMap.

In a few environments we have utilized `existingSecretForConfig` but we use ephemeral clusters and it's much simpler to have the config created in-cluster as a secret compared to creating the secret ourselves and placing it in-cluster next to the Helm-chart via gitops.

rgarcia6520:
Signed-off-by: garcia.ryan <garcia.ryan@solute.us>

Since there are options for credentials to be fed into the Loki config, it should be created as a kubernetes Secret opposed to a ConfigMap, this matches setup for (now deprecated) loki single-binary helm-chart. 

Fixes #7357

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/README.md
+++ b/production/helm/loki/README.md
@@ -1,6 +1,6 @@
 # loki
 
-![Version: 3.2.1](https://img.shields.io/badge/Version-3.2.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.6.1](https://img.shields.io/badge/AppVersion-2.6.1-informational?style=flat-square)
+![Version: 3.3.0](https://img.shields.io/badge/Version-3.3.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.6.1](https://img.shields.io/badge/AppVersion-2.6.1-informational?style=flat-square)
 
 Helm chart for Grafana Loki in simple, scalable mode
 


trevorwhitney (changes requested):
Do you think it should always be a secret? I'm torn on this because we do our secret injections through other means, and I find it much nicer to have a configmap I can edit by hand when in a pinch (ie. a production outage). We already have the option `.Values.loki.existingSecretForConfig` for those who want to put the config in a secret. I understand they lose the benefit of the generated config, so I'm open to an option for putting this in a secret, but I don't think a secret should be the only option.

rgarcia6520:
> I'd say the way should be to use environment and run Loki with expand env or support #7279

That would be fine. Only data that would need obfuscating in the secret are the AWS credentials, so if those could be placed in their own secret and mounted as environment vars (if values are present) or yes more reliable support for using IAM Roles would also work.

trevorwhitney:
I'm also open to an option of opting in to putting the config in a secret, I just don't think it should by default always go into a secret.

gclawes:
Mimir already offers the ability to specify the config storage type (`ConfigMap` or `Secret`), I'd suggest just porting this functionality over:

https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/values.yaml#L78-L81
https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/templates/_helpers.tpl#L109-L123
https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/templates/mimir-config.yaml#L3-L7

trevorwhitney:
I would be happy to approve a PR that does this in the way mimir does as suggested by @gclawes.

trevorwhitney:
Closing this and replacing with an issue to make the config resource type configurable: https://github.com/grafana/loki/issues/8268

trevorwhitney:
https://github.com/grafana/loki/issues/8268

PR #7359, opened 2022-10-06
Enable Loki to configure Alertmanagers per tenant

### LINKED ISSUE #5847: alert manager url per tenant
Is there any way to add an alert manager URL for each tenant?
what does "# Each Alertmanager URL is treated as a separate group in the configuration." mean?

aminesnow:
This PR adds the ability for Loki to configure Alertmanagers per tenant.
Alertmanager can be configured globally, and the configuration may be overridden per tenant, allowing the possibility of defining an Alertmanager per tenant using the `limits_config`.

Fixes #5847

cc: @periklis

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


docs/sources/configuration/_index.md
```diff
@@ -660,28 +661,39 @@ alertmanager_client:
...
 [notification_queue_capacity: <int> | default = 10000]
 
 # HTTP timeout duration when sending notifications to the Alertmanager.
 # CLI flag: -ruler.notification-timeout
+# Deprecated: Use `alertmanagers_per_tenant` instead
 [notification_timeout: <duration> | default = 10s]
 
+# Configure alertmanager per tenant.
+# A map with tenant ID as key.
+alertmanagers_per_tenant:
```
dannykopping, line 694: This doesn't really fit with the way we configure tenant-specific settings in Loki.
https://grafana.com/docs/loki/latest/configuration/#limits_config

Could you refactor this to fit the existing model please?

dannykopping (changes requested):
I really like this idea, but I think the config isn't idiomatic to Loki.

docs/sources/configuration/_index.md
aminesnow: Hey @dannykopping, thanks for the comment, I think you're right but that means we leave the existing config as a global config that is overridden per tenant in the limits config right?

dannykopping: Correct, I think that's the only way we can support it right now that is consistent with the rest of Loki's configuration.

periklis:
@trevorwhitney @dannykopping Adding the backport label for 2.7 on this. I would like to see this landing to move forward with the Loki Operator agenda where we have the demand to point to multiple Alertmanager targets.

docs/sources/configuration/_index.md
```diff
@@ -2308,6 +2308,9 @@ The `limits_config` block configures global and per-tenant limits in Loki.
 # CLI flag: -ruler.max-rule-groups-per-tenant
 [ruler_max_rule_groups_per_tenant: <int> | default = 0]
 
+# Ruler alertmanager configuration per tenant.
+[ruler_alertmanager_config: <alertmanager_config>]
```
dannykopping, line 2312: I think this is stale now?

pkg/ruler/base/manager.go
```diff
@@ -185,6 +182,31 @@ func (r *DefaultMultiTenantManager) getOrCreateNotifier(userID string) (*notifie
...
+			if err != nil {
+				return nil, err
+			}
+
+			amCfg = tenantAmCfg
+		}
+
+		nCfg, err = buildNotifierConfig(amCfg, r.cfg.ExternalLabels)
+		if err != nil {
+			return nil, err
```
dannykopping, line 202: Please wrap these errors with a clear description of what went wrong

pkg/ruler/base/notifier.go
```diff
@@ -88,10 +76,46 @@ func (rn *rulerNotifier) stop() {
 	rn.wg.Wait()
 }
 
+func getAlertmanagerTenantConfig(amConfig ruler_config.AlertManagerConfig, amOverrides ruler_config.AlertManagerConfig) (ruler_config.AlertManagerConfig, error) {
+	if amOverrides.AlertmanagerURL != "" {
```
dannykopping, line 80: Why do we have to do this? Surely the zero value of `amConfig.AlertmanagerURL` is `""` anyway?

pkg/ruler/base/notifier.go
```diff
@@ -88,10 +76,46 @@ func (rn *rulerNotifier) stop() {
...
+	}
+
+	return amConfig, nil
+}
+
 // Builds a Prometheus config.Config from a ruler.Config with just the required
 // options to configure notifications to Alertmanager.
-func buildNotifierConfig(rulerConfig *Config) (*config.Config, error) {
-	amURLs := strings.Split(rulerConfig.AlertmanagerURL, ",")
+func buildNotifierConfig(amConfig ruler_config.AlertManagerConfig, externalLabels labels.Labels) (*config.Config, error) {
```
dannykopping, line 117: `ruler_config.AlertManagerConfig` is a big struct to be copying each time we call this function; can you use a pointer pls?

production/docker/config/loki-runtime.yaml
```diff
@@ -0,0 +1,11 @@
+overrides:
```
dannykopping, line 1: I suspect this was used for your tests, but I'm skeptical about committing this to the actual repo

production/docker/config/loki.yaml
```diff
@@ -104,3 +104,17 @@ querier:
 compactor:
   working_directory: /tmp/compactor
   shared_store: s3
+
```
dannykopping, line 107: Ditto - just for your testing?

dannykopping (commented):
Overall LGTM @aminesnow!
I've left a few comments, and I think we should remove most of the changes to `production/docker`

docs/sources/configuration/_index.md
aminesnow: What do you mean?

pkg/ruler/base/notifier.go
aminesnow: exactly, so we should only use the override when we have a non zero value, hence the check

production/docker/config/loki-runtime.yaml
aminesnow: indeed I used it for tests, and the reason why I wanted to commit it is so that anyone else wanting to test out Loki using docker-compose could do so easily with an already made config. Is this not the place to put such a thing?

pkg/ruler/base/notifier.go
dannykopping: Can't we just set the value, though? If the value is blank, then there's no override
Anyway, it's a nit - nothing serious

production/docker/config/loki-runtime.yaml
dannykopping: This setup is intended for production use, so I don't think these testing configs belong here.
I do like that you've setup alertmanager, so we can keep that, but the rest should be removed.

dannykopping:
> @trevorwhitney @dannykopping Adding the backport label for 2.7 on this. I would like to see this landing to move forward with the Loki Operator agenda where we have the demand to point to multiple Alertmanager targets.

I'll get back to you later today about this.

pkg/ruler/base/notifier.go
aminesnow: well the problem is, if the global config sets a URL, but no value is set per tenant, the the override URL would be empty, which would erase the global URL in the config (unless I misunderstood something, this would not be the desired outcome)

production/docker/config/loki-runtime.yaml
aminesnow: I see. I will remove those configs then

docs/sources/configuration/_index.md
aminesnow: for the global config yes, but not for the limits config. This part of the doc is for the limits config

dannykopping: Ah! Gotcha

pkg/ruler/base/notifier.go
dannykopping: I see, thanks :+1:

pkg/ruler/base/notifier.go
dannykopping: On second thought, this function is called only a couple times so ignore my comment

dannykopping:
> @trevorwhitney @dannykopping Adding the backport label for 2.7 on this. I would like to see this landing to move forward with the Loki Operator agenda where we have the demand to point to multiple Alertmanager targets.

@trevorwhitney just a reminder about this one

dannykopping (approved):
LGTM, thanks @aminesnow!!

PR #7360, opened 2022-10-06
Fix Flaky Integration Tests

MasslessParticle:
Allow the OS kernel to decide what's an open port instead of trying to guess

### DIFF
--- a/integration/cluster/cluster.go
+++ b/integration/cluster/cluster.go
@@ -8,7 +8,6 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
-	"net/url"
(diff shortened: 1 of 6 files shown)


DylanGuedes (approved):
this is neat, good catch. Btw, how did a test fail with the previous approach? I made it to bind the free port before instantiating the next components to make sure no two components would fight for the same port, no idea how it was still be possible :joy:

MasslessParticle:
We run the tests concurrently on the pr branch and the main branch to get our test diff. I think the two integration tests running at the same time exposed a race condition in `getFreePort`

PR #7362, opened 2022-10-06
TSDB explicit fingerprint

owen-d:
This PR does two things:
* Includes the series fingerprint in TSDB WALs. Doing this ensures we recover fingerprint used in chunk addresses after restart even if loki is upgraded and the label hashing algorithm changes.
* Removes much of the extra hashing in the TSDB write path + manager. Occasionally, the fingerprint may be different than the hash of labels. Historically this has happened in a few places:
  * When the `__name__="logs"` label was injected for non-tsdb indices (not applicable to TSDB)
  * When building a multitenant tsdb index and adding a synthetic `__loki_tenant__` label
  * In a previous bug(https://github.com/grafana/loki/pull/7355) when empty label values altered calculated fingerprints.

Explicitly storing/retrieving the desired fingerprint throughout helps avoid calculating it in error.

### DIFF
--- a/pkg/storage/stores/tsdb/head.go
+++ b/pkg/storage/stores/tsdb/head.go
@@ -159,12 +159,12 @@ func updateMintMaxt(mint, maxt int64, mintSrc, maxtSrc *atomic.Int64) {
 }
 
 // Note: chks must not be nil or zero-length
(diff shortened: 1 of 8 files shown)


sandeepsukhani (approved):
LGTM

PR #7370, opened 2022-10-07
Measure Per-Stream rates

MasslessParticle:
This PR implements the ability for ingesters to measure per-stream rates at 1/s resolution.

It also fixes bugs in the kind of Limits used in the RateStore and the Ingester API

### DIFF
--- a/pkg/distributor/ratestore.go
+++ b/pkg/distributor/ratestore.go
@@ -6,8 +6,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/grafana/loki/pkg/validation"
-
 	"github.com/weaveworks/common/instrument"
 
 	"github.com/grafana/dskit/services"
@@ -46,10 +44,6 @@ type ingesterClient struct {
 	client logproto.StreamDataClient
 }
 
-type overrides interface {
-	AllByUserID() map[string]*validation.Limits
-}
-
 type rateStore struct {
 	services.Service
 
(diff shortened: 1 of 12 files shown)


pkg/distributor/ratestore.go
```diff
@@ -62,10 +56,10 @@ type rateStore struct {
 	maxParallelism         int
 	rateRefreshFailures    *prometheus.CounterVec
 	refreshDuration        *instrument.HistogramCollector
-	overrides              overrides
+	overrides              Limits
```
DylanGuedes, line 59: nit: maybe other Loki maintainers will be confused with overrides being used as limits here?

pkg/ingester/stream_rate_calculator.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"sync"
+	"time"
+)
+
+const (
+	// defaultStripeSize is the default number of entries to allocate in the
+	// stripeSeries list.
+	defaultStripeSize = 1 << 15
```
DylanGuedes, line 11: that's ~32k. Doesn't that mean we're guaranteed to have a conflict if the number of streams is higher than 32k? If so, I feel like 32k is a low value. WDYT?

pkg/distributor/ratestore.go
```diff
@@ -126,11 +120,12 @@ func (s *rateStore) updateAllRates(ctx context.Context) error {
...
 	limits := s.overrides.AllByUserID()
 	if limits == nil {
-		return false
+		// There aren't any tenant limits, check the default
+		return s.overrides.ShardStreams("fake").Enabled
 	}
 
-	for _, l := range limits {
-		if l.ShardStreams.Enabled {
+	for user, _ := range limits {
```
DylanGuedes, line 127:
```suggestion
	for _, user := range limits {
```

DylanGuedes (approved):

PR #7380, opened 2022-10-10
[performance] metrics query: range vector support streaming agg when no overlap

liguozhong:

We try to speed up metrics query, we have 102 recording rules for BI analysis. Through the metrics query of logql。

`avg_over_time({log_type="service_metrics", module="eerouter",operation="MonitorContainer"}[59s] | json | memoryLimitInMB>=4096 | unwrap containersRetiring | __error__="") by (accountID, memoryLimitInMB)
`

 but logql often timeouts, which leads us to continuously reduce the range from [60s] to [30s] or even [5s]. 
_### **We hope that the granularity of bi can be at least 60s.**_

 Therefore, our team focused on analyzing the range function of loki code, and we found that the current implementation is to obtain all data points(r.window) within 60s, and then do the calculation.
```go

func (r *batchRangeVectorIterator) Next() bool {
	....
	r.load(rangeStart, rangeEnd)
	return true
}

func (r *batchRangeVectorIterator) load(start, end int64) {
	for lbs, sample, hasNext := r.iter.Peek(); hasNext; lbs, sample, hasNext = r.iter.Peek() {
		....
		p := promql.Point{
			T: sample.Timestamp,
			V: sample.Value,
		}
		series.Points = append(series.Points, p)
	}
}

func (r *batchRangeVectorIterator) At() (int64, promql.Vector) {
	...
        for _, series := range r.window {
		r.at = append(r.at, promql.Sample{
			Point: promql.Point{
				V: r.agg(series.Points),
				T: ts,
			},
			Metric: series.Metric,
		})
	}
	return ts, r.at
}

func countOverTime(samples []promql.Point) float64 {
	return float64(len(samples))
}
```

 This is a bit like the batch calculation of hadoop or hive. Our team has some experience in storm before, and we modified it a little. 

we Introduce the aggregation calculation xxx_over_time of the range into a streaming calculation, which can reduce a huge batch cache(r.window). This is very bad for our current data volume . Our data may reach 1 million points/60s. 

streaming calculation do not need w.windows. continue calculation each sample.
```
type CountOverTime struct {
	count float64
}

func (a *CountOverTime) agg(sample promql.Point) {
	a.count++
}

func (a *CountOverTime) at() float64 {
	return a.count
}
```

 The streaming mode cannot work well in the case of overlap, so in the case of overlap, this pr still use loki's existing batch agg calculation mode.

Fixes #<issue number>

```
Benchmark_RangeVectorIteratorCompare
Benchmark_RangeVectorIteratorCompare/streaming_agg
Benchmark_RangeVectorIteratorCompare/streaming_agg-12         	  133740	      8583 ns/op	    3967 B/op	      65 allocs/op
Benchmark_RangeVectorIteratorCompare/batch_agg
Benchmark_RangeVectorIteratorCompare/batch_agg-12             	  100306	     11885 ns/op	   34702 B/op	      35 allocs/op
PASS
```

benchmark report。Less memory reduces gc risk. and has a faster speed
```
StreamingAgg
Benchmark_RangeVectorIterator-12    	  132342	      8214 ns/op	    3968 B/op	      66 allocs/op

BatchAgg
Benchmark_RangeVectorIterator-12    	   95154	     12521 ns/op	   34748 B/op	      36 allocs/op

```

BI analysis example
image
range = [30s] ,duration = 31s. very slow.
image
error log
`msg="GET /loki/api/v1/query_range?direction=BACKWARD&end=1665472454000000000&limit=593&query=sum(count_over_time({log_type="service_metrics",module="api_server",operation="InvokeFunction"}|json|runtime="custom-container"|registryType="acree"|acclType="Default"|__error__=""[59s]))&shards=0_of_16&start=1665471552000000000&step=2.000000 (504) 1m12.284221346s Response: \"Request timed out, decrease the duration of the request or add more label matchers (prefer exact match over regex match) to reduce the amount of data processed.\\n\" ws: false; X-Query-Queue-Time: 58.69µs; X-Scope-Orgid: test; uber-trace-id: 3d3c62f404360341:0c0f61a3108670b4:670a8d2a689862e1:0;
`

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/logql/range_vector_test.go
```diff
@@ -56,6 +56,37 @@ func newPoint(t time.Time, v float64) promql.Point {
...
+		(30 * time.Second).Nanoseconds(),
+		0,
+		time.Unix(10, 0),
+		time.Unix(100, 0),
+	}
+
+	for i := 0; i < b.N; i++ {
+		i := 0
+		it, err := newRangeVectorIterator(newfakePeekingSampleIterator(),
+			&syntax.RangeAggregationExpr{Operation: "count_over_time"}, tt.selRange,
```
kavirajk, line 77: nit: `s/count_over_time/syntax.OpRangeTypeCount/`

Just to be more static check friendly.

```suggestion
			&syntax.RangeAggregationExpr{Operation: syntax.OpRangeTypeCount}, tt.selRange,
```

pkg/logql/range_vector_test.go
```diff
@@ -179,12 +210,14 @@ func Test_RangeVectorIterator(t *testing.T) {
 		t.Run(
 			fmt.Sprintf("logs[%s] - step: %s - offset: %s", time.Duration(tt.selRange), time.Duration(tt.step), time.Duration(tt.offset)),
 			func(t *testing.T) {
-				it := newRangeVectorIterator(newfakePeekingSampleIterator(), tt.selRange,
+				it, err := newRangeVectorIterator(newfakePeekingSampleIterator(),
+					&syntax.RangeAggregationExpr{Operation: "count_over_time"}, tt.selRange,
```
kavirajk, line 214:
```suggestion
					&syntax.RangeAggregationExpr{Operation: syntax.OpRangeTypeCount}, tt.selRange,
```

pkg/logql/range_vector_test.go
```diff
@@ -201,8 +234,11 @@ func Test_RangeVectorIteratorBadLabels(t *testing.T) {
 			Labels:  "{badlabels=}",
 			Samples: samples,
 		}))
-	it := newRangeVectorIterator(badIterator, (30 * time.Second).Nanoseconds(),
+	it, err := newRangeVectorIterator(badIterator,
+		&syntax.RangeAggregationExpr{Operation: "count_over_time"}, (30 * time.Second).Nanoseconds(),
```
kavirajk, line 238:
```suggestion
		&syntax.RangeAggregationExpr{Operation: syntax.OpRangeTypeCount}, (30 * time.Second).Nanoseconds(),
```

kavirajk (commented):
Took quick look. And it looks super nice. thanks for this PR @liguozhong.

will take deeper look today and wait for few more eyes before merging.

vlad-diachenko (commented):
Hey @liguozhong . thanks for your contribution. Looks great. 
Also, I would like to see these things before merging it:
1. tests that compare the results from both approaches. let's say, we can create some mock data, and run some queries against this data using `streamingAggregator` and `batchRangeVectorIterator` to compare the accuracy of the results. 
2. We need to mention these changes in https://github.com/grafana/loki/blob/main/CHANGELOG.md
3. It would be awesome if you could add more comments to the methods that you added with a description of how these methods work. for example, https://github.com/grafana/loki/blob/bc56d7e411b7f7393b5c632a0be190c891d62b4d/pkg/logql/range_vector.go#L146  describes this method only in general. It would be helpful so see something like this: https://github.com/grafana/loki/blob/bc56d7e411b7f7393b5c632a0be190c891d62b4d/pkg/logql/range_vector.go#L280-L284

liguozhong:
> Hey @liguozhong . thanks for your contribution. Looks great. Also, I would like to see these things before merging it:
> 
> 1. tests that compare the results from both approaches. let's say, we can create some mock data, and run some queries against this data using `streamingAggregator` and `batchRangeVectorIterator` to compare the accuracy of the results.
> 2. We need to mention these changes in https://github.com/grafana/loki/blob/main/CHANGELOG.md
> 3. It would be awesome if you could add more comments to the methods that you added with a description of how these methods work. for example, https://github.com/grafana/loki/blob/bc56d7e411b7f7393b5c632a0be190c891d62b4d/pkg/logql/range_vector.go#L146
>      describes this method only in general. It would be helpful so see something like this: https://github.com/grafana/loki/blob/bc56d7e411b7f7393b5c632a0be190c891d62b4d/pkg/logql/range_vector.go#L280-L284

done, thank you for taking time to help me review this PR.

liguozhong:
`

level=info ts=2022-11-14T11:26:56.866609673Z caller=metrics.go:143 component=frontend org_id=1662_qaawmopdln latency=slow query="avg_over_time({log_type=\"service_metrics\", module=\"api_server\", operation=\"InvokeFunction\"} \|= `\"runtime\":\"custom-container\"` \| regexp `\"registryType\":\"(?P<registryType>.*?)\"(,\|\\{\|\\[\|\\})` \| regexp `\"acclType\":\"(?P<acclType>.*?)\"(,\|\\{\|\\[\|\\})` \| regexp `\"latency\":(?P<latency>.*?)(,\|\\{\|\\[\|\\})` \| unwrap latency \| __error__=\"\" [59s]) by (registryType, acclType)" query_type=metric range_type=range length=0s start_delta=16h57m16.866596501s end_delta=16h57m16.866596725s step=1s duration=1m12.497572269s status=200 limit=5000 returned_lines=0 throughput=304MB total_bytes=22GB total_entries=3 queue_time=15.283388s subqueries=2 cache_chunk_req=0 cache_chunk_hit=0 cache_chunk_bytes_stored=162018 cache_chunk_bytes_fetched=0 cache_index_req=4497 cache_index_hit=4497 cache_result_req=0 cache_result_hit=0
`

 total_bytes=22GB  duration=1m12.497572269s
hi , @kavirajk   if you have time, please help continue to review this PR. At present, we have more such slow queries and expect to speed up

vlad-diachenko:
@liguozhong sorry for the delay. reviewing it )

liguozhong:
> @liguozhong sorry for the delay. reviewing it )

Thanks, I am very patient. Our team’s plan for the second half of the year is to optimize the metrics query. 
We will spend a long time on logql performance optimization, mainly to make our recording rule execute more smoothly. The recording rule is replacing the original BI report in our Hadoop, which is very flexible. But there is a disadvantage that it allows slower. Accelerating the metrics query will make loki more powerful.

vlad-diachenko (approved):
LGTM. thanks @liguozhong

liguozhong:
image

memory 20Gb -> 5Gb .

liguozhong:
image

less code:400 query

PR #7381, opened 2022-10-10
use a channel to signal completion of index read operation for tsdb and boltdb shipper

sandeepsukhani:
In PR #7367, I was relying on cancellation of context to detect if the index read operation has finished or not, but it is error prone since a cancelled or timed-out query could cancel the context while we are still reading the index. Since `tsdb` doesn't can't check the context for each read operation, we could still panic in this scenario. I am adding a channel where the consumer of the index would explicitly signal the completion of the read operation to the `indexshipper`

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/indexshipper/downloads/index_set_test.go
+++ b/pkg/storage/stores/indexshipper/downloads/index_set_test.go
@@ -40,10 +40,10 @@ func TestIndexSet_Init(t *testing.T) {
 		indexSet, stopFunc := buildTestIndexSet(t, userID, tempDir)
 		require.Len(t, indexSet.index, len(indexesSetup))
 		verifyIndexForEach(t, indexesSetup, func(callbackFunc index.ForEachIndexCallback) error {
-			ctx, cancel := context.WithCancel(context.Background())
-			defer cancel()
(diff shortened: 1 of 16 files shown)


pkg/storage/stores/indexshipper/downloads/index_set_test.go
```diff
@@ -88,10 +88,10 @@ func TestIndexSet_doConcurrentDownload(t *testing.T) {
 				require.Len(t, indexSet.index, tc)
 			}
 			verifyIndexForEach(t, indexesSetup, func(callbackFunc index.ForEachIndexCallback) error {
-				ctx, cancel := context.WithCancel(context.Background())
-				defer cancel()
+				doneChan := make(chan struct{})
+				defer close(doneChan)
 
-				return indexSet.ForEach(ctx, callbackFunc)
+				return indexSet.ForEach(context.Background(), doneChan, callbackFunc)
```
kavirajk, line 94: Q: May be unrelated to this change. But what is the purpose of calling `ForEach` with already closed `donechan` here?

kavirajk (approved):
LGTM 👍 just one clarification.

PR #7393, opened 2022-10-11
Fix stream sharding

MasslessParticle:
Promtail doesn't add a stream hash to push requests. As a result, we weren't accounting for the current rate in our sharding calculation and only sharding on individual large pushes. 

This PR ensures that streams have a hash as they're ingested. It also add a couple of metrics that were useful in debugging this issue.

### DIFF
--- a/pkg/distributor/ratestore_metrics.go
+++ b/pkg/distributor/ratestore_metrics.go
@@ -11,6 +11,7 @@ type ratestoreMetrics struct {
 	streamCount         prometheus.Gauge
 	maxStreamShardCount prometheus.Gauge
 	maxStreamRate       prometheus.Gauge
+	maxUniqueStreamRate prometheus.Gauge
 	refreshDuration     *instrument.HistogramCollector
 }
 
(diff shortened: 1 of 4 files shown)


pkg/distributor/ratestore_metrics.go
```diff
@@ -36,6 +37,11 @@ func newRateStoreMetrics(reg prometheus.Registerer) *ratestoreMetrics {
 			Name:      "rate_store_max_stream_rate_bytes",
 			Help:      "The maximum stream rate for any stream reported by ingesters during a sync operation. Sharded Streams are combined.",
 		}),
+		maxUniqueStreamRate: promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+			Namespace: "loki",
+			Name:      "rate_store_max_unique_stream_rate_bytes",
+			Help:      "The maximum stream rate for any stream reported by ingesters during a sync operation. Sharded Streams are considered seperate.",
```
DylanGuedes, line 43:
```suggestion
			Help:      "The maximum stream rate for any stream reported by ingesters during a sync operation. Sharded Streams are considered separate.",
```

DylanGuedes (approved):
lgtm! just a nit

Issue #5010, opened 2021-12-28
FluentD plugin line_format issues


RomanBadiornyi:
Hi guys,

**FluentD plugin incorrectly parses properties with key value format and does not respect indentation and line breaks with json**

When I use FluentD loki plugin of [1.2.16 version](https://rubygems.org/gems/fluent-plugin-grafana-loki/versions/1.2.16)
I get problems with either visualization of log event (format: json) or parsing of detected fields (format: key_value)
I can see in recent changes there was introduced support for [ndjson in promtail](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#loki_push_api)
It seems to me that I face similar kind of issue, while using json format, but with FluentD plugin.

**To Reproduce**
Use `line_format key_value`
**Expected behavior**
All key value pairs send by FluentD visible in the UI
**Actual behavior**
values are truncated, some of the keys like message is truncated as well in case of multi line log event
![keyvalye_format_output](https://user-images.githubusercontent.com/949814/147581861-380b07b2-59f3-472f-ba1e-89ca8e19d60f.PNG)

**To Reproduce**
Use `line_format json`
**Expected behavior**
Multiline json is broken into multiple lines as it does when key_value format used
**Actual behavior**
Multi line logs contains un-escaped \r\n, making it hard to read such things as stack traces.
![json_format_output](https://user-images.githubusercontent.com/949814/147581850-c4736e3c-b268-4ed5-8a3d-f2391decc6dd.PNG)

PR #7396, opened 2022-10-12
fluentd: Add un-escaping of control characters in JSON

drbugfinder-work:

Like mentioned in https://github.com/grafana/loki/issues/5010 we also want to use line_format json but noticed that newlines in the message text would be escaped. This change will un-escape double-escaped control characters (applies for \n, \r, \s, \t) e.g. newline characters "\\\n" by "\n" in the record.

This will fix e.g. newline errors when parsing the JSON message later (for example Grafana Dashboard, which shows newline errors otherwise).

Fixes https://github.com/grafana/loki/issues/5010

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/clients/cmd/fluentd/lib/fluent/plugin/out_loki.rb
+++ b/clients/cmd/fluentd/lib/fluent/plugin/out_loki.rb
@@ -278,6 +278,7 @@ module Fluent
           case @line_format
           when :json
             line = Yajl.dump(record)
+            line = line.gsub("\\n", "\n").gsub("\\r", "\r").gsub("\\s", "\s").gsub("\\t", "\t")
           when :key_value
             formatted_labels = []
             record.each do |k, v|


drbugfinder-work:
cc @dannykopping @owen-d

dannykopping (approved):
Seems reasonable to me
Thanks for your contribution @drbugfinder-work

drbugfinder-work:
@dannykopping thanks for your review!

Is there any release schedule for this component? The current version 1.2.18 is some commits behind the main-branch.

dannykopping:
@drbugfinder-work we have a 2.7 release coming in the next couple weeks

drbugfinder-work:
@dannykopping ok, so the update of the output plugin is coupled to the Loki release, correct?

dannykopping:
TBH I'm not sure how the output plugin is released - not my area. I'll ask the team

chaudum:
Hey @drbugfinder-work I am trying to understand what this changes does actually fix. There is a reason why special characters in encoded JSON are escaped, `Yajl.dump(record)` should handle that correctly.

Maybe you could give a full example of where the existing JSON formatting fails. As far as I understood, #5010 describes a problem with Detected Fields in Grafana not Loki.

chaudum:
I'm pretty sure that this is only a problem with Grafana and the detected fields and their pretty printing, but I will see if I can reproduce it.

chaudum:
I was not able to reproduce the issue with Loki 2.6.1 and Grafana 9.1.6

I used a Docker compose setup with the following `fluentd.conf`.

```
<source>
  @type tail
  <parse>
    @type json
  </parse>
  tag all
  path /var/log/testjson.log
</source>

<match all>
  @type loki
  url "#{ENV['LOKI_URL']}"
  username "#{ENV['LOKI_USERNAME']}"
  password "#{ENV['LOKI_PASSWORD']}"
  extra_labels {"source":"fluentd"}
  line_format json
  <buffer>
    flush_interval 10s
    flush_at_shutdown true
  </buffer>
  buffer_chunk_limit 1m
</match>
```



Pretty JSON enabled in Grafana:

chaudum:
That previous screenshot was actually from the wrong log source (had also a promtail running in the same docker compose env). Here's the screenshot from the logs sent by fluentd:

chaudum:
@drbugfinder-work  I am going to revert this change, as it would create non-valid JSON data. Special chars need to be escaped:

```console
$ js
Welcome to Node.js v18.10.0.
Type ".help" for more information.
> JSON.parse("{\"message\":\"a\nb\nc\"}")
Uncaught SyntaxError: Unexpected token 
 in JSON at position 13
> JSON.parse("{\"message\":\"a\\nb\\nc\"}")
{ message: 'a\nb\nc' }
```

drbugfinder-work:
> That previous screenshot was actually from the wrong log source (had also a promtail running in the same docker compose env). Here's the screenshot from the logs sent by fluentd:
> 
> 

@chaudum In all of your screenshots, you can see the error that we are referring to. Grafana tells you that the format is wrong: 
196709675-b89ace68-1f89-46c6-b913-4919246358d6

chaudum:
@drbugfinder-work Ah I see, you're confused by the button in Grafana. However, that seems to be an issue with Grafana, not Loki. I quickly looked at the grafana/grafana repo and stubled across issue https://github.com/grafana/grafana/issues/45502 which was transferred to a [discussion](https://github.com/grafana/grafana/discussions/46521). I cross-referenced your issue there. Feel free to add your input there as well :)

PR #7399, opened 2022-10-12
Helm: Allow users to define remote write for metrics.

jeschkies:
Previously users had to create their own `MetricsInstance` for the Grafana Agent operator to send metrics. This introduces the option for them to do so.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/templates/monitoring/grafana-agent.yaml
+++ b/production/helm/loki/templates/monitoring/grafana-agent.yaml
@@ -21,7 +21,14 @@ spec:
     instanceSelector:
       matchLabels:
         {{- include "loki.selectorLabels" $ | nindent 8 }}
-      # cluster label for logs is added in the LogsInstance
+  {{- with $.Values.monitoring.serviceMonitor}}
+  {{- if .metricsInstance.remoteWrite}}
+  metrics:
+    instanceSelector:
+      matchLabels:
+        {{- include "loki.selectorLabels" $ | nindent 8 }}
+  {{- end }}
+  {{- end }}
 
 ---
 
--- a/production/helm/loki/templates/monitoring/metrics-instance.yaml
+++ b/production/helm/loki/templates/monitoring/metrics-instance.yaml
@@ -1,7 +1,7 @@
-{{- if .Values.monitoring.selfMonitoring.enabled }}
-{{- with .Values.monitoring.selfMonitoring.logsInstance }}
+{{- if .Values.monitoring.serviceMonitor.enabled }}
+{{- with .Values.monitoring.serviceMonitor.metricsInstance }}
 apiVersion: monitoring.grafana.com/v1alpha1
-kind: LogsInstance
+kind: MetricsInstance
 metadata:
   name: {{ include "loki.fullname" $ }}
   namespace: {{ .namespace | default $.Release.Namespace }}
@@ -15,15 +15,14 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
-  clients:
-    {{- include "loki.logsInstanceClient" $ | nindent 4}}
-    {{- with .clients}}
+  {{- with .remoteWrite}}
+  remoteWrite:
     {{- toYaml . | nindent 4 }}
-    {{- end }}
+  {{- end }}
 
-  podLogsNamespaceSelector: {}
+  serviceMonitorNamespaceSelector: {}
 
-  podLogsSelector:
+  serviceMonitorSelector:
     matchLabels:
       {{- include "loki.selectorLabels" $ | nindent 6 }}
 {{- end -}}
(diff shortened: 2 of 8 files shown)


production/helm/loki/templates/monitoring/metrics-instance.yaml
```diff
@@ -0,0 +1,30 @@
+{{- if .Values.monitoring.serviceMonitor.enabled }}
+{{- with .Values.monitoring.serviceMonitor.metricsInstance }}
```
jeschkies, line 2: Is `metricsInstance` is `null` this will be empty, won't it?

production/helm/loki/templates/monitoring/grafana-agent.yaml
```diff
@@ -18,6 +18,10 @@ spec:
   serviceAccountName: {{ include "loki.fullname" $ }}-grafana-agent
   enableConfigReadAPI: {{ .enableConfigReadAPI }}
   logs:
+    instanceSelector:
+      matchLabels:
+        {{- include "loki.selectorLabels" $ | nindent 8 }}
+  metrics:
```
trevorwhitney, line 24: shouldn't we only include the metrics section if a metrics instance has been defined?

production/helm/loki/templates/monitoring/metrics-instance.yaml
trevorwhitney: yes, I believe so, `null` or an empty object. But you can test via the `helm template` command

production/helm/loki/templates/monitoring/metrics-instance.yaml
```diff
@@ -0,0 +1,30 @@
...
+    {{- end }}
+spec:
+  {{- with .remoteWrite}}
+  remoteWrite:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+
+  # Supply an empty namespace selector to look in all namespaces. Remove
+  # this to only look in the same namespace as the MetricsInstance CR
+  serviceMonitorNamespaceSelector: {}
```
trevorwhitney, line 25: do we need this? what happens if omitted, I would assume the default is same namespace? alternatively we know the namespace, so we could also be explicit?

production/helm/loki/templates/monitoring/servicemonitor.yaml
```diff
@@ -13,6 +13,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
   labels:
+    instance: primary
```
trevorwhitney, line 16: why are we adding this here and removing from pod logs?

trevorwhitney (changes requested):
thanks for this, great addition! sorry for the long delay on reviewing, just a few small things.

production/helm/loki/templates/monitoring/servicemonitor.yaml
jeschkies: That's just an error on my side. Thanks 😊

production/helm/loki/templates/monitoring/grafana-agent.yaml
jeschkies: ``
{{- with .Values.monitoring.serviceMonitor.metricsInstance }}
```
at the top should do this, shouldn't it?

File: production/helm/loki/templates/monitoring/metrics-instance.yaml (hunk 3 above)
Comment (jeschkies) [reply production/helm/loki/templates/monitoring/metrics-instance.yaml:25]: Ok, let's use `.Release.Namespace`

Comment (trevorwhitney) [reply production/helm/loki/templates/monitoring/metrics-instance.yaml:25]: loki namespace is in `.Release.Namespace`

**Review by trevorwhitney — APPROVED** [review]
LGTM, one small question, take it or leave it :)

Thanks Karsten!

File: production/helm/loki/templates/monitoring/logs-instance.yaml [hunk 5]
```diff
@@ -17,13 +17,16 @@ metadata:
 spec:
   clients:
     {{- include "loki.logsInstanceClient" $ | nindent 4}}
+    {{- with .clients}}
```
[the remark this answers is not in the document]
jeschkies: I was wondering the same but I didn't manage to append the clients.  Maybe this soups be a follow up. Isn't it possible to disable the default client?

production/helm/loki/templates/monitoring/metrics-instance.yaml
jeschkies: @trevorwhitney turns out we can only match on labels of the namespace (https://github.com/prometheus-operator/prometheus-operator/issues/1331#issuecomment-388917575). I'd say we move on without a selector and follow up.

trevorwhitney: since we're not creating the namespace resources ourselves we can't control what labels they're created with, so I agree with leaving this empty for now.

PR #7401, opened 2022-10-12
Promtail: Add handler timeout for GCP Logs Push target

thepalbi:
After been running the GCP Logs Push target in actual workloads, we've noticed that GCP sends a lot of traffic, and adjusts that send rate according to the rate at which the receiver can process. To perform that adjustment, GCP considers as a NACK [non 2xx HTTP responses](https://cloud.google.com/pubsub/docs/push#receive_push). 

This PR adds a target handler timeout (that includes as well the `api.Entry` channel send) to allow the user to configure this maximum processing time, therefore giving GCP notice that the sending rate is too high.

This new timeout is optional, and disabled by default.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/clients/pkg/promtail/targets/gcplog/push_target.go
+++ b/clients/pkg/promtail/targets/gcplog/push_target.go
@@ -1,11 +1,11 @@
 package gcplog
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
-	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -103,6 +103,14 @@ func (h *pushTarget) run() error {
 func (h *pushTarget) push(w http.ResponseWriter, r *http.Request) {
 	defer r.Body.Close()
 
+	// Create no-op context.WithTimeout returns to simplify logic
+	ctx := r.Context()
+	cancel := context.CancelFunc(func() {})
+	if h.config.PushTimeout != 0 {
+		ctx, cancel = context.WithTimeout(r.Context(), h.config.PushTimeout)
+	}
+	defer cancel()
+
 	pushMessage := PushMessage{}
 	bs, err := io.ReadAll(r.Body)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


clients/pkg/promtail/targets/gcplog/push_target.go
```diff
@@ -132,11 +143,33 @@ func (h *pushTarget) push(w http.ResponseWriter, r *http.Request) {
 
 	level.Debug(h.logger).Log("msg", fmt.Sprintf("Received line: %s", entry.Line))
 
-	h.entries <- entry
+	if err := h.doSendEntry(ctx, entry); err != nil {
+		h.metrics.gcpPushErrors.WithLabelValues("timeout").Inc()
```
kgeckhart, line 147: Just wondering if we need to track timeouts in this custom metric or if we can depend on the existing `promtail_gcp_push_target_..._request_duration_seconds_count` metric looking for a timeout status code instead.

clients/pkg/promtail/targets/gcplog/push_target.go
```diff
@@ -132,11 +143,33 @@ func (h *pushTarget) push(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, err.Error(), http.StatusServiceUnavailable)
+		return
+	}
+
 	h.metrics.gcpPushEntries.WithLabelValues().Inc()
 	w.WriteHeader(http.StatusNoContent)
 }
 
+func (h *pushTarget) doSendEntry(ctx context.Context, entry api.Entry) error {
+	// Since this setting is configured at the target level, it will be ignored for follow-up request due to branch prediction
```
kavirajk, line 158: > it will be ignored for follow-up request due to branch prediction

Didn't quite understand what this means? can you clarify please?

clients/pkg/promtail/targets/gcplog/push_target.go
```diff
@@ -132,11 +143,33 @@ func (h *pushTarget) push(w http.ResponseWriter, r *http.Request) {
...
+	if h.config.PushTimeout != 0 {
+		select {
```
kavirajk, line 160: wondering what is the problem in having this `select` irrespective of `h.config.PushTimeout`.? and wrap the error under `ctx.Done()` with `ctx.Err()`?

kavirajk (commented):
Looks overall good. thanks @thepalbi Left few minor comments.

kavirajk (approved):
LGTM. thanks @thepalbi

PR #7413, opened 2022-10-13
Loki: Implement new rate limit for pushes when stream sharding

DylanGuedes:
Introduce a new rate limit effective when stream sharding is enabled
- It helps with the backwards compatibility of the stream sharding feature
- It is completely ignored if not set to a value different than 0 (default)
- Even if it is configured, it won't be used by ingesters if sharding is disabled
- Allow clusters running stream sharding to control sharded streams streams rate limiting without touching the per-stream-rate-limit, making regressions impossible if a rollback is necessary
- Adds a new flag named `sharding_supported`, sent by distributors to ingesters. It will guarantee that, in case of a rollback, even if distributors are rolled out before ingesters, ingesters will behave the intended way.


MasslessParticle:
I don't we'll need this, after all. We should hold off on merging it for now

PR #7419, opened 2022-10-13
Adds support for IRSA credentials for the s3 storage mode

### LINKED ISSUE #3069: AWS IRSA requires delicate dance
**Describe the bug**

For AWS IAM roles for service accounts (IRSA) to work, it's important that the `.WithCredentials` method is never called https://github.com/grafana/loki/blob/8c1fe88409fe832b45da1f8f3cdd8116d81f4048/vendor/github.com/cortexproject/cortex/pkg/chunk/aws/s3_storage_client.go#L140. This means you need to pass an s3 url and dodge the other branches.

---

This appears to have been fixed in master of cortex https://github.com/cortexproject/cortex/blob/d775e195f186fe4e2407ca2c643bf7f2350bd6cd/pkg/chunk/aws/s3_storage_client.go#L190. So fix should be to simply update the dependency.

### LINKED ISSUE #7403: [Loki+Helm] S3 backend error: incorrect region, the bucket is not in 'eu-west-1' region at endpoint ''
**Describe the bug**
I'm running Loki in EKS using IRSA to use the s3 storage backend, and I'm getting the following errors in the `loki-read` pods:

```
│ level=error ts=2022-10-12T21:06:28.774105533Z caller=cached_client.go:65 msg="failed to build cache" err="BucketRegionError: incorrect region, the bucke │
│ t is not in 'eu-west-1' region at endpoint ''\n\tstatus code: 301, request id: , host id: "                                                              │
│ level=error ts=2022-10-12T21:06:28.774123664Z caller=compactor.go:432 msg="failed to run compaction" err="BucketRegionError: incorrect region, the bucke │
│ t is not in 'eu-west-1' region at endpoint ''\n\tstatus code: 301, request id: , host id: "                                                              │
│ level=error ts=2022-10-12T21:06:44.83911778Z caller=reporter.go:203 msg="failed to delete corrupted cluster seed file, deleting it" err="BucketRegionErr │
│ or: incorrect region, the bucket is not in 'eu-west-1' region at endpoint ''\n\tstatus code: 301, request id: , host id: "                               │
│ level=error ts=2022-10-12T21:07:21.961894672Z caller=ruler.go:497 msg="unable to list rules" err="BucketRegionError: incorrect region, the bucket is not │
│  in 'eu-west-1' region at endpoint ''\n\tstatus code: 301, request id: , host id: "
```

My helm chart values.yaml config is simple:
```yaml
        serviceAccount:
          name: loki-test
          annotations:
            eks.amazonaws.com/role-arn: arn:aws:iam::012345678901:role/my-role
        loki:
          storage:
            type: s3
            s3:
              s3: s3://eu-west-1/my-bucket
```

I can also confirm that the service account is configured correctly in a separate pod running the aws cli.  I've also tried multiple permutations of the `s3:` storage config, but I've only gotten different errors around assuming the role, and figured that this was the closest that I got to a working solution.

I've also confirmed the bucket region:
```bash
$ aws s3api get-bucket-location --bucket my-bucket
{
    "LocationConstraint": "eu-west-1"
}
```

**To Reproduce**
Using the following helm chart & version:
```yaml
chart: loki
repoURL: https://grafana.github.io/helm-charts
targetRevision: 3.2.1
```

**Expected behavior**
Able to connect to s3 bucket

**Environment:**
 - Infrastructure: EKS
 - Deployment tool: helm & argocd

sklarsa:
Per [my issue from yesterday](https://github.com/grafana/loki/issues/7403) and an [older issue I found](https://github.com/grafana/loki/issues/3069), loki seems to be incompatible with [AWS IRSA Roles](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html), or IAM Roles for Service Accounts.  These are the AWS-recommended way to access AWS resources running inside EKS.  Effectively, a properly-annotated k8s ServiceAccount will run [AssumeRoleWithWebIdentity](https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role-with-web-identity.html) inside a pod to grant the pod specified AWS IAM permissions.

On pod startup, AWS injects the `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` values into its environment.  AWS calls made by the pod use these credentials to assume the IAM role that is annotated on the pod's ServiceAccount.

Currently, loki expects an AWS access key and secret key and does not have the ability to use the environment variables that are set inside the pod.

This is tricky to test, since it requires a live EKS cluster to properly test.  I have created Docker image based off my forked branch, [sklarsa/loki-test:0.0.1](https://hub.docker.com/layers/sklarsa/loki-test/0.0.1/images/sha256-e941a9ffb686c8d12c98372b8548ecec4d214e53f1b2360d11668756bd9f589b?context=repo).  I'm currently running this container in a cluster and it appears to be connecting to S3.  I'm getting other errors, but I'm not sure if they're related or not.

I can work on updating tests, but I'd like to get some initial feedback to make sure that I'm on the right track here.

Fixes #7403
Fixes #3069

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/storage/chunk/client/aws/s3_storage_client.go
+++ b/pkg/storage/chunk/client/aws/s3_storage_client.go
@@ -48,11 +48,12 @@ var (
 	errUnsupportedSignatureVersion = errors.New("unsupported signature version")
 )
 
-type FileSystemTokenFetcher struct {
+type fileSystemTokenFetcher struct {
 	Path string
 }
 
-func (t FileSystemTokenFetcher) FetchToken(c credentials.Context) ([]byte, error) {
+// Implements the sts.TokenFetcher interface by reading the token from a file path
+func (t fileSystemTokenFetcher) FetchToken(c credentials.Context) ([]byte, error) {
 	return os.ReadFile(t.Path)
 }
 
@@ -257,17 +258,30 @@ func buildS3Client(cfg S3Config, hedgingCfg hedging.Config, hedging bool) (*s3.S
 		creds := credentials.NewStaticCredentials(cfg.AccessKeyID, cfg.SecretAccessKey.String(), "")
 		s3Config = s3Config.WithCredentials(creds)
 	} else if cfg.UseIrsa {
+
 		// If static credentials are not used, fall back to use IRSA (IAM Roles for Service Accounts)
 		// https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
 		// This is the recommended approach that is used to access AWS resources from inside k8s pods
+
+		// Use a new session with the environment defaults.  This will be used to AssumeRoleWithWebIdentity
+		// and allow the pod to assume the IRSA-specified role
 		stsSession, err := session.NewSession()
 		if err != nil {
 			return nil, err
 		}
 		stsCfg := sts.New(stsSession)
+
 		roleArn := os.Getenv("AWS_ROLE_ARN")
+		if roleArn == "" {
+			return nil, errors.New("AWS_ROLE_ARN environment variable expected but not found (or is empty)")
+		}
+
 		tokenPath := os.Getenv("AWS_WEB_IDENTITY_TOKEN_FILE")
-		fetcher := FileSystemTokenFetcher{Path: tokenPath}
+		if tokenPath == "" {
+			return nil, errors.New("AWS_WEB_IDENTITY_TOKEN_FILE environment variable expected but not found (or is empty)")
+		}
+
+		fetcher := fileSystemTokenFetcher{Path: tokenPath}
 		provider := stscreds.NewWebIdentityRoleProviderWithOptions(stsCfg, roleArn, "loki", fetcher)
 		creds := credentials.NewCredentials(provider)
 		s3Config = s3Config.WithCredentials(creds)
--- a/pkg/storage/chunk/client/aws/s3_storage_client.go
+++ b/pkg/storage/chunk/client/aws/s3_storage_client.go
@@ -17,11 +17,13 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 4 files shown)


pkg/storage/chunk/client/aws/s3_storage_client.go
```diff
@@ -77,6 +88,7 @@ type S3Config struct {
 	SignatureVersion string              `yaml:"signature_version"`
 	SSEConfig        bucket_s3.SSEConfig `yaml:"sse"`
 	BackoffConfig    backoff.Config      `yaml:"backoff_config"`
+	UseIrsa			 bool				 `yaml:"use_irsa"`
```
kavirajk, line 91: I would prefer to call it `UseIRSA` (on the struct field) just to be explicit about the acronym. (`use_irsa` is fine for yaml).

production/helm/loki/values.yaml
```diff
@@ -212,6 +212,7 @@ loki:
       s3ForcePathStyle: false
       insecure: false
       http_config: {}
+      use_irsa: true
```
kavirajk, line 215: this would be a breaking change no?. Breaks the existing helm users who uses ACCESSKEY based S3 credentials.

I would prefer to keep `false` as default and document this additional configs on the S3Config page (use_irsa + additional ENVs that need to be set)
https://grafana.com/docs/loki/latest/configuration/#s3_storage_config

kavirajk (commented):
Thanks @sklarsa for the PR. I like the changes, left few comments.

sklarsa:
Thanks for the review! Based on [this comment in the issue I raised](https://github.com/grafana/loki/issues/7403#issuecomment-1278600391), I still have hope that this is just a configuration issue and won't require code changes.  I'll keep you updated, and make those fixes in the meantime

sklarsa:
It looks like this was just a case of user error and misconfiguration :-(

I guess the AWS golang client is smart enough to use the environment's default credential chain if no credentials are explicitly passed to the client config.

okgolove:
@sklarsa could you show your result config?

sklarsa:
the `loki:` part of my `values.yaml`...:
```yaml
loki:
  storage:
    type: s3
    bucketNames:
      chunks: loki-chunks-bucket
      ruler: loki-ruler-bucket
      admin: loki-admin-bucket
    s3:
      region: us-east-2
  storage_config:
    boltdb_shipper:
      cache_ttl: 48h

  compactor:
    retention_enabled: true
```

PR #7423, opened 2022-10-14
Fix enable of internal server for modules

periklis:
Makes adding the internal server modules to the module map more robust. The previous implementation accidentally dropped other modules.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -571,23 +571,24 @@ func (t *Loki) setupModuleManager() error {
 	}
 
 	if t.Cfg.InternalServer.Enable {
-		depsToUpdate := []string{Distributor, Ingester, Querier, QueryFrontendTripperware, QueryScheduler, Ruler, TableManager, Compactor, IndexGateway}
-
-		for _, dep := range depsToUpdate {
-			var idx int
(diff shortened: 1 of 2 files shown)


pkg/loki/loki.go
```diff
@@ -571,22 +571,27 @@ func (t *Loki) setupModuleManager() error {
...
 
-			lhs := deps[dep][0 : idx+1]
-			rhs := deps[dep][idx+2:]
+			if idx == -1 {
+				continue
+			}
 
-			deps[dep] = append(lhs, InternalServer)
-			deps[dep] = append(deps[dep], rhs...)
+			if len(ds) == idx {
```
owen-d, line 587: I don't think this clause is possible. The only thing setting `idx` is the range over `ds` so it could never find an index at `len(ds)`, no?

Is this supposed to be `if len(ds)-1 == idx` to indicate a shortcut if the server is the last dependency index? If so, I don't think that optimization is needed 🤔

owen-d (commented):
Looks good, just one question

owen-d (approved):

PR #7427, opened 2022-10-14
Loki: Implement random entries assignment when sharding

DylanGuedes:
Modify sharding to assign random shards when the number of shards is bigger than the number of entries.
This helps us to avoid hot-spotting the first shards. Ex:
- Let's say we calculate the number of shards for a given stream to be 8, but we receive a push request for that same stream with only 3 entries. Previously, we would assign the 3 entries for the first 3 shards. That means that, if we regularly receive push requests for that stream with only 3 entries, the last 5 shards wouldn't receive any load. By randomizing the entries assignment we help spreading the load between all shards.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -444,7 +444,7 @@ func (d *Distributor) shardStream(stream logproto.Stream, streamSize int, userID
 		return d.divideEntriesBetweenShards(logger, userID, len(stream.Entries), shardStreamsCfg, stream, shardsOrder)
 	}
 
-	return d.divideEntriesBetweenShards(logger, userID, len(stream.Entries), shardStreamsCfg, stream, nil)
+	return d.divideEntriesBetweenShards(logger, userID, shardCount, shardStreamsCfg, stream, nil)
 }
 
 func (d *Distributor) randomizeShardsOrder(shardCount int) []int {
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"math"
+	"math/rand"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/distributor/distributor_test.go
```diff
@@ -906,15 +785,6 @@ func TestShardCountFor(t *testing.T) {
 			wantShards:     1,
 			wantErr:        false,
 		},
-		{
```
DylanGuedes, line 909: I removed this scenario because it is impossible to have `streamSize=2` if number of entries is 0.

pkg/distributor/distributor.go
```diff
@@ -432,13 +439,36 @@ func (d *Distributor) shardStream(stream logproto.Stream, streamSize int, userID
...
+func (d *Distributor) divideEntriesBetweenShards(logger log.Logger, userID string, shardCount int, shardStreamsCfg *shardstreams.Config, stream logproto.Stream, alternateShardsOrder []int) ([]uint32, []streamTracker) {
 	derivedKeys := make([]uint32, 0, shardCount)
 	derivedStreams := make([]streamTracker, 0, shardCount)
+	streamLabels := labelTemplate(stream.Labels)
+	streamPattern := streamLabels.String()
+
 	for i := 0; i < shardCount; i++ {
-		shard, ok := d.createShard(shardStreamsCfg, stream, streamLabels, streamPattern, shardCount, i)
+		j := i
+		if alternateShardsOrder != nil {
```
MasslessParticle, line 468: It's preferred to check `len(alternateShardsOrder) > 0` because it works for a 0-length slice and a nil slice while `alternateShardsOrder != nil` only works for a nil slice.

pkg/distributor/distributor.go
```diff
@@ -486,12 +516,12 @@ func (d *Distributor) createShard(streamshardCfg *shardstreams.Config, stream lo
 	}, true
 }
 
-func (d *Distributor) boundsFor(stream logproto.Stream, totalShards, shardNumber int, loggingEnabled bool) (int, int, bool) {
+func (d *Distributor) boundsFor(stream logproto.Stream, totalShards, spot int, loggingEnabled bool) (int, int, bool) {
```
MasslessParticle, line 519: What do you think about renaming `spot` to `index` or `idx`? I know the type is going to be awkward but it took me awhile to understand what was meant by `spot`?

MasslessParticle (approved):
This looks awesome. Just a couple of small fixes

pkg/distributor/distributor.go
DylanGuedes: idx/index feels much better, agreed

pkg/distributor/distributor.go
DylanGuedes: good call, pushed.

PR #7429, opened 2022-10-14
Fix result cache misses on sharded queries

MasslessParticle:
Headers weren't being propagated from query responses on sharded queries causing result cache misses. This PR:
- Adds headers to `logproto.Result` so we can save response headers as they come back from queries
- Metric-query results are turned into numbers well away from where responses are sent to the frontend: save all headers in the context so they can be pulled out when the final result is calculated
- Propagate headers in the final responses


pkg/logqlmodel/metadata/context.go
```diff
@@ -0,0 +1,82 @@
...
+	headers := make([]*definitions.PrometheusResponseHeader, 0, len(c.headers))
+	for k, vs := range c.headers {
+		header := definitions.PrometheusResponseHeader{
+			Name:   k,
+			Values: vs,
+		}
+		headers = append(headers, &header)
+	}
+
+	sort.Slice(headers, func(i, j int) bool {
```
trevorwhitney, line 63: what's the reason for sorting the headers?

trevorwhitney (approved):
Looks pretty straightforward to me, LGTM!

pkg/logqlmodel/metadata/context.go
MasslessParticle: It keeps the test the test deterministic 🙂

PR #7444, opened 2022-10-17
Results cache fix improvements

MasslessParticle:
Move middleware to someplace more sensible and incorporate feedback missed in the review for this.


owen-d (approved):
Needs a bit of linting to pass CI but the code LGTM

Issue #7276, opened 2022-09-28
Allow for multiple store configurations from the same provider


dannykopping:
**Is your feature request related to a problem? Please describe.**
Loki currently supports [period_configs](https://grafana.com/docs/loki/latest/configuration/#period_config) for specifying index & chunk schema changes over time. A user can select a `store` with each config, and this effectively allows users to switch providers should they need to. Queries that span multiple periods will route requests to the appropriate store.

Example:

```yaml
storage_config:
    azure:
        account_key: ${AZURE_BLOB_SECRET_ACCESS_KEY}
        account_name: lokiprod
        container_name: data

schema_config:
    configs:
        - from: "2020-07-31"
            index:
            period: 24h
            prefix: loki_prod_index_
            object_store: azure
            schema: v11
            store: boltdb-shipper
        - from: "2022-01-24"
            index:
            period: 24h
            prefix: loki_prod_index_
            object_store: azure
            schema: v12 # <-- changed from v11 to v12
            store: boltdb-shipper
```

We currently do not support multiple store configurations for the _same_ provider, however.

At Grafana Labs, we require this feature due to a misconfiguration of an Azure Blob bucket. Its replication policy was set to be [LRS](https://learn.microsoft.com/en-us/azure/storage/common/storage-redundancy#locally-redundant-storage) (locally redundant within a zone), but we actually need it to be [ZRS](https://learn.microsoft.com/en-us/azure/storage/common/storage-redundancy#zone-redundant-storage) (zonally redundant within a region). Migrating the data from one bucket to another is impractical and slow, so we would prefer to setup a new bucket, and read/write new chunks/indexes to this bucket. Data in the old bucket will be deleted according to its [retention period](https://grafana.com/docs/loki/latest/configuration/#limits_config).

**Describe the solution you'd like**
I would like to be able to define multiple store configurations from the same provider, and use these configurations within [period_configs](https://grafana.com/docs/loki/latest/configuration/#period_config) definitions.

Additionally, if we could add a path prefix to prepend to all path lookups in the bucket, that would also solve some outstanding feature-requests (listed below), and it seems like related work. _This is not required to deliver on the main value of this ticket, but it might be possible with some drive-by refactoring that needs to be done anyway._

```yaml
storage_config:
    azure:
        account_key: ${AZURE_BLOB_SECRET_ACCESS_KEY}
        account_name: lokiprod
        container_name: data
        type: azure # <--- new "type" field, derive from key name for backwards-compatibility?
[...]

PR #7447, opened 2022-10-18
compactor: multi-store support

ashwanthgoli:
This PR adds multi-store support for compactors. Since loki allows users to configure mutiple stores using schema_config, compactor should be able to operate on multiple object stores that contain index. Currently, it can perform compaction on indexes in a single store.

To maintain backward compatibility: if `boltdb.shipper.compactor.shared-store` is set, compactor will only operate on that store, else compactor will be initialized to operate on all the object store indexes (`boltdb, tsdb`) defined in the schema config.

This PR also adds a new config option to define where delete requests are to be stored - `boltdb.shipper.compactor.delete-request-store`. If it's not set, `boltdb.shipper.compactor.shared-store` is used for storing them, this is to ensure no config changes are required by the users when upgrading. Refer to [docs/sources/upgrading/_index.md](https://github.com/grafana/loki/blob/a85d01c9cd3c62a62b29a20274b71d3c51f1d11a/docs/sources/upgrading/_index.md) for more details.

**Which issue(s) this PR fixes:**
Fixes https://github.com/grafana/loki/issues/7276

**Checklist**
- [X] Documentation added
- [X] Tests updated
- [X] `CHANGELOG.md` updated
- [X] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`


pkg/loki/modules.go
```diff
@@ -999,22 +1000,46 @@ func (t *Loki) initCompactor() (services.Service, error) {
...
+	objectClients := make(map[string]client.ObjectClient)
+	if sharedStoreType := t.Cfg.CompactorConfig.SharedStoreType; sharedStoreType != "" {
+		if !config.UsingObjectStorageIndex(t.Cfg.SchemaConfig.Configs) {
+			level.Info(util_log.Logger).Log("msg", "Not using object storage index, not starting compactor")
+			return nil, nil
+		}
+
+		objectClient, err := storage.NewObjectClient(sharedStoreType, t.Cfg.StorageConfig, t.clientMetrics)
+		if err != nil {
+			return nil, err
```
dannykopping, line 1017: Please wrap all returned errors with clear descriptions for debugging purposes

Example:
```suggestion
			return nil, fmt.Errorf("cannot create object store client for store: %w", err)
```

pkg/storage/config/schema_config.go
```diff
@@ -240,17 +240,32 @@ func usingForPeriodConfigs(configs []PeriodConfig, fn func(PeriodConfig) bool) b
 	return false
 }
 
+func usingObjectStoreIndex(cfg PeriodConfig) bool {
+	switch cfg.IndexType {
+	case BoltDBShipperType, TSDBType:
+		return true
+	default:
+		return false
+	}
```
dannykopping, line 249:
```suggestion
	return cfg.IndexType == BoltDBShipperType || cfg.IndexType == TSDBType
```

pkg/storage/config/schema_config.go
```diff
@@ -240,17 +240,32 @@ func usingForPeriodConfigs(configs []PeriodConfig, fn func(PeriodConfig) bool) b
...
-			return true
-		default:
-			return false
+	return usingForPeriodConfigs(configs, usingObjectStoreIndex)
+}
+
+// FilterObjectStoreIndex returns a new slice retaining only the periodConfigs
+// that use object store index.
+func FilterObjectStoreIndex(in []PeriodConfig) []PeriodConfig {
+	var out []PeriodConfig
```
dannykopping, line 261:
```suggestion
	out := make([]PeriodConfig, 0, len(in))
```

Let's preallocate where possible to avoid unnecessary slice growth

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -98,13 +99,14 @@ type Config struct {
 // RegisterFlags registers flags.
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.WorkingDirectory, "boltdb.shipper.compactor.working-directory", "", "Directory where files can be downloaded for compaction.")
-	f.StringVar(&cfg.SharedStoreType, "boltdb.shipper.compactor.shared-store", "", "Shared store used for storing boltdb files. Supported types: gcs, s3, azure, swift, filesystem")
+	f.StringVar(&cfg.SharedStoreType, "boltdb.shipper.compactor.shared-store", "", "Shared store used for storing boltdb files. Supported types: gcs, s3, azure, swift, filesystem. If not set, compactor will be initialized to operate on all the object store indexes defined in the schema config.")
```
dannykopping, line 102: We'll need some documentation / upgrade guide docs on this pls

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -98,13 +99,14 @@ type Config struct {
...
 	f.DurationVar(&cfg.CompactionInterval, "boltdb.shipper.compactor.compaction-interval", 10*time.Minute, "Interval at which to re-run the compaction operation.")
 	f.DurationVar(&cfg.ApplyRetentionInterval, "boltdb.shipper.compactor.apply-retention-interval", 0, "Interval at which to apply/enforce retention. 0 means run at same interval as compaction. If non-zero, it should always be a multiple of compaction interval.")
 	f.DurationVar(&cfg.RetentionDeleteDelay, "boltdb.shipper.compactor.retention-delete-delay", 2*time.Hour, "Delay after which chunks will be fully deleted during retention.")
 	f.BoolVar(&cfg.RetentionEnabled, "boltdb.shipper.compactor.retention-enabled", false, "(Experimental) Activate custom (per-stream,per-tenant) retention.")
 	f.IntVar(&cfg.RetentionDeleteWorkCount, "boltdb.shipper.compactor.retention-delete-worker-count", 150, "The total amount of worker to use to delete chunks.")
+	f.StringVar(&cfg.DeleteRequestStore, "boltdb.shipper.compactor.delete-request-store", "", "Store used for managing delete requests. If not set, shared_store is used as a fallback.")
```
dannykopping, line 109: Do we need to even provide this as an option? Under what circumstances do you see a separate store being needed?
I do like that you're falling back to the shared store, though.

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -178,7 +186,13 @@ func NewCompactor(cfg Config, objectClient client.ObjectClient, schemaConfig con
 		defaultRetentionStats.Set(limits.DefaultLimits().RetentionPeriod.String())
 	}
 	if cfg.SharedStoreType == "" {
-		return nil, errors.New("compactor shared_store_type must be specified")
+		objectTypes := make([]string, len(objectClients))
+		i := 0
+		for key := range objectClients {
+			objectTypes[i] = key
+			i++
+		}
```
dannykopping, line 194:
```suggestion
		objectTypes := make([]string, 0, len(objectClients))
		for key := range objectClients {
			objectTypes = append(objectTypes, key)
		}
```

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -178,7 +186,13 @@ func NewCompactor(cfg Config, objectClient client.ObjectClient, schemaConfig con
...
+		level.Info(util_log.Logger).Log("msg", "compactor shared_store_type not specified. Initializing compactor to operator on the following object stores", "stores", objectTypes)
```
dannykopping, line 195: Does this get printed correctly? I'd imagine we'd need to concat all the values of the slice...?

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -227,53 +241,82 @@ func NewCompactor(cfg Config, objectClient client.ObjectClient, schemaConfig con
...
-		if err := c.initDeletes(r, limits); err != nil {
+		if err := retention.MoveMarkersToSharedStoreDir(filepath.Join(c.cfg.WorkingDirectory, "retention"), c.cfg.SharedStoreType); err != nil {
 			return err
 		}
+	}
 
-		c.tableMarker, err = retention.NewMarker(retentionWorkDir, c.expirationChecker, c.cfg.RetentionTableTimeout, chunkClient, r)
-		if err != nil {
-			return err
+	c.containers = make(map[string]storeContainer)
```
dannykopping, line 268: Please preallocate

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -227,53 +241,82 @@ func NewCompactor(cfg Config, objectClient client.ObjectClient, schemaConfig con
...
+			if _, ok := objectClient.(*local.FSObjectClient); ok {
+				encoder = client.FSEncoder
+			}
+
+			chunkClient := client.NewClient(objectClient, encoder, schemaConfig)
+
+			retentionWorkDir := filepath.Join(c.cfg.WorkingDirectory, "retention")
+			sc.sweeper, err = retention.NewSweeper(retentionWorkDir, objectType, chunkClient, c.cfg.RetentionDeleteWorkCount, c.cfg.RetentionDeleteDelay, r)
+			if err != nil {
+				return err
```
dannykopping, line 284: Wrap errors here please

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -506,8 +551,13 @@ func (c *Compactor) CompactTable(ctx context.Context, tableName string, applyRet
 		return fmt.Errorf("index processor not found for index type %s", schemaCfg.IndexType)
 	}
 
-	table, err := newTable(ctx, filepath.Join(c.cfg.WorkingDirectory, tableName), c.indexStorageClient, indexCompactor,
-		schemaCfg, c.tableMarker, c.expirationChecker, c.cfg.UploadParallelism)
+	sc, ok := c.containers[schemaCfg.ObjectType]
+	if !ok {
+		return fmt.Errorf("storeContainer not found for %s", schemaCfg.ObjectType)
```
dannykopping, line 556: Will a user be able to understand what this error means?

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -167,9 +166,18 @@ type Compactor struct {
...
+}
+
+type storeContainer struct {
+	tableMarker        retention.TableMarker
+	sweeper            *retention.Sweeper
+	indexStorageClient shipper_storage.Client
 }
 
-func NewCompactor(cfg Config, objectClient client.ObjectClient, schemaConfig config.SchemaConfig, limits *validation.Overrides, r prometheus.Registerer) (*Compactor, error) {
+func NewCompactor(cfg Config, objectClients map[string]client.ObjectClient, schemaConfig config.SchemaConfig, limits *validation.Overrides, r prometheus.Registerer) (*Compactor, error) {
```
dannykopping, line 180: Nit: I'm finding the use of the term "object" in this PR to be a bit confusing; these are object _store_ clients - so I think the more clear name would be `storeClients`, `storeType`, etc.

pkg/storage/stores/indexshipper/compactor/retention/retention.go
```diff
@@ -404,3 +408,29 @@ func (c *chunkRewriter) rewriteChunk(ctx context.Context, ce ChunkEntry, tableIn
...
 }
+
+// with multi-store support, markers need to be stored in store specific dir
+// MoveMarkersToSharedStoreDir checks for markers in retention dir and moves them as needed
+func MoveMarkersToSharedStoreDir(workingDir string, sharedStoreType string) error {
+	markersDir := filepath.Join(workingDir, markersFolder)
+	info, err := os.Stat(markersDir)
+	if os.IsNotExist(err) {
+		return nil
+	} else if err == nil && info.IsDir() {
```
dannykopping, line 419: No need for an `else` if you're returning in the alternate branch

dannykopping (commented):
I can't comment on the compaction behaviour because I'm not too familiar with it, but overall the changes LGTM.

I've left a few nits for clarity and idiomatic Go practices

pkg/storage/stores/indexshipper/compactor/compactor.go
ashwanthgoli: agree, I'll do the needful

pkg/storage/stores/indexshipper/compactor/compactor.go
ashwanthgoli: might not be helpful unless they grok the code. I'll try to add more context to the log message

pkg/storage/stores/indexshipper/compactor/compactor.go
ashwanthgoli: oh, I overlooked it. Will check

pkg/storage/stores/indexshipper/compactor/compactor.go
ashwanthgoli: It need not be a separate store, users can choose from one of the object stores used for index.

When it comes to index, we know where to route reads/writes because of the schema_config, but I believe the same cannot be done when storing delete requests as they don't have any period associated with them.

We could implicitly assume that delete requests should always go to the object store referred to in the latest schema_config entry, but this would result in compactor not processing any pending requests from older stores. I thought making this configurable would allow users to change it as they see fit.

I could also document this better

docs/sources/upgrading/_index.md
```diff
@@ -95,6 +95,21 @@ level=info ts=2022-12-20T15:27:54.858554127Z caller=metrics.go:147 component=fro
 
 These statistics are also displayed when using `--stats` with LogCLI.
 
+#### Multi-store support for compactor
+
+In releases prior to 2.7.4, setting `-boltdb.shipper.compactor.shared-store` configured the following:
+- store used for managing delete requests.
+- store on which index compaction should be performed.
+
+If `-boltdb.shipper.compactor.shared-store` not set, it defaults to the `object_store` configured in the latest `period_config` that uses either tsdb or boltdb-shipper index.
```
JStickler, line 104:
```suggestion
If `-boltdb.shipper.compactor.shared-store` is not set, it defaults to the `object_store` configured in the latest `period_config` that uses either the tsdb or boltdb-shipper index.
```

docs/sources/upgrading/_index.md
```diff
@@ -95,6 +95,21 @@ level=info ts=2022-12-20T15:27:54.858554127Z caller=metrics.go:147 component=fro
...
+
+In releases 2.7.4 and later, compactor supports index compaction on multiple stores.
```
JStickler, line 106:
```suggestion
In releases 2.7.4 and later, the Compactor supports index compaction on multiple stores.
```

docs/sources/upgrading/_index.md
```diff
@@ -95,6 +95,21 @@ level=info ts=2022-12-20T15:27:54.858554127Z caller=metrics.go:147 component=fro
...
+If `-boltdb.shipper.compactor.shared-store` is set, the two points mentioned above would still hold true to ensure backward compatibility. But loki will not set any defaults on this config, if not set compaction would be performed on all the object stores that contain either boltdb-shipper or tsdb index.
```
JStickler, line 107:
```suggestion
If `-boltdb.shipper.compactor.shared-store` is set, the two points mentioned above would still hold true to ensure backward compatibility. But Loki will not set any defaults for this setting. If no configuration is specified, compaction will be performed on all the object stores that contain either a boltdb-shipper or tsdb index.
```

docs/sources/upgrading/_index.md
```diff
@@ -95,6 +95,21 @@ level=info ts=2022-12-20T15:27:54.858554127Z caller=metrics.go:147 component=fro
...
+
+A new config option `-boltdb.shipper.compactor.delete-request-store` is added that decides where delete requests should be stored. If `-boltdb.shipper.compactor.shared-store` is set, it takes precedence over `-boltdb.shipper.compactor.delete-request-store` to decide the store for delete requests.
```
JStickler, line 109:
```suggestion
A new config option `-boltdb.shipper.compactor.delete-request-store` decides where delete requests should be stored. If `-boltdb.shipper.compactor.shared-store` is set, it takes precedence over `-boltdb.shipper.compactor.delete-request-store` to decide the store for delete requests.
```

docs/sources/upgrading/_index.md
```diff
@@ -95,6 +95,21 @@ level=info ts=2022-12-20T15:27:54.858554127Z caller=metrics.go:147 component=fro
...
+
+`-boltdb.shipper.compactor.delete-request-store` defaults to the `object_store` configured in the latest `period_config` that uses either tsdb or boltdb-shipper index. This is to ensure any pending delete requests in the latest period are handled - useful for the case where `-boltdb.shipper.compactor.shared-store` was not previously set.
```
JStickler, line 111:
```suggestion
`-boltdb.shipper.compactor.delete-request-store` defaults to the `object_store` configured in the latest `period_config` that uses either a tsdb or boltdb-shipper index. This is to ensure any pending delete requests in the latest period are handled.  This is useful for the case where `-boltdb.shipper.compactor.shared-store` was not previously set.
```

JStickler (commented):
[Doc squad] I had a couple of small suggestions for wording.

docs/sources/configuration/_index.md
```diff
@@ -1917,6 +1919,11 @@ The `compactor` block configures the compactor component, which compacts index s
 # CLI flag: -boltdb.shipper.compactor.retention-table-timeout
 [retention_table_timeout: <duration> | default = 0s]
 
+# Store used for managing delete requests. If not set,
+# -boltdb.shipper.compactor.shared-store is used as fallback.
```
JStickler, line 1923: "as fallback"  Would "default" be a better word choice, since fallback implies something that you use in an emergency?

docs/sources/upgrading/_index.md
```diff
@@ -95,6 +95,21 @@ level=info ts=2022-12-20T15:27:54.858554127Z caller=metrics.go:147 component=fro
 
 These statistics are also displayed when using `--stats` with LogCLI.
 
+#### Multi-store support for compactor
+
+In releases prior to 2.7.4, setting `-boltdb.shipper.compactor.shared-store` configured the following:
```
JStickler, line 100: Not sure when this PR is going live/getting released, do we need to update 2.7.4 to read 2.7.5 now?

docs/sources/upgrading/_index.md
```diff
@@ -95,6 +95,21 @@ level=info ts=2022-12-20T15:27:54.858554127Z caller=metrics.go:147 component=fro
...
+- store used for managing delete requests.
+- store on which index compaction should be performed.
+
+If `-boltdb.shipper.compactor.shared-store` is not set, it defaults to the `object_store` configured in the latest `period_config` that uses either the tsdb or boltdb-shipper index.
+
+In releases 2.7.4 and later, the Compactor supports index compaction on multiple stores.
```
JStickler, line 106: Does the release version need updating now that 2.7.4 has gone out?

JStickler (approved):
[Docs squad}  Documentation LGTM.  Couple of small suggestions.

docs/sources/upgrading/_index.md
ashwanthgoli: agree, I'll update it

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -226,53 +234,76 @@ func NewCompactor(cfg Config, objectClient client.ObjectClient, schemaConfig con
...
-		c.sweeper, err = retention.NewSweeper(retentionWorkDir, chunkClient, c.cfg.RetentionDeleteWorkCount, c.cfg.RetentionDeleteDelay, r)
-		if err != nil {
-			return err
+		if err := c.initDeletes(objectClient, r, limits); err != nil {
+			return fmt.Errorf("failed to init delete store: %w", err)
 		}
 
-		if err := c.initDeletes(r, limits); err != nil {
-			return err
+		if err := retention.MigrateMarkers(filepath.Join(c.cfg.WorkingDirectory, "retention"), deleteRequestStore); err != nil {
```
sandeepsukhani, line 267: deletion markers might not always belong to objects in the delete requests store. There is a high chance when someone wants to start using a new bucket, they would change the bucket for delete requests store as well. In that case we will try deleting objects from new bucket here while the objects were stored in older bucket. Will think more and see what we can do here.

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -226,53 +234,76 @@ func NewCompactor(cfg Config, objectClient client.ObjectClient, schemaConfig con
...
-	c.indexStorageClient = shipper_storage.NewIndexStorageClient(objectClient, c.cfg.SharedStoreKeyPrefix)
-	c.metrics = newMetrics(r)
 
 	if c.cfg.RetentionEnabled {
-		var encoder client.KeyEncoder
-		if _, ok := objectClient.(*local.FSObjectClient); ok {
-			encoder = client.FSEncoder
+		deleteRequestStore := c.cfg.DeleteRequestStore
+		// if -boltdb.shipper.compactor.shared-store is set, use it instead to ensure backward compatibility.
+		if c.cfg.SharedStoreType != "" {
```
sandeepsukhani, line 254: I think we should give precedence to `c.cfg.DeleteRequestStore` since it is a new and explicit config. Any change in that config would have been made by user deliberately. Also, it looks inconsistent not honoring an explicit config.

sandeepsukhani (commented):
overall changes look good to me with the initial pass. I need to think more on how we approach the deletion markers migration.

pkg/storage/config/schema_config.go
```diff
@@ -305,17 +305,15 @@ func usingForPeriodConfigs(configs []PeriodConfig, fn func(PeriodConfig) bool) b
...
-	}
+// UsingObjectStoreIndex returns true if the period config uses either boltdb-shipper or tsdb index.
+func UsingObjectStoreIndex(cfg PeriodConfig) bool {
+	return cfg.IndexType == BoltDBShipperType || cfg.IndexType == TSDBType
+}
 
-	return usingForPeriodConfigs(configs, fn)
+// ContainsObjectStorageIndex returns true if the current or any of the upcoming periods
+// use an object store index.
+func ContainsObjectStorageIndex(configs []PeriodConfig) bool {
```
sandeepsukhani, line 315: I called it `UsingObjectStoreIndex` because it checked whether current or upcoming index is an object storage based index. This new name makes it confusing. Maybe swap `UsingObjectStoreIndex` and `ContainsObjectStorageIndex` names?

pkg/storage/stores/indexshipper/compactor/retention/retention.go
```diff
@@ -424,3 +427,50 @@ func (c *chunkRewriter) rewriteChunk(ctx context.Context, ce ChunkEntry, tableIn
 
 	return wroteChunks, linesDeleted, nil
 }
+
+// since compactor supports multiple stores, markers need to be written to store specific dir.
+// MigrateMarkers checks for markers in retention dir and migrates them.
+func MigrateMarkers(workingDir string, store string) error {
```
sandeepsukhani, line 433: Let us be more specific and call it `CopyMarkers`? Also maybe make the comment more descriptive saying we are copying markers to store specific marker directory?

sandeepsukhani (approved):
The changes look great. Just have some non-blocking changes for code readablitiy so just approving the PR. Will merge it once the feedback has been addressed.

sandeepsukhani (approved):
LGTM

PR #7426, opened 2022-10-14
Add compactor delete client tls client config

periklis:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

**Checklist**
- [x] Reviewed the `CONTRIBUTING.md` guide
- [x] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #7453, opened 2022-10-18
Add single compactor http client for delete and gennumber clients

periklis:
This PR completes work started with #7426 as there are two types of http clients for using the compactor retention API, one for the deletes and one for the generation numbers. However #7426 introduced by a hindsight on the existence of generation number client the `boltdb.shipper.compactor.delete_client` which is replaced in this PR with a single `boltdb.shipper.compactor.client`. Latter enables to set the TLS settings for the underlying http client when calling the compactor retention API.

Furthermore this change set includes also a fix for enabling the generation number middlewares only if the compactor retention API is enabled. Currently this produced non-blocking but confusing 404 errors like the following on each query and rule execution:

```
level=error ts=2022-10-18T14:13:41.598649348Z caller=delete_requests_client.go:211 msg="error getting delete requests from the store" err="unexpected status code: 404"
ts=2022-10-18T14:13:41.598697295Z caller=spanlogger.go:80 user=application level=error msg="failed loading deletes for user" err="unexpected status code: 404"
```

Fixes #<issue number>

Added backport label to fix the CLI flags from `boltdb.shipper.compactor.delete_client` to `boltdb.shipper.compactor.client` before we go with #7455 out or soon after. (cc @MichelHollands @trevorwhitney)

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -350,7 +350,7 @@ func (t *Loki) initQuerier() (services.Service, error) {
 	toMerge := []middleware.Interface{
 		httpreq.ExtractQueryMetricsMiddleware(),
 	}
-	if t.supportIndexDeleteRequest() {
+	if t.supportIndexDeleteRequest() && t.Cfg.CompactorConfig.RetentionEnabled {
 		toMerge = append(
 			toMerge,
 			queryrangebase.CacheGenNumberHeaderSetterMiddleware(t.cacheGenerationLoader),
@@ -660,7 +660,8 @@ func (t *Loki) initQueryFrontendTripperware() (_ services.Service, err error) {
 		t.Cfg.QueryRange,
 		util_log.Logger,
 		t.overrides,
-		t.Cfg.SchemaConfig, t.cacheGenerationLoader,
+		t.Cfg.SchemaConfig,
+		t.cacheGenerationLoader, t.Cfg.CompactorConfig.RetentionEnabled,
 		prometheus.DefaultRegisterer,
 	)
 	if err != nil {
@@ -679,7 +680,13 @@ func (t *Loki) initCacheGenerationLoader() (_ services.Service, err error) {
 		if err != nil {
 			return nil, err
 		}
-		client, err = generationnumber.NewGenNumberClient(compactorAddress, &http.Client{Timeout: 5 * time.Second})
+
+		httpClient, err := compactor.NewCompactorHTTPClient(t.Cfg.CompactorClient)
+		if err != nil {
+			return nil, err
+		}
+
+		client, err = generationnumber.NewGenNumberClient(compactorAddress, httpClient)
 		if err != nil {
 			return nil, err
 		}
@@ -1112,7 +1119,7 @@ func (t *Loki) initUsageReport() (services.Service, error) {
 }
 
 func (t *Loki) deleteRequestsClient(clientType string, limits *validation.Overrides) (deletion.DeleteRequestsClient, error) {
-	if !t.supportIndexDeleteRequest() {
+	if !t.supportIndexDeleteRequest() || !t.Cfg.CompactorConfig.RetentionEnabled {
 		return deletion.NewNoOpDeleteRequestsStore(), nil
 	}
 
@@ -1121,7 +1128,7 @@ func (t *Loki) deleteRequestsClient(clientType string, limits *validation.Overri
 		return nil, err
 	}
 
-	httpClient, err := deletion.NewDeleteHTTPClient(t.Cfg.DeleteClient)
+	httpClient, err := compactor.NewCompactorHTTPClient(t.Cfg.CompactorClient)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/querier/queryrange/queryrangebase/results_cache.go
+++ b/pkg/querier/queryrange/queryrangebase/results_cache.go
@@ -163,6 +163,7 @@ type resultsCache struct {
 	merger               Merger
 	cacheGenNumberLoader CacheGenNumberLoader
 	shouldCache          ShouldCacheFn
+	retentionEnabled     bool
 	metrics              *ResultsCacheMetrics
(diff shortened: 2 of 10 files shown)


trevorwhitney:
@periklis thanks for adding the backport label, aiming on end of week for actually cutting 2.7 so should be plenty of time to get this in.

pkg/storage/stores/indexshipper/compactor/compactor_client.go
```diff
@@ -0,0 +1,42 @@
...
+// Config for compactor's generation-number client
+type ClientConfig struct {
+	TLSEnabled bool             `yaml:"tls_enabled"`
+	TLS        tls.ClientConfig `yaml:",inline"`
+}
+
+// RegisterFlags adds the flags required to config this to the given FlagSet.
+func (cfg *ClientConfig) RegisterFlags(f *flag.FlagSet) {
+	prefix := "boltdb.shipper.compactor.client"
+	f.BoolVar(&cfg.TLSEnabled, prefix+".tls-enabled", false,
```
xperimental, line 20: Is the separate `tls-enabled` configuration necessary? If I remember correctly, then the configuration contains the whole URL to the compactor including the scheme, so that should be enough for the client to decide whether to use TLS or not.

periklis:
@trevorwhitney Reminder about this one

pkg/loki/modules.go
```diff
@@ -395,7 +395,7 @@ func (t *Loki) initQuerier() (services.Service, error) {
 	toMerge := []middleware.Interface{
 		httpreq.ExtractQueryMetricsMiddleware(),
 	}
-	if t.supportIndexDeleteRequest() {
+	if t.supportIndexDeleteRequest() && t.Cfg.CompactorConfig.RetentionEnabled {
```
trevorwhitney, line 398: what's this additional check for, why wasn't it previously needed? You check it in combination with `t.supportIndexDeleteRequests` in multiple places so I'm curious if it should be moved into the `t.supportIndexDeleteRequests` function?

pkg/loki/modules.go
```diff
@@ -1160,7 +1167,7 @@ func (t *Loki) initUsageReport() (services.Service, error) {
 }
 
 func (t *Loki) deleteRequestsClient(clientType string, limits *validation.Overrides) (deletion.DeleteRequestsClient, error) {
-	if !t.supportIndexDeleteRequest() {
+	if !t.supportIndexDeleteRequest() || !t.Cfg.CompactorConfig.RetentionEnabled {
```
trevorwhitney, line 1170: again, this seems to add to the argument this additional check should be moved into that function.

pkg/querier/queryrange/queryrangebase/results_cache.go
```diff
@@ -214,7 +217,7 @@ func (s resultsCache) Do(ctx context.Context, r Request) (Response, error) {
 		return s.next.Do(ctx, r)
 	}
 
-	if s.cacheGenNumberLoader != nil {
+	if s.cacheGenNumberLoader != nil && s.retentionEnabled {
```
trevorwhitney, line 220: again I'm not quite sure why we are having to add this check now but it wasn't needed before? seems out of scope with adding TLS support to the client, so I'm curious what motivated it?

trevorwhitney (commented):
Looks good Peri! One question and one suggestions:

* why did the simplification of the client require so much additional checking of whether or not retention was enabled that wasn't previously needed? My assumption here is that previously we were creating the client but just not using it when retention was not enabled, but I'm assuming something about this change makes that problematic?
* how would you feel about moving the `t.CfgCompactorConfig.RetentionEnabled` check into `t.supportIndexDeleteRequests`, or are there cases we would support the index delete requests even if retention was not enabled?

pkg/loki/modules.go
periklis: As mentioned above in the description this extra check is needed when the deletes-based retention is entirely disabled. This is the case by default IIRC and produces red-herring errors in the querier and ruler logs like this:
```
level=error ts=2022-10-18T14:13:41.598649348Z caller=delete_requests_client.go:211 msg="error getting delete requests from the store" err="unexpected status code: 404"
ts=2022-10-18T14:13:41.598697295Z caller=spanlogger.go:80 user=application level=error msg="failed loading deletes for user" err="unexpected status code: 404"
```

I found moving it into the `supportIndexDeleteRequests` obscures the fact that this method is dedicated to check if the underlying index store supports deletes. Adding `RetentionEnabled` in there would make the check for retention hidden/implicit with the store capabilities. OTH as proposed in the implementation it is more explicit for the reader to know that we need an index store that supports deletes and explicitly opt-in to use the feature.

pkg/querier/queryrange/queryrangebase/results_cache.go
periklis: Although it seems out of scope, but on the other hand fixing TLS makes the 404s visible in the querier/ruler logs. I know if you run the entire system withough TLS you see the 404s anyway if retention is disabled. In any case both bits missing TLS support and 404s on disabled retentions are hindsights of the compactor client. IMHO the PR is just removing these hindsights and we may need to just re-title/change the changelog entry to be more informative in that direction. WDYT?

pkg/loki/modules.go
periklis: Same as above, it is more explicit for the reader that we need an index store that supports deletes and explicitly opt-in to use the feature.

pkg/querier/queryrange/queryrangebase/results_cache.go
trevorwhitney: > fixing TLS makes the 404s visible in the querier/ruler log

sounds like good enough justification to me! thanks for that added context.

trevorwhitney (approved):
Thanks for the added context. LGTM!

PR #7469, opened 2022-10-19
Loki: log when a query starts in the logql engine

slim-bean:

There are cases where a query cannot finish (OOM of a querier/ruler) and most of our logs are emitted upon completion of a query. This can help us find queries which are OOM crashing a component.

Note: It only covers queries executed by the engine, so only instant and range queries. Another PR will be necessary to cover labels/series endpoints. However this approach seemed the best to be able to get this information from ruler executed queries.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/engine.go
+++ b/pkg/logql/engine.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	log2 "github.com/grafana/loki/pkg/util/log"
 	"math"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/logql/engine.go
```diff
@@ -192,6 +193,12 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
 	log, ctx := spanlogger.New(ctx, "query.Exec")
 	defer log.Finish()
 
+	if GetRangeType(q.params) == InstantType {
```
DylanGuedes, line 196: WDYT of adding a runtime config (like log_queries) and only logging this in case it is enabled? not sure how costly it will be to always log all queries

slim-bean: The challenge here is we often don't know the tenant that is responsible for the OOM's so we'd need a way to enable it for everyone.

We already log metrics.go with a result for every query at info level so it seemed reasonable to me to log the start of the query also at info level.

I think this is an acceptable tradeoff, we can always revisit though.

dannykopping: Ooh I like that idea, then we can log all query types

dannykopping: @slim-bean what's the rationale around only enabling for instant queries? Surely we also see OOMs for large metric queries sometimes? This will definitely help in the case of rulers though, so hell yeah from that perspective

DylanGuedes: Fair enough. I like that we're starting with level=info so that a new rollout isn't necessary in case you need to debug something. On the other hand, my concern with logging all queries is that we'll end up logging even subqueries AFAIK, so the volume will be huge I suppose.

Anyway, LGTM!

For the future, would be really cool if we could have a runtime configuration section that although loaded in runtime isn't per-tenant. Would be really helpful for scenarios similar to this one.

DylanGuedes (approved):

owen-d (approved):
LGTM

PR #7480, opened 2022-10-20
Remove prometheus CRDs

jkroepke:

Remove Prometheus CRDs from loki chart.

Why?

CRDs are not fully managed by helm. If loki chart is installed before kube-prometheus-stack helm chart, ServiceMonitor and PrometheusRule CRD are not in sync/still outdated.

In umbrella chart scenarios, where kube-prometheus-stack and loki are bundled to one helm charts, the versions of the installed CRDs maybe unexpectable. (No one knows, if the CRD are picked from loki or kube-prometheus-stack first) 

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated


production/helm/loki/values.yaml
```diff
@@ -545,7 +545,7 @@ monitoring:
   # Recording rules for monitoring Loki, required for some dashboards
   rules:
     # -- If enabled, create PrometheusRule resource with Loki recording rules
-    enabled: true
+    enabled: false
```
trevorwhitney, line 548: why are we changing the defaults here? self-monitoring should be on by default, that is a goal of this chart.

production/helm/loki/values.yaml
```diff
@@ -579,7 +579,7 @@ monitoring:
   # ServiceMonitor configuration
   serviceMonitor:
     # -- If enabled, ServiceMonitor resources for Prometheus Operator are created
-    enabled: true
+    enabled: false
```
trevorwhitney, line 582: again here, self monitoring enabled should be the default.

trevorwhitney (changes requested):
I am open to removing the CRDs, as it feels hacky to me as well, but we need to keep self-monitoring on by default.

production/helm/loki/values.yaml
jkroepke: From my point of view, its common practice is disable custom resources by default. Helm charts should not depend agaist custom resources unless the CRDs are managed by the helm chart.

The [grafana ](https://github.com/grafana/helm-charts/blob/2cf626aad06a12506c014aa541f85b8646f0916c/charts/grafana/values.yaml#L173)and [tempo](https://github.com/grafana/helm-charts/blob/2cf626aad06a12506c014aa541f85b8646f0916c/charts/tempo/values.yaml#L199) chart are follow this principle, too.

I can revert both mentioned changes to get an non working configuration by default. This is fine for me.

---

But include foreign CRDs feel really hackly, since helm CRD behavior is unexpectable.

jkroepke: Please take a look at fe5dfbf, if this fits for you.

production/helm/loki/CHANGELOG.md
```diff
@@ -11,7 +11,13 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
-## 3.2.1 
+## 3.3.0
+
+- [CHANGE] Remove ServiceMonitor and PrometheusRule CRD
+- [CHANGE] Set `monitoring.serviceMonitor.enabled` to `false` by default.
+- [CHANGE] Set `monitoring.rules.enabled` to `false` by default.
```
trevorwhitney, line 18: no longer true
```suggestion
```

trevorwhitney (approved):
I prefer this approach, thank you!

jkroepke:
I remove by fork by accident. #7499 is a new PR

PR #7481, opened 2022-10-20
[mixin] Add cortex-gw-internal to dashboards

gouthamve:
We are now accepting data through the internal gateway as well. This updates the dashboards to reflect that.

No alerts or recording rules reference `cortex-gw`. 

Does this need a CHANGELOG entry?

### DIFF
--- a/production/loki-mixin/dashboards/dashboard-utils.libsonnet
+++ b/production/loki-mixin/dashboards/dashboard-utils.libsonnet
@@ -167,7 +167,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
       tooltip: { sort: 2 },  // Sort descending.
     },
   containerCPUUsagePanel(title, containerName)::
-    self.CPUUsagePanel(title, 'container="%s"' % containerName),
+    self.CPUUsagePanel(title, 'container=~"%s"' % containerName),
 
   memoryWorkingSetPanel(title, matcher)::
     $.panel(title) +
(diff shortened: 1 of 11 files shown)


dannykopping (commented):
I'm a little bit unsure why we have `cortex-gw` in our mixins at all since this is proprietary Grafana Labs middleware; I'm not sure we should be adding the other variant here too.

I'll raise this with the team tomorrow

production/loki-mixin/dashboards/dashboard-utils.libsonnet
```diff
@@ -167,7 +167,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
       tooltip: { sort: 2 },  // Sort descending.
     },
   containerCPUUsagePanel(title, containerName)::
-    self.CPUUsagePanel(title, 'container="%s"' % containerName),
+    self.CPUUsagePanel(title, 'container=~"%s"' % containerName),
```
dannykopping, line 170: Curious why these changes were made? They seem unrelated to the purpose of the PR.
The compiled versions generally seem to not need the regex, and I'm concerned about containers like `query-scheduler` where `-` has a semantic meaning in regex.

dannykopping (approved):
Left a nit, approving to unblock
We will remove all mention of cortex-gw from our mixin, and move this into our own tooling

production/loki-mixin/dashboards/dashboard-utils.libsonnet
gouthamve: You're right! Fixing it.

dannykopping:
@gouthamve can you please recompile the mixins to make CI happy?
I'll merge after that

gouthamve:
@dannykopping Its now done! Thanks!

PR #7492, opened 2022-10-21
Improve rate store refresh latency

MasslessParticle:
Rather than iterating streams and getting rates per stream, keep all necessary information in the rate calculator and just hand back the current rates when asked.

This change took p99 latencies from ~1.8s to ~400ms and p50 latences from ~1s to 6ms.

### DIFF
--- a/pkg/ingester/stream_rate_calculator.go
+++ b/pkg/ingester/stream_rate_calculator.go
@@ -3,6 +3,8 @@ package ingester
 import (
 	"sync"
 	"time"
+
+	"github.com/grafana/loki/pkg/logproto"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


pkg/ingester/stream_rate_calculator.go
```diff
@@ -58,30 +63,57 @@ func (c *StreamRateCalculator) updateLoop() {
...
+}
 
-	c.locks[i].RLock()
-	defer c.locks[i].RUnlock()
+func (c *StreamRateCalculator) Rates() []*logproto.StreamRate {
+	c.rateLock.RLock()
+	defer c.rateLock.RUnlock()
+
+	rates := make([]*logproto.StreamRate, 0, len(c.allRates))
+	for _, r := range c.allRates {
```
DylanGuedes, line 95: hmm can't you just use `copy(src, dst)` here? if so, do you know if it is more performatic than iterating over etc?

pkg/ingester/stream_rate_calculator.go
```diff
@@ -58,30 +63,57 @@ func (c *StreamRateCalculator) updateLoop() {
 }
 
 func (c *StreamRateCalculator) updateRates() {
+	rates := make([]logproto.StreamRate, 0, defaultStripeSize)
```
DylanGuedes, line 66: maybe c.size instead of defaultstripesize?

DylanGuedes (approved):
lgtm, just 2 nits

pkg/ingester/stream_rate_calculator.go
MasslessParticle: We can't `copy...` here because we're changing the  type from `logproto.StreamRate` to `*logproto.StreamRate`. I'm doing it this way so no pointers are ever shared among instances (I kept making memory leaks)

DylanGuedes: oops, thanks! :smile:

PR #7490, opened 2022-10-21
Indexshipper revert locking issues

owen-d:
*Note: Reverting these will reintroduce an older bug (race condition around mmap access), but with less complexity than the current bug. I will followup with a second PR that fixes our underlying issue as an accompanying PR*

This reverts two locking related PRs in the indexshipper. They both attempted to introduce safe locking practices, but the problem is structural and neither fixed the underlying problem. The issue occurs when

```
1) ForEach
  * `RLock`s table1, acquires indexset: tenant="1", the post-compacted user.
    * `Lock`s indexset for tenant="1", the post compacted user
    * Does not yet call `RUnlock`, which will be called at end of query
  * `RUnlock`s table1
2) Sync
  * `RLock`s table1
    * _Blocks_ calling `Lock` on the tenant="1" indexset (still held from (1)
3) ForEach
  * _Blocks_ `RLock`ing table1, intending to acquire the tenant="" indexset for the pre-compacted users
4) Deadlock!

This is because the locking pattern looks like:


* Sync
  * RLock table
    * Lock indexset
    * Unlock indexset
  * RUnlock Table

* ForEach
  * RLock table
  * RUnlock table
    * RLock indexset
    * RUnlock indexset at end of query
```

Note: This can also happen in any order of lock acquisitions on the indexSets because `RWMutex` prioritized write locks:
```
// If a goroutine holds a RWMutex for reading and another goroutine might
// call Lock, no goroutine should expect to be able to acquire a read lock
// until the initial read lock is released. In particular, this prohibits
// recursive read locking. This is to ensure that the lock eventually becomes
// available; a blocked Lock call excludes new readers from acquiring the
// lock.
```

A recent goroutine capture during this deadlock shows:
* Goroutines waiting for read lock on the indexset
```
goroutine profile: total 34599
32434 @ 0x43ccf6 0x44dc7e 0x44dc55 0x46a285 0x1c2af6d 0x1c2af48 0x1c29eff 0x1c2d62e 0x1c4062b 0x1c42723 0x1a430af 0x1c42175 0x1c49bb0 0xbbd99a 0x1c49b05 0x1c30ac7 0xf10570 0x1fbffe2 0xb2a71a 0x1d2bc50 0xb2a71a 0xbe7022 0x1e9f166 0xb2a71a 0xbe762a 0xb2a71a 0xb2e27b 0xb2a71a 0xbe8ac2 0xb2a71a 0xb2a5be 0xae4fdb
[...]

PR #7497, opened 2022-10-22
Index gw lock bugfix

owen-d:
While I'm still in favor of a longer refactor to address the complexity described in https://github.com/grafana/loki/pull/7490, I'm opening this PR as a shorter, more immediate alternative that should fix the deadlocks. This also reduces the number of total lock calls in many cases as we lock to act over entire groups rather than individual items.

### DIFF
--- a/pkg/storage/stores/indexshipper/downloads/table.go
+++ b/pkg/storage/stores/indexshipper/downloads/table.go
@@ -145,15 +145,19 @@ func (t *table) Close() {
 }
 
 func (t *table) ForEach(ctx context.Context, userID string, doneChan <-chan struct{}, callback index.ForEachIndexCallback) error {
+	ids := []string{userID, ""}
+	sets, err := t.getOrCreateIndexSets(ctx, ids, true)
+	if err != nil {
+		return err
+	}
+
 	// iterate through both user and common index
-	for _, uid := range []string{userID, ""} {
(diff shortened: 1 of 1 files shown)


pkg/storage/stores/indexshipper/downloads/table.go
```diff
@@ -347,18 +382,19 @@ func (t *table) EnsureQueryReadiness(ctx context.Context, userIDs []string) erro
...
+	sets, err := t.getOrCreateIndexSets(ctx, userIDs, false)
+	if err != nil {
+		return err
+	}
 
+	return concurrency.ForEachJob(ctx, len(userIDs), maxDownloadConcurrency, func(ctx context.Context, idx int) error {
+		indexSet := sets[idx]
 		return indexSet.AwaitReady(ctx)
 	})
 }
```
sandeepsukhani, line 394: we bound the number of index sets we download at a time to `maxDownloadConcurrency`. This change would cause us to download all of them simultaneously. If there are too many index sets to download, it would not leave a lot of resources for the query operation.

`indexSet.Init` function initiates the download operation, and `indexSet.AwaitReady` just blocks the function until the `indexSet` gets ready to query after the initial download of the index. I think it would be better to bind the number of index sets we download at a time.

slim-bean (approved):
LGTM

PR #7499, opened 2022-10-23
Remove prometheus CRDs

jkroepke:

Remove Prometheus CRDs from loki chart.

Why?

CRDs are not fully managed by helm. If loki chart is installed before kube-prometheus-stack helm chart, ServiceMonitor and PrometheusRule CRD are not in sync/still outdated.

In umbrella chart scenarios, where kube-prometheus-stack and loki are bundled to one helm charts, the versions of the installed CRDs maybe unexpectable. (No one knows, if the CRD are picked from loki or kube-prometheus-stack first) 

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated


jkroepke:
I remove my fork by accident, it replaces https://github.com/grafana/loki/pull/7480

jkroepke:
> Would it help if you could disable the CRDs?

There is no away to disable single CRDs. I could run `helm install --skip-crds`, but that would break the Installation of Grafana Agent Operator, since the CRDs manadory.

On top, there are issues with having the loki chart as part of an umbrella chart. For example together with the kube-prometheus-stack.

---

At the end, I do not get the point here. What the point to create service Monitors without having `kube-prometheus-stack` installed?

jeschkies (approved):
Sorry. I misinterpreted this change. I thought you've removed the `ServiceMonitor` CRD `kind: ServiceMonitor` when you are removing `kind: CustomResourceDefiniton`.

trevorwhitney:
@jeschkies ahh, yes, definitely still need our `ServiceMonitor`s :)

jkroepke:
No, I only want to remove the CRDs, since they should be owned by the kube-prometheus-stack helm chart.

The kind: ServiceMonitor manifeste are totally fine here.

trevorwhitney:
@jkroepke I think the Grafana Agent chart is a slightly different problem, because the Grafana Agent can act on service monitor resources independent of a prometheus operator. So, in a deployment where you only are using Grafana Agent, that chart needs to include those CRDs

PR #7500, opened 2022-10-23
[Helm] Add support for azure blob storage

jkroepke:
Support Azure Blob Storage in helm chart. See https://grafana.com/docs/loki/latest/storage/#azure-storage-account-1

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -204,7 +204,23 @@ gcs:
   bucket_name: {{ $.Values.loki.storage.bucketNames.chunks }}
   chunk_buffer_size: {{ .chunkBufferSize }}
   request_timeout: {{ .requestTimeout }}
-  enable_http2: {{ .enableHttp2}}
+  enable_http2: {{ .enableHttp2 }}
+{{- end -}}
+{{- else if eq .Values.loki.storage.type "azure" -}}
+{{- with .Values.loki.storage.azure }}
+azure:
+  account_name: {{ .accountName }}
+  {{- with .accountKey }}
+  account_key: {{ . }}
+  {{- end }}
+  container_name: {{ $.Values.loki.storage.bucketNames.chunks }}
+  use_managed_identity: {{ .useManagedIdentity }}
+  {{- with .userAssignedId }}
+  user_assigned_id: {{ . }}
+  {{- end }}
+  {{- with .requestTimeout }}
+  request_timeout: {{ . }}
+  {{- end }}
 {{- end -}}
 {{- else -}}
 {{- with .Values.loki.storage.filesystem }}
@@ -253,7 +269,23 @@ gcs:
   bucket_name: {{ $.Values.loki.storage.bucketNames.ruler }}
   chunk_buffer_size: {{ .chunkBufferSize }}
   request_timeout: {{ .requestTimeout }}
-  enable_http2: {{ .enableHttp2}}
+  enable_http2: {{ .enableHttp2 }}
+{{- end -}}
+{{- else if eq .Values.loki.storage.type "azure" -}}
+{{- with .Values.loki.storage.azure }}
+azure:
+  account_name: {{ .accountName }}
+  {{- with .accountKey }}
+  account_key: {{ . }}
+  {{- end }}
+  container_name: {{ $.Values.loki.storage.bucketNames.ruler }}
+  use_managed_identity: {{ .useManagedIdentity }}
+  {{- with .userAssignedId }}
+  user_assigned_id: {{ . }}
+  {{- end }}
+  {{- with .requestTimeout }}
+  request_timeout: {{ . }}
+  {{- end }}
 {{- end -}}
 {{- end -}}
 {{- end -}}
--- a/production/helm/loki/values.yaml
+++ b/production/helm/loki/values.yaml
@@ -123,13 +123,7 @@ loki:
       configs:
         - from: 2022-01-11
           store: boltdb-shipper
-          {{- if eq .Values.loki.storage.type "s3" }}
-          object_store: s3
-          {{- else if eq .Values.loki.storage.type "gcs" }}
-          object_store: gcs
-          {{- else }}
-          object_store: filesystem
-          {{- end }}
+          object_store: {{ .Values.loki.storage.type }}
           schema: v12
           index:
             prefix: loki_index_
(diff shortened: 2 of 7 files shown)


production/helm/loki/values.yaml
```diff
@@ -138,7 +140,7 @@ loki:
 
     ruler:
       storage:
-      {{- if or .Values.minio.enabled (eq .Values.loki.storage.type "s3") (eq .Values.loki.storage.type "gcs") }}
+      {{- if or .Values.minio.enabled (eq .Values.loki.storage.type "s3") (eq .Values.loki.storage.type "gcs") (eq .Values.loki.storage.type "azure") }}
```
trevorwhitney, line 143: probably getting time to move this into a template, but we can do that in a later PR.

production/helm/loki/templates/_helpers.tpl
```diff
@@ -248,6 +253,11 @@ gcs:
   request_timeout: {{ .requestTimeout }}
   enable_http2: {{ .enableHttp2}}
 {{- end -}}
+{{- else if eq .Values.loki.storage.type "azure" -}}
+{{- with .Values.loki.storage.azure }}
+azure:
+{{- toYaml . }}
```
trevorwhitney, line 259: What's the argument against following the same pattern as above, where we explicitly pull out the individual config fields that are valid?

What I like about this is it also encourages us to add a default section under `.Values.loki.storage.azure` so people know what options are available, since Loki only supports specific ones.

Would you be open to changing it to follow that pattern?

trevorwhitney (changes requested):
Thanks for adding this, definitely something we need!

How would you feel about changing this PR a bit to follow the same pattern we use for s3 and gcs where we provide defaults in `values.yaml` of all the available options, and then explicitly pick them off in the template rather than using `toYaml`?

jkroepke:
I hope, its fine now.

the serviceAccount label is needed for [workload identity](https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview).

production/helm/loki/values.yaml
jkroepke: please, later PR.

Tolsto:
@jkroepke As you've mentioned Azure Workload Identity: Were you able to test this successfully with Loki?

jkroepke:
> @jkroepke As you've mentioned Azure Workload Identity: Were you able to test this successfully with Loki?

[Yes](https://github.com/jkroepke/homelab/blob/main/aks/argocd/loki.yaml#L101). 

Workload Identity is not official supported by loki, since the Azure Go SDK is [not aware](https://github.com/Azure/azure-sdk-for-go/issues/15615) of Workload Identity.

But with the [Workload Identity Migration Sidecar](https://learn.microsoft.com/en-us/azure/aks/workload-identity-migration-sidecar), loki works well with Workload Identity using the user assigned managed identity approch.

Tolsto:
> > @jkroepke As you've mentioned Azure Workload Identity: Were you able to test this successfully with Loki?
> 
> [Yes](https://github.com/jkroepke/homelab/blob/main/aks/argocd/loki.yaml#L101).
> 
> Workload Identity is not official supported by loki, since the Azure Go SDK is [not aware](https://github.com/Azure/azure-sdk-for-go/issues/15615) of Workload Identity.
> 
> But with the [Workload Identity Migration Sidecar](https://learn.microsoft.com/en-us/azure/aks/workload-identity-migration-sidecar), loki works well with Workload Identity using the user assigned managed identity approch.

Oh, great. Didn't know about the sidecar yet. Just started to add native support for it to Loki...

jkroepke:
> > > @jkroepke As you've mentioned Azure Workload Identity: Were you able to test this successfully with Loki?
> > 
> > 
> > [Yes](https://github.com/jkroepke/homelab/blob/main/aks/argocd/loki.yaml?rgh-link-date=2022-10-27T22%3A03%3A16Z#L101).
> > Workload Identity is not official supported by loki, since the Azure Go SDK is [not aware](https://github.com/Azure/azure-sdk-for-go/issues/15615) of Workload Identity.
> > But with the [Workload Identity Migration Sidecar](https://learn.microsoft.com/en-us/azure/aks/workload-identity-migration-sidecar), loki works well with Workload Identity using the user assigned managed identity approch.
> 
> Oh, great. Didn't know about the sidecar yet. Just started to add native support for it to Loki...

The sidecar is a good workaround for now. It works for all components which are supporting user assigned MIs

Tolsto:
> > > > @jkroepke As you've mentioned Azure Workload Identity: Were you able to test this successfully with Loki?
> > > 
> > > 
> > > [Yes](https://github.com/jkroepke/homelab/blob/main/aks/argocd/loki.yaml?rgh-link-date=2022-10-27T22%3A03%3A16Z#L101).
> > > Workload Identity is not official supported by loki, since the Azure Go SDK is [not aware](https://github.com/Azure/azure-sdk-for-go/issues/15615) of Workload Identity.
> > > But with the [Workload Identity Migration Sidecar](https://learn.microsoft.com/en-us/azure/aks/workload-identity-migration-sidecar), loki works well with Workload Identity using the user assigned managed identity approch.
> > 
> > 
> > Oh, great. Didn't know about the sidecar yet. Just started to add native support for it to Loki...
> 
> The sidecar is a good workaround for now. It works for all components which are supporting user assigned MIs

On second thoughts, the sidecar doesn't work very well for me because the container requires privileged access to set up the redirection.
I've added a change that brings native Workload Identity support to Loki: https://github.com/grafana/loki/pull/7540

jkroepke:
@trevorwhitney rebased

jeschkies:
> I've added a change that brings native Workload Identity support to Loki: https://github.com/grafana/loki/pull/7540

Shall we block this change then until https://github.com/grafana/loki/pull/7540 landed?

jkroepke:
@jeschkies In out production system, we want to use SAS Token (.azure.accountKey) instead Managed Identities. 

I recommend that the PR #7540 should include the required changes for the helm chart. I'm bit tired of rebasing the branch.

DarChaos21:
We also need this feature. Any timeline on this? 

I appreciate your efforts.

jkroepke:
@trevorwhitney any updates on this PR?

jeschkies (approved):
Heck, if it works we should not wait for #7540

jeschkies:
@jkroepke you need to update the Helm Chart README by running `docker run --rm --volume "$(pwd):/helm-docs" -u "$(id -u)" jnorwood/helm-docs:v1.8.1`inside `production/helm/loki`.

jeschkies:
@jkroepke would you follow up on @trevorwhitney's comments above if required?

PR #7516, opened 2022-10-25
allow optionally setting extend-writes

cstyan:
Fixes higher than anticipated 500s on writes when using zone aware ingesters.


cstyan:
closing, since we decided to go with https://github.com/grafana/loki/pull/7517 instead

PR #3163, opened 2021-01-12
Uses custom json-iter decoder for log entries.

cyriltovena:
Previously we were using json.Unmarshal for each line. However json-iter uses a Pool for each calls and I believe this can cause to increase memory usage.

For each line we would put in a pool the iterator to re-use it, once put in a pool, the last data is retained, since we handle millions of lines, this can cause problem, using a custom extensions, keep using a pool but at the root object only, not for each line.

On top of that we're going to process that json payload 50% faster.

```
❯ benchcmp  before.txt after.txt2
benchmark                          old ns/op     new ns/op     delta
Benchmark_DecodePushRequest-16     13509236      6677037       -50.57%
benchmark                          old allocs     new allocs     delta
Benchmark_DecodePushRequest-16     106149         38719          -63.52%
benchmark                          old bytes     new bytes     delta
Benchmark_DecodePushRequest-16     10350362      5222989       -49.54%
```

Signed-off-by: Cyril Tovena <cyril.tovena@gmail.com>

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

PR #7522, opened 2022-10-26
Improves performance of json responses

cyriltovena:

This allows to reuse stream memory allocations between responses. Related https://github.com/grafana/loki/pull/3163 but this time this is for the encoding.

```
name        old time/op    new time/op    delta
_Encode-16    29.2µs ±12%    25.2µs ± 1%  -13.85%  (p=0.016 n=5+4)

name        old alloc/op   new alloc/op   delta
_Encode-16    24.9kB ± 6%    16.4kB ± 8%  -34.20%  (p=0.008 n=5+5)

name        old allocs/op  new allocs/op  delta
_Encode-16       145 ± 0%       129 ± 0%  -11.03%  (p=0.008 n=5+5)
```

### DIFF
--- a/pkg/util/marshal/marshal.go
+++ b/pkg/util/marshal/marshal.go
@@ -32,7 +32,11 @@ func WriteQueryResponseJSON(v logqlmodel.Result, w io.Writer) error {
 		},
 	}
 
-	return jsoniter.NewEncoder(w).Encode(q)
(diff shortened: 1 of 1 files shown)


chaudum (approved):
LGTM

Failing CI seems to be unrelated and I've never seen that failure case before.

PR #7530, opened 2022-10-26
[helm] Fix incorrect ingress paths when using single binary deployment

### LINKED ISSUE #7080: [Helm] 3.0.0 ingress does not support single binary mode
**Describe the bug**

The ingress that comes with the Helm chart is configured for a scalable deployment only. This is because the paths in the ingress use services suffixed with `read` and `write`. A single binary deployment does not include these services, only `loki` and `loki-headless`. 

https://github.com/grafana/loki/blob/c23e541fe7d6aa4e35e720df4e191f034325d24f/production/helm/loki/values.yaml#L759-L770

https://github.com/grafana/loki/blob/c23e541fe7d6aa4e35e720df4e191f034325d24f/production/helm/loki/templates/ingress.yaml#L36-L53

I have worked around this in the meantime by using the `headless` suffix and emptying the default `write` and `read` lists:

```yaml
paths:
  write: []
  read: []
  headless:
    - /api/prom/push
    - /loki/api/v1/push
    - /api/prom/tail
    - /loki/api/v1/tail
    - /loki/api
    - /api/prom/rules
    - /loki/api/v1/rules
    - /prometheus/api/v1/rules
    - /prometheus/api/v1/alerts
```

**Expected behavior**

The Helm chart uses the `loki` service when deploying an ingress in single binary mode.

trevorwhitney:
The helm ingress template was copied over from the simple scalable chart and does not work when deploying the chart in single binary configuration. This fixes that by adding a set of single binary routes, and conditionally configuring the ingress resource based on deployment type.

Fixes #7080

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -327,6 +327,75 @@ Return if ingress supports pathType.
 {{- end -}}
 
 {{/*
+Generate list of ingress service paths based on deployment type
+*/}}
+{{- define "loki.ingress.servicePaths" -}}
+{{- if (eq (include "loki.deployment.isScalable" .) "true") -}}
+{{- include "loki.ingress.scalableServicePaths" . }}
+{{- else -}}
+{{- include "loki.ingress.singleBinaryServicePaths" . }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Ingress service paths for scalable deployment
+*/}}
+{{- define "loki.ingress.scalableServicePaths" -}}
+{{- include "loki.ingress.servicePath" (dict "ctx" . "svcName" "read" "paths" .Values.ingress.paths.read )}}
+{{- include "loki.ingress.servicePath" (dict "ctx" . "svcName" "write" "paths" .Values.ingress.paths.write )}}
+{{- end -}}
+
+{{/*
+Ingress service paths for single binary deployment
+*/}}
+{{- define "loki.ingress.singleBinaryServicePaths" -}}
+{{- include "loki.ingress.servicePath" (dict "ctx" . "svcName" "singleBinary" "paths" .Values.ingress.paths.singleBinary )}}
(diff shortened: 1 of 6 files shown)


trevorwhitney:
@calvinbui would you mind double checking this works in your env since you filed the bug?

jeschkies (changes requested):
Please merge `main` and generate the reference with `make -BC docs sources/installation/helm/reference.md`

MichelHollands:
@trevorwhitney This is looking good. Can you rebase and run the command Karsten mentioned? I can approve and merge afterwards.

trevorwhitney:
@jeschkies ran the command you mentioned and it nuked the reference.md, which I assume is not the goal?

jeschkies:
@trevorwhitney calling `BUILD_IN_CONTAINER=false make -BC docs sources/installation/helm/reference.md` from root dir should work.

jeschkies:
Here is the change for the reference https://github.com/trevorwhitney/loki/pull/1

production/helm/loki/templates/_helpers.tpl
```diff
@@ -326,6 +326,75 @@ Return if ingress supports pathType.
...
+Ingress service paths for single binary deployment
+*/}}
+{{- define "loki.ingress.singleBinaryServicePaths" -}}
+{{- include "loki.ingress.servicePath" (dict "ctx" . "svcName" "singleBinary" "paths" .Values.ingress.paths.singleBinary )}}
+{{- end -}}
+
+{{/*
+Ingress service path helper function
+Params:
+  ctx = . context
```
jeschkies, line 358: What is the context?

jeschkies (approved):
Go Template is just super hard to read. I don't grasp the template completely but it seems mostly copy and paste.

PR #7534, opened 2022-10-27
Improves performance of distributor validations

cyriltovena:
Most of the time we don't need to create those strings which create an alloc per entry.


owen-d (approved):

salvacorts (approved):
LGTM 🎉

kavirajk (approved):
LGTM 👍

Issue #7529, opened 2022-10-26
Helm breaking change in config adaptions


tyriis:
[this commit](https://github.com/grafana/loki/commit/9bc6e85604bcfb3d486eb4bc48305fdc163f0589)
creates breaking changes in existing configured clusters.

As an end-user i want to have trust in minor version changes. I manage my cluster with renovate dependency upgrade for flux, so I got an update for the latest 3.3.0 chart without the need of a manual review. I can of course create an exception for Loki and review minor changes, but I would expect and prefer a proper versioning here.

`MetricsInstance` is part of Grafana operator currently in `beta`. Grafana operator is not a dependency of loki. Please be more careful with breaking changes, better keep downwards compatibility by disabling new things and introduce them in a major release.

Keep up with your great work, and thanks a lot for delivering such powerful tools to the opensource community!

PR #7525, opened 2022-10-26
[Helm] Add toggle for MetricsInstance

jkroepke:
Signed-off-by: Jan-Otto Kröpke <jok@cloudeteer.de>

**What this PR does / why we need it**:

Chart version 3.3.0 introduced a new MetricsInstance CR. The only way to disable the MetricsInstance is `--set loki.monitoring.serviceMonitor.metricsInstance=null`, but this raises helm warning... 

```
coalesce.go:220: warning: cannot overwrite table with non table for loki.monitoring.serviceMonitor.metricsInstance (map[annotations:map[] labels:map[] remoteWrite:<nil>])
coalesce.go:220: warning: cannot overwrite table with non table for loki.monitoring.serviceMonitor.metricsInstance (map[annotations:map[] labels:map[] remoteWrite:<nil>])
coalesce.go:220: warning: cannot overwrite table with non table for loki.monitoring.serviceMonitor.metricsInstance (map[annotations:map[] labels:map[] remoteWrite:<nil>])
```

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Checklist**
- [ ] Reviewed the `CONTRIBUTING.md` guide
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #7535, opened 2022-10-27
[Helm] Create metricsInstance only if available

agologan:
If you're using prometheus operator and not Grafana Agent a metricsInstance shouldn't be auto-created.

#7529

I would also propose moving the definition at the same level with serviceMonitor and adding .enabled i.e.:
```
  serviceMonitor:
    enabled: true
    ...
  metricsInstance:
    enabled: true
    annotations: {}
    labels: {}
    remoteWrite: null
```
This of course would be a breaking change.
**L.E.** Just noticed this other PR: #7525

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/templates/monitoring/metrics-instance.yaml
+++ b/production/helm/loki/templates/monitoring/metrics-instance.yaml
@@ -1,5 +1,6 @@
-{{- if and .Values.monitoring.serviceMonitor.enabled .Values.monitoring.serviceMonitor.metricsInstance.enabled  }}
+{{- if .Values.monitoring.serviceMonitor.enabled }}
 {{- with .Values.monitoring.serviceMonitor.metricsInstance }}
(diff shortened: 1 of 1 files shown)


trevorwhitney (changes requested):
I'd like to see a combination of this and #7525. I like where the enabled flag was added in #7525 as I feel it's a bit less disruptive, but I also like you capabilities check. Would you mind refactoring once #7525 is in to add your check to that same conditional where the new `enabled` is checked?

agologan:
@trevorwhitney updated. 
While I could've just written this big **and**, used the same pattern present in other files to be a bit more consistent.

trevorwhitney (approved):
lgtm, looks good, thanks!

PR #7538, opened 2022-10-27
Loki: Fix index_set log message

DylanGuedes:
As of now, this log message isn't correctly showing the list of files (instead it logs the value as "unsupported value type").

### DIFF
--- a/pkg/storage/stores/indexshipper/downloads/index_set.go
+++ b/pkg/storage/stores/indexshipper/downloads/index_set.go
@@ -278,14 +278,7 @@ func (t *indexSet) sync(ctx context.Context, lock, bypassListCache bool) (err er
 		return err
 	}
 
-	toDownloadBuilder := strings.Builder{}
-	for i, f := range toDownload {
-		if i != 0 {
-			toDownloadBuilder.WriteString(",")
-		}
-		toDownloadBuilder.WriteString(f.Name)
-	}
-	level.Debug(t.logger).Log("msg", fmt.Sprintf("updates for table %s. toDownload: %s, toDelete: %s", t.tableName, toDownloadBuilder.String(), toDelete))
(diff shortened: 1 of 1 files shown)


pkg/storage/stores/indexshipper/downloads/index_set.go
```diff
@@ -278,7 +278,14 @@ func (t *indexSet) sync(ctx context.Context, lock, bypassListCache bool) (err er
 		return err
 	}
 
-	level.Debug(t.logger).Log("msg", fmt.Sprintf("updates for table %s. toDownload: %s, toDelete: %s", t.tableName, toDownload, toDelete))
```
sandeepsukhani, line 281: This logline is being printed fine

pkg/storage/stores/indexshipper/downloads/index_set.go
```diff
@@ -289,7 +296,7 @@ func (t *indexSet) sync(ctx context.Context, lock, bypassListCache bool) (err er
 	// it means the cache is not valid anymore since compaction would have happened after last index list cache refresh.
 	// Let us return error to ask the caller to re-run the sync after the list cache refresh.
 	if !bypassListCache && len(downloadedFiles) == 0 && len(toDownload) > 0 {
-		level.Error(t.logger).Log("msg", "we skipped downloading all the new files, possibly removed by compaction", "files", toDownload)
```
sandeepsukhani, line 292: I think the simplest fix here would be to do this
```
level.Error(t.logger).Log("msg", "we skipped downloading all the new files, possibly removed by compaction", "files", fmt.Sprint(toDownload))
```

DylanGuedes: good call, ty

MichelHollands (approved):
LGTM

PR #7540, opened 2022-10-28
Loki: Add support for Azure Workload Identity authentication

Tolsto:

Adds support for [Azure Workload Identity](https://azure.github.io/azure-workload-identity/docs/) authentication.


jkroepke:
Please validate, if it re-reads the token after a hour, since the token file is rotating each hour.

pkg/storage/chunk/client/azure/blob_storage_client.go
```diff
@@ -378,6 +382,37 @@ func (b *BlobStorage) getServicePrincipalToken() (*adal.ServicePrincipalToken, e
 
 	resource := fmt.Sprintf("https://%s.%s", b.cfg.StorageAccountName, endpoint)
 
+	if b.cfg.UseFederatedToken {
+		var environmentName string
+		if b.cfg.Environment == azureGlobal {
+			environmentName = "AzurePublicCloud"
+		} else {
+			environmentName = b.cfg.Environment
+		}
```
kavirajk, line 391: Can you add `AzurePublicCloud` as part of the list of const `Environments` top of this file.

Also you could avoid `else` here. 
```
		environmentName := azurePublicCloud

		if b.cfg.Environment != azureGlobal {
			environmentName = b.cfg.Environment
		}

```
looks more readable IMHO.

Q: Also curious what's the difference between `AzureGlobal` vs `AzurePublicCloud` environments?

kavirajk (commented):
Thanks for PR @Tolsto. Added one minor feedback. Can you please add small test to check if `adal.NewServicePrincipalTokenFromFederatedToken` works correctly with this additional flag and configs?

pkg/storage/chunk/client/azure/blob_storage_client.go
Tolsto: I don't know where `AzureGlobal` is coming from but it doesn't exist in the Azure Go SDKs. It's `AzurePublicCloud` there, i.e. you cannot do an environment lookup via `EnvironmentFromName` with `AzureGlobal`

jkroepke: image

The workload identity webhook is exposing a authority host `AZURE_AUTHORITY_HOST` which is the `env.ActiveDirectoryEndpoint`. Consider to prefer the environment variable instead a own cloud env lookup

Tolsto: I thought about that but decided to go with the reference implementation outlined by the person who added the code to adal. I also looked at the new [azidentity module where AZURE_AUTHORITY_HOST gets only used as a fallback option.](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/azidentity.go#L79) It will only honor it if no cloud configuration has been set. However, the current implementation of Azure storage in Loki requires the set cloud environment name to be set, i.e. adding fallback logic is imho futile.

Tolsto:
> FYI: Workload Identity will be official supported jan 2023 - [Azure/azure-sdk-for-go#15615](https://github.com/Azure/azure-sdk-for-go/issues/15615)

Doesn't help us much as Loki doesn't use `azure-sdk-for-go` for the Azure storage integration but the discontinued [azure-storage-blob](https://github.com/Azure/azure-storage-blob-go) SDK.

weisdd (approved):

pkg/storage/chunk/client/azure/blob_storage_client.go
```diff
@@ -48,6 +51,11 @@ var (
 		azureUSGovernment: "blob.core.usgovcloudapi.net",
 	}
 
+	defaultAuthFunctions = authFunctions{
+		NewOAuthConfigFunc: adal.NewOAuthConfig,
+		NewServicePrincipalTokenFromFederatedTokenFunc: adal.NewServicePrincipalTokenFromFederatedToken,
```
weisdd, line 56: Have you tried to use loki with your WI patch for longer than 24h? I guess it would have the same issue that I discovered while preparing a patch for external-dns and cert-manager. - Federated token is valid for 1h by default, and AAD token received in exchange - for 24h. So, before you can renew an access token, you need to renew the federated token by re-reading it from disk. Unfortunately, adal doesn't expose any native method to do so, but you can trick it through a `customRefreshFunc`. You could find implementation details here: https://github.com/kubernetes-sigs/external-dns/pull/3111
NOTE: it's not like I tested loki with your patch, I'm just assuming you'll hit the same issue.

weisdd (changes requested):

pkg/storage/chunk/client/azure/blob_storage_client.go
weisdd: @Tolsto always happy to help :)

Tolsto: Can confirm that it now keeps working after 24h.

weisdd (changes requested):

pkg/storage/chunk/client/azure/blob_storage_client.go
weisdd: Happy to hear it keeps on working now. The implementation looks fine to me.

pkg/storage/chunk/client/azure/blob_storage_client.go
```diff
@@ -32,6 +34,7 @@ import (
 const (
 	// Environment
 	azureGlobal       = "AzureGlobal"
+	azurePublicCloud  = "AzurePublicCloud"
```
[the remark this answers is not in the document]
weisdd: Ah, alright, I see your point. Never mind then :)

weisdd (approved):

Tolsto:
@kavirajk Anything missing for this PR to move forward?

p4ranoidandro1d:
Would be a nice addition, because actual workaround using sidecar requires enabling privilege escalation.

Tolsto:
ping @kavirajk

pkg/storage/chunk/client/azure/blob_storage_client_test.go
```diff
@@ -24,6 +28,58 @@ func (fn RoundTripperFunc) RoundTrip(req *http.Request) (*http.Response, error)
...
 }
 
+type FederatedTokenTestSuite struct {
+	suite.Suite
+	config                      *BlobStorage
+	mockOAuthConfig             *adal.OAuthConfig
+	mockedServicePrincipalToken *adal.ServicePrincipalToken
+}
+
+func (suite *FederatedTokenTestSuite) SetupTest() {
```
jeschkies, line 38: Interesting. Are you using the suite to get a test setup code?

jeschkies (commented):
Unfortunately I don't know Azure well enough to accept this change. I does look good to me though.

docs/sources/configuration/_index.md
```diff
@@ -829,6 +829,13 @@ The `azure_storage_config` configures Azure as a general storage for different d
 # CLI flag: -<prefix>.azure.use-managed-identity
 [use_managed_identity: <boolean> | default = false]
 
+# Use a Federated Token to authenticate to the Azure storage account.
```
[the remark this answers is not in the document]
Tolsto: I'd suppose so.

https://azure.github.io/azure-workload-identity/docs/topics/federated-identity-credential.html
https://learn.microsoft.com/en-us/azure/active-directory/develop/workload-identity-federation

pkg/storage/chunk/client/azure/blob_storage_client_test.go
Tolsto: I use the suite to set up an isolated environment with mocks for testing the Azure federated token functionality. There might be other ways to do it, but I come from the Ruby/Rspec world, and using a suite was the most familiar approach for me.

docs/sources/configuration/_index.md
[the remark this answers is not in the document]
jeschkies: Could you add the second link in the documentation? I think that would be helpful.

Also, I know this is a big ask but could you provide documentation similar to https://grafana.com/docs/loki/next/storage/#aws-deployment-s3-single-store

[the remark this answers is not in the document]
Tolsto: I don't think that the second link is very helpful as it describes the federated token architecture in general. >90% of Loki users will use it with the reference implementation (Azure Workload Identity) in a Kubernetes cluster.
I'm willing to provide documentation if this PR can then get merged in the very near future.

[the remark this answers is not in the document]
jeschkies: Sounds good to me 😊

jeschkies (approved):

jeschkies:
@Tolsto once you've fixed the merge conflicts we can land the change.

PR #7543, opened 2022-10-28
[logcli] Add parallel flags

SN9NV:
Requesting a large range, a day, for example, takes a very long time to download. You can download in parallel by starting many processes, but this just made my job so much easier.

TBH, I am not good at testing, so IDK how such a thing can be tested.

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


MichelHollands:
@SN9NV I'm not sure we can accept this as is. The log lines will be returned out of order which will be very confusing for users. Is there a way you can add ordering? Also this needs rebasing.

SN9NV:
One of the ways I thought we could add ordering is if we created an output file for each part. Then the user can order the logs using the filename.

But I guess that would be some work. Adding new flags and stuff.
But if that is what you want, I can give it a try.

SN9NV:
> @SN9NV I'm not sure we can accept this as is. The log lines will be returned out of order which will be very confusing for users. Is there a way you can add ordering? Also this needs rebasing.

I have pushed a new bit of functionality which allows the user to create output files, one for each partition range, then the user can order file files as they please. I think we can leave that up to them. Not so sure on the naming. I couldn't think of another tool which does this to get some naming inspiration.

The reason for the prefix is so that you can place the files in a specific directory, or "label" a specific download set so many in the same directory don't collide.

pkg/logcli/query/query.go
```diff
@@ -156,9 +173,93 @@ func (q *Query) DoQuery(c client.Client, out output.LogOutput, statistics bool)
...
+		rq.Start = start
+		rq.End = end
+
+		jobs <- rq
 
+		if end.Equal(q.End) {
+			break
 		}
+
+		start = end
```
jeschkies, line 250: Checkout the comment in https://github.com/grafana/loki/blob/cf23194d05d33105958f9e1ef70da3ed9717a81c/pkg/logcli/query/query.go#L142 I think you need to specify a limit.

jeschkies (changes requested):
I like the ist but I'd wie the results into one file. As I wouldn't expect all results to arrive at once you could use a channel to the writer that synchronizes the writes. Maybe a simple mutex will do.

SN9NV:
> I like the ist but I'd wie the results into one file. As I wouldn't expect all results to arrive at once you could use a channel to the writer that synchronizes the writes. Maybe a simple mutex will do.

I'm not quite sure what you are saying.

Once you start the workers in parallel, sorting the results into one file would mean that you would need all the memory to hold all the messages until the worker with the lowest value is done. This is why I have saved one file per section.

The previous comment said we should try to keep the ordering of the log messages. And you comment seems to contradict that.

jeschkies:
> Once you start the workers in parallel, sorting the results into one file would mean that you would need all the memory to hold all the messages until the worker with the lowest value is done. This is why I have saved one file per section.

You don't have to keep them in memory. You could insert them into a file. That's a little bit more complicated though. It's like insertion sort with a file.

jeschkies:
@SN9NV I've done a quick research. What I mean is called [External Sorting](https://en.wikipedia.org/wiki/External_sorting). However, we probably do not need it this complicated. I know this is some extra work but what do you think about naming the part files with timestamps and then merge them afterwards? It would be an append only operation. The difficult part is to figure out if we need deduplication.

jeschkies:
@SN9NV I'm sorry that you closed the PR. I liked the idea.

SN9NV:
I moved the code to another branch. I guess that's why it was closed. I'll open another PR from the new branch.

PR #7548, opened 2022-10-30
Prevent IndexShipper from leaking mutex accesses.

owen-d:
# What

Inverts control in the `index-shipper` so consumers don't manage accesses to mutexes. This prevents deadlocks and no longer requires consumers have internal knowledge of the `index-shipper` itself.

This PR reverts a few previous attempts. Look at the last commit to see the net-additions.

# Background:

TSDB previously didn't use mmap, which meant we could freely pass around TSDBs as `[]byte`. Since they were immutable, we didn't need to know when an underlying index was cleaned up/closed; this coordination was unnecessary and the garbage collector would clean up dangling memory for us. After introducing TSDB, we needed to control access to avoid use-after-free bugs. A few attempts at controlling this became increasingly complex & error prone. Instead of requiring _consumers_ to prevent deadlocks on index-shipper internals, it makes more sense that references to the index-shipper's resources should not leak outside the `ForEach` functions that give temporary access to them. This _inverts control_ from the consumer (code which queries the indices) to the producer (the index-shipper).

### DIFF
--- a/pkg/storage/stores/indexshipper/downloads/table.go
+++ b/pkg/storage/stores/indexshipper/downloads/table.go
@@ -28,7 +28,7 @@ const (
 
 type Table interface {
 	Close()
-	ForEach(ctx context.Context, userID string, doneChan <-chan struct{}, callback index.ForEachIndexCallback) error
+	ForEach(ctx context.Context, userID string, callback index.ForEachIndexCallback) error
 	DropUnusedIndex(ttl time.Duration, now time.Time) (bool, error)
 	Sync(ctx context.Context) error
 	EnsureQueryReadiness(ctx context.Context, userIDs []string) error
@@ -144,7 +144,7 @@ func (t *table) Close() {
 	t.indexSets = map[string]IndexSet{}
 }
 
-func (t *table) ForEach(ctx context.Context, userID string, doneChan <-chan struct{}, callback index.ForEachIndexCallback) error {
+func (t *table) ForEach(ctx context.Context, userID string, callback index.ForEachIndexCallback) error {
 	// iterate through both user and common index
 	for _, uid := range []string{userID, ""} {
 		indexSet, err := t.getOrCreateIndexSet(ctx, uid, true)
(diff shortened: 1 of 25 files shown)


periklis:
@owen-d Is the leackage serious enough to add the backport label for 2.7.0?

pkg/storage/stores/indexshipper/downloads/table.go
```diff
@@ -145,24 +146,48 @@ func (t *table) Close() {
...
-	sets, err := t.getOrCreateIndexSets(ctx, ids, true)
-	if err != nil {
-		return err
+func (t *table) ForEachConcurrent(ctx context.Context, userID string, callback index.ForEachIndexCallback) error {
+	g, ctx := errgroup.WithContext(ctx)
+
+	// iterate through both user and common index
+	users := []string{userID, ""}
+
+	for i := range []string{userID, ""} {
```
DylanGuedes, line 155:
```suggestion
	for i := range users {
```

pkg/storage/stores/tsdb/index_shipper_querier.go
```diff
@@ -29,36 +31,35 @@ func newIndexShipperQuerier(shipper indexShipperIterator, tableRanges config.Tab
...
+				if !ok {
+					return fmt.Errorf("unexpected shipper index type: %T", idx)
+				}
+				if multitenant {
+					impl = NewMultiTenantIndex(impl)
+				}
+
+				return f(ctx, impl)
+			}); err != nil {
+				return err
```
DylanGuedes, line 56: i think here you're allowing `itr` to be of type `err`. I'm not sure the NewMultiIndex at line 62 will be able of handling it though :thinking:

sandeepsukhani: The return here is part of the function definition cast to `indexIterFunc`. It looks fine to me. `itr` is holding a function definition and not capturing the output of a function call.

sandeepsukhani:
> @owen-d Is the leackage serious enough to add the backport label for 2.7.0?

the code that we are fixing here never made it to any release so not required.

sandeepsukhani (approved):
Let us leave a note somewhere on lazy iteration possibly going over different index files within the same index query. For e.g, `Bounds` and `GetChunkRefs` might go through different index files if a sync happened between the calls.

The code otherwise LGTM.

pkg/storage/stores/tsdb/index_shipper_querier.go
DylanGuedes: ohhh you right, thank you

PR #7555, opened 2022-10-31
Loki: Implement timeouts migration

DylanGuedes:
Improve the experience of migrating to our new timeouts behavior, which is the unifying of the engine:timeout and querier:query_timeout into a new configuration named limits_config:query_timeout.
This PR makes sure users won't be surprised by their timeouts being shorter, but at the same time, it makes it easier to migrate to the new configuration.

### DIFF
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -29,6 +29,7 @@ import (
 	"github.com/grafana/loki/pkg/distributor"
 	"github.com/grafana/loki/pkg/ingester"
 	"github.com/grafana/loki/pkg/ingester/client"
+	"github.com/grafana/loki/pkg/logql"
 	"github.com/grafana/loki/pkg/loki/common"
 	"github.com/grafana/loki/pkg/lokifrontend"
 	"github.com/grafana/loki/pkg/querier"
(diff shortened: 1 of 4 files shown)


pkg/loki/loki.go
```diff
@@ -221,6 +222,64 @@ func (c *Config) Validate() error {
...
+		return nil
+	}
+
+	if perTenantTimeoutIsDefault && !engineTimeoutIsDefault {
+		if err := c.LimitsConfig.QueryTimeout.Set(c.Querier.Engine.Timeout.String()); err != nil {
+			return fmt.Errorf("couldn't set per-tenant query_timeout as the engine timeout value: %w", err)
+		}
+		level.Warn(util_log.Logger).Log("msg",
+			fmt.Sprintf(
+				"using configured engine timeout (%q) for all queries. Be aware that engine timeout (%q) is deprecated and will be removed in the next major version.",
```
owen-d, line 275:
```suggestion
				"using configured engine timeout (%q) as the default (can be overridden per-tenant in the limits_config). Be aware that engine timeout (%q) is deprecated and will be removed in the next major version.",
```

pkg/loki/loki.go
```diff
@@ -221,6 +222,64 @@ func (c *Config) Validate() error {
...
+				c.LimitsConfig.QueryTimeout.String(),
+			),
+		)
+		return nil
+	}
+
+	if !perTenantTimeoutIsDefault && !engineTimeoutIsDefault {
+		level.Warn(util_log.Logger).Log("msg",
+			fmt.Sprintf(
+				"using configured per-tenant timeout (%q) for all queries, configured engine timeout (%q) is deprecated and will be ignored.",
```
owen-d, line 261:
```suggestion
				"using configured per-tenant timeout (%q) as the default (can be overridden per-tenant in the limits_config). Configured engine timeout (%q) is deprecated and will be ignored.",
```

owen-d (approved):
Nit, but LGTM

PR #7556, opened 2022-10-31
Ensure labels on sharded streams are sorted before they're sent anywhere

MasslessParticle:
Pushing streams with unsorted labels to ingesters results in a bug where ingesters duplicate chunks for streams. This causes a memory leak and eventual OOM crash when ingesters recover from the WAL because one of the duplicate streams is orphaned.

This PR ensures that sharded streams have sorted labels to avoid this bug.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -6,6 +6,7 @@ import (
 	"math"
 	"math/rand"
 	"net/http"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -498,6 +499,8 @@ func labelTemplate(lbls string) labels.Labels {
 	copy(streamLabels, baseLbls)
 	streamLabels[len(baseLbls)] = labels.Label{Name: ingester.ShardLbName, Value: ingester.ShardLbPlaceholder}
 
+	sort.Sort(labels.Labels(streamLabels))
+
 	return streamLabels
 }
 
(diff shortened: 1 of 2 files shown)


pkg/distributor/distributor_test.go
```diff
@@ -470,7 +470,7 @@ func Test_TruncateLogLines(t *testing.T) {
 func TestStreamShard(t *testing.T) {
 	// setup base stream.
 	baseStream := logproto.Stream{}
-	baseLabels := "{app='myapp', job='fizzbuzz'}"
```
DylanGuedes, line 473: hmm out of curiosity, why was this label removed from the test case?

pkg/distributor/distributor.go
```diff
@@ -508,7 +511,13 @@ func (d *Distributor) createShard(streamshardCfg *shardstreams.Config, stream lo
...
-	lbls[len(lbls)-1] = labels.Label{Name: ingester.ShardLbName, Value: shardLabel}
+	for i := 0; i < len(lbls); i++ {
+		if lbls[i].Name == ingester.ShardLbName {
+			lbls[i].Value = shardLabel
+			break
+		}
+	}
+
 	return logproto.Stream{
 		Labels:  strings.Replace(streamPattern, ingester.ShardLbPlaceholder, shardLabel, 1),
```
DylanGuedes, line 522: hmm maybe we don't need this anymore after the change of line 514?

DylanGuedes (approved):
lgtm! just 2 questions

pkg/distributor/distributor.go
MasslessParticle: iirc, we did this to save allocations. labels -> string is pretty expensive. Instead, we generate a template once and do a string replacement.

PR #7560, opened 2022-10-31
Shard streams round-robin rather in batches. 

MasslessParticle:
To avoid hot spots, keep track of shard numbers across pushes. When tracking shard numbers across pushes, it's no longer necessary to randomize shards, in the case where the desired number of shards is greater than the number of entries. Instead, we just partition the entries among the shards in round robin fashion. Over time, this should give us a more even distribution of logs in shards, particularly for large streams that are made of small individual pushes.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"flag"
 	"math"
-	"math/rand"
 	"net/http"
 	"sort"
 	"strconv"
@@ -87,7 +86,8 @@ type Distributor struct {
 	validator        *Validator
 	pool             *ring_client.Pool
 
-	rateStore RateStore
+	rateStore    RateStore
+	shardTracker *ShardTracker
 
 	// The global rate limiter requires a distributors ring to count
 	// the number of healthy instances.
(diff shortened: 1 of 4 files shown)


pkg/distributor/distributor.go
```diff
@@ -440,40 +422,35 @@ func (d *Distributor) shardStream(stream logproto.Stream, streamSize int, userID
...
 }
 
-func (d *Distributor) divideEntriesBetweenShards(logger log.Logger, userID string, shardCount int, shardStreamsCfg *shardstreams.Config, stream logproto.Stream, alternateShardsOrder []int) ([]uint32, []streamTracker) {
-	derivedKeys := make([]uint32, 0, shardCount)
-	derivedStreams := make([]streamTracker, 0, shardCount)
-	streamLabels := labelTemplate(stream.Labels)
-	streamPattern := streamLabels.String()
+func (d *Distributor) createShards(stream logproto.Stream, totalShards int, userID string, shardStreamsCfg *shardstreams.Config) ([]uint32, []streamTracker) {
+	var (
+		streamLabels   = labelTemplate(stream.Labels)
```
owen-d, line 442: I think `labelTemplate` needs to sort the labels after adding the shard label since we rely on labels to be sorted coming from the distributor.

MasslessParticle: Already covered in https://github.com/grafana/loki/pull/7556

DylanGuedes (approved):
LGTM!
do you mind adding to the PR description why moving from the random assignment to round-robin?

PR #7563, opened 2022-10-31
Refactor CacheSplitter interface

trevorwhitney:

This is another approach to fixing the LBAC cache previously fixed by https://github.com/grafana/loki/pull/7542. Instead of injecting a "callback" function, this refactors the `CacheSplitter` interface a little, as well as pushing the `QuerySplitDuration` limit down into the `queryrangebase` package.


MichelHollands:
This is no longer needed, we went with another solution.

PR #7568, opened 2022-11-01
simplify stream rate calculator and reduce stripe size in ingester

owen-d:
I found a few things to simplify/change in the rate calculator:
* Reduce default stripe size to 1024. I don't think 32k locks are needed/worth the cost they incur
* Remove the `rates` field in the `StreamRateCalculator` as it was only used during one function and isn't necessary to keep around.

### DIFF
--- a/pkg/ingester/stream_rate_calculator.go
+++ b/pkg/ingester/stream_rate_calculator.go
@@ -10,7 +10,7 @@ import (
 const (
 	// defaultStripeSize is the default number of entries to allocate in the
 	// stripeSeries list.
-	defaultStripeSize = 1 << 15
+	defaultStripeSize = 1 << 10
 
 	// The intent is for a per-second rate so this is hard coded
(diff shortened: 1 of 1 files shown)


DylanGuedes (approved):

pkg/ingester/stream_rate_calculator.go
```diff
@@ -10,7 +10,7 @@ import (
 const (
 	// defaultStripeSize is the default number of entries to allocate in the
 	// stripeSeries list.
-	defaultStripeSize = 1 << 15
+	defaultStripeSize = 1 << 10
```
MasslessParticle, line 13: Will this be enough? We typically have 5,000 - 10,000 streams per ingester.

MasslessParticle (commented):
One question

MasslessParticle:
superseded by https://github.com/grafana/loki/pull/7583

PR #7152, opened 2022-09-14
logcli: Added exponential backoff for retries

LaserPhaser:
Added new option for configuring wait time between retries

**What this PR does / why we need it**:
In case of errors on server side we may need for a while to do the retry properly.

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [ ] Is this an important fix or new feature? Add an entry in the `CHANGELOG.md`.
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #6671, opened 2022-07-13
[Optimization] logql client: When logql fails to execute, return the failure information to the caller

liguozhong:
…ail msg to the client caller.

**What this PR does / why we need it**:
When logql fails to execute, return the failure information to the caller. 
Solve the problem that grafana's  loki alarm cannot see the cause of failure.

I wrote a testcase, you can see the details of the error through this PR.
image

**Which issue(s) this PR fixes**:
Fixes #<issue number>

The loki alarm of grafana uses this logql client, and it is not possible to see the cause of the error when there is an error

**Special notes for your reviewer**:

image

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] Is this an important fix or new feature? Add an entry in the `CHANGELOG.md`.
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #7579, opened 2022-11-02
Retry pr

MasslessParticle:
This is the same PR as https://github.com/grafana/loki/pull/7152 but I botched the merge when updating the code. This one works


kavirajk (approved):

liguozhong:
This PR removes the return value, which makes it impossible for the client to observe the cause of the error through code.
ref pr:
https://github.com/grafana/loki/pull/6671

PR #7582, opened 2022-11-02
Improve CRI format documentation

DanCech:
The documentation doesn't mention the flags/tags component in a CRI log line.  I went with flags since that's the name it's given in our regex.


MichelHollands (approved):
LGTM

PR #7584, opened 2022-11-03
Revert "Add way to override userId for caching (#7542)"

trevorwhitney:
This reverts commit 16761723f4f2dbdeece4697377239b99d25368fe.

This code was added in #7542 but I'm still not clear what caching bug it fixes?


MichelHollands:
This is no longer needed.

PR #7594, opened 2022-11-03
Add config for optional pod labels

Tolsto:
Currently, pod labels for Loki services can only
be added via the `selectorLabels` key. However, it's a
confusing name as it actually doesn't change the
selectors but only adds additional pod labels.

Additionally, it's currently not possible to add labels 
for the gateway and single binary deployments.

This change adds a new `podLabels` key for all Loki services
analog to `podAnnotations` in order to prevent adding the
confusing `selectorLabels` key to more sections and keep
everything consistent.

The `selectorLabels` keys could then be removed with
the next major version bump of this chart.

### DIFF
--- a/production/helm/loki/templates/gateway/deployment-gateway.yaml
+++ b/production/helm/loki/templates/gateway/deployment-gateway.yaml
@@ -29,6 +29,12 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
       labels:
+        {{- with .Values.loki.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+        {{- with .Values.gateway.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
         {{- include "loki.gatewaySelectorLabels" . | nindent 8 }}
     spec:
       serviceAccountName: {{ include "loki.serviceAccountName" . -}}
--- a/docs/sources/installation/helm/reference.md
+++ b/docs/sources/installation/helm/reference.md
@@ -814,6 +814,15 @@ See values.yaml
 </td>
 		</tr>
 		<tr>
+			<td>gateway.podLabels</td>
+			<td>object</td>
+			<td>Additional labels for gateway pods</td>
+			<td><pre lang="json">
+{}
+</pre>
+</td>
+		</tr>
+		<tr>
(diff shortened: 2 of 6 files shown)


production/helm/loki/templates/gateway/deployment-gateway.yaml
```diff
@@ -29,6 +29,12 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
       labels:
+        {{- with .Values.gateway.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+        {{- with .Values.loki.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
```
trevorwhitney, line 37: how does k8s handle if the same label is specified on both, and the template ends up with a duplicate? would it be better to use a template here and leverage helm's merging capabilities?

trevorwhitney (commented):
thanks for the change, just one small question.

production/helm/loki/templates/gateway/deployment-gateway.yaml
Tolsto: Fair point. If two label keys are duplicated then the latter label wins. I don't quite see the value of using a template/merge here as the only choice is which label should get precedent, the one defined in the generic `loki` section or the more specific label in the service section.
Currently, the labels from the `loki` section get precedent. However, the behavior for annotations is the other way around and that actually makes more sense to me. 
I'll switch the order for the labels.

Tolsto:
@trevorwhitney Is there anything missing for this PR to move forward?

trevorwhitney (approved):
lgtm! thanks for the improvement!

trevorwhitney:
There we go :) Feel free to move that extra empty line in a subsequent commit (just realized I could have done an empty commit instead). Good to merge once CI passes. Thanks again!

Tolsto:
@trevorwhitney rebased branch onto main. CI seems to fail because of an unrelated issue.

Tolsto:
@trevorwhitney Can you merge this?

trevorwhitney:
look like there was a CI run pending, just triggered it.

Tolsto:
@trevorwhitney Please run the helm linter again, I just fixed the issue.

PR #7206, opened 2022-09-21
[bugfix] promtail:Concurrent map read and map write in promtail pod

liguozhong:
**What this PR does / why we need it**:

```
type noopPipeline struct {
	mu    sync.RWMutex
}

```
noopPipeline.struct add lock
```
type noopPipeline struct {
	cache map[uint64]*noopStreamPipeline
	mu    sync.RWMutex
}
```

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Checklist**
- [ ] Reviewed the `CONTRIBUTING.md` guide
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

Issue #7180, opened 2022-09-16
Concurrent map read and map write in promtail pod


athrunecho:
**Describe the bug**
Some of my promtail pods in CrashLoopBackOff and restart over 1000 times.
After I check the pod logs, I found it start with a fatal error log:
fatal error: concurrent map read and map write

**To Reproduce**
The issue occurs suddenly and after sometime whe nI check the pod goes fine.

**Expected behavior**
promtail should normally work. 

**Environment:**
Kubenetes: v1.22.11
promtail: v2.1.0

**Screenshots, Promtail config, or terminal output**
![Screen Shot 2022-09-14 at 14 04 59](https://user-images.githubusercontent.com/25002032/190552464-70323083-ea90-4f71-b70f-4ceed8b1f3d3.png)

liguozhong:
Please provide a more comprehensive panic log.

athrunecho:
![1](https://user-images.githubusercontent.com/25002032/191156961-d403a47a-c1d3-4358-a928-058e14e1be92.png)

![2](https://user-images.githubusercontent.com/25002032/191157631-5ee173cc-c3e1-4ab3-9ba1-044a8050c4e2.png)

![3](https://user-images.githubusercontent.com/25002032/191157646-74e25af0-63b6-4c8d-b345-37a87c9ee362.png)

liguozhong:
Thank you for the detailed log provided. According to the log, I found that there is a code that is relatively dangerous. A lock is added to protect thread safety, but it will lead to performance degradation. I need to wait for the loki team review.

liguozhong:
promtail v2.1.0 code.

```go
// Process implements Stage
func (m *matcherStage) Process(lbs model.LabelSet, extracted map[string]interface{}, t *time.Time, entry *string) {
	for _, filter := range m.matchers {
		if !filter.Matches(string(lbs[model.LabelName(filter.Name)])) {
			return
		}
	}

	sp := m.pipeline.ForStream(labels.FromMap(util.ModelLabelSetToMap(lbs)))
	if newLine, newLabels, ok := sp.ProcessString(*entry); ok {
		switch m.action {
		case MatchActionDrop:
			// Adds the drop label to not be sent by the api.EntryHandler
			lbs[dropLabel] = model.LabelValue(m.dropReason)
		case MatchActionKeep:
			*entry = newLine
			for k := range lbs {
				delete(lbs, k)
			}
			for _, l := range newLabels.Labels() {
[...]

PR #7597, opened 2022-11-04
promtail: ratelimiting by label

redbaron:

Fixes #<issue number>

Limit stage can now takes optional `by_label_name` param, which tracks distinct values of that label and ratelimit entries independently.

Log lines without expected label are not considered for ratelimiting and passed to the next pipeline stage.

To avoid stalling whole stage by just few labels it requires 'drop', to be set to true.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 
 * [7462](https://github.com/grafana/loki/pull/7462) **MarNicGit**: Allow excluding event message from Windows Event Log entries.
 * [7602](https://github.com/grafana/loki/pull/7602) **vmax**: Add decolorize stage to Promtail to easily parse colored logs.
+* [7597](https://github.com/grafana/loki/pull/7597) **redbaron**: allow ratelimiting by label
 
 ##### Fixes
 * [7771](https://github.com/grafana/loki/pull/7771) **GeorgeTsilias**: Handle nil error on target Details() call.
@@ -78,7 +79,6 @@ Check the history of the branch FIXME.
 * [6179](https://github.com/grafana/loki/pull/6179) **chaudum**: Add new HTTP endpoint to delete ingester ring token file and shutdown process gracefully
 * [5997](https://github.com/grafana/loki/pull/5997) **simonswine**: Querier: parallize label queries to both stores.
 * [5406](https://github.com/grafana/loki/pull/5406) **ctovena**: Revise the configuration parameters that configure the usage report to grafana.com.
-* [7597](https://github.com/grafana/loki/pull/7597) **redbaron**: allow ratelimiting by label
 * [7264](https://github.com/grafana/loki/pull/7264) **bboreham**: Chunks: decode varints directly from byte buffer, for speed.
 * [7263](https://github.com/grafana/loki/pull/7263) **bboreham**: Dependencies: klauspost/compress package to v1.15.11; improves performance.
 * [7270](https://github.com/grafana/loki/pull/7270) **wilfriedroset**: Add support for `username` to redis cache configuration.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ Check the history of the branch FIXME.
 * [6179](https://github.com/grafana/loki/pull/6179) **chaudum**: Add new HTTP endpoint to delete ingester ring token file and shutdown process gracefully
 * [5997](https://github.com/grafana/loki/pull/5997) **simonswine**: Querier: parallize label queries to both stores.
 * [5406](https://github.com/grafana/loki/pull/5406) **ctovena**: Revise the configuration parameters that configure the usage report to grafana.com.
+* [7597](https://github.com/grafana/loki/pull/7597) **redbaron**: allow ratelimiting by label
 * [7264](https://github.com/grafana/loki/pull/7264) **bboreham**: Chunks: decode varints directly from byte buffer, for speed.
 * [7263](https://github.com/grafana/loki/pull/7263) **bboreham**: Dependencies: klauspost/compress package to v1.15.11; improves performance.
 * [7270](https://github.com/grafana/loki/pull/7270) **wilfriedroset**: Add support for `username` to redis cache configuration.
--- a/clients/pkg/logentry/stages/limit.go
+++ b/clients/pkg/logentry/stages/limit.go
@@ -2,24 +2,34 @@ package stages
 
 import (
 	"context"
+	"fmt"
+
+	"github.com/go-kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/model"
+
+	"github.com/grafana/loki/pkg/util"
 
 	"github.com/go-kit/log"
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
-	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/time/rate"
 )
 
 const (
 	ErrLimitStageInvalidRateOrBurst = "limit stage failed to parse rate or burst"
+	ErrLimitStageByLabelMustDrop    = "When ratelimiting by label, drop must be true"
+	MinReasonableMaxDistinctLabels  = 10000 // 80bytes per rate.Limiter ~ 1MiB memory
 )
 
 var ratelimitDropReason = "ratelimit_drop_stage"
 
 type LimitConfig struct {
-	Rate  float64 `mapstructure:"rate"`
-	Burst int     `mapstructure:"burst"`
-	Drop  bool    `mapstructure:"drop"`
+	Rate              float64 `mapstructure:"rate"`
+	Burst             int     `mapstructure:"burst"`
+	Drop              bool    `mapstructure:"drop"`
+	ByLabelName       string  `mapstructure:"by_label_name"`
+	MaxDistinctLabels int     `mapstructure:"max_distinct_labels"`
 }
 
 func newLimitStage(logger log.Logger, config interface{}, registerer prometheus.Registerer) (Stage, error) {
@@ -34,12 +44,30 @@ func newLimitStage(logger log.Logger, config interface{}, registerer prometheus.
 		return nil, err
 	}
 
-	r := &limitStage{
-		logger:      log.With(logger, "component", "stage", "type", "limit"),
-		cfg:         cfg,
-		dropCount:   getDropCountMetric(registerer),
-		rateLimiter: rate.NewLimiter(rate.Limit(cfg.Rate), cfg.Burst),
+	logger = log.With(logger, "component", "stage", "type", "limit")
+	if cfg.ByLabelName != "" && cfg.MaxDistinctLabels < MinReasonableMaxDistinctLabels {
+		level.Warn(logger).Log(
+			"msg",
+			fmt.Sprintf("max_distinct_labels was adjusted up to the minimal reasonable value of %d", MinReasonableMaxDistinctLabels),
+		)
+		cfg.MaxDistinctLabels = MinReasonableMaxDistinctLabels
 	}
+
+	r := &limitStage{
+		logger:    logger,
+		cfg:       cfg,
+		dropCount: getDropCountMetric(registerer),
+	}
+
+	if cfg.ByLabelName != "" {
+		r.dropCountByLabel = getDropCountByLabelMetric(registerer)
+		newRateLimiter := func() *rate.Limiter { return rate.NewLimiter(rate.Limit(cfg.Rate), cfg.Burst) }
+		gcCb := func() { r.dropCountByLabel.Reset() }
+		r.rateLimiterByLabel = util.NewGenMap[model.LabelValue, *rate.Limiter](cfg.MaxDistinctLabels, newRateLimiter, gcCb)
+	} else {
+		r.rateLimiter = rate.NewLimiter(rate.Limit(cfg.Rate), cfg.Burst)
+	}
+
 	return r, nil
 }
 
@@ -47,15 +75,22 @@ func validateLimitConfig(cfg *LimitConfig) error {
 	if cfg.Rate <= 0 || cfg.Burst <= 0 {
 		return errors.Errorf(ErrLimitStageInvalidRateOrBurst)
 	}
+
+	if cfg.ByLabelName != "" && !cfg.Drop {
+		return errors.Errorf(ErrLimitStageByLabelMustDrop)
+	}
 	return nil
 }
 
 // limitStage applies Label matchers to determine if the include stages should be run
 type limitStage struct {
-	logger      log.Logger
-	cfg         *LimitConfig
-	rateLimiter *rate.Limiter
-	dropCount   *prometheus.CounterVec
+	logger             log.Logger
+	cfg                *LimitConfig
+	rateLimiter        *rate.Limiter
+	rateLimiterByLabel util.GenerationalMap[model.LabelValue, *rate.Limiter]
+	dropCount          *prometheus.CounterVec
+	dropCountByLabel   *prometheus.CounterVec
+	byLabelName        model.LabelName
 }
 
 func (m *limitStage) Run(in chan Entry) chan Entry {
@@ -63,7 +98,7 @@ func (m *limitStage) Run(in chan Entry) chan Entry {
 	go func() {
 		defer close(out)
 		for e := range in {
-			if !m.shouldThrottle() {
+			if !m.shouldThrottle(e.Labels) {
 				out <- e
 				continue
 			}
@@ -72,7 +107,21 @@ func (m *limitStage) Run(in chan Entry) chan Entry {
 	return out
 }
 
-func (m *limitStage) shouldThrottle() bool {
+func (m *limitStage) shouldThrottle(labels model.LabelSet) bool {
+	if m.cfg.ByLabelName != "" {
+		labelValue, ok := labels[model.LabelName(m.cfg.ByLabelName)]
+		if !ok {
+			return false // if no label found, dont ratelimit
+		}
+		rl := m.rateLimiterByLabel.GetOrCreate(labelValue)
+		if rl.Allow() {
+			return false
+		}
+		m.dropCount.WithLabelValues(ratelimitDropReason).Inc()
+		m.dropCountByLabel.WithLabelValues(m.cfg.ByLabelName, string(labelValue)).Inc()
+		return true
+	}
+
 	if m.cfg.Drop {
 		if m.rateLimiter.Allow() {
 			return false
@@ -87,4 +136,10 @@ func (m *limitStage) shouldThrottle() bool {
 // Name implements Stage
 func (m *limitStage) Name() string {
 	return StageTypeLimit
+}
+
+func getDropCountByLabelMetric(registerer prometheus.Registerer) *prometheus.CounterVec {
+	return util.RegisterCounterVec(registerer, "logentry", "dropped_lines_by_label_total",
+		"A count of all log lines dropped as a result of a pipeline stage",
+		[]string{"label_name", "label_value"})
 }
--- a/docs/sources/clients/promtail/stages/limit.md
+++ b/docs/sources/clients/promtail/stages/limit.md
@@ -17,6 +17,13 @@ limit:
 
   # The cap in the quantity of burst lines that Promtail will push to Loki
   [burst: <int>]
+   
+  # Ratelimit each label value independently. If label is not found, log line is not
+  # considered for ratelimiting. Drop must be true if this is set.
+  [by_label_name: <string>]  
+    
+  # When ratelimiting by label is enabled, keep track of this many last used labels
+  [max_distinct_labels: <int> | default = 10000]  
 
   # When drop is true, log lines that exceed the current rate limit will be discarded.
   # When drop is false, log lines that exceed the current rate limit will only wait
@@ -56,3 +63,18 @@ Given the pipeline:
 ```
 
 Would throttle any log line and drop logs when rate limit.
+
+#### Ratelimit by a label
+
+Given the pipeline:
+
+```yaml
+- limit:
+    rate: 10
+    burst: 10
+    drop: true
+    by_label_name: "namespace"
+```
+
+Would ratelimit messages originating from each namespace independently.
+Any message without namespace label will not be ratelimited.
--- a/pkg/util/map.go
+++ b/pkg/util/map.go
@@ -1,0 +1,39 @@
+package util
+
+type GenerationalMap[K comparable, V any] struct {
+	oldgen map[K]V
+	newgen map[K]V
+
+	maxSize int
+	newV    func() V
+	gcCb    func()
+}
+
+// NewGenMap created which maintains at most maxSize recently used entries
+func NewGenMap[K comparable, V any](maxSize int, newV func() V, gcCb func()) GenerationalMap[K, V] {
+	return GenerationalMap[K, V]{
+		newgen:  make(map[K]V),
+		maxSize: maxSize,
+		newV:    newV,
+		gcCb:    gcCb,
+	}
+}
+
+func (m *GenerationalMap[K, T]) GetOrCreate(key K) T {
+	v, ok := m.newgen[key]
+	if !ok {
+		if v, ok = m.oldgen[key]; !ok {
+			v = m.newV()
+		}
+		m.newgen[key] = v
+
+		if len(m.newgen) == m.maxSize {
+			m.oldgen = m.newgen
+			m.newgen = make(map[K]T)
+			if m.gcCb != nil {
+				m.gcCb()
+			}
+		}
+	}
+	return v
+}
--- a/pkg/util/metrics_helper.go
+++ b/pkg/util/metrics_helper.go
@@ -805,3 +805,23 @@ type CollectorVec interface {
 	prometheus.Collector
 	Delete(labels prometheus.Labels) bool
 }
+
+// RegisterCounterVec registers new CounterVec with given name,namespace and labels.
+// If metric was already registered it returns existing instance.
+func RegisterCounterVec(registerer prometheus.Registerer, namespace, name, help string, labels []string) *prometheus.CounterVec {
+	vec := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: namespace,
+		Name:      name,
+		Help:      help,
+	}, labels)
+	err := registerer.Register(vec)
+	if err != nil {
+		if existing, ok := err.(prometheus.AlreadyRegisteredError); ok {
+			vec = existing.ExistingCollector.(*prometheus.CounterVec)
+		} else {
+			// Same behavior as MustRegister if the error is not for AlreadyRegistered
+			panic(err)
+		}
+	}
+	return vec
+}
--- a/clients/pkg/logentry/stages/limit_test.go
+++ b/clients/pkg/logentry/stages/limit_test.go
@@ -38,10 +38,31 @@ pipeline_stages:
     drop: true
 `
 
+var testLimitByLabelYaml = `
+pipeline_stages:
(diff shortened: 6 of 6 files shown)


liguozhong:
hi, thanks for your contribution.
I have a question.
The configuration of this PR is probably like this.

```yaml
- limit:
    rate: 10
    burst: 10
    drop: true
    by_label_name: "namespace"
```
Is it possible to achieve the same effect with match+rate stage in such a scenario.
```yaml
match:
  selector: '{namespace="foo"}"'
  stages:
  - limit:
      rate: 5
      drop: true
match:
  selector: '{namespace!="foo"}"'
  stages:
  - limit:
      rate: 100
      drop: false
```

redbaron:
No, because in your case "namespace=abc" and "namespace=xyz" will be ratelimited together, only "namespace=foo" will have it's own ratelimiter.

liguozhong:
If user configure rate limit based on traceid label, will this pr make promtail oom.

- limit:
    rate: 10
    burst: 10
    drop: true
    by_label_name: "traceID"

redbaron:
@liguozhong , added max size cap. Implementation is completely free until size cap is reached at which point additional cost is just 1 lookup +  1 insert per each in-use label value.

liguozhong:
> @liguozhong , added max size cap. Implementation is completely free until size cap is reached at which point additional cost is just 1 lookup + 1 insert per each in-use label value once.

👍，thanks for your contribution.

Reimplementing a map with limited expiration requires more consideration, such as whether the map is a thread-safe map ? Why not use an existing map?

```
import	lru "github.com/hashicorp/golang-lru"
*lru.Cache
```

redbaron:
I considered using LRU, but decided against , because LRU is not what we need. We don't have requirements to take out entry on insert beyond capacity, our map values are small enough (pointer) that we can afford to keep 2x of them with neglible increase in memory use.


Core logic of capped map ended up being just 13 lines, therefore I couldn't justify pulling another much much more complicated dependency, especially given that we don't  need LRU.

Keep in mind, that end goal is to prevent crash on misconfiguration, like with your trace id example, nothing else. Normal mode of operation of properly configured system will not reach capacity , therefore we should optimize for that. Proposed  solution in a steady state is indistinguishable from a simple map (which it is) both for CPU and memory utilization.

as for multithreading, my undertandig is that  Stage.Run never runs concurrently, therefore thread safety is  a non requirement in this particular case.

liguozhong:
[bugfix] promtail:Concurrent map read and map write in promtail pod #7206

https://github.com/grafana/loki/pull/7206

No. promtail stage is not always thread safety. for example issue https://github.com/grafana/loki/issues/7180

redbaron:
These are pipelines, right?  Within pipeline instance, every stage's `Run` is called once. Limit stage spawns a single goroutine in it's Run, so there is no concurrent access to the same stage Instance.

liguozhong:
> [bugfix] promtail:Concurrent map read and map write in promtail pod #7206

[bugfix] promtail:Concurrent map read and map write in promtail pod https://github.com/grafana/loki/pull/7206

From this issue, in extreme cases, the pipeline may not be thread-safe. I'm still analyzing this issue. Why does map concurrent access occur?

liguozhong:
Today I carefully checked the pipeline code, and the 2.7.0 code should be thread-safe.

redbaron:
good to merge, then? :)

liguozhong:
this new metrics may cause prometheus OOM due to high cardinality label values.
Wouldn't it be better to replace metrics with log?

liguozhong:
and add testcase for generationalMap

redbaron:
> such as introducing a callback interface and triggering close interfaces to recycle resources when a key expires.
this is not an easy thing

I don't understand what it is for. None of that is required for the current use of it.

redbaron:
> this new metrics may cause prometheus OOM due to high cardinality label values.

I am not familiar with the prometheus metrics instrumentation, how unbounded cardinality problem is resolved elsewhere?

liguozhong:
thanks.

I don't recommend introducing such overly detailed metrics, just print out the specific label through the log.
Maybe we can wait for more developers in the community to take a look.

liguozhong:
The "lookups" variable does not seem to be referenced.

clients/pkg/logentry/stages/limit.go
```diff
@@ -12,14 +15,52 @@ import (
...
+func (m *generationalMap[K, T]) GetOrCreate(key K) T {
+	v, ok := m.newgen[key]
+	if !ok {
+		if v, ok = m.oldgen[key]; !ok {
+			v = m.newV()
+		}
+		m.newgen[key] = v
+
+		if len(m.newgen) == m.maxSize {
+			m.oldgen = m.newgen
```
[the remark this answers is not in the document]
liguozhong: `r.rateLimiterByLabel.Reset()`

clients/pkg/logentry/stages/match.go
```diff
@@ -114,11 +114,23 @@ func newMatcherStage(logger log.Logger, jobName *string, config interface{}, reg
 }
 
 func getDropCountMetric(registerer prometheus.Registerer) *prometheus.CounterVec {
+	return getLogEntryMetric(registerer, "dropped_lines_total",
+		"A count of all log lines dropped as a result of a pipeline stage",
+		[]string{"reason"})
+}
+
+func getDropCountByLabelMetric(registerer prometheus.Registerer) *prometheus.CounterVec {
+	return getLogEntryMetric(registerer, "dropped_lines_by_label",
```
liguozhong, line 123: `	return getLogEntryMetric(registerer, "dropped_lines_by_label_total",
`

liguozhong, line 123:
```suggestion
	return getLogEntryMetric(registerer, "dropped_lines_by_label_total",
```

clients/pkg/logentry/stages/limit.go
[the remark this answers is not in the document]
redbaron: We call reset via GC callback in:  https://github.com/grafana/loki/pull/7597/files#diff-f1c58a87f36177864bda1e254f57ee1f44d6f3ffb01432aeb2be3f75f0552a65R100

clients/pkg/logentry/stages/match.go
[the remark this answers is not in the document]
redbaron: In intentended usecase it will not be high cardinality metric. In case of pathalogical misconfiguration, like ratelimiting based on trace id, we reset all counts every  `max_distinct_labels` unique labels values processed to prevent OOM.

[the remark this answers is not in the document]
redbaron: I'd like to remind that ratelimiting itself is there to prevent excessive resource utilization in case of a misbehaving app. In a normal case no ratelimiting will happen and no dropped count metric will ever be created. Existence of the dropped counter by metric is a rare occasion and is definitely of interest to system administrators.

Dropped counter for every label when label itself is high cardinality is a rare occasion squared, but even that case is now handled.

clients/pkg/logentry/stages/limit.go
```diff
@@ -35,35 +38,49 @@ func newLimitStage(logger log.Logger, config interface{}, registerer prometheus.
...
 // limitStage applies Label matchers to determine if the include stages should be run
 type limitStage struct {
-	logger      log.Logger
-	cfg         *LimitConfig
-	rateLimiter *rate.Limiter
-	dropCount   *prometheus.CounterVec
+	logger             log.Logger
+	cfg                *LimitConfig
+	rateLimiter        *rate.Limiter
+	rateLimiterByLabel map[model.LabelValue]*rate.Limiter
```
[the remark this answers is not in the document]
redbaron: Gave my motivation not to use LRU in https://github.com/grafana/loki/pull/7597#issuecomment-1315244423

redbaron:
> Maybe we can wait for more developers in the community to take a look.

Agree, maybe @kavirajk or @trevorwhitney can have a look

liguozhong:
I left 2 small comment, other parts LGTM.
thank you very much for your contribution. great job👍

redbaron:
> I left 2 small comment, other parts LGTM.
thank you very much for your contribution. great job👍

I can't see any new comments after my last message

liguozhong:
> > I left 2 small comment, other parts LGTM.
> > thank you very much for your contribution. great job👍
> 
> I can't see any new comments after my last message

I submit again yet

clients/pkg/logentry/stages/match.go
```diff
@@ -114,11 +114,23 @@ func newMatcherStage(logger log.Logger, jobName *string, config interface{}, reg
 }
 
 func getDropCountMetric(registerer prometheus.Registerer) *prometheus.CounterVec {
+	return getLogEntryMetric(registerer, "dropped_lines_total",
+		"A count of all log lines dropped as a result of a pipeline stage",
+		[]string{"reason"})
+}
+
+func getDropCountByLabelMetric(registerer prometheus.Registerer) *prometheus.CounterVec {
```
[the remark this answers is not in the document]
redbaron: done, moved this function to `limit.go`. Change to `match.go` was still required, because `limit.go` is  using `getDropCountMetric` from it and that function was refactored to become a wrapper around new `getLogEntryMetric`

clients/pkg/logentry/stages/limit.go
```diff
@@ -12,14 +15,56 @@ import (
 
 const (
 	ErrLimitStageInvalidRateOrBurst = "limit stage failed to parse rate or burst"
+	ErrLimitStageByLabelMustDrop    = "When ratelimiting by label, drop must be true"
+	MinReasonableMaxDistinctLabels  = 10000 // 80bytes per rate.Limiter ~ 1MiB memory
 )
 
+type generationalMap[K comparable, V any] struct {
```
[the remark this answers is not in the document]
redbaron: done

clients/pkg/logentry/stages/match.go
```diff
@@ -114,11 +114,17 @@ func newMatcherStage(logger log.Logger, jobName *string, config interface{}, reg
...
 func getDropCountMetric(registerer prometheus.Registerer) *prometheus.CounterVec {
+	return getLogEntryMetric(registerer, "dropped_lines_total",
+		"A count of all log lines dropped as a result of a pipeline stage",
+		[]string{"reason"})
+}
+
+func getLogEntryMetric(registerer prometheus.Registerer, name, help string, labels []string) *prometheus.CounterVec {
 	dropCount := prometheus.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "logentry",
-		Name:      "dropped_lines_total",
```
liguozhong, line 119: It would be better if this PR can not modify `match.go`. Other parts LGTM

redbaron: removed any changes from match.go

liguozhong (approved):
LGTM, thanks for your contributor👍

redbaron:
@kavirajk , would you kindly have a look?

pkg/util/map.go
```diff
@@ -0,0 +1,39 @@
...
+	oldgen map[K]V
+	newgen map[K]V
+
+	maxSize int
+	newV    func() V
+	gcCb    func()
+}
+
+// NewGenMap created which maintains at most maxSize recently used entries
+func NewGenMap[K comparable, V any](maxSize int, newV func() V, gcCb func()) GenerationalMap[K, V] {
```
kavirajk, line 13: This would our first generic code in Loki I believe. Will wait for few more eyes from contributors.

Can we make it bit more readable may be? failing to understand what are `newV` and `gcCb` means here. Can we have may be better naming and some descriptive comment on those?

If we going with this generic map then, a have few tests for this map would be good.

kavirajk: Just started to think why we need this Map in the first place? I see it's getting used only in one place and could be simple `map[string]*rate.Limiter`?. Trying to understand the rationale here.

clients/pkg/logentry/stages/limit.go
```diff
@@ -34,36 +42,61 @@ func newLimitStage(logger log.Logger, config interface{}, registerer prometheus.
 		return nil, err
 	}
 
+	logger = log.With(logger, "component", "stage", "type", "limit")
+	if cfg.ByLabelName != "" && cfg.MaxDistinctLabels < MinReasonableMaxDistinctLabels {
+		level.Warn(logger).Log(
+			"msg",
+			fmt.Sprintf("max_distinct_labels was adjusted up to the minimal reasonable value of %d", MinReasonableMaxDistinctLabels),
```
kavirajk, line 49: Any rationale behind not to have `< MinReasonableMaxDistrinctLabels`?

pkg/util/metrics_helper.go
```diff
@@ -805,3 +805,23 @@ type CollectorVec interface {
 	prometheus.Collector
 	Delete(labels prometheus.Labels) bool
 }
+
+// RegisterCounterVec registers new CounterVec with given name,namespace and labels.
+// If metric was already registered it returns existing instance.
+func RegisterCounterVec(registerer prometheus.Registerer, namespace, name, help string, labels []string) *prometheus.CounterVec {
```
kavirajk, line 811: Q: what's the problem in using existing `prometheus.MustRegister` helper here? I would let it panic in case of duplicate registration rather than handling it via `promtheus.AlreadyRegisteredError` handling. Am I missing any specific use case here?

docs/sources/clients/promtail/stages/limit.md
```diff
@@ -56,3 +63,18 @@ Given the pipeline:
 ```
 
 Would throttle any log line and drop logs when rate limit.
+
+#### Ratelimit by a label
```
kavirajk, line 67: I think it's worth mentioning what is the difference between these two.
```
- limit:
    rate: 10
    burst: 10
    drop: true
    by_label_name: "namespace"
```
with this

```
match:
  selector: '{namespace="foo"}"'
  stages:
  - limit:
      rate: 5
      drop: true
match:
  selector: '{namespace!="foo"}"'
  stages:
  - limit:
      rate: 100
      drop: false
```

Like @liguozhong already mentioned, even my first thought would be to use latter config instead of this new config.

I see you explained, with `by_label_name` we can get separate ratelimiter per label. 

Wondering what advantage does it serve?

kavirajk (commented):
thanks @redbaron for the contribution and @liguozhong for the review.

Added a few comments on my own. My main concern is what advantage does this `by_label_name` serve over using `matcher` stage with any labels..

clients/pkg/logentry/stages/limit.go
```diff
@@ -34,36 +42,61 @@ func newLimitStage(logger log.Logger, config interface{}, registerer prometheus.
...
-		rateLimiter: rate.NewLimiter(rate.Limit(cfg.Rate), cfg.Burst),
+		logger:    logger,
+		cfg:       cfg,
+		dropCount: getDropCountMetric(registerer),
+	}
+
+	if cfg.ByLabelName != "" {
+		r.dropCountByLabel = getDropCountByLabelMetric(registerer)
+		newRateLimiter := func() *rate.Limiter { return rate.NewLimiter(rate.Limit(cfg.Rate), cfg.Burst) }
+		gcCb := func() { r.dropCountByLabel.Reset() }
```
vlad-diachenko, line 63: we do not need to reset the counter. we only increment it .....

vlad-diachenko (changes requested):
I definitely like this feature but I think we could improve the implementation before merging it ;) 

PS: generics are one love ;)

pkg/util/map.go
redbaron: newV - new value
gcCb - garbage collector callback

I am happy to use any names you suggest for these or any other identifiers.

Purpose of this map is to keep maximum number of last used entries and GC the rest. Difference to LRU is that it doesn't pessimize common case of NOT reaching the limit.

clients/pkg/logentry/stages/limit.go
```diff
@@ -88,3 +135,9 @@ func (m *limitStage) shouldThrottle() bool {
 func (m *limitStage) Name() string {
 	return StageTypeLimit
 }
+
+func getDropCountByLabelMetric(registerer prometheus.Registerer) *prometheus.CounterVec {
+	return util.RegisterCounterVec(registerer, "logentry", "dropped_lines_by_label_total",
+		"A count of all log lines dropped as a result of a pipeline stage",
+		[]string{"label_name", "label_value"})
```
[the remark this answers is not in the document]
redbaron: ratelimiting itself is a safeguard against misbehaving apps. Expectation is that in the common case ratelimit wont be hit, but if it is hit you'd want to  know which app is being ratelimited, hence we report both label name and value.

In case too many label values are being ratelimited (and hence their counters created) on a grossly misconfigured setup, we cap memory usage by `Reset()` counter every time we GC to maintain maximum configured capacity.

clients/pkg/logentry/stages/limit.go
```diff
@@ -12,14 +16,18 @@ import (
...
 
 type LimitConfig struct {
-	Rate  float64 `mapstructure:"rate"`
-	Burst int     `mapstructure:"burst"`
-	Drop  bool    `mapstructure:"drop"`
+	Rate              float64 `mapstructure:"rate"`
+	Burst             int     `mapstructure:"burst"`
+	Drop              bool    `mapstructure:"drop"`
+	ByLabelName       string  `mapstructure:"by_label_name"`
+	MaxDistinctLabels int     `mapstructure:"max_distinct_labels"`
```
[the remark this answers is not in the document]
redbaron: I considered separate stage, but decidee against it as configs are too similar. Any ratelimiter need rate and burst params. There just 2 new params and users is not forced to use them:

- `by_label_name`  opt-in, offers change mode of operation to do ratelimiting per label, not globally per stage.
- `max_distinct_labels` have reasonable default of 10000, so  they are unlikely need to change it at all and if they do, then they operate on  a scale requiring a lot of engineering effort and increasing a limit wont be any problem for them.

clients/pkg/logentry/stages/limit.go
```diff
@@ -34,36 +42,61 @@ func newLimitStage(logger log.Logger, config interface{}, registerer prometheus.
...
+		logger:    logger,
+		cfg:       cfg,
+		dropCount: getDropCountMetric(registerer),
+	}
+
+	if cfg.ByLabelName != "" {
+		r.dropCountByLabel = getDropCountByLabelMetric(registerer)
+		newRateLimiter := func() *rate.Limiter { return rate.NewLimiter(rate.Limit(cfg.Rate), cfg.Burst) }
+		gcCb := func() { r.dropCountByLabel.Reset() }
+		r.rateLimiterByLabel = util.NewGenMap[model.LabelValue, *rate.Limiter](cfg.MaxDistinctLabels, newRateLimiter, gcCb)
```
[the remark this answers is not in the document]
redbaron: > for each label value, we will have a separate instance of *rate.Limiter . it will blow up our memory in case we have a lot of instances of stages... 

We don't instantiate Limiter stages dynamically here, stage instances are mapped 1:1 to what defined in the config.  Each stage has fixed `label_name` it cares about and that name is configured by user statically, so there should be no cardinality problem.

clients/pkg/logentry/stages/limit.go
redbaron: > and user need to increase it should they need more.

Just to clarify this bit, user need to increase limit in case if number actively hit ratelimiters exceed it or they'll have innacurate ratelimiting (more entrics will go through then should have). Should some labels values be transient and NOT being in active use (lets say pod_name) , then  it is perfectly fine for the system to reach max limit and get then GCed, there is no need to increase limit to accomodate for them.

pkg/util/map.go
```diff
@@ -0,0 +1,39 @@
...
+		if len(m.newgen) == m.maxSize {
+			m.oldgen = m.newgen
+			m.newgen = make(map[K]T)
+			if m.gcCb != nil {
+				m.gcCb()
+			}
+		}
+	}
+	return v
+}
```
Supra89kren, line 39: sorry, for such late comment, but this code must be thread-safe....please add mutex.
I am pretty sure that this method will be called from multiple goroutines...

Supra89kren:
hey @redbaron . thanks for rapid reply ;) 
In general, I misunderstood the goal of this implementation. Now it's much clear. will double check the PR and clean up my irrelevant comments

pkg/util/map.go
```diff
@@ -0,0 +1,39 @@
...
+func (m *GenerationalMap[K, T]) GetOrCreate(key K) T {
+	v, ok := m.newgen[key]
+	if !ok {
+		if v, ok = m.oldgen[key]; !ok {
+			v = m.newV()
+		}
+		m.newgen[key] = v
+
+		if len(m.newgen) == m.maxSize {
+			m.oldgen = m.newgen
```
Supra89kren, line 31: can we add some mechanism to clean up oldgen after some time if it's untouched?
I think that it might be a case when we have a spike of values, so we populate `oldgen`, and after that we remove stabilise our values, but rateLimitters will live in oldgen forever...

clients/pkg/logentry/stages/limit.go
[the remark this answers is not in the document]
Supra89kren: ok, i see.

clients/pkg/logentry/stages/limit.go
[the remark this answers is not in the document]
Supra89kren: sure, make sense.

clients/pkg/logentry/stages/limit.go
```diff
@@ -12,14 +16,18 @@ import (
 
 const (
 	ErrLimitStageInvalidRateOrBurst = "limit stage failed to parse rate or burst"
+	ErrLimitStageByLabelMustDrop    = "When ratelimiting by label, drop must be true"
+	MinReasonableMaxDistinctLabels  = 10000 // 80bytes per rate.Limiter ~ 1MiB memory
```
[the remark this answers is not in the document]
Supra89kren: irrelevant comment) i misunderstood the idea of the feature

clients/pkg/logentry/stages/limit.go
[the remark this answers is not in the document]
vlad-diachenko: I see a lot of cases when we might continue reporting thousands of the metrics, (for each label value) ...
for example: 
user set `MaxDistinctLabels` to 100_000.
90_000 of streams are got rate limited, so we started reporting 90_000 metrics... and we will continue reporting these metrics forever if we do not reach `MaxDistinctLabels`... until we call `gcCb`.
Moreover, if you have 100 instances of promtail, and all of them report 90_000 metrics you will have 9_000_000 metrics ...
I also think that it would be useful to see this metric, but we can not do it by adding such unique values as labels of the metrics...

vlad-diachenko:
hey @redbaron . thanks for rapid reply ;) 
In general, I misunderstood the goal of this implementation. Now it's much clear. will double check the PR and clean up my irrelevant comments

pkg/util/map.go
vlad-diachenko: you might be right ;) i have double checked and have not found any proves that it can be called from a few goroutines ;) thanks

clients/pkg/logentry/stages/limit.go
vlad-diachenko: if we do not add label value as a label to the metric, we can avoid using `reset()` function

clients/pkg/logentry/stages/limit.go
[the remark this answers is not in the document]
redbaron: In general having high cardinality label in Loki is a nogo anyway. Also reporting stale metrics potentially forever is a long standing problem with Prometheus instrumentation and whole ecosystem learned to live with it. 

Label value cardinality is specific to every use case, but in general it is expected that:
- low cardinality labels are used 
- ratelimits are set so that in normal operation mode they are not triggered

For scenario you worry about both have be false and even then worst thing happening is just 1 extra metric per label.

As an administrator of log shipping I need to know which teams to contact to tame their log usage and this metric gives me this visibility. In the absence of per label metric  all I have is global drop count and it wont be enough.

clients/pkg/logentry/stages/limit.go
redbaron: true that

clients/pkg/logentry/stages/limit.go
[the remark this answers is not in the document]
vlad-diachenko: completely agree with you ;) it is super useful to have this metric with label_value.... and considering that in normal operation mode it should not be a case, then I agree that we can leave it ;)

redbaron:
> hey @redbaron . thanks for rapid reply ;) In general, I misunderstood the goal of this implementation. Now it's much clear. will double check the PR and clean up my irrelevant comments

If I read output correctly, it reports files I didn't change in this PR.

vlad-diachenko:
> > hey @redbaron . thanks for rapid reply ;) In general, I misunderstood the goal of this implementation. Now it's much clear. will double check the PR and clean up my irrelevant comments
> 
> If I read output correctly, it reports files I didn't change in this PR.

level=info msg="Fix issue &result.Issue{FromLinter:\"goimports\", Text:\"File is not `goimports`-ed with -local github.com/grafana/loki/pkg,github.com/grafana/loki/tools\", Severity:\"\", SourceLines:[]string{\"\"}, Replacement:(*result.Replacement)(0xc098805850), Pkg:(*packages.Package)(0xc001b4f380), LineRange:(*result.Range)(nil), Pos:token.Position{Filename:\"clients/pkg/logentry/stages/limit.go\", Offset:0, Line:9, Column:0}, HunkPos:0, ExpectNoLint:false, ExpectedNoLintLinter:\"\"} with range {9 9}" 

clients/pkg/logentry/stages/limit.go

import must be sorted and split into 3 groups:
1. standard libraries 
2. packages from this/grafana repo
3. external modules

try to move the imports and run `make lint` command. 

I will be partially unavailable during the next few days... might be slow in replies... let me know if you need any help

pkg/util/map.go
redbaron: Potentially I could trigger gc on timer, but it doesn't feel right to do, because:

- it affects ratelimiting accuracy
- because of the above GC period likely need to be exposed as configuration variable for users to tune
- net win is 1-2MiB tops in memory and with no win in CPU utilization

overall it doesn't worth it IMHO

redbaron:
@vlad-diachenko , there were unrelated errors reported, so I squashed and rebased on top of latest master.

CHANGELOG.md
```diff
@@ -78,6 +78,7 @@ Check the history of the branch FIXME.
 * [6179](https://github.com/grafana/loki/pull/6179) **chaudum**: Add new HTTP endpoint to delete ingester ring token file and shutdown process gracefully
 * [5997](https://github.com/grafana/loki/pull/5997) **simonswine**: Querier: parallize label queries to both stores.
 * [5406](https://github.com/grafana/loki/pull/5406) **ctovena**: Revise the configuration parameters that configure the usage report to grafana.com.
+* [7597](https://github.com/grafana/loki/pull/7597) **redbaron**: allow ratelimiting by label
```
vlad-diachenko, line 81: Hey @redbaron . sorry, for the delay. was on sick leave. 
Thanks for addressing all the comments. I am happy to merge this PR. 
However, the last thing that I ask you to fix is to move the changelog entry that you added. now, it's placed in V2.7.0  changelogs ... could you please move it to the next line after https://github.com/grafana/loki/pull/7597/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR36

vlad-diachenko: maybe it was moved when you pulled the latest main.... sorry for so many cycles of the review ;)

vlad-diachenko: nevermind, I am fixing it

vlad-diachenko (approved):
LGTM . thanks @redbaron 🚀

PR #7599, opened 2022-11-04
Keep rates in the ratestore for 10 minutes 

MasslessParticle:
This PR accounts for intermittent streams that might get dropped on a per-second update interval

### DIFF
--- a/pkg/distributor/ratestore.go
+++ b/pkg/distributor/ratestore.go
@@ -47,13 +47,19 @@ type ingesterClient struct {
 	client logproto.StreamDataClient
 }
 
+type expiringRate struct {
+	createdAt time.Time
+	rate      int64
+}
+
 type rateStore struct {
 	services.Service
 
 	ring            ring.ReadRing
 	clientPool      poolClientFactory
-	rates           map[string]int64
+	rates           map[string]expiringRate
 	rateLock        sync.RWMutex
+	rateKeepAlive   time.Duration
 	ingesterTimeout time.Duration
 	maxParallelism  int
 	limits          Limits
@@ -67,8 +73,10 @@ func NewRateStore(cfg RateStoreConfig, r ring.ReadRing, cf poolClientFactory, l 
 		clientPool:      cf,
 		maxParallelism:  cfg.MaxParallelism,
 		ingesterTimeout: cfg.IngesterReqTimeout,
+		rateKeepAlive:   10 * time.Minute,
 		limits:          l,
 		metrics:         newRateStoreMetrics(registerer),
+		rates:           make(map[string]expiringRate),
 	}
 
 	rateCollectionInterval := util.DurationWithJitter(cfg.StreamRateUpdateInterval, 0.2)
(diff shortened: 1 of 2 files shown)


pkg/distributor/ratestore.go
```diff
@@ -98,11 +106,28 @@ func (s *rateStore) updateAllRates(ctx context.Context) error {
...
-
 	s.rateLock.Lock()
 	defer s.rateLock.Unlock()
-	s.rates = rates
+
+	for stream, rate := range rates {
+		s.rates[stream] = expiringRate{
+			rate:      rate,
+			createdAt: time.Now(),
+		}
```
DylanGuedes, line 116: i think you can check for `time.Since(rate.createdAt) > s.rateKeepAlive` and update maxRate here and get rid of the next iteration. the bigO complexity is the same but I feel like it should have better performance since you would avoid calling `delete`

pkg/distributor/ratestore.go
```diff
@@ -244,7 +264,7 @@ func (s *rateStore) RateFor(tenant string, streamHash uint64) int64 {
 	s.rateLock.RLock()
 	defer s.rateLock.RUnlock()
 
-	return s.rates[key(tenant, streamHash)]
+	return s.rates[key(tenant, streamHash)].rate
```
DylanGuedes, line 267: double checking: there's no way `streamHash` won't have an entry in s.rates, right? because if it is possible, it means invoking `.rate` will panic, so a nilcheck would be necessary

pkg/distributor/ratestore_test.go
```diff
@@ -156,6 +156,33 @@ func TestRateStore(t *testing.T) {
...
+			}, 1),
+		}
+
+		tc.rateStore.rateKeepAlive = 2 * time.Second
+		_ = tc.rateStore.StartAsync(context.Background())
+		defer tc.rateStore.StopAsync()
+
+		require.Eventually(t, func() bool {
+			return 25 == tc.rateStore.RateFor("tenant 1", 0)
+		}, time.Second, time.Millisecond)
```
DylanGuedes, line 180: nit: checking every millisecond with a 1 second sounds too much imo
```suggestion
		}, time.Second, 100*time.Millisecond)
```

DylanGuedes (approved):
a few suggestions, but lgtm!

pkg/distributor/ratestore.go
MasslessParticle: I think we need to do both: There is no guarantee that both maps contain the same entries so first we merge them and then we iterate over the whole map to make sure we don't miss anything.

pkg/distributor/ratestore.go
MasslessParticle: `expiringRate` is a value not a pointer so this would just return 0.

MasslessParticle: Here is an example: https://go.dev/play/p/DLx0w-K2d9A

DylanGuedes: ops, thats right. resolved!

pkg/distributor/ratestore.go
DylanGuedes: got it, at first I thought that all entries on `s.rates` would be on the initial `rates`.

PR #7619, opened 2022-11-07
Loki: Modified heroku drain target to make any url query parameters available as labels

cadrake:

This PR reads the url query parameters included in Heroku drain POST request and creates `__param_<name>` labels similar to how the parameters work with prometheus metrics and metric endpoints as a way to allow setting the true application name when ingesting logs from a Heroku drain.

Fixes #<issue number>

The existing `app` field coming from Heroku only specifies application source (either the application itself or Heroku) and thus is insufficient for uniquely identifying a log lines source.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


cadrake:
Note that I didnt update documentation because the section on Heroku drains already mentions url parameters being available as `__param_<name>` labels; however, that looks to have possibly been a typo or copied from another section and not updated.

thepalbi:
> Note that I didnt update documentation because the section on Heroku drains already mentions url parameters being available as `__param_<name>` labels; however, that looks to have possibly been a typo or copied from another section and not updated.

Hi @cadrake , thanks for bringing this up. As for what you mention in the docs, I think you refer to this:

> The __param_<name> label is set to the value of the first passed URL parameter called <name>.

That quoted sentence comes from the [relabel_configs](https://grafana.com/docs/loki/next/clients/promtail/configuration/#relabel_configs) section, and it refers that if you use a relabel_config with `__param_<name>` it will affect how the URL takes shape in the scraped targets. Almost sure this doesn't apply in this case, but I agree the documentation is kid of misleading, will take this back and correct it.

However, I think what you brought up in this PR is a nice addition to the Heroku target. Commenting on the PR.

clients/pkg/promtail/targets/heroku/target.go
```diff
@@ -117,6 +117,12 @@ func (h *Target) drain(w http.ResponseWriter, r *http.Request) {
 			ts = message.Timestamp
 		}
 
+		// Create __param_<name> labels from query parameters
+		params := r.URL.Query()
+		for k, v := range params {
+			lb.Set(fmt.Sprintf("__param_%s", k), strings.Join(v, ","))
```
thepalbi, line 123: Can you change the temporary label name to something like `__heroku_drain_param_%s` to be consistent with the other extracted labels.

thepalbi (changes requested):
In order for this PR to be sound, we need some other changes:
- [ ] Update docs [here](https://github.com/grafana/loki/blob/main/docs/sources/clients/promtail/configuration.md#available-labels-2) and [here](https://github.com/grafana/loki/blob/main/docs/sources/clients/promtail/scraping.md#heroku-drain) adding the new available labels
- [ ] Add a test case covering query params being extracted AND dropped
- [ ] Add entry to changelog [here](https://github.com/grafana/loki/blob/main/CHANGELOG.md?plain=1#L20)

clients/pkg/promtail/targets/heroku/target.go
cadrake: Sure

cadrake:
For the unit tests, can I piggy-back off of the existing tests or would you like me to create new ones?

thepalbi:
> For the unit tests, can I piggy-back off of the existing tests or would you like me to create new ones?

I think copying one of [these](https://github.com/grafana/loki/blob/main/clients/pkg/promtail/targets/heroku/target_test.go#L213) as a new test exercising the URL with query params will do

cadrake:
I ended up expanding `TestHerokuDrainTarget` to run additional tests with and without query parameters, also updated all the documentation and changelog

docs/sources/clients/promtail/scraping.md
```diff
@@ -458,14 +458,34 @@ The [server](../configuration.md#server) section configures the HTTP server crea
...
-Before using a `heroku_drain` target, Heroku should be configured with the URL where the Promtail instance will be listening. 
+Before using a `heroku_drain` target, Heroku should be configured with the URL where the Promtail instance will be listening.
 Follow the steps in [Heroku HTTPS Drain docs](https://devcenter.heroku.com/articles/log-drains#https-drains) for using the Heroku CLI
 with a command like the following:
 
 ```
 heroku drains:add [http|https]://HOSTNAME:8080/heroku/api/v1/drain -a HEROKU_APP_NAME
 ```
 
+### Getting the Heroku application name
```
thepalbi, line 469: 😮‍💨

docs/sources/clients/promtail/scraping.md
```diff
@@ -53,9 +53,9 @@ There are different types of labels present in Promtail:
 - The `__path__` label is a special label which Promtail uses after discovery to
   figure out where the file to read is located. Wildcards are allowed, for example `/var/log/*.log` to get all files with a `log` extension in the specified directory, and `/var/log/**/*.log` for matching files and directories recursively. For a full list of options check out the docs for the [library](https://github.com/bmatcuk/doublestar) Promtail uses.
 
-- The `__path_exclude__` label is another special label Promtail uses after 
-  discovery, to exclude a subset of the files discovered using `__path__` from 
-  being read in the current scrape_config block. It uses the same 
+- The `__path_exclude__` label is another special label Promtail uses after
```
thepalbi, line 56: nit: there's a bunch of extra-spaces deleted that adds unnecessary diffs, can we remove them?

docs/sources/clients/promtail/configuration.md
```diff
@@ -1091,7 +1091,7 @@ The list of labels below are discovered when consuming kafka:
 - `__meta_kafka_partition`: The partition id where the message has been read.
 - `__meta_kafka_member_id`: The consumer group member id.
 - `__meta_kafka_group_id`: The consumer group id.
-- `__meta_kafka_message_key`: The message key. If it is empty, this value will be 'none'. 
+- `__meta_kafka_message_key`: The message key. If it is empty, this value will be 'none'.
```
thepalbi, line 1094: nit: Same in this file regarding deleted spaces making the diff noisy

thepalbi (approved):
LGTM. Left some nits regarding extra-spaces. Will bump this through Loki to get a review from them, since it's required.

docs/sources/clients/promtail/scraping.md
cadrake: Yeah sorry that was my editor, Ill fix that

clients/pkg/promtail/targets/heroku/target.go
```diff
@@ -117,6 +117,12 @@ func (h *Target) drain(w http.ResponseWriter, r *http.Request) {
 			ts = message.Timestamp
 		}
 
+		// Create __heroku_drain_param_<name> labels from query parameters
+		params := r.URL.Query()
+		for k, v := range params {
+			lb.Set(fmt.Sprintf("__heroku_drain_param_%s", k), strings.Join(v, ","))
```
slim-bean, line 123: Is there a reason to prefix the keys? would it be better for users to just use the url param key as the label name?

slim-bean: oh I see, I think the intent is to use a relabel rule to turn these into actual labels sent to Loki?

That seems reasonable to me 👍

slim-bean (approved):
LGTM, thank you!!

PR #7622, opened 2022-11-07
Docs: Add missing changelog entry for stream sharding

DylanGuedes:
Add missing changelog entry for the new stream sharding feature.
This will be better detailed under the release notes doc.


trevorwhitney (approved):
Thanks!

PR #7631, opened 2022-11-08
Swap direction in "The label identifier is always on the right side of the operation." sentence

osg-grafana:


dannykopping:
@osg-grafana please mark `loki-team` as the reviewer in future; PR reviews are a shared responsibility between the team

PR #7635, opened 2022-11-08
Refactor nix to make it easier to import into other projects

trevorwhitney:

This makes it easier for projects that depend on Loki (like enterprise logs) to more easily import the Loki nix flake in order to use the overlays and configurations it contains.

It also removes the `build-vars.nix` because it was super annoying without really getting us much.


trevorwhitney:
@simonswine @owen-d are probably best to review this one. thanks!

jdbaldry (approved):
LGTM. 

TIL: about `self.rev`. Do you know of any documentation I can read to understand this better?

I'm not sure if my review is enough for merging but I can at least it confirm that I was able to hop into the shell environment and build Loki with Nix. It's the first time in a while I've built Loki and it's super cool to see us use Nix in some our projects!

trevorwhitney:
@jdbaldry `self.rev` is set to the commit sha of `self` (since that is treated as a git repo), but only if the git tree is clean. Otherwise it's not set.

This is not well documented. I think I pieced this knowledge together from many searches across discourse.nixos.org, like this [comment](https://discourse.nixos.org/t/flakes-accessing-selfs-revision/11237/4) or [this post](https://discourse.nixos.org/t/flakes-embedding-git-commit-into-the-binary/17784)

PR #7638, opened 2022-11-08
[helm] Ignore single binary headless service from service-monitor

trevorwhitney:

Ignore the headless service from the Prometheus service-monitor as it was resulting in duplicate metrics. Replaces https://github.com/grafana/loki/pull/7049


MichelHollands (approved):
LGTM

PR #7650, opened 2022-11-09
Add 3rd target (`backend`) to SSD/Scalable mode

trevorwhitney:

This adds a 3rd target to SSD/Scalable mode, as well as a config flag to run the legacy `read` mode from the original 2 target configuration in order to give people time to migrate before we remove this option in Loki 3.0 (hopefully).

The 3rd target has the two major advantages:
1. Allows the read path to be run as a deployment and thus auto-scaled using our existing auto-scaling logic for queriers
2. Creates consistency with Mimir since they went with a 3 target model for their SSD deployment.

I don't want to merge this until I've had the chance to also update the helm chart, use this version, and test everything in a real environment.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -235,6 +235,13 @@ func (c *Config) Validate() error {
 		return err
 	}
 
+	// Honor the legacy scalable deployment topology
+	if c.LegacyReadTarget {
+		if c.isModuleEnabled(Backend) {
+			return fmt.Errorf("invalid target, cannot run backend target with legacy read mode")
+		}
+	}
+
 	return nil
 }
 
@@ -638,12 +645,7 @@ func (t *Loki) setupModuleManager() error {
 		deps[QueryFrontend] = append(deps[QueryFrontend], QueryScheduler)
 	}
 
-	// Honor the legacy scalable deployment topology
 	if t.Cfg.LegacyReadTarget {
-		if t.Cfg.isModuleEnabled(Backend) {
-			return fmt.Errorf("cannot run backend target with legacy read mode")
-		}
-
 		deps[Read] = append(deps[Read], QueryScheduler, Ruler, Compactor, IndexGateway)
 	}
 
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -71,7 +71,6 @@ type Config struct {
 	UseBufferedLogger bool `yaml:"use_buffered_logger"`
 	UseSyncLogger     bool `yaml:"use_sync_logger"`
 
-	// TODO(trevorwhitney): remove this with Loki 3.0
 	LegacyReadTarget bool `yaml:"legacy_read_target,omitempty"`
 
 	Common              common.Config               `yaml:"common,omitempty"`
@@ -117,7 +116,9 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&c.UseBufferedLogger, "log.use-buffered", true, "Uses a line-buffered logger to improve performance.")
 	f.BoolVar(&c.UseSyncLogger, "log.use-sync", true, "Forces all lines logged to hold a mutex to serialize writes.")
 
-	f.BoolVar(&c.LegacyReadTarget, "legacy-read-mode", false, "Set to true to enable the legacy read mode. This will be removed in Loki 3.0.")
+	//TODO(trevorwhitney): flip this to false with Loki 3.0
+	f.BoolVar(&c.LegacyReadTarget, "legacy-read-mode", true, "Set to false to disable the legacy read mode and use new scalable mode with 3rd backend target. "+
+		"The default will be flipped to false in the next Loki release.")
 
 	c.registerServerFlagsWithChangedDefaultValues(f)
 	c.Common.RegisterFlags(f)
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -71,6 +71,9 @@ type Config struct {
 	UseBufferedLogger bool `yaml:"use_buffered_logger"`
 	UseSyncLogger     bool `yaml:"use_sync_logger"`
 
+	// TODO(trevorwhitney): remove this with Loki 3.0
+	LegacyReadTarget bool `yaml:"legacy_read_target,omitempty"`
+
 	Common              common.Config               `yaml:"common,omitempty"`
 	Server              server.Config               `yaml:"server,omitempty"`
 	InternalServer      internalserver.Config       `yaml:"internal_server,omitempty"`
@@ -113,6 +116,8 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 		"garbage collection. Larger ballasts result in fewer garbage collection passes, reducing compute overhead at the cost of memory usage.")
 	f.BoolVar(&c.UseBufferedLogger, "log.use-buffered", true, "Uses a line-buffered logger to improve performance.")
 	f.BoolVar(&c.UseSyncLogger, "log.use-sync", true, "Forces all lines logged to hold a mutex to serialize writes.")
+
+	f.BoolVar(&c.LegacyReadTarget, "legacy-read-mode", false, "Set to true to enable the legacy read mode. This will be removed in Loki 3.0.")
 
 	c.registerServerFlagsWithChangedDefaultValues(f)
 	c.Common.RegisterFlags(f)
@@ -586,6 +591,7 @@ func (t *Loki) setupModuleManager() error {
 	mm.RegisterModule(All, nil)
 	mm.RegisterModule(Read, nil)
 	mm.RegisterModule(Write, nil)
+	mm.RegisterModule(Backend, nil)
 
 	// Add dependencies
 	deps := map[string][]string{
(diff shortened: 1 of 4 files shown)


DylanGuedes (changes requested):
Could you also update the related docs? Relevant ones I can think of are: the configuration page and the architectures pages

DylanGuedes (commented):
not sure if that's something for you to fix on this PR or is out of the scope, but running loki on the read mode with the `/cmd/loki/loki-local-config.yaml` won't work anymore (it used to work though). That doesn't seem too relevant but I thought it was worth mentioning it as users will probably use `loki-local-config.yaml` to start testing things and their initial experience with SSD will be potentially affected by this not working

the error is due to the compactor-address not being set:
```
query filtering for deletes requires 'compactor_address' to be configured
error initialising module: cache-generation-loader
```

potential solutions are:
- adding the compactor address to the loki-local-config.yaml
- making the compactor address parameter optional somehow

pkg/loki/loki.go
```diff
@@ -111,6 +114,8 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&c.UseBufferedLogger, "log.use-buffered", true, "Uses a line-buffered logger to improve performance.")
 	f.BoolVar(&c.UseSyncLogger, "log.use-sync", true, "Forces all lines logged to hold a mutex to serialize writes.")
 
+	f.BoolVar(&c.LegacyReadTarget, "legacy-read-mode", false, "Set to true to enable the legacy read mode. This will be removed in Loki 3.0.")
```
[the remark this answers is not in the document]
trevorwhitney: I wasn't sure which direction to go here. On by default is less disruptive, but might create more pain when we want to remove it in 3.0? Or maybe we flip it in 3.0, and remove it later?

trevorwhitney:
@DylanGuedes I've always thought of `loki-local-config.yaml` to mainly be fore single binary, so I'm not too worried about breaking it. I don't know how we could default the address without knowing the environment they're running in. Making it optional might work but I think it's out of scope for this work. WDYT?

pkg/loki/loki.go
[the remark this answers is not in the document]
DylanGuedes: Yes, I vote for having it as truthy by default, otherwise it will break setups without users expecting it. 

Regarding flipping it in 3.0 and removing it later or removing it entirely in 3.0, both are ok in my opinion.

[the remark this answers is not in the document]
trevorwhitney: Cool, agreed :+1: flipped!

DylanGuedes:
> > @DylanGuedes I've always thought of `loki-local-config.yaml` to mainly be fore single binary, so I'm not too worried about breaking it. I don't know how we could default the address without knowing the environment they're running in. Making it optional might work but I think it's out of scope for this work. WDYT?
> 
> The legacy read targets work with that default configuration even though they run a compactor, so although what you said is true, there's a way to overcome it (I don't know how though joy ). That said, do you mind looking at how it was done? That seems unfair from my side since I don't know it myself but I'm advocating in favor of it because, if the amount of work is low, I think your work will significantly benefit from it.
> 
> Out of curiosity, there's any configuration file in this repo that you normally use to test SSD stuff? In case there's none and you always write a new configuration file from scratch, besides we adapting the loki-local-config.yaml another option here would be for you to copy a configuration file that you normally use to the repo. I used to use the loki-local-config.yaml to test SSD stuff btw

shiit, I thought that the deletion mode was a general config that we could check if it was disabled or not, such that we could add the following check:

```diff
func (t *Loki) compactorAddress() (string, error) {
	legacyReadMode := t.Cfg.LegacyReadTarget && t.Cfg.isModuleEnabled(Read)
	if t.Cfg.isModuleEnabled(All) || legacyReadMode || t.Cfg.isModuleEnabled(Backend) {
		// In single binary or read modes, this module depends on Server
		return fmt.Sprintf("http://127.0.0.1:%d", t.Cfg.Server.HTTPListenPort), nil
	}

+       if t.Cfg.isModuleEnabled(Read) && t.Cfg.Compactor.CompactorMode == Disabled {
+               return fmt.Sprintf("http://127.0.0.1:%d", t.Cfg.Server.HTTPListenPort), nil
+	}

	return t.Cfg.Common.CompactorAddress, nil
}
```
However... deletion mode is per-tenant now :cry: 
You can consider that thread resolved then, I thought it would be easier to work around this.

trevorwhitney:
> Out of curiosity, there's any configuration file in this repo that you normally use to test SSD stuff? 

Yes, I use the `enterprise-logs` [k3d config](https://github.com/trevorwhitney/loki/blob/add-backend-target-to-ssd/tools/dev/k3d/environments/enterprise-logs/main.jsonnet) which runs the helm chart in SSD mode. We should probably add a configuration for plain `loki` in this mode.

pkg/loki/loki.go
```diff
@@ -111,6 +113,10 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&c.UseBufferedLogger, "log.use-buffered", true, "Uses a line-buffered logger to improve performance.")
 	f.BoolVar(&c.UseSyncLogger, "log.use-sync", true, "Forces all lines logged to hold a mutex to serialize writes.")
 
+	//TODO(trevorwhitney): flip this to false with Loki 3.0
+	f.BoolVar(&c.LegacyReadTarget, "legacy-read-mode", true, "Set to false to disable the legacy read mode and use new scalable mode with 3rd backend target. "+
```
jeschkies, line 117: Is there some validation for `legacy-read-mode=true` and `target=backend`?

pkg/loki/loki.go
```diff
@@ -546,28 +553,38 @@ func (t *Loki) setupModuleManager() error {
...
 		deps[QueryFrontend] = append(deps[QueryFrontend], QueryScheduler)
 	}
 
+	// Honor the legacy scalable deployment topology
+	if t.Cfg.LegacyReadTarget {
+		if t.Cfg.isModuleEnabled(Backend) {
+			return fmt.Errorf("cannot run backend target with legacy read mode")
+		}
+
+		deps[Read] = append(deps[Read], QueryScheduler, Ruler, Compactor, IndexGateway)
```
jeschkies, line 585: What about 

```go
if t.Cfg.isModuleEnabled(Read) {
    deps[QueryFrontend] = append(deps[QueryFrontend], QueryScheduler)
}
```

pkg/loki/loki.go
```diff
@@ -546,28 +553,38 @@ func (t *Loki) setupModuleManager() error {
...
 	// first to initialize the ring that will also be used by the query frontend
-	if (t.Cfg.isModuleEnabled(QueryFrontend) && t.Cfg.isModuleEnabled(QueryScheduler)) || t.Cfg.isModuleEnabled(Read) || t.Cfg.isModuleEnabled(All) {
+	if (t.Cfg.isModuleEnabled(QueryFrontend) && t.Cfg.isModuleEnabled(QueryScheduler)) || t.Cfg.isModuleEnabled(All) {
 		deps[QueryFrontend] = append(deps[QueryFrontend], QueryScheduler)
 	}
 
+	// Honor the legacy scalable deployment topology
+	if t.Cfg.LegacyReadTarget {
+		if t.Cfg.isModuleEnabled(Backend) {
+			return fmt.Errorf("cannot run backend target with legacy read mode")
```
jeschkies, line 582: I think this should happen when the config is validated.

DylanGuedes (approved):
approving to remove the request changes tag (but make sure to read through Karsten comments)

pkg/loki/loki.go
```diff
@@ -546,28 +553,38 @@ func (t *Loki) setupModuleManager() error {
 		IngesterQuerier:          {Ring},
 		IndexGatewayRing:         {RuntimeConfig, Server, MemberlistKV},
 		All:                      {QueryScheduler, QueryFrontend, Querier, Ingester, Distributor, Ruler, Compactor},
-		Read:                     {QueryScheduler, QueryFrontend, Querier, Ruler, Compactor, IndexGateway},
+		Read:                     {QueryFrontend, Querier},
 		Write:                    {Ingester, Distributor},
+		Backend:                  {QueryScheduler, Ruler, Compactor, IndexGateway},
```
dannykopping, line 558: Curious why the ruler is in this target... can you expand on that pls?

pkg/loki/loki.go
trevorwhitney: yes: https://github.com/trevorwhitney/loki/blob/5135a4325d78cd2e2e29f503c1ca9b54ad9a51d5/pkg/loki/loki.go#L644

pkg/loki/loki.go
trevorwhitney: Two reasons, 1) because that's what mimir did, 2) because we want to be able to run the Read target as a deployment that can easily be killied, OOMed, scaled up or down. How does the Ruler respond to unexpected interruptions? I do see we run it as a deployment, how problematic will it be if a ruler gets interrupted because a Querier is OOM killed?

dannykopping: Gotcha, thanks for the context.
The ruler can respond just fine to interruptions, but it's better that they're a statefulset so that rules can be ~consistently sharded

pkg/loki/modules_test.go
```diff
@@ -161,67 +162,106 @@ func TestMultiKVSetup(t *testing.T) {
...
+		target      string
+	}
 
-	t.Run("When IndexGateway is running independent of Read target", func(t *testing.T) {
-		t.Run("IndexGateway respects configured simple mode", func(t *testing.T) {
-			cfg := minimalWorkingConfig(t, dir, IndexGatewayRing)
-			cfg.IndexGateway.Mode = indexgateway.SimpleMode
+	for _, tc := range []ringModeTestCase{
+		{
+			name:   "leagcy read",
```
earimont-ib, line 176: typo leagcy

PR #7651, opened 2022-11-09
adds Dropbox to the adopters list

afayngelerindbx:
Dropbox recently deployed Loki. This PR adds us to the adopters list


dannykopping (approved):
:tada:

PR #7652, opened 2022-11-10
Update Ingester Rate Calculations

MasslessParticle:
- Use weighted moving average for rates
- Keep Streams around until they are removed

Previously, the rate calculator only kept the most recent second. This led to issues where bursty streams that didn't get pushes every second resulted in undersharding those streams

### DIFF
--- a/pkg/ingester/stream_rate_calculator.go
+++ b/pkg/ingester/stream_rate_calculator.go
@@ -14,6 +14,11 @@ const (
 
 	// The intent is for a per-second rate so this is hard coded
 	updateInterval = time.Second
+
+	// The factor used to weight the moving average. Must be in the range [0, 1.0].
+	// A larger factor weights recent samples more heavily while a smaller
+	// factor weights historic samples more heavily.
+	smoothingFactor = .7
 )
 
 // stripeLock is taken from ruler/storage/wal/series.go
@@ -35,7 +40,7 @@ type StreamRateCalculator struct {
 
 func NewStreamRateCalculator() *StreamRateCalculator {
 	calc := &StreamRateCalculator{
-		size:     defaultStripeSize,
+		size: defaultStripeSize,
 		// Lookup pattern: tenant -> fingerprint -> rate
 		samples:  make([]map[string]map[uint64]logproto.StreamRate, defaultStripeSize),
 		locks:    make([]stripeLock, defaultStripeSize),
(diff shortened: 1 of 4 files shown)


pkg/ingester/stream_rate_calculator.go
```diff
@@ -66,31 +71,79 @@ func (c *StreamRateCalculator) updateLoop() {
 }
 
 func (c *StreamRateCalculator) updateRates() {
-	rates := make([]logproto.StreamRate, 0, c.size)
+	rates := c.rates()
+	samples := c.currentSamplesPerTenant()
+	samples = updateSamples(rates, samples)
+
+	updatedRates := make([]logproto.StreamRate, 0, c.size)
```
DylanGuedes, line 78: not sure if I'm missing something, but why do you need this `updatedReads` instead of just using the evaluated `samples` from line 76?

pkg/ingester/stream_rate_calculator.go
```diff
@@ -66,31 +71,79 @@ func (c *StreamRateCalculator) updateLoop() {
...
 
 	for i := 0; i < c.size; i++ {
 		c.locks[i].Lock()
 
-		tenantRates := c.samples[i]
-		for _, tenant := range tenantRates {
-			for _, streamRate := range tenant {
-				rates = append(rates, logproto.StreamRate{
+		for tenantID, tenant := range c.samples[i] {
+			existingRates := ratesForTenant(rates, tenantID)
```
DylanGuedes, line 105: sorry, I found this part confusing; do you mind clarifying why do we have to check the rates variable here? I think the confusion is because:
- `rates` is a new variable populated from the `existingRates`
- the `existingRates` variable is populated here from the `rates`

which at the end looks like a cyclic dependency :thinking:

DylanGuedes (approved):
LGTM! Just a few questions/suggestions

pkg/ingester/stream_rate_calculator.go
MasslessParticle: This is because the data structure is now a `map[string]map[uint64]logproto.StreamRate` the 0 value of a map is `nil` so we have to do all this to make sure the inner map is initialized: https://github.com/grafana/loki/blob/82d838a9cc1c485084fe51e844d8bde0ed129bcd/pkg/ingester/stream_rate_calculator.go#L205

pkg/ingester/stream_rate_calculator.go
```diff
@@ -117,13 +170,45 @@ func (c *StreamRateCalculator) Record(tenant string, streamHash, streamHashNoSha
 	c.samples[i][tenant] = tenantMap
 }
 
-func (c *StreamRateCalculator) getTenant(idx uint64, tenant string) map[uint64]logproto.StreamRate {
+func (c *StreamRateCalculator) Remove(tenant string, streamHash uint64) {
+	i := streamHash & uint64(c.size-1)
+
+	c.locks[i].Lock()
```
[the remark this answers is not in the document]
MasslessParticle: This is a fair question. I think the worst thing that could happen is returning an out of date rate map. In that case, the stream will be omitted in the next call. The biggest issue is that `samples` and `allRates` shouldn't grow forever

DylanGuedes (approved):
lgtm!

PR #7653, opened 2022-11-10
Promtail docs: Fix docker stage code example

MatthiasKunnen:
The current, 2.6.x, documentation on [configuring promtail _pipeline_stages_](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#pipeline_stages) contains a [docker stage](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#docker) which wraps the json, labels, timestamp, and output stage. The unwrapped configuration is shown as example. However, this example is outdated as can be seen by the lack of `expressions`. I have stumbled across this as I tried the example in my own config as a starter to modify to my own needs. It took me some time to realize what `failed to make file target manager: invalid json stage config: JMES expression is required` meant. This PR updates the example to match the current code at <https://github.com/grafana/loki/blob/a6df09bc97f64584c2efbeb2550a2d59526f39c0/clients/pkg/logentry/stages/extensions.go#L16-L42>

Going back in history points to #576 as the origin of the divergence.

I found another usage of outdated configuration in a test file but was unsure whether it needed updating:
https://github.com/grafana/loki/blob/a6df09bc97f64584c2efbeb2550a2d59526f39c0/clients/pkg/promtail/scrapeconfig/scrapeconfig_test.go#L32-L40

**Checklist**
- [x] Reviewed the `CONTRIBUTING.md` guide
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated
- [x] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #7654, opened 2022-11-10
Fix json stage fields in example

MichelHollands:
Fix the json section of a scrape config used in a unit test. The unit test doesn't fail as the unmarshaling is very generic. The json stage is unit tested in a different place.
This was mentioned as a comment on another PR: https://github.com/grafana/loki/pull/7653.

### DIFF
--- a/clients/pkg/promtail/scrapeconfig/scrapeconfig_test.go
+++ b/clients/pkg/promtail/scrapeconfig/scrapeconfig_test.go
@@ -30,14 +30,15 @@ pipeline_stages:
   - regex:
       expr: "./*"
   - json:
-      timestamp:
-        source: time
-        format: RFC3339
-      labels:
-        stream:
-          source: json_key_name.json_sub_key_name
-      output:
-        source: log
+      expressions:
+        timestamp:
+          source: time
+          format: RFC3339
+        labels:
+          stream:
+            source: json_key_name.json_sub_key_name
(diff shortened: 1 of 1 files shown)


jeschkies (approved):

clients/pkg/promtail/scrapeconfig/scrapeconfig_test.go
```diff
@@ -30,14 +30,15 @@ pipeline_stages:
...
-        source: log
+      expressions:
+        timestamp:
+          source: time
+          format: RFC3339
+        labels:
+          stream:
+            source: json_key_name.json_sub_key_name
+        output:
+          source: log
```
MatthiasKunnen, line 41: This might be my inexperience with Promtail but if this is supposed to following the [pipeline stage config](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#pipeline_stages), shouldn't it be more like this?
```suggestion
      expressions:
        timestamp: time
        stream: stream
        output: log
  - labels:
      stream:
  - timestamp:
      source: time
      format: RFC3339
  - output:
      source: output
```

The only documentation I can find regarding the original structure is in this historical design document: https://github.com/grafana/loki/blob/3039ca06f4d0fe964943bebcc29d9a4991143635/docs/sources/design-documents/labels.md?plain=1#L97-L104

PR #7660, opened 2022-11-10
Add diff-config command to flake.nix

trevorwhitney:

Adds @slim-bean's config diffing command to `flake.nix` so it can diff the current checked out version against the last released version without having to create docker images.


trevorwhitney:
closing due to lack of interest

PR #7384, opened 2022-10-10
operator: Add provision for the user to set proxy env variables in the loki pods

shwetaap:
Signed-off-by: Shweta Padubidri <spadubid@redhat.com>

Add provision for the user to set proxy env variables in the loki pods

**Which issue(s) this PR fixes**:
Fixes Log2860

**Special notes for your reviewer**:

**Checklist**
- [ ] Reviewed the `CONTRIBUTING.md` guide
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #7682, opened 2022-11-13
operator: Refactor proxy env variables

periklis:
Follow-up of #7384 for core replacement and using `mergo.Merge`

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Main
 
+- [7682](https://github.com/grafana/loki/pull/7682) **periklis**: Refactor cluster proxy to use configv1.Proxy on OpenShift
 - [7617](https://github.com/grafana/loki/pull/7617) **Red-GV**: Modify ingestionRate for respective shirt size
 - [7592](https://github.com/grafana/loki/pull/7592) **aminesnow**: Update API docs generation using gen-crd-api-reference-docs
 - [7448](https://github.com/grafana/loki/pull/7448) **periklis**: Add TLS support for compactor delete client
--- a/operator/apis/loki/v1/lokistack_types.go
+++ b/operator/apis/loki/v1/lokistack_types.go
@@ -343,13 +343,6 @@ type ClusterProxy struct {
 	// +kubebuilder:validation:optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="NoProxy"
 	NoProxy string `json:"noProxy,omitempty"`
-	// ReadVarsFromEnv defines a flag to use Operator-lib provides a helper function
-	//
-	// +optional
-	// +kubebuilder:validation:optional
-	// +kubebuilder:default:=false
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch",displayName="ReadVarsFromEnv"
-	ReadVarsFromEnv bool `json:"readVarsFromEnv,omitempty"`
 }
 
 // ObjectStorageTLSSpec is the TLS configuration for reaching the object storage endpoint.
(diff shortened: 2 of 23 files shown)


operator/internal/manifests/env_vars.go
```diff
@@ -0,0 +1,85 @@
...
+				Name:  httpProxyKey,
+				Value: spec.HTTPProxy,
+			},
+			{
+				Name:  strings.ToLower(httpProxyKey),
+				Value: spec.HTTPProxy,
+			},
+			{
+				Name:  httpsProxyKey,
+				Value: spec.HTTPSProxy,
```
xperimental, line 40: I think you need a check if the value of a setting is empty and only add the env-var in that case, so that it is possible to set `httpProxy` without also setting `httpsProxy` and `noProxy`. Currently only setting `httpProxy` for example looks like this:

```
        env:
        - name: HTTP_PROXY
          value: test
        - name: http_proxy
          value: test
        - name: HTTPS_PROXY
        - name: https_proxy
        - name: NO_PROXY
        - name: no_proxy
```

operator/internal/manifests/env_vars.go
```diff
@@ -0,0 +1,85 @@
...
+	noProxyKey    = "NO_PROXY"
+)
+
+func configureProxyEnv(spec *lokiv1.ClusterProxy, pod *corev1.PodSpec) error {
+	if spec == nil {
+		return nil
+	}
+
+	for _, envVar := range proxy.ProxyEnvNames {
+		resetProxyVar(pod, envVar)
```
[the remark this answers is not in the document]
periklis: For now this is the only place we set environment variables. We can't use `WithOvveride` here because it would replace the entire env vars slice and we may introduce somewhere else env vars. This would make debugging hard for no reason. 

In general I suggested to put here a reset because double entries spam the kube audit log. I imagine you can introduce env vars on the pod controller level. So this is just for safety.

operator/internal/manifests/env_vars.go
```diff
@@ -0,0 +1,111 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	httpProxyKey  = "HTTP_PROXY"
+	httpsProxyKey = "HTTPS_PROXY"
+	noProxyKey    = "NO_PROXY"
+)
+
+var ProxyEnvNames = []string{
```
xperimental, line 16: Can be private now.

operator/internal/manifests/env_vars.go
```diff
@@ -0,0 +1,111 @@
+package manifests
```
xperimental, line 1: Should we rename this file, because it's not "generic environment variables" but only "proxy variables". Maybe `proxy_env.go`?

operator/internal/manifests/env_vars.go
```diff
@@ -0,0 +1,111 @@
...
+		if found {
+			podSpec.Containers[i].Env = append(podSpec.Containers[i].Env[:index], podSpec.Containers[i].Env[index+1:]...)
+		}
+	}
+}
+
+// getEnvVar matches the given name with the envvar name
+func getEnvVar(name string, envVars []corev1.EnvVar) (bool, int) {
+	for i, env := range envVars {
+		if env.Name == name || env.Name == strings.ToLower(name) {
```
xperimental, line 61: The `ToLower` is unnecessary, because the list of variables already includes both permutations.

operator/internal/manifests/env_vars_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"strings"
+	"testing"
+
+	lokiv1 "github.com/grafana/loki/operator/apis/loki/v1"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestContainerEnvVars_ReadVarsFromCustomResource(t *testing.T) {
```
xperimental, line 13: Should we add a test for "the container already has one of the environment variables but we are overriding it"?

xperimental (approved):
lgtm and seems to work fine in tests when adding the fix for the global resource watches. Have not tested with a functional cluster-proxy yet though, will do that later today, but the code looks fine.

PR #7695, opened 2022-11-15
[helm] Fix read-only /tmp in single-binary mode

MiggL:
When running in single-binary mode, these warnings may appear:
`level=warn ts=2022-11-15T08:01:39.777179378Z caller=marker.go:214 msg="failed to process marks" path=/var/loki/compactor/retention/markers/1668476560013344947 err="open /tmp/marker-view-912253853: read-only file system"`

This was previously fixed in the old chart: grafana/helm-charts#967

Fixes grafana/helm-charts#609, again.

**Checklist**

### DIFF
--- a/production/helm/loki/Chart.yaml
+++ b/production/helm/loki/Chart.yaml
@@ -4,7 +4,7 @@ name: loki
 description: Helm chart for Grafana Loki in simple, scalable mode
 type: application
 appVersion: 2.6.1
-version: 3.4.1
+version: 3.4.2
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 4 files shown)


stefanandres:
Yes, please merge this. This breaks all retentions for existing installations that upgraded. I had to add the workaround like 10months ago back to our installations:

```
  # https://github.com/grafana/loki/pull/7695
  extraVolumeMounts:
    - name: tmp
      mountPath: /tmp
  extraVolumes:
    - name: tmp
      emptyDir: {}
```

production/helm/loki/Chart.yaml
```diff
@@ -4,7 +4,7 @@ name: loki
 description: Helm chart for Grafana Loki in simple, scalable mode
 type: application
 appVersion: 2.6.1
-version: 3.3.4
+version: 3.3.5
```
jeschkies, line 7: I know this is a little annoying but main is at `3.4.0` now.

jeschkies (approved):

production/helm/loki/Chart.yaml
MiggL: Fixed. Also updated `production/helm/loki/CHANGELOG.md`

PR #7703, opened 2022-11-15
Loki: Use default limits when cant get the tenantId

### LINKED ISSUE #7696: After upgrading to 2.7 multi tenant queries fails
**Describe the bug**
After upgrading to 2.7 multi tenant queries fails.

It seems we encounter this bug because of this function only checks if query has single tenant id for per tenant override for timeout parameter. 
https://github.com/grafana/loki/blob/v2.7.0/pkg/querier/http.go#L486

**To Reproduce**
Query loki querier with multiple X-Scope-OrgID header

exp:  `curl --location -g --request GET 'localhost:3100/loki/api/v1/query_range?query={namespace=~"loki",_cluster="cluster"}' \
--header 'X-Scope-OrgID: tenant1|tenant2'`
returns: `multiple org IDs present`

**Expected behavior**
It should return query range response

mbdeveci:
This PR is a quick fix to not fail and use default limits if can not get the single tenantId, which makes multi-tenant queries work again.
Fixes #7696 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/querier/http.go
+++ b/pkg/querier/http.go
@@ -487,14 +487,7 @@ func WrapQuerySpanAndTimeout(call string, q *QuerierAPI) middleware.Interface {
 	return middleware.Func(func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 			log, ctx := spanlogger.New(req.Context(), call)
-			userID, err := tenant.TenantID(ctx)
-			if err != nil {
-				level.Error(log).Log("msg", "couldn't fetch tenantID", "err", err)
-				serverutil.WriteError(httpgrpc.Errorf(http.StatusBadRequest, err.Error()), w)
-				return
-			}
-
-			// Enforce the query timeout while querying backends
+			userID, _ := tenant.TenantID(ctx)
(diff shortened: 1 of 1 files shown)


DylanGuedes:
Hey thank you for your contribution. Since the bug was introduced by me I paid special attention to it and because of that, I noticed this solution isn't enough. Reasons:
- In some places we rely on the engine timeout and this PR doesn't fix for it
- You're just picking whatever returns from tenant.TenantID. I think a better approach is to pick the longest timeout across all tenants given.

That said, I'm writing another solution on https://github.com/grafana/loki/pull/7708. Do you mind if we close this in favor of it?

mbdeveci:
Hey, your solution is the right approach. I just wanted to provide a quick solution to make it work as it was. I'm closing this issue

DylanGuedes:
> Hey, your solution is the right approach. I just wanted to provide a quick solution to make it work as it was. I'm closing this issue

Thanks! If you don't mind, I added you to the signed-off-by of that PR since your report was extremely useful in finding the issue and getting notified that something was wrong.

PR #7708, opened 2022-11-16
Loki: Fix multitenant querying

DylanGuedes:
We recently broke multitenant querying because of recent changes to how timeouts working across Loki. This PR fixes this by:
- Adapt the timeout wrapper to work with multitenant queries. It will take the shortest timeout across all given tenants
- Adapt the query engine timeout assigning to work with multitenant queries. It will take the shortest timeout between all the tenants
- Adapt query sharding to use smallest max query parallelism across given tenants
- Add a functional test to ensure multitenant is behaving as expected

Signed-off-by: DylanGuedes <djmgguedes@gmail.com>
Signed-off-by: Mehmet Burak Devecí <mhmtbrkdvc@gmail.com>

https://github.com/grafana/loki/issues/7696

The regression was probably introduced by https://github.com/grafana/loki/pull/7555

### DIFF
--- a/integration/multi_tenant_queries_test.go
+++ b/integration/multi_tenant_queries_test.go
@@ -1,0 +1,66 @@
+package integration
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/loki/integration/client"
+	"github.com/grafana/loki/integration/cluster"
+)
+
+func TestMultiTenantQuery(t *testing.T) {
+	clu := cluster.New()
+	defer func() {
+		assert.NoError(t, clu.Cleanup())
+	}()
+
+	var (
+		tAll = clu.AddComponent(
+			"all",
+			"-target=all",
+		)
+	)
+
+	require.NoError(t, clu.Run())
+
+	cliTenant1 := client.New("org1", "", tAll.HTTPURL())
+	cliTenant2 := client.New("org2", "", tAll.HTTPURL())
+	cliMultitenant := client.New("org1|org2", "", tAll.HTTPURL())
+
+	// ingest log lines for tenant 1 and tenant 2.
+	require.NoError(t, cliTenant1.PushLogLineWithTimestamp("lineA", cliTenant1.Now.Add(-45*time.Minute), map[string]string{"job": "fake1"}))
+	require.NoError(t, cliTenant2.PushLogLineWithTimestamp("lineB", cliTenant2.Now.Add(-45*time.Minute), map[string]string{"job": "fake2"}))
+
+	// check that tenant1 only have access to log line A.
+	matchLines(t, cliTenant1, `{job="fake2"}`, []string{})
+	matchLines(t, cliTenant1, `{job=~"fake.*"}`, []string{"lineA"})
+	matchLines(t, cliTenant1, `{job="fake1"}`, []string{"lineA"})
+
+	// check that tenant2 only have access to log line B.
+	matchLines(t, cliTenant2, `{job="fake1"}`, []string{})
+	matchLines(t, cliTenant2, `{job=~"fake.*"}`, []string{"lineB"})
+	matchLines(t, cliTenant2, `{job="fake2"}`, []string{"lineB"})
+
+	// check that multitenant has access to all log lines on same query.
+	matchLines(t, cliMultitenant, `{job=~"fake.*"}`, []string{"lineA", "lineB"})
+	matchLines(t, cliMultitenant, `{job="fake1"}`, []string{"lineA"})
+	matchLines(t, cliMultitenant, `{job="fake2"}`, []string{"lineB"})
+	matchLines(t, cliMultitenant, `{job="fake3"}`, []string{})
+}
+
+func matchLines(t *testing.T, client *client.Client, labels string, expectedLines []string) {
+	resp, err := client.RunRangeQuery(context.Background(), labels)
+	require.NoError(t, err)
+
+	var lines []string
+	for _, stream := range resp.Data.Stream {
+		for _, val := range stream.Values {
+			lines = append(lines, val[1])
+		}
+	}
+	require.ElementsMatch(t, expectedLines, lines)
+}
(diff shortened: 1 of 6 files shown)


chaudum:
@DylanGuedes can you check with https://github.com/grafana/loki/pull/7703 so you don't do things twice? :)

DylanGuedes:
> @DylanGuedes can you check with #7703 so you don't do things twice? :)

Good call. I left a comment on his PR recommending that we proceed with this PR instead, since I'm adding a test that checks that this is really working and I noticed we also have to adapt things in other places.

JordanRushing:
Does it make sense to opt for the lower of multiple limits rather than the higher? The former doesn't introduce unanticipated increase in resources whereas the latter does if I understand correctly.

pkg/logql/engine.go
```diff
@@ -240,8 +240,15 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
 }
 
 func (q *query) Eval(ctx context.Context) (promql_parser.Value, error) {
-	userID, _ := tenant.TenantID(ctx)
-	queryTimeout := q.limits.QueryTimeout(userID)
+	// set as timeout the longest timeout across all tenants.
```
jeschkies, line 243: So, we usually take the most restrictive limit. That'd be the smallest timeout. What is the reason to take the longest here?

pkg/querier/queryrange/querysharding.go
```diff
@@ -99,17 +99,25 @@ func (ast *astMapperware) Do(ctx context.Context, r queryrangebase.Request) (que
 		return ast.next.Do(ctx, r)
 	}
 
-	userID, err := tenant.TenantID(ctx)
+	// use biggest max query parallelism across all given tenants
```
jeschkies, line 102: Same here. We should take the most restrictive.

DylanGuedes:
> Does it make sense to opt for the lower of multiple limits rather than the higher? The former doesn't introduce unanticipated increases in resource consumption whereas the latter does if I understand correctly.

Good question. I feel like we can find arguments for both, but personally, I think going for the higher values is better because it avoids that scenario where a smaller tenant makes queries not operational for bigger tenants if they have limits that are smaller enough.

jeschkies (changes requested):
This is a really great find. Thanks for these awesome tests! However, we should always use the most restrictive limit. See https://github.com/grafana/loki/pull/5626 for the original introduction.

pkg/querier/http.go
```diff
@@ -487,22 +487,30 @@ func WrapQuerySpanAndTimeout(call string, q *QuerierAPI) middleware.Interface {
...
+				// Enforce the query timeout while querying backends
+				queryTimeout := q.limits.QueryTimeout(tenantID)
+				// TODO: remove this clause once we remove the deprecated query-timeout flag.
+				if q.cfg.QueryTimeout != 0 { // querier YAML configuration.
+					level.Warn(log).Log("msg", "deprecated querier:query_timeout YAML configuration identified. Please migrate to limits:query_timeout instead.", "call", "WrapQuerySpanAndTimeout", "org_id", tenantID)
+					queryTimeout = q.cfg.QueryTimeout
+				}
+				if longestTimeout < queryTimeout {
+					longestTimeout = queryTimeout
+
```
JordanRushing, line 509: small nit: this might be extra

integration/multi_tenant_queries_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/loki/integration/client"
+	"github.com/grafana/loki/integration/cluster"
+)
+
+func TestMultiTenantQuery(t *testing.T) {
```
JordanRushing, line 15: Nice test!

JordanRushing (approved):
Approved pending additional dialogue about picking the higher vs. lower limit for a multi-tenant query.

pkg/logql/engine.go
DylanGuedes: To avoid a smaller tenant messing up with other tenants. Full answer on https://github.com/grafana/loki/pull/7708#issuecomment-1322287483.

DylanGuedes:
> This is a really great find. Thanks for these awesome tests! However, we should always use the most restrictive limit. See #5626 for the original introduction.

Thanks, I hope those will help! Regarding the limits: I'm not strongly opinionated for proceeding with the bigger values, but personally, I think they work better here. That said, I'll wait a little more for others to share opinions before reverting the change etc.

pkg/querier/http.go
```diff
@@ -487,22 +487,30 @@ func WrapQuerySpanAndTimeout(call string, q *QuerierAPI) middleware.Interface {
 	return middleware.Func(func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 			log, ctx := spanlogger.New(req.Context(), call)
-			userID, err := tenant.TenantID(ctx)
+			tenantIDS, err := tenant.TenantIDs(ctx)
```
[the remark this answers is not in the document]
DylanGuedes: addressed by https://github.com/grafana/loki/pull/7708/commits/c64a892b3b4771babb3a73fa044991ce40a5cb6d.

DylanGuedes:
@jeschkies @JordanRushing I pushed a commit using the smallest limits instead, WDYT?

pkg/querier/http.go
jeschkies: Sorry for testing your patience but you should use [SmallestPositiveNonZeroDurationPerTenant](https://github.com/grafana/loki/blob/main/pkg/util/validation/limits.go#L683)

DylanGuedes: No worries and good call. I pushed https://github.com/grafana/loki/pull/7708/commits/cd034e33b3335b4df2004805a89031a10acc7145 to address this, WDYT?

DylanGuedes:
> See my comment. I am happy to take over the PR if you give me right access to your repository.

I pushed a commit addressing it, WDYT?

jeschkies (approved):
🎉

PR #7715, opened 2022-11-17
[CI/CD] fixed step `deploy`

vlad-diachenko:
migrated the `deploy` step configuration to use the value from `.tag` file instead of using `DRONE_SOURCE_BRANCH`  env variable

**Why this change is needed?**

During the initial implementation, I assumed that we always have env variable `DRONE_SOURCE_BRANCH` defined ... However, it's undefined on tag event .. that is why during `Loki 2.7.0` release, the step `deploy` ([drone_build](https://drone.grafana.net/grafana/loki/17477/20/2)) failed  and used value `next` for the release name: 
```
+ export RELEASE_NAME=$([[ $DRONE_SOURCE_BRANCH =~ $RELEASE_BRANCH_REGEXP ]] && echo $DRONE_SOURCE_BRANCH | grep -oE "([0-9\.x]+)" | sed "s/\./-/g" || echo "next")
sh: ^release-([0-9\.x]+)$: unknown operand
+ echo $RELEASE_NAME
next
```

That is why I changed the implementation to use the value from file `.tag`. Because we always have the value there (on push to main and on tag). So, if the value in `.tag` file matches the regex `^([0-9]+\.[0-9]+\.[0-9]+)$` then it understands that it's a build is related to release, then it creates a release name based on this value. Otherwise, it uses `next` as a release name. Example:
* for value from `.tag`: `23.12.34`, it defines release name `23-12-x`. ([drone build](https://drone.grafana.net/grafana/loki/17560/1/2))
* for value from `.tag`: `retest-2.7.x-a698052-amd64`,  it defines release name `next` ([drone build](https://drone.grafana.net/grafana/loki/17561/1/2))


.drone/drone.yml
```diff
@@ -1203,19 +1203,19 @@ steps:
...
-    echo $DRONE_SOURCE_BRANCH | grep -oE "([0-9\.x]+)" | sed "s/\./-/g" || echo "next")
+  - export RELEASE_TAG=$(cat .tag)
+  - export RELEASE_NAME=$([[ $RELEASE_TAG =~ $RELEASE_TAG_REGEXP ]] && echo $RELEASE_TAG
+    | grep -oE "([0-9]+\.[0-9]+)" | sed "s/\./-/g" | sed "s/$/-x/" || echo "next")
   - echo $RELEASE_NAME
-  - export RELEASE_TAG=$(cat .tag) && echo $RELEASE_TAG
   - echo $PLUGIN_CONFIG_TEMPLATE > updater-config.json
   - sed -i "s/\"{{release}}\"/\"$RELEASE_NAME\"/g" updater-config.json
   - sed -i "s/{{version}}/$RELEASE_TAG/g" updater-config.json
+  - cat updater-config.json
```
kavirajk, line 1214: is this intentional? (looks like added for debugging). 

Can we remove it to avoid noisy json?

.drone/drone.jsonnet
```diff
@@ -629,21 +629,21 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
...
         commands: [
           'apk add --no-cache bash git',
           'git fetch origin --tags',
-          'echo $(./tools/image-tag)',
           'echo $(./tools/image-tag) > .tag',
-          // if the branch name matches the pattern `release-D.D.x` then RELEASE_NAME="D-D-x", otherwise RELEASE_NAME="next"
-          'export RELEASE_NAME=$([[ $DRONE_SOURCE_BRANCH =~ $RELEASE_BRANCH_REGEXP ]] && echo $DRONE_SOURCE_BRANCH | grep -oE "([0-9\\.x]+)" | sed "s/\\./-/g" || echo "next")',
+          'export RELEASE_TAG=$(cat .tag)',
+          // if the tag matches the pattern `D.D.D` then RELEASE_NAME="D-D-x", otherwise RELEASE_NAME="next"
+          'export RELEASE_NAME=$([[ $RELEASE_TAG =~ $RELEASE_TAG_REGEXP ]] && echo $RELEASE_TAG | grep -oE "([0-9]+\\.[0-9]+)" | sed "s/\\./-/g" | sed "s/$/-x/" || echo "next")',
```
kavirajk, line 640: Can we give some name to `([0-9]+\\.[0-9]+)`? similar to `RELEASE_TAG_REGEXP`?

.drone/drone.yml
```diff
@@ -1203,19 +1203,19 @@ steps:
 - commands:
   - apk add --no-cache bash git
   - git fetch origin --tags
-  - echo $(./tools/image-tag)
   - echo $(./tools/image-tag) > .tag
-  - export RELEASE_NAME=$([[ $DRONE_SOURCE_BRANCH =~ $RELEASE_BRANCH_REGEXP ]] &&
-    echo $DRONE_SOURCE_BRANCH | grep -oE "([0-9\.x]+)" | sed "s/\./-/g" || echo "next")
+  - export RELEASE_TAG=$(cat .tag)
+  - export RELEASE_NAME=$([[ $RELEASE_TAG =~ $RELEASE_TAG_REGEXP ]] && echo $RELEASE_TAG
+    | grep -oE "([0-9]+\.[0-9]+)" | sed "s/\./-/g" | sed "s/$/-x/" || echo "next")
```
kavirajk, line 1209: Can we give some name to `([0-9]+\.[0-9]+`? similar to `RELEASE_TAG_REGEXP`?

Q: is it different than the one used above? (`([0-9]+\\.[0-9]+)`) only difference I see is `backslash` escaped in later (`\\.`). Is this intentional?

.drone/drone.yml
vlad-diachenko: removed

.drone/drone.yml
vlad-diachenko: in jsonnet we used `\\.` because we need to generate value `\.` in yaml file. if we do not escape `\` in jsonnet, it generates just `.` in yaml file )))

kavirajk (approved):
Thanks @vlad-diachenko. LGTM!

PR #7723, opened 2022-11-18
Improve the JSON parser performance.

cyriltovena:
Sorry I couldn't help myself and worked a bit on the JSON parser by using another library that allows way less allocations.

On top of this, the parser code is also much simpler to understand.

I realize that we could possibly also intern/cache (limited obviously) extracted JSON value per key. I would do it per key because otherwise, a high cardinality key can take over the intern/cache which will make it worthless. Happy to walk someone through the idea, I think this would be a big jump too.

Would be great to run and test this in our dev environment using some JSON queries before releasing it.

```
name                             old time/op    new time/op    delta
_Parser/json/no_labels_hints-10    2.53µs ± 0%    2.16µs ± 2%  -14.59%  (p=0.008 n=5+5)
_Parser/json/labels_hints-10       1.93µs ± 0%    1.47µs ± 0%  -23.96%  (p=0.008 n=5+5)

name                             old alloc/op   new alloc/op   delta
_Parser/json/no_labels_hints-10      656B ± 0%      280B ± 0%  -57.32%  (p=0.008 n=5+5)
_Parser/json/labels_hints-10         512B ± 0%      176B ± 0%  -65.62%  (p=0.008 n=5+5)

name                             old allocs/op  new allocs/op  delta
_Parser/json/no_labels_hints-10      46.0 ± 0%      18.0 ± 0%  -60.87%  (p=0.008 n=5+5)
_Parser/json/labels_hints-10         39.0 ± 0%      12.0 ± 0%  -69.23%  (p=0.008 n=5+5)
```

### DIFF
--- a/pkg/logql/log/parser.go
+++ b/pkg/logql/log/parser.go
@@ -5,9 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"strings"
 	"unicode/utf8"
 
+	"github.com/buger/jsonparser"
 	"github.com/grafana/loki/pkg/logql/log/jsonexpr"
 	"github.com/grafana/loki/pkg/logql/log/logfmt"
 	"github.com/grafana/loki/pkg/logql/log/pattern"
@@ -23,12 +23,17 @@ const (
 	duplicateSuffix = "_extracted"
 	trueString      = "true"
 	falseString     = "false"
+	// How much stack space to allocate for unescaping JSON strings; if a string longer
+	// than this needs to be escaped, it will result in a heap allocation
+	unescapeStackBufSize = 64
 )
 
 var (
 	_ Stage = &JSONParser{}
 	_ Stage = &RegexpParser{}
 	_ Stage = &LogfmtParser{}
+
+	trueBytes = []byte("true")
 
 	errUnexpectedJSONObject = fmt.Errorf("expecting json object(%d), but it is not", jsoniter.ObjectValue)
 	errMissingCapture       = errors.New("at least one named capture must be supplied")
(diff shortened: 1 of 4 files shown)


pkg/logql/log/parser.go
```diff
@@ -23,170 +23,162 @@ const (
...
 		return falseString
 	default:
-		iter.Skip()
 		return ""
 	}
 }
 
+func unescapeJSONString(b []byte) string {
+	var stackbuf [unescapeStackBufSize]byte // stack-allocated array for allocation-free unescaping of small strings
+	bU, err := jsonparser.Unescape(b, stackbuf[:])
```
jeschkies, line 168: This is really nice. Do you know how much overhead `Unescape` is?

I'm wondering if we should go the other way around and use the escape raw string. This makes it a little bit more complicated later on when the labels are matched or read out. Overall I'm wondering if labels should be lazy. I've talked to Ed about this. If we can somehow can ensure the original line is not garbage collected until the process is done it would be enough to keep an index or rather unsafe pointer into the line. No allocation is required.

cyriltovena: Most of the cpu is there, not sure if we can change this without risk. However I do like the lazy idea give it a go. What would be great is to limit extraction but I wasn't able to because the sanitize function is irreversible.

MasslessParticle:
@cyriltovena @jeschkies Is this ready to be merged?

MasslessParticle:
@jeschkies Should we convert this to a draft PR, then?

owen-d (approved):
This is clean, through and through. A pleasure to read, sorry it took me so long.

@jeschkies happy to see what you come up with, but I don't think it should block this PR

vlad-diachenko (commented):
looks awesome !!

vlad-diachenko:
@cyriltovena should we mention such changes in the changelog? I believe that it might worth to mention because it improves json parser performance a lot ... ))

cyriltovena:
> @cyriltovena should we mention such changes in the changelog? I believe that it might worth to mention because it improves json parser performance a lot ... ))

Sure however it would be better to try it in a cluster see the real difference. May be you can compare before and after ? I'd love to know.

PR #7731, opened 2022-11-21
Examples: Add healthckecks in docker example

bitkill:
Adds healthchecks for the loki `docker compose` example, this makes the getting started example more complete,
and can be now used with the `--wait` command.

Use the following commands to test:
```sh
cd examples/getting-started
docker compose up -d --wait
docker compose ps
```

**Checklist**
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [7684](https://github.com/grafana/loki/pull/7684) **kavirajk**: Add missing `embedded-cache` config under `cache_config` doc.
 * [6360](https://github.com/grafana/loki/pull/6099) **liguozhong**: Hide error message when ctx timeout occurs in s3.getObject
 * [7602](https://github.com/grafana/loki/pull/7602) **vmax**: Add decolorize filter to easily parse colored logs.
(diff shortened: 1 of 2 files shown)


kavirajk:
@bitkill thanks for the PR. Two clarifications.
1. After reading the PR description, still I'm failing to understand what exactly is the problem we are trying to fix with these changes? Can you clarify?
2. How do you come up with `retry` and `timeout` values?

bitkill:
>     1. After reading the PR description, still I'm failing to understand what exactly is the problem we are trying to fix with these changes? Can you clarify?
I wanted to be able to have a better base for testing loki locally, and waiting for the systems to be healthy is a good way to automate some tests that I am running.
In the middle of a bash script, I need to be able to launch the docker-compose and wait for it to be all working, to then start to generate logs and send them to loki (without failure).
This also helps me to write ECS containerDefinitions and load balancer healthchecks faster.
>     2. How do you come up with `retry` and `timeout` values?
These were a bit arbitrary, just tried to make the initial wait time (for all services to be healthy) fast, so it wouldn't take a long time just because we are waiting 30s between checks.

bitkill:
@kavirajk forgot to tag you in the response yesterday. response above :point_up:

kavirajk (approved):
LGTM. thanks for the clarification!

PR #7735, opened 2022-11-21
[bugfix] scheduler: Gracefully shutdown querier when using query-scheduler

### LINKED ISSUE #7722: [deadlock] scheduler: if querier OOM restart .
**Describe the bug**
I deploy frontend+scheduler+querier 3 components in k8s to complete logql API query. 

According to my observation, when the queryer is killed by k8s due to OOM, the logql query will be unavailable. And the log keeps reporting errors. 

frontend log
```log
level=error ts=2022-11-18T10:52:00.862957605Z caller=retry.go:77 org_id=123_qaawmopdln 
msg=\"error processing request\" try=0 err=\"context canceled\"
```

The phenomenon is somewhat similar to deadlock. But I'm still investigating the rootcause.

**To Reproduce**
I deploy loki without helm, it's diffcult to reproduce .

**Expected behavior**
if querier was killed, logql should be available.

**Environment:**
 - Infrastructure: [Kubernetes]
 - Deployment tool: [deploy yaml one by one]
 - loki code branch : github master

scheduler log
image

image

prometheus metrics
image

querier cpu and memory
image

frontend querylog
image

liguozhong:
Gracefully shutdown querier when using query-scheduler
This PR is an attempt to fix a bug that my loki cluster is unavailable for logql. The source code and ideas are from mimir of LGTM. Thanks mimir and pr author @pracucci 

Fixes #7722

mimir PR
https://github.com/grafana/mimir/pull/1756
https://github.com/grafana/mimir/pull/1767

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/querier/worker/scheduler_processor.go
```diff
@@ -71,29 +78,33 @@ func (sp *schedulerProcessor) notifyShutdown(ctx context.Context, conn *grpc.Cli
...
 			continue
 		}
 
-		if err := sp.querierLoop(c, address); err != nil {
-			// E.Welch I don't know how to do this any better but context cancelations seem common,
-			// likely because of an underlying connection being close,
-			// they are noisy and I don't think they communicate anything useful.
-			if !strings.Contains(err.Error(), "context canceled") {
+		if err := sp.querierLoop(c, address, inflightQuery); err != nil {
+			// Do not log an error is the query-scheduler is shutting down.
```
jeschkies, line 103:
```suggestion
			// Do not log an error if the query-scheduler is shutting down.
```

pkg/querier/worker/scheduler_processor.go
```diff
@@ -114,23 +125,31 @@ func (sp *schedulerProcessor) querierLoop(c schedulerpb.SchedulerForQuerier_Quer
...
+			if parentSpanContext != nil {
+				queueSpan, spanCtx := opentracing.StartSpanFromContextWithTracer(ctx, tracer, "querier_processor_runRequest", opentracing.ChildOf(parentSpanContext))
+				defer queueSpan.Finish()
 
-			sp.runRequest(ctx, logger, request.QueryID, request.FrontendAddress, request.StatsEnabled, request.HttpRequest)
+				ctx = spanCtx
+			}
+			logger := util_log.WithContext(ctx, sp.log)
 
-			sp.metrics.inflightRequests.Dec()
```
jeschkies, line 133: Where did this go?

pkg/querier/worker/util.go
```diff
@@ -0,0 +1,58 @@
...
+
+	go func() {
+		// Wait until it's safe to cancel the execution context, which is when one of the following conditions happen:
+		// - The worker context has been canceled and there's no inflight query
+		// - The execution context itself has been explicitly canceled
+		select {
+		case <-workerCtx.Done():
+			level.Debug(logger).Log("msg", "querier worker context has been canceled, waiting until there's no inflight query")
+
+			for inflightQuery.Load() {
```
jeschkies, line 40: What happens when the query is never processed? Also, isn't there a potential race condition between testing the flag and setting it in the querier loop? It could be false here but then the next query is received.

jeschkies (changes requested):
This is great work but I need to understand the synchronization of `inflightQuery` a little better.

How do you feel about using a `WaitGroup` instead. This would also avoid using a busy loop. A mutex would be easier to reason about as well.

pkg/querier/worker/scheduler_processor.go
liguozhong: Hi, this was my mistake, I simply ‘copied’ scheduler mimir's code to loki, ignoring the inflightRequests metrics. 
I have submitted a commit which fixes this issue.

pkg/querier/worker/util.go
liguozhong: https://github.com/grafana/mimir/blob/main/pkg/querier/worker/util.go

This util.go is completely copied from mimir. I will deploy this PR to my loki cluster and run it for a while to verify whether this PR will cause unexpected race risks.

liguozhong:
> How do you feel about using a `WaitGroup` instead. This would also avoid using a busy loop. A mutex would be easier to reason about as well.

Thanks you for your timely review. this pr is really important to me .I've been trying to fix #7722 for 18 days.

I prefer to keep the current code, which will make loki and mimir use the same scheduler code, even if there is a problem with this code, it can be fixed together with the mimir community.

dannykopping:
@jeschkies can you take another pass at this pls?

liguozhong:
Good news, I deployed this PR to my loki cluster and Fixes https://github.com/grafana/loki/issues/7722. 

At present, the recording rule has been running stably for 1 day. This PR seems to be useful.

image

pkg/querier/worker/util.go
jeschkies: Hm. We somehow need to document this. I'll try to find the original author.

pracucci: > Also, isn't there a potential race condition between testing the flag and setting it in the querier loop? It could be false here but then the next query is received.

When the querier shutdowns it's expected to cancel the context and so the call to `request, err := c.Recv()` (done in `schedulerProcessor.querierLoop()`) to return error because of the canceled context (I mean the querier context, not the query execution context).

Is there a race? Yes, there's a race between the call to `c.Recv()` and the sequent call to `inflightQuery.Store(true)`, but the time window is very short and we ignored it in Mimir (all in all we want to gracefully handle the 99.9% of cases).

pracucci: > What happens when the query is never processed?

Can you elaborate this?

jeschkies: I was wondering if we can end up in a state were the query is inflight but we shut down. I guess it times out.

pracucci: I think that race condition still exists (I found it very hard to guarantee to never happen) but in practice should be very unlikely.

jeschkies: @liguozhong would you mind add a small comment summarizing Marco's answer?

jeschkies (approved):
Thanks for you hard work and patience. Could you add a comment on the possible race condition.

liguozhong:
> Thanks for you hard work and patience. Could you add a comment on the possible race condition.

done

Issue #7737, opened 2022-11-21
[helm] Allow to provide volumeClaimTemplates selector from Values


TeddyAndrieux:
**Is your feature request related to a problem? Please describe.**

In the previous helm chart, I was able to add a selector to the `storage` `volumeClaimTemplates` using Values file

**Describe the solution you'd like**

Just add a field to the Values file for it

**Describe alternatives you've considered**

Manually patch the StatefulSet or the helm template or stay with an older helm chart

**Additional context**

I will likely open the PR to fix it soon

PR #7743, opened 2022-11-22
helm/loki: Allow to provide selector for Loki persistence volume

TeddyAndrieux:

Add the ability to add a selector for the Loki persistence volumes.

Fixes: #7737

**Checklist**
- [x] Documentation added
- [x] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/Chart.yaml
+++ b/production/helm/loki/Chart.yaml
@@ -4,7 +4,7 @@ name: loki
 description: Helm chart for Grafana Loki in simple, scalable mode
 type: application
 appVersion: 2.6.1
-version: 3.3.4
+version: 3.4.0
 home: https://grafana.github.io/helm-charts
 sources:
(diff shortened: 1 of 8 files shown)


production/helm/loki/Chart.yaml
```diff
@@ -4,7 +4,7 @@ name: loki
 description: Helm chart for Grafana Loki in simple, scalable mode
 type: application
 appVersion: 2.6.1
-version: 3.3.4
+version: 3.4.0
```
jeschkies, line 7: Would 3.3.5 suffice? I'm not sure about the impact.

TeddyAndrieux: I personally have no strong opinion on this
But since it was like a "new feature" ( from the helm chart point of view) I bumped the minor version, even if the change is really minor

PR #7745, opened 2022-11-22
Return error on JSON serialization error.

jeschkies:
Clean up of https://github.com/grafana/loki/pull/7721


pkg/util/marshal/marshal_test.go
```diff
@@ -722,7 +721,7 @@ func randEntries(rand *rand.Rand) []logproto.Entry {
 func Test_EncodeResult_And_ResultValue_Parity(t *testing.T) {
 	f := func(w wrappedValue) bool {
 		var buf bytes.Buffer
-		js := jsoniter.NewStream(jsoniter.ConfigFastest, &buf, 0)
+		js := json.NewStream(json.ConfigFastest, &buf, 0)
```
dannykopping, line 724: It seems that the test here knows too much about the implementation detail, no?

pkg/util/marshal/query.go
```diff
@@ -264,7 +264,7 @@ func encodeStreams(streams logqlmodel.Streams, s *jsoniter.Stream) error {
 
 		err := encodeStream(stream, s)
 		if err != nil {
-			return nil
+			return err
```
dannykopping, line 267: Did a test fail in CI and that's why we're adding this? Or do we need another to test this?

chaudum: Yes, ideally we have a test case for the error case as well.

jeschkies: Alas. You are right...

jeschkies: ...and there was a bug indeed.

pkg/util/marshal/query.go
```diff
@@ -185,28 +185,35 @@ func EncodeResult(v logqlmodel.Result, s *jsoniter.Stream) error {
...
 
 	s.WriteObjectField("resultType")
 	s.WriteString(string(v.Data.Type()))
 
 	s.WriteMore()
 	s.WriteObjectField("result")
-	encodeResult(v.Data, s)
+	err := encodeResult(v.Data, s)
+	if err != nil {
+		return err
```
dannykopping, line 207: Can we wrap these errors for more clarity?

dannykopping (approved):
LGTM, with a clarity nit

PR #7751, opened 2022-11-23
fix logs deletion integration test to properly validate deleted data

sandeepsukhani:
Currently, the integration test for deletion does not validate whether the delete request properly deleted the requested data or not. It just currently checks if the delete request got processed. This was due to the singleton `boltdb-shipper` client causing problems in flushing chunks.

I have used `ResetBoltDBIndexClientWithShipper` to reset the boltdb-shipper client to get the chunk flush from the ingester working. The test now does the checks to ensure that the right data got deleted by the compactor. This should help us catch any bugs earlier for any future changes in deletion code.

**Checklist**
- [x] Tests updated

### DIFF
--- a/integration/loki_micro_services_delete_test.go
+++ b/integration/loki_micro_services_delete_test.go
@@ -117,7 +117,7 @@ func TestMicroServicesDeleteRequest(t *testing.T) {
 		cliIngester.Now = now
 		metrics, err := cliIngester.Metrics()
 		require.NoError(t, err)
-		checkMetricValue(t, "loki_ingester_memory_chunks", metrics, 1)
+		checkMetricValue(t, "loki_ingester_chunks_flushed_total", metrics, 1)
 
 		// reset boltdb-shipper client and restart querier
 		storage.ResetBoltDBIndexClientWithShipper()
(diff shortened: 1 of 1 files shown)


integration/loki_micro_services_delete_test.go
```diff
@@ -203,31 +186,12 @@ func TestMicroServicesDeleteRequest(t *testing.T) {
...
-		time.Sleep(5 * time.Second)
-
-		// Check remote write was successful.
-		require.EqualValues(t, []bool{true, true}, remoteCalled, "one or both of the remote write target were not called")
+		assert.ElementsMatch(t, []string{"lineA", "lineC", "lineD"}, lines, "lineB should not be there")
 	})
 }
 
-func checkLabelValue(t *testing.T, metricName, metrics, tenantID string, expectedValue float64) {
+func checkUserLabelValue(t *testing.T, metricName, metrics, tenantID string, expectedValue float64) {
```
chaudum, line 194: nit: What about naming it `checkUserLabelAndMetricValue`?

integration/loki_micro_services_delete_test.go
```diff
@@ -160,6 +109,37 @@ func TestMicroServicesDeleteRequest(t *testing.T) {
...
 
+	t.Run("flush-logs-and-restart-ingester-querier", func(t *testing.T) {
+		// restart ingester which should flush the chunks
+		require.NoError(t, tIngester.Restart())
+		// ensure that ingester has 0 chunks in memory
+		cliIngester = client.New(tenantID, "", tIngester.HTTPURL())
+		cliIngester.Now = now
+		metrics, err := cliIngester.Metrics()
+		require.NoError(t, err)
+		checkMetricValue(t, "loki_ingester_memory_chunks", metrics, 1)
```
chaudum, line 120: This assertion contradicts with the comment on line 115. Shouldn't there be no chunks in memory?

sandeepsukhani: sorry, my bad. I have updated the metric.
The value is 1 because the metric is lazily registered however the registry still has the previous metric value since it is global.

chaudum (approved):
lgtm

PR #7753, opened 2022-11-23
operator: Check for mandatory CA configmap name in ObjectStorageTLS spec

periklis:
Followup on #7744 to check for the `ObjectStorageTLSSpec.CA` field as mandatory before looking up for ConfigMap. This is a corner case where some one opens the dialog for `ObjectStorageTLSSpec` (e.g. in the OpenShift Console UI) and introduces doing so the default value for `CAKey` but does not select any CA ConfigMap. The same can happen when a user introduced in a YAML manifest the `caKey` field but not the `ca`.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/operator/controllers/loki/lokistack_controller.go
+++ b/operator/controllers/loki/lokistack_controller.go
@@ -142,14 +142,12 @@ func (r *LokiStackReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 func handleDegradedError(ctx context.Context, c client.Client, req ctrl.Request, err error) (ctrl.Result, error) {
 	var degraded *status.DegradedError
 	if errors.As(err, &degraded) {
-		err = status.SetDegradedCondition(ctx, c, req, degraded.Message, degraded.Reason)
-		if err != nil {
-			return ctrl.Result{}, err
+		derr := status.SetDegradedCondition(ctx, c, req, degraded.Message, degraded.Reason)
+		if derr != nil {
+			return ctrl.Result{}, derr
 		}
 
-		return ctrl.Result{
-			Requeue: degraded.Requeue,
-		}, nil
+		return ctrl.Result{Requeue: degraded.Requeue}, err
 	}
 
 	if err != nil {
--- a/operator/controllers/loki/lokistack_controller.go
+++ b/operator/controllers/loki/lokistack_controller.go
@@ -142,12 +142,14 @@ func (r *LokiStackReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 func handleDegradedError(ctx context.Context, c client.Client, req ctrl.Request, err error) (ctrl.Result, error) {
 	var degraded *status.DegradedError
 	if errors.As(err, &degraded) {
-		derr := status.SetDegradedCondition(ctx, c, req, degraded.Message, degraded.Reason)
-		if derr != nil {
-			return ctrl.Result{}, derr
+		err = status.SetDegradedCondition(ctx, c, req, degraded.Message, degraded.Reason)
+		if err != nil {
+			return ctrl.Result{}, err
 		}
 
-		return ctrl.Result{Requeue: degraded.Requeue}, err
+		return ctrl.Result{
+			Requeue: degraded.Requeue,
+		}, nil
 	}
 
 	if err != nil {
--- a/operator/internal/handlers/lokistack_create_or_update.go
+++ b/operator/internal/handlers/lokistack_create_or_update.go
@@ -33,6 +33,10 @@ import (
 	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+const (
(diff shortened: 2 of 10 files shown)


operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -101,7 +101,7 @@ func CreateOrUpdateLokiStack(
 
 	objStore.Schemas = storageSchemas
 
-	if stack.Spec.Storage.TLS != nil {
+	if stack.Spec.Storage.TLS != nil && stack.Spec.Storage.TLS.CA != "" {
```
xperimental, line 104: Should we set [the validation](https://github.com/periklis/loki/blob/82161fe329f6aef99c52368a47a30f096887d14e/operator/apis/loki/v1/lokistack_types.go#L361-L362) on the CRD attribute to "required"?

Should there be a "degraded error" when TLS is set but CA is empty?

periklis: > Should we set [the validation](https://github.com/periklis/loki/blob/82161fe329f6aef99c52368a47a30f096887d14e/operator/apis/loki/v1/lokistack_types.go#L361-L362) on the CRD attribute to "required"?

I think you are right we should set the field `CA` to required in the first place. Means whenever someone creates the spec block, they will be required to enter a configmap (i.e. in the OpenShift Console uncollapsing the block creates the block in the first place). 

> Should there be a "degraded error" when TLS is set but CA is empty?

The handler does already lookup for the ConfigMap first and if not found (includes the empty name case) sets the resource to a degraded condition.

periklis: Need to circle back to the original solution because even if `ObjectStorageTLSSpec` is uncollapsed the console renderer makes the `caName` field required, i.e.:
image
image
image

This is really an unfortunate combination where we need to keep the default on `caKey` but make the entire struct optional. :(

xperimental: Hm, I assumed that the editor would not add the whole `tls` block unless the user actively added it, because that is marked as `optional` as well. An alternative/extension would be to remove the default from the `caKey` and do the default-handling in code instead.

periklis: There is more to add here, i.e. the handling of degraded errors is broken too :(. But making the `caKey` optional and move the default back to code is a good option.

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -33,6 +33,10 @@ import (
 	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+const (
+	caKeyKey = "service-ca.crt"
```
Red-GV, line 37: Would it be better to call this variable `defaultCAKey`?

operator/controllers/loki/lokistack_controller.go
```diff
@@ -142,14 +142,12 @@ func (r *LokiStackReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-			return ctrl.Result{}, err
+		derr := status.SetDegradedCondition(ctx, c, req, degraded.Message, degraded.Reason)
+		if derr != nil {
+			return ctrl.Result{}, derr
 		}
 
-		return ctrl.Result{
-			Requeue: degraded.Requeue,
-		}, nil
+		return ctrl.Result{Requeue: degraded.Requeue}, err
```
xperimental, line 150: I think this can be reverted to the state before. The new variable is not really necessary and the `return ctrl.Result{Requeue: degraded.Requeue}, nil` is intentional, so that the value of `degraded.Requeue` is actually used to decide whether the request should be retried or not.

operator/internal/handlers/lokistack_create_or_update.go
xperimental:
```suggestion
	defaultCAKey = "service-ca.crt"
```

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -123,7 +135,12 @@ func CreateOrUpdateLokiStack(
 			}
 		}
 
-		objStore.TLS = &storageoptions.TLSConfig{CA: cm.Name, Key: stack.Spec.Storage.TLS.CAKey}
+		caKey := stack.Spec.Storage.TLS.CAKey
+		if caKey == "" {
+			caKey = caKeyKey
+		}
```
xperimental, line 141: This works, but I think we have used the "initialize with default, override if set" style before:

```suggestion
		caKey := defaultCAKey 
		if stack.Spec.Storage.TLS.CAKey != "" {
			caKey = stack.Spec.Storage.TLS.CAKey
		}
```

PR #7754, opened 2022-11-23
index-shipper: add support for multiple stores

### LINKED ISSUE #7276: Allow for multiple store configurations from the same provider

ashwanthgoli:

Currently loki initializes a single instance of index-shipper to [handle all the table ranges](https://github.com/grafana/loki/blob/ff7b46297345b215fbf49c2cd4c364d125b6290b/pkg/storage/factory.go#L188) (from across periods) for a given index type `boltdb-shipper, tsdb`. Since index-shipper only has the object client handle to the store defined by `shared_store_type`, it limits the index uploads to a single store. Setting `shared_store_type` to a different store at a later point in time would mean losing access to the indexes stored in the previously configured store.

With this PR, we initialize a separate index-shipper & table manager for each period if `shared_store_type` is not explicity configured. This offers the flexibility to store index in multiple stores (across providers).

**Note**:
- usage of `shared_store_type` in this commit text refers to one of these config options depending on the index in use: `-boltdb.shipper.shared-store`,  `-tsdb.shipper.shared-store`
- `shared_store_type` used to default to the `object_store` from the latest `period_config` if not explicitly configured. This PR removes these defaults in favor of supporting index uploads to multiple stores.

Fixes #7276

All the instances of downloads table manager operate on the same cacheDir. But it shouldn't be a problem as the tableRanges do not overlap across periods.

**Checklist**
- [ ] Documentation added
- [X] Tests updated
- [x] `CHANGELOG.md` updated


docs/sources/upgrading/_index.md
```diff
@@ -95,6 +95,10 @@ level=info ts=2022-12-20T15:27:54.858554127Z caller=metrics.go:147 component=fro
 
 These statistics are also displayed when using `--stats` with LogCLI.
 
+#### Index shipper multi-store support
+If `-boltdb.shipper.shared-store`, `-tsdb.shipper.shared-store` are not explicitly configured, they used to default to the `object_store` configured in the latest `period_config` of the correspoding index type.
```
JStickler, line 99:
```suggestion
In releases prior to 2.7.4, if you did not explicitly configure `-boltdb.shipper.shared-store`, `-tsdb.shipper.shared-store`, those values default to the `object_store` configured in the latest `period_config` of the corresponding index type.
```

docs/sources/upgrading/_index.md
```diff
@@ -95,6 +95,10 @@ level=info ts=2022-12-20T15:27:54.858554127Z caller=metrics.go:147 component=fro
...
+Going forward, these defaults are removed in favor of uploading indexes to multiple stores. If `shared-store` is not explicitly configured, boltdb & tsdb indexes will be shipped to the `object_store` configured for that period.
```
JStickler, line 100:
```suggestion
In releases 2.7.4 and later, these defaults are removed in favor of uploading indexes to multiple stores. If you do not explicitly configure a `shared-store`, the boltdb & tsdb indexes will be shipped to the `object_store` configured for that period.
```

JStickler (commented):
[Doc squad]  Added more specifics about what releases this change applies to (assuming this is going to go into the next release).

pkg/storage/stores/tsdb/store.go
```diff
@@ -62,57 +52,53 @@ type newStoreFactoryFunc func(
...
 
-// NewStore creates a new store if not initialized already.
-// Each call to NewStore will always build a new stores.ChunkWriter even if the store was already initialized since
-// fetcher.Fetcher instances could be different due to periodic configs having different types of object storage configured
-// for storing chunks.
-// It also helps us make tsdb store a singleton because
-// we do not need to build store for each schema config since we do not do any schema specific handling yet.
-// If we do need to do schema specific handling, it would be a good idea to abstract away the handling since
-// running multiple head managers would be complicated and wasteful.
+var tsdbMetrics *Metrics
```
sandeepsukhani, line 55: same here

pkg/storage/factory.go
```diff
@@ -269,33 +273,46 @@ func NewIndexClient(name string, cfg Config, schemaCfg config.SchemaConfig, limi
...
 			return nil, err
 		}
 
-		tableRanges := getIndexStoreTableRanges(config.BoltDBShipperType, schemaCfg.Configs)
-
-		boltDBIndexClientWithShipper, err = shipper.NewShipper(cfg.BoltDBShipperConfig, objectClient, limits,
-			ownsTenantFn, tableRanges, registerer)
+		tableRange := getTableRange()
+		instanceName := fmt.Sprintf("%s_%d", periodCfg.ObjectType, tableRange.Start)
+		shipper, err := shipper.NewShipper(instanceName, cfg.BoltDBShipperConfig, objectClient, limits,
```
sandeepsukhani, line 306: I think you can use `instanceName` as a static label in registerer to get rid of global metrics instances.
It would also be good to pass a logger to `instanceName` in it to be able to filter out logs by schema.

pkg/storage/store.go
```diff
@@ -161,10 +161,21 @@ func (s *store) init() error {
 			return err
 		}
 
-		w, idx, stop, err := s.storeForPeriod(p, chunkClient, f)
+		// getTableRange is used only for initializing boltdb and tsdb index clients.
+		// Computing TableRange for other indexTypes might cause a panic if PeriodConfig.IndexTables.Period is not configured.
```
sandeepsukhani, line 165: The `PeriodConfig.IndexTables.Period` should always be set. Infact we should do a validation for it.
It would be good to pass `config.TableRange` instead of a function to the factory functions.

sandeepsukhani (changes requested):
I have done first pass on the code, it looks good and nice coverage on the tests.
I have left some suggestions. Will do another review once we take care of those suggestions.

pkg/storage/stores/indexshipper/downloads/table_manager.go
```diff
@@ -27,8 +25,7 @@ const (
 	daySeconds           = int64(24 * time.Hour / time.Second)
 )
 
-// regexp for finding the trailing index bucket number at the end of table name
-var extractTableNumberRegex = regexp.MustCompile(`[0-9]+$`)
+var downloadMetrics *metrics
```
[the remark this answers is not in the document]
ashwanthgoli: yeah, I tried using `WrapRegistererWith` but it ends up registering the metric more than once resulting in a panic.
I am looking at [CurryWith](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#HistogramVec.CurryWith), but the [NewHistogramCollector](https://github.com/grafana/loki/blob/25db6bcff287dcf6dc60a8250645573c56566b55/vendor/github.com/weaveworks/common/instrument/instrument.go#L46) used [here](https://github.com/grafana/loki/blob/25db6bcff287dcf6dc60a8250645573c56566b55/pkg/storage/stores/shipper/shipper_index_client.go#L170) doesn't accept [ObserverVec](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#ObserverVec), so I am planning to do the metric refactor in a separate pr after updating the weaveworks pkg. But even this won't get rid of the global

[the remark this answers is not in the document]
ashwanthgoli: I am guessing we didn't notice it so far since `index-gateway` client is suppressing it here - https://github.com/grafana/loki/blob/25db6bcff287dcf6dc60a8250645573c56566b55/pkg/storage/stores/indexshipper/gatewayclient/gateway_client.go#L120
and we've only initialized singleton botldb and tsdb index clients.

pkg/storage/store.go
ashwanthgoli: 0 seems like a valid option when configuring [non-periodic](https://github.com/grafana/loki/blob/25db6bcff287dcf6dc60a8250645573c56566b55/pkg/storage/config/schema_config.go#L529) tables for older indexes, CMIIW

pkg/storage/stores/indexshipper/downloads/table_manager.go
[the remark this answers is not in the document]
ashwanthgoli: removed the global and did a sanity check with multiple boltdb clients, it works. Not sure why it failed last time I tried

pkg/storage/factory.go
ashwanthgoli: we are already adding a component label here, that should suffice I think.
https://github.com/grafana/loki/blob/6d527e78e9474a8d5e59a5d3de989be17dfc5087/pkg/storage/store.go#L219

pkg/storage/stores/tsdb/store.go
```diff
@@ -62,48 +52,43 @@ type newStoreFactoryFunc func(
...
```
sandeepsukhani, line 55: this global should be removed too

pkg/storage/store.go
sandeepsukhani: can we handle that scenario in config.TableRange methods? If `Period` is 0 and table name is same as `Prefix` then just assume table is within the range.

sandeepsukhani (commented):
thanks for taking care of the feedback! Left some minor nits.
I think we should pass a logger to boltdb-shipper and tsdb index clients with schema `from` as a kv added to all the logs. This would help us debug any issues.

docs/sources/upgrading/_index.md
```diff
@@ -95,6 +95,10 @@ level=info ts=2022-12-20T15:27:54.858554127Z caller=metrics.go:147 component=fro
 
 These statistics are also displayed when using `--stats` with LogCLI.
 
+#### Index shipper multi-store support
+In releases prior to 2.7.4, if you did not explicitly configure `-boltdb.shipper.shared-store`, `-tsdb.shipper.shared-store`, those values default to the `object_store` configured in the latest `period_config` of the corresponding index type.
+In releases 2.7.4 and later, these defaults are removed in favor of uploading indexes to multiple stores. If you do not explicitly configure a `shared-store`, the boltdb & tsdb indexes will be shipped to the `object_store` configured for that period.
```
JStickler, line 100:
```suggestion
In releases 2.7.4 and later, these defaults are removed in favor of uploading indexes to multiple stores. If you do not explicitly configure a `shared-store`, the boltdb and tsdb indexes will be shipped to the `object_store` configured for that period.
```

JStickler (approved):
[Docs squad] One small change to conform to the [Style Guide.](https://grafana.com/docs/writers-toolkit/style-guide/capitalization-punctuation/#punctuation-guidelines)

sandeepsukhani (changes requested):
I think it would be a good idea to add some tests in pkg/storage/store_test.go covering various scenarios like multiple boltdb-shipper stores, multiple tsdb stores, combination of both etc.

pkg/storage/stores/shipper/index/table_manager.go
```diff
@@ -163,25 +168,31 @@ func (tm *TableManager) handoverIndexesToShipper(force bool) {
...
-			level.Error(util_log.Logger).Log("msg", "failed to snapshot table for reads", "table", table.name, "err", err)
+			level.Error(tm.logger).Log("msg", "failed to snapshot table for reads", "table", table.name, "err", err)
 		}
 	}
 }
 
 func (tm *TableManager) loadTables() (map[string]*Table, error) {
+	// as we are running multiple instances of TableManger - one for each period
+	// existing tables should be migrated to period specific directory
+	if err := migrateTables(tm.cfg.IndexDir, tm.tableRange, tm.logger); err != nil {
```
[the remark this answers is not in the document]
ashwanthgoli: I began with this approach since overlapping chunks were being indexed on both sides of a period boundary which would've caused conflicting writes from multiple table managers. I think we can remove this now given this problem is fixed with #8251

pkg/storage/stores/tsdb/manager.go
```diff
@@ -281,11 +287,47 @@ func (m *tsdbManager) BuildFromWALs(t time.Time, ids []WALIdentifier) (err error
 	return nil
 }
 
-func indexBuckets(from, through model.Time, tableRanges config.TableRanges) (res []string) {
+func migrateMultitenantDir(dir string, tableRange config.TableRange, logger log.Logger) error {
```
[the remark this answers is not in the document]
ashwanthgoli: Initially I was migrating WALs based on the ts, but removed this code given that the ts is the creation timestamp of the WAL and is not helpful when deciding which period it belongs to. I felt it would be okay not to migrate WALs as long as we gracefully shutdown ingesters during the migration which takes care of [building the prev head](https://github.com/grafana/loki/blob/ashwanth/index-shipper-multi-store/pkg/storage/stores/tsdb/head_manager.go#L220). 
 
But if we want to retain the WALs during migration, we could move them to the latest period dir and have that instance of head manager process them, wdyt?

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -257,17 +257,61 @@ func (m *HeadManager) Start() error {
...
+		return errors.Wrap(err, "rotating tsdb head")
+	}
+
+	m.wg.Add(1)
+	go m.loop()
+
+	return nil
+}
+
+func (m *HeadManager) buildLegacyWALs() error {
```
sandeepsukhani, line 287: lets name it a bit better to make it clearer
```suggestion
func (m *HeadManager) buildTSDBFromLegacyWALs() error {
```

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -257,17 +257,61 @@ func (m *HeadManager) Start() error {
...
+	for _, d := range managerRequiredDirs(parentDir) {
+		if err := util.EnsureDirectory(d); err != nil {
+			return errors.Wrapf(err, "ensuring required legacy directory exists: %s", d)
+		}
+	}
+
+	return m.buildWALs(parentDir, true)
+}
+
+func (m *HeadManager) buildWALs(dir string, legacy bool) error {
```
sandeepsukhani, line 308: lets name it a bit better to make it clearer
```suggestion
func (m *HeadManager) buildTSDBFromWALs(dir string, legacy bool) error {
```

pkg/storage/stores/tsdb/store.go
```diff
@@ -32,78 +33,64 @@ type store struct {
...
-// NewStore creates a new store if not initialized already.
-// Each call to NewStore will always build a new stores.ChunkWriter even if the store was already initialized since
-// fetcher.Fetcher instances could be different due to periodic configs having different types of object storage configured
-// for storing chunks.
-// It also helps us make tsdb store a singleton because
-// we do not need to build store for each schema config since we do not do any schema specific handling yet.
-// If we do need to do schema specific handling, it would be a good idea to abstract away the handling since
-// running multiple head managers would be complicated and wasteful.
+// NewStore creates a new tsdb index ReaderWriter.
 var NewStore = func() newStoreFactoryFunc {
```
sandeepsukhani, line 58: `newStoreFactoryFunc` was just a hack to make tsdb store instances a singleton. I think we can get rid of it.

sandeepsukhani (approved):

sandeepsukhani:
Great job Ashwanth! The changes look great to me. I think we can merge the PR once we get to test it in a dev cell and ensure things work as expected.

sandeepsukhani (approved):
The changes look great and thanks for taking care of all the feedback. Just two minor suggestions, otherwise we are good to merge this.

pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -279,31 +308,32 @@ func (m *HeadManager) Start() error {
...
-		m.metrics.walTruncations.WithLabelValues(statusFailure).Inc()
-		return errors.New("cleaning (removing) wal dir")
-	}
-	m.metrics.walTruncations.WithLabelValues(statusSuccess).Inc()
-
-	err = m.Rotate(now)
-	if err != nil {
-		return errors.Wrap(err, "rotating tsdb head")
+	if legacy {
+		for _, grp := range walsByPeriod {
```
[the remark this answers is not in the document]
sandeepsukhani: I think not since wal is parent dir of store specific wals. Resolving this.

Erichorng:
[https://github.com/grafana/loki/issues/10397#issue-1873340694](url)

PR #7760, opened 2022-11-23
[helm] Bump helm Loki and GEL version

trevorwhitney:

Bumps Loki to 2.7.0 and GEL to 1.6.0 in the helm chart.

### DIFF
--- a/production/helm/loki/values.yaml
+++ b/production/helm/loki/values.yaml
@@ -272,7 +272,7 @@ enterprise:
   enabled: false
 
   # Default verion of GEL to deploy
-  version: v1.5.2
+  version: v1.6.0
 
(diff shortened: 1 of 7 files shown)


DylanGuedes (approved):
lgtm

production/helm/loki/values.yaml
```diff
@@ -320,8 +320,6 @@ enterprise:
     registry: docker.io
     # -- Docker image repository
     repository: grafana/enterprise-logs
-    # -- Overrides the image tag whose default is the chart's appVersion
-    tag: v1.4.0
```
jeschkies, line 324: Why was this removed?

trevorwhitney: it's not used

PR #7762, opened 2022-11-23
Loki: Add new sharding metrics (histogram based)

DylanGuedes:
Add histogram-version of two existing stream sharding metrics. The new metrics are:
- `rate_store_stream_shards`: Will show us the distribution of shards reported by ingesters. This will help us on monitoring the number of shards we're dealing with. As of now we only have access to the max number of shards seen, which is helpful but not enough.
- `rate_store_stream_rate_bytes`: Same as below: will help us understand the distribution of stream sizes instead of only showing us the max stream size seen.

### DIFF
--- a/pkg/distributor/ratestore.go
+++ b/pkg/distributor/ratestore.go
@@ -123,7 +123,7 @@ func (s *rateStore) updateAllRates(ctx context.Context) error {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


MasslessParticle (approved):
LGTM. I think we can get away with dropping `_hist_` from the metric names. We're not using that convention anywhere else so I'm a little wary of adding it here unless we need to.

PR #7769, opened 2022-11-24
Split parallelism across Period Configs

owen-d:
One of things we watch while updating non-TSDB period configs to TSDB period configs is the difference in query parallelism. TSDB dynamically shards queries into (potentially) much smaller units of work compared to the static shard factors uses prior. To account for this, we use much higher query parallelism configurations with TSDB period configs.

This creates a potential problem when querying across `non-tsdb, tsdb` period boundaries: we may want a query parallelism of 512 for the tsdb portion but only 64 for the non-tsdb portion! However, we only had one limit to specify this per tenant, meaning this would be too high when querying non-tsdb periods or too low when querying tsdb ones.

This PR

* Introduces `tsdb_max_query_parallelism` (default `512`) to `limits_config`
* Uses `tsdb_max_query_parallelism` and `max_query_parallelism` limits to find a better parallelism _per query_ by weighting the two respective configs by the proportion of each query spent on TSDB or non-TSDB period configurations.


pkg/querier/queryrange/limits.go
```diff
@@ -332,3 +347,113 @@ func (rt limitedRoundTripper) do(ctx context.Context, r queryrangebase.Request)
...
+		if i+1 < len(configs) {
+			through, _ = minMaxModelTime(end, configs[i+1].From.Time)
+		}
+
+		dur := through.Sub(from)
+
+		if i+1 < len(configs) && configs[i+1].From.Time.Before(end) {
+			dur = configs[i+1].From.Time.Sub(from)
+		}
+		if ty := configs[i].IndexType; ty == config.TSDBType {
```
sandeepsukhani, line 414: can we do this here?
```suggestion
		if configs[i].IndexType == config.TSDBType {
```

pkg/querier/queryrange/limits.go
```diff
@@ -332,3 +347,113 @@ func (rt limitedRoundTripper) do(ctx context.Context, r queryrangebase.Request)
...
+		// or
+		// there is overlap with current config
+		finalOrFuture := i == len(configs)-1 || configs[i].From.After(end)
+		if finalOrFuture {
+			return true
+		}
+		// qEnd not before start && qStart not after end
+		overlapCurrent := !end.Before(configs[i].From.Time) && !start.After(configs[i+1].From.Time)
+		return overlapCurrent
+	})
```
sandeepsukhani, line 379: we can simplify this by doing it like how we do it in the store here
https://github.com/grafana/loki/blob/c8c8303a6a6767b8e6e196e9e160dbb2c6887273/pkg/storage/stores/composite_store.go#L209-L219

pkg/querier/queryrange/limits.go
```diff
@@ -332,3 +347,113 @@ func (rt limitedRoundTripper) do(ctx context.Context, r queryrangebase.Request)
...
+
+	// There was no overlapping index. This can only happen when a time
+	// was requested before the first period config start. In that case, just
+	// use the first period config. It should error elsewhere.
+	if i == len(configs) {
```
sandeepsukhani, line 384: I think we will never hit this condition since we always return at least one true response [here](https://github.com/grafana/loki/pull/7769/files#diff-212968864cd729346d8b76a2372d463f9ae49616aca6a3ba947e70d92fddd6cbR372) i.e when `i == len(configs)-1`.

sandeepsukhani (approved):
Nice job! The code looks great to me.
I have left some minor non-blocking feedback, but I will probably merge this after testing it internally, and then we can do a follow-up PR to address the feedback.

sandeepsukhani:
I think I am going to go ahead and merge this. Will fix any problems that I find with testing the changes.

PR #7784, opened 2022-11-27
[bugfix]  Remove hardwired '127.0.0.1' values in default addresses

isodude:
While running all-in-one Loki instance, it's not possible to listen on anything other than IPv4 since compactor_address is then set to http://127.0.0.1. This enables listening on only IPv6.

Also if frontent_address, scheduler_ring and scheduler_address is not set, the default value is to connect to '127.0.0.1:<grpc_listen_port'. Which means that grpc_listen_address is ignored.

It's not possible to use compactor while using an all-in-one instance and listening to IPv6.

Let's not ignore grpc_listen_address to not confuse users.

As noted in the first commit, the default behavior is changed from defaulting to `127.0.0.1` to `localhost`. Not sure how major that change is though.
My /etc/hosts says
```
$ grep localhost /etc/hosts
127.0.0.1 localhost
```

**Checklist**
- [ ] ~~Documentation added~~
- [ ] ~~Tests updated~~
- [ ] ~~`CHANGELOG.md` updated~~
- [ ] ~~Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 
 * [7708](https://github.com/grafana/loki/pull/7708) **DylanGuedes**: Fix multitenant querying.
 
+* [7784](https://github.com/grafana/loki/pull/7784) **isodude**: Fix default values of connect addresses for compactor and querier workers to work with IPv6.
+
 ##### Changes
 
 #### Promtail
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -703,7 +703,11 @@ func (t *Loki) supportIndexDeleteRequest() bool {
 func (t *Loki) compactorAddress() (string, error) {
 	if t.Cfg.isModuleEnabled(All) || t.Cfg.isModuleEnabled(Read) {
 		// In single binary or read modes, this module depends on Server
-		return fmt.Sprintf("http://127.0.0.1:%d", t.Cfg.Server.HTTPListenPort), nil
+		proto := "http"
+		if len(t.Cfg.Server.HTTPTLSConfig.TLSCertPath) > 0 && len(t.Cfg.Server.HTTPTLSConfig.TLSKeyPath) > 0 {
+			proto = "https"
+		}
+		return fmt.Sprintf("%s://%s:%d", proto, t.Cfg.Server.HTTPListenAddress, t.Cfg.Server.HTTPListenPort), nil
 	}
 
 	if t.Cfg.Common.CompactorAddress == "" {
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -339,6 +339,7 @@ func (t *Loki) initQuerier() (services.Service, error) {
 	querierWorkerServiceConfig := querier.WorkerServiceConfig{
 		AllEnabled:            t.Cfg.isModuleEnabled(All),
 		ReadEnabled:           t.Cfg.isModuleEnabled(Read),
+		GrpcListenAddress:     t.Cfg.Server.GRPCListenAddress,
 		GrpcListenPort:        t.Cfg.Server.GRPCListenPort,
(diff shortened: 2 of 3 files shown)


isodude:
Sigh, why does not Github add a sign-off in their commits.

isodude:
Seems that compactor_client is not possible to configure though.
https://github.com/isodude/loki/blob/c6b55e75122ae0761102395651a4da647c639039/pkg/loki/loki.go#L78

isodude:
> Seems that compactor_client is not possible to configure though. https://github.com/isodude/loki/blob/c6b55e75122ae0761102395651a4da647c639039/pkg/loki/loki.go#L78

Actually, running with `target: all` that works as well.
I am now getting logs back when sending a query.

isodude:
For those in the dark I am running with client certificates, to ensure that the compactor_client connects with the proper information one needs the above patch and this config:
```
target: all
server:
  http_listen_address: "[${IPV6}]"
  http_listen_port: 443
  http_tls_config:
    cert_file: ${CREDENTIALS_DIRECTORY}/default.crt
    key_file: ${CREDENTIALS_DIRECTORY}/default.key
    client_auth_type: RequireAndVerifyClientCert
    client_ca_file: ${CREDENTIALS_DIRECTORY}/ca.crt
compactor_client:
  tls_enabled: true
  tls_cert_path: ${CREDENTIALS_DIRECTORY}/default.crt
  tls_key_path: ${CREDENTIALS_DIRECTORY}/default.key
  tls_ca_path: ${CREDENTIALS_DIRECTORY}/ca.crt
  tls_server_name: "${HOSTNAME}"
common:
  compactor_address: "[${IPV6}]"
```

chaudum (approved):

isodude:
Shall I do the same here, in the same PR?
https://github.com/grafana/loki/blob/c3e92b876c9a2079a9989c3954f420336e3c8182/pkg/querier/worker_service.go#L115

chaudum:
> Shall I do the same here, in the same PR?
> 
> https://github.com/grafana/loki/blob/c3e92b876c9a2079a9989c3954f420336e3c8182/pkg/querier/worker_service.go#L115

Good catch! Yes, you can do it on the same PR. Make sure you change the PR title, otherwise it may get confusing.

isodude:
> > Shall I do the same here, in the same PR?
> > https://github.com/grafana/loki/blob/c3e92b876c9a2079a9989c3954f420336e3c8182/pkg/querier/worker_service.go#L115
> 
> Good catch! Yes, you can do it on the same PR. Make sure you change the PR title, otherwise it may get confusing.

Sure! I'm testing out the patch right now.

isodude:
> > Shall I do the same here, in the same PR?
> > https://github.com/grafana/loki/blob/c3e92b876c9a2079a9989c3954f420336e3c8182/pkg/querier/worker_service.go#L115
> 
> Good catch! Yes, you can do it on the same PR. Make sure you change the PR title, otherwise it may get confusing.

PR updated and commit force pushed.

I changed the commit message of the first commit.

chaudum:
> As noted in the first commit, the default behavior is changed from defaulting to 127.0.0.1 to localhost. Not sure how major that change is though.

That should not cause any problems.

@isodude May I ask one last change before merging: Could you add a changelog entry under Loki / Fixes, stating something like this: "Fix default values of connect addresses for compactor and querier workers to work with IPv6"

isodude:
> @isodude May I ask one last change before merging: Could you add a changelog entry under Loki / Fixes, stating something like this: "Fix default values of connect addresses for compactor and querier workers to work with IPv6"

Sounds fine. Pushed.

chaudum:
@isodude Thanks for your contribution!

PR #7785, opened 2022-11-28
Querier/Ruler: query blocker

dannykopping:
Block malicious or expensive queries using a per-tenant runtime configuration.

N/A

This does not apply to `labels` or `series` queries, nor does it allow for defining other dimensions like query time range. I wanted to keep things simple to start with.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/operations/blocking-queries.md
+++ b/docs/sources/operations/blocking-queries.md
@@ -36,6 +36,8 @@ The available query types are:
 - `filter`: a query with a log filter, e.g. `{env="prod"} |= "error"`
 - `limited`: a query without a filter or a metric aggregation
 
+**Note:** the order of patterns is preserved, so the first matching pattern will be used
+
 ## Observing blocked queries
 
 Blocked queries are logged, as well as counted in the `loki_blocked_queries` metric on a per-tenant basis.
--- a/docs/sources/operations/blocking-queries.md
+++ b/docs/sources/operations/blocking-queries.md
@@ -1,0 +1,45 @@
+---
+title: Blocking Queries
+weight: 60
+---
+# Blocking Queries
+
+In certain situations, you may not be able to control the queries being sent to your Loki installation. These queries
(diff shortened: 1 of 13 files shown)


docs/sources/operations/blocking-queries.md
```diff
@@ -0,0 +1,45 @@
...
+      - pattern: '.*prod.*'
+        regex: true
+
+      # block all metric queries
+      - types: metric
+
+      # block any filter or limited queries matching this regex pattern 
+      - pattern: '.*prod.*'
+        regex: true
+        types: filter,limited
```
periklis, line 30: Is the order of the patterns important? If yes we should mention that first match will be first serve

pkg/logql/blocker.go
```diff
@@ -0,0 +1,94 @@
...
+	logger := log.With(qb.logger, "user", tenant, "type", typ)
+
+	query := qb.q.params.Query()
+	for _, p := range patterns {
+
+		// if no pattern is given, assume we want to match all queries
+		if p.Pattern == "" {
+			p.Pattern = ".*"
+			p.Regex = true
+		}
```
periklis, line 49: This is also mention worthy in the docs WDYT?

periklis (commented):
Looks good to me, `blocked` as naming seems a bit weird. Usually folks skim though pairs like allow/deny lists

docs/sources/operations/blocking-queries.md
dannykopping: That's a good call; the order is deterministic since this is a list - i'll update the docs.

pkg/logql/blocker.go
dannykopping: I don't think so; it's an implementation detail.

This example points out that blocking can be done purely on types.

```yaml
...
      # block all metric queries
      - types: metric
...
```

chaudum (approved):
Really nice addition to control query execution.

PR #7791, opened 2022-11-28
helm/loki: Allow to override default loki `target`

### LINKED ISSUE #7788: [helm] Allow to add Loki `target`

Using the helm chart I cannot add any Loki `target`

I would like from a helm value like `extraTargets` being able to add `target` to Loki command

Manual patch

**Additional context**

TeddyAndrieux:

Add the ability to change the Loki `target` argument. E.g. one might want to use the single-binary which has target `all` but also the `table-manager` which is not part of it. So one can override the target with `all,table-manager`.

Fixes #7788 

**Checklist**
- [x] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/installation/helm/reference.md
+++ b/docs/sources/installation/helm/reference.md
@@ -324,6 +324,7 @@ null
     "runAsNonRoot": true,
     "runAsUser": 10001
   },
+  "targetModule": "tokengen",
   "tolerations": []
 }
 </pre>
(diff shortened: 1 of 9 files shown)


jeschkies:
Thanks @TeddyAndrieux. Out of curiosity why do you need this?

TeddyAndrieux:
> Thanks @TeddyAndrieux. Out of curiosity why do you need this?

In our case, it's because we deploy as single-binary and we still use the "table-manager" but this one is not part of target `all` since Loki 2.4

jeschkies (approved):

PR #7793, opened 2022-11-28
Update delve version to be compatible with Go 1.19

dannykopping:
In order to use the `production/docker` with an interactive debugger, we need to upgrade the Delve tool to a version that works with the version of Go which Loki uses: 1.19

N/A

This is step 1 of 2 in the [release process](https://github.com/grafana/loki/blob/main/docs/sources/maintaining/release-loki-build-image.md) of the build image.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/cmd/loki/Dockerfile.debug
+++ b/cmd/loki/Dockerfile.debug
@@ -1,9 +1,7 @@
 # Directories in this file are referenced from the root of the project not this folder
 # This file is intended to be called from the root like so:
 # docker build -t grafana/loki -f cmd/loki/Dockerfile.debug .
-ARG BUILD_IMAGE
-
-FROM $BUILD_IMAGE as build
+FROM grafana/loki-build-image:0.24.2 as build
 ARG GOARCH="amd64"
 COPY . /src/loki
 WORKDIR /src/loki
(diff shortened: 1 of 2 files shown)


cmd/loki/Dockerfile.debug
```diff
@@ -1,8 +1,9 @@
 # Directories in this file are referenced from the root of the project not this folder
 # This file is intended to be called from the root like so:
 # docker build -t grafana/loki -f cmd/loki/Dockerfile.debug .
+ARG BUILD_IMAGE
```
chaudum, line 4: Where is the BUILD_IMAGE variable set?

DylanGuedes, line 4: I'm really unfamiliar with ARG but could you somehow have a default value for it? So that we could keep as default the value `grafana/loki-build-image`

dannykopping: https://github.com/grafana/loki/blob/main/Makefile#L89

dannykopping: I tried to get a default value working, but it proved tricky.
I've just hardcoded the image version now as it is everywhere else too.

dannykopping: I've just hardcoded the image version now as it is everywhere else too.

DylanGuedes (approved):

Issue #7595, opened 2022-11-04
add grpc support for delete client


ashwanthgoli:
**Is your feature request related to a problem? Please describe.**
To perform query filtering, queriers make HTTP calls to compactor for fetching delete requests.
Ideally we want to use gRPC for interservice communication and users can work with the HTTP api for managing delete requests.

**Describe the solution you'd like**
Add gRPC methods for fetching delete requests and generation numbers, use these methods in querier

**Describe alternatives you've considered**
NA

**Additional context**
NA

periklis:
@ashwanthgoli This is something I hit when fixing the client's TLS config (See #7453). We decided within the maintainers group to implement this post-releasing v2.7.0.

Segflow:
Implemented in #7804. Closing as done.

PR #7804, opened 2022-11-29
use grpc for communicating with compactor for query time filtering of data requested for deletion

sandeepsukhani:
Add grpc support to compactor for getting delete requests and gen number for query time filtering.
Since these requests are internal to Loki, it would be good to use grpc instead of HTTP same as all the internal requests we do in Loki.

I have added a new config for accepting the grpc address of the compactor. I tried having just the existing config and detecting if it is a grpc server, but it was hard to do it reliably, considering the different deployment modes we support. I think it is safe to keep it the same and eventually deprecate the existing config.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


periklis:
@sandeepsukhani You are addressing this issue btw: #7595

MasslessParticle (approved):
This looks good. Nice work!

periklis (approved):
Looks also good to me

pkg/loki/modules.go
```diff
@@ -713,7 +713,7 @@ func (t *Loki) supportIndexDeleteRequest() bool {
 func (t *Loki) compactorAddress() (string, bool, error) {
 	if t.Cfg.isModuleEnabled(All) || t.Cfg.isModuleEnabled(Read) {
 		// In single binary or read modes, this module depends on Server
-		return fmt.Sprintf("127.0.0.1:%d", t.Cfg.Server.GRPCListenPort), true, nil
+		return fmt.Sprintf("%s:%d", t.Cfg.Server.HTTPListenAddress, t.Cfg.Server.GRPCListenPort), true, nil
```
periklis, line 716:
```suggestion
		return fmt.Sprintf("%s:%d", t.Cfg.Server.GRPCListenAddress, t.Cfg.Server.GRPCListenPort), true, nil
```
Did you mean GRPC?

sandeepsukhani: yeah, sorry, my bad, messed up while merging conflicts. Thanks for noticing it and pointing it out!

PR #7805, opened 2022-11-29
Fix alertmanager per tenant override config behavior

aminesnow:
**Context:** #7359 implement multi tenancy for the alertmanager config. However, the behavior of the per-tenant alertmanager config is erroneous, as it tries to merge the global alertmanager config with the per-tenant config. This does not work because there is no way of knowing if certain values were set by the user or if they were undefined and have the default Golang zero value (typically booleans).

This PR changes the behavior of the alertmanager config to override all the configuration instead of trying to merge the global config with the per-tenant config. It also applies default values when appropriate.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/ruler/base/ruler.go
+++ b/pkg/ruler/base/ruler.go
@@ -70,9 +70,9 @@ const (
 	errListAllUser = "unable to list the ruler users"
 
 	// Alertmanager default values
-	AlertmanagerRefreshIntervalDefault = 1 * time.Minute
-	NotificationQueueCapacityDefault   = 10000
-	NotificationTimeoutDefault         = 10 * time.Second
+	alertmanagerRefreshIntervalDefault           = 1 * time.Minute
+	alertmanagerNotificationQueueCapacityDefault = 10000
+	alertmanagerNotificationTimeoutDefault       = 10 * time.Second
 )
 
 // Config is the configuration for the recording rules server.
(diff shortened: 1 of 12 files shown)


dannykopping (commented):
@aminesnow thanks for fixing this, however how can we be sure this is fixed without a test?

aminesnow:
> @aminesnow thanks for fixing this, however how can we be sure this is fixed without a test?

I'll add one

pkg/ruler/base/ruler.go
```diff
@@ -68,6 +68,11 @@ const (
 
 	// errors
 	errListAllUser = "unable to list the ruler users"
+
+	// Alertmanager default values
+	AlertmanagerRefreshIntervalDefault = 1 * time.Minute
+	NotificationQueueCapacityDefault   = 10000
+	NotificationTimeoutDefault         = 10 * time.Second
```
xperimental, line 75: It should be possible to make these private (added a common prefix for easier grouping):

```suggestion
	alertmanagerRefreshIntervalDefault             = 1 * time.Minute
	alertmanagerNotificationQueueCapacityDefault   = 10000
	alertmanagerNotificationTimeoutDefault         = 10 * time.Second
```

xperimental (commented):
Looks fine for me. I'm not approving, because it is not in `operator/` though.

aminesnow:
@dannykopping Does this look good to you?

pkg/ingester/stream_rate_calculator.go
```diff
@@ -35,7 +35,7 @@ type StreamRateCalculator struct {
 
 func NewStreamRateCalculator() *StreamRateCalculator {
 	calc := &StreamRateCalculator{
-		size:     defaultStripeSize,
+		size: defaultStripeSize,
```
dannykopping, line 38: Not sure why these files are included in the PR, I assume because of some auto-formatting run over the whole project. I'd suggest reverting since they're not related to the PR, but if it's a PITA then don't worry.

dannykopping (approved):
LGTM, with a nit.
Thanks!

pkg/ingester/stream_rate_calculator.go
aminesnow: Probably because of the `make lint` I ran. I'll see about reverting it.
Also, do you think we can get this PR on a 2.7.x release? This behavior is wrong and should be backported IMO

aminesnow: It was actually a rebase issue. I hope you don't mind me rebasing to fix this, as you already approved the PR

dannykopping: Sounds good, I agree it should be backported. No prob about rebasing once a review is complete :+1: thanks!

PR #7831, opened 2022-12-01
guard against divide by 0 when splitting parallelism

trevorwhitney:

We saw a spike in divide by zero panics in the code introduced in #7769. I was able to reproduce this error via a test that calculates `WeightedParallelism` with a start that's after the end. Not sure if this is possible, but we definitely saw this happening in our ops environment, so something is causing it, and the fix should guard against it in any case.

**Checklist**
- [X] Tests updated

### DIFF
--- a/pkg/querier/queryrange/limits.go
+++ b/pkg/querier/queryrange/limits.go
@@ -9,19 +9,19 @@ import (
 	"time"
 
 	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/tenant"
 	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/timestamp"
 	"github.com/weaveworks/common/httpgrpc"
 	"github.com/weaveworks/common/user"
 
-	"github.com/grafana/dskit/tenant"
-
 	"github.com/grafana/loki/pkg/logproto"
 	"github.com/grafana/loki/pkg/logql"
 	"github.com/grafana/loki/pkg/querier/queryrange/queryrangebase"
 	"github.com/grafana/loki/pkg/storage/config"
 	"github.com/grafana/loki/pkg/util"
+	util_log "github.com/grafana/loki/pkg/util/log"
 	"github.com/grafana/loki/pkg/util/spanlogger"
 	"github.com/grafana/loki/pkg/util/validation"
 )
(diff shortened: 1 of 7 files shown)


MasslessParticle (approved):

dannykopping (commented):
It seems to me like we should be returning an error to indicate that an expectation failed (such as end time before start time) and let the calling function handle this gracefully with an error.

Returning 0 or 1 here doesn't carry a lot of semantic meaning upon first glance, but admittedly I'm not too familiar with this code.

sandeepsukhani:
> It seems to me like we should be returning an error to indicate that an expectation failed (such as end time before start time) and let the calling function handle this gracefully with an error.
> 
> Returning 0 or 1 here doesn't carry a lot of semantic meaning upon first glance, but admittedly I'm not too familiar with this code.

The reason we are not handling all the negative cases is to avoid having checks all over the place. This code is supposed just to find the right parallelism parameter based on its overlap across tsdb vs non-tsdb schema configs. If we do all the checks here, then we will have to ensure the same checks happen both here and in querier.

pkg/querier/queryrange/limits.go
```diff
@@ -366,11 +366,13 @@ func WeightedParallelism(
 	tsdbMaxQueryParallelism := l.TSDBMaxQueryParallelism(user)
 	regMaxQueryParallelism := l.MaxQueryParallelism(user)
 	if tsdbMaxQueryParallelism+regMaxQueryParallelism == 0 {
+		level.Info(util_log.Logger).Log("msg", "querying disabled for tenant")
 		return 0
 	}
 
 	// query end before start would anyways error out so just short circuit and return 1
 	if end < start {
+		level.Warn(util_log.Logger).Log("msg", "query end time before start, letting downstream code handle it gracefully", "start", start, "end", end)
```
dannykopping, line 375: Do we include the tenant in these logs? And how will we be able to identify which query caused this warning log?

dannykopping (commented):
LGTM but I'm struggling to see how these logs will be operationally useful in this current form

pkg/querier/queryrange/limits.go
sandeepsukhani: yeah, good callout! I have made some changes to log `orgID` and `traceID` with the logs.

dannykopping (approved):
LGTM!

trevorwhitney:
Great work @sandeepsukhani, thanks for picking this up and making it better!

PR #7826, opened 2022-12-01
Update Prometheus dependency to latest release (v2.40.4)

rfratto:
Closes #7811, which is needed for Grafana Agent to update to v2.40 and add support for native histograms.

I did not add support for native histograms to Loki, sorry :)

PR #7841, opened 2022-12-02
loki-build-image: update faillint

rfratto:
faillint needs an update for Go 1.19 where it fails unexpectedly. First observed in #7826.

See fatih/faillint#36


slim-bean (approved):
LGTM

PR #7845, opened 2022-12-04
promtail: add RFC3164 syslog support

aparcar:
I'm no Go programmer so somewhat glued ends together, please be patient. If this is of general interest I'm happy to fill the gaps like testing, documentation and change log. Since this requires changes of a vendor package, I created [a PR](https://github.com/influxdata/go-syslog/pull/46) there as well.

It adds RFC3164 support to promtail. Commonly used for OpenWrt and Ubiquiti syslogs.

I opened a PR directly instead of an issue first.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/clients/pkg/promtail/scrapeconfig/scrapeconfig.go
+++ b/clients/pkg/promtail/scrapeconfig/scrapeconfig.go
@@ -194,6 +194,9 @@ type SyslogTargetConfig struct {
 	// message should be pushed to Loki
 	UseRFC5424Message bool `yaml:"use_rfc5424_message"`
 
+	// IsRFC3164Message defines wether the log is formated as RFC3164
+	IsRFC3164Message bool `yaml:"is_rfc3164_message"`
+
 	// MaxMessageLength sets the maximum limit to the length of syslog messages
 	MaxMessageLength int `yaml:"max_message_length"`
 
--- a/clients/pkg/promtail/targets/heroku/target_test.go
+++ b/clients/pkg/promtail/targets/heroku/target_test.go
@@ -100,7 +100,7 @@ func TestHerokuDrainTarget(t *testing.T) {
 			args: args{
 				RequestBodies: []string{testPayload},
 				RequestParams: map[string][]string{
-					"some_query_param": []string{"app_123", "app_456"},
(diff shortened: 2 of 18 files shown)


MasslessParticle:
This is a reasonable idea. Given this is setup as a configuration option, it'd probably be cleaner to implement it as it's own target rather than having `if rfc3164...` everywhere.

Of course, also it'd need tests

chaudum (commented):
How likely do you think it is to get https://github.com/influxdata/go-syslog/pull/46 into the upstream library (the project is rather stale)?

chaudum:
> This is a reasonable idea. Given this is setup as a configuration option, it'd probably be cleaner to implement it as it's own target rather than having `if rfc3164...` everywhere.
> 
> Of course, also it'd need tests

I'm not sure about making a separate target, but I agree that we should have a better abstraction for the various syslog implementations.

E.g. in the config, there are would now be options for `UseRFC5424Message` and `IsRFC3164Message`, which are hard to distinguish and are not mutually exclusive.

Maybe the syslog target should have a single option `syslog_format` which could default to `rfc5424` and also takes `rfc3164` as possible value.

That configuration change would probably be breaking and therefore something for Loki/Promtail 3.0?

chaudum:
> How likely do you think it is to get [influxdata/go-syslog#46](https://github.com/influxdata/go-syslog/pull/46) into the upstream library (the project is rather stale)?

@MasslessParticle @aparcar What do you think about forking the `go-syslog` project to the `grafana` Github org and reference the fork as long as the addition is not added to the upstream project?

MasslessParticle:
> Maybe the syslog target should have a single option syslog_format which could default to rfc5424 and also takes rfc3164 as possible value.

This is totally reasonable. Putting the syslog stuff behind an interface would help keep the implementation cleaner

MasslessParticle:
Sounds like there's still some demand for this but we may have to pick it up ourselves

aparcar:
Uhm where can I follow your work? I can also give this another go

catap:
@aparcar, @MasslessParticle any news?

PR #7849, opened 2022-12-04
rerun make drone so the newest build image is used

rfratto:
Follow up on #7848, where the Drone config needed to be updated as well


slim-bean (approved):
thanks for updating the docs!

PR #7880, opened 2022-12-08
consider range and offset in queries while looking for schema config for query sharding

sandeepsukhani:
We disable query sharding when a query touches multiple schema configs since they might have different sharding factors and might handle sharding differently. This check was done purely on the start and end time of the query without considering `range` and `offset`, which would change the length of the actual data being queried.

Besides being incorrect, it also causes us to fail queries when migrating between tsdb and non-tsdb stores since both handle query sharding differently. For e.g., if the prev schema is `boltdb-shipper` and starting today, `tsdb` index is being used, so if we do a query `sum(rate({foo="bar"}[24h])` even with start and end within `tsdb` range, we will fail the query with error `incompatible index shard` if dynamic sharding goes with shard factor other than `32`(default shard factor in ingester for inverted index). The reason being the `range` here is `24h` which causes us to process data from previous schema as well.

This PR takes care of the issue by factoring in `range` and `offset` in the queries while looking for schema config for query sharding.

**Checklist**
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/querier/queryrange/split_by_interval.go
+++ b/pkg/querier/queryrange/split_by_interval.go
@@ -276,25 +276,35 @@ func splitByTime(req queryrangebase.Request, interval time.Duration) ([]queryran
 	return reqs, nil
 }
 
-// maxRangeVectorDuration returns the maximum range vector duration within a LogQL query.
-func maxRangeVectorDuration(q string) (time.Duration, error) {
-	expr, err := syntax.ParseSampleExpr(q)
+// maxRangeVectorAndOffsetDuration returns the maximum range vector and offset duration within a LogQL query.
+func maxRangeVectorAndOffsetDuration(q string) (time.Duration, time.Duration, error) {
+	expr, err := syntax.ParseExpr(q)
 	if err != nil {
-		return 0, err
+		return 0, 0, err
 	}
-	var max time.Duration
+
+	if _, ok := expr.(syntax.SampleExpr); !ok {
+		return 0, 0, nil
+	}
+	
+	var maxRVDuration, maxOffset time.Duration
 	expr.Walk(func(e interface{}) {
-		if r, ok := e.(*syntax.LogRange); ok && r.Interval > max {
-			max = r.Interval
+		if r, ok := e.(*syntax.LogRange); ok {
+			if r.Interval > maxRVDuration {
+				maxRVDuration = r.Interval
+			}
+			if r.Offset > maxOffset {
+				maxOffset = r.Offset
+			}
 		}
 	})
-	return max, nil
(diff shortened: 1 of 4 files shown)


pkg/querier/queryrange/split_by_interval.go
```diff
@@ -276,25 +276,35 @@ func splitByTime(req queryrangebase.Request, interval time.Duration) ([]queryran
...
-	expr, err := syntax.ParseSampleExpr(q)
+// maxRangeVectorAndOffsetDuration returns the maximum range vector and offset duration within a LogQL query.
+func maxRangeVectorAndOffsetDuration(q string) (time.Duration, time.Duration, error) {
+	expr, err := syntax.ParseExpr(q)
 	if err != nil {
-		return 0, err
+		return 0, 0, err
+	}
+
+	if _, ok := expr.(syntax.SampleExpr); !ok {
```
kavirajk, line 286: Note this also changes the previous behavior of returning non-nil errors for `LogSelectorExpr` but now it would return `nil` error. 

Do you think this would break anything?

pkg/querier/queryrange/split_by_interval.go
```diff
@@ -276,25 +276,35 @@ func splitByTime(req queryrangebase.Request, interval time.Duration) ([]queryran
...
+		return 0, 0, nil
 	}
-	var max time.Duration
+	
+	var maxRVDuration, maxOffset time.Duration
 	expr.Walk(func(e interface{}) {
-		if r, ok := e.(*syntax.LogRange); ok && r.Interval > max {
-			max = r.Interval
+		if r, ok := e.(*syntax.LogRange); ok {
+			if r.Interval > maxRVDuration {
```
kavirajk, line 293: Q: whats the point of comparing it with `maxRvDuration` and `maxOffset`. They have only zero values right? (from the above declarations)

why not just `maxRVDuration, maxOffset = r.Internval, r.Offset`? Given, there can be only one `LogRange` expression in a given LogQL.? (can LogQL contains multiple Range expression? I doubt it though)

kavirajk (commented):
Nice catch :)

LGTM. Feel like we should lock this behavior with some tests?

pkg/querier/queryrange/split_by_interval.go
sandeepsukhani: I think it is okay since previous usage was only in `metric` query path and `range` and `offset` are only used in `metric` queries.

pkg/querier/queryrange/split_by_interval.go
kavirajk: Gotcha! :+1:

PR #3573, opened 2021-04-02
Fixes a race when using specific tenant and multi-client.

cyriltovena:
This was because each client would try to mutate the original set of labels.
Instead of cloning the map which can be expensive, I created a little helper that generate the string
from the labelset without a set of label name.

Fixes #3571

I also added a test to see if this was reproducible and it was indeed:

```
~/go/src/github.com/grafana/loki master*
❯ go test -timeout 30s -tags dev,gofuzz -race -run ^TestMultiClient_Handle_Race$ github.com/grafana/loki/pkg/promtail/client -v -count=1 -timeout=0s
=== RUN   TestMultiClient_Handle_Race
==================
WARNING: DATA RACE
Read at 0x00c0000b77d0 by goroutine 22:
  runtime.mapaccess2_faststr()
      /usr/local/Cellar/go/1.16.2/libexec/src/runtime/map_faststr.go:107 +0x0
  github.com/grafana/loki/pkg/promtail/client.(*client).getTenantID()
      /Users/ctovena/go/src/github.com/grafana/loki/pkg/promtail/client/client.go:376 +0xcb
  github.com/grafana/loki/pkg/promtail/client.(*client).processEntry()
      /Users/ctovena/go/src/github.com/grafana/loki/pkg/promtail/client/client.go:407 +0x9a
  github.com/grafana/loki/pkg/promtail/client.(*client).run()
      /Users/ctovena/go/src/github.com/grafana/loki/pkg/promtail/client/client.go:235 +0x36a

Previous write at 0x00c0000b77d0 by goroutine 21:
  runtime.mapdelete_faststr()
      /usr/local/Cellar/go/1.16.2/libexec/src/runtime/map_faststr.go:297 +0x0
  github.com/grafana/loki/pkg/promtail/client.(*client).processEntry()
      /Users/ctovena/go/src/github.com/grafana/loki/pkg/promtail/client/client.go:408 +0x144
  github.com/grafana/loki/pkg/promtail/client.(*client).run()
      /Users/ctovena/go/src/github.com/grafana/loki/pkg/promtail/client/client.go:235 +0x36a

Goroutine 22 (running) created at:
  github.com/grafana/loki/pkg/promtail/client.New()
      /Users/ctovena/go/src/github.com/grafana/loki/pkg/promtail/client/client.go:198 +0x7dc
  github.com/grafana/loki/pkg/promtail/client.TestMultiClient_Handle_Race()
[...]

PR #1254, opened 2019-11-13
pkg/promtail/client: Handle fingerprint hash collisions

pstibrany:
**What this PR does / why we need it**:
Due to fingerprint hash collisions, Promtail client can mix entries for different labels into the same stream. This can lead to having entries with incorrect labels, and possible out of order errors.

**Which issue(s) this PR fixes**:
Not sure if there is issue for this.

**Special notes for your reviewer**:
I've used large prime number to find next possible fingerprint value. Any prime (except 2) would work here, as would simple increment. I wanted to avoid using next fingerprint (+1), because fast fingerprint creates similar values for similar labels. 

**Checklist**
- [x] Documentation added (no extra documentation)
- [x] Tests updated

PR #7882, opened 2022-12-08
Promtail: Improve LabelsMap CPU performance

RutgerKe:

The labelsMapToString function takes a relative high amount of CPU in a simple promtail pipeline. By replacing the fmt.Sprintf with a string. Builder we achieve a 50+% improvement in the function.  In my local promtail installation I get about a 10% global cpu reduction. 

```
name                 old time/op    new time/op    delta
LabelsMapToString-8    1.10µs ± 3%    0.30µs ± 4%  -72.40%  (p=0.008 n=5+5)

name                 old alloc/op   new alloc/op   delta
LabelsMapToString-8      344B ± 0%      184B ± 0%  -46.51%  (p=0.008 n=5+5)

name                 old allocs/op  new allocs/op  delta
LabelsMapToString-8      14.0 ± 0%       6.0 ± 0%  -57.14%  (p=0.008 n=5+5)
```

The code does get more complex, and this might be a trade-off. Considering promtail is running on many machines (ie. kubernetes nodes) I think the gain is worth it.

Two alternatives:
- Only replace the `lstrs = append(lstrs, fmt.Sprintf("%s=%q", l, v))` call, as it takes most of the time
- Get rid of the explicit `b.Grow` calls and the tracking of size. This only adds a slight improvement

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/clients/pkg/promtail/client/batch.go
+++ b/clients/pkg/promtail/client/batch.go
@@ -72,7 +72,10 @@ func (b *batch) add(entry api.Entry) error {
 }
 
 func labelsMapToString(ls model.LabelSet, without ...model.LabelName) string {
+	var b strings.Builder
+	var labelsSize int
 	lstrs := make([]string, 0, len(ls))
+
 Outer:
 	for l, v := range ls {
 		for _, w := range without {
@@ -80,11 +83,45 @@ Outer:
 				continue Outer
 			}
 		}
-		lstrs = append(lstrs, fmt.Sprintf("%s=%q", l, v))
+
+		str, s := labelNameValueToString(b, l, v)
+		lstrs = append(lstrs, str)
+		labelsSize += s
+	}
+	if len(lstrs) == 0 {
+		return "{}"
 	}
 
 	sort.Strings(lstrs)
-	return fmt.Sprintf("{%s}", strings.Join(lstrs, ", "))
+	return labelListToString(b, labelsSize, lstrs)
+}
+
+// labelListToString replicates 'fmt.Sprintf("{%s}", strings.Join(lstrs, ", "))', but in a more efficent way. We achieve this by using a string.Builder and keeping track of the size at an earlier stage
+func labelListToString(b strings.Builder, labelsSize int, lstrs []string) string {
+	b.Reset()
+	b.Grow(labelsSize + 2 + (2 * (len(lstrs) - 1)))
+	b.WriteString("{")
+	for i := 0; i < len(lstrs)-1; i++ {
+		b.WriteString(lstrs[i])
+		b.WriteString(", ")
+	}
+	// Write the last element seperatly so we have no trailing comma (', ')
+	b.WriteString(lstrs[len(lstrs)-1])
+	b.WriteString("}")
+	return b.String()
+}
+
+// labelNameValueToString replicates 'fmt.Sprintf("%s=%q", l, v)', but in a more efficent way. We achieve this by using a string.Builder. We also return the size for a later concatenation.
+func labelNameValueToString(b strings.Builder, l model.LabelName, v model.LabelValue) (string, int) {
+	// The strings.Builder
+	b.Reset()
+	stringSize := len(l) + len(v) + 3 // 3 for the equals (=) and two quotes (")
+	b.Grow(stringSize)
+	b.WriteString(string(l))
+	b.WriteString("=\"")
+	b.WriteString(string(v))
+	b.WriteString("\"")
+	return b.String(), stringSize
 }
 
 // sizeBytes returns the current batch size in bytes
--- a/clients/pkg/promtail/client/batch.go
+++ b/clients/pkg/promtail/client/batch.go
@@ -2,13 +2,14 @@ package client
 
 import (
 	"fmt"
-	"sort"
+
 	"strings"
 	"time"
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/snappy"
 	"github.com/prometheus/common/model"
+	"golang.org/x/exp/slices"
 
 	"github.com/grafana/loki/clients/pkg/promtail/api"
 
@@ -71,57 +72,36 @@ func (b *batch) add(entry api.Entry) error {
 	return nil
 }
 
-func labelsMapToString(ls model.LabelSet, without ...model.LabelName) string {
+func labelsMapToString(ls model.LabelSet, without model.LabelName) string {
 	var b strings.Builder
-	var labelsSize int
-	lstrs := make([]string, 0, len(ls))
+	totalSize := 2
+	lstrs := make([]model.LabelName, 0, len(ls))
 
-Outer:
 	for l, v := range ls {
-		for _, w := range without {
-			if l == w {
-				continue Outer
-			}
+		if l == without {
+			continue
 		}
 
-		str, s := labelNameValueToString(b, l, v)
-		lstrs = append(lstrs, str)
-		labelsSize += s
-	}
-	if len(lstrs) == 0 {
-		return "{}"
+		lstrs = append(lstrs, l)
+		totalSize += len(l) + 2 + len(v) + 3
 	}
 
-	sort.Strings(lstrs)
-	return labelListToString(b, labelsSize, lstrs)
-}
+	b.Grow(totalSize)
+	b.WriteByte('{')
+	slices.Sort(lstrs)
+	for i, l := range lstrs {
+		if i > 0 {
+			b.WriteString(", ")
+		}
 
-// labelListToString replicates 'fmt.Sprintf("{%s}", strings.Join(lstrs, ", "))', but in a more efficent way. We achieve this by using a string.Builder and keeping track of the size at an earlier stage
-func labelListToString(b strings.Builder, labelsSize int, lstrs []string) string {
-	b.Reset()
-	b.Grow(labelsSize + 2 + (2 * (len(lstrs) - 1)))
-	b.WriteString("{")
-	for i := 0; i < len(lstrs)-1; i++ {
-		b.WriteString(lstrs[i])
-		b.WriteString(", ")
+		b.WriteString(string(l))
+		b.WriteString(`="`)
+		b.WriteString(strings.ReplaceAll(string(ls[l]), `"`, `\"`))
+		b.WriteString(`"`)
 	}
-	// Write the last element seperatly so we have no trailing comma (', ')
-	b.WriteString(lstrs[len(lstrs)-1])
-	b.WriteString("}")
+	b.WriteByte('}')
+
 	return b.String()
-}
-
-// labelNameValueToString replicates 'fmt.Sprintf("%s=%q", l, v)', but in a more efficent way. We achieve this by using a string.Builder. We also return the size for a later concatenation.
-func labelNameValueToString(b strings.Builder, l model.LabelName, v model.LabelValue) (string, int) {
-	// The strings.Builder
-	b.Reset()
-	stringSize := len(l) + len(v) + 3 // 3 for the equals (=) and two quotes (")
-	b.Grow(stringSize)
-	b.WriteString(string(l))
-	b.WriteString("=\"")
-	b.WriteString(string(v))
-	b.WriteString("\"")
-	return b.String(), stringSize
 }
 
 // sizeBytes returns the current batch size in bytes
--- a/clients/pkg/promtail/client/batch.go
+++ b/clients/pkg/promtail/client/batch.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"fmt"
+	"strconv"
 
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


clients/pkg/promtail/client/batch.go
```diff
@@ -72,19 +72,56 @@ func (b *batch) add(entry api.Entry) error {
...
+	b.WriteString("}")
+	return b.String()
+}
+
+// labelNameValueToString replicates 'fmt.Sprintf("%s=%q", l, v)', but in a more efficent way. We achieve this by using a string.Builder. We also return the size for a later concatenation.
+func labelNameValueToString(b strings.Builder, l model.LabelName, v model.LabelValue) (string, int) {
+	// The strings.Builder
+	b.Reset()
+	stringSize := len(l) + len(v) + 3 // 3 for the equals (=) and two quotes (")
+	b.Grow(stringSize)
```
jeschkies, line 119: I'm not sure you need this. The builder should grow automatically. After a few runs it should not allocate anymore.

clients/pkg/promtail/client/batch.go
```diff
@@ -72,19 +72,56 @@ func (b *batch) add(entry api.Entry) error {
...
 
 	sort.Strings(lstrs)
-	return fmt.Sprintf("{%s}", strings.Join(lstrs, ", "))
+	return labelListToString(b, labelsSize, lstrs)
+}
+
+// labelListToString replicates 'fmt.Sprintf("{%s}", strings.Join(lstrs, ", "))', but in a more efficent way. We achieve this by using a string.Builder and keeping track of the size at an earlier stage
+func labelListToString(b strings.Builder, labelsSize int, lstrs []string) string {
+	b.Reset()
+	b.Grow(labelsSize + 2 + (2 * (len(lstrs) - 1)))
```
jeschkies, line 102: Again, I'm not sure you need this. Did you measure without?

jeschkies (commented):
This is pretty cool. Please see my comments :slightly_smiling_face:

clients/pkg/promtail/client/batch_test.go
```diff
@@ -164,3 +164,17 @@ func TestHashCollisions(t *testing.T) {
...
+	labelSet := make(model.LabelSet)
+	labelSet["label"] = "value"
+	labelSet["label1"] = "value"
+	labelSet["label2"] = "value"
+	labelSet["__tenant_id__"] = "value"
+
+	ReservedLabelTenantID := "__tenant_id__"
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_ = labelsMapToString(labelSet, model.LabelName(ReservedLabelTenantID))
```
kavirajk, line 178: I'm bit concerned about using `_` to store the result here. Mainly because of some internal compiler optimizations. Compiler may think why call this func `labelsMapToString` in tight loop if result is not used at all.

I would prefer to have some variable outside loop to store the result.

```
var result string
for i:=0; i<b.N; i++ {
    result = ....
}
```

and compare the results again.

RutgerKe: Thanks, I re-ran the benchmark with the code below and get the same results :relieved: 

```
// store the result to a package level variable
// so the compiler cannot eliminate the Benchmark itself.
var result string

func BenchmarkLabelsMapToString(b *testing.B) {
	labelSet := make(model.LabelSet)
	labelSet["label"] = "value"
	labelSet["label1"] = "value"
	labelSet["label2"] = "value"
	labelSet["__tenant_id__"] = "value"

	ReservedLabelTenantID := "__tenant_id__"
	b.ResetTimer()
	var r string
	for i := 0; i < b.N; i++ {
		// store in r prevent the compiler eliminating the function call.
		r = labelsMapToString(labelSet, model.LabelName(ReservedLabelTenantID))
	}
	result = r
}
```

clients/pkg/promtail/client/batch.go
RutgerKe: We actually create a new builder for every log line, so its called quite a bit. It does have an impact.

Or would you say I should extract the builder and re-use it more?

clients/pkg/promtail/client/batch.go
```diff
@@ -72,19 +72,56 @@ func (b *batch) add(entry api.Entry) error {
...
+	b.WriteString("{")
+	for i := 0; i < len(lstrs)-1; i++ {
+		b.WriteString(lstrs[i])
+		b.WriteString(", ")
+	}
+	// Write the last element seperatly so we have no trailing comma (', ')
+	b.WriteString(lstrs[len(lstrs)-1])
+	b.WriteString("}")
```
RutgerKe, line 110: We are actually using the resulting value as a key in a map, and I don't think it has to be a valid label, just unique. So I can take away the wrapping `{` and some other constants, I don't expect performance gains, but it'll simplify the code a bit.

RutgerKe (commented):
Thanks for the reviews!

Some context: I looked at old PRs around these lines of code and found [this comment](https://github.com/grafana/loki/pull/1254#issuecomment-554951515), already talking about optimizing this specific logic. (it used to be `labelset.String`, [this PR](https://github.com/grafana/loki/pull/3573/files) extracted the .String function and modified it slightly).

clients/pkg/promtail/client/batch.go
colega: This allows allocating the needed space from the beginning, instead of growing in smaller bits which causes allocations and copying of the underlying slice.

clients/pkg/promtail/client/batch.go
```diff
@@ -72,19 +72,56 @@ func (b *batch) add(entry api.Entry) error {
...
+
+// labelNameValueToString replicates 'fmt.Sprintf("%s=%q", l, v)', but in a more efficent way. We achieve this by using a string.Builder. We also return the size for a later concatenation.
+func labelNameValueToString(b strings.Builder, l model.LabelName, v model.LabelValue) (string, int) {
+	// The strings.Builder
+	b.Reset()
+	stringSize := len(l) + len(v) + 3 // 3 for the equals (=) and two quotes (")
+	b.Grow(stringSize)
+	b.WriteString(string(l))
+	b.WriteString("=\"")
+	b.WriteString(string(v))
```
colega, line 122: :exclamation: Note that this was quoting the value previously, which escaped the quotes within the value: https://go.dev/play/p/1ijtqdBBwJu

You need to do something like:

```go
var (
    quoteBytes := []byte{'"'}
    escapedQuoteBytes := []byte{'"'}
)

func labelNameValueToString(b strings.Builder, l model.LabelName, v model.LabelValue) (string, int) {
    // ...
    b.Write(bytes.Replace(v, quoteBytes, escapedQuoteBytes)
    // ...
}
```

clients/pkg/promtail/client/batch.go
```diff
@@ -72,19 +72,56 @@ func (b *batch) add(entry api.Entry) error {
...
+		b.WriteString(", ")
+	}
+	// Write the last element seperatly so we have no trailing comma (', ')
+	b.WriteString(lstrs[len(lstrs)-1])
+	b.WriteString("}")
+	return b.String()
+}
+
+// labelNameValueToString replicates 'fmt.Sprintf("%s=%q", l, v)', but in a more efficent way. We achieve this by using a string.Builder. We also return the size for a later concatenation.
+func labelNameValueToString(b strings.Builder, l model.LabelName, v model.LabelValue) (string, int) {
```
colega, line 115: Note that you're passing a copy of a builder here, an empty copy, so it's basically useless.

clients/pkg/promtail/client/batch.go
RutgerKe: Thanks! Good catch. As I realized later and mentioned in [a comment](https://github.com/grafana/loki/pull/7882#discussion_r1045622980) its actually just a key in a map, and it doesn't have to stay the same in this PR.

The place where we turn this map of streams into push requests we discard the key: https://github.com/grafana/loki/blob/05d21106817438f523c6161d0a1312812d6db44c/clients/pkg/promtail/client/batch.go#L125

colega: We also [set it in the stream](https://github.com/grafana/loki/blob/b2a85ad7945320765c064836f27b967a2d3fa647/clients/pkg/promtail/client/batch.go#L67-L68) so it's forwarded in the request. (And then that `Labels` field is parsed later)

inkel: Rather than using `strings.Replace` you might want to check [`strconv.Quote`](https://pkg.go.dev/strconv#Quote) which will properly escape all characters that need to be escaped.

colega: I explicitly tried to avoid that as it would cause more allocations here, but you're right in the fact that it quotes more than just a quote rune.

In that case, I'd replace strings.Builder by a byte buffer, and use `strconv.AppendQuote` on that buffer instead, and would do `*(*string)(unsafe.Pointer(&buf))` at the end (which is what `strings.Buidlder` does in its `String()` method.

clients/pkg/promtail/client/batch.go
jeschkies: Reusing the builder with reset soups be even faster.

clients/pkg/promtail/client/batch.go
RutgerKe: What would be the benefit of that last suggestion? I'm open to adding that, but I do feel like it hurts the readability a bit.

colega: It does hurt the readability. The benefit is that we would save yet more allocations (`strconv.Quote` does a few even if there's nothing to change in the quoted string, while `strconv.AppendQuote` can make it in zero allocations)

clients/pkg/promtail/client/batch.go
```diff
@@ -71,20 +73,36 @@ func (b *batch) add(entry api.Entry) error {
...
+	b.WriteByte('{')
+	slices.Sort(lstrs)
+	for i, l := range lstrs {
+		if i > 0 {
+			b.WriteString(", ")
+		}
+
+		b.WriteString(string(l))
+		b.WriteString(`="`)
+		b.WriteString(strconv.Quote(string(ls[l])))
```
slim-bean, line 100: strconv.Quote looks like it creates a string with quotes, the quotes added before and after this end up doubling up on the quotes and making an invalid label string.

inkel (approved):

clients/pkg/promtail/client/batch.go
```diff
@@ -71,20 +73,35 @@ func (b *batch) add(entry api.Entry) error {
...
-			if l == w {
-				continue Outer
-			}
+		if l == without {
+			continue
 		}
-		lstrs = append(lstrs, fmt.Sprintf("%s=%q", l, v))
+
+		lstrs = append(lstrs, l)
+		totalSize += len(l) + 2 + len(v) + 3
```
owen-d, line 87: Can we get a comment explaining the constants in `totalSize` here?

owen-d, line 87:
```suggestion
		// guess size increase: 2 for `, ` between labels and 3 for the `=` and quotes around label value
		totalSize += len(l) + 2 + len(v) + 3
```

owen-d (approved):
Nit, but lgtm. Thanks and sorry for the wait on review!

clients/pkg/promtail/client/batch_test.go
```diff
@@ -164,3 +164,23 @@ func TestHashCollisions(t *testing.T) {
...
+// store the result to a package level variable
+// so the compiler cannot eliminate the Benchmark itself.
+var result string
+
+func BenchmarkLabelsMapToString(b *testing.B) {
+	labelSet := make(model.LabelSet)
+	labelSet["label"] = "value"
+	labelSet["label1"] = "value"
+	labelSet["label2"] = "value"
+	labelSet["__tenant_id__"] = "value"
```
owen-d, line 177:
```suggestion
	labelSet["__tenant_id__"] = "another_value"
```

Changing this to avoid tripping `goconst` linter suggesting we make `"value"` a constant

owen-d (approved):

clients/pkg/promtail/client/batch_test.go
```diff
@@ -164,3 +164,23 @@ func TestHashCollisions(t *testing.T) {
...
 }
+
+// store the result to a package level variable
+// so the compiler cannot eliminate the Benchmark itself.
+var result string
+
+func BenchmarkLabelsMapToString(b *testing.B) {
+	labelSet := make(model.LabelSet)
+	labelSet["label"] = "value"
+	labelSet["label1"] = "value"
```
owen-d, line 175:
```suggestion
	labelSet["label1"] = "value2"
```

clients/pkg/promtail/client/batch_test.go
```diff
@@ -164,3 +164,23 @@ func TestHashCollisions(t *testing.T) {
...
+	labelSet["label2"] = "value"
```
owen-d, line 176:
```suggestion
	labelSet["label2"] = "value3"
```

owen-d: goconst 😭

PR #7899, opened 2022-12-09
Fix sizing tool.

jeschkies:
**What this PR does / why we need it**
The current sizing tool is broken because it would not allow cross origin requests. See LogQL Analyzer (https://github.com/grafana/loki/blob/main/docs/sources/logql/analyzer/script.js#L115) for the proper way to call its API.


trevorwhitney (approved):
lgtm

PR #7920, opened 2022-12-12
[helm] add third scalable target `backend`

trevorwhitney:

Adds the new 3rd scalable deployment mode target (backend) to the helm chart. This new target was added in #7650

This PR also adds a new loki dev env in `tools/dev/k3d`. It's currently hardcoding to a Loki image that has this target in it so I can test this new mode. We can change that after the next release or weekly to point to something a bit more canonical.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


production/helm/loki/templates/read/deployment-read.yaml
```diff
@@ -0,0 +1,135 @@
...
+      securityContext:
+        {{- toYaml .Values.loki.podSecurityContext | nindent 8 }}
+      terminationGracePeriodSeconds: {{ .Values.read.terminationGracePeriodSeconds }}
+      containers:
+        - name: read
+          image: {{ include "loki.image" . }}
+          imagePullPolicy: {{ .Values.loki.image.pullPolicy }}
+          args:
+            - -config.file=/etc/loki/config/config.yaml
+            - -target={{ .Values.read.targetModule }}
```
jeschkies, line 56: Where is this defined?

production/helm/loki/templates/backend/service-backend-headless.yaml
```diff
@@ -0,0 +1,25 @@
+{{- $isSimpleScalable := eq (include "loki.deployment.isScalable" .) "true" -}}
+{{- if and $isSimpleScalable (not .Values.read.legacyReadTarget ) }}
+---
+apiVersion: v1
+kind: Service
```
jeschkies, line 5: Why is this needed? I see a service below. Is there some duplication?

jeschkies (commented):
Nice to see this happening. How would one migrate from Helm Chart 3.x?

production/helm/loki/templates/backend/service-backend-headless.yaml
trevorwhitney: I don't know k8s well enough to answer this, but I'm following the pattern of our other statefulsets that have both a service and a headless service. The headless service is used as the statefulset's `spec.serviceName`, whereas the other service is used for load-balancing.

production/helm/loki/templates/read/deployment-read.yaml
trevorwhitney: https://github.com/trevorwhitney/loki/blob/0621044bcb4df2388721e604ff46b61e22506529/production/helm/loki/values.yaml#L825

jeschkies (changes requested):
I think this is good to go. However, what is the migration process from 3.x? What documentation should we add?

trevorwhitney:
@jeschkies migration guide added

PR #7925, opened 2022-12-13
fix bugs in logs results caching and its tests

sandeepsukhani:
When a logs query results in an empty response, we cache it to avoid doing that query again and respond straight away with an empty response. However, we cache a single entry per time split interval with the query itself to keep things simple. For example, if the time split config for tenant `A` is `30m`, then queries for intervals `10m`-`20m` and `21m`-`25m` would have the same cache key.

Here is roughly how cache hit is handled:

* If the new query is within the cached query bounds, return empty results
* If the start of new query is before the start time of the cached query, do a query from `newQuery.Start` to `cachedQuery.Start`
* If the response of last query is also empty, set `cachedQuery.Start` = `newQuery.Start`
* If the end of new query is after the end time of the cached query, do a query from `cachedQuery.End` to `newQuery.Start`
* If the response of last query is also empty, set `cachedQuery.End` = `newQuery.End`
* If we have changes in `cachedQuery.Start/End`, update it in the cache.

The problem here is when we do queries to fill the gap, we sometimes do queries for the range outside of what the user requested and respond back without reducing the response to what the user requested. For example, if the cached query is from `21m`-`25m` and the user query is from `10m`-`15m`, we will query for the whole gap i.e `10m`-`21m`. If there are logs from `15m`-`21m` in the response, we will unexpectedly send it back to the user.

This PR takes care of this issue by extracting the data and sending back only the user's requested logs.

I have also found the tests for logs results cache were incorrect. They heavily use [mergeLokiResponse](https://github.com/grafana/loki/blob/e2842c69c5e089389b697faa792e368f42f84c19/pkg/querier/queryrange/codec.go#L1014) for building the test data, but they were not setting the `limit` in test queries which was causing `mergeLokiResponse` to send back empty results. This means were were always comparing two empty results, which would always be the same.
This PR also takes care of fixing it, and relevant changes to correct the test.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/querier/queryrange/log_result_cache.go
```diff
@@ -180,8 +180,7 @@ func (l *logResultCache) handleHit(ctx context.Context, cacheKey string, cachedR
 	result := emptyResponse(cachedRequest)
 	// if the request is the same and cover the whole time range,
 	// we can just return the cached result.
-	if !lokiReq.GetStartTs().After(cachedRequest.GetStartTs()) && lokiReq.GetStartTs().Equal(cachedRequest.GetStartTs()) &&
-		!lokiReq.GetEndTs().Before(cachedRequest.GetEndTs()) && lokiReq.GetEndTs().Equal(cachedRequest.GetEndTs()) {
```
sandeepsukhani, line 184: this check looked wrong to me, so I fixed it as well.

pkg/querier/queryrange/log_result_cache.go
```diff
@@ -274,6 +273,45 @@ func (l *logResultCache) handleHit(ctx context.Context, cacheKey string, cachedR
...
+		ErrorType:  r.ErrorType,
+		Error:      r.Error,
+		Statistics: r.Statistics,
+		Data: LokiData{
+			ResultType: r.Data.ResultType,
+			Result:     []logproto.Stream{},
+		},
+	}
+	for _, stream := range r.Data.Result {
+		if stream.Entries[0].Timestamp.After(end) || stream.Entries[len(stream.Entries)-1].Timestamp.Before(start) {
```
kavirajk, line 292: The condition one above uses `ts.Nano()` where as here we use `time.After` for comparison. Can we make both same?

kavirajk (approved):
LGTM :+1:  Just minor nit about consistency in time comparisons.

Either we stick to converting to `UnixNano()` and do `< , >, =` on it, Or do `time.Compare` . but not both.

pkg/querier/queryrange/log_result_cache.go
sandeepsukhani: I went with different ways to compare the time for convenience and readability.
It is hard to do `<=` or `>=` with `time.After` and `time.Before` while keeping it simple to read.
If I do the same here, then it would make the code a little inefficient due to having to do the conversion for each log line. I think it would be good to keep it as is, but please feel free to push back if you feel otherwise.

PR #7926, opened 2022-12-13
Fix validation for pattern and rexexp parsers to they don't panic in query-frontend

MichelHollands:

Running the query `topk(10,sum by(namespace)(count_over_time({application="nginx", site!="eu-west-1-dev"} |= "/artifactory/" != "api" != "binarystore" | regexp `` [1d])))` causes a panic.

This PR adds validation for regexp and pattern parsers at parse time. If they don't have a (valid) parameter an error will be returned. 

**Checklist**
- [ ] Documentation added
- [X] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ##### Fixes
 
+* [7926](https://github.com/grafana/loki/pull/7926) **MichelHollands**: Fix validation for pattern and rexexp parsers to they don't panic in query-frontend.
 * [7720](https://github.com/grafana/loki/pull/7720) **sandeepsukhani**: fix bugs in processing delete requests with line filters.
 * [7708](https://github.com/grafana/loki/pull/7708) **DylanGuedes**: Fix multitenant querying.
 * [7784](https://github.com/grafana/loki/pull/7784) **isodude**: Fix default values of connect addresses for compactor and querier workers to work with IPv6.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 
 ##### Fixes
 
-* [7926](https://github.com/grafana/loki/pull/7926) **MichelHollands**: Fix validation for pattern and rexexp parsers to they don't panic in query-frontend.
+* [7926](https://github.com/grafana/loki/pull/7926) **MichelHollands**: Fix validation for pattern and regexp parsers.
 * [7720](https://github.com/grafana/loki/pull/7720) **sandeepsukhani**: fix bugs in processing delete requests with line filters.
 * [7708](https://github.com/grafana/loki/pull/7708) **DylanGuedes**: Fix multitenant querying.
 * [7784](https://github.com/grafana/loki/pull/7784) **isodude**: Fix default values of connect addresses for compactor and querier workers to work with IPv6.
--- a/pkg/logql/rangemapper_test.go
+++ b/pkg/logql/rangemapper_test.go
@@ -1750,4 +1750,6 @@ func Test_FailQuery(t *testing.T) {
 	require.Error(t, err)
 	_, _, err = rvm.Parse(`topk(0, sum(count_over_time({app="foo"} | json |  __error__="" [15m])))`)
 	require.Error(t, err)
+	_, _, err = rvm.Parse(`topk(10,sum by(namespace)(count_over_time({application="nginx", site!="eu-west-1-dev"} |= "/artifactory/" != "api" != "binarystore" | regexp[1d])))`)
+	require.Error(t, err)
 }
--- a/pkg/logql/rangemapper_test.go
+++ b/pkg/logql/rangemapper_test.go
@@ -1752,7 +1752,7 @@ func Test_FailQuery(t *testing.T) {
 	_, _, err = rvm.Parse(`topk(0, sum(count_over_time({app="foo"} | json |  __error__="" [15m])))`)
 	require.Error(t, err)
 
-	// Fix for bug where missing or empty parameters for regexp and pattern parsers threw a panic
+	// Check fixes for bug where missing or empty parameters for regexp and pattern parsers threw a panic
 	// Missing parameter to regexp parser
 	_, _, err = rvm.Parse(`topk(10,sum by(namespace)(count_over_time({application="nginx", site!="eu-west-1-dev"} |= "/artifactory/" != "api" != "binarystore" | regexp [1d])))`)
 	require.ErrorIs(t, err, logqlmodel.ErrParse)
--- a/pkg/logql/rangemapper_test.go
+++ b/pkg/logql/rangemapper_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/grafana/loki/pkg/logqlmodel"
 	"github.com/stretchr/testify/require"
 )
 
@@ -1751,5 +1752,5 @@ func Test_FailQuery(t *testing.T) {
 	_, _, err = rvm.Parse(`topk(0, sum(count_over_time({app="foo"} | json |  __error__="" [15m])))`)
 	require.Error(t, err)
 	_, _, err = rvm.Parse(`topk(10,sum by(namespace)(count_over_time({application="nginx", site!="eu-west-1-dev"} |= "/artifactory/" != "api" != "binarystore" | regexp[1d])))`)
-	require.Error(t, err)
+	require.ErrorIs(t, err, logqlmodel.ErrParse)
 }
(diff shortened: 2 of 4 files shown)


pkg/logql/rangemapper_test.go
```diff
@@ -1750,4 +1751,18 @@ func Test_FailQuery(t *testing.T) {
...
+	_, _, err = rvm.Parse(`topk(10,sum by(namespace)(count_over_time({application="nginx", site!="eu-west-1-dev"} |= "/artifactory/" != "api" != "binarystore" | regexp [1d])))`)
+	require.ErrorIs(t, err, logqlmodel.ErrParse)
+	// Empty parameter to regexp parser
+	_, _, err = rvm.Parse(`topk(10,sum by(namespace)(count_over_time({application="nginx", site!="eu-west-1-dev"} |= "/artifactory/" != "api" != "binarystore" | regexp \"\" [1d])))`)
+	require.ErrorIs(t, err, logqlmodel.ErrParse)
+	// Empty parameter to pattern parser
+	_, _, err = rvm.Parse(`topk(10,sum by(namespace)(count_over_time({application="nginx", site!="eu-west-1-dev"} |= "/artifactory/" != "api" != "binarystore" | pattern \"\" [1d])))`)
+	require.ErrorIs(t, err, logqlmodel.ErrParse)
+	// Empty parameter to json parser
+	_, _, err = rvm.Parse(`topk(10,sum by(namespace)(count_over_time({application="nginx", site!="eu-west-1-dev"} |= "/artifactory/" != "api" != "binarystore" | json [1d])))`)
```
dannykopping, line 1766: Why should this one fail?

pkg/logql/syntax/ast.go
```diff
@@ -332,6 +332,19 @@ type LabelParserExpr struct {
 }
 
 func newLabelParserExpr(op, param string) *LabelParserExpr {
+	if op == OpParserTypeRegexp {
```
dannykopping, line 335: Can we not handle this error higher up? A missing param to these parsers is a syntax error; specifically handling this here seems like we're addressing the symptom not the cause.

pkg/logql/rangemapper_test.go
MichelHollands: I assume you're referring to the pattern test. The pattern parser throws an error if there are no captures: see https://github.com/grafana/loki/blob/main/pkg/logql/log/pattern/ast.go#L20.

dannykopping: I'm referring to the json test

pkg/logql/syntax/ast.go
dannykopping: I mean couldn't we handle this in the grammar?
It's strange to me that these queries are even submitted because they don't match the grammar

```yacc
...
  | REGEXP STRING  { $$ = newLabelParserExpr(OpParserTypeRegexp, $2) }
  | PATTERN STRING { $$ = newLabelParserExpr(OpParserTypePattern, $2) }
```

In any case, I'm surprised why this error isn't handled by the `func (m MultiStageExpr) stages()` function, because instantiating a new pattern parser with an invalid expression should return an error.

It feels awkward to special-case these here, and it makes it possible us to add new parsers later and forget to special-case those here too.

pkg/logql/rangemapper_test.go
MichelHollands: That was added to make sure I didn't mess up the other stages. Note that `rvm.Parse` doesn't return an error there.

dannykopping: Ah, I see. Thanks!

pkg/logql/syntax/ast.go
MichelHollands: These queries match the grammar. They pass an empty string ``. The STRING token in our grammar matches this. I'll see if I can create a special string type for these parsers that only matches if it's not an empty string.
The `Stage()` method only seems to be called from `Pipeline()`, which is done after parsing. Perhaps this was done to avoid creating a regex until needed.

dannykopping: In the query you posted in the description, I don't see an empty string param:

```logql
topk(10,sum by(namespace)(count_over_time({application="nginx", site!="eu-west-1-dev"} |= "/artifactory/" != "api" != "binarystore" | regexp[1d])))
```

If no param is supplied, do you mean it defaults to the zero value of `STRING` which I guess would be `""`?

MichelHollands: That is actually not the query that was posted, it is the query printed when the original panic occurred. Susana found this in the [logs](https://admin-ops-us-east-0.grafana.net/grafana/explore?orgId=1&left=%7B%22datasource%22:%22loki-ops%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-ops%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%7Bcluster%3D%5C%22prod-us-central-0%5C%22,job%3D%5C%22loki-prod%2Fquery-frontend%5C%22%7D%20%21%3D%20%5C%22level%3D%5C%22%22,%22queryType%22:%22range%22,%22hide%22:true%7D,%7B%22refId%22:%22B%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-ops%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%7Bcluster%3D%5C%22prod-us-central-0%5C%22,job%3D%5C%22loki-prod%2Fquery-frontend%5C%22%7D%20%7C%3D%20%5C%22regexp%5B%5C%22%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221670417578824%22,%22to%22:%221670421178824%22%7D%7D). Note the second line which has the GET command that has `regexp+%60%60`.  These last 2 characters translate to  ``.

MichelHollands: That test case can be removed if needed.

dannykopping: I'd suggest updating the description to reflect the actual query that causes the panic though

MichelHollands: Done.

dannykopping (approved):
LGTM!

periklis:
@dannykopping Sounds like another backportworthy PR for 2.7.x?

dannykopping:
> @dannykopping Sounds like another backportworthy PR for 2.7.x?

Good idea!

dannykopping (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 
 ##### Fixes
 
+* [7926](https://github.com/grafana/loki/pull/7926) **MichelHollands**: Fix validation for pattern and rexexp parsers to they don't panic in query-frontend.
```
ssncferreira, line 20: The panic in the query-frontend is to identify these sort of problems, that otherwise would be missed. Therefore, I think we can just mention the fix of the parser

```suggestion
* [7926](https://github.com/grafana/loki/pull/7926) **MichelHollands**: Fix validation for pattern and regexp parsers.
```

pkg/logql/rangemapper_test.go
```diff
@@ -1750,4 +1751,17 @@ func Test_FailQuery(t *testing.T) {
 	require.Error(t, err)
 	_, _, err = rvm.Parse(`topk(0, sum(count_over_time({app="foo"} | json |  __error__="" [15m])))`)
 	require.Error(t, err)
+	// Check fixes for bug where missing or empty parameters for regexp and pattern parsers threw a panic
+	// Missing parameter to regexp parser
+	_, _, err = rvm.Parse(`topk(10,sum by(namespace)(count_over_time({application="nginx", site!="eu-west-1-dev"} |= "/artifactory/" != "api" != "binarystore" | regexp [1d])))`)
+	require.ErrorIs(t, err, logqlmodel.ErrParse)
+	// Empty parameter to regexp parser
+	_, _, err = rvm.Parse(`topk(10,sum by(namespace)(count_over_time({application="nginx", site!="eu-west-1-dev"} |= "/artifactory/" != "api" != "binarystore" | regexp ` + "``" + ` [1d])))`)
```
ssncferreira, line 1759: Does it also fail if we use other empty string with for instance `""` or `''`?

ssncferreira (commented):
LGTM 👍 just a couple of comments

pkg/logql/rangemapper_test.go
ssncferreira: And shouldn't we also add equivalent unit tests in `pkg/logql/syntax/ast_test.go`? 🤔

MichelHollands: Yes, it also fails with a panic with the other empty strings (I just ran the test again without the fixes).
Additional unit test were added as well.

PR #7927, opened 2022-12-13
WIP: Size based retention

### LINKED ISSUE #6876: Loki Block Disk Space Cleanup Proposal
## **Is your feature request related to a problem? Please describe.**

Currently, Loki users who are **not** using object-based storage may accidentally fill all available space on a disk or PVC with logs on block-based storage. This is especially common with Loki running in small adhoc k8s clusters like microk8s.

[Loki provides endpoints for deleting logs based on dates or queries](https://grafana.com/docs/loki/latest/api/#request-log-deletion), but no “automatic” method for ensuring that block storage remains within a defined threshold.

If block storage fills, Loki [will crash](https://github.com/grafana/loki/issues/2314) #2314  and require manual intervention. Several issues have been opened requesting a feature to manage this, as existing workarounds (cronjobs, scheduled runners, etc) require rebuilding indexes. (See [this](https://github.com/grafana/loki/issues/2314#issuecomment-1015374839) and [this](https://github.com/grafana/loki/issues/2314#issuecomment-1015562649))

### New config option: `size_based_retention_percentage`

Similar to [`–storage.tsdb.retention.size` in Prometheus,](https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects) it should be possible to programmatically configure a size based retention policy for Loki.

Based on the [complete local example](https://grafana.com/docs/loki/latest/configuration/examples/#complete-local-configyaml), a new option `size_based_retention_percentage` can be added:

```yaml
storage_config:
  boltdb:
    directory: /tmp/loki/index

  filesystem:
    directory: /tmp/loki/chunks
    size_based_retention_percentage: 80
```

### New argument: `-store.size_based_retention_percentage`

Loki binary has already a time based retention argument:

```bash
$ /usr/bin/loki -h | grep -A1 "store.retention"
  -store.retention value
        How long before chunks will be deleted from the store. (requires compactor retention enabled). (default 31d)
```

Following this example, the size_based_retention_percentage in the config file can be translated into `-store.size_based_retention_percentage`

### Possible Implementation

[Loki’s chunk-tool](https://github.com/grafana/loki/tree/main/cmd/chunks-inspect) already contains much of the logic necessary to determine the “real” storage usage from given chunks, and the datetimes for each. Extending and integrating this into an experimental argument to Loki which runs a Ticker. This Ticker can check the disk usage for `block` volumes, if present in the configuration, and if they exceed the interval, calculate the date/time ranges to delete via the endpoint to bring it back to a threshold.

This Timer could **potentially** expose a new metric on how fast disk usage is growing as well (`loki_chunk_store_bytes_minutes`), and emit logs from Loki itself if the log growth rate is sufficiently high to repeatedly trigger pruning.

Optionally, behavior similar to ceph (to determine how aggressive pruning should be) may be possible. 

This argument should be configurable to “fail” (and log errors or another mechanism) if it is in conflict with the `retention_period`.

#### Tentative deletion process

- Check disk usage
  - Re-calculate and update loki_chunk_store_bytes_minutes metric:
    - (actual_disk_usage - last_disk_usage) / minutes_between_reads
  - If disk usage is greater than the size_based_retention_percentage:
    - Calculate the amount of data (bytes) to be deleted:
      - (actual_disk_usage - size_based_retention_percentage) * disk_size
    - Iterate over chunks until the sum of the chunk's “Data length”  is greater than the amount of data to be deleted and get the “Through” value of the last chunk.
    - Perform a [log deletion](https://grafana.com/docs/loki/latest/api/#request-log-deletion) using that “Through” value as “end” parameter.

We thought about implementing an external process to delete old logs based on the size of the volume like  [this](https://github.com/grafana/loki/issues/2314#issuecomment-1015374839) and [this](https://github.com/grafana/loki/issues/2314#issuecomment-1015562649), but seems too hacky for a professional use case. 

**Additional context**

This is a WIP feature proposal, please feel free to share your thoughts!

Abuelodelanada:
## What this PR does / why we need it:

This `WorkInProgress` RP is the first approach to try to solve #6876.
I'm opening the PR at this stage to share the progress we made and fundamentally to read suggestions about it! 

## Which issue(s) this PR fixes:
Fixes #6876

## Special notes for your reviewer:

To activate the **size_based_retention** policy you have to add the `size_based_retention_percentage` in the `storage.filesystem` section:

```yaml
common:
  path_prefix: /tmp/loki
  storage:
    filesystem:
      chunks_directory: /tmp/loki/chunks
      rules_directory: /tmp/loki/rules
      size_based_retention_percentage: 2
```

Besides in the `compactor` section, the value for `retention_enabled` must be `true`, for instance:

```yaml
compactor:
  working_directory: /tmp/loki/retention
  shared_store: filesystem
  compaction_interval: 30s
  retention_enabled: true
  retention_delete_delay: 1m
```

## How to test:

1. Build loki: `make loki`
2. Let's say loki will store chunks in a filesystem with a disk usage near 2%, use a loki config like this:

```yaml
auth_enabled: false

server:
  http_listen_port: 3100
  grpc_listen_port: 9096

common:
  instance_addr: 127.0.0.1
  path_prefix: /tmp/loki
  storage:
    filesystem:
      chunks_directory: /tmp/loki/chunks
      rules_directory: /tmp/loki/rules
      size_based_retention_percentage: 2
  replication_factor: 1
  ring:
    kvstore:
      store: inmemory


storage_config:
  boltdb_shipper:
    active_index_directory: /tmp/loki/boltdb-shipper-active
    shared_store: filesystem
  filesystem:
    directory: /tmp/loki/chunks


schema_config:
  configs:
    - from: 2020-10-24
      store: boltdb-shipper
      object_store: filesystem
      schema: v11
      index:
        prefix: index_
        period: 24h

ingester:
  wal:
    enabled: true
    dir: /tmp/loki/chunks/wal
    flush_on_shutdown: true

compactor:
  working_directory: /tmp/loki/retention
  shared_store: filesystem
  compaction_interval: 30s
  retention_enabled: true
  retention_delete_delay: 1m

limits_config:
  ingestion_rate_mb: 10
```
3. Run loki: `./cmd/loki/loki -config.file=cmd/loki/loki-local-config.yaml`
4. Send logs to loki using for instance promtail.
5. Check that Loki actually receives the logs, for instance using Grafana:

image

6. When disk usage reaches  (or exceeds) `2%`  Loki will start deleting chunk files and you will see in the logs:

```
level=info ts=2023-04-14T17:17:59.614606507Z caller=retention.go:135 msg="Detected disk usage percentage" diskUsage=52.61%
level=info ts=2023-04-14T17:17:59.614636609Z caller=retention.go:156 msg="Block size retention exceeded, removing chunks from file" filepath=miguelito-1681483439590834149-1681491600.gz
level=info ts=2023-04-14T17:18:59.611874456Z caller=retention.go:135 msg="Detected disk usage percentage" diskUsage=52.60%
level=info ts=2023-04-14T17:18:59.614467894Z caller=retention.go:135 msg="Detected disk usage percentage" diskUsage=52.60%
level=info ts=2023-04-14T17:18:59.614573707Z caller=retention.go:156 msg="Block size retention exceeded, removing chunks from file" filepath=compactor-1681492684.gz
```
8. Loki will check for disk usage [every minute](https://github.com/grafana/loki/pull/7927/files#diff-8357ce3aacb6cace0adbbbb88de67c6808980aa2e7859697355571864b5057ddR35).

### Settings

Despite of the fact it is working seems the solution may be improved a lot, for instance we need 2 settings in the config file:

- `size_based_retention_percentage` in `storage.filesystem` section
- `retention_enabled: true` in `compactor` section.

I would love to avoid the need for `retention_enabled: true`.... 

### About the author

Since I'm mostly a Python/PHP Software Engineer writing my first lines in Go and it's my first contact with the Loki codebase, there are surely **many things** that can be improved. Please feel free to comment!! 😄

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


Abuelodelanada:
Hi @sandeepsukhani @owen-d

Please, about this PR, may I have your thoughts ?

MasslessParticle:
This is a good start @Abuelodelanada. I'm most concerned about the index here. When there are index entries for chunks that don't exist, Loki returns 500 errors so it's important that the index gets updated -- probably before the actual chunks are deleted. 

Looking back at the discussion, @sandeepsukhani suggested using the `indexProcessor.OpenCompactedFile` to get a `CompactedIndex` type. That type has the function `ForEachChunk` which allows you to iterate a delete chunks.

This approach has the advantage of updating the index while you find the chunks that need deleted.

To achieve the suggested approach, you'd need to:
1. Understand when you need to prune like your already do
2. Get the oldest index files
3. open them with `OpenCompactedFile`
4. Iterate and delete the chunks 
5. rewrite the index
6. Delete the chunks from disk

The existing retention code does everything from 1-5 as part of a "mark" phase and then cleans up all the found chunks as part of a delete phase

@sandeepsukhani, please let me know if I've missed anything

As far as your golang, I think it looks good! The only think I noticed is the pattern where you call a returned error `error` rather than `err`. It'd be better to use `err`. It's convention/idiomatic and it doesn't collide with the builtin `error` interface.

Abuelodelanada:
Hello @MasslessParticle thank you so much for the review!


About the index, and the steps you mentioned let me go one by one to understand if we are on the same page:

> 1. Understand when you need to prune like your already 

I think then we are ok [doing that here](https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/indexshipper/compactor/retention/retention.go#L107)

> 2. Get the oldest index files

At this point it's not clear to me how to get those index files, that's why in the PR we are [sorting and deleting the chunk files calling the new function `DeleteChunksBasedOnBlockSize`](https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/indexshipper/compactor/retention/deletion.go#L14). Our approach here was: _"Let's delete the oldest chunks and then re build the index"_

Do you have an example of how to do this?


> 3. open them with OpenCompactedFile

Do you mean [`OpenCompactedIndexFile` ?](https://github.com/grafana/loki/blob/3170edf3e3cf9cc354c41cf01556eb6024016d4f/pkg/storage/stores/indexshipper/compactor/table.go#L49). If so, [is this a good example](https://github.com/grafana/loki/blob/3170edf3e3cf9cc354c41cf01556eb6024016d4f/pkg/storage/stores/indexshipper/compactor/table.go#L256) about how to use it?

> 4. Iterate and delete the chunks

As far as I understand I should use [`ChunkIterator`](https://github.com/grafana/loki/blob/3170edf3e3cf9cc354c41cf01556eb6024016d4f/pkg/storage/stores/indexshipper/compactor/retention/retention.go#L47)

> 5. rewrite the index

I have no idea about how to do this.

> 6. Delete the chunks from disk

Is it enough just deleting the files from disk with `os.Remove`?

MasslessParticle:
> Do you mean [OpenCompactedIndexFile ?](https://github.com/grafana/loki/blob/3170edf3e3cf9cc354c41cf01556eb6024016d4f/pkg/storage/stores/indexshipper/compactor/table.go#L49). If so, [is this a good example](https://github.com/grafana/loki/blob/3170edf3e3cf9cc354c41cf01556eb6024016d4f/pkg/storage/stores/indexshipper/compactor/table.go#L256) about how to use it?

Yes! Sorry for the typo. The return type of `OpenCompactedIndexFile` implements `ChunkIterator` so you can use that to iterate actual chunks. 

Loki's index is made of tables that themselves have index files.

To get the index sets so you can open them:

- To get the tables: https://github.com/grafana/loki/blob/d97aaca08941430ec878b4a73662c9eaad2dbca1/pkg/storage/stores/indexshipper/compactor/compactor.go#L569
- To sort the tables by time: https://github.com/grafana/loki/blob/65bef5c360b31f091523e8fcb59b89b8b2cf6025/pkg/storage/stores/indexshipper/compactor/compactor.go#L718
- To get the index files from the tables: https://github.com/grafana/loki/blob/d97aaca08941430ec878b4a73662c9eaad2dbca1/pkg/storage/stores/indexshipper/compactor/table.go#L119
- Finally, to turn that into an `indexSet`: https://github.com/grafana/loki/blob/d97aaca08941430ec878b4a73662c9eaad2dbca1/pkg/storage/stores/indexshipper/compactor/table.go#L143
- `ForEachChunk` takes a callback that indicates whether the chunk is deleted. The `compactedIndex` keeps track of this. This is where you'll also keep track of which chunks you're going to delete from storage.
- If you delete all the chunks in a `compactedIndex`, you need to call `CleanupSeries` https://github.com/grafana/loki/blob/65bef5c360b31f091523e8fcb59b89b8b2cf6025/pkg/storage/stores/indexshipper/compactor/retention/retention.go#L246
- Based on what's returned from your iterations, you'll need to decide what to do with your table. Look to https://github.com/grafana/loki/blob/65bef5c360b31f091523e8fcb59b89b8b2cf6025/pkg/storage/stores/indexshipper/compactor/index_set.go#L164-L183 for an example of what to do if you delete the whole table
- Finally, you'll want to call `done` on your `index_set`: https://github.com/grafana/loki/blob/65bef5c360b31f091523e8fcb59b89b8b2cf6025/pkg/storage/stores/indexshipper/compactor/index_set.go#L285 which actually handles the index file uploads/drops

It's worth studying this `runRetention` function linked above as it implements all the same index operations you'll need to

Abuelodelanada:
Hi @MasslessParticle, thanks! 

I have some doubts about how to get the `ChunkIterator` with the implementation I have right now.

I mean, I [initialize the `sizeBasedRetention` in the compactor](https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/indexshipper/compactor/compactor.go#L275), then if everything is ok, the service is started in the [`compactor.starting` function](https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/indexshipper/compactor/compactor.go#L333). This `TimerService` is executed [every minute](https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/indexshipper/compactor/retention/retention.go#L33)

Right now, the method that deletes the chunks is [`RunIteration`](https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/indexshipper/compactor/retention/retention.go#L107)... so, how would you iterate the chunks at this point using `ChunkIterator`?, As far as I understand I cannot get `OpenCompactedIndexFile` (`t.indexCompactor.OpenCompactedIndexFile(...`) there

MasslessParticle:
@Abuelodelanada It looks like I missed the step of turning an `indexSet` into a `CompactedIndex`, sorry about that.

`CompactedIndex` implements `ForEachChunk`. This is the current code that converts `indexSet -> CompactedIndex` https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/shipper/index/compactor/table_compactor.go#L445

This function is complex because it does actual compaction but you'll only need the parts necessary to get a valid `CompactedIndex` type

Abuelodelanada:
Thanks @MasslessParticle !

I feel I have a gap in the knowledge between what I have said in my last comment:


> I mean, I [initialize the `sizeBasedRetention` in the compactor](https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/indexshipper/compactor/compactor.go#L275), then if everything is ok, the service is started in the [`compactor.starting` function](https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/indexshipper/compactor/compactor.go#L333). This `TimerService` is executed [every minute](https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/indexshipper/compactor/retention/retention.go#L33)
>
> Right now, the method that deletes the chunks is [`RunIteration`](https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/indexshipper/compactor/retention/retention.go#L107)... so, how would you iterate the chunks at this point using `ChunkIterator`?, As far as I understand I cannot get `OpenCompactedIndexFile` (`t.indexCompactor.OpenCompactedIndexFile(...`) there 

And what you have said in your last comment:


> `CompactedIndex` implements `ForEachChunk`. This is the current code that converts `indexSet -> CompactedIndex`
> 
> https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/shipper/index/compactor/table_compactor.go#L445
> 
> This function is complex because it does actual compaction but you'll only need the parts necessary to get a valid `CompactedIndex` type

Sorry but it is not obvious to me how to do this at [`RunIteration`](https://github.com/grafana/loki/blob/ea68bd8785167133e2437c136f17e0603e2efb7a/pkg/storage/stores/indexshipper/compactor/retention/retention.go#L107)...

MasslessParticle:
Hello @Abuelodelanada! I'm working up some detailed feedback for how to accomplish this but I've run out of time this week. I hope to have it out early next week.

Abuelodelanada:
Hi @MasslessParticle thanks!

There is no rush, I will be AFK (on holiday) for the next two weeks.

Abuelodelanada:
Hi @MasslessParticle 

Have you been able to make some progress about how to accomplish what I have asked in [this comment](https://github.com/grafana/loki/pull/7927#issuecomment-1402085592)?

MasslessParticle:
@Abuelodelanada Hello! Sorry for the delay.

I'm made some feedback in the form of this Draft PR: https://github.com/grafana/loki/pull/8581

LMKWYT

Abuelodelanada:
Hi @MasslessParticle 

After some weeks of radio silence (we were working hard), we return to this PR.

We are at a point this is kind of working as expected, using a config file like this one:

```yaml
auth_enabled: false

server:
  http_listen_port: 3100
  grpc_listen_port: 9096

common:
  instance_addr: 127.0.0.1
  path_prefix: /tmp/loki
  storage:
    filesystem:
      chunks_directory: /tmp/loki/chunks
      rules_directory: /tmp/loki/rules
      size_based_retention_percentage: 2
  replication_factor: 1
  ring:
    kvstore:
      store: inmemory


storage_config:
  boltdb_shipper:
    active_index_directory: /tmp/loki/boltdb-shipper-active
    shared_store: filesystem
  filesystem:
    directory: /tmp/loki/chunks


schema_config:
  configs:
    - from: 2020-10-24
      store: boltdb-shipper
      object_store: filesystem
      schema: v11
      index:
        prefix: index_
        period: 24h

ingester:
  wal:
    enabled: true
    dir: /tmp/loki/chunks/wal
    flush_on_shutdown: true

compactor:
  working_directory: /tmp/loki/retention
  shared_store: filesystem
  compaction_interval: 30s
  retention_enabled: true
  retention_delete_delay: 1m

limits_config:
  ingestion_rate_mb: 10

```

Note the `size_based_retention_percentage` under `common` section and `retention_enabled: true` under compactor.

May you give it a try and let us known your thoughts?

MasslessParticle:
@Abuelodelanada 🎉

I'm a buried at the moment but I will test when I get the chance!

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -401,9 +443,114 @@ func (c *Compactor) starting(ctx context.Context) (err error) {
...
 
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
+func (c *Compactor) sizeBasedCompactionInterval(ctx context.Context) error {
+	if exceeded, err := c.sizeBasedRetention.ThresholdExceeded(); !exceeded {
```
MasslessParticle, line 454: This doesn't handle the error when `exceeded == true`

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -149,18 +151,19 @@ type Compactor struct {
...
 	DeleteRequestsHandler     *deletion.DeleteRequestHandler
 	DeleteRequestsGRPCHandler *deletion.GRPCRequestHandler
 	deleteRequestsManager     *deletion.DeleteRequestsManager
-	expirationChecker         retention.ExpirationChecker
-	metrics                   *metrics
-	running                   bool
-	wg                        sync.WaitGroup
-	indexCompactors           map[string]IndexCompactor
-	schemaConfig              config.SchemaConfig
+	//expirationChecker         retention.ExpirationChecker
```
MasslessParticle, line 159: Should delete this

pkg/storage/stores/indexshipper/compactor/retention/retention.go
```diff
@@ -90,7 +227,17 @@ type Marker struct {
 	markTimeout      time.Duration
 }
 
+var mu sync.Mutex
```
MasslessParticle, line 230: I don't think we need the mutex here. The compactor should probably own the marker metrics and pass it to markers.

MasslessParticle (commented):
Sorry again for the delay to look at this. I can't run it because things have drifted but I have done a review.

This is looking good and makes sense. I've added comments about a couple of concerns. I'm also concerned there there's a lot of logic here and very few tests. I notice that you merged my original suggestions. That's fair to avoid rework but also those were illustrative and should probably be tested

MasslessParticle (commented):
Sorry again for the delay to look at this. I can't run it because things have drifted but I have done a review.

This is looking good and makes sense. I've added comments about a couple of concerns. I'm also concerned there there's a lot of logic here and very few tests. I notice that you merged my original suggestions. That's fair to avoid rework but also those were illustrative and should probably be tested

pkg/storage/stores/indexshipper/compactor/compactor.go
Abuelodelanada: Hi @MasslessParticle 

We are not handling this situation, because [the function `ThresholdExceeded()`](
https://github.com/grafana/loki/pull/7927/files#diff-8357ce3aacb6cace0adbbbb88de67c6808980aa2e7859697355571864b5057ddR124-R134) does not return this combination: `true, err`. The possible returns are:

- `false, err`
- `false, nil`
- `true, nil`

Do you think we should modify this anyway?

MasslessParticle: For uniform error handling, I think it would be best to modify this to handle `true, err`. This guards against future changes to `ThresholdExceeded`

pkg/loki/modules.go
```diff
@@ -1076,6 +1076,11 @@ func (t *Loki) initCompactor() (services.Service, error) {
 	// Set some config sections from other config sections in the config struct
 	t.Cfg.CompactorConfig.CompactorRing.ListenPort = t.Cfg.Server.GRPCListenPort
 
+	if !(config.UsingObjectStorageIndex(t.Cfg.SchemaConfig.Configs) || config.UsingBoltDBtorageIndex(t.Cfg.SchemaConfig.Configs)) {
```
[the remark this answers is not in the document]
Abuelodelanada: You are right! This was left here by mistake.

pkg/storage/stores/indexshipper/compactor/compactor.go
Abuelodelanada: Reviewing the code again seems that there is no point in checking all the combinations. 

We need just 2 guards:

- Log a message and `return nil` if there is an error, or
- `return nil` if doesn't exceed threshold.

In other situations we should continue with the execution.

Pushing these changes with less nesting, let me know your thoughts.

Abuelodelanada:
> Sorry again for the delay to look at this. I can't run it because things have drifted but I have done a review.
> 
> This is looking good and makes sense. I've added comments about a couple of concerns. I'm also concerned there there's a lot of logic here and very few tests. I notice that you merged my original suggestions. That's fair to avoid rework but also those were illustrative and should probably be tested

I'm working on the test right now.

pkg/storage/stores/indexshipper/compactor/compactor.go
MasslessParticle: This makes sense. Do we return `nil` on error to not exit compactions? That makes sense, but we should probably add a metric here so an alert can be raised on successive failures.

pkg/storage/stores/indexshipper/compactor/retention/retention.go
Abuelodelanada: I did not implement this.
This section of the code was written by @rbarry82 who sadly passed away a month ago.

Do you mean something like this?

```golang
func NewMarker(workingDirectory string, expiration ExpirationChecker, markTimeout time.Duration, chunkClient client.Client, metrics *markerMetrics) (*Marker, error) {
	return &Marker{
		workingDirectory: workingDirectory,
		expiration:       expiration,
		markerMetrics:    metrics,
		chunkClient:      chunkClient,
		markTimeout:      markTimeout,
	}, nil
}
```

MasslessParticle: I'm sorry to hear about the loss of your colleague :slightly_frowning_face: 

Metrics can only be registered once. If we try to register them a second time, a runtime panic occurs. The code, as is prevents registering a second time but it's a little akward.

I think we could change
```go
func NewMarker(workingDirectory string, expiration ExpirationChecker, markTimeout time.Duration, chunkClient client.Client, r prometheus.Registerer) (*Marker, error)
```
to take metrics like this
```
func NewMarker(workingDirectory string, expiration ExpirationChecker, markTimeout time.Duration, chunkClient client.Client, metrics *markerMetrics) (*Marker, error)
```

But then we need a place to instantiate the metrics. We could probably do it in the compactor's `init` function and make it a field on the compactor.

If we do this, the mutex can be deleted altogether.

Abuelodelanada: About mutex, now I remember that Ryan added [the following to `Compactor` struct:](https://github.com/grafana/loki/blob/69f78bee44ba8d2f5a223243a2fdd2b100cad692/pkg/storage/stores/indexshipper/compactor/compactor.go#L182)

```golang
	// Size based compaction means that two compactions might try to happen at the same time.
	// Use this to ensure size-based and normal compaction can't step on eachother.
	compactionMtx sync.Mutex
```

Does make sense o he was wrong?



About your suggestion, I have this draft right now (didn't push it yet)... Is it close to what you suggest?


```diff
modified   pkg/storage/stores/indexshipper/compactor/compactor.go
@@ -182,6 +182,7 @@ type Compactor struct {
 	// Size based compaction means that two compactions might try to happen at the same time.
 	// Use this to ensure size-based and normal compaction can't step on eachother.
 	compactionMtx sync.Mutex
+	markerMetrics *retention.markerMetrics
 
 	// one for each object store
 	storeContainers map[string]storeContainer
@@ -310,6 +311,8 @@ func (c *Compactor) init(objectStoreClients map[string]client.ObjectClient, sche
 		}
 	}
 
+	c.markerMetrics = retention.newMarkerMetrics(r)
+
 	c.storeContainers = make(map[string]storeContainer, len(objectStoreClients))
 	for objectStoreType, objectClient := range objectStoreClients {
 		var sc storeContainer
@@ -338,7 +341,7 @@ func (c *Compactor) init(objectStoreClients map[string]client.ObjectClient, sche
 				return fmt.Errorf("failed to init sweeper: %w", err)
 			}
 
-			sc.tableMarker, err = retention.NewMarker(retentionWorkDir, c.expirationChecker, c.cfg.RetentionTableTimeout, chunkClient, r)
+			sc.tableMarker, err = retention.NewMarker(retentionWorkDir, c.expirationChecker, c.cfg.RetentionTableTimeout, chunkClient, c.markerMetrics)
 			if err != nil {
 				return fmt.Errorf("failed to init table marker: %w", err)
 			}
modified   pkg/storage/stores/indexshipper/compactor/retention/retention.go
@@ -5,9 +5,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync"
 	"os"
 	"path/filepath"
+	"sync"
 
 	"time"
 
@@ -227,21 +227,11 @@ type Marker struct {
 	markTimeout      time.Duration
 }
 
-var mu sync.Mutex
-var metrics *markerMetrics
-
-func NewMarker(workingDirectory string, expiration ExpirationChecker, markTimeout time.Duration, chunkClient client.Client, r prometheus.Registerer) (*Marker, error) {
-	mu.Lock()
-	defer mu.Unlock()
-
-	if metrics == nil {
-		metrics = newMarkerMetrics(r)
-	}
-
+func NewMarker(workingDirectory string, expiration ExpirationChecker, markTimeout time.Duration, chunkClient client.Client, metrics *markerMetrics) (*Marker, error) {
 	return &Marker{
 		workingDirectory: workingDirectory,
 		expiration:       expiration,
-		markerMetrics:    newMarkerMetrics(r),
+		markerMetrics:    metrics,
 		chunkClient:      chunkClient,
 		markTimeout:      markTimeout,
 	}, nil
```

simskij:
This is taking a lot longer than we expected as we've bumped into quite a lot of issues with the delta between when we originally forked and the current state. In combination with @rbarry82's unfortunate passing, things are going slower than usual. Work is however still ongoing and we're still looking to get this across the finish line.

VenkateswaranJ:
Any update on this?

simskij:
> Any update on this?

Some, but nothing substantial. I'll get back to you all at the beginning of next year with more details.

dmitriy-kharchenko:
Folks, this feature is very needed for World! Let's fix conflicts and merge it!

mlladb:
Any update on this? I've run into the situation twice this week where a sudden spike in logs filled up disks, causing backlog pressure and instability in Loki-write pods. Having log retention based on disk usage would be a great. Without it I need to ssh onto the worker node and free up some space manually.

Etalon-Mikhail-Tymchuk:
Can't believe this is not implemented in 2026.

JStickler:
@Abuelodelanada have you abandoned this PR?  Should we close it and let someone else attempt this fix?

JStickler:
Closing as abandoned and one contributor has not signed the CLA.

PR #7942, opened 2022-12-14
[Helm] Allow single binary with Object Storage

jkroepke:

Allow run loki in single binary mode with object storage. See also https://grafana.com/docs/loki/latest/fundamentals/architecture/deployment-modes/

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/installation/helm/reference.md
+++ b/docs/sources/installation/helm/reference.md
@@ -235,6 +235,15 @@ null
 </td>
 		</tr>
 		<tr>
+			<td>deployment.singleBinary</td>
+			<td>string</td>
+			<td></td>
+			<td><pre lang="json">
+""
+</pre>
+</td>
+		</tr>
+		<tr>
 			<td>enterprise.adminApi</td>
 			<td>object</td>
 			<td>If enabled, the correct admin_client storage will be configured. If disabled while running enterprise, make sure auth is set to `type: trust`, or that `auth_enabled` is set to `false`.</td>
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -26,14 +26,24 @@ singleBinary fullname
 Return if deployment mode is simple scalable
 */}}
 {{- define "loki.deployment.isScalable" -}}
+{{- $singleBinary := (.Values.deployment.singleBinary | toString) }}
+{{- if ne $singleBinary "" }}
+  {{- ne $singleBinary "true" }}
+{{- else }}
   {{- eq (include "loki.isUsingObjectStorage" . ) "true" }}
+{{- end -}}
 {{- end -}}
 
 {{/*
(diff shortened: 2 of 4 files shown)


trevorwhitney (changes requested):
So we originally intentionally didn't expose this option to simplify configurations. The thought was if you're running single binary, you're only running 1, and therefore you don't need object storage. Can you describe your use case a little more?

jkroepke:
object storage is always preferred over local disk. 

less costs, redundant by default, can not run out of space.

from my point of view, object storage should be always preferred over local disk. If object storage is supported, there is no reason to use a local disk

jeschkies:
@trevorwhitney I agree with @jkroepke. I'm just Eldon how it should be implemented.

jkroepke:
Rebased

djotanov:
We need this as well. I don't know why single binary option was disabled for object storage in this refactored helm chart. Main reason we can't upgrade to this new chart.

trevorwhitney:
I agree this configuration should be possible, however I am reticent to expose `deployment` type as a top level config. Loki has always had the problem of very complicated configs that force operators to have extensive knowledge about how Loki is implemented in order to run it. One of the major goals of this recent helm investment was to provide a way to run Loki that only required the operator to describe their system (ie. how big, where do we store things) and the chart would provide a recommended configuration for that scenario.

My understanding here is that there is a desire to run a Loki that is HA/replicated but for a log volume that does not merit 3 read and 3 write pods. I think there is a way to express/infer this through the values without requiring the operator to pick between single-binary and scalable. I have some ideas, but need some time to put a PR together.

Thanks to everyone who's participated in this thread. I'm sure you are all experienced users of Loki, so this distinction may seem trivial to you, as you're already familiar with the various deployment configurations. However, I want to make sure this chart remains accessible to those who have never used Loki before, and who don't know about nor care about the different deployment configurations.

jeschkies:
> What is Eldon?

@jkroepke that was some odd auto correct. 

I might I'm wondering how it would be implemented.

jkroepke:
Current workaround: To use this helm chart and run a single instance loki with an blob storage, you have to create an umbrella chart and add this helm chart as dependency. 

Create a _helpers.tpl with following content:

```
{{- define "loki.deployment.isScalable" -}}
false
{{- end -}}

{{- define "loki.deployment.isSingleBinary" -}}
true
{{- end -}}
```

trevorwhitney:
I have created an [issue](https://github.com/grafana/loki/issues/8185) to track this and discuss options. At the very least I agree this should be possible (even if it means overriding the whole config file), but I'm not ready to merge this solution. Let's move the discussion to the [issue](https://github.com/grafana/loki/issues/8185).

trevorwhitney:
https://github.com/grafana/loki/pull/8284

PR #7946, opened 2022-12-15
config_wrapper: Add support for named stores

### LINKED ISSUE #7276: Allow for multiple store configurations from the same provider

ashwanthgoli:

This PR adds support for named stores which allows users to define multiple object store configurations for a storage type. Named stores can be referred to from other sections of the config similar to how provider type is being used `i.e aws, gcs...`. Named store reference should be of the format: `<store_type>.<key_name>`

Below is an example configuration with two named stores defined for aws `store-1, store-2`. Schema config is referring to one of them as `aws.store-1`
```
storage_config:
  aws:
    endpoint: s3://common-bucket
    region: us-east1
    access_key_id: abc123
    secret_access_key: def789
  named_stores:
    aws:
      store-1:
        endpoint: s3://foo-bucket
        region: us-west1
        access_key_id: 123abc
        secret_access_key: 789def
      store-2:
        endpoint: s3://bar-bucket
        region: us-west2
        access_key_id: 456def
        secret_access_key: 789abc
schema_config:
    configs:
        - from: "2020-07-31"
            index:
              period: 24h
              prefix: loki_prod_index_
            object_store: aws.store-1
            schema: v11
            store: boltdb-shipper
```
Fixes #7276

Configuration defined in common storage config has no effect on the named stores. Common storage config only overrides storageConfig if it's not explicitly set - same behavior as before. This should be a non-intrusive change.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


tools/doc-generator/parse/root_blocks.go
```diff
@@ -217,8 +222,20 @@ var (
 		},
 		{
 			Name:       "local_storage_config",
-			StructType: reflect.TypeOf(local.Config{}),
+			StructType: reflect.TypeOf(local.FSConfig{}),
```
ssncferreira, line 225: Good catch ⭐

docs/sources/configuration/_index.md
```diff
@@ -1817,6 +1652,18 @@ hedging:
...
+# Configures additional object stores for a given storage provider.
+# Example:
+# storage_config:
+#   named_stores:
+#     aws:
+#       store-1:
+#         endpoint: s3://foo-bucket
+#         region: us-west1
+# Named store from this example can be used by setting object_store to
+# aws.store-1 in period_config.
```
ssncferreira, line 1664: nit: add a lit of the supported providers

ssncferreira (commented):
LGTM 👍 and great job updating the `_index.md` file with the storage configuration flags 🏆

pkg/storage/factory.go
```diff
@@ -134,6 +145,29 @@ func (cfg *Config) Validate() error {
...
 		return errors.Wrap(err, "invalid tsdb config")
 	}
+
+	return cfg.validateNamedStores()
+}
+
+func (cfg *Config) validateNamedStores() error {
+	for name, cfg := range cfg.NamedStores.AWS {
+		if err := cfg.Validate(); err != nil {
+			return errors.Wrap(err, fmt.Sprintf("invalid AWS Storage config with name %s", name))
```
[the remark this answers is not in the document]
ashwanthgoli: rest of them don't expose a `Validate()` method

pkg/storage/factory.go
```diff
@@ -331,21 +374,86 @@ func (c *ClientMetrics) Unregister() {
...
+			if !ok {
+				return nil, fmt.Errorf("Unrecognized named azure storage config %s", name)
+			}
+		}
+
+		return azure.NewBlobStorage(&azureCfg, clientMetrics.AzureMetrics, cfg.Hedging)
 	case config.StorageTypeSwift:
-		return openstack.NewSwiftObjectClient(cfg.Swift, cfg.Hedging)
+		swiftCfg := cfg.Swift
+		if namedStore != "" && cfg.NamedStores.Swift != nil {
```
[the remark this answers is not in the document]
ashwanthgoli: I forgot to remove this check. `nil` maps are safe to read, so we would throw an error if the namedStore is not found

pkg/storage/factory.go
```diff
@@ -331,21 +374,86 @@ func (c *ClientMetrics) Unregister() {
...
-	switch name {
+	var (
+		namedStore string
+		storeType  = name
+	)
+
+	// extact storeType and lookup key for named stores
+	if words := strings.SplitN(name, ".", 2); len(words) == 2 {
+		storeType, namedStore = words[0], words[1]
+	}
```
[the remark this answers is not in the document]
ashwanthgoli: `NewObjectClient` has references from other packages, so I am thinking to use the same contract as before, otherwise the callers have to perform the split to make the call.

pkg/storage/factory.go
```diff
@@ -134,6 +145,29 @@ func (cfg *Config) Validate() error {
 	if err := cfg.TSDBShipperConfig.Validate(); err != nil {
 		return errors.Wrap(err, "invalid tsdb config")
 	}
+
+	return cfg.validateNamedStores()
+}
+
+func (cfg *Config) validateNamedStores() error {
```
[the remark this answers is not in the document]
ashwanthgoli: other stores don't expose a `Validate` method

pkg/loki/config_wrapper_test.go
```diff
@@ -579,6 +579,46 @@ storage_config:
...
+        access_key_id: 123abc
+        secret_access_key: 789def
+      store-2:
+        endpoint: s3://bar-bucket
+        region: us-west2
+        access_key_id: 456def
+        secret_access_key: 789abc`
+			config, _ := testContext(namedStoresConfig, nil)
+
+			// should be set by common config
```
ssncferreira, line 605: In this case, since we defined storage in 1) common config, and in 2) named storage, the common config will be ignored and the named storage will be the one used, right? Any way we can assert that in the test?

pkg/storage/factory_test.go
```diff
@@ -45,6 +50,60 @@ func TestFactoryStop(t *testing.T) {
...
+					Period: time.Hour * 168,
+				},
+			},
+		},
+	}
+
+	limits, err := validation.NewOverrides(validation.Limits{}, nil)
+	require.NoError(t, err)
+
+	t.Run("refer to named store", func(t *testing.T) {
```
ssncferreira, line 91: Not sure about this test name and the following. Maybe 
```suggestion
	t.Run("period config referring to configured named store", func(t *testing.T) {
```
wdyt?

ssncferreira, line 91: Not sure about this test name and the following. Maybe 
```suggestion
	t.Run("period config referring to configured named store", func(t *testing.T) {
```
wdyt?

ssncferreira (approved):
LGTM 🌟 thanks for addressing the comments. Just adding some minor comments about the unit tests 🙂

pkg/loki/config_wrapper_test.go
ashwanthgoli: common config won't be ignored as it overrides StorageConfig (if it's not configured). And users can choose to use one or both by setting objectType across periodConfigs.
`aws` for using configuration from StorageConfig (which actually comes from common config)
`aws.named-store-key-name` for using the named store

pkg/storage/factory_test.go
ashwanthgoli: 👍

pkg/loki/config_wrapper_test.go
ssncferreira: Ah got it 👍 not necessary for this PR, but maybe it would be useful to add this information to period_config, that one can set it via the StorageConfig or via named_stored

ssncferreira (approved):
LGTM 🏆  the `check-doc` CI step is failing, so you need to re-run `make doc` to update the _index.md file

pkg/storage/factory.go
```diff
@@ -198,11 +232,20 @@ func NewIndexClient(name string, cfg Config, schemaCfg config.SchemaConfig, limi
...
 func NewChunkClient(name string, cfg Config, schemaCfg config.SchemaConfig, clientMetrics ClientMetrics, registerer prometheus.Registerer) (client.Client, error) {
-	switch name {
+	var (
+		storeType = name
+	)
+
+	// extact storeType for named stores
+	if words := strings.SplitN(name, ".", 2); len(words) == 2 {
+		storeType = words[0]
+	}
```
dannykopping, line 242: I think this should be its own function with a test associated

tools/doc-generator/parse/parser.go
```diff
@@ -375,11 +375,15 @@ func getFieldCustomType(t reflect.Type) (string, bool) {
 	}
 }
 
-func getFieldType(t reflect.Type) (string, error) {
+func getFieldType(t reflect.Type, rootBlocks []RootBlock) (string, error) {
```
dannykopping, line 378: nit: in future, please try keep unrelated changes in a separate PR

But! Kudos for fixing this :1st_place_medal:

pkg/storage/factory_test.go
```diff
@@ -45,6 +50,60 @@ func TestFactoryStop(t *testing.T) {
 	store.Stop()
 }
 
+func TestNamedStores(t *testing.T) {
```
dannykopping, line 53: Maybe I missed it, but I was expecting to find a test that shows that named and non-named stores can be used together?
Something like "period A is using the default GCS bucket, but period B is using a named store with a different bucket also in GCS"

pkg/storage/factory_test.go
```diff
@@ -45,6 +50,60 @@ func TestFactoryStop(t *testing.T) {
...
+			},
+		},
+	}
+
+	limits, err := validation.NewOverrides(validation.Limits{}, nil)
+	require.NoError(t, err)
+
+	t.Run("period config referring to configured named store", func(t *testing.T) {
+		store, err := NewStore(cfg, config.ChunkStoreConfig{}, schemaConfig, limits, cm, nil, util_log.Logger)
+		require.NoError(t, err)
```
dannykopping, line 93: I'm not sure just testing for the absense of an error is sufficient here. I think we should be validating that the expected config was selected.

dannykopping (commented):
Looks great! I think the tests are insufficient though

pkg/storage/factory.go
```diff
@@ -198,11 +232,20 @@ func NewIndexClient(name string, cfg Config, schemaCfg config.SchemaConfig, limi
 
 // NewChunkClient makes a new chunk.Client of the desired types.
 func NewChunkClient(name string, cfg Config, schemaCfg config.SchemaConfig, clientMetrics ClientMetrics, registerer prometheus.Registerer) (client.Client, error) {
-	switch name {
+	var (
+		storeType = name
+	)
+
+	// extact storeType for named stores
+	if words := strings.SplitN(name, ".", 2); len(words) == 2 {
```
owen-d, line 240: Rather than requiring we use `<type>.<name>` as a convention in the configuration, what if we allowed specifying only the store's name and resolved the type from that. As a bonus, this would also work with stores that used the reserved `.` characters, such as `name: "my.s3.compatible.store"`. I think this would be easier to use and understand (I had to check test cases to realize the convention of putting the `.` in the store name to look it up). Note: if we made this change, all the individual types would be reserved store names (for example, you couldn't redefine another store with name `aws`).

pkg/storage/factory_test.go
ashwanthgoli: cannot access the internals of store as it is private to a different pkg. And I feel mocking might not be an option here since the client init happens a few layers below.

I am thinking of adding an integration test: push a few log lines and ensure that the chunks and index are stored in the dir configured for the filesystem named store. wdyt?

pkg/storage/factory_test.go
ashwanthgoli: problem is when using boltdb or tsdb indexes, current code only intializes a single instance based on `sharedStoreType` or the objectType from the latest period config.

Next [PR](https://github.com/grafana/loki/pull/7754) will remove this limitation and adds tests with two stores [1](https://github.com/grafana/loki/pull/7754/files#diff-d383bd6a92c29d4c84d56fdd2ac42de2991fd4c2c38eb0a380e1c5d8099fa946) [2](https://github.com/grafana/loki/pull/7754/files#diff-556b2c27e0ba1456d41699b8a0f631c10bce22f01d15b860f87a8ce9b3a7ef16)

dannykopping: OK cool :+1:

pkg/storage/factory.go
ashwanthgoli: I was also a bit worried that this convention might not be the most intuitive.
Just to make sure I understand it correctly: we want to drop the type prefix when referring to the named store, but the way we are configuring names stores can remain the same right?

pkg/storage/factory.go
```diff
@@ -65,6 +71,98 @@ type NamedStores struct {
...
+			switch name {
+			case config.StorageTypeAWS, config.StorageTypeAWSDynamo, config.StorageTypeS3,
+				config.StorageTypeGCP, config.StorageTypeGCPColumnKey, config.StorageTypeBigTable, config.StorageTypeBigTableHashed, config.StorageTypeGCS,
+				config.StorageTypeAzure, config.StorageTypeBOS, config.StorageTypeSwift, config.StorageTypeCassandra,
+				config.StorageTypeFileSystem, config.StorageTypeInMemory, config.StorageTypeGrpc:
+				return fmt.Errorf("named store %s should not match with the name of a predefined storage type", name)
+			}
+
+			if objectType, ok := seen[name]; ok {
+				return fmt.Errorf("named store %s is already defined under %s", name, objectType)
```
dannykopping, line 109: What value will `objectType` typically have? It's excluded from the test, and I'm not sure what it'll be from looking at this implementation

pkg/storage/factory.go
```diff
@@ -65,6 +71,98 @@ type NamedStores struct {
...
+			}
+			seen[name] = objectType
+		}
+	}
+
+	return nil
+}
+
+func (ns NamedStores) lookupStoreType(name string) (string, bool) {
```
dannykopping, line 118: This feels a bit weird to me; why are we looking it up in the YAML rather than capturing this type when we instantiate the named store?

pkg/storage/factory.go
```diff
@@ -65,6 +71,98 @@ type NamedStores struct {
...
+		if v := v.FieldByIndex(field.Index).MapIndex(reflect.ValueOf(name)); v.IsValid() {
+			return objectType, true
+		}
+	}
+
+	return "", false
+}
+
+func (ns NamedStores) validate() error {
+	for name, awsCfg := range ns.AWS {
```
dannykopping, line 145: Howcome we're only validating these 3, out of interest?

pkg/storage/factory_test.go
```diff
@@ -50,23 +54,65 @@ func TestFactoryStop(t *testing.T) {
 	store.Stop()
 }
 
+func TestCassandraInMultipleSchemas(t *testing.T) {
```
dannykopping, line 57: I'm curious why Cassandra is being tested specifically here?

pkg/storage/factory_test.go
```diff
@@ -91,56 +137,102 @@ func TestNamedStores(t *testing.T) {
 	t.Run("period config referring to configured named store", func(t *testing.T) {
 		store, err := NewStore(cfg, config.ChunkStoreConfig{}, schemaConfig, limits, cm, nil, util_log.Logger)
 		require.NoError(t, err)
+		defer store.Stop()
 
-		store.Stop()
+		// FSObjectClient creates the configured dir on init, ensure that correct cfg is picked by checking for it.
+		_, err = os.Stat(cfg.NamedStores.Filesystem["named-store"].Directory)
```
dannykopping, line 143: You should first check that this directory does not exist before instantiating the store, to account for false-positives.

pkg/storage/factory_test.go
```diff
@@ -91,56 +137,102 @@ func TestNamedStores(t *testing.T) {
...
+		require.NoError(t, err)
+
+		// dir specified in StorageConfig/FSConfig should not be created as we are not referring to it.
+		_, err = os.Stat(cfg.FSConfig.Directory)
+		require.Error(t, err)
```
dannykopping, line 148: This seems redundant

```suggestion
```

pkg/storage/factory.go
```diff
@@ -65,6 +71,98 @@ type NamedStores struct {
 	Swift      map[string]openstack.SwiftConfig     `yaml:"swift"`
 }
 
+func (ns NamedStores) checkForDuplicates() error {
```
dannykopping, line 74: Can you please explain the rationale for approaching validation in this way?
Wouldn't it be easier to just retrieve the config and scan through all period configs and find duplicates across store types? Or are you doing it this way to spot duplicates within the same type?

This seems quite complicated, and I'm curious what problems you ran into here.

pkg/storage/factory.go
```diff
@@ -65,6 +71,98 @@ type NamedStores struct {
...
+
+			name := key.Interface().(string)
+
+			// return error if the name matches a predefined storage type
+			switch name {
+			case config.StorageTypeAWS, config.StorageTypeAWSDynamo, config.StorageTypeS3,
+				config.StorageTypeGCP, config.StorageTypeGCPColumnKey, config.StorageTypeBigTable, config.StorageTypeBigTableHashed, config.StorageTypeGCS,
+				config.StorageTypeAzure, config.StorageTypeBOS, config.StorageTypeSwift, config.StorageTypeCassandra,
+				config.StorageTypeFileSystem, config.StorageTypeInMemory, config.StorageTypeGrpc:
+				return fmt.Errorf("named store %s should not match with the name of a predefined storage type", name)
```
dannykopping, line 105: nit: i'd suggest changing these from `%s` to `%q` so that error messages will be a little more clear:

> named store aws should not match...
vs
> named store "aws" should not match...

pkg/storage/factory_test.go
```diff
@@ -91,56 +137,102 @@ func TestNamedStores(t *testing.T) {
...
+	}
 
-	// build schema with cassandra in multiple periodic configs
-	schemaCfg := DefaultSchemaConfig("cassandra", "v1", model.Now().Add(-7*24*time.Hour))
-	newSchemaCfg := schemaCfg.Configs[0]
-	newSchemaCfg.Schema = "v2"
-	newSchemaCfg.From = config.DayTime{Time: model.Now()}
+	storeType, ok := ns.lookupStoreType("store-1")
+	assert.True(t, ok)
+	assert.Equal(t, "gcs", storeType)
```
dannykopping, line 207: Please use constants for these store type values

pkg/storage/factory.go
ashwanthgoli: only 3 out of the supported named stores have a validation method

pkg/storage/factory.go
ashwanthgoli: finding duplicates would require checking the keys from each of the maps defined under NamedStores. Doing it this way to avoid the amount of redundant code needed to lookup each field.
I could drop this for the simpler approach too.

pkg/storage/factory_test.go
ashwanthgoli: not sure, I didn't touch this
must have moved it when adding new tests

pkg/storage/factory.go
dannykopping: Simpler is usually better; it's less prone to bugs.
No need to change it now, just a suggestion for the future.

pkg/storage/factory.go
dannykopping: Again, I think it'd be far simpler to just keep a mapping in the `NamedStores` struct of identifier to types than doing all this fancy footwork.

I don't want to hold up this PR further, so we can leave it as-is, but in general I'd suggest erring on the side of simplicity and readability over conciseness; it'll work out better in the long run (IMHO).

pkg/storage/factory.go
dannykopping: Again, just a suggestion for the future:
It'd be cleaner to iterate over each config and see if it satisfies 

```golang
interface {
  Validate() error
}
```

than special-casing these 3 stores. It also has the benefit of expressing the intent more clearly (as opposed to this code which seems to only validate these 3 for some unknown reason).

dannykopping (approved):
LGTM, thanks for being patient with the review feedback and simplifying the approach!

sandeepsukhani (approved):
LGTM

PR #7951, opened 2022-12-16
Add count to template functions

MichelHollands:

Add a `count` template function to count occurrences of a substring. For example if `label="abc abd abc"` 

```
{{ .label | count "abc" }}
``` 

would return 2.

**Checklist**
- [x] Documentation added
- [X] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/logql/log/fmt.go
```diff
@@ -43,6 +44,10 @@ var (
 			r := regexp.MustCompile(regex)
 			return r.ReplaceAllLiteralString(s, repl)
 		},
+		"count": func(substr string, s string) string {
+			count := strings.Count(s, substr)
```
dannykopping, line 48: I wonder if we should support a regex pattern here as well?

dannykopping (commented):
I like it!

I could see myself using this with `unwrap`. Does this work with `label_format`?

pkg/logql/log/fmt.go
MichelHollands: Yes we can. It's been added.

MichelHollands:
> Does this work with `label_format`?

Yes, unit tests have been added for this as well.

pkg/logql/log/fmt.go
```diff
@@ -43,6 +43,14 @@ var (
 			r := regexp.MustCompile(regex)
 			return r.ReplaceAllLiteralString(s, repl)
 		},
+		"count": func(regexsubstr string, s string) int {
+			r := regexp.MustCompile(regexsubstr)
```
dannykopping, line 47: This will panic if the regex is invalid
Please add a test case for that as well

docs/sources/logql/template_functions.md
```diff
@@ -697,3 +697,21 @@ Example of a query to print a `-` if the `http_request_headers_x_forwarded_for`
 ```logql
 {job="access_log"} | json | line_format `{{.http_request_headers_x_forwarded_for | default "-"}}`
 ```
+
+## count
+
+`count` counts occurrences of the regex (`regexsubstr`) in (`src`).
+
+Signature: `count(regexsubstr string, src string) int`
```
dannykopping, line 705:
```suggestion
`count` counts occurrences of the regex (`regex`) in (`src`).

Signature: `count(regex string, src string) int`
```

dannykopping (commented):
Thanks for the addition! Spotted a potential panic

pkg/logql/log/fmt.go
MichelHollands: Tests have been added.

pkg/logql/log/fmt_test.go
```diff
@@ -511,6 +511,26 @@ func mustNewLabelsFormatter(fmts []LabelFmt) *LabelsFormatter {
 	return lf
 }
 
+// Test_Panic checks if a custom function that use regular expressions panics.
+// This panic is caught inside the text/template library when the function is called.
+func Test_Panic(t *testing.T) {
```
dannykopping, line 516: Why do we want it to panic? Can't we validate the regex and return an error message?

MichelHollands: Yes indeed, that also works.

pkg/logql/log/fmt.go
```diff
@@ -35,13 +35,30 @@ var (
...
-		"regexReplaceAllLiteral": func(regex string, s string, repl string) string {
-			r := regexp.MustCompile(regex)
-			return r.ReplaceAllLiteralString(s, repl)
+		"count": func(regexsubstr string, s string) (int, error) {
+			r, err := regexp.Compile(regexsubstr)
+			if err != nil {
+				return 0, err
+			}
+			matches := r.FindAllStringIndex(s, -1)
+			if matches == nil {
```
ashwanthgoli, line 58: nit:`len` works on a nil slice
https://go.dev/play/

ashwanthgoli (approved):
LGTM

pkg/logql/log/fmt.go
MichelHollands: Good catch, it's changed.

PR #7953, opened 2022-12-16
Only log flush error when there is one.

jeschkies:


MichelHollands (approved):
LGTM

dannykopping (changes requested):
This error message won't make much sense to the user; can we make it more clear what is the problem exactly?

pkg/util/log/log.go
```diff
@@ -170,7 +170,9 @@ func CheckFatal(location string, err error, logger log.Logger) {
 
 	logger.Log("err", errStr)
 	err = Flush()
-	fmt.Fprintln(os.Stderr, "Could not flush logger", err)
+	if err != nil {
```
owen-d, line 173: This looks unnecessary as the first line of this `CheckFatal` function is
```
	if err == nil {
		return
	}
```

owen-d:
Closing, please reopen if I've missed something.

PR #7957, opened 2022-12-16
Fix min and max aggregations when using instant queries.

MasslessParticle:
Instant queries use the new `streamRangeVectorIterator` which has a new set of `aggregators`. Before, aggregations were based on the first point in the range to be aggregated. With  `streamRangeVectorIterator`, points are instead treated one at a time. 

The aggregations were simply checking whether the value was less than the min or greater than the max without any initialization. This PR fixes these aggregations by initializing `min` or `max` to `math.Nan` so the first point is always considered.


JordanRushing:
Code changes and the new test LGTM 👍 

Looks like CI is failing; maybe re-run? Also we should use the Loki PR template for the description? 😄

JordanRushing (approved):

PR #7966, opened 2022-12-19
fix query-frontend request load balancing when using k8s service

sandeepsukhani:
We noticed an imbalance in the requests sent to the Query Frontend pods using `query-frontend` k8s service. This seems to be caused by `query-frontend` being a headless service that resolves to each QF pod IP and leaves it up to the client to load balance the requests as mentioned [here](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services).

This PR fixes the issue by creating two separate services for pod IP discovery and load balancing of queries:
* `query-frontend` to be used for load balancing incoming Loki queries.
* `query-frontend-headless` to be used for discovering QF pod IPs from queriers to connect as workers.

**Checklist**
- [x] `CHANGELOG.md` updated


dannykopping (approved):
LGTM!
I think we should add this to the upgrade guide as well

dannykopping (approved):
LGTM!

Issue #7876, opened 2022-12-07
Support exemplars like datapoints on Logql metrics queries in Loki


ajitpatil:
Loki datasource already supports linking specific fields like traceId to Jaeger datasource. It shows as a link on a popup in raw logs view. But similar feature could be added with exemplar-like datapoints on the time-series panels with log metrics queries in Loki datasource. 

![image](https://user-images.githubusercontent.com/772666/206277670-514c6197-1ea3-4db8-99a3-694177154de9.png)

ajitpatil:
@liguozhong Thanks for linking this feature request to that.

liguozhong:
> @liguozhong Thanks for linking this feature request to that.

Hi, @ajitpatil，Thans for the detailed picture design.
 I submit a PR to support the http api. If this PR is successfully merged, you need to file an issue in the github of grafana/grafana to support the display of exemplars data including explore and panel in the grafana UI. This issue will be a long process, looking forward to having this powerful feature.

https://github.com/grafana/loki/pull/7974

PR #7974, opened 2022-12-20
[new feature] API: Introduce exemplar api

liguozhong:
image
Provides loki's http api for this feature.

Introduce exemplar http api ( /loki/api/v1/query_exemplars ).
prometheus exemplar doc link: https://prometheus.io/docs/prometheus/latest/querying/api/#querying-exemplars

query_range:
`http://localhost:3100/loki/api/v1/query_range?direction=BACKWARD&limit=1732&query=count_over_time({job="varlogs"}[1m])&start=1671449435000000000&end=1671453036000000000&step=2
`
image

VS
query_exemplars:
`http://localhost:3100/loki/api/v1/query_exemplars?direction=BACKWARD&limit=1732&query=count_over_time({job="varlogs"}[1m])&start=1671449435000000000&end=1671453036000000000&step=2`
image

Fixes #https://github.com/grafana/loki/issues/7876

`Why the api parameters of /loki/api/v1/query_exemplars are exactly the same as those of /loki/api/v1//query_range？
` 
because the label of loki can be dynamically generated during the logql query process through parsers such as | json and | regex, which is different from the fixed label of prometheus. Therefore, the parameters of loki's exemplars api cannot use label match like prometheus, but should use the same logql parameters as query_range to make better use of the schema on read loki design.

Therefore, this exemplar code PR becomes relatively long, which will cause the reviewer to work very hard. I am sorry for this.

`Why is the test coverage rate dropping dramatically?
`
Because the amount of code in this PR is too much, this PR only completes the prototype of the exemplar feature with the smallest amount of code, and the testing and doc will be completed in another PR.
```
-           ingester	-1.8%
+        distributor	0%
-            querier	-3.2%
+ querier/queryrange	0%
-               iter	-24%
-            storage	-7.7%
-           chunkenc	-6%
-              logql	-15.6%
+               loki	0%
```

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


slim-bean:
This is very interesting @liguozhong! 

I am wondering though if it might be simpler to implement such a feature in just the front end?

If you were to click on a metric graph at any point couldn't Grafana make a query for the corresponding log line(s) in a lookup query?  This won't be perfect however because a point on a graph can encompass many log lines.

A related question? How are the exemplars chosen?

liguozhong:
> If you were to click on a metric graph at any point couldn't Grafana make a query for the corresponding log line(s) in a lookup query? This won't be perfect however because a point on a graph can encompass many log lines.

@slim-bean , Thanks for such a timely review. This idea is great. Compared with the implementation of this PR, this idea can reduce a lot of s3 query calculations. I hadn't thought of this way before. But maybe we can provide 2 kinds of exemplar api option.

Although the amount of code in this PR is too bloated, from a mathematical point of view, the implementation of this PR is the correct answer. At the same time, this PR is similar to the design of Prometheus's exemplar. As far as I know, the exemplar concept was designed by Google's monarch system after a lot of practice,exemplar is a very good monitoring standard. loki can also follow the exemplar standard.

I suggest that we can provide two exemplar APIs for users to choose from. Comparing the calculation amount of this PR is basically the same as that of /query_range, it is necessary to initiate a very large amount of query requests to s3.

liguozhong:
> A related question? How are the exemplars chosen?

`range_exemplar.go` code here completes the logic of how to select the exemplar. From the code point of view, the last log in the range is selected as the exemplar.

```go
func (r *streamRangeExemplarIterator) load(start, end int64) {
  for lbs, sample, hasNext := r.iter.Peek(); hasNext; lbs, sample, hasNext = r.iter.Peek() {
    rangeAgg, ok = r.windowRangeAgg[lbs]
    p := exemplar.Exemplar{
      Ts:     sample.TimestampMs,
      Value:  sample.Value,
      Labels: logproto.FromLabelAdaptersToLabels(sample.Labels),
    }
    rangeAgg.agg(p)
    _ = r.iter.Next()
  }
}
func (r *streamRangeExemplarIterator) At() (int64, []exemplar.QueryResult) {
	if r.exemplars == nil {
		r.exemplars = make([]exemplar.QueryResult, 0, len(r.windowRangeAgg))
	}
	r.exemplars = r.exemplars[:0]
	ts := r.current/1e+6 + r.offset/1e+6
	for lbs, rangeAgg := range r.windowRangeAgg {
		exp := exemplar.Exemplar{
			Labels: rangeAgg.at().Labels,
			Ts:     ts,
		}

		eps := make([]exemplar.Exemplar, 0)
		eps = append(eps, exp)

		r.exemplars = append(r.exemplars, exemplar.QueryResult{
			SeriesLabels: r.metrics[lbs],
			Exemplars:    eps,
		})
	}
	return ts, r.exemplars
}

```
```go
type ExemplarAgg struct {
	exemplar exemplar.Exemplar
}

func (a *ExemplarAgg) agg(exemplar exemplar.Exemplar) {
	a.exemplar = exemplar
}

func (a *ExemplarAgg) at() exemplar.Exemplar {
	return a.exemplar
}
```

owen-d (commented):
This is a really cool PR, but I have a few reservations.

* Performance: Exemplars are likely meant to run alongside queries. How can we ensure this doesn't double query load?
* Selection: All logs are processed in order to select the last log line as an exemplar (related to performance). I don't think we can accept this approach, especially on large datasets. 
* Maintainability: This PR duplicates most of the read path -- I don't think this is a sustainable pattern.

Considering the downsides, do we need this now? I don't think so. I'd be interested in figuring out a more scalable & sustainable approach to this problem. Perhaps we could extend the QuerySample API to include exemplars, scanning the data only once? It feels bad to abandon such a large and well constructed PR and I feel terrible blocking it -- in the future it's worth talking about possible solutions before writing all the code (github issues are great places to do this).

liguozhong:
I agree with you, this PR took us 1 week, and the code is difficult to maintain, not suitable for merging into master. 
But this PR is quite useful for our company, before the master branch implements the exemplar feature. Since we are going to implement `logreduce`, we will maintain this PR separately in our private branch.

owen-d:
I'm going to close this PR in the meantime since we don't plan to accept it.

PR #7975, opened 2022-12-20
feat(logql): Support drop labels in logql pipeline

### LINKED ISSUE #7368: Log query fallback parse method
There is no way to define a fallback parsing method for log queries. 

For example, If I have a query like below and if `logfmt` can't parse the line I would like to be able to show the whole line or try parsing with `json`
```
{namespace=~"$namespace", container=~"$container", pod=~"$pod"} 
| logfmt 
| __error__ = ""
```

Maybe there could be if else conditions for steps so we can continue if `__error__ = ""` and parse with another method in case there is an error

### LINKED ISSUE #7870: Querying mixed format logs
I have an application that emits logs in a JSON format with useful extra fields alongside a string message.
The pod it runs in has a sidecar that emits plain text logs.

I ideally want to be able to view logs from both containers (`app` and `sidecar`) in the same query.
e.g.
`{pod="foobar"} | json | line_format "{{ default __line__ .message }}"`

The log query for this actually works just fine, except it has the `__error__` label added which then affects the way Grafana displays the lines (greyed out with a warning icon).

The metrics query (with a `count_over_time`) returns an actual error from Loki however.
I can use a label *filter* (`__error__!="JSONParserErr"`) to allow the query to work but then it doesn't actually count or display the non-json lines.

1) Some kind of flag on the `json` parser to ignore errors and not insert the `__error__` label

2) The ability to `label_format` errors away. 
 Formatting `__error__` is [explicitly](https://github.com/grafana/loki/blob/v2.7.0/pkg/logql/log/fmt.go#L293) blocked at the moment. `label_format __error__=""` returns an error.
Trying to rename it just has no effect: `label_format foo=__error__ `

**Additional context**
The display issue from the log query could easily be a Grafana issue instead to just allow hiding errors on the UI side.

The metrics query seems like it can only be fixed in Loki.

adityacs:
This PR introduces `drop` stage in logql pipeline. 

Fixes #7870, Fixes #7368

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated
- [x] Changes that require user attention or interaction to upgrade are documented in `docs/sources/logql/log_queries.md`


pkg/logql/log/fmt.go
```diff
@@ -299,14 +300,17 @@ func NewLabelsFormatter(fmts []LabelFmt) (*LabelsFormatter, error) {
...
 }
 
-func validate(fmts []LabelFmt) error {
+func validate(fmts []LabelFmt, formatRenameErrors bool) error {
 	// it would be too confusing to rename and change the same label value.
 	// To avoid confusion we allow to have a label name only once per stage.
 	uniqueLabelName := map[string]struct{}{}
 	for _, f := range fmts {
-		if f.Name == logqlmodel.ErrorLabel {
+		if f.Name == logqlmodel.ErrorLabel && !formatRenameErrors {
```
owen-d, line 308: @cyriltovena Any idea why we prevented renaming the error label previously? I'm inclined to not treat it in a special way.

owen-d (commented):
We can already use `| __error__=""` to ignore errors and `| label_format x="{{.__error__}}"` to extract errors to new labels, but I'm definitely interested in a way to remove existing labels. It sounds like your `label_format_rename_errors` is a nice syntactic sugar for a combination of these.

Would it make more sense to start with a `| strip <label> <label>` operation which could remove labels from the results? I think this would result in cleaner code as well since we wouldn't have to have special cases for the `__error__` label besides preventing it from being rewritten (current behavior).

adityacs:
@owen-d 
> We can already use | __error__="" to ignore errors and | label_format x="{{.__error__}}" to extract errors to new labels

`__error__=""` drops the lines which contains the error. With `ignore_errors` we are just removing the `__error__` label but keeping the line. 

With `label_format dst={{__error__}}`, we can extract but we are trying to remove `__error__` label here. Instead of just removing, with `label_format_rename_error`, we can capture the error with different key. 

>Would it make more sense to start with a | strip <label> <label> operation which could remove labels from the results? I think this would result in cleaner code as well since we wouldn't have to have special cases for the __error__ label besides preventing it from being rewritten (current behavior).

I am totally ok with this. This would be a nice approach. However, I think we should still support renaming `__error__` labels(with `label_format` or `label_format_rename_errors`), this would allow user to know there was an error in that line. If the user wants to remove or ignore the errors,  we could support `strip <label>`  instead of `ignore_errors`.

owen-d:
> __error__="" drops the lines which contains the error. With ignore_errors we are just removing the __error__ label but keeping the line.

Good distinction there 👍 . 

I think we could achieve what you're suggesting with something like
```
| label_format dst={{__error__}} | strip __error__
```

This would be a little more verbose, but it won't have to treat the `__error__` label in a special way and could be used for other use cases, which I like. WDYT?

adityacs:
```| label_format dst={{__error__}} | strip __error__```

Yes, this would be do.

Should we say `label_remove` or `label_drop` or `srtrip_label` which would be more intuitive instead of `strip`?

Also, should we support `strip` with a match condition i.e. strip labels only if it matches some value? Something like`|strip __error__=JSONParserErr`.

**Comment by owen-d** [comment]
I'm inclined to just `| drop <label> <label>` and `| drop <label>=<match> <label>=<match>`

**Comment by adityacs** [comment]
>`| drop <label> <label>` and `| drop <label>=<match> <label>=<match>` 
Yes, this works. Will make changes accordingly.

**Comment by adityacs** [comment]
@owen-d Not sure if we should support `__error__details__` explicitly. For `match` expression it would be too verbose to match by `__error_details__`.  For example, something like this would be too verbose `|drop __error_details__="expecting json object(6), but it is not"`.

What do you suggest?

**Review by owen-d — COMMENTED** [review]
I'd like to support `drop` with `__error_details__`. Is there a reason you don't think so? For instance, `| drop __error_details__=~"unexpected json:.*"` would be helpful.

**Comment by adityacs** [comment]
@owen-d Have included `__error_details__` in `drop`

File: docs/sources/logql/log_queries.md [hunk 2]
```diff
@@ -556,3 +556,41 @@ In both cases, if the destination label doesn't exist, then a new one is created
...
+
+The `=` operator after the label name is a **label matching operator**.
+The following label matching operators are supported:
+
+- `=`: exactly equal
+- `!=`: not equal
+- `=~`: regex matches
+- `!~`: regex does not match
+
+The `| drop` expression will drop the given labels in the pipeline. For example, for the query `{job="varlogs"}|json|drop level, method="GET"`, with below log line
```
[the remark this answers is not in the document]
adityacs: Have added one more example with `regex match`. The examples shown will cover the above examples. Let me know if we should add more.

docs/sources/logql/log_queries.md
```diff
@@ -556,3 +556,55 @@ In both cases, if the destination label doesn't exist, then a new one is created
...
+Similary, this expression can be used to drop `__error__` labels as well. For example, for the query `{job="varlogs"}|json|drop __error__`, with below log line
+
+```
+INFO GET / loki.net 200
+```
+
+the result will be
+
+```
+{} INFO GET / loki.net 200
```
owen-d, line 593: I think this is how it should work, but currently I expect it to keep `__error_details__` when `| drop __error__` is used. I think it _should_ also remove `__error_details__` like this doc suggests.

owen-d (approved):
This looks great. I'm approving now, but I think `| drop __error__` should also remove `__error_details__` so we can more ergonomically use `| drop __error__` instead of needing to do `| drop __error__ __error_details__`. Can we make that last change?

docs/sources/logql/log_queries.md
adityacs: @owen-d Yeah, that was the earlier behaviour I added but later removed that while adding support for `__error__details__`.  I will send a follow up PR for this change.

What should be the behaviour in this case `|drop __error_details__` or `|drop __error_details__="some error details"`? Should we remove `__error__` with this?

PR #7978, opened 2022-12-20
Gracefully shut down query frontend

chaudum:

This PR implements a graceful shutdown of the query frontend. When shutting down, the service needs to wait for all inflight requests that have already been sent to the scheduler/queriers. Otherwise, the service may shut down before the querier sends back the results to the frontend, resulting in failed queries.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/lokifrontend/frontend/v2/frontend.go
```diff
@@ -144,19 +146,29 @@ func NewFrontend(cfg Config, ring ring.ReadRing, log log.Logger, reg prometheus.
...
+func (f *Frontend) stopping(e error) error {
+	t := time.NewTicker(500 * time.Millisecond)
+	defer t.Stop()
+
+	start := time.Now()
+
+	for range t.C {
+		inflight := f.requests.count()
+		if time.Since(start) > f.cfg.GracefulShutdownTimeout || inflight <= 0 {
+			break
```
jeschkies, line 161: I would add a log message here.

chaudum: I moved the log message from below up here.

pkg/lokifrontend/frontend/v2/frontend.go
```diff
@@ -144,19 +146,30 @@ func NewFrontend(cfg Config, ring ring.ReadRing, log log.Logger, reg prometheus.
...
 }
 
-func (f *Frontend) stopping(_ error) error {
+func (f *Frontend) stopping(e error) error {
+	t := time.NewTicker(500 * time.Millisecond)
+	defer t.Stop()
+
+	start := time.Now()
+
+	for range t.C {
```
jeschkies, line 158: I'm not a big fan of active polling. Ideally we would wait for done with a timeout and `select`. However, this is probably too complicated for this change.

jeschkies (commented):
How is the frontend signaled not to queue new items?

pkg/lokifrontend/frontend/v2/frontend.go
```diff
@@ -144,19 +146,30 @@ func NewFrontend(cfg Config, ring ring.ReadRing, log log.Logger, reg prometheus.
...
+		}
+		level.Debug(f.log).Log("msg", "outstanding queries waiting to complete", "inflight", inflight, "elapsed", elapsed)
+	}
+
 	return errors.Wrap(services.StopAndAwaitTerminated(context.Background(), f.schedulerWorkers), "failed to stop frontend scheduler workers")
 }
 
 // RoundTripGRPC round trips a proto (instead of a HTTP request).
 func (f *Frontend) RoundTripGRPC(ctx context.Context, req *httpgrpc.HTTPRequest) (*httpgrpc.HTTPResponse, error) {
-	if s := f.State(); s != services.Running {
```
jeschkies, line 156: I think we should refuse new requests. There is `service.Stopping`.

jeschkies (changes requested):

pkg/lokifrontend/frontend/v2/frontend.go
```diff
@@ -144,19 +146,30 @@ func NewFrontend(cfg Config, ring ring.ReadRing, log log.Logger, reg prometheus.
...
+	t := time.NewTicker(500 * time.Millisecond)
+	defer t.Stop()
+
+	start := time.Now()
+
+	for range t.C {
+		inflight := f.requests.count()
+		elapsed := time.Since(start)
+		if elapsed > f.cfg.GracefulShutdownTimeout || inflight <= 0 {
+			level.Debug(f.log).Log("msg", "no outstanding queries waiting to complete", "inflight", f.requests.count(), "elapsed", elapsed)
```
salvacorts, line 162: This log message is not correct. It may be printed when elapsed is larger than  `GracefulShutdownTimeout` even though there may be outstanding inflight requests.

Maybe separate it into two _if_ statements with different logs?

pkg/lokifrontend/frontend/v2/frontend.go
```diff
@@ -144,19 +146,30 @@ func NewFrontend(cfg Config, ring ring.ReadRing, log log.Logger, reg prometheus.
...
-func (f *Frontend) stopping(_ error) error {
+func (f *Frontend) stopping(e error) error {
+	t := time.NewTicker(500 * time.Millisecond)
+	defer t.Stop()
+
+	start := time.Now()
+
+	for range t.C {
+		inflight := f.requests.count()
+		elapsed := time.Since(start)
```
salvacorts, line 160: Maybe we can use `time.After` and `time.Tick` here?
Example: https://go.dev/play/p/aavdcV5r96k

pkg/lokifrontend/frontend/v2/frontend.go
```diff
@@ -259,13 +272,22 @@ func (f *Frontend) QueryResult(ctx context.Context, qrReq *frontendv2pb.QueryRes
 			level.Warn(f.log).Log("msg", "failed to write query result to the response channel", "queryID", qrReq.QueryID, "user", userID)
 		}
 	}
+	// TODO(chaudum): In case the the userIDs do not match, we do not send a
```
salvacorts, line 275: I understand this TODO will be addressed on a separate PR?

salvacorts (approved):
LGTM. Just a few nits

pkg/lokifrontend/frontend/v2/frontend.go
```diff
@@ -144,19 +146,30 @@ func NewFrontend(cfg Config, ring ring.ReadRing, log log.Logger, reg prometheus.
 }
 
 func (f *Frontend) starting(ctx context.Context) error {
-	return errors.Wrap(services.StartAndAwaitRunning(ctx, f.schedulerWorkers), "failed to start frontend scheduler workers")
+	return errors.Wrap(services.StartAndAwaitRunning(context.Background(), f.schedulerWorkers), "failed to start frontend scheduler workers")
```
salvacorts, line 149: I'd add a comment here explaining why do we create a new context for the child.

pkg/lokifrontend/frontend/v2/frontend.go
chaudum: Right

pkg/lokifrontend/frontend/v2/frontend_test.go
```diff
@@ -252,6 +251,94 @@ func TestFrontendFailedCancellation(t *testing.T) {
...
+		go func() {
+			_, err := f.RoundTripGRPC(user.InjectOrgID(ctx, "test"), &httpgrpc.HTTPRequest{})
+			require.NoError(t, err)
+		}()
+	}
+
+	require.Eventually(t, func() bool {
+		return f.requests.count() == inflightRequests
+	},
+		3*delayResponse, // wait at least 2*delayResponse to wait for queries to be finished and removed from inlight requests map
```
vlad-diachenko, line 283: nit:
it's too short... I would make it at least 1s ... it would prevent us from having flaky tests ;) 
if 10 requests are executed sequentially (it is unbelievable but possible;))) ), then the test will take > 200ms.... and this check will be failed.

pkg/lokifrontend/frontend/v2/frontend.go
chaudum: These are not "new" requests, these are sub-requests from a query request.

> During the time of the shutdown, the RoundTripGRCP function still needs to accept any round-tripping requests. This is, because the frontend round tripper is wrapped into a query range tripper ware, which splits incoming (HTTP) query requests into multiple (gRPC) query requests if needed.

pkg/lokifrontend/frontend/v2/frontend.go
```diff
@@ -50,6 +51,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.SchedulerAddress, "frontend.scheduler-address", "", "DNS hostname used for finding query-schedulers.")
 	f.DurationVar(&cfg.DNSLookupPeriod, "frontend.scheduler-dns-lookup-period", 10*time.Second, "How often to resolve the scheduler-address, in order to look for new query-scheduler instances. Also used to determine how often to poll the scheduler-ring for addresses if the scheduler-ring is configured.")
 	f.IntVar(&cfg.WorkerConcurrency, "frontend.scheduler-worker-concurrency", 5, "Number of concurrent workers forwarding queries to single query-scheduler.")
+	f.DurationVar(&cfg.GracefulShutdownTimeout, "frontend.graceful-shutdown-timeout", 5*time.Minute, "Time to wait for inflight requests to finish before forcefully shutting down. This needs to be aligned with the query timeout and the graceful termination period of the process orchestrator.")
```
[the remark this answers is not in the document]
chaudum: 5m is the default query timeout. We should wait for that time, otherwise queries may fail.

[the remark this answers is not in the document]
chaudum: > I'd suggest aligning the default value of GracefulShutdownTimeout to the default of terminationGracePeriodSeconds. That's 30s.

We should do it the other way round. Or at least set both. That's why the documentation says

> This needs to be aligned with the query timeout and the graceful termination period of the process orchestrator.

pkg/lokifrontend/frontend/v2/frontend.go
chaudum: Right, I'll revert https://github.com/grafana/loki/pull/7978/commits/09da99f99032f5d4ab9cd5240fd76cf1bf556057 again.

pkg/lokifrontend/frontend/v2/frontend_test.go
```diff
@@ -252,6 +251,94 @@ func TestFrontendFailedCancellation(t *testing.T) {
...
+	// wait for Stopping state
+	require.Eventually(t, func() bool {
+		t.Log(f.State())
+		return f.State() == services.Stopping
+	}, delayResponse/2, 2*time.Millisecond)
+
+	// send (sub-)request
+	// This request still needs to be able to pass the RoundTripGRCP function,
+	// because it may be a sub-request of a query request that was split earlier
+	// into multiple sub-requests.
```
vlad-diachenko, line 335: why do we need it? I believe that another query-frontend can handle it... 
if we started the shutdown process for this query-frontend, it should not receive any new requests... Wdyt?

pkg/scheduler/scheduler.go
```diff
@@ -297,10 +297,10 @@ func (s *Scheduler) FrontendLoop(frontend schedulerpb.SchedulerForFrontend_Front
 		switch msg.GetType() {
 		case schedulerpb.ENQUEUE:
 			err = s.enqueueRequest(frontendCtx, frontendAddress, msg)
-			switch {
-			case err == nil:
```
vlad-diachenko, line 301: :)

vlad-diachenko (approved):
LGTM. just a few not critical questions/comments

pkg/lokifrontend/frontend/v2/frontend_test.go
chaudum: These 10 requests are executed concurrently.

pkg/lokifrontend/frontend/v2/frontend_test.go
chaudum: It's a test for the very counter-intuitive behaviour. By intuition, you would say that the RoundTripGRCP must not access any more incoming requests if the state is not `Running`. However, the opposite is the case, because the function is not the entry-point for requests, but rather a grcp round tripper that is wrapped into the query splitting tripperware, that produces potentially lots of internal gRPC requests that are handled by this function.

vlad-diachenko: ok, i see.

pkg/lokifrontend/frontend/v2/frontend_test.go
vlad-diachenko: yep, I saw ;) anyway, the check happens every tick and completes as soon the function returns true, so, making timeout bigger just removed the possibility of this test to became flaky test ;) not critical ;)

docs/sources/configuration/_index.md
```diff
@@ -666,6 +666,12 @@ The `frontend` block configures the Loki query-frontend.
 # frontend.grpc-client-config
 [grpc_client_config: <grpc_client>]
 
+# Time to wait for inflight requests to finish before forcefully shutting down.
+# This needs to be aligned with the query timeout and the graceful termination
+# period of the process orchestrator.
+# CLI flag: -frontend.graceful-shutdown-timeout
+[graceful_shutdown_timeout: <duration> | default = 5m]
```
ssncferreira, line 673: 🤩

pkg/lokifrontend/frontend/v2/frontend.go
ssncferreira: Not sure I understand the logic here. The query-frontend is the one responsible for splitting a HTTP request into multiple HTTP over GRPC requests that are then propagated to the queriers correct? So is this `RoundTripGRPC` method the one that handles the HTTP over GRPC responses from the queriers?

pkg/lokifrontend/frontend/v2/frontend.go
```diff
@@ -259,13 +286,22 @@ func (f *Frontend) QueryResult(ctx context.Context, qrReq *frontendv2pb.QueryRes
...
+	// requests map.
 
 	return &frontendv2pb.QueryResultResponse{}, nil
 }
 
 // CheckReady determines if the query frontend is ready.  Function parameters/return
 // chosen to match the same method in the ingester
 func (f *Frontend) CheckReady(_ context.Context) error {
+	if s := f.State(); s != services.Running {
+		return fmt.Errorf("%v", s)
```
ssncferreira, line 302: nit
```suggestion
		return fmt.Errorf("Frontend service not running. Current state: %v", s)
```

pkg/lokifrontend/frontend/v2/frontend_test.go
ssncferreira, line 283:
```suggestion
		3*delayResponse, // wait at least 3*delayResponse to wait for queries to be finished and removed from inlight requests map
```

ssncferreira (commented):
LGTM 🌟 Great work! Just don't understand why the `RoundTripGRPC` still needs to accept incoming requests. From the comments, it seems that these are not new requests, so are they follow-up requests from the queriers. But in which cases do the queriers send follow sub-requests to the query frontend? 🤔

pkg/lokifrontend/frontend/v2/frontend.go
chaudum: We already prepend that text in the HTTP handler.

https://github.com/grafana/loki/blob/089ec1b05f5ec15a8851d0e8230153e0eeb4dcec/pkg/loki/loki.go#L569-L576

pkg/lokifrontend/frontend/v2/frontend.go
jeschkies: Alright. Then ignore my request. How are we rejecting Incoming requests though?

chaudum: The frontend has a `/ready` endpoint that calls the `CheckReady(context.Context) error` handler. This endpoint returns `503` when the frontend service is not in `Running` state.

https://github.com/grafana/loki/blob/5a8e9900565545cdbf76437dc294945ce056a5a0/pkg/lokifrontend/frontend/v2/frontend.go#L301-L303

If running in K8s this will remove the instance from the query frontend service and therefore does not get any new requests from upstream (e.g. gateway).

jeschkies (approved):

PR #7981, opened 2022-12-20
Add missing period_config root block to doc generator

ssncferreira:
Update doc generator tool to validate if slice element is a root block.

Add missing `period_config` root block to doc generator tool.

**Checklist**
- [ ] Documentation updated

### DIFF
--- a/pkg/storage/config/schema_config.go
+++ b/pkg/storage/config/schema_config.go
@@ -91,13 +91,13 @@ type PeriodConfig struct {
 	// used when working with config
 	From DayTime `yaml:"from" doc:"description=The date of the first day that index buckets should be created. Use a date in the past if this is your only period_config, otherwise use a date when you want the schema to switch over. In YYYY-MM-DD format, for example: 2018-04-15."`
 	// type of index client to use.
-	IndexType string `yaml:"store"`
+	IndexType string `yaml:"store" doc:"description=store and object_store below affect which <storage_config> key is used.\nWhich store to use for the index. Either aws, aws-dynamo, gcp, bigtable, bigtable-hashed, cassandra, boltdb or boltdb-shipper. "`
 	// type of object client to use; if omitted, defaults to store.
-	ObjectType  string              `yaml:"object_store"`
-	Schema      string              `yaml:"schema"`
-	IndexTables PeriodicTableConfig `yaml:"index"`
-	ChunkTables PeriodicTableConfig `yaml:"chunks"`
-	RowShards   uint32              `yaml:"row_shards"`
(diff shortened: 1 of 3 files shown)


pkg/storage/config/schema_config.go
```diff
@@ -369,17 +369,17 @@ func (cfg *PeriodConfig) VersionAsInt() (int, error) {
...
+	Tags   Tags          `yaml:"tags" doc:"description=A map to be added to all managed tables."`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
 func (cfg *PeriodicTableConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	g := struct {
-		Prefix string         `yaml:"prefix"`
-		Period model.Duration `yaml:"period"`
-		Tags   Tags           `yaml:"tags"`
+		Prefix string
```
vlad-diachenko, line 380: when we will parse the yaml file, we will expect properties starting from capital "P". I believe that it will brake the parsing process....

pkg/storage/config/schema_config.go
```diff
@@ -395,9 +395,9 @@ func (cfg *PeriodicTableConfig) UnmarshalYAML(unmarshal func(interface{}) error)
 // MarshalYAML implements the yaml.Marshaler interface.
 func (cfg PeriodicTableConfig) MarshalYAML() (interface{}, error) {
 	g := &struct {
-		Prefix string         `yaml:"prefix"`
-		Period model.Duration `yaml:"period"`
-		Tags   Tags           `yaml:"tags"`
+		Prefix string
+		Period model.Duration
+		Tags   Tags
 	}{
```
vlad-diachenko, line 401: it will produce yaml with properties starting from capital letters...

vlad-diachenko (commented):
Looks great, could you please check my comments...

pkg/storage/config/schema_config.go
ssncferreira: Great catch @vlad-diachenko this completely breaks the parsing. Removed this change in commit: https://github.com/grafana/loki/pull/7981/commits/f585d81c3a6bdce2e58846c85d60c39ea5e24908

pkg/storage/config/schema_config.go
ssncferreira: Addressed in https://github.com/grafana/loki/pull/7981/commits/f585d81c3a6bdce2e58846c85d60c39ea5e24908

vlad-diachenko (approved):
LGTM ;) thanks for addressing the comments ;) 🥇

PR #7983, opened 2022-12-20
Add README file to doc-generator tool

ssncferreira:

Add README file to doc-generator tool

### DIFF
--- a/tools/doc-generator/README.md
+++ b/tools/doc-generator/README.md
@@ -1,0 +1,23 @@
+# Documentation generation tool
+
+The doc-generator tool is used to automatically generate the configuration flags documentation from the metadata information
+provided in the code. 
+
+## Run 
+
+The tool receives as input a template file and generates the list of configuration value blocks: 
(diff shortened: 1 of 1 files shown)


tools/doc-generator/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The doc-generator tool is used to automatically generate the configuration flags documentation from the metadata information
+provided in the code. 
+
+## Run 
+
+The tool receives as input a template file and generates the list of configuration value blocks: 
+
+```shell
+go run ./tools/doc-generator docs/sources/configuration/index.template > docs/sources/configuration/_index.md
```
vlad-diachenko, line 11: looks great ;) however, we could simplify adding a target to Makefile that would do everything that we need ))

ssncferreira: There is already a Makefile target: https://github.com/grafana/loki/blob/main/Makefile#L732-L733

PR #7984, opened 2022-12-20
Add empty k3d cluster for testing helm chart

trevorwhitney:
Add a new k3d cluster which is a cluster that has all Loki dependencies for purposes of testing the helm charts using `helm install`. The other environments use the helm charts, but deploy them using `tanka` and therefore can't catch certain race conditions or issues that occur with helm installs.


owen-d (approved):
skimmed, lgtm

PR #7988, opened 2022-12-21
store: write overlapping chunks to multiple stores

### LINKED ISSUE #7847: No duplication on collisions between period configs
**Describe the bug**
When you add another period config with a different `object_store` to the `schema_config`, seems based on the code (https://github.com/grafana/loki/blob/v2.7.0/pkg/storage/stores/composite_store.go#L204-L250) it should duplicate the chunk into both stores so for example ruler can fetch the chunk from both old and new period configs for the chunks that are started from the day before the new period config and end after the new period config.

But in this slice:
https://github.com/grafana/loki/blob/v2.7.0/pkg/storage/stores/series_store_write.go#L66-L67
We are preventing from storing the chunk in the new period config store cause it first uploads to the old period config (chunk has collision between both period configs) and then as it's cached by the first period config (and the cache is shared between stores), the new period config won't write it.

What happens is now for example ruler tries to fetch the chunks from the new period config (based on the expr and...) and it couldn't find it and will reproduce lots of 404 reqs to the new store and also failed to evaluate the rules.

I was going to add an `isDuplicate` to the `forStores` callback func, so it can pass it as a bool and `PutOne` func can check for the cache only if it was not duplicated - but I see it will going to add that arg for other functions that don't need it (like GetSeries and...).
```diff
diff --git a/pkg/storage/stores/composite_store.go b/pkg/storage/stores/composite_store.go
index fd8124ae0..8bced2abc 100644
--- a/pkg/storage/stores/composite_store.go
+++ b/pkg/storage/stores/composite_store.go
@@ -201,7 +201,7 @@ func (c compositeStore) Stop() {
        }
 }
 
-func (c compositeStore) forStores(ctx context.Context, from, through model.Time, callback func(innerCtx context.Context, from, through model.Time, store Store) error) error {
+func (c compositeStore) forStores(ctx context.Context, from, through model.Time, callback func(innerCtx context.Context, from, through model.Time, store Store, isDuplicate bool) error) error {
        if len(c.stores) == 0 {
                return nil
        }
@@ -230,6 +230,8 @@ func (c compositeStore) forStores(ctx context.Context, from, through model.Time,
                return b
        }
 
+       // Check if chunk should be duplicated to stores because of the collision between period configs.
+       isDuplicate := j-i > 1
        start := from
        for ; i < j; i++ {
                nextSchemaStarts := model.Latest
@@ -238,7 +240,7 @@ func (c compositeStore) forStores(ctx context.Context, from, through model.Time,
                }
 
                end := min(through, nextSchemaStarts-1)
-               err := callback(ctx, start, end, c.stores[i])
+               err := callback(ctx, start, end, c.stores[i], isDuplicate)
                if err != nil {
                        return err
                }
diff --git a/pkg/storage/stores/series_store_write.go b/pkg/storage/stores/series_store_write.go
index 3dde483a6..f69e23e57 100644
--- a/pkg/storage/stores/series_store_write.go
+++ b/pkg/storage/stores/series_store_write.go
@@ -58,7 +58,7 @@ func (c *Writer) Put(ctx context.Context, chunks []chunk.Chunk) error {
 }
 
 // PutOne implements Store
-func (c *Writer) PutOne(ctx context.Context, from, through model.Time, chk chunk.Chunk) error {
+func (c *Writer) PutOne(ctx context.Context, from, through model.Time, chk chunk.Chunk, isDuplicate bool) error {
        log, ctx := spanlogger.New(ctx, "SeriesStore.PutOne")
        defer log.Finish()
        writeChunk := true
@@ -81,7 +81,7 @@ func (c *Writer) PutOne(ctx context.Context, from, through model.Time, chk chunk
        chunks := []chunk.Chunk{chk}
 
        // chunk not found, write it.
-       if writeChunk {
+       if writeChunk || isDuplicate {
                err := c.fetcher.Client().PutChunks(ctx, chunks)
                if err != nil {
                        return err
```

Do you have any better idea of how it can be fixed?

**To Reproduce**
1. Started Loki (v2.6.1)
2. Add two period configs with different object stores and check the behavior after the current date changes to the next period config.

**Expected behavior**
Put chunks on both stores when the collision happens in period configs.

ashwanthgoli:

When a chunk falls on the period boundary it should be written to both the stores as it gets referred to by the indexes on both sides. But we currently skip the writes for subsequent periods if the chunk is found in the cache. This PR ensures that overlapping chunks get written to both the stores

Fixes #7847 

Downside of this approach is that we end up writing the overlapping chunk multiple times, once from each instance in the replica set. But these duplicate writes would be infrequent given this occurs only on schema changes

**Checklist**
- [X] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * [7937](https://github.com/grafana/loki/pull/7937) **ssncferreira**: Deprecate CLI flag `-ruler.wal-cleaer.period` and replace it with `-ruler.wal-cleaner.period`.
 * [7906](https://github.com/grafana/loki/pull/7906) **kavirajk**: Add API endpoint that formats LogQL expressions and support new `fmt` subcommand in `logcli` to format LogQL query.
 * [7966](https://github.com/grafana/loki/pull/7966) **sandeepsukhani**: Fix query-frontend request load balancing when using k8s service.
+* [7988](https://github.com/grafana/loki/pull/7988) **ashwanthgoli** store: write overlapping chunks to multiple stores.
 
 ##### Changes
 
--- a/pkg/storage/stores/series_store_write.go
+++ b/pkg/storage/stores/series_store_write.go
@@ -61,12 +61,21 @@ func (c *Writer) Put(ctx context.Context, chunks []chunk.Chunk) error {
 func (c *Writer) PutOne(ctx context.Context, from, through model.Time, chk chunk.Chunk) error {
 	log, ctx := spanlogger.New(ctx, "SeriesStore.PutOne")
 	defer log.Finish()
-	writeChunk := true
+
+	var (
+		writeChunk = true
+		overlap    bool
+	)
+
+	// always write the chunk if it spans multiple periods to ensure that it gets added to all the stores
+	if chk.From < from || chk.Through > through {
+		overlap = true
+	}
 
 	// If this chunk is in cache it must already be in the database so we don't need to write it again
 	found, _, _, _ := c.fetcher.Cache().Fetch(ctx, []string{c.schemaCfg.ExternalKey(chk.ChunkRef)})
 
-	if len(found) > 0 {
+	if !overlap && len(found) > 0 {
 		writeChunk = false
 		DedupedChunksTotal.Inc()
 	}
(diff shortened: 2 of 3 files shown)


sandeepsukhani (approved):
LGTM

MichelHollands (approved):
LGTM

PR #7989, opened 2022-12-21
[new feature] logql:support sort and sort_desc

### LINKED ISSUE #7933: `sort_desc` missing from LogQL
**Describe the bug**
`sort_desc` is missing from LogQL. It's present in PromQL.

As a user I expect LogQL to be a strict superset of PromQL.

**Environment:**
Hosted Loki instance on GrafanaCloud

liguozhong:

sort()
sort(v instant-vector) returns vector elements sorted by their sample values, in ascending order.

sort_desc()
Same as sort, but sorts in descending order.

ex:
`sort(rate(({app=~"foo|bar"} |~".+bar")[1m])) without (app) `

prometheud doc: https://prometheus.io/docs/prometheus/latest/querying/functions/#sort

Fixes #7933

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/logql/engine.go
```diff
@@ -334,7 +334,13 @@ func (q *query) evalSample(ctx context.Context, expr syntax.SampleExpr) (promql_
 	}
 
 	if GetRangeType(q.params) == InstantType {
-		sort.Slice(vec, func(i, j int) bool { return labels.Compare(vec[i].Metric, vec[j].Metric) < 0 })
+		sortByValue, err := Sortable(q.params)
+		if err != nil {
+			return nil, err
+		}
+		if !sortByValue {
```
dannykopping, line 341: I'm confused; why is this `!sortByValue`?

dannykopping (commented):
This is awesome, thanks @liguozhong!!

pkg/logql/engine.go
```diff
@@ -334,7 +334,13 @@ func (q *query) evalSample(ctx context.Context, expr syntax.SampleExpr) (promql_
...
+			sort.Slice(vec, func(i, j int) bool { return labels.Compare(vec[i].Metric, vec[j].Metric) < 0 })
```
adityacs, line 342: I may be wrong. Just to clarify though. Should this not be 
```
sort.Slice(vec, func(i, j int) bool { return vec[i].V < vec[j].V })
```

pkg/logql/engine.go
```diff
@@ -334,7 +334,13 @@ func (q *query) evalSample(ctx context.Context, expr syntax.SampleExpr) (promql_
 	}
 
 	if GetRangeType(q.params) == InstantType {
-		sort.Slice(vec, func(i, j int) bool { return labels.Compare(vec[i].Metric, vec[j].Metric) < 0 })
+		sortByValue, err := Sortable(q.params)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
liguozhong: done，thanks

pkg/logql/engine.go
liguozhong: `			sort.Slice(vec, func(i, j int) bool { return labels.Compare(vec[i].Metric, vec[j].Metric) < 0 })
`
Because this line of code exists before this PR, the meaning of this line of code seems to be` sortByLabel()`, but sort and sort_desc need `sortByValue()`, so when we have `sort` and `sort_desc` in our logql grammar, we should not execute `sortByLabel ()`

pkg/logql/engine.go
liguozhong: Because this line of code exists before this PR, the meaning of this line of code seems to be sortByLabel(), but sort and sort_desc need sortByValue(), so when we have sort and sort_desc in our logql grammar, we should not execute sortByLabel ()


sortByValue is here to complete the sorting, and only here can the syntax tree be correctly parsed to accurately locate the need for sort or sort_desc.

```go
} else if expr.Operation == syntax.OpTypeSortDesc {
					result[groupingKey].heap = make(vectorByValueHeap, 0)
					heap.Push(&result[groupingKey].heap, &promql.Sample{
						Point:  promql.Point{V: s.V},
						Metric: s.Metric,
					})
				} else if expr.Operation == syntax.OpTypeSort {
					result[groupingKey].reverseHeap = make(vectorByReverseValueHeap, 0)
					heap.Push(&result[groupingKey].reverseHeap, &promql.Sample{
						Point:  promql.Point{V: s.V},
						Metric: s.Metric,
					})
				}
```

pkg/logql/engine.go
liguozhong: The purpose of modifying this line of code is to make this test pass. 
I debugged for a long time at the beginning, and it was correctly sorted by value in the front, but the final output is indeed in the string order of the label.

pkg/logql/engine.go
adityacs: Thank you. My bad, I didn't see this code earlier.

liguozhong: Thanks for taking the time to review this PR，👍

owen-d (changes requested):
I've been wanting us to add `sort` for a long time -- great start. A few notes:
* I think we need to prevent `by|without` groupings on `sort` operations. This will keep compatibility with promql and allowing `sort by (foo)` doesn't make much sense anyway when `sort` orders by _value_ instead of _labels_.

Some small changes and this will be ready to merge :)

pkg/logql/evaluator.go
```diff
@@ -387,7 +429,7 @@ func vectorAggEvaluator(
 				}
 				continue // Bypass default append.
 
-			case syntax.OpTypeBottomK:
+			case syntax.OpTypeBottomK, syntax.OpTypeSort:
 				// The heap keeps the lowest value on top, so reverse it.
 				sort.Sort(sort.Reverse(aggr.reverseHeap))
```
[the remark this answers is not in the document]
liguozhong: done,thanks.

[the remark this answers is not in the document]
liguozhong: and thank you very much for your review, otherwise this PR will be a PR with a bug.

owen-d (commented):
I've left one suggestion, then LGTM!

owen-d (approved):

SHaaD94:
Hi guys.

What is the use case for this sort function? 
API already has a direction [https://grafana.com/docs/loki/latest/api/#query-loki](parameter)

Issue #1540, opened 2020-01-17
Promtail syslog input cannot recover from crashes


rfratto:
If Promtail happens to crash, when it restarts, it will continue reading files and journald from the last saved position in the syslog. Aside from a few edge cases, such as log files rolling, the position file provides durability to logs ingested by promtail. This durability currently does not apply for the syslog target, as logs are received over the network. 

One potential implementation to solve this problem is a WAL for syslog data.

slim-bean:
I'm not sure how much a WAL will help though? if promtail is not running it can't receive the logs being sent and will miss them, this problem feels like it would need to be more effectively addressed by the sender.

Given that some batching is done in promtail you would lose those logs in memory, (BTW this would also be the case for reading log files and journal files as well) so we could consider a WAL here possibly, it would at least be a tiny WAL

rfratto:
I guess the question is if we want to provide a guarantee that once you send logs to Promtail, they will eventually be sent to Loki.

It's probably a small problem that the positions file is updated before the batch is sent, but that's another discussion for another day.

rfratto:
Sorry, I see what you're saying now. I'm just talking about Promtail crashing before a batch is sent. I'll update the description to make that clearer. I didn't realize until now that's a problem for all target types, not just syslog.

owen-d:
If I understand this correctly we can solve this problem for file-based targets (which can be reread) by just deferring updating the positions file until we've seen successful response codes from loki.

rfratto:
That'd solve most problems for file-based targets, unless the log files are rotated after we read them. Admittedly, it's probably really unlikely to Promtail to need to recover after a log file rotated.

adityacs:
Thanks @rfratto  for opening this discussion. Few more points to add to this

It's better that we have some kind of fallback mechanism in Promtail. If sending the `batch` is not successful we delete the batch or create a new batch. So, we lose the logs because of this, especially when using docker driver. Also, as @rfratto mentioned we lose the logs even when promtail crashes. 

It would be good to support some fallback mechanism so that we don’t lose the logs if they were unable to be sent to Loki instance. We can support `filesystem` and `in memory` buffer. With `filesystem` the batches can be compressed and stored in a filesystem. With `in memory` buffer, we can hold the logs till we reach the buffer maximum. We can consider WAL as well here. Even fluent bit has `filesystem` and `buffer` features https://docs.fluentbit.io/manual/administration/buffering-and-storage

[...]

PR #7993, opened 2022-12-21
Add WAL support for Promtail

thepalbi:
## What this PR does?

This PR adds a WAL for Promtail. The work done on this PR is based on the foundations layed in this other PR https://github.com/grafana/loki/pull/6325.

## Why we need this?

Right now Promtail does not provide good delivery guarantees when there's some failures. Therefore, entries scraped by targets, and waiting to be sent to Loki by remote write are lost if some failure occurs. 

This PR solves this by adding a WAL in-between the scraped entries pipeline, and the remote write client.

## Architecture

The WAL is implemented in such a way that all logic is contained and abstracted inside the client that performs remote_writes to loki when entries are received. 

The tenets that were taken into accounts for this implementation are:
- The write side of the WAL should be fully decoupled from the read side.
- The WAL should provide some durability guarantees, in that if an entry reaches the client, it will eventually be sent.
- The WAL should be able to cleanup already sent entries.

Depending if the WAL is disabled or enabled, this will act in one of two ways:

### WAL disabled
```
clients:
  - url: https://123:asd@logs-prod3.grafana.net/loki/api/v1/push
    wal:
      enabled: false
      dir: ble
```
If the WAL is disabled, the client will work as always. When an entry is read from the channel where all targets and pipelines push to, it's added to a batch. Once the batch reaches certain size or age threshold, it's sent.
Note that the code has been refactored so that the main loop is generic enough to support both usages, but the non-WAL implementation should be the same.

### WAL enabled

```
clients:
  - url: https://123:asd@logs-prod3.grafana.net/loki/api/v1/push
    wal:
      enabled: true
      dir: /tmp/wal
```

The diagram below shows the main running actors that take action in how the WAL interacts inside Promtail. Basically, we'll have three main components:
- **Write side**: This takes part in the same main loop as in the WAL-disabled mode. It does two main things when an entry is received: first it logs it to the corresponding WAL; after it check if sufficient entries have been seen for a batch to be sent. If that's the case, a new segment is cut in the WAL (the read side will react to this event).
- **WAL**: This component is the WAL that sits in the middle between the write and send side. It re-uses prometheus WAL implementation, and Loki's WAL entries format.
- **Read side**: In order to have a fully decoupled read side, the implementation takes an approach similar to the [prometheus WAL watcher](https://github.com/prometheus/prometheus/blob/main/tsdb/wlog/watcher.go). When a WAL is created for a `(client config, tenant)` pair, a watched is launched for that WAL. This watched keeps listening to segments of a WAL while they are being written. When entries are read, they are added to a WIP batch, and once a segment is cut, the batch is sent.

Note that to be able to keep track of the batch size/age limits, and be able to cleanup WAL data related to entries already sent, a 1 segment to 1 batch relations was chosen. The segment cut event allows the reader side to know when enough data has been read for a batch to be sent, and also, allows that segment to be deleted once the send routine completes successfully.

promtail_wal drawio

## Which issue(s) this PR fixes
- Related to https://github.com/grafana/loki/issues/1540
- Resolves https://github.com/grafana/cloud-onboarding/issues/2569
- Required for https://github.com/grafana/cloud-onboarding/issues/2489

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


clients/pkg/promtail/client/watcher.go
```diff
@@ -0,0 +1,326 @@
...
+		if err := w.watch(currentSegment, currentSegment >= lastSegment); err != nil {
+			return err
+		}
+
+		// For testing: stop when you hit a specific segment.
+		if currentSegment == w.MaxSegment {
+			return nil
+		}
+
+		// we now a new segment has been cut, upon advancing the segment pointer, emit the send batch call
```
cstyan, line 125: This is only true because we're doing a single batch per segment, correct?

cstyan (commented):
Just a quick review for now, I'll take another pass later this week. It would be helpful if you could clean up any unnecessary comments or whitespace changes just to clean up the diff view for the next reviews.

I've left a few comments, but I have some thoughts as well.

Right now we're only writing a single batch to each segment, and we also have a WAL dir per client/tenant combo. Just thinking how this could potentially be problematic; we could run into IO/inode problems, or segment file ID #'s problems since we send a lot of batches over time.

The current implementation is the simplest though, and probably achieves our goals of just having resiliency for individual batch sends.

Should we take things further towards prometheus' approach; a single WAL with a reader (per client/tenant combo, in prometheus the reader is just per client), and some kind of WAL cleanup routine on an interval?

thepalbi:
Thanks for taking a look @cstyan , I wanted some eyes before doing some further work.

> It would be helpful if you could clean up any unnecessary comments or whitespace changes just to clean up the diff view for the next reviews.

Sure, I'll do another cleanup pass so the diff is easier.

> Right now we're only writing a single batch to each segment, and we also have a WAL dir per client/tenant combo. Just thinking how this could potentially be problematic; we could run into IO/inode problems, or segment file ID #'s problems since we send a lot of batches over time.

Yhea I had that in mind when taking the 1 segment per batch decision. The main driver is that's the only way I could do a cleanup of a segment if and only if that entries contained in it were successfully sent. That's the purpose for the `SegmentEnd` callback on the `WatcherConsumer` interface. 

Another idea I had is to keep the same writer/watcher architecture, but for them to be slightly different:
- **Writer**: Instead of keeping a shallow copy of the batches to keep track of the size / age, and "cutting a segment" to trigger a batch sent, the writer side would just log to the WAL.
- **Watcher**:  The watcher would read from the WAL, and do the whole batch building process. Once the age/size thresholds are reached, the batch would be sent.

The only problem with this is I'm not sure on how the watcher side would have certainty that all entries in a segment have been sent, in order to clean it up. Any ideas? 

> Should we take things further towards prometheus' approach; a single WAL with a reader (per client/tenant combo, in prometheus the reader is just per client), and some kind of WAL cleanup routine on an interval?

I think this is a starting point. I think we could keep a single WAL per client config, using the `UserID` field in [Loki's WAL record](https://github.com/grafana/loki/blob/bc10888c47d21a523764778c2661abd67097e4c4/pkg/ingester/encoding.go#L36). What do you think?

clients/pkg/promtail/client/client.go
```diff
@@ -484,3 +671,116 @@ func (c *client) UnregisterLatencyMetric(labels prometheus.Labels) {
...
+	return nil
+}
+
+func (c *clientConsumer) ConsumeEntries(samples ingester.RefEntries) error {
+	var entry api.Entry
+	if l, ok := c.series[uint64(samples.Ref)]; ok {
+		entry.Labels = l
+		for _, e := range samples.Entries {
+			entry.Entry = e
+			// Using replay since we know the batch needs to be sent once the segment ends
```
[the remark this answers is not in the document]
thepalbi: So `replay` does basically the same action as `batch.add`, but without doing some checks like maximum batch size, or max stream size, since these have already been performed by the WAL writing side.

clients/pkg/promtail/client/watcher.go
thepalbi: Yes on the sending the batch, but if that line is reached we know the previous segment has been read fully, so we are good to call the `SegmentEnd` callback. Simplifying the comment since the watcher's `SegmentEnd` callback can be used for something other.

clients/pkg/promtail/client/client.go
```diff
@@ -287,14 +467,15 @@ func (c *client) run() {
 			// If adding the entry to the batch will increase the size over the max
 			// size allowed, we do send the current batch and then create a new one
 			if batch.sizeBytesAfter(e) > c.cfg.BatchSize {
-				c.sendBatch(tenantID, batch)
-
-				batches[tenantID] = newBatch(c.maxStreams, e)
+				_ = dispatchBatch(tenantID, batch)
```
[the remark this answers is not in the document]
thepalbi: So two answers:
> we want to ensure the batch has been sent successfully before creating a new one and clearing the previous WAL segment

Sending the batch, and if successful deleting the corresponding segment occurs on the **WAL reading** side, that is the watcher and the configured callbacks. For example, [here's](https://github.com/grafana/loki/blob/05553737f80d0a5dda4de421381ea01aa4fdce0f/clients/pkg/promtail/client/client.go#L779-L785) the piece of code under the `SegmentEnd` callback that sends the batch and if no error occurs, deletes the batch.

In case of failure when sending the batch (considering the `sendBatch` does some retries depending on the client config), the segment won't be deleted and would be picked up by the next Promtail restart.

> we should be checking the error here and later in run as well, should we not?
I think we could be logging the error, but in the writer side if the `dispatch` fails it means that cutting a segment in the WAL fails. Therefore, all follow up entries will be written to the same segment, therefore not lost. Only problem with this is that the batch that corresponds to that segment could keep growing, and once it's read, the send operation failing.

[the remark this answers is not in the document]
thepalbi: Logged error on failed batch `dispatches`

cstyan:
> Yhea I had that in mind when taking the 1 segment per batch decision. The main driver is that's the only way I could do a cleanup of a segment if and only if that entries contained in it were successfully sent. That's the purpose for the SegmentEnd callback on the WatcherConsumer interface.

Right, this is an exact "problem" we already have with Prometheus Remote Write. 

In Prometheus, TSDB has ownership over the WAL and the Watcher/Remote Write are just consuming the data from it. If a remote write falls behind in sending out samples, and TSDB truncates the WAL to clean up old data that it has already committed to disk in a block, that data is lost to remote write forever. Granted, that doesn't happen often in reality and we would probably want to have some kind of time based commitment of WAL data retention anyways so it doesn't grow in an unbounded way.

We could discuss a method for being able to indicate from a client to the rest of the codebase that all data from a given segment # has been sent, making it safe to clean up that segment. That could be a follow up PR imo.

> Another idea I had is to keep the same writer/watcher architecture, but for them to be slightly different:
> - Writer: Instead of keeping a shallow copy of the batches to keep track of the size / age, and "cutting a segment" to trigger a batch sent, the writer side would just log to the WAL.
> - Watcher: The watcher would read from the WAL, and do the whole batch building process. Once the age/size thresholds are reached, the batch would be sent.

I think the WAL Write and Watcher architecture you're describing here is actually what we do want. Coupling the writing to the WAL to an individual client, or more accurately the passing of a sample from the scrape code to a client, does feel a bit like a smell. It does feel like at least the writing to the WAL should happen via one single interface, and reading from the WAL to build a batch can happen via another. If the Client struct happens to implement both, or include a watcher that it starts, that's fine. Does that make sense?

clients/pkg/promtail/client/client.go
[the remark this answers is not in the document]
cstyan: Can you update the comment to include that detail, so that it's more clear that at the moment we're coupling this side effect of the WAL writing behaviour.

I'm not sure we would want to keep things as coupled as they are currently (side effect of how Cyril/I first implemented the WAL) but it should be clear that this call would need to be changed to `batch.add` when we decouple things in the future.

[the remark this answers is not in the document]
thepalbi: Adding comment, and agree that this should be `batch.add` once we decouple the writing side completely from the batch making.

clients/pkg/promtail/client/client.go
```diff
@@ -267,15 +436,28 @@ func (c *client) run() {
...
+			// after writing to wal, keep track of batch size to know when to cut the batch. That is either
+			// send it, or cut a segment in the wal
 			batch, ok := batches[tenantID]
 
 			// If the batch doesn't exist yet, we create a new one with the entry
 			if !ok {
-				batches[tenantID] = newBatch(c.maxStreams, e)
+				b := c.newBatch()
+				batches[tenantID] = b
+				_ = addEntryToBatch(b, e)
```
kgeckhart, line 460: Here and https://github.com/grafana/loki/blob/daa332547b94a0f323592b2b6fcfabae30e7e1ed/clients/pkg/promtail/client/client.go#L477-L477 we ignore the `addEntryToBatch` error. But further https://github.com/grafana/loki/blob/daa332547b94a0f323592b2b6fcfabae30e7e1ed/clients/pkg/promtail/client/client.go#L483-L483 checks and logs the error. Should we be checking and logging at all of these?

If we should be checking the error it might be useful to refactor so there's only a single call to `addEntryToBatch` at the end with the if statements ensuring the batch is sent/created ahead of it.

thepalbi:
@cstyan 

> We could discuss a method for being able to indicate from a client to the rest of the codebase that all data from a given segment # has been sent, making it safe to clean up that segment. That could be a follow up PR imo.

Agree, I'm down on developing / discussing this idea in a follow up PR.

> If the Client struct happens to implement both, or include a watcher that it starts, that's fine. Does that make sense?

So at the moment everything WAL-regarding is implemented behind the remote_write client. I think that's something we should keep since it both keeps the changes contained into a single component, which makes impact of changes smaller; and could be re-used by the Grafana Agent team IIUC. That's the approach you mean by the above right?

> I think the WAL Write and Watcher architecture you're describing here is actually what we do want. Coupling the writing to the WAL to an individual client, or more accurately the passing of a sample from the scrape code to a client, does feel a bit like a smell. It does feel like at least the writing to the WAL should happen via one single interface, and reading from the WAL to build a batch can happen via another. 

As for this, I agree. I think the flow from each scraping target should be:
```
writer side:

target 1-- | 
target 2--| --> single channel where all write to --> loop that reads entries and write to WAL
...              |
target n--|

reader side:
watcher --> fan out to every other client
```
while right now we have:

```
writer side:

target 1--|
target 2--| --> single channel where all write to --> loop that reads entries and handle in client
...             |
target n--|

client i's channel --> log in client/tenant specific WAL

reader side:
client/tenant specific watcher --> build and send batch
```

If we do the former, we'd be having one single WAL for all clients and tenants, which makes sense, but won't let us keep the design in this PR. Do you think we should start with this PR's design, and continue building the discussed solution in follow ups?

clients/pkg/promtail/client/watcher.go
```diff
@@ -0,0 +1,333 @@
...
+
+	currentSegment := lastSegment
+	level.Debug(w.logger).Log("msg", "Tailing WAL", "currentSegment", currentSegment, "lastSegment", lastSegment)
+	for !isClosed(w.quit) {
+		w.metrics.currentSegment.WithLabelValues(w.name).Set(float64(currentSegment))
+		level.Debug(w.logger).Log("msg", "Processing segment", "currentSegment", currentSegment)
+
+		// On start, we have a pointer to what is the latest segment. On subsequent calls to this function,
+		// currentSegment will have been incremented, and we should open that segment.
+		if err := w.watch(currentSegment, currentSegment >= lastSegment); err != nil {
```
kgeckhart, line 114:
```suggestion
		if err := w.watch(currentSegment, true); err != nil {
```
Isn't `currentSegment >= lastSegment` always going to be true or am I misreading the code here? `currentSegment` starts out equal to `lastSegment` and only ever gets larger while lastSegment seems to stay the same.

kgeckhart (commented):
I had two questions so far, I might have more questions later I'm still trying to digest it

clients/pkg/promtail/client/watcher.go
thepalbi: Uhh you are right! That bit comes from Prometheus'es WAL watcher in which I based this implementation. In their case, the watcher itself replays the WAL, but in our case, we just watch from the last segment onwards. Cleaning up here all logic related to replaying from the watcher itself. Should simplify a little bit.

clients/pkg/promtail/client/client.go
thepalbi: Great idea! Refactoring the code to have a single call to `addEntryToBatch`

thepalbi:
@cstyan moving this to a ready state as I think it has reached some maturity. When you have some minute, can you answer https://github.com/grafana/loki/pull/7993#issuecomment-1376260433. I think that should give us next steps

clients/pkg/promtail/client/client.go
```diff
@@ -484,3 +681,130 @@ func (c *client) UnregisterLatencyMetric(labels prometheus.Labels) {
...
+			entry.Entry = e
+			// Using replay since we know the batch needs to be sent once the segment ends. This coupling happens because
+			// the bucketing of incoming entries into batches is taken care by the WAL writer. This should be changed
+			// to `batch.add` once we decouple the batch creation from the WAL writer completely. See
+			// https://github.com/grafana/loki/pull/7993#discussion_r1062706323 for details.
+			c.currentBatch.replay(entry)
+		}
+	} else {
+		// if series is not present for sample, just log for now
+		level.Debug(c.logger).Log("series for sample not found")
```
cstyan, line 797: we should potentially also have a metric for this

clients/pkg/promtail/client/watcher.go
cstyan: I'm still a bit confused here as well. We need to replay the WAL in case anything is unsent and also to cache labels data for streams that already existed prior to the most recent start up. I'm not seeing how we do that caching and reuse of stream ref IDs.

cstyan:
> So at the moment everything WAL-regarding is implemented behind the remote_write client. I think that's something we should keep since it both keeps the changes contained into a single component, which makes impact of changes smaller; and could be re-used by the Grafana Agent team IIUC. That's the approach you mean by the above right?

I'm not sure I agree. I feel like this coupling of the WAL to a client is part of what's making this PR a bit harder for me to review. At the moment it's hard for me to follow where the line is drawn between code that receives log lines from a discovery target, writing of that data to the WAL, the routine that then reads from the WAL, and the eventual building of a batch in the client itself.

Let me know if this makes sense, it feels like we should be organizing the flow something like this:

A sample is scraped by a discovery target, which should have an interface to something that can receive a log line. Right now this interface would be implemented by the base target itself. We should implement the interface again for writing to the WAL, and the struct that implements the interface should actually fan out writes to multiple WALs per tenant ID.

We would then have a routine that reads the WAL and passes the log lines to the client, again via an interface so we don't couple things too closely again. However, we need to start one of these reading routines per client, because we don't want reading and sending of more data to all clients to be blocked if there's a connection issue for a single client.

TL; DR, In general I think we're heading in the right direction but something here still feels too coupled/hard to follow and reason about. Perhaps we should break this current implementation into a few PRs; one to write to the WAL, and a second to read from it and pass that data to clients. We could then have a third to implement the other feature we discussed earlier.

thepalbi:
@cstyan 

> At the moment it's hard for me to follow where the line is drawn between code that receives log lines from a discovery target, writing of that data to the WAL, the routine that then reads from the WAL, and the eventual building of a batch in the client itself.

Yhea, I agree the model is a bit confusing to be honest. The main driver to keep everything enclosed in the client itself is to keep impact of changes minimal. Regardless, if we fully decouple it by changing where each scraping target writes to for example, I think the model will be much cleaner.

Let's go that route. Also doing it that way it will help splitting the PR as you mentioned. I'm thinking first on how the interfaces will interact and what it will look like. Once we agree I'll start the implementation. Should we keep going with the discussion in this PR? If not I can open an issue, so we discuss over there + track the upcoming PRs.

Leaving a question in the meantime:

> and the struct that implements the interface should actually fan out writes to multiple WALs per tenant ID.

If we go the fully decoupled route, with targets writing entries directly to WAL, what's the purpose of keeping one WAL per tenant? If we do one WAL per tenant, and we want to avoid a single client blocking all others, we'd have to keep one watcher per tenant WAL, per client. That is if we got N tenants and M client, N*M watchers running. Not saying there's anything bad with that, just not seeing the point of keeping many WALs right now.

clients/pkg/promtail/client/watcher.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/prometheus/prometheus/tsdb/wlog"
+
+	"github.com/grafana/loki/pkg/ingester"
+)
+
+const (
+	readPeriod         = 10 * time.Millisecond
+	checkpointPeriod   = 5 * time.Second
+	segmentCheckPeriod = 100 * time.Millisecond
+)
```
rfratto, line 24: Something to be aware of here:

I've found that this adds a 4-5% CPU usage to the baseline resource consumption of the process. See prometheus/prometheus#11625.

The CPU utilization is negligible if something is writing to the WAL frequently enough, but if writes to the WAL are infrequent (i.e., slow moving log files), then the CPU usage will be higher than expected.

This could impact users with limited hardware capabilities, like Raspberry Pis.

thepalbi:
Closing this PR as we are taking a multi-PR approach, tracked in this issue: https://github.com/grafana/loki/issues/8197

Issue #7996, opened 2022-12-22
promtail `cri` stage should not insert newlines while joining partial log lines


umezawatakeshi:
**Describe the bug**
[`cri` stage](https://grafana.com/docs/loki/latest/clients/promtail/stages/cri/) is able [to join partial log lines](https://github.com/grafana/loki/pull/6177) which are splitted by cri-o and to treat them as single log entry. However, it inserts newlines `'\n'` while joining lines by mistake. They should become single line.

**To Reproduce**
1. Start Loki v2.7.1
2. Start Promtail v2.7.1 and tail Pod logs. `cri` stage should be enabled.
3. Deploy a Pod which outputs single long (~20KB) log line. The line does not contain newline, of course.
4. Query that log. Single log entry is returned but it contains newline.

**Expected behavior**
The returned log entry should not contain newline.

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: helm

**Screenshots, Promtail config, or terminal output**

umezawatakeshi:
This issue was accidentally closed by completion of my PR in my organization...

PR #7997, opened 2022-12-22
fix(promtail): Fix cri tags extra new lines.

kavirajk:
When re-creating single lines from multiple partial lines, `\n` should be avoided to join the partial lines.

Related: https://github.com/grafana/loki/pull/6177#discussion_r976113761

Example
if Original log line is 
```
{"a": "very long json field split in two lines"}
```

And got split into partial lines via CRI
```
2022-09-20T17:51:01.214529932Z stdout P {"a": "very long json field 
2022-09-20T17:51:01.214697597Z stdout F split in two lines"}
```
The result should be without \n when joining both of those lines.

Fixes: https://github.com/grafana/loki/issues/7996

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/clients/pkg/logentry/stages/extensions_test.go
+++ b/clients/pkg/logentry/stages/extensions_test.go
@@ -107,48 +107,31 @@ func TestCRI_tags(t *testing.T) {
 		{
 			name: "tag P",
 			lines: []string{
-				"2019-05-07T18:57:50.904275087+00:00 stdout P partial line 1",
-				"2019-05-07T18:57:50.904275087+00:00 stdout P partial line 2",
+				"2019-05-07T18:57:50.904275087+00:00 stdout P partial line 1 ",
+				"2019-05-07T18:57:50.904275087+00:00 stdout P partial line 2 ",
 				"2019-05-07T18:57:55.904275087+00:00 stdout F log finished",
 				"2019-05-07T18:57:55.904275087+00:00 stdout F another full log",
 			},
 			expected: []string{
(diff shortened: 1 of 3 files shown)


clients/pkg/logentry/stages/extensions_test.go
```diff
@@ -107,50 +107,33 @@ func TestCRI_tags(t *testing.T) {
 		{
 			name: "tag P",
 			lines: []string{
-				"2019-05-07T18:57:50.904275087+00:00 stdout P partial line 1",
-				"2019-05-07T18:57:50.904275087+00:00 stdout P partial line 2",
+				"2019-05-07T18:57:50.904275087+00:00 stdout P partial line 1 ",
```
MasslessParticle, line 110: Do partial lines come in with a space at the end or should we be joining with `" "`?

MasslessParticle (commented):
This looks good. I'm concerned about the spacing, though.

clients/pkg/logentry/stages/extensions_test.go
kavirajk: I don't think we have to handle the spaces here, cri will just split without any adding any extra characters. We should join as it is.

MasslessParticle: Ok, it looks like cri-o splits lines wherever. Thanks!

MasslessParticle (approved):
lgtm

PR #8002, opened 2022-12-23
Update log queries regex documentation

ssncferreira:

Update log queries regex documentation with a clarification on the difference between the newline character in:
* label matching operator: `.*` does not match newlines
* line filter expressions: `.*` matches newlines

**Checklist**
- [x] Documentation updated


MichelHollands (approved):
LGTM

PR #8025, opened 2023-01-02
make query-frontend service grpc lb friendly

sandeepsukhani:
In PR #7966, I made changes to Query Frontend service. Part of it was creating a new service.
To make the service properly load balance grpc requests, we need to use `grpclbServiceFor`, but since we do not use grpc in QF, I went with the usual method, i.e. I used `serviceFor`. Thinking more about it, I think there is no harm in keeping it grpc LB friendly, and it would save us from any possible bugs if we make QF accept queries via grpc.


kavirajk (approved):
LGTM. Does it need any upgrade guide entries?

sandeepsukhani:
Thanks for the review Kavi!

> LGTM. Does it need any upgrade guide entries?

I have added it to the upgrade guide already in the previous PR.

PR #8028, opened 2023-01-02
Fix chunks-inspect reports zstd compressed chunks as lz4-256k

jfolz:
Make the `chunks-inspect` command correctly report Zstandard compressed chunks as `zstd` encoding.

The `chunks-inspect` command falsely reports Zstandard compressed chunks as `lz4-256k`.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/cmd/chunks-inspect/loki.go
+++ b/cmd/chunks-inspect/loki.go
@@ -43,7 +43,7 @@ var (
 	enclz4_1M   = Encoding{code: 6, name: "lz4-1M", readerFn: func(reader io.Reader) (io.Reader, error) { return lz4.NewReader(reader), nil }}
 	enclz4_4M   = Encoding{code: 7, name: "lz4-4M", readerFn: func(reader io.Reader) (io.Reader, error) { return lz4.NewReader(reader), nil }}
 	encFlate    = Encoding{code: 8, name: "flate", readerFn: func(reader io.Reader) (io.Reader, error) { return flate.NewReader(reader), nil }}
-	encZstd     = Encoding{code: 9, name: "lz4-256k", readerFn: func(reader io.Reader) (io.Reader, error) {
+	encZstd     = Encoding{code: 9, name: "zstd", readerFn: func(reader io.Reader) (io.Reader, error) {
 		r, err := zstd.NewReader(reader)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/chunks-inspect/loki.go
```diff
@@ -43,7 +43,7 @@ var (
 	enclz4_1M   = Encoding{code: 6, name: "lz4-1M", readerFn: func(reader io.Reader) (io.Reader, error) { return lz4.NewReader(reader), nil }}
 	enclz4_4M   = Encoding{code: 7, name: "lz4-4M", readerFn: func(reader io.Reader) (io.Reader, error) { return lz4.NewReader(reader), nil }}
 	encFlate    = Encoding{code: 8, name: "flate", readerFn: func(reader io.Reader) (io.Reader, error) { return flate.NewReader(reader), nil }}
-	encZstd     = Encoding{code: 9, name: "lz4-256k", readerFn: func(reader io.Reader) (io.Reader, error) {
+	encZstd     = Encoding{code: 9, name: "zstd", readerFn: func(reader io.Reader) (io.Reader, error) {
```
DylanGuedes, line 46: doesn't that mean that we won't be inspecting `lz4-256k` chunks anyway? I'm not sure if `lzy4-256` are compatible but in case they are, maybe the best thing here would be to duplicate this to have both supports.

DylanGuedes (changes requested):
Hey, thank you for your contribution!
Added a question, lmk if it makes sense.

cmd/chunks-inspect/loki.go
jfolz: I think this was just a copy & paste error. A decoder for `lz4-256k` is added in [line 42](https://github.com/grafana/loki/blob/43c11a9a287afba695f7502367d1b7c41ab4e612/cmd/chunks-inspect/loki.go#L42).

DylanGuedes: ohhh good call, thank you. Approved!

DylanGuedes (approved):
lgtm!

owen-d (approved):
Thanks, nice catch.

PR #7839, opened 2022-12-02
Operator: Configure Alertmanager per-tenant

aminesnow:
**What this PR does / why we need it**:
This PR adds the ability to configure the operator to send alerts to an Alertmanager per-tenant. In addition, when the operator is deployed on OCP, it automatically configures the `application` tenant to send alerts to the `user-workload` Alertmanager if it's enabled. 

**Special notes for your reviewer**:
Depends on: #7805 

**Checklist**
- [x] Reviewed the `CONTRIBUTING.md` guide
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #8038, opened 2023-01-04
operator: Add watch on the Alertmanager in OCP's user-workload-monitoring namespace

aminesnow:
Following #7839, this PR adds a watch on the Alertmanager in the user-workload-monitoring namespace that is used when the operator is running on an OpenShift cluster. 

@periklis let me know if this is what you had in mind.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -5,6 +5,7 @@
 - [8087](https://github.com/grafana/loki/pull/8087) **xperimental**: Fix status not updating when state of pods changes
 - [8068](https://github.com/grafana/loki/pull/8068) **periklis**: Use lokistack-gateway replicas from size table
 - [8068](https://github.com/grafana/loki/pull/8068) **periklis**: Use lokistack-gateway replicas from size table
+- [8038](https://github.com/grafana/loki/pull/8038) **aminesnow**: Add watch on the Alertmanager in OCP's user-workload-monitoring namespace
 - [7839](https://github.com/grafana/loki/pull/7839) **aminesnow**: Configure Alertmanager per-tenant
 - [7910](https://github.com/grafana/loki/pull/7910) **periklis**: Update Loki operand to v2.7.1
 - [7815](https://github.com/grafana/loki/pull/7815) **periklis**: Apply delete client changes for compat with release-2.7.x
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Main
 
-- [8173](https://github.com/grafana/loki/pull/8173) **periklis**: Remove custom webhook cert mounts for OLM-based deployment (OpenShift)
 - [8001](https://github.com/grafana/loki/pull/8001) **aminesnow**: Add API validation to Alertmanager header auth config
+- [8173](https://github.com/grafana/loki/pull/8173) **periklis**: Remove custom webhook cert mounts for OLM-based deployment (OpenShift)
 - [8087](https://github.com/grafana/loki/pull/8087) **xperimental**: Fix status not updating when state of pods changes
 - [8068](https://github.com/grafana/loki/pull/8068) **periklis**: Use lokistack-gateway replicas from size table
 - [8068](https://github.com/grafana/loki/pull/8068) **periklis**: Use lokistack-gateway replicas from size table
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -1,11 +1,11 @@
 ## Main
 
-- [8001](https://github.com/grafana/loki/pull/8001) **aminesnow**: Add API validation to Alertmanager header auth config
+- [8038](https://github.com/grafana/loki/pull/8038) **aminesnow**: Add watch on the Alertmanager in OCP's user-workload-monitoring namespace
 - [8173](https://github.com/grafana/loki/pull/8173) **periklis**: Remove custom webhook cert mounts for OLM-based deployment (OpenShift)
+- [8001](https://github.com/grafana/loki/pull/8001) **aminesnow**: Add API validation to Alertmanager header auth config
 - [8087](https://github.com/grafana/loki/pull/8087) **xperimental**: Fix status not updating when state of pods changes
 - [8068](https://github.com/grafana/loki/pull/8068) **periklis**: Use lokistack-gateway replicas from size table
 - [8068](https://github.com/grafana/loki/pull/8068) **periklis**: Use lokistack-gateway replicas from size table
-- [8038](https://github.com/grafana/loki/pull/8038) **aminesnow**: Add watch on the Alertmanager in OCP's user-workload-monitoring namespace
 - [7839](https://github.com/grafana/loki/pull/7839) **aminesnow**: Configure Alertmanager per-tenant
 - [7910](https://github.com/grafana/loki/pull/7910) **periklis**: Update Loki operand to v2.7.1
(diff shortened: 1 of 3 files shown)


operator/controllers/loki/lokistack_controller.go
```diff
@@ -178,7 +187,8 @@ func (r *LokiStackReconciler) buildController(bld k8s.Builder) error {
 		Owns(&rbacv1.ClusterRole{}, updateOrDeleteOnlyPred).
 		Owns(&rbacv1.ClusterRoleBinding{}, updateOrDeleteOnlyPred).
 		Owns(&rbacv1.Role{}, updateOrDeleteOnlyPred).
-		Owns(&rbacv1.RoleBinding{}, updateOrDeleteOnlyPred)
+		Owns(&rbacv1.RoleBinding{}, updateOrDeleteOnlyPred).
+		Watches(&source.Kind{Type: &corev1.Service{}}, r.enqueueAllLokiStacksHandler(), createUpdateOrDeletePred)
```
periklis, line 191: :scream: Why is this possible at all? Using the same source Kind for owned and watched resources?

TL;DR; Although the controller runtime API seems to allow this I suggest to dive into the code behind this to see what happens with caching, listers and informers when you register the same source Kind twice. This seems odd and dangerous.

Besides this the `Watches` clause seems to be also very eager sort of that any service across the cluster will trigger a reconciliation loop for all LokiStack instances. Can't we limit this to the two particular OCP services and only for the LokiStacks with tenant modes `openshift-logging` and `openshift-network`?

periklis (changes requested):

operator/controllers/loki/lokistack_controller.go
aminesnow: I agree that it's reckless to have all the SVCs trigger the reconciliation when we are interested in only one specific SVC, so I'll edit this to be more specific. 

However, I don't see an issue with having an Owns and a Watches on the same resource; Owns is just a Watches under the hood, and if I'm not mistaken, we can have multiple Watches on the same resource in order to react to different events differently. If it's not possible then please enlighten me as I might have misunderstood the behavior.

operator/controllers/loki/lokistack_controller.go
```diff
@@ -224,3 +235,31 @@ func (r *LokiStackReconciler) enqueueAllLokiStacksHandler() handler.EventHandler
...
+		if obj.GetName() == openshift.MonitoringSVCOperated && obj.GetNamespace() == openshift.MonitoringUserwWrkloadNS {
+
+			for _, stack := range lokiStacks.Items {
+				requests = append(requests, reconcile.Request{
+					NamespacedName: types.NamespacedName{
+						Namespace: stack.Namespace,
+						Name:      stack.Name,
+					},
+				})
+			}
```
periklis, line 258: I suggest to omit any stack that has tenant mode other than `openshift-logging` or `openshift-network`. These are the only two tenants we reconcile OpenShift related Alertmanager configuration.

operator/controllers/loki/lokistack_controller.go
```diff
@@ -224,3 +235,31 @@ func (r *LokiStackReconciler) enqueueAllLokiStacksHandler() handler.EventHandler
...
+	ctx := context.TODO()
+	return handler.EnqueueRequestsFromMapFunc(func(obj client.Object) []reconcile.Request {
+		lokiStacks := &lokiv1.LokiStackList{}
+		if err := r.Client.List(ctx, lokiStacks); err != nil {
+			r.Log.Error(err, "Error getting LokiStack resources in event handler")
+			return nil
+		}
+		var requests []reconcile.Request
+
+		if obj.GetName() == openshift.MonitoringSVCOperated && obj.GetNamespace() == openshift.MonitoringUserwWrkloadNS {
```
periklis, line 249: What about the AM in the `openshift-monitoring` namespace?

periklis (changes requested):

operator/controllers/loki/lokistack_controller.go
aminesnow: what about it?

periklis: Don't we need to check for it here?

aminesnow: maybe, maybe not, but this is not the PR for that

operator/controllers/loki/lokistack_controller.go
aminesnow: done

operator/controllers/loki/lokistack_controller.go
periklis: Why not?

aminesnow: I'd rather get this one right and merged then do another PR for the other AM.

periklis: I see this being the same for both AM as both services have the name `alertmanager-operated` but differ only in a namespace.

periklis: :shrug:

operator/controllers/loki/lokistack_controller.go
```diff
@@ -224,3 +235,34 @@ func (r *LokiStackReconciler) enqueueAllLokiStacksHandler() handler.EventHandler
...
+			for _, stack := range lokiStacks.Items {
+				if stack.Spec.Tenants.Mode == lokiv1.OpenshiftLogging ||
+					stack.Spec.Tenants.Mode == lokiv1.OpenshiftNetwork {
+					requests = append(requests, reconcile.Request{
+						NamespacedName: types.NamespacedName{
+							Namespace: stack.Namespace,
+							Name:      stack.Name,
+						},
+					})
+				}
```
periklis, line 260: Need to check here if  `stack.Spec.Tenants` is not nil.

periklis (changes requested):

operator/controllers/loki/lokistack_controller.go
aminesnow: good catch, thnx!

operator/CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
 ## Main
 
-- [8173](https://github.com/grafana/loki/pull/8173) **periklis**: Remove custom webhook cert mounts for OLM-based deployment (OpenShift)
 - [8001](https://github.com/grafana/loki/pull/8001) **aminesnow**: Add API validation to Alertmanager header auth config
+- [8173](https://github.com/grafana/loki/pull/8173) **periklis**: Remove custom webhook cert mounts for OLM-based deployment (OpenShift)
 - [8087](https://github.com/grafana/loki/pull/8087) **xperimental**: Fix status not updating when state of pods changes
 - [8068](https://github.com/grafana/loki/pull/8068) **periklis**: Use lokistack-gateway replicas from size table
 - [8068](https://github.com/grafana/loki/pull/8068) **periklis**: Use lokistack-gateway replicas from size table
+- [8038](https://github.com/grafana/loki/pull/8038) **aminesnow**: Add watch on the Alertmanager in OCP's user-workload-monitoring namespace
```
periklis, line 8: You pulled in some entries through the last rebase. In addition Line8 should be on the top.

periklis (approved):

operator/CHANGELOG.md
aminesnow: Got confused and moved the wrong line to the top :sweat_smile:

periklis (approved):

PR #8057, opened 2023-01-06
Helm: Update tokengen to only create kubectl secret if admin token file does not exist

ssncferreira:
Update tokengen to only create kubectl secret if admin token file does not exist.

When installing GEL with the grafana-loki helm chart (with enterprise enabled), if you uninstall it and reinstall it again it fails because the enterprise-logs-admin-token secret already exists:
```
Defaulted container "create-secret" out of: create-secret, loki (init)
error: failed to create secret secrets "enterprise-logs-admin-token" already exists
```

The solution consists in only creating the kubectl secret if the token file does not yet exist.

### DIFF
--- a/production/helm/loki/templates/tokengen/job-tokengen.yaml
+++ b/production/helm/loki/templates/tokengen/job-tokengen.yaml
@@ -75,7 +75,7 @@ spec:
           image: {{ include "loki.kubectlImage" . }}
           imagePullPolicy: {{ .Values.kubectlImage.pullPolicy }}
           command:
-            - /bin/bash
+            - ! test -f /shared/admin-token && /bin/bash
             - -euc
             - |
               kubectl create secret generic "{{ include "enterprise-logs.adminTokenSecret" . }}" --from-file=token=/shared/admin-token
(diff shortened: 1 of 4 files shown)


jeschkies (approved):
🤯 I've just mentioned yesterday that the Tippen generator should be idempotent.

How is the check working? Is the token persisted to disk across the lifetime of the pod?

production/helm/loki/templates/tokengen/job-tokengen.yaml
```diff
@@ -75,7 +75,7 @@ spec:
           image: {{ include "loki.kubectlImage" . }}
           imagePullPolicy: {{ .Values.kubectlImage.pullPolicy }}
           command:
-            - /bin/bash
+            - ! test -f /shared/admin-token && /bin/bash
```
jeschkies, line 78: Trevor just explained it to me. The init container job would not create the file if it's already in the object storage. Could you add this as a comment?

ssncferreira:
Closing in favor of: https://github.com/grafana/loki/pull/8467

PR #8066, opened 2023-01-06
[lambda-promtail] allow generic logs to be pulled from s3, add a test for s3 log parsing

cstyan:
This has come up a few times; confusion around whether or not lambda-promtail can pull any generic logs from S3, or only AWS Load Balancer logs. If I understand the original implementation, it's only the parsing of the timestamp that ends up dropping logs that don't match the expected format.

I've also added a basic test of parsing an s3 log event.

### DIFF
--- a/tools/lambda-promtail/lambda-promtail/s3.go
+++ b/tools/lambda-promtail/lambda-promtail/s3.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"compress/gzip"
 	"context"
-	"fmt"
 	"io"
 	"regexp"
 	"time"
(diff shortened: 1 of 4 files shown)


tools/lambda-promtail/lambda-promtail/s3.go
```diff
@@ -113,34 +128,38 @@ func getLabels(record events.S3EventRecord) (map[string]string, error) {
...
+		return false, err
+	}
+	return true, nil
+}
 
-		err = parseS3Log(ctx, batch, labels, obj)
-		if err != nil {
-			return err
-		}
+func processS3Event(ctx context.Context, ev *events.S3Event) (bool, error) {
```
chaudum, line 149: What is the reason for changing the return value from `error` to `(bool, error)`?
First, there is no combination where `error` is not `nil` but the `bool` is true. And second, the `bool` return value is discarded in the only usage of `processS3Event()` (main.go L148)

jeschkies (commented):
Could you add some documentation? A request for this feature came up quite a few times.

mmisiewicz-g:
Have a look at: https://github.com/grafana/loki/pull/8254

This change adds parsing for a couple more date formats from different types of AWS S3 logs.
I also had to change the regex for file path/name because AWS WAF logs have hours and minutes in the path.
I dont see any filename regex change here so this will not process AWS WAF logs.

cstyan:
@mmisiewicz-g this should pick up any logs stored in an S3 bucket, it just wouldn't necessarily attach the filename as a label since as you said, I haven't updated the regex.

MasslessParticle:
Closing this for inactivity

PR #8071, opened 2023-01-09
Helm: fix ruler config storage regression

agologan:

While #7528 fixed an issue with null storage when using filesystem it changed the behavior of other storage configs.
Fix reintroduces previous behavior where by default you get the computed `rulerStorageConfig` and appends/overlays `rulerConfig`.
This enables the use of `.Values.storage` adding to `rulerConfig` only the extra parameters one would need e.g. `enable_api: true`.

Other possible solution:
`rulerConfig` was not added as a mutually exclusive option to the default storage config but it could be if we `tpl`-it so users could use something along the lines of (to avoid defining the storage config twice):
```
rulerConfig:
  enable_api: true
  storage: |
    {{- include "loki.rulerStorageConfig" . | nindent 4}}
```
The problem with this approach is there are other places which use the additive pattern and not mutually exclusive like `commonConfig` which would then need to be revisited.

N/A

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


production/helm/loki/templates/_helpers.tpl
```diff
@@ -287,24 +288,18 @@ azure:
...
-{{- define "loki.shouldIncludeRulerConfig" }}
-{{- or (not (empty .Values.loki.rulerConfig)) (.Values.minio.enabled) (eq .Values.loki.storage.type "s3") (eq .Values.loki.storage.type "gcs") }}
-{{- end }}
-
 {{/* Loki ruler config */}}
 {{- define "loki.rulerConfig" }}
-{{- if eq (include "loki.shouldIncludeRulerConfig" .) "true" }}
 ruler:
+  storage:
+    {{- include "loki.rulerStorageConfig" . | nindent 4}}
```
agologan, line 300: moved the storage config first so it can be overriden by rulerConfig if need be

michelefa1988:
I was running into this blocked, and download the helm chart locally did fix the issue

MasslessParticle (approved):

michelefa1988:
@MasslessParticle when can this PR be released? I'm quite blocked without it

agologan:
They usually cut release pretty often, but you can do without it until then by redeclaring the storage config in rulerConfig.
Something along the lines of:
```
rulerConfig:
  enable_alertmanager_v2: true
  notification_timeout: 30s
  ...
  storage:
    type: s3
      s3:
        bucketnames: lokirulerbucket
        insecure: false
        region: us-east-1
        s3forcepathstyle: false
```
to match whatever storage config you have
```
storage:
  bucketNames:
    chunks: lokichunksbucket
    ruler: lokirulerbucket
    admin: lokiadminbucket
  type: s3
  s3:
    region: us-east-1
```

PR #8083, opened 2023-01-10
Add docker-driver pipeline

MichelHollands:

Copy the docker-driver pipeline from CirclCI to drone. This was tested by running with a trigger that creates an image on every branch: see the 'docker-driver' step in the https://drone.grafana.net/grafana/loki/18655 run.
See https://hub.docker.com/r/grafana/loki-docker-driver/tags for the built images: main and main-63360ab.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -758,6 +758,14 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
           }) { when: { event: ['tag'] } },
     ],
   },
+  pipeline('docker-driver') {
+    trigger+: onPRs,
+    steps: [
+      make('docker-driver', container=false) {
+        depends_on: ['clone'],
+      },
+    ],
+  },
 ]
 + [
   lambda_promtail(arch)
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -759,10 +759,19 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
     ],
   },
   pipeline('docker-driver') {
-    trigger+: onTagOrMain,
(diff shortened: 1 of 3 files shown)


salvacorts (approved):

.drone/drone.jsonnet
```diff
@@ -758,6 +758,38 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
...
+        commands: [
+          'make docker-driver-push',
+        ],
+        volumes: [
+          {
+            name: 'docker',
+            path: '/var/run/docker.sock',
+          },
+        ],
+        privileged: true,
```
jeschkies, line 781: I wonder whether we should use something like https://buildah.io/ instead. What do you think?

Makefile
```diff
@@ -421,6 +421,13 @@ clients/cmd/docker-driver/docker-driver:
 	CGO_ENABLED=0 go build $(GO_FLAGS) -o $@ ./$(@D)
 
 docker-driver-push: docker-driver
+ifndef DOCKER_PASSWORD
```
ssncferreira, line 424: This requires the docker username and password to be defined as environment variables? These credentials are coming from the vault right?
https://github.com/grafana/loki/blob/main/.drone/drone.jsonnet#L36-L37

.drone/drone.jsonnet
MichelHollands: That would be nice but I'm not sure we can. This builds a docker driver which is uploaded using `docker plugin push`. I can't find anything similar for podman or buildah.

ssncferreira (approved):
LGTM 🐳

PR #8087, opened 2023-01-10
operator: Fix status not updating when state of pods changes

xperimental:
The operator sometimes fails to update the `.status` field of the `LokiStack` resource to reflect the current state of the associated Pods and set conditions accordingly.

I have identified two reasons for this during testing:

- Out-of-date cached copy of the resource leading to a conflict error
- Not listening to Pod status updates and subsequently not trying to update the status at all

For me, most of the conflicts happened in the controllers for the rules, trying to update the annotations, so I have added retry logic to these controllers.

Instead of adding another watcher for the Pod resource updates, I have updated the predicate used for Deployment and StatefulSet resources to also include updates that only affect the status field. This could be slimmed down to only look at certain fields that we are interested in (the replica counts) or simplified into looking at all changes, skipping the compare operation.

[LOG-3430](https://issues.redhat.com/browse/LOG-3430)

**Checklist**
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/operator/apis/loki/v1/lokistack_types.go
+++ b/operator/apis/loki/v1/lokistack_types.go
@@ -739,7 +739,7 @@ const (
 	// ConditionReady defines the condition that all components in the Loki deployment are ready.
 	ConditionReady LokiStackConditionType = "Ready"
 
-	// ConditionPending defines the conditioin that some or all components are in pending state.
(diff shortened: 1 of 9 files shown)


Red-GV (commented):
lgtm

MasslessParticle (approved):

PR #8088, opened 2023-01-10
Display replicas, cores and nodes in sizing tool.

jeschkies:
No every users wants to deploy Loki with Helm. That's why we are going to display the required replicas as well.

image

This change also fixes two bugs in the algorithm and introduces some tests for its invariant.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/sizing/algorithm_test.go
```diff
@@ -33,3 +33,14 @@ func Test_AlgorithTest_Algorith(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func Test_CoresNodeInvariant(t *testing.T) {
```
jeschkies, line 37: @trevorwhitney I may misunderstand the logic of the algorithm but it fails here in that it demands more cores than is available on the nodes.

trevorwhitney: That sounds like a bug. It should add additional nodes if there aren't enough cores. After a brief look at the code nothing jumps out, would have to dig in more.

pkg/sizing/algorithm.go
```diff
@@ -33,7 +34,7 @@ func calculateClusterSize(nt NodeType, bytesDayIngest float64, qperf QueryPerf)
 	replicasOnLastNode := math.Mod(numWriteReplicasNeeded, writeReplicasPerNode)
 
 	coresOnLastNode := 0.0
-	if replicasOnLastNode >= 0.0 {
+	if replicasOnLastNode > 0.0 {
```
jeschkies, line 37: This was the bug in the algorithm. Glad the test caught it.

pkg/sizing/algorithm_test.go
```diff
@@ -32,4 +32,21 @@ func Test_AlgorithmTest_Algorithm(t *testing.T) {
 	if err := quick.Check(f, nil); err != nil {
 		t.Error(err)
 	}
+
+	// Sanity check for 1TB/Day
+	size := calculateClusterSize(NodeTypesByProvider["AWS"]["t2.xlarge"], 1e12, Basic)
+	require.Equalf(t, 4, size.TotalNodes, "given ingest=1PB/Day totla nodes must be big")
```
jeschkies, line 38: This catches the other bug.

pkg/sizing/algorithm.go
```diff
@@ -23,7 +24,7 @@ const (
 func calculateClusterSize(nt NodeType, bytesDayIngest float64, qperf QueryPerf) ClusterSize {
 
 	// 1 Petabyte per day is maximum. We use decimal prefix https://en.wikipedia.org/wiki/Binary_prefix
-	bytesDayIngest = math.Min(bytesDayIngest, 1e12)
+	bytesDayIngest = math.Min(bytesDayIngest, 1e15)
```
jeschkies, line 27: This was also an issue.

pkg/sizing/algorithm.go
trevorwhitney: nice!

docs/sources/installation/sizing/index.md
```diff
@@ -86,7 +103,8 @@ This tool helps to generate a Helm Charts `values.yaml` file based on specified
 </style>
 
 <script>
-const API_URL = `https://logql-analyzer.grafana.net/next/api/sizing`
+//const API_URL = `https://logql-analyzer.grafana.net/next/api/sizing`
```
trevorwhitney, line 106: don't think you want to commit the localhost url

trevorwhitney (approved):
Looks great. I think there was an accidental localhost url committed, otherwise LGTM!

docs/sources/installation/sizing/index.md
jeschkies: Yes. Thanks for reminding me. This is for testing.

PR #8107, opened 2023-01-11
Change delimiter in  test  "sed" command to ':'

MasslessParticle:
The current delimiter is `/` which breaks when a branch name contains that character. `:` is disallowed in branch names so should serve as a suitable delimiter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,7 +289,7 @@ lint:
 ########
 
 test: all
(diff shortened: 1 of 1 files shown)


MasslessParticle:
Closed in favor of: https://github.com/grafana/loki/pull/8108

PR #8109, opened 2023-01-11
Fixed a typo and made a small edit to the Helm monolithic setup doc. 

09jvilla:
Its storage type, not filesystem type. Also "before you begin" seemed a bit vague given the list below, so I added that the list below was software requirements.


JStickler (commented):
LGTM {docs squad approval)

MasslessParticle (approved):

PR #8122, opened 2023-01-12
Remove flaky helm test

MasslessParticle:
This test is causing tens of timeouts across all PRs in our system. Delete it to improve CI stability

### DIFF
--- a/integration/loki_simple_scalable_test.go
+++ b/integration/loki_simple_scalable_test.go
@@ -12,69 +12,6 @@ import (
 	"github.com/grafana/loki/integration/cluster"
 )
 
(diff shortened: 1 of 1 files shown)


MichelHollands (approved):
LGTM

owen-d (approved):
This looks like a great well written test, but our integration test suite may not be reliable enough to run this reliably. I'm ok removing it

trevorwhitney (approved):
A test that is constantly failing does not increase confidence in the behavior of the code, and in my opinion is worse than having no test at all for that functionality. I'm in favor of removing it as well!

PR #8127, opened 2023-01-13
skip compacting tables that do not fit in start and end date of desired schema config

sandeepsukhani:
Before compacting tables, we look up the schema config for them. If we do not find a schema config for a table, we skip them. However, the schema lookup code was considering the start time of the schema and not the end time.
This can cause the compactor to try compacting tables with the same prefix as one of the previous schemas, but the table interval falls into one of the subsequent schemas with a different config.

This PR fixes the issue and simplifies the code by reusing existing schema lookup code.
I have also updated test cases to verify the fix works.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/indexshipper/compactor/compactor_test.go
+++ b/pkg/storage/stores/indexshipper/compactor/compactor_test.go
@@ -152,6 +152,53 @@ func Test_schemaPeriodForTable(t *testing.T) {
 	indexFromTime := func(t time.Time) string {
 		return fmt.Sprintf("%d", t.Unix()/int64(24*time.Hour/time.Second))
 	}
+	tsdbIndexTablePrefix := fmt.Sprintf("%stsdb_", indexTablePrefix)
+	schemaCfg := config.SchemaConfig{Configs: []config.PeriodConfig{
+		{
+			From:       dayFromTime(start),
+			IndexType:  "boltdb",
+			ObjectType: "filesystem",
+			Schema:     "v9",
+			IndexTables: config.PeriodicTableConfig{
+				Prefix: indexTablePrefix,
+				Period: time.Hour * 24,
+			},
+			RowShards: 16,
+		},
+		{
+			From:       dayFromTime(start.Add(25 * time.Hour)),
+			IndexType:  "boltdb",
+			ObjectType: "filesystem",
+			Schema:     "v12",
+			IndexTables: config.PeriodicTableConfig{
+				Prefix: indexTablePrefix,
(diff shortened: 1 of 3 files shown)


pkg/storage/stores/indexshipper/compactor/compactor_test.go
```diff
@@ -163,14 +210,16 @@ func Test_schemaPeriodForTable(t *testing.T) {
 		{"first table", schemaCfg, indexTablePrefix + indexFromTime(dayFromTime(start).Time.Time()), schemaCfg.Configs[0], true},
 		{"4 hour after first table", schemaCfg, indexTablePrefix + indexFromTime(dayFromTime(start).Time.Time().Add(4*time.Hour)), schemaCfg.Configs[0], true},
 		{"second schema", schemaCfg, indexTablePrefix + indexFromTime(dayFromTime(start.Add(28*time.Hour)).Time.Time()), schemaCfg.Configs[1], true},
-		{"third schema", schemaCfg, indexTablePrefix + indexFromTime(dayFromTime(start.Add(75*time.Hour)).Time.Time()), schemaCfg.Configs[2], true},
+		{"third schema", schemaCfg, tsdbIndexTablePrefix + indexFromTime(dayFromTime(start.Add(75*time.Hour)).Time.Time()), schemaCfg.Configs[2], true},
+		{"unexpected table prefix", schemaCfg, indexTablePrefix + indexFromTime(dayFromTime(start.Add(75*time.Hour)).Time.Time()), config.PeriodConfig{}, false},
 		{"now", schemaCfg, indexTablePrefix + indexFromTime(time.Now()), schemaCfg.Configs[3], true},
+		{"unexpected table number", schemaCfg, tsdbIndexTablePrefix + indexFromTime(time.Now()), config.PeriodConfig{}, false},
```
ashwanthgoli, line 216: this one is again due to "unexpected table prefix" right?

ashwanthgoli (approved):
lgtm

pkg/storage/stores/indexshipper/compactor/compactor_test.go
sandeepsukhani: Yeah, it can be considered both ways actually. Let me shuffle things a bit to simplify it.

sandeepsukhani:
I have fixed `TestDropLabelsPipeline` test, which [keeps failing](https://drone.grafana.net/grafana/loki/18943/3/5) due to 2 PRs being merged at the same time, causing some discrepancy in generated error from a function and the expected error in a test.

PR #8129, opened 2023-01-13
LID: Remote Rule Evaluation

dannykopping:
Introduces a [LID](https://grafana.com/docs/loki/next/lids/) for community consideration on the topic of remote rule evaluation to improve `ruler` performance and reliability in large multi-tenant setups.


docs/sources/lids/0001-Introduction.md
```diff
@@ -12,7 +12,7 @@ title: "0001: Introducing LIDs"
 
 **Type:** Process
 
-**Status:** In Discussion
+**Status:** Accepted
```
chaudum, line 15: Nitpicking: Accepting the LID 0001 should not be part of proposing another LID. Would be great if we could keep that in a separate PR (I know it's overhead ...)

docs/sources/lids/_DRAFT_RemoteRuleEvaluation.md
```diff
@@ -0,0 +1,84 @@
...
+
+Recording rules produce metric series which are sent to a Prometheus-compatible source. Alerting rules send notifications to Alertmanager when a condition is met. Both of these rule types can play a vital role in an organisation's observability strategy, and so their reliable evaluation is essential.
+
+## Problem Statement
+
+Rule evaluations can contain expensive queries. The `ruler` initialises a `querier`, but the `querier` does not have the capability to accelerate queries; the `query-frontend` component is responsible for query acceleration through splitting, sharding, caching, and other techniques.
+
+An expensive rule query can cause an entire `ruler` instance to use excessive resources and even crash. This is highly problematic for the following reasons:
```
chaudum, line 33:
```suggestion
An expensive rule query can cause an entire `ruler` instance to use excessive resources and even crash by running out of memory. This is highly problematic for the following reasons:
```

docs/sources/lids/_DRAFT_RemoteRuleEvaluation.md
```diff
@@ -0,0 +1,84 @@
...
+
+Care must be taken by operators who run large multi-tenant setups to ensure that large volumes of queries can be received, queued, and processed in an acceptable timeframe. The `query-scheduler` component is highly recommended in these situations, as it will enable the `query-frontend` and `query` components to scale out to accommodate the load. Shuffle-sharding should also be implemented to ensure that tenants with particularly large workloads do not starve out the query resources of other tenants. Alerting should also be put in place to notify operators if rule evaluations are being routinely missed or a tenant's query queue becomes full.
+
+Rule-based sharding and remote evaluation are _mutually exclusive_. It is recommended that operators first implement remote evaluation to improve `ruler` reliability, and _then_ further investigate rule-based sharding if rule evaluations are still being missed due to the sequential execution of rule groups, or advise their tenants to split these rule groups up.
```
chaudum, line 84: I don't understand the first sentence here. Why are they mutually exclusive, when you later describe that you first should enable remove evaluation, and then rule-based sharding.

Did you mean that you shall not enable both simultaneously without prior testing of one or another?

docs/sources/lids/_DRAFT_RemoteRuleEvaluation.md
```diff
@@ -0,0 +1,84 @@
...
+
+**Pros:**
+- Takes full advantage of Loki's query acceleration techniques, leading to faster and more efficient rule evaluation
+- Operationally simple as existing `query-frontend`/`query-scheduler`/`querier` setup can be used
+- Per-tenant isolation available in Loki's query path (shuffle-sharding, per-tenant queues) can be used to reduce or eliminate the noisy neighbour problem
+
+**Cons:**
+- Increased interdependence in components, increased cross-component networking
+- Reusing the same `query-frontend`/`query-scheduler`/`querier` setup can cause expensive queries to starve rule evaluations of query resources, and vice versa
+  - Additional complexity introduced if this setup needs to be duplicated for rule evaluations (recommended)
```
chaudum, line 72: Since you say that separating the rule evaluation path from the regular query path is the recommended way, I think it makes sense to add a few words about that preferred setup. While you describe the problem with a common query and rule execution path in the section "Other Notes" below, you don't describe the possible solution.

docs/sources/lids/0001-Introduction.md
dannykopping: Fair point :slightly_smiling_face: I was being lazy.

docs/sources/lids/_DRAFT_RemoteRuleEvaluation.md
dannykopping: > this setup needs to be duplicated

I think that's describing what needs to be done, no? Maybe I should make this a bit more clear.

docs/sources/lids/_DRAFT_RemoteRuleEvaluation.md
dannykopping: Yeah, I should clarify this more. I meant that they are separate features and one does not require the other. Good call-out, I'll tighten up the language here.

docs/sources/lids/_DRAFT_RemoteRuleEvaluation.md
```diff
@@ -0,0 +1,84 @@
...
+
+Rule-based sharding distributes rules evenly across all available `ruler` instances, each in their own rule group. Consequentially, each rule that belongs to a `ruler` instance will be evaluated concurrently (as they're each in their own rule group). For tenants with hundreds or thousands of rules, this can result in large batches of queries being sent to the `query-frontend` in quick succession, should they all use the same interval or happen to overlap.
+
+Care must be taken by operators who run large multi-tenant setups to ensure that large volumes of queries can be received, queued, and processed in an acceptable timeframe. The `query-scheduler` component is highly recommended in these situations, as it will enable the `query-frontend` and `query` components to scale out to accommodate the load. Shuffle-sharding should also be implemented to ensure that tenants with particularly large workloads do not starve out the query resources of other tenants. Alerting should also be put in place to notify operators if rule evaluations are being routinely missed or a tenant's query queue becomes full.
```
salvacorts, line 82:
```suggestion
Care must be taken by operators who run large multi-tenant setups to ensure that large volumes of queries can be received, queued, and processed in an acceptable timeframe. The `query-scheduler` component is highly recommended in these situations, as it will enable the `query-frontend` and `querier` components to scale out to accommodate the load. Shuffle-sharding should also be implemented to ensure that tenants with particularly large workloads do not starve out the query resources of other tenants. Alerting should also be put in place to notify operators if rule evaluations are being routinely missed or a tenant's query queue becomes full.
```

docs/sources/lids/_DRAFT_RemoteRuleEvaluation.md
```diff
@@ -0,0 +1,84 @@
...
+
+Loki's current `ruler` implementation is sufficient for small installations running relatively simple or inexpensive queries.
+
+**Pros**:
+- Nothing to be done
+
+**Cons:**
+- Loki's `ruler` will remain unreliable and inefficient when used in large multi-tenant environments with expensive queries.
+
+### Proposal 1: Remote Execution
```
salvacorts, line 60: IIRC, we discussed the pros and cons of two ways of achieving remote execution:
1. Re-use the existing frontend+scheduler+queriers
2. Create a separate set of frontend+scheduler+queriers

I think we did not discard any of those two alternatives, so I think it may be worth adding a third proposal with (2.).

salvacorts (commented):
Thanks for putting this together Danny!

docs/sources/lids/_DRAFT_RemoteRuleEvaluation.md
chaudum: Yes, kind of, but I think it should be more prominent as part of the proposal if it's the recommended way to run the rule execution in the future.
I guess it's a similar concern that @salvacorts had in [his comment](https://github.com/grafana/loki/pull/8129/files#r1069481832). My comment should not block the LID from being merged, though.

dannykopping:
Apologies for the rebase, I messed up the branch somehow.

docs/sources/lids/_DRAFT_RemoteRuleEvaluation.md
dannykopping: The workload is referring to user queries. I'll add some more detail here to improve the clarity, thanks!

dannykopping: See https://github.com/grafana/loki/pull/8129/commits/1aad3e950eca4289a92b32f10c97e5484c32b822

chaudum (approved):

salvacorts (approved):

owen-d (approved):
LGTM. I'd also introduce some jitter into rule evaluations to improve their queueing distribution and _always_ shard by rule instead of group.

slim-bean:
our Mimir friends I believe implemented this using a separate query-frontend and querier pool and I think we should do the same thing, I think keeping the isolation between alerts/rules and user queries is important.

dannykopping:
> our Mimir friends I believe implemented this using a separate query-frontend and querier pool and I think we should do the same thing, I think keeping the isolation between alerts/rules and user queries is important. That may be an implementation detail not in this doc but it sort of reads like the intent here is to use the same resources for both and I would vote for separate resources, otherwise LGTM!

I'm very much advocating for using separate resources 👍

> If rule evaluations and tenant queries are slowing each other down, the read path setup would need to be duplicated so that tenant queries and rule evaluations would not share the same query execution resources.

PR #8145, opened 2023-01-13
fixes bug with queryIngesterWithin logic in asyncStore ingester stats…

owen-d:
Fixes a previous mistake in the logic calculating when to skip querying ingesters in the async store Statistics method. Notably `through.After` should be `through.Before` when skipping querying ingesters as that's when there's no overlap with the `query-ingesters-within` period:

```go
	// OLD CODE BELOW
	if a.queryIngestersWithin != 0 {
		// don't query ingesters if the query does not overlap with queryIngestersWithin.
		if !through.After(model.Now().Add(-a.queryIngestersWithin)) { // <----- should be through.Before
			return a.Store.Stats(ctx, userID, from, through, matchers...)
		}
	}
```

 I discovered the problem while debugging querier OOMs during a boltdb-shipper -> tsdb migration and ultimately found this happened under the following circumstances:

* Queries over high volumes of _recent_ data wouldn't query ingesters for index metadata
* Without index metadata, we only checked storage metadata
* There is a delay before we ship the index to storage, meaning we don't see it if ingesters are skipped
* Calculating ideal shard factors without recent data for queries that only touch recent data _dramatically_ underestimates the desired shard factor
* Queries aren't split enough and get scheduled onto too few querier replicas
* They oom. We had some fun examples like a single replica trying to download 419,000 chunks/query

To be clear, this is still a hypothesis, but a plausible one, especially after finding the boolean logic error this PR fixes.

Instead of changing this one line, I took the opportunity to refactor this into a shared utility used by our other `GetChunkRefs` method which is already tested, ensuring the logic works as expected. I also added some more logging visibility into this code so we can understand what the difference is when querying statistics from storage vs ingesters. Finally, I added a helper to prepare our `Stats` objects to be logged which is now used in a few places.

### DIFF
--- a/pkg/logproto/extensions.go
+++ b/pkg/logproto/extensions.go
@@ -1,8 +1,10 @@
 package logproto
 
 import (
+	"strings"
 	"sync/atomic" //lint:ignore faillint we can't use go.uber.org/atomic with a protobuf struct without wrapping it.
 
+	"github.com/dustin/go-humanize"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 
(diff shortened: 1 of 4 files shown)


MasslessParticle (approved):
Awesome catch

PR #8147, opened 2023-01-14
Helm: Expose single binary service configuration in values

### LINKED ISSUE #8146: Helm: Expose single binary service configuration in values
The Service used in single-binary mode in the Helm chart does not expose any configuration options the same way the Service for the gateway does.

The same options that `gateway.service` has available should be configurable for the single-binary Service via `values.yaml`. Something like the following should be possible:

```yaml
singleBinary:
  service:
    type: NodePort # default is ClusterIP
    nodePort: 31000 # default is null
```

An optionally enableable secondary Service with all of the configuration options exposed in `values.yaml`. It seems like overkill given that the ingress will point to the already existing service anyway.

**Additional context**
Having these options exposed is very helpful for dev setups where something like a simple `NodePort` is much simpler for accessing Loki than going through an ingress. This is possible to do with the simple-scalable configuration since the gateway's service can be configured to be of type `NodePort` or whatever you want. However the single-binary configuration is more desirable on resource-constrained setups like those used for dev environments.

One potentially weird thing is that the single-binary Service has a gRPC port in addition to the HTTP port. I'm not sure how much it will harm the ergonomics to have both a `.service.port` and a `.service.grpcPort`.

sapslaj:
Exposes service configuration used in single-binary mode via `values.yaml`. This way it is just as configurable as the gateway Service in simple-scalable mode.

Fixes #8146

Like I mentioned in #8146, this service has an extra port for gRPC that the gateway Service does not have. The solution I came up with is just to add `grpcPort` and `grpcNodePort` but I didn't know if this was desirable or not. Otherwise this this should behave exactly the same as before with the current defaults (I confirmed this via `helm diff`).

**Checklist**
- [x] Reviewed the [`CONTRIBUTING.md`](../CONTRIBUTING.md) guide (**required**)
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/values.yaml
+++ b/production/helm/loki/values.yaml
@@ -902,6 +902,25 @@ singleBinary:
     storageClass: null
     # -- Selector for persistent disk
     selector: null
+  service:
+    # -- Port of the single binary service
+    port: 3100
+    # -- GRPC port of the single binary service
+    grpcPort: 9095
+    # -- Type of the single binary service
+    type: ClusterIP
+    # -- ClusterIP of the single binary service
+    clusterIP: null
+    # -- (int) Node port if service type is NodePort
+    nodePort: null
+    # -- (int) GRPC node port if service type is NodePort
+    grpcNodePort: null
+    # -- Load balancer IPO address if service type is LoadBalancer
+    loadBalancerIP: null
+    # -- Annotations for the single binary service
+    annotations: {}
+    # -- Labels for single binary service
+    labels: {}
 
 # Use either this ingress or the gateway, but not both at once.
 # If you enable this, make sure to disable the gateway.
--- a/docs/sources/installation/helm/reference.md
+++ b/docs/sources/installation/helm/reference.md
@@ -3136,6 +3136,87 @@ null
 </td>
 		</tr>
 		<tr>
+			<td>singleBinary.service.annotations</td>
+			<td>object</td>
+			<td>Annotations for the single binary service</td>
+			<td><pre lang="json">
+{}
+</pre>
+</td>
+		</tr>
+		<tr>
+			<td>singleBinary.service.clusterIP</td>
+			<td>string</td>
+			<td>ClusterIP of the single binary service</td>
+			<td><pre lang="json">
+null
+</pre>
+</td>
+		</tr>
+		<tr>
+			<td>singleBinary.service.grpcNodePort</td>
+			<td>int</td>
+			<td>GRPC node port if service type is NodePort</td>
+			<td><pre lang="json">
+null
+</pre>
+</td>
+		</tr>
+		<tr>
+			<td>singleBinary.service.grpcPort</td>
+			<td>int</td>
+			<td>GRPC port of the single binary service</td>
+			<td><pre lang="json">
+9095
+</pre>
+</td>
+		</tr>
+		<tr>
+			<td>singleBinary.service.labels</td>
+			<td>object</td>
+			<td>Labels for single binary service</td>
+			<td><pre lang="json">
+{}
+</pre>
+</td>
+		</tr>
+		<tr>
+			<td>singleBinary.service.loadBalancerIP</td>
+			<td>string</td>
+			<td>Load balancer IPO address if service type is LoadBalancer</td>
+			<td><pre lang="json">
+null
+</pre>
+</td>
+		</tr>
+		<tr>
+			<td>singleBinary.service.nodePort</td>
+			<td>int</td>
+			<td>Node port if service type is NodePort</td>
+			<td><pre lang="json">
+null
+</pre>
+</td>
+		</tr>
+		<tr>
+			<td>singleBinary.service.port</td>
+			<td>int</td>
+			<td>Port of the single binary service</td>
+			<td><pre lang="json">
+3100
+</pre>
+</td>
+		</tr>
+		<tr>
+			<td>singleBinary.service.type</td>
+			<td>string</td>
+			<td>Type of the single binary service</td>
+			<td><pre lang="json">
+"ClusterIP"
+</pre>
+</td>
+		</tr>
+		<tr>
 			<td>singleBinary.targetModule</td>
 			<td>string</td>
 			<td>Comma-separated list of Loki modules to load for the single binary</td>
--- a/production/helm/loki/templates/single-binary/service.yaml
+++ b/production/helm/loki/templates/single-binary/service.yaml
@@ -7,16 +7,35 @@ metadata:
   name: {{ include "loki.name" . }}
   labels:
     {{- include "loki.labels" . | nindent 4 }}
+    {{- with .Values.singleBinary.service.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- with .Values.singleBinary.service.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
-  type: ClusterIP
+  type: {{ .Values.singleBinary.service.type }}
+  {{- with .Values.singleBinary.service.clusterIP }}
+  clusterIP: {{ . }}
+  {{- end }}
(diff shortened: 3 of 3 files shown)


sapslaj:
Hmm I'm not sure what to do about that lint failure. It wants to make this change:

```diff
diff --git a/production/helm/loki/README.md b/production/helm/loki/README.md
index 911a73ea3..9354f6dc0 100644
--- a/production/helm/loki/README.md
+++ b/production/helm/loki/README.md
@@ -17,4 +17,4 @@ Helm chart for Grafana Loki in simple, scalable mode
 | https://charts.min.io/ | minio(minio) | 4.0.12 |
 | https://grafana.github.io/helm-charts | grafana-agent-operator(grafana-agent-operator) | 0.2.3 |
 
-[Find more information in the Loki Helm Chart documentation](https://grafana.com/docs/loki/next/installation/helm).
+[Find more information in the Loki Helm Chart [documentation](https://grafana.com/docs/loki/next/installation/helm).
```

It looks like it's coming from the template, but that isn't technically valid markdown. I didn't make any changes to either the rendered README.md or the template. Any advice?

production/helm/loki/values.yaml
```diff
@@ -902,6 +902,25 @@ singleBinary:
     storageClass: null
     # -- Selector for persistent disk
     selector: null
+  service:
+    # -- Port of the single binary service
+    port: 3100
+    # -- GRPC port of the single binary service
+    grpcPort: 9095
```
trevorwhitney, line 909: we intentionally don't expose http and grpc ports in the configuration for the other target services. this allows us to keep the chart simpler with less config. what is the reason for making these configurable? I would much prefer we keep them in line with the others.

production/helm/loki/values.yaml
```diff
@@ -902,6 +902,25 @@ singleBinary:
...
+    # -- Type of the single binary service
+    type: ClusterIP
+    # -- ClusterIP of the single binary service
+    clusterIP: null
+    # -- (int) Node port if service type is NodePort
+    nodePort: null
+    # -- (int) GRPC node port if service type is NodePort
+    grpcNodePort: null
```
trevorwhitney, line 917: what's the reason for exposing GRPC port?

trevorwhitney (changes requested):
Thanks for the contribution! Can you describe your use case a little bit more. We are being very intentional about keeping this chart opinionated so that we don't need to expose every possible value as something in the `values.yaml`. One example of that is we don't allow grpc or http ports to be configured for the other targets (ie. read, write, backend). This allows us to make assumptions in the gateway, allowing for a simpler template and less configuration.

A few questions:
1. Why not use the gateway in front of single binary targets?
2. Why do the http and grpc ports need to be configurable?
3. If the answer to 2 is because you need http on 80, could we add an aditional port to the backend service that's configurable, and have that be the only one (similar to what we do in the gateway's service).

sapslaj:
@trevorwhitney Thanks for looking at this! I mentioned the use case in a separate issue (8146), but I can TLDR it here. Having a way to access Loki in a single binary configuration _without_ enabling an ingress is useful for dev/test setups via the service types `NodePort` or `LoadBalancer`.

The gateway isn't enabled in single binary mode, so the only way to access Loki is via that single binary service. AFAIK there isn't a way to easily enable it in some instances since the condition for enabling it is ANDed with `loki.deployment.isScalable`, which seems to be based on whether the storage backend is using object storage or file system. In my dev/test setup, I'm using the filesystem since MinIO can be a bit of a memory hog, and setting up a cloud provider bucket is quite a bit of extra work for something ephemeral-ish. Seems like a strange ask to require using a completely different storage backend to enable a frontend proxy.

https://github.com/grafana/loki/blob/65bef5c360b31f091523e8fcb59b89b8b2cf6025/production/helm/loki/templates/gateway/service-gateway.yaml#L1-L2
https://github.com/grafana/loki/blob/65bef5c360b31f091523e8fcb59b89b8b2cf6025/production/helm/loki/templates/_helpers.tpl#L28-L30
https://github.com/grafana/loki/blob/65bef5c360b31f091523e8fcb59b89b8b2cf6025/production/helm/loki/templates/_helpers.tpl#L446-L448

So to answer question 1: You can't, from what I can tell, at least not without a good deal of effort to add support for single binary. It looks like the nginx config template directs different subpaths to either the read or write services. That all would have to be duplicated just for single binary mode or make the `loki.readFullname` and `loki.writeFullname` helpers render different values based on whether it's in single binary or not. This PR was quite a bit easier to implement since there isn't a dependency on the gateway in single binary mode. One could argue that the single binary service performs the same function as the gateway service. This is corroborated by places where the Loki host is determined depending on whether the config is for single binary or not. So if anything making this service configurable aligns with what gateway already offers.

For 2 and 3, [k3s's ServiceLB/`klipper-lb`](https://docs.k3s.io/networking#service-load-balancer) will forward traffic for each of the ports on that service. What if there's already a service with port 3100 forwarded? Guess you're not getting traffic to that service! Hence why I opted to make it configurable. And it's for every port in the service, which is why the gRPC port also needs to be configurable. I made the node ports configurable just for completeness.

A middle ground might be to add a dedicated, optionally enableable, Service that can be fully configured just like the gateway's service. I didn't go that route initially for several reasons. I can imagine confusion over which service does what if we had two services with the same exact pod selector, but one was configurable via `values.yaml` and the other wasn't.

I did check through, and it looks like that port _is_ hardcoded in a few places, but it also seems to be hardcoded for the gateway service as well, so this particular issue isn't limited to just the single binary service.

https://github.com/grafana/loki/blob/65bef5c360b31f091523e8fcb59b89b8b2cf6025/production/helm/loki/templates/_helpers.tpl#L456-L463

We'd probably have to do something like this (there's a more elegant way to do this but just illustrating the point):
```tpl
{{- $url := printf "%s.%s.svc.%s.:%d" (include "loki.gatewayFullname" .) .Release.Namespace .Values.global.clusterDomain .Values.gateway.service.port }}
{{- if and $isSingleBinary (not .Values.gateway.enabled)  }}
  {{- $url = printf "%s.%s.svc.%s.:%d" (include "loki.singleBinaryFullname" .) .Release.Namespace .Values.global.clusterDomain .Values.singleBinary.service.port }}
{{- end }}
```

mateuszdrab:
@trevorwhitney I just wanted to add my 2 cents that I'd like to use the gateway with single binary for the sake of configuration simplicity/reusing values; but as mentioned by @sapslaj, this is not possible in single binary mode. 
Additionally, I have some additional nginx snippets added at the gateway level, which work around some issues with tracing headers (rewriting W3C header format to Jaeger format).
Is this something that we can work towards resolving?

sapslaj:
@mateuszdrab I like the idea of being able to use the gateway with single binary long-term. I just went the path of least resistance at first and made this PR to upstream my changes. My goal is simply "have a `values.yaml` configurable (Kubernetes) Service to access Loki in single binary mode." If that service is the gateway's service then that works for me! Nginx is very lightweight so it shouldn't have any noticeable negative impact for my use cases.

I have no problems taking a stab at adding single binary support to the gateway, but it's going to end up being a pretty big PR. It also might be potentially a major version bump since `gateway.enabled` is `true` by default (and has to be for non-single-binary) so removing the check for `isScalable` will cause the gateway to get created by default in single binary mode. There's are a lot of other little decisions like that if we go forward with adding single binary support to gateway, so I'd like to make sure that's something the maintainers (@trevorwhitney et al.) are willing to support.

trevorwhitney:
Thanks @sapslaj for the concise use case. I like the option you and @mateuszdrab mentioned about using the gateway for single binary. I think for consistency between modes, and simplicity of configuration (not having to expose all these confiugrations of the single binary service) that I too would prefer enabling the gateway for single binary. I think this is also beneficial to those who are looking to run single binary in an HA mode with object storage support, which is something I'm also look into (see https://github.com/grafana/loki/issues/8185)

trevorwhitney:
Closing. Moving to issue to enable the gateway for single binary mode: https://github.com/grafana/loki/issues/8269

PR #8150, opened 2023-01-16
fix(logql): remove __error_details__ label on drop __error__ or other way around

adityacs:
This is follow up PR to #7975

**Checklist**
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/logql/log/drop_labels.go
+++ b/pkg/logql/log/drop_labels.go
@@ -47,15 +47,20 @@ func isErrorDetailsLabel(name string) bool {
 	return name == logqlmodel.ErrorDetailsLabel
 }
 
+func resetError(lbls *LabelsBuilder) {
+	lbls.ResetError()
+	lbls.ResetErrorDetails()
+}
+
 func dropLabelNames(name string, lbls *LabelsBuilder) {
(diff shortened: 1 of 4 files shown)


pkg/logql/log/drop_labels.go
```diff
@@ -47,15 +47,20 @@ func isErrorDetailsLabel(name string) bool {
...
-	if isErrorDetailsLabel(name) {
-		lbls.ResetErrorDetails()
+	if isErrorLabel(name) || isErrorDetailsLabel(name) {
+		resetError(lbls)
 		return
 	}
+	// if isErrorDetailsLabel(name) {
+	// 	resetError(lbls)
+	// 	return
+	// }
```
chaudum, line 63: Can be removed completely, right?

chaudum (approved):
Thanks for the follow up!

pkg/logql/log/drop_labels.go
adityacs: Oh! yes

adityacs: removed

adityacs: Thanks for the review

PR #8151, opened 2023-01-16
deletion: fix log deletion with line filters

sandeepsukhani:
When a delete request is issued with a line filter, [the current deletion code](https://github.com/grafana/loki/blob/d317e659c7583db5e2fcd4828155fc85c45dce60/pkg/storage/stores/indexshipper/compactor/retention/retention.go#L179-L203) always assumes that the chunk would be re-written with a new id and marks the existing chunk for deletion. However, if no lines match the line filter, then the chunk is not changed. This causes us to delete the whole chunk unexpectedly.

This PR takes care of the issue by detecting if there was any data deleted from the chunk. If there is no change in the chunk then it should be a no-op.
Additionally, I have refactored the code to simplify partial chunk deletion by accepting `time.Time` in [filter.Func](https://github.com/grafana/loki/blob/f34661e4caa8444b81c036339a18b70df1d745a9/pkg/util/filter/filter_function.go#L5). This way we do not have to calculate the chunk remainders beforehand and use this `filter.Func` to decide whether the log line is supposed to be deleted or not while iterating through the logs in the chunk. This should not add much overhead since the chunk slicer anyways iterates through the log lines for building a new chunk.

I have also improved the integration tests to cover all the deletion modes and test the deletion with/without query time filtering.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated


integration/loki_micro_services_delete_test.go
```diff
@@ -129,66 +195,92 @@ func TestMicroServicesDeleteRequest(t *testing.T) {
...
+		validateQueryResponse(resp)
 	})
 
 	// Wait until delete request is finished
 	t.Run("wait-until-delete-request-processed", func(t *testing.T) {
+		tenantLimits := tCompactor.GetTenantLimits(tenantID)
+		tenantLimits.DeletionMode = "filter-and-delete"
+		require.NoError(t, tCompactor.SetTenantLimits(tenantID, tenantLimits))
+
+		// all the delete requests should have been processed
```
MichelHollands, line 242: How do you know the delete requests have been processed?

pkg/storage/stores/indexshipper/compactor/retention/retention.go
```diff
@@ -332,75 +339,85 @@ func newChunkRewriter(chunkClient client.Client, tableName string, chunkIndexer
 	}
 }
 
-func (c *chunkRewriter) rewriteChunk(ctx context.Context, ce ChunkEntry, tableInterval model.Interval, intervalFilters []IntervalFilter) (bool, error) {
+func (c *chunkRewriter) rewriteChunk(ctx context.Context, ce ChunkEntry, tableInterval model.Interval, filterFunc filter.Func) (bool, bool, error) {
```
MichelHollands, line 342: Can you add a comment that explains the return values? It's only at the end of this long function that you can see what they mean.

MasslessParticle: +1

MasslessParticle (approved):
This is way easier to follow. LGTM with a couple of comments

MichelHollands (approved):

integration/loki_micro_services_delete_test.go
sandeepsukhani: I am verifying that below in `require.Eventually`. Sorry, am I missing something?

MichelHollands (approved):
LGTM

MasslessParticle (approved):

PR #8158, opened 2023-01-16
Bug fix: isFunction consume all reader left error when matcher label is function token

### LINKED ISSUE #7639: 【bugs】seems logql query length is limited?
**Describe the bug**
logql query size too large，loki return error：
```
parse error at line 1, col 6: literal not terminated
```

**To Reproduce**
loki version: ``` 2.6.1```
logql query within grafana like:
```
{ip=~"10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5"}
```

**Expected behavior**
above query return success

**Environment:**
os version: ```CentOS Linux release 7.4.1708 (Core)```

no

suikammd:
Recently I met this a problem, just like this issue [issue 7639](https://github.com/grafana/loki/issues/7639) reported. When a `matcher` uses a `function` token as label (e.g. ip), Loki will false consume all left text. It will try to judge whether it is a `function` by using the func blew:
```
func isFunction(sc scanner.Scanner) bool {
	var sb strings.Builder
	sc = trimSpace(sc)
	for r := sc.Next(); r != scanner.EOF; r = sc.Next() {
		sb.WriteRune(r)
		switch sb.String() {
		case "(":
			return true
		case "by", "without":
			sc = trimSpace(sc)
			return sc.Next() == '('
		}
	}
	return false
}
```
if there is a `(` followed or `by (` or `without (` followed, it will consider a `function`. if no matches it will consume all left text. If the label value is too long, we will get a truncated label value and it is not a valid string because we do not find the left `"`. 
Why this happens? 
Inside Scanner there's a Reader and a srcBuf buffers (array with length `1 + 1024`) data consumed from Reader. Func `isFunction` seems to just identify it is a `function` or not and will not modify Reader, but in fact, this func may consumes the whole Reader! It means is not a immutable borrow of Reader but a mutable borrow. If no matches, this func will keep calling `Scanner.Next()` and consumes the raw Reader.
 
Fixes #7639

In fact, I tried several other more reasonable solutions by clone entire struct of Reader of Scanner, but it seems hard to get consume status. Another solution is wrap Scanner and reimplement all funcs it has, but it was hard to reimplement Scan func. So I just modify `isFunction` a little to workaround. This mr in fact can not completely fix this problem, but will significantly reduce the probability of error happening. We should implement a brand new Scanner like prometheus does.


kavirajk:
@suikammd thanks for the PR. 

Totally agree the original error message `parse error at line 1, col 6: literal not terminated` is definitely confusing if your label value is too large and truncated.

Ideally, I would expect to change this error message to something useful. Say if you selector looks something like (taken from your example)
```
{ip=~"10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5"}
```

In that case instead of truncating the label value and display wierd parsing error message, we should show message like `label "ip" has value that is too large (maximum length is xx)` or something similar.

I'm failing to understand how changing `isFunction` would be the fix here.

suikammd:
> @suikammd thanks for the PR.
> 
> Totally agree the original error message `parse error at line 1, col 6: literal not terminated` is definitely confusing if your label value is too large and truncated.
> 
> Ideally, I would expect to change this error message to something useful. Say if you selector looks something like (taken from your example)
> 
> ```
> {ip=~"10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5|10.5.17.5"}
> ```
> 
> In that case instead of truncating the label value and display wierd parsing error message, we should show message like `label "ip" has value that is too large (maximum length is xx)` or something similar.
> 
> I'm failing to understand how changing `isFunction` would be the fix here.

In fact it can not completely fix the problem, just make things less likely to happen. (So you do not need to merge this pr, though this pr does the same things as the original implementation.)
We need to self reimplement a scanner to solve this problem instead of using text scanner. Text scanner does not support fall back. [Prometheus](https://github.com/prometheus/prometheus/blob/main/promql/parser/lex.go) implement a scanner here. when not meet expectations, it supports falling back to the original state.

PR #8166, opened 2023-01-16
[Helm] Allow node port on single binary service

jkroepke:
Signed-off-by: Jan-Otto Kröpke <jok@cloudeteer.de>

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/installation/helm/reference.md
+++ b/docs/sources/installation/helm/reference.md
@@ -3151,6 +3151,60 @@ null
 </td>
 		</tr>
 		<tr>
+			<td>singleBinary.service.annotations</td>
+			<td>object</td>
+			<td>Annotations for the gateway service</td>
+			<td><pre lang="json">
+{}
+</pre>
+</td>
+		</tr>
+		<tr>
+			<td>singleBinary.service.clusterIP</td>
+			<td>string</td>
+			<td>ClusterIP of the single binary service</td>
+			<td><pre lang="json">
+null
+</pre>
+</td>
+		</tr>
+		<tr>
+			<td>singleBinary.service.labels</td>
+			<td>object</td>
+			<td>Labels for gateway service</td>
+			<td><pre lang="json">
+{}
+</pre>
+</td>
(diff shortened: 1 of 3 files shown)


sapslaj:
Hey @jkroepke looks like we had the same idea. #8146 #8147

trevorwhitney (changes requested):
Can you please provide a description and explain why you need this functionality? We are intentionally limiting what can be configured on the "internal" services.

jkroepke:
We have an central grafana instance (from loki perspective outside of the cluster) and we do not have flexibility of using ingress (we have only public ingress) and we do not want to expose loki to public internet.

The central grafana outside of the cluster needs direct access. node port is the easiest available option for us. In that case, the single binary loki "internal" service needs to be exposed to an other cluster.

jkroepke:
@trevorwhitney how to generate the docs?

> make -C docs sources/installation/helm/reference.md

generates no change

MasslessParticle:
@jkroepke This was an error on our end. We fixed it so a rebase should get this to pass. Sorry about that.

trevorwhitney:
As discussed in https://github.com/grafana/loki/pull/8147, enabling the gateway for single binary mode https://github.com/grafana/loki/issues/8269 will solve a few issues at once, while keeping the config smaller and different deployment modes more consistent. Closing in favor of that approach.

jkroepke:
@trevorwhitney I'm bit confused, keeping the configuration small while integrate an automatic way to identity customer need feels like a conflict (https://github.com/grafana/loki/issues/8185)

trevorwhitney:
@jkroepke does my implementation clarify at all? https://github.com/grafana/loki/pull/8284. This requires no additional config (keeping it small), and only required the change of a single default value (which I believe was done in a non-breaking way). It does make it so the gateway is always enabled, but I feel like that's a good thing from a consistency perspective.

Issue #8174, opened 2023-01-17
Config reload with /reload endpoint causes metrics scraping error


buildsville:
**Describe the bug**
If multiple `scrape_config` set and `metrics` is set in `pipeline_stages`, reloading the config with the /reload endpoint will cause an error when scraping /metrics endpoint.

**To Reproduce**
1. Started Promtail 2.7.1 with below

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: promtail
spec:
  selector:
    matchLabels:
      app: promtail
  template:
    metadata:
      labels:
        app: promtail
    spec:
      containers:
        - name: promtail
          image: grafana/promtail:2.7.1
          args:
            - -config.file=/etc/promtail/promtail.yaml
          volumeMounts:
            - name: config
              mountPath: /etc/promtail
            - name: tmp
              mountPath: /tmp
        - name: log
          image: busybox
          command:
            - sh
            - -c
            - while true; do echo hello; sleep 1; done >/tmp/hello.log
          volumeMounts:
            - name: tmp
              mountPath: /tmp
      volumes:
        - name: config
          configMap:
            name: promtail
        - name: tmp
          emptyDir: {}
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: promtail
data:
  promtail.yaml: |
    server:
      enable_runtime_reload: true
      http_listen_port: 3101

    client:
      url: http://localhost:3101/

    positions:
      filename: /tmp/positions.yaml

    scrape_configs:
      - job_name: local1
        static_configs:
          - targets: [localhost]
            labels:
              job: local1
              __path__: /tmp/hello.log
        pipeline_stages:
          - metrics:
              debug_metrics:
                type: Counter
                source: job
                config:
                  action: inc

      - job_name: local2
        static_configs:
          - targets: [localhost]
            labels:
              job: local2
              __path__: /tmp/hello.log
        pipeline_stages:
          - metrics:
              debug_metrics:
                type: Counter
                source: job
                config:
                  action: inc
```

2. Modify ConfigMap as below. (Add a label to one of the job)

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: promtail
data:
  promtail.yaml: |
    server:
      enable_runtime_reload: true
      http_listen_port: 3101

    client:
      url: http://localhost:3101/

    positions:
      filename: /tmp/positions.yaml

    scrape_configs:
      - job_name: local1
        static_configs:
          - targets: [localhost]
            labels:
              job: local1
[...]

PR #8178, opened 2023-01-17
【RFR】fix Config reload with /reload endpoint causes metrics scraping error

liguozhong:
fix Config reload with /reload endpoint causes metrics scraping error

pipeline.go adds a Close() interface, which releases the resources configured last time when promtail reloads. Currently, only the metrics stage implements the Close() interface

Fixes #https://github.com/grafana/loki/issues/8174

Because the Unregister function of prometheus only works on collectorsByID, the original metrics stage uses a custom Collector that belongs to uncheckedCollectors. Therefore, metrics.go is re-implemented, and the original collector is replaced with the original sdk of prometheus/counter/gauge/histogram.
```go		
c.registry.Unregister(m)
```

```go
type Registry struct {
	collectorsByID        map[uint64]Collector // ID is a hash of the descIDs.
	uncheckedCollectors   []Collector
}
```

```go
// Unregister implements Registerer.
func (r *Registry) Unregister(c Collector) bool {
	var (
		descChan    = make(chan *Desc, capDescChan)
		descIDs     = map[uint64]struct{}{}
		collectorID uint64 // All desc IDs XOR'd together.
	)
	go func() {
		c.Describe(descChan)
		close(descChan)
	}()
	for desc := range descChan {
		if _, exists := descIDs[desc.id]; !exists {
			collectorID ^= desc.id
			descIDs[desc.id] = struct{}{}
		}
	}

	r.mtx.RLock()
	if _, exists := r.collectorsByID[collectorID]; !exists {
		r.mtx.RUnlock()
		return false
	}
	r.mtx.RUnlock()

	r.mtx.Lock()
	defer r.mtx.Unlock()

	delete(r.collectorsByID, collectorID)
	for id := range descIDs {
		delete(r.descIDs, id)
	}
	// dimHashesByName is left untouched as those must be consistent
	// throughout the lifetime of a program.
	return true
}
```

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


liguozhong:
> Thanks for the quick fix @liguozhong! I think we should add a test case for this, though.

OK, I will be on vacation for the next two weeks, and I will come back in two weeks to continue to complete this PR

owen-d:
Enjoy the time off!

liguozhong:
done.

Because the Unregister function of prometheus only works on collectorsByID, the original metrics stage uses a custom Collector that belongs to uncheckedCollectors. Therefore, metrics.go is re-implemented, and the original collector is replaced with the original sdk of prometheus/counter/gauge/histogram.

```
=== RUN   TestPipeline_Reload
    pipeline_test.go:497: 
        	Error Trace:	/Users/fuling/go/src/github.com/grafana/loki/clients/pkg/logentry/stages/pipeline_test.go:497
        	Error:      	Received unexpected error:
        	            	collected metric "promtail_custom_debug_metrics" { label:<name:"action" value:"GET" > label:<name:"match" value:"false" > label:<name:"status_code" value:"200" > label:<name:"stream" value:"stderr" > counter:<value:1 > } was collected before with the same name and label values
        	Test:       	TestPipeline_Reload
--- FAIL: TestPipeline_Reload (0.00s)

```

cstyan:
@liguozhong is this PR still relevant?

chaudum:
Promtail is superseded by [Grafana Agent](https://grafana.com/docs/agent/latest/) and [Grafana Alloy](https://grafana.com/docs/alloy/latest/).

PR #8024, opened 2023-01-02
[loki-canary] Support passing loki address as environment variable

jijotj:
**What this PR does / why we need it**:
- Replace flag with kingpin which has more capabilities to override defaults
- Override default for `addr`, `user` and `pass` from environment the variables `LOKI_ADDRESS`, `LOKI_USERNAME` and `LOKI_PASSWORD` respectively.

**Which issue(s) this PR fixes**:
Fixes #8021

**Special notes for your reviewer**:

**Checklist**
- [x] Reviewed the `CONTRIBUTING.md` guide
- [x] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #8183, opened 2023-01-17
Revert "[loki-canary] Support passing loki address as environment variable (#8024)"

JordanRushing:
This PR reverts commit f9ed9cb7d98dcdaa6fe2f1b7e97ca86af2864121 which introduced breaking config changes for the Loki canary. The relevant change introduced support for configuring a new `LOKI_ADDRESS` canary environment variable by adopting a new library for flags.

This is the easiest path forward to avoid forcing breaking config changes on canary operators; such a change is likely better suited to join other configuration changes in a future major Loki release.

N/A

Related PR:
- https://github.com/grafana/loki/pull/8024

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/cmd/loki-canary/main.go
+++ b/cmd/loki-canary/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


slim-bean (approved):
LGTM

slim-bean:
@jijotj sorry to have revert this 😬 but unfortunately for all of us the Go stdlib flag parser allows for single dash var names even against long variables.

kingpin will fail to parse all of these as it expects a single dash to be a short name for a config.

Unfortunately this means basically everyones canary install is broken by this change 😞 

I think if we were to make a change as big as this we'd want to tie it to a major release.

Since your original PR you were only looking to configure a few vars from the environment, I wonder if a smaller PR to just set those manually could be a compromise here?

jijotj:
Oh cool, I chose kingpin because it was already used in the Loki ecosystem by logcli. Thanks for suggesting a compromise, I'll raise a PR to do just that.

Issue #4143, opened 2021-08-11
Push official images to a different registry, in addition to Docker Hub


onedr0p:
**Is your feature request related to a problem? Please describe.**

Dockerhub rate limits

**Describe the solution you'd like**

Push loki and promtail to an additional registry like Quay or GHCR

**Additional context**

https://github.com/grafana/grafana/issues/27677

cgill27:
Google has a pull through caching mirror of dockerhub, use it instead of dockerhub when trying to get around dockerhub rate limits (if not paying for dockerhub account): mirror.gcr.io

onedr0p:
@cgill27 that doesn't seem to work...

```
~/Downloads on ☁️  N. Virginia
✖ docker pull mirror.gcr.io/grafana/loki:2.3.0
Error response from daemon: manifest for mirror.gcr.io/grafana/loki:2.3.0 not found: manifest unknown: Failed to fetch "2.3.0" from request "/v2/grafana/loki/manifests/2.3.0".

~/Downloads on ☁️  N. Virginia
✖ docker pull mirror.gcr.io/library/grafana/loki:2.3.0
Error response from daemon: manifest for mirror.gcr.io/library/grafana/loki:2.3.0 not found: manifest unknown: Failed to fetch "2.3.0" from request "/v2/library/grafana/loki/manifests/2.3.0".
```

Nemric:
I had this problem for all of my promtail instances and updates
I've created a [docker.io/registry container that act as a proxy](https://docs.docker.com/registry/configuration/#proxy) registry for all my docker hub images, it works great !

onedr0p:
Still interested in this becoming a reality, thanks!

onedr0p:
Not stake

MatteoJoliveau:
Definitely an issue for us as well, especially when running on AWS spot instances (which get frequently rotated, causing re-pulling). GitHub has a free container registry that is available automatically on all public repositories, it should be easy to integrate in your CI workflow.

onedr0p:
Given the recent dockerhub announcement of removing free team orgs, I'm wondering if the current `grafana` org will be affected? And if so perhaps this could finally have a high priority?

https://github.com/docker/hub-feedback/issues/2314

remram44:
It looks like the grafana account is a "verified publisher", which should make it exempt from rate limits.

danieljkemp:
> It looks like the grafana account is a "verified publisher", which should make it exempt from rate limits.

[...]

PR #8187, opened 2023-01-18
fix(helm): Use GHCR and bump nginx-unprivileged version

onedr0p:

There are no rate limits on GHCR, and also there's a newer version of nginx available.

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/values.yaml
+++ b/production/helm/loki/values.yaml
@@ -974,11 +974,11 @@ gateway:
     type: RollingUpdate
   image:
     # -- The Docker registry for the gateway image
-    registry: docker.io
+    registry: ghcr.io
     # -- The gateway image repository
     repository: nginxinc/nginx-unprivileged
     # -- The gateway image tag
-    tag: 1.19-alpine
+    tag: 1.23-alpine
     # -- The gateway image pull policy
     pullPolicy: IfNotPresent
   # -- The name of the PriorityClass for gateway pods


production/helm/loki/values.yaml
```diff
@@ -974,11 +974,11 @@ gateway:
     type: RollingUpdate
   image:
     # -- The Docker registry for the gateway image
-    registry: docker.io
+    registry: ghcr.io
```
trevorwhitney, line 977: We use `docker.io` for everything else, as well as the other "grafana labs supported" helm charts (ie. `mimir-distributed`). This is also where we publish our images, and link to in documentation. I would like to maintain that consistency and don't think we should make this change.

production/helm/loki/values.yaml
```diff
@@ -974,11 +974,11 @@ gateway:
...
     # -- The gateway image repository
     repository: nginxinc/nginx-unprivileged
     # -- The gateway image tag
-    tag: 1.19-alpine
+    tag: 1.23-alpine
```
trevorwhitney, line 981: I'm ok with this change if you want to separate it out.

trevorwhitney (changes requested):
Thanks for the contribution! For consistency reasons, I don't want to change the default registry, nor do I want to have a mix of different registries. A lot of people install this chart in a controlled environment where individual registries have to be approved, and most of them already allowed `docker.io`. Users who are concerned about rate limits can always provide different registries via values.

I would be open to upgrading Nginx if you would be will to separate that out. Thanks!

production/helm/loki/values.yaml
onedr0p: It's pretty unfortunate Grafana is so engrained to dockerhub given their 100/pulls per 6 hours rate limits and image retention nonsense. I'd love to see Grafana ditch dockerhub overall for quay, or GHCR or anything else that's more user friendly.

I know this is neither here nor there but if I worked for Grafana I'd make it my grand mission to migrate away from dockerhub for all projects, they are not friendly to the small guys anymore.

It's not great UX spinning up a fresh Kubernetes cluster and seeing logging or monitoring not working due to rate limits. Sure this can be solved with a pull thru cache, but Grafana could solve the problem for everyone by just migrating away from dockerhub. Sorry for my rant, it's just a passion of mine to see people migrate away from Dockerhub.

onedr0p:
It's pretty unfortunate Grafana is so engrained to dockerhub given their 100/pulls per 6 hours rate limits and image retention nonsense. I'd love to see Grafana ditch dockerhub overall for quay, or GHCR or anything else that's more user friendly.

I know this is neither here nor there but if I worked for Grafana I'd make it my grand mission to migrate away from dockerhub for all projects, they are not friendly to the small guys anymore.

It's not great UX spinning up a fresh Kubernetes cluster and seeing logging or monitoring not working due to pull rate limits. Sure this can be solved with a pull thru cache, but Grafana could solve the problem for everyone by just migrating away from dockerhub. Sorry for my rant, it's just a passion of mine to see people migrate away from Dockerhub.

I would love to see this one day. #4143

PR #8188, opened 2023-01-18
Helm: Default grafana dashboard label

nervo:
Add a default `grafana_dashboard` grafana dashboard label, so that dashboards could be detected by grafana out-of-the-box.

kube-prometheus-stack chart make the same choice here: https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L812

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/values.yaml
+++ b/production/helm/loki/values.yaml
@@ -510,8 +510,9 @@ monitoring:
     namespace: null
     # -- Additional annotations for the dashboards ConfigMap
     annotations: {}
-    # -- Additional labels for the dashboards ConfigMap
-    labels: {}
+    # -- Labels for the dashboards ConfigMap
+    labels:
+      grafana_dashboard: "1"
 
   # Recording rules for monitoring Loki, required for some dashboards
   rules:
--- a/docs/sources/installation/helm/reference.md
+++ b/docs/sources/installation/helm/reference.md
@@ -2007,9 +2007,11 @@ true
 		<tr>
 			<td>monitoring.dashboards.labels</td>
 			<td>object</td>
-			<td>Additional labels for the dashboards ConfigMap</td>
+			<td>Labels for the dashboards ConfigMap</td>
 			<td><pre lang="json">
-{}
+{
+  "grafana_dashboard": "1"
+}
 </pre>
 </td>
 		</tr>


production/helm/loki/values.yaml
```diff
@@ -490,8 +490,9 @@ monitoring:
     namespace: null
     # -- Additional annotations for the dashboards ConfigMap
     annotations: {}
-    # -- Additional labels for the dashboards ConfigMap
-    labels: {}
+    # -- Labels for the dashboards ConfigMap
+    labels:
```
trevorwhitney, line 494: what happens if someone provides custom labels? will it merge them or clobber this default? curious if instead we should just add this to the template. is this pattern documented anywhere?

trevorwhitney (changes requested):
This is a great idea, thank you. What are your thoughts on adding to the template itself? I'm worried about how this will work when someone also provides labels. Also, is this dashboard discovery pattern documented anywhere you know of?

production/helm/loki/values.yaml
nervo: Dicts are merged, so that custom labels are simply added, and this one can only be overwritten.

I don't think it's a good idea to add this to the template, this is simply a de facto standard defined in the official grafana chart (right here https://github.com/grafana/helm-charts/blob/7f6a6634d3c095f0e448810be781bcfe5238f9d0/charts/grafana/values.yaml#L822 ) and as you can see, its value can be customized ("1" by default, although this is by incident :) https://github.com/grafana/helm-charts/issues/1457 ). Under the hood, a sidecar container (https://github.com/kiwigrid/k8s-sidecar) is used to grab configmaps by labels.

Anyway, this leaves the final user with following choices:
- don't touch anything in default loki and grafana charts values, and let magic happen 🪄 
- change default label value in grafana chart, and set the same value in loki chart
- change default label **name** in grafana chart, add the same to loki chart. The old one `grafana_dashboard` is still there, but who cares ? Imho, advantages outweigh the disadvantages :)

trevorwhitney (approved):
thanks for the explanation! let's move ahead with this, thank you!

PR #7640, opened 2022-11-09
[helm] Add a flag to enable/disable persistentVolumeClaimRetentionPolicy feature

Guidurand:
**What this PR does / why we need it**:

add a flag to enable/disable persistentVolumeClaimRetentionPolicy feature which is available since kubernetes 1.23 but in alpha (StatefulSetAutoDeletePVC) and disabled by default on most cloud provider

I've set the flag to true by default, to not change the actual behavior
 
**Which issue(s) this PR fixes**:
Fixes #7249

**Special notes for your reviewer**:
@trevorwhitney 

**Checklist**
- [ X] Reviewed the `CONTRIBUTING.md` guide
- [ X] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #8193, opened 2023-01-18
Helm: Generalize enableStatefulSetAutoDeletePVC

nervo:
- This pr https://github.com/grafana/loki/pull/7640 introduced persistence `enableStatefulSetAutoDeletePVC` but *only*  on read pods
- This pr https://github.com/grafana/loki/pull/7920 introduced persistence `enableStatefulSetAutoDeletePVC` backend value, but did not implement it in related statefulset definition.

I propose to generalize persistence `enableStatefulSetAutoDeletePVC` on *all* statefulsets:
- read
- write
- backend
- singleBinary

To introduce this PR, i've set all default values to `true` but it could be relevant to think deeper about all specific component default value.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/values.yaml
+++ b/production/helm/loki/values.yaml
@@ -700,6 +700,8 @@ write:
   # -- Tolerations for write pods
   tolerations: []
   persistence:
+    # -- Enable StatefulSetAutoDeletePVC feature
+    enableStatefulSetAutoDeletePVC: false
     # -- Size of persistent disk
     size: 10Gi
     # -- Storage class to be used.
@@ -978,6 +980,8 @@ singleBinary:
   # -- Tolerations for single binary pods
   tolerations: []
   persistence:
+    # -- Enable StatefulSetAutoDeletePVC feature
+    enableStatefulSetAutoDeletePVC: true
     # -- Enable persistent disk
     enabled: true
     # -- Size of persistent disk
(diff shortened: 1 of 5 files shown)


production/helm/loki/values.yaml
```diff
@@ -678,6 +678,8 @@ write:
   # -- Tolerations for write pods
   tolerations: []
   persistence:
+    # -- Enable StatefulSetAutoDeletePVC feature
+    enableStatefulSetAutoDeletePVC: true
```
trevorwhitney, line 682: I'm nervous about this as the default for write. I think we should default to `false`, as I want to err on the side of not losing WALs when someone doesn't intentionally set this value.

trevorwhitney (changes requested):
Looks great! Totally onboard, but I think we should change the default for `write`. Thoughts?

production/helm/loki/values.yaml
nervo: I'm agree with you, let's go `false` by default for write. And maybe `false` by default too for singleBinary ?
What about backend ? `true` seems relevant imho.

trevorwhitney (approved):
Agreed with `false` for `write` and `singleBinary`. I think keeping it `true` for `read` and `backend` are good! Thanks for the contribution!

PR #8211, opened 2023-01-19
Use latest release of doc-validator image

jdbaldry:
- Fixes regression where non-Markdown files are considered for linting.
- Uses a canonical algorithm for converting headings to anchor identifiers taken from the Hugo source.


dannykopping (approved):
LGTM, thanks!

PR #8212, opened 2023-01-19
ingester: Add `ingester_memory_streams_labels_bytes` metric

kavirajk:

This is to give visibility on `labels` size of in-memory streams on the ingesters.

It caculate the size of labels via `len(labels.String())`. Meaning if you have 
```go
	lbs := labels.Labels{
		{Name: "foo", Value: "bar"},
		{Name: "instance", Value: "loki"},
	}
```
size would be number of bytes in `{foo="bar", instance="loki"}` which is `28`. NOTE: `len(string)` gives [bytes length in Go](https://pkg.go.dev/builtin#len)

Fixes #NA

Wanted this visibility for some of the work I'm doing it with ingester in-memory streams.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 * [8047](https://github.com/grafana/loki/pull/8047) **bboreham**: Dashboards: add k8s resource requests to CPU and memory panels.
 * [8061](https://github.com/grafana/loki/pull/8061) **kavirajk**: Remove circle from Loki OSS
 * [8131](https://github.com/grafana/loki/pull/8131) **jeschkies**: Compile Promtail ARM and ARM64 with journald support.
-* [8212](https://github.com/grafana/loki/pull/8212) **kavirajk**: ingester: Add `ingester_memory_streams_labels_size metric` for more visibility of size of metadata of in-memory streams.
+* [8212](https://github.com/grafana/loki/pull/8212) **kavirajk**: ingester: Add `ingester_memory_streams_labels_bytes metric` for more visibility of size of metadata of in-memory streams.
 
 ##### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * [8047](https://github.com/grafana/loki/pull/8047) **bboreham**: Dashboards: add k8s resource requests to CPU and memory panels.
 * [8061](https://github.com/grafana/loki/pull/8061) **kavirajk**: Remove circle from Loki OSS
 * [8131](https://github.com/grafana/loki/pull/8131) **jeschkies**: Compile Promtail ARM and ARM64 with journald support.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 * [8047](https://github.com/grafana/loki/pull/8047) **bboreham**: Dashboards: add k8s resource requests to CPU and memory panels.
 * [8061](https://github.com/grafana/loki/pull/8061) **kavirajk**: Remove circle from Loki OSS
 * [8131](https://github.com/grafana/loki/pull/8131) **jeschkies**: Compile Promtail ARM and ARM64 with journald support.
+* [8212](https://github.com/grafana/loki/pull/8212) **kavirajk**: ingester: Add `ingester_memory_streams_labels_size metric` for more visibility of size of metadata of in-memory streams.
```
MichelHollands, line 29:
```suggestion
* [8212](https://github.com/grafana/loki/pull/8212) **kavirajk**: ingester: Add `ingester_memory_streams_labels_bytes metric` for more visibility of size of metadata of in-memory streams.
```

MichelHollands: The title of the PR also shows a different metric

pkg/ingester/instance.go
```diff
@@ -53,6 +53,11 @@ var (
 		Name:      "ingester_memory_streams",
 		Help:      "The total number of streams in memory per tenant.",
 	}, []string{"tenant"})
+	memoryStreamsLabels = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
MichelHollands, line 56: `memoryStreamsLabelsBytes` would be a better name for the variable. The current name suggests the count of the labels to me.

kavirajk: good point 👍 Done.

MichelHollands (approved):
LGTM

pkg/ingester/instance.go
```diff
@@ -300,6 +306,7 @@ func (i *instance) createStreamByFP(ls labels.Labels, fp model.Fingerprint) *str
 
 	i.streamsCreatedTotal.Inc()
 	memoryStreams.WithLabelValues(i.instanceID).Inc()
+	memoryStreamsLabels.WithLabelValues(i.instanceID).Add(float64(len(s.labels.String())))
```
dannykopping, line 309: `labels.Labels` has a `Len()` function which you should use instead I believe.

kavirajk: Actually `Len()` of `labels.Labels` just return total number of elements in `labels.Labels` slice. We need a byte size of all the labels.

```
type Labels []Label

func (ls Labels) Len() int           { return len(ls) }

```

dannykopping: Oops!

dannykopping (approved):
LGTM

kavirajk:
Do not merge this. Currently it's per tenant metric. Going to change to non-per tenant metric. Thinking bit about it, I actually need only metadata size per ingester level (at least now). So don't want to create un-necessary per-tenant metric here.

PR #8215, opened 2023-01-19
Make secret job fail when not logged in

trevorwhitney:
* Also replace enterprise-logs cluster with helm installed version

This PR fixes the `secrets` target in the k3d makefile to fail (and remove failed files) when not logged into 1password. Without this change empty files were being created, causing the target to not run again, but the installation to fail.

This PR also removes the jsonnet Enterprise Logs cluster in favor of the helm installed one.

I would like to eventually remove `loki-distributed`, but we need to first provide values for a loki cluster via helm.


tools/dev/k3d/Makefile
```diff
@@ -66,11 +58,20 @@ secrets: secrets/grafana.jwt secrets/gel.jwt
...
+	@if ! test -s "$(CURDIR)/secrets/grafana.jwt"; then \
+		echo "Failed to fetch Grafana license from 1Password, are you logged in?"; \
+		rm -f $(CURDIR)/secrets/grafana.jwt; \
+		exit 1; \
+	fi
 
 secrets/gel.jwt:
 	mkdir -p secrets/
-	op document get "loki/gel.jwt" --output=$(CURDIR)/secrets/gel.jwt
+	op document get "loki/gel.jwt" --output=$(CURDIR)/secrets/gel.jwt || rm -f $(CURDIR)/secrets/gel.jwt
```
ssncferreira, line 70: Nice 🌟

tools/dev/k3d/Makefile
```diff
@@ -66,11 +58,20 @@ secrets: secrets/grafana.jwt secrets/gel.jwt
 
 secrets/grafana.jwt:
 	mkdir -p secrets/
-	op document get "loki/grafana.jwt" --output=$(CURDIR)/secrets/grafana.jwt
+	op document get "loki/grafana.jwt" --output=$(CURDIR)/secrets/grafana.jwt || rm -f $(CURDIR)/secrets/grafana.jwt
+	@if ! test -s "$(CURDIR)/secrets/grafana.jwt"; then \
+		echo "Failed to fetch Grafana license from 1Password, are you logged in?"; \
+		rm -f $(CURDIR)/secrets/grafana.jwt; \
```
ssncferreira, line 64: This instruction is not needed, right? Because the file doesn't exist

ssncferreira (approved):
LGTM 🤩

tools/dev/k3d/Makefile
ssncferreira: Ah, forget it! The file can exist, but be empty. In that case, we should also add an `rm -f` to the `secrets/gel.jwt` condition.

trevorwhitney: Yup, that's the thought :smile:

tools/dev/k3d/Makefile
```diff
@@ -66,11 +58,20 @@ secrets: secrets/grafana.jwt secrets/gel.jwt
...
+		exit 1; \
+	fi
 
 secrets/gel.jwt:
 	mkdir -p secrets/
-	op document get "loki/gel.jwt" --output=$(CURDIR)/secrets/gel.jwt
+	op document get "loki/gel.jwt" --output=$(CURDIR)/secrets/gel.jwt || rm -f $(CURDIR)/secrets/gel.jwt
+	@if ! test -s "$(CURDIR)/secrets/gel.jwt"; then \
+		echo "Failed to fetch GEL license from 1Password, are you logged in?"; \
```
ssncferreira, line 72: We should also rm the gel license file in this case, right? As it might be empty:
```suggestion
		echo "Failed to fetch GEL license from 1Password, are you logged in?"; \
		rm -f $(CURDIR)/secrets/gel.jwt; \
```

jeschkies (approved):

tools/dev/k3d/Makefile
trevorwhitney: yes, thank you!

PR #8218, opened 2023-01-19
[helm] add local k3d loki cluster

trevorwhitney:

This removes the `loki-distributed` k3d cluster, and adds on OSS `loki` cluster that is installed using `helm install`. This follows the pattern used by the `enterprise-logs` cluster, and both clusters share the same base `helm-cluster` jsonnet configuration.

To spin up a local Loki cluster, run `make k3d-loki` from the root of the repo.
To spin up a local GEL cluster, run `make k3d-enterprise-logs` from the root of the repo.
To tear either cluster down, run `make k3d-down` from the root of the repo.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


trevorwhitney:
cc @jeschkies @ssncferreira as I know you both have used these k3d envs for GEL

Makefile
```diff
@@ -773,3 +774,12 @@ check-example-config-doc: generate-example-config-doc
 		echo -e "(Don't forget to check in the generated files when finished)\n"; \
 		exit 1; \
 	fi
+
+k3d-loki:
```
jeschkies, line 778: The name is a little confusing. How about `tanka-loki` instead?

jeschkies (changes requested):

Makefile
trevorwhitney: maybe `helm-loki-cluster`? it doesn't use `tanka` to install Loki, so I think `tanka-loki` is also misleading.

trevorwhitney: or `k3d-loki-cluster`? It spins up a `k3d` cluster using `tanka`, which it then `helm installs` Loki into.

jeschkies: How about `dev-k3d-...`

jeschkies (approved):

PR #8224, opened 2023-01-20
Loki: Remove validation of matchers when performing label values lookup.

slim-bean:
the curren validation fails on != style matchers when they are the only matchers present, requiring at least one `=` matcher.

However for looking up label values in TSDB, by design it uses a `desired_label != ""` style query, as such it should be allowed to also include these as matchers.

Additional matchers on this Label Value queries typically only exist with LBAC headers adding additional labels.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/storage/stores/shipper/indexgateway/gateway.go
+++ b/pkg/storage/stores/shipper/indexgateway/gateway.go
@@ -2,7 +2,7 @@ package indexgateway
 
 import (
 	"context"
-	"errors"
+	"fmt"
 	"sync"
 
 	"github.com/go-kit/log"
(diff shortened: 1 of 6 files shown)


pkg/storage/stores/shipper/indexgateway/gateway.go
```diff
@@ -218,10 +219,16 @@ func (g *Gateway) LabelValuesForMetricName(ctx context.Context, req *logproto.La
...
 	if req.Matchers != syntax.EmptyMatchers {
-		matchers, err = syntax.ParseMatchers(req.Matchers)
+		expr, err := syntax.ParseExprWithoutValidation(req.Matchers)
 		if err != nil {
 			return nil, err
 		}
+
+		matcherExpr, ok := expr.(*syntax.MatchersExpr)
+		if !ok {
+			return nil, errors.New("invalid label matchers found")
```
dannykopping, line 229:
```suggestion
			return nil, fmt.Errorf("invalid label matchers found of type %T", expr)
```

I'm hoping we have another layer of indirection above this that gives a more human-readable response though

PR #8231, opened 2023-01-21
Lambda-promtail: Add support for processing SQS messages, add promtailClient Type, add logger, upgrade dependencies and fix unexpected flushing behaviors

CCOLLOT:

## Context
I want to :
+ Use lambda-promtail to ingest AWS VPC flow logs from S3 files, deploying it as an AWS lambda function.
+ Configure the s3 Notification to send messages to SQS instead of directly to lambda in order to be able to spin up a secondary SQS queue to be used as a dead-letter queue lambda can send unsuccessfully processed events to.
    + This way I am able to use the `AWS SQS redrive` feature to process logs after fixing an outage in the ingestion flow and I am not limited by lambda giving up after 2 (default) retries.

I faced several issues:
+ lambda-promtail tries to flush a batch when it gets bigger than `BATCH_SIZE`. The current implementation does not reset the size of the batch, leading to a flush to loki/promtail for every single added line after reaching `BATCH_SIZE` (and making the below file descriptor issue much worse because the `http.Client` is currently created in the `send` function.
+ lambda-promtail reached the quota of file_descriptors of AWS lambda (1024) very quickly due to http.Client being created thousands of times when processing large files (a VPC flow log file can reach dozens of MB very quick, meaning hundreds of thousands of lines.

## Added Feature
+ Add support for processing events from an SQS queue
     + Side effect: add support for on-failure recovery using and SQS queue.
     + If merged, I would happily upload my terraform sample code to setup the S3 --> SQS main--> lambda --> SQS DLQ -->redrive to SQS main

## Fixed bugs

+ The `flushBatch` function would reset the stream after flushing but it wouldn't reset the size of the batch, leading to lambda-promtail flushing every single following line to `loki/promtail`. In my case it led to logs being processed very slowly by lambda when the total size is way above the default `batchSize`. I would reach the 15 min timeout on AWS lambda for almost every single file.
+ The `http.Client`used to flush logs to `loki/promtail` was being recreated thousands of times, which made the process quickly run out of file descriptors (limited to 1024 in AWS lambda). I set it as a global variable to avoid impacting other parts of the code for now but if required we could do it a little cleaner via dependency injection.

After fixing those bugs my average lambda execution time for vpc flow log files (size between 10mb and 100mb) went from timing out at 15 min to ~3 seconds.

## Upgrades:
+ Go version to `1.19`
+ Dependencies using `go get -u ./...`

## Other
+ Reorganized imports 
+ Fixed some typos

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md`


tools/lambda-promtail/lambda-promtail/main.go
```diff
@@ -165,6 +174,11 @@ func handler(ctx context.Context, ev map[string]interface{}) error {
 		return processCWEvent(ctx, evt)
 	case *events.KinesisEvent:
 		return processKinesisEvent(ctx, evt)
+	case *events.SQSEvent:
+		return processSQSEvent(ctx, evt)
+	//When setting up S3Notification to SQS, a test event is first sent to the queue
```
cstyan, line 179: is this behaviour documented somewhere?

tools/lambda-promtail/lambda-promtail/main_test.go
```diff
@@ -1,9 +1,10 @@
 package main
 
 import (
+	"testing"
+
```
cstyan, line 5: can we avoid whitespace/import reordering changes, just makes reviews for new features easier :) thank you

tools/lambda-promtail/lambda-promtail/promtail.go
```diff
@@ -129,7 +129,7 @@ func (b *batch) flushBatch(ctx context.Context) error {
 	}
 
 	b.streams = make(map[string]*logproto.Stream)
-
+	b.size = 0
```
cstyan, line 132: what is this for? fixing one of the bugs you're referring to I assume?

tools/lambda-promtail/lambda-promtail/s3.go
```diff
@@ -51,7 +52,7 @@ func getS3Object(ctx context.Context, labels map[string]string) (io.ReadCloser,
 		s3Client = s3.NewFromConfig(cfg)
 		s3Clients[labels["bucket_region"]] = s3Client
 	}
-
+	fmt.Println("fetching", labels["key"])
```
cstyan, line 55: did you mean to include this println?

cstyan (commented):
few minor comments

reviewing this I also started thinking  that it might make sense to implement a `client` type within lambda promtail that we could attach functions to, and it could store the `promtailClient` reference rather than creating more package level globals

tools/lambda-promtail/lambda-promtail/s3.go
CCOLLOT: I used it for debugging purposes and ended up leaving it there since it's a useful piece of information (it helps immediately know what event/file is being processed) and because it prints only 1 line per lambda invocation (so it's not like it's going to cost a lot in Cloudwatch logs).

As the code is growing maybe we should have a more structured approach.
Maybe we could remove this line and open up another MR to use the standard logger (and set the LOG_LEVEL or DEBUG using an environment variable)? WDYT?

tools/lambda-promtail/lambda-promtail/promtail.go
CCOLLOT: > what is this for? fixing one of the bugs you're referring to I assume?

Yes because batch.size is incremented as we scan the file and after each added line we compare batch.size and (max)batchSize to know if we need to flush the batch to the target (loki/promtail). 

It needs to be reset (to 0) after batch.size gets bigger than batchSize otherwise we end up sending a request to the target to flush a single line for each new line (so it becomes super slow because of the thousands of round-trip calls)

I think I will update this PR to provide a proper reset() method to make it clearer

tools/lambda-promtail/lambda-promtail/main.go
CCOLLOT: Yes [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html#:~:text=Example%20messages,following%20test%20message.)

CCOLLOT:
Thanks! I will update this PR with your comment. 
Would you like me to implement the client type in this PR ?

tools/lambda-promtail/lambda-promtail/main_test.go
CCOLLOT: Sure I will remove them 😅

CCOLLOT:
Thanks! I will update this PR with your comments (and the new client type)

tools/lambda-promtail/lambda-promtail/main_test.go
cstyan: thanks!

whitespace/import cleanup PRs are welcome! I just want to be sure I can easily see everything relevant to the things you're trying to fix here with this PR :)

tools/lambda-promtail/lambda-promtail/main.go
cstyan: can we include that link in this comment please

tools/lambda-promtail/lambda-promtail/promtail.go
cstyan: a `resetBatch` function sounds like a good idea :+1:

tools/lambda-promtail/lambda-promtail/s3.go
cstyan: I'm fine with keeping as is or implementing something more structured.

CCOLLOT: Okay I implemented the logger in my latest commit, I went with go-kit since it's the one used in the main project

tools/lambda-promtail/lambda-promtail/promtail.go
CCOLLOT: Done in my latest commit 👍

cstyan:
Looks like there's a conflict with your CHANGELOD.md changes. Might be easiest to remove that commit, merge upstream main into your branch, and then make your changelog entry again.  I'll have a look at the client and logger changes in the morning.

tools/lambda-promtail/lambda-promtail/promtail_client.go
```diff
@@ -0,0 +1,53 @@
...
+	"crypto/tls"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/dskit/backoff"
+)
+
+type IPromtailClient interface {
```
cstyan, line 14: we can just name the interface `client`

tools/lambda-promtail/lambda-promtail/s3.go
```diff
@@ -39,41 +40,30 @@ const (
...
-	obj, err := s3Client.GetObject(ctx,
-		&s3.GetObjectInput{
-			Bucket:              aws.String(labels["bucket"]),
-			Key:                 aws.String(labels["key"]),
-			ExpectedBucketOwner: aws.String(labels["bucketOwner"]),
-		})
-
-	if err != nil {
-		fmt.Printf("Failed to get object %s from bucket %s on account %s\n", labels["key"], labels["bucket"], labels["bucketOwner"])
-		return nil, err
```
cstyan, line 64: why are we removing the object fetching code here and moving it?

tools/lambda-promtail/lambda-promtail/promtail_client.go
CCOLLOT: Sure, will do! 👍  Actually I thought that it might get confusing later because there are other clients (s3clients etc...) that could be mocked and passed around using interfaces.  I guess it's not an issue to call this interface `client` for now and we still can move the clients to dedicated packages later on.

docs/sources/clients/lambda-promtail/_index.md
```diff
@@ -109,6 +109,12 @@ This workflow allows ingesting AWS loadbalancer logs stored on S3 to Loki.
 
 Cloudfront [real-time logs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/real-time-logs.html) can be sent to a Kinesis data stream. The data stream can be mapped to be an [event source](https://docs.aws.amazon.com/lambda/latest/dg/invocation-eventsourcemapping.html) for lambda-promtail to deliver the logs to Loki.
 
+### Triggering Lambda-Promtail via SQS
+For AWS services supporting sending messages to SQS (ex: S3 with an S3 Notification to SQS), events can be processed through an `SQS queue with lambda trigger` instead of directly configuring the source service to trigger lambda. Lambda-promtail will retrieve the nested events from the SQS messages' body and process them as if them came directly from the source service.
```
JStickler, line 113:
```suggestion
For AWS services supporting sending messages to SQS (for example, S3 with an S3 Notification to SQS), events can be processed through an `SQS queue with lambda trigger` instead of directly configuring the source service to trigger lambda. Lambda-promtail will retrieve the nested events from the SQS messages' body and process them as if them came directly from the source service.
```

docs/sources/clients/lambda-promtail/_index.md
```diff
@@ -109,6 +109,12 @@ This workflow allows ingesting AWS loadbalancer logs stored on S3 to Loki.
...
+
+### On-Failure log recovery using SQS
+Triggering lambda-promtail through SQS allows handling on-failure recovery of the logs using a secondary SQS queue as a dead-letter-queue. Lambda can be set up such as unsuccessfully processed messages will be sent to the DLQ. After fixing the issue, operators will be able to reprocess the messages by sending back messages from the DLQ to the source queue using the SQS `DLQ redrive` feature.
```
JStickler, line 116:
```suggestion
Triggering lambda-promtail through SQS allows handling on-failure recovery of the logs using a secondary SQS queue as a dead-letter-queue (DLQ). You can configure lambda so that unsuccessfully processed messages will be sent to the DLQ. After fixing the issue, operators will be able to reprocess the messages by sending back messages from the DLQ to the source queue using the SQS `DLQ redrive` feature.
```

JStickler (commented):
[Docs squad]  Couple of small suggestions.

docs/sources/clients/lambda-promtail/_index.md
CCOLLOT: Thanks for your suggestion! I will include it in the PR

docs/sources/clients/lambda-promtail/_index.md
CCOLLOT: Thanks for your suggestion! I will include it in the PR

CCOLLOT: Done, also added some links to the AWS documentation

JStickler (approved):
[docs sqad]  docs look good, I'll leave it to engineering to approve the code.

tools/lambda-promtail/lambda-promtail/s3.go
cstyan: :+1: got it

cstyan:
I appreciate the bug fixes :+1: 

Having never used SQS myself I can't speak to the validity of the SQS specific code or how you're talking about using it, but I trust that you or other users will be using it enough to point out any issues that need fixing going forward.

I'll leave this for a day just to have one last read through and see if anyone else has comments, and then merge.

CCOLLOT:
> I appreciate the bug fixes :+1: 
> 
> Having never used SQS myself I can't speak to the validity of the SQS specific code or how you're talking about using it, but I trust that you or other users will be using it enough to point out any issues that need fixing going forward.
> 
> I'll leave this for a day just to have one last read through and see if anyone else has comments, and then merge.

Glad I could help. I've been running this code for days/weeks now and it runs smoothly. 

Thanks for the review. ✌️

cstyan:
hey @CCOLLOT my bad for leaving this unmerged, do you mind just resolving the conflicts? I will merge after that

cstyan (approved):

PR #8234, opened 2023-01-23
Helm - Loki: fix fullname for services

houqp:

Loki services were created with the short name, but the ingress is referencing these services using the fullname.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/Chart.yaml
+++ b/production/helm/loki/Chart.yaml
@@ -4,7 +4,7 @@ name: loki
 description: Helm chart for Grafana Loki in simple, scalable mode
 type: application
 appVersion: 2.7.0
-version: 4.4.0
+version: 4.4.1
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -449,7 +449,7 @@ Create the service endpoint including port for MinIO.
 
 {{/* Configure the correct name for the memberlist service */}}
 {{- define "loki.memberlist" -}}
-{{ include "loki.name" . }}-memberlist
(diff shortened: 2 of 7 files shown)


jeschkies (approved):

trevorwhitney (changes requested):
How does this affect the included dashboards? I think this will affect the `job` label, and will break many of the dashboards, which are looking for `job=~"$namespace/(loki|enterprise-logs)-$target`. Instead let's fix the ingress.

houqp:
@trevorwhitney  should we fix the dashboards instead? The problem with the current setup being if you deploy multiple instances into the same namespace with different fullnames, it will result in conflicts. I do think this is a legitimate bug based on the variable name and what's in the ingress rule.

trevorwhitney:
> if you deploy multiple instances into the same namespace with different fullnames, it will result in conflicts

Is that a common use case? I wouldn't ever think to do that, and would use different namespaces.

That being said, I would very much welcome some better templating and compiling of the dashboards, but we may need to fix that upstream in the [mixin](https://github.com/grafana/loki/tree/main/production/loki-mixin) or improve the process by which we compile that mixin into the helm chart.

Currently those dashboards were just copied over from the compiled mixin, and a few things chnaged manually, but I don't think that's tenable long term, and have on my list to push the changes from https://github.com/grafana/loki/pull/8261 upstream back into the mixin. 

A better solution might be a job that compiles mixin at installation time so we can dynamically build the matcher? Or maybe we need to add something to our build pipeline that compiles the mixin with a templated matcher (ie `job="{{ .Release.namespace}}/{{ include 'loki.fullWriteName' }}`)?

MasslessParticle:
There's been no movement on this in 5 months. I'm going to close for inactivity

PR #8244, opened 2023-01-23
Loki: Log hashed queries emitted by `metrics.go`

DylanGuedes:
- Adds a new `query_hash` label pair to queries logged by `metrics.go`. As the name suggests, `query_hash` is just the hashed version of the query being logged.
  - Especially useful when grouping results/behavior of different queries. Ex, you can easily find specific query patterns with `sum by(query_hash)` .

Ex:
```
level=info ts=2023-01-23T14:45:52.743780661Z caller=metrics.go:149 component=frontend org_id=fake latency=fast query="{stream=\"stdout\",pod=\"loki-canary-abcde\"}" query_hash=1360642565 query_type=limited range_type=range length=20s start_delta=58m27.642790018s end_delta=58m7.64279027s step=1s duration=4.761564089s status=200 limit=1000 returned_lines=0 throughput=78kB total_bytes=369kB total_entries=41 store_chunks_download_time=3.037076ms queue_time=1m5.039459998s subqueries=18 cache_chunk_req=3 cache_chunk_hit=3 cache_chunk_bytes_stored=0 cache_chunk_bytes_fetched=2881268 cache_chunk_download_time=2.49811ms cache_index_req=33 cache_index_hit=33 cache_index_download_time=61.548315ms cache_result_req=1 cache_result_hit=0 cache_result_download_time=217.84µs
```

N/A

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/engine.go
+++ b/pkg/logql/engine.go
@@ -203,10 +203,11 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
 	log, ctx := spanlogger.New(ctx, "query.Exec")
 	defer log.Finish()
 
+	queryHash := hashedQuery(q.params.Query())
 	if GetRangeType(q.params) == InstantType {
-		level.Info(logutil.WithContext(ctx, q.logger)).Log("msg", "executing query", "type", "instant", "query", q.params.Query())
+		level.Info(logutil.WithContext(ctx, q.logger)).Log("msg", "executing query", "type", "instant", "query", q.params.Query(), "query_hash", queryHash)
 	} else {
-		level.Info(logutil.WithContext(ctx, q.logger)).Log("msg", "executing query", "type", "range", "query", q.params.Query(), "length", q.params.End().Sub(q.params.Start()), "step", q.params.Step())
+		level.Info(logutil.WithContext(ctx, q.logger)).Log("msg", "executing query", "type", "range", "query", q.params.Query(), "length", q.params.End().Sub(q.params.Start()), "step", q.params.Step(), "query_hash", queryHash)
 	}
 
 	rangeType := GetRangeType(q.params)
--- a/pkg/logql/metrics.go
+++ b/pkg/logql/metrics.go
@@ -2,6 +2,7 @@ package logql
 
 import (
 	"context"
+	"hash/fnv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


pkg/logql/metrics.go
```diff
@@ -108,11 +109,12 @@ func RecordRangeAndInstantQueryMetrics(
 
 	queryTags, _ := ctx.Value(httpreq.QueryTagsHTTPHeader).(string) // it's ok to be empty.
 
-	logValues := make([]interface{}, 0, 20)
+	logValues := make([]interface{}, 0, 21)
```
dannykopping, line 112: By my count, there are `30` items

dannykopping (approved):
Great idea @DylanGuedes!

pkg/logql/metrics.go
DylanGuedes: Haha your math is better than mine. Fixed!

slim-bean:
We should also add the hash to the "executing query" log lines to make correlating started vs finished queries much easier

pkg/logql/engine.go
```diff
@@ -203,10 +203,11 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
 	log, ctx := spanlogger.New(ctx, "query.Exec")
 	defer log.Finish()
 
+	queryHash := hashedQuery(q.params.Query())
```
slim-bean, line 206: I think we should verify that the same hash is produced in metrics.go vs this executing query line for queries like:

```
sum(rate(...)) by (label)
```

This query will show up in metrics.go as:

```
sum by (label) (rate(...))
```

And i'm not exactly sure where this re-write happens?

DylanGuedes: Good catch. I pushed a test that checks whether the hashed query emitted by engine vs metrics.go are equal or not. Apparently, `sum(...) by(label)` resulted in the same query hash in both places so not sure I'm missing something :thinking: 

if you know of other interesting queries for me to test I can add them as new test cases in the scenario I implemented on `engine_test.go`

slim-bean: ok, no worries! thanks for checking, my recollection here is likely missing something.

slim-bean (approved):
LGTM!

PR #8245, opened 2023-01-23
stores/series: always return entries to pool

bboreham:

Instead of doing the `Get` inside a function and `Put` outside, do them both outside. This makes it clearer that it's ok to `defer` before checking for errors, which avoids dropping a pooled slice as garbage on any error.

A particular case is `CardinalityExceededError` where the dropped slice will be large, and which is treated as benign unless it happens on all matchers.

Change the pooled objects to pointer type to avoid the Go runtime creating a pointer wrapper, and we can stop silencing the warning about this from the staticcheck linter.

**Checklist**
- NA Documentation added
- NA Tests updated
- NA `CHANGELOG.md` updated - no user-visible difference.
- NA Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

### DIFF
--- a/pkg/storage/stores/series/series_index_store.go
+++ b/pkg/storage/stores/series/series_index_store.go
@@ -316,15 +316,15 @@ func (c *indexReaderWriter) LabelValuesForMetricName(ctx context.Context, userID
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


MasslessParticle (approved):
LGTM, Thanks!

PR #8250, opened 2023-01-24
logs more metadata when failing to recover head from WAL in TSDB

owen-d:
Ran into this in a set of confusing circumstances. I haven't figured out the cause yet, so I'm adding some extra logging to get further next time I see it.

### DIFF
--- a/pkg/storage/stores/tsdb/head_manager.go
+++ b/pkg/storage/stores/tsdb/head_manager.go
@@ -524,11 +524,11 @@ func recoverHead(dir string, heads *tenantHeads, wals []WALIdentifier) error {
 				if len(rec.Chks.Chks) > 0 {
 					tenant, ok := seriesMap[rec.UserID]
 					if !ok {
-						return errors.New("found tsdb chunk metas without user in WAL replay")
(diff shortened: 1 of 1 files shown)


pkg/storage/stores/tsdb/head_manager.go
```diff
@@ -524,11 +524,11 @@ func recoverHead(dir string, heads *tenantHeads, wals []WALIdentifier) error {
 				if len(rec.Chks.Chks) > 0 {
 					tenant, ok := seriesMap[rec.UserID]
 					if !ok {
-						return errors.New("found tsdb chunk metas without user in WAL replay")
+						return fmt.Errorf("found tsdb chunk metas without user in WAL replay (period=%s): %+v", id.String(), *rec)
```
kavirajk, line 527: Just double checking `%+v` of this complex nested struct is what we need in error message?

```
type WALRecord struct {
	UserID      string
	Series      record.RefSeries
	Fingerprint uint64
	Chks        ChunkMetasRecord
}

```

JordanRushing (approved):

pkg/storage/stores/tsdb/head_manager.go
owen-d: Yeah, see https://go.dev/play/p/a-aLqq7s_Ix

PR #8251, opened 2023-01-24
index-store: fix indexing of chunks overlapping multiple schemas

sandeepsukhani:
We noticed that when a chunk overlaps multiple schemas, the index entries are being built for whole chunk range for both schemas. We should be building index entries in each stores, limited to their schema interval.
For example, `schema[0]` is `boltdb-shipper` starting at `0` and `schema[1]` is `tsdb` starting at `10`. If chunk 8-11 is flushed, it should be indexed in `schema[0]` with index entries for range 8-9, and for `schema[1]`, it should be 10-11.

However, due to my mistake in changeset [here](https://github.com/grafana/loki/commit/c8c8303a6a6767b8e6e196e9e160dbb2c6887273#diff-c2c327d47d019a916c9c1b5e668bae2e5b05a89533a90001e42cbf334ed861eaL101) I changed the behaviour, now we are indexing chunk for 9-11 in both the stores. This is causing us to build boltdb files that are out of range for the store, and if the prefix for both `boltdb-shipper` and `tsdb` are the same in the above example, we will end up adding a boltdb file to `tsdb` index table. This causes compaction and index gateways to fail to read the files since the file names and formats are different for both index types.

I have fixed the issue and added a test to verify the indexing of chunks when the same index table name prefix is used for `boltdb-shipper` and `tsdb` index stores.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/storage/stores/indexshipper/shipper.go
+++ b/pkg/storage/stores/indexshipper/shipper.go
@@ -70,6 +70,10 @@ type Config struct {
 	IngesterDBRetainPeriod time.Duration
 }
 
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	cfg.RegisterFlagsWithPrefix("", f)
+}
+
 // RegisterFlagsWithPrefix registers flags.
 func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	cfg.IndexGatewayClientConfig.RegisterFlagsWithPrefix(prefix+"shipper.index-gateway-client", f)
--- a/pkg/storage/stores/series/series_index_store.go
+++ b/pkg/storage/stores/series/series_index_store.go
@@ -82,8 +82,8 @@ func NewIndexReaderWriter(schemaCfg config.SchemaConfig, schema series_index.Ser
 	}
 }
 
-func (c *indexReaderWriter) IndexChunk(ctx context.Context, chk chunk.Chunk) error {
-	writeReqs, keysToCache, err := c.calculateIndexEntries(ctx, chk.From, chk.Through, chk)
+func (c *indexReaderWriter) IndexChunk(ctx context.Context, from, through model.Time, chk chunk.Chunk) error {
+	writeReqs, keysToCache, err := c.calculateIndexEntries(ctx, from, through, chk)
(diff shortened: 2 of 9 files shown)


pkg/storage/stores/indexshipper/shipper.go
```diff
@@ -70,6 +70,10 @@ type Config struct {
 	IngesterDBRetainPeriod time.Duration
 }
 
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	cfg.RegisterFlagsWithPrefix("", f)
+}
+
```
chaudum, line 76: This change is unrelated, right? Could be reverted.

pkg/storage/stores/series/series_index_store.go
```diff
@@ -82,8 +82,8 @@ func NewIndexReaderWriter(schemaCfg config.SchemaConfig, schema series_index.Ser
 	}
 }
 
-func (c *indexReaderWriter) IndexChunk(ctx context.Context, chk chunk.Chunk) error {
-	writeReqs, keysToCache, err := c.calculateIndexEntries(ctx, chk.From, chk.Through, chk)
+func (c *indexReaderWriter) IndexChunk(ctx context.Context, from, through model.Time, chk chunk.Chunk) error {
+	writeReqs, keysToCache, err := c.calculateIndexEntries(ctx, from, through, chk)
```
kavirajk, line 86: :+1:

kavirajk (approved):
Nice catch :) also love the test case :heart: 

how did you find this bug? Happed in any of our prod cell?

chaudum (approved):

pkg/storage/stores/indexshipper/shipper.go
sandeepsukhani: It was causing the new test I added to fail while setting the defaults, and it should have been added in the first place anyways.

sandeepsukhani:
> how did you find this bug? Happed in any of our prod cell?

It luckily didn't cause any problems in any of our prod cells since we used a new prefix. We did saw it causing problems in our secondary ops cells while testing this exact use case.

PR #8257, opened 2023-01-24
Setting Julie Stickler as codeowner for docs

dannykopping:
Welcome @JStickler!!

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -14,4 +14,4 @@
 
 # Documentation owner: Grafana Docs Squad
 # Grafana Docs Squad does not own the content, but assists with editing and updates.
-/docs/ @grafana/docs-logs
+/docs/ @JStickler


kavirajk (approved):
:tada: 

Q: why not add @JStickler to `@grafana/docs-logs` squad instead?

dannykopping:
> 🎉
> 
> Q: why not add @JStickler to `@grafana/docs-logs` squad instead?

We don't need the whole group to be informed whenever there's a Loki docs PR; that was a temporary measure when we had no tech writer.

kavirajk:
>We don't need the whole group to be informed whenever there's a Loki docs PR;

Yea. but that's why it's @grafana/docs-logs. Specific to logs?. My concern is mainly creating PR everytime. Instead add or remove people to the right group which we already have it as codeowner.

dannykopping:
> > We don't need the whole group to be informed whenever there's a Loki docs PR;
> 
> Yea. but that's why it's @grafana/docs-logs. Specific to logs?. My concern is mainly creating PR everytime. Instead add or remove people to the right group which we already have it as codeowner.

Ah, I missed that detail. Well, it's functionally the same for now - if we add any more tech writers later to the Loki team we'll move it back to the group

osg-grafana:
@GrafanaWriter, can you weigh in here given that it is now an exception, given all the other subteams that we use.

osg-grafana:
> > We don't need the whole group to be informed whenever there's a Loki docs PR;
> 
> Yea. but that's why it's @grafana/docs-logs. Specific to logs?. My concern is mainly creating PR everytime. Instead add or remove people to the right group which we already have it as codeowner.

Yes, this is the logic. We now have to manage that Loki does things differently, so heavily in favor or reverting this PR.

osg-grafana:
Relates to https://github.com/grafana/technical-documentation/issues/520.

dannykopping:
@osg-grafana I'm happy to revert this change.

https://github.com/orgs/grafana/teams/docs-logs/members seems to only have yourself and Kim in it; should you replace yourselves with Julie?

osg-grafana:
> @osg-grafana I'm happy to revert this change.
> 
> https://github.com/orgs/grafana/teams/docs-logs/members seems to only have yourself and Kim in it; should you replace yourselves with Julie?

Yes, please replace me and Kim with Julie. Thank you!

dannykopping:
@osg-grafana I don't have permission to do so

PR #8259, opened 2023-01-24
Logproto: Extract push.proto from logproto package to the separate module

mar4uk:
This PR is aimed to extract push proto to a separate module. It will allow using push proto in other repositories without bringing the whole Loki as a dependency.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated


pkg/push/push.pb.go
```diff
@@ -0,0 +1,1292 @@
...
+	cc *grpc.ClientConn
+}
+
+func NewPusherClient(cc *grpc.ClientConn) PusherClient {
+	return &pusherClient{cc}
+}
+
+func (c *pusherClient) Push(ctx context.Context, in *PushRequest, opts ...grpc.CallOption) (*PushResponse, error) {
+	out := new(PushResponse)
+	err := c.cc.Invoke(ctx, "/push.Pusher/Push", in, out, opts...)
```
owen-d, line 452: I'm worried this may be a breaking change since the path is used in the method name here.

pkg/push/push.proto
```diff
@@ -0,0 +1,40 @@
+syntax = "proto3";
+
+package push;
```
owen-d, line 3: To address breaking concerns, I wonder if we can name this with the same package as before: `package logproto;`

owen-d (commented):
I like this PR, but I think we'll have to address the proto package name changing and breaking method compatibility. See https://github.com/grafana/loki/pull/6852 for another example of this in the past.

Hopefully we can have separate golang pkgs (the new push pkg and the pre-existing logproto) both register the same grpc package namespace (logproto). If that doesn't work, we could look at code-generating a compatible grpc package from the same definition subset.

pkg/push/push.proto
mar4uk: I need the `push` package to be a separate module, so I can import it outside of loki. That's why it has its own name.
If push would be a part of logproto package then I would need to import the whole logproto package from the outside.
That's will be ok, but logproto can't be a separate module right now, because it has dependencies on other loki packages:
github.com/grafana/loki/pkg/querier/queryrange/queryrangebase/definitions
github.com/grafana/loki/pkg/util
github.com/grafana/loki/pkg/storage/stores/tsdb/index
github.com/grafana/loki/pkg/logqlmodel/stats

To make logproto a separate module I would need to remove these dependencies from it. As I understood that dependencies were brought to logproto on purpose

pkg/push/push.pb.go
mar4uk: can I write some test to check if I broke something? What does this path affect?

pkg/push/push.proto
mar4uk: it seems working! renamed package from `push` to `logproto`

PR #8260, opened 2023-01-24
Helm: ensure valid label values when version has digest

sathieu:

Fix this:
```console
$ helm install loki grafana/loki \
  --set loki.image.tag=2.7.1@sha256:d69f377ecfdbb3f72086a180dcd7c2f02c795cf1867bbeb61606b42a8d41a557
ServiceAccount "loki" is invalid: [
  metadata.labels: Invalid value: "2.7.1@sha256:d69f377ecfdbb3f72086a180dcd7c2f02c795cf1867bbeb61606b42a8d41a557": must be no more than 63 characters,
  metadata.labels: Invalid value: "2.7.1@sha256:d69f377ecfdbb3f72086a180dcd7c2f02c795cf1867bbeb61606b42a8d41a557": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
],
Service "loki" is invalid: [
  metadata.labels: Invalid value: "2.7.1@sha256:d69f377ecfdbb3f72086a180dcd7c2f02c795cf1867bbeb61606b42a8d41a557": must be no more than 63 characters,
  metadata.labels: Invalid value: "2.7.1@sha256:d69f377ecfdbb3f72086a180dcd7c2f02c795cf1867bbeb61606b42a8d41a557": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
],
[...]
```

To ensure it is allowed, and leads to:

```yaml    
      app.kubernetes.io/version: "2.7.1-sha256-d69f377ecfdbb3f72086a180dcd7c2f02c795cf1867bbeb616"
```

See https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

(see also https://gitlab.com/kubitus-project/kubitus-installer/-/jobs/3649809256)

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -1,4 +1,12 @@
 {{/*
+Enforce valid label value.
+See https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+*/}}
+{{- define "loki.validLabelValue" -}}
+{{- (regexReplaceAllLiteral "[^a-zA-Z0-9._-]" . "-") | trunc 63 | trimSuffix "-" | trimSuffix "_" | trimSuffix "." }}
+{{- end }}
+
+{{/*
 Expand the name of the chart.
 */}}
 {{- define "loki.name" -}}
@@ -80,7 +88,7 @@ Common labels
 helm.sh/chart: {{ include "loki.chart" . }}
 {{ include "loki.selectorLabels" . }}
 {{- if or (.Chart.AppVersion) (.Values.loki.image.tag) }}
-app.kubernetes.io/version: {{ .Values.loki.image.tag | default .Chart.AppVersion | quote }}
+app.kubernetes.io/version: {{ include "loki.validLabelValue" (.Values.loki.image.tag | default .Chart.AppVersion) | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end }}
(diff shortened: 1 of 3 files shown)


production/helm/loki/templates/_helpers.tpl
```diff
@@ -80,7 +80,7 @@ Common labels
 helm.sh/chart: {{ include "loki.chart" . }}
 {{ include "loki.selectorLabels" . }}
 {{- if or (.Chart.AppVersion) (.Values.loki.image.tag) }}
-app.kubernetes.io/version: {{ .Values.loki.image.tag | default .Chart.AppVersion | quote }}
+app.kubernetes.io/version: {{ (regexReplaceAllLiteral "@.*" (.Values.loki.image.tag | default .Chart.AppVersion) "") | quote }}
```
trevorwhitney, line 83: 2 questions.

1. should we replace it with a `:` instead of nothing so there's still a visible separator
2. we probably also want to enforce it's truncated to 63 characters?

trevorwhitney (changes requested):
Thanks for bringing this up, nice find. 2 small questions regarding the implementation, but +1 to moving forward with this.

production/helm/loki/templates/_helpers.tpl
sathieu: Thanks for your review @trevorwhitney.

As `:` is not allowed, I've replaced it with `-`. I've moved this to a new defined template `"loki.validLabelValue"`. Back to you :ping_pong:!

trevorwhitney (approved):
Looks great, thanks for the contribution!

PR #8263, opened 2023-01-24
fix remaining docs issues

MasslessParticle:
Turns out the `docs-validator` doesn't do an entire check of the docs. This PR fixes some build issues and some additional linting issues.


JStickler (approved):
LGTM

Issue #8197, opened 2023-01-18
WAL support for Promtail


thepalbi:
This issue is a spin off the the following PR: https://github.com/grafana/loki/pull/7993
Design is being discussed in the following [doc](https://docs.google.com/document/d/1dDPRtGDb_ckNjtnbQ4DT8ZkfC2635l8tAXD5QOZ7gjc/edit#).

## Steps

- [x] Re-license some of Loki's WAL bits https://github.com/grafana/loki/pull/8315
- [x] Implement WAL writer side https://github.com/grafana/loki/pull/8267
- [x] Implement WAL reader side https://github.com/grafana/loki/pull/8302
- [x] Improve resource consumption of WAL watcher in low-resources hardware. See [this](https://github.com/prometheus/prometheus/issues/11625) and [this](https://github.com/grafana/loki/pull/8302#pullrequestreview-1353613973)
  - https://github.com/grafana/loki/pull/9135 
- [ ] Refactor all remote write clients instantiation under the new `client.Manager` https://github.com/grafana/loki/pull/9908
- [ ] [Design] how to efficiently cleanup old segments
  - This should aim in a perfect scenario, to cleanup old segments once all entries they had were remote-written by each rw client.
- [ ] [implementation] how to efficiently cleanup old segments
- [ ] Implement WAL replay
  - Might be implemented just letting the watchers re-read all existing segments
  - Requires a better old segment cleanup, if not, turning on and off promtail will end up in a lot of duplicates
- [x] Tune WAL for buffering
  - [x] Batch series and samples wal writes https://github.com/grafana/loki/pull/8489

## After
- [x] Revamp the WAL entries approach, avoiding the need to write series in each WAL entry (maybe already tackled by https://github.com/grafana/loki/pull/8489)

thepalbi:
Closing as this was implemented in https://grafana.com/docs/agent/next/flow/reference/components/loki.write/

PR #8267, opened 2023-01-24
Promtail WAL support: Implement writer side

thepalbi:

This PR is the first in a series that will implement WAL support into Promtail. The main objective of this PR is implement the writer side of WAL. 
Note that enabling WAL with the implementation as is will cause entries scraped to be written to WAL, but not remote written. Therefore, should not be used until the whole implementation is completed.

Apart form this, this PR includes a new primitive that will be extender in the following editions: `client.Manager`. This new component will be responsible for instantiating the correct remote write client, and handle the orchestration from the channel scraping targets write to, to the clients themselves. Also, will handle the graceful shutdown of bespoke clients.

Part of #8197 

- Since there's a licensing conflict between primitives under the `ingester` package (AGPL), dependencies have been copied over under promtail, which is Apache.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


clients/pkg/promtail/wal/reader.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"fmt"
+
+	"github.com/prometheus/common/model"
+
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+
+	"github.com/grafana/loki/pkg/ingester"
+	"github.com/grafana/loki/pkg/util"
+	wal2 "github.com/grafana/loki/pkg/util/wal"
```
cstyan, line 12: unfortunately this code is AGPL so we'll need to either need to go through the necessary bits to move it to a different path that can be licensed differently, or duplicate the functionality

clients/pkg/promtail/client/manager.go
```diff
@@ -0,0 +1,49 @@
...
+
+// Manager manages remote write client instantiation, and connects the related components to orchestrate the flow of api.Entry
+// from the scrape targets, to the remote write clients themselves.
+//
+// Right now it just supports instantiating the WAL writer side of the future-to-be WAL enabled client. In follow-up
+// work, tracked in https://github.com/grafana/loki/issues/8197, this Manager will be responsible for instantiating all client
+// types: Logger, Multi and WAL.
+type Manager struct {
+	clientChannelHandler api.EntryHandler
+}
```
cstyan, line 19: We want each steps PR to be as standalone as possible and ideally not introduce a codepath that does nothing, let me know if this makes sense:
- in this PR the Manager should also contain a field `clients []Client`
- manager should also implement the Client interface, and it's functions would just range over it's internal `clients`

This would allow us to instantiate the Manager and any current implementation of the `Client` interface, and still be able to write both to the WAL and to the client correctly and thus "enable" the WAL but not break remote writing of log lines.

clients/pkg/promtail/client/manager.go
```diff
@@ -0,0 +1,49 @@
...
+
+// NewManager creates a new Manager
+func NewManager(reg prometheus.Registerer, logger log.Logger, walCfg wal.Config) (*Manager, error) {
```
cstyan, line 22: this should take the general `clients` block of the promtail config and instantiate all the clients

clients/pkg/promtail/client/manager.go
```diff
@@ -0,0 +1,49 @@
...
+	// TODO: refactor this to instantiate all clients types
+	pWAL, err := wal.New(walCfg, logger, reg)
+	if err != nil {
+		return nil, err
+	}
+	walWriter := wal.NewWriter(pWAL, logger)
```
cstyan, line 28: new writer should be created in promtails `main.go`, and passed to all the scrape jobs from there

here is where we would create the watcher for each client in a follow up PR

clients/pkg/promtail/client/client_test.go
```diff
@@ -319,7 +311,7 @@ func TestClient_Handle(t *testing.T) {
 			receivedReqsChan := make(chan receivedReq, 10)
 
 			// Start a local HTTP server
-			server := httptest.NewServer(createServerHandler(receivedReqsChan, testData.serverResponseStatus))
```
thepalbi, line 322: refactored out some testing utils

clients/pkg/promtail/wal/encoding.go
```diff
@@ -0,0 +1,256 @@
+package wal
```
thepalbi, line 1: should I add a comment on top of the copied files due to licensing, clarifying that?

clients/pkg/promtail/client/manager.go
cstyan: I don't think I agree. The components that bring logs into the system and store them to the WAL should be separated from the components that need to read logs from the WAL and send them over the wire to Loki.

The discovery/scrape targets are responsible for reading log files or grabbing logs data from an external system, and thus should probably be managed by a single overarching struct which also either creates the WAL instance and passes an interface for it to those target routines, or takes the interface as part of it's own construction and can pass it to the targets that way.

The client manager struct is then completely separate, and only takes in some interface for receiving logs (which in reality could be just the existing channel passing we do, or the WAL watcher interface/channel we're working on now, as long as both satisfy the same interface) that it can then duplicate and send to all clients that have been started.

thepalbi: After taking a second look I agree with you. It's not part of the client manager concern to route entries emitted by the scraping target to the WAL itself. The WAL itself should act as barrier between the scraping side (scraping targets + writing to the WAL) and the remote write side (reading from the WAL + remote writing).

The only part I'm not that convinced is where to the logic of, for the time being, writing entries produced by scraping target to both the WAL and the clients. I feel it's a smell to create a separate components that does this fanout since it will be removed shorty after.

I took a stab at adding this inside the WAL writer, I think it's not a bad place considering it will live there just until we get [this done](https://github.com/grafana/loki/pull/8302).

clients/pkg/promtail/wal/reader.go
thepalbi: Re-licensed this bit as discussed. This [PR](https://github.com/grafana/loki/pull/8315) contains those changes to keep them separate. Will rebase after.

clients/pkg/promtail/promtail.go
```diff
@@ -162,7 +166,20 @@ func (p *Promtail) reloadConfig(cfg *config.Config) error {
...
+	if cfg.WAL.Enabled {
+		pWAL, err := wal.New(cfg.WAL, p.logger, p.reg)
+		if err != nil {
+			return fmt.Errorf("error starting WAL: %w", err)
+		}
+		p.walWriter = wal.NewWriter(pWAL, p.logger, p.client)
+		entriesSink = p.walWriter
+	}
+
+	tms, err := targets.NewTargetManagers(p, p.reg, p.logger, cfg.PositionsConfig, entriesSink, cfg.ScrapeConfig, &cfg.TargetConfig)
```
cstyan, line 182: I think it probably makes more sense to have promtail have a field that is an entry handler, and during the loading of the config here we create an entry handler that wraps both the list of clients and the walWriter, rather than doing the propagation of the channel data within the wal writer.

clients/pkg/promtail/client/manager.go
```diff
@@ -0,0 +1,107 @@
...
+	clients := make([]Client, 0, len(clientCfgs))
+	for _, cfg := range clientCfgs {
+		client, err := New(metrics, cfg, maxStreams, maxLineSize, maxLineSizeTruncate, logger)
+		if err != nil {
+			return nil, err
+		}
+
+		// Don't allow duplicate clients, we have client specific metrics that need at least one unique label value (name).
+		if _, ok := clientsCheck[client.Name()]; ok {
+			return nil, fmt.Errorf("duplicate client configs are not allowed, found duplicate for URL: %s", cfg.URL)
```
kgeckhart, line 47: Does this error need to reference the name instead of the URL?

thepalbi: Yhea I think it probably should, since it's whats checking, just copied for this the [current impl](https://github.com/grafana/loki/blob/main/clients/pkg/promtail/client/multi.go#L39). Will update it!

clients/pkg/promtail/promtail.go
thepalbi: Yhea same, wasn't that happy with how it looked adding the fanout inside the writer. Moved it to promtail.go, and added a new slim component that does the fanout.

thepalbi: In [this commit](https://github.com/grafana/loki/pull/8267/commits/5af6e20ec688caf4e327f32346680f1626c02061) the fanout is moved out of the `Writer`, and done in `promtail.go` via a `EntryHandlerFanouter`, component that takes n entry handlers and exposes a single one, performing a fanout. Could be re-used when the `client.Manager` needs to implement the same functionality as the current multiclient.

thepalbi:
@cstyan , the latest changes should be from [this commit](https://github.com/grafana/loki/pull/8267/commits/a8533b953d334742c05a2e8b98c21dfaa8b71cea) onwards. I tried to remove all unnecessary changes after the rebase, and sorted all imports to make the reviewal process easier. To give a summary for what I changes since last time:
- Do fanout to wal writer and client manager as mentioned [here](https://github.com/grafana/loki/pull/8267#discussion_r1089563543)
- Implemented simple old wal segments cleaner, this is the [starting commit](https://github.com/grafana/loki/pull/8267/commits/442052537f2dace69af210d3ea5a8d65a7e22b42), but you could just look for in under `pkg/promtail/wal/writer.go`
- Cleanups

I implemented the old segments cleanup in this PR to at least support alongside with this PR, a very simple cleanup system allows a wal-enabled promtail to run without running the host out of disk

clients/pkg/promtail/wal/writer.go
```diff
@@ -54,6 +78,30 @@ func (wrt *Writer) start() {
 			wrt.toClients.Chan() <- e
 		}
 	}()
+	// WAL cleanup routine that cleans old segments
+	wrt.wg.Add(1)
+	go func() {
+		defer wrt.wg.Done()
+		// By cleaning every 10th of the configured threshold for considering a segment old, we are allowing a maximum slip
```
cstyan, line 96: :+1:

clients/pkg/promtail/utils/entries.go
```diff
@@ -0,0 +1,45 @@
...
+	multiplex.wg.Add(1)
+	go func() {
+		defer multiplex.wg.Done()
+		for e := range multiplex.entries {
+			for _, handler := range multiplex.handlers {
+				handler.Chan() <- e
+			}
+		}
+	}()
+	return multiplex
```
cstyan, line 32: shouldn't we be running each channel send in a go routine and, have the `wg.Add(multiplex.Entries*multiplex.handlers)` and then inside each goroutinne `defer wg.Done()`

clients/pkg/promtail/utils/entries.go
```diff
@@ -0,0 +1,45 @@
+package utils
+
+import (
+	"sync"
+
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+)
+
+// EntryHandlerFanouter implements api.EntryHandler, fanning out received entries to one or multiple channels.
+type EntryHandlerFanouter struct {
```
cstyan, line 10: nit: can we call this `FanoutEntryHandler`

clients/pkg/promtail/utils/entries.go
```diff
@@ -0,0 +1,45 @@
...
+	entries  chan api.Entry
+	handlers []api.EntryHandler
+
+	once sync.Once
+	wg   sync.WaitGroup
+}
+
+func NewEntryHandlerFanouter(handlers ...api.EntryHandler) *EntryHandlerFanouter {
```
cstyan, line 18: name would change here too

clients/pkg/promtail/wal/writer.go
```diff
@@ -0,0 +1,258 @@
...
+// Also, since Writer is responsible for all changing operations over the WAL, therefore a routine is run for cleaning
+// old segments.
+type Writer struct {
+	entries     chan api.Entry
+	log         log.Logger
+	wg          sync.WaitGroup
+	once        sync.Once
+	wal         WAL
+	entryWriter *entryWriter
+	toClients   api.EntryHandler
```
cstyan, line 41: it looks like `toClients` and related tests are no longer needed

clients/pkg/promtail/utils/entries.go
thepalbi: That's actually not a bad idea, we'd have:
1. one main go routine which listens to the channel the `FanoutEntryHandler` exposes, FEH for shorten
2. upon receiving one entry, the fanout is done with one go routine per channel send
3. the go routine in 1) will only read the next message from the channel once all in 2) have completed

We need that wait until all have finished in 3), because if not, we could be reading yet another message without knowing the previous one has been fanned out properly.

Like it!

cstyan (commented):
almost there, I think we just need a nice way to shutdown the fanout handler and ensure data that was queued in it's channel is sent

clients/pkg/promtail/utils/entries.go
```diff
@@ -6,38 +6,51 @@ import (
...
 		entries:  make(chan api.Entry),
 		handlers: handlers,
 	}
-	multiplex.wg.Add(1)
+	eh.start()
+	return eh
+}
+
+func (eh *FanoutEntryHandler) start() {
+	eh.wg.Add(1)
```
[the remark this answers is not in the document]
thepalbi: that wg is the one that allows us to wait on `fanout.Stop()` that the main routine gracefully stopped right? Without it, we might return from stop, while the routine is sending an entry it already read from the `eh.entries` channel.

clients/pkg/promtail/utils/entries.go
```diff
@@ -6,38 +6,51 @@ import (
...
 	entries  chan api.Entry
 	handlers []api.EntryHandler
 
 	once sync.Once
 	wg   sync.WaitGroup
 }
 
-func NewEntryHandlerFanouter(handlers ...api.EntryHandler) *EntryHandlerFanouter {
-	multiplex := &EntryHandlerFanouter{
+func NewFanoutEntryHandler(handlers ...api.EntryHandler) *FanoutEntryHandler {
```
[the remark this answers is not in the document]
thepalbi: What this new channel you mention for? Right now when we stop the following occurs:
1. Within `eh.Stop()`
  1. the `eh.entries` channel is closed
  2. the routine running stop (main one) blocks until `eh.wg` is cleared
2. within the routine in `eh.start()`:
  1. If it was waiting for a new entry to be sent in line 31 -> break out of the for loop, signal `eh.wg` and return
  2. if it was sending an entry -> N go routines are running sending entries, and the routine in 2) is waiting for them to finish

I think we could add some kind of timeout mechanism that is started upon calling `Stop()`, that from that moment on if the entry being processed in 2) is not sent before T time, exit. wdyt?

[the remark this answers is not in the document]
thepalbi: Explored the idea [here](https://github.com/grafana/loki/compare/pablo/wal-part-1...pablo/wal-fanout-with-timeout?expand=1), let me know what you think 😃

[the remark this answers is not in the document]
cstyan: Ah yeah you're right, I missed the stop function. A select loop to try and send remaining data before exiting is what I was thinking. Prometheus' `runShard` [function](https://github.com/prometheus/prometheus/blob/main/storage/remote/queue_manager.go#L1328) hard exists on a specific context being cancelled, and the hard shutdown is enforced after a timeout within the stop function , see [here](https://github.com/prometheus/prometheus/blob/main/storage/remote/queue_manager.go#L1119-L1152).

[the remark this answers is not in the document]
thepalbi: Got it, then it's something like I thought [here](https://github.com/grafana/loki/compare/pablo/wal-part-1...pablo/wal-fanout-with-timeout?expand=1). Like those examples, will review it and implement something that mechanism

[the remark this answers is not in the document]
thepalbi: @cstyan done, let me know what you think!

clients/pkg/promtail/utils/entries.go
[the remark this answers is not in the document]
thepalbi: Removed that wait group in favour of a hand-made one with a channel in order to be able to wait on it, much like in [here](https://github.com/prometheus/prometheus/blob/main/storage/remote/queue_manager.go#L1072)

clients/pkg/promtail/promtail.go
```diff
@@ -139,14 +148,44 @@ func (p *Promtail) reloadConfig(cfg *config.Config) error {
...
+		p.walWriter, err = wal.NewWriter(cfg.WAL, p.logger, p.reg)
+		if err != nil {
+			return fmt.Errorf("failed to create wal writer: %w", err)
+		}
+		entryHandlers = append(entryHandlers, p.walWriter)
+	}
+
+	p.entriesFanout = utils.NewFanoutEntryHandler(timeoutUntilFanoutHardStop, entryHandlers...)
+
+	tms, err := targets.NewTargetManagers(p, p.reg, p.logger, cfg.PositionsConfig, p.entriesFanout, cfg.ScrapeConfig, &cfg.TargetConfig)
```
cstyan, line 188: last comment, I think; it feels to me like the creation of the WAL writer should be above within the previous `else if cfg.WAL.Enabled`, and we could reassign the fanout writer to `p.client`

if the usage of the actual Client type for `p.client` makes that difficult lets just find a way to combine these two if statements so that we don't make our job harder when refactoring, we could make `p.walWriter` a pointer, and check if the pointer isn't null when deciding whether to create the entries fanout?

thepalbi: Yhea also wasn't fond of that double if now that you mention. Moved a bit the code so that we have a single if branch related to wal-enabled behaviour. I think we can keep the fanout if wal is disabled, since it adds the graceful shutdown which is not there otherwise.

Also, we should keep a Client implementing interface in `p.client`, since it's part of the interface the main promtail objects exposes. In the future, this will be just the `client.Manager`.

Let me know wdyt!

cstyan (approved):
LGTM, thanks for all your hard work here @thepalbi 

I'll wait until Monday morning before merging just in case any else has been meaning to take a look at this.

PR #8272, opened 2023-01-25
Remove creationTimestamp: null

patsevanton:
CreationTimestamp is a timestamp representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations. Clients may not set this value. It is represented in RFC3339 form and is in UTC. https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L179

Signed-off-by: Anton Patsev <patsev.anton@gmail.com>

Which of these should I fill out in this PR ?
Should i change version operator?

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/operator/bundle/manifests/loki-operator-controller-manager-metrics-service_v1_service.yaml
+++ b/operator/bundle/manifests/loki-operator-controller-manager-metrics-service_v1_service.yaml
@@ -3,7 +3,6 @@ kind: Service
 metadata:
   annotations:
     service.beta.openshift.io/serving-cert-secret-name: loki-operator-metrics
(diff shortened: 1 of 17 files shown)


periklis:
@patsevanton If you are a user of the Loki Operator and you spotted an error by the generated `creationTimestamp: null` instances then please let me know more about. If no error occurred, I will advise to leave the manifests in the operator folder untouched. FWIW these files are generated by the operator-sdk (underlying kubebuilder) and so far this worked without any incidents.

PR #8281, opened 2023-01-25
Fix pubsub pull target

MasslessParticle:
- Retry when the subscription returns an error instead of hanging forever
- If the number of retries is exceeded, return the underlying error

This PR also does some substantial work on the tests to allow testing error behaviors.

### DIFF
--- a/clients/pkg/promtail/targets/gcplog/pull_target.go
+++ b/clients/pkg/promtail/targets/gcplog/pull_target.go
@@ -1,20 +1,34 @@
 package gcplog
 
 import (
-	"context"
-	"sync"
-
 	"cloud.google.com/go/pubsub"
+	"context"
+	"fmt"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/backoff"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/relabel"
 	"google.golang.org/api/option"
+	"io"
+	"sync"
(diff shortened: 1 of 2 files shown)


clients/pkg/promtail/targets/gcplog/pull_target.go
```diff
@@ -83,42 +101,57 @@ func (t *pullTarget) run() error {
...
+		})
+		if lastError != nil {
+			level.Error(t.logger).Log("msg", "failed to receive pubsub messages", "error", lastError)
+			t.metrics.gcplogErrors.WithLabelValues(t.config.ProjectID).Inc()
+			t.metrics.gcplogTargetLastSuccessScrape.WithLabelValues(t.config.ProjectID, t.config.Subscription).SetToCurrentTime()
+			t.backoff.Wait()
+		}
+	}
+
+	if t.ctx.Err() == nil && t.backoff.Err() != nil {
```
DylanGuedes, line 150: hmm why do you want to skip this check when `t.ctx.Err()`?

DylanGuedes (approved):
lgtm, just a question

clients/pkg/promtail/targets/gcplog/pull_target.go
MasslessParticle: `backoff.Err()` returns the `ctx.Err()` if the context has been canceled. We handle the context-canceled case already so this ensures we're only handling errors from the pubsub

PR #8283, opened 2023-01-25
Retry getting logs from pubsub forever

MasslessParticle:
Exiting from the run loop of a started target isn't meaningful. Retry forever


dannykopping (approved):
LGTM!

PR #8261, opened 2023-01-24
[helm] Fix helm dashboards

trevorwhitney:
**What this PR does / why we need it**:

This PR adds documentation on how to get the included Dashboards to work for a fresh helm installation. It also modifies the local k3d loki cluster to install the necessary relabelings to the `kube-prometheus-stack`, `kube-state-metrics`, and `node-exporter` helm charts so those metrics will work with the included dashboards.
**Checklist**
- [x] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [x] Documentation added
- ~[ ] Tests updated~
- [ ] `CHANGELOG.md` updated
- ~[ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`~

Issue #8185, opened 2023-01-17
[Helm] Allow single binary with Object Storage 


trevorwhitney:
**Is your feature request related to a problem? Please describe.**
Currently, the Helm chart only allows two modes of operation, single binary with local filesystem storage and scalable with object storage. This is missing a set of use cases which are two small to merit the resources of a replicated scalable deployment (3x read and 3x write pods), but would still like replication (3x single binary pods).

**Describe the solution you'd like**
 A way that exposes this configuration without exposing the "deployment type". Ideally, a newcomer to Loki should not need to know about the difference between "single binary" and "scalable" deployment. Instead, they should provide information about the size of their environment, and their replication needs, and the chart should provide a "recommended" deployment.

At the very least, this deployment method should be possible with the helm chart, even if it requires overriding the whole config file.

**Describe alternatives you've considered**
An alternative was provided in https://github.com/grafana/loki/pull/7942, however it forces the operator to know about the different deployment types.

djotanov:
Personally I think that the operator should know the difference between single binary and microservices deployment type, and that is already nicely documented in LOKI documentation: https://grafana.com/docs/loki/latest/fundamentals/architecture/deployment-modes/
Deployment type should be a conscious choice instead of imposed by this chart.
Hiding that in this chart is unnecessarily limiting this chart, and actually complicating things.

trevorwhitney:
@djotanov I disagree. I don't see a reason to add an additional configuration value when we can infer the deployment model from existing configuration options. I think the less config there is the better. But please share your feedback on #8284 to encourage wider discussion. Thanks!

silverlyra:
I’m in the process of upgrading a Loki 2.4 deployment (Helm chart 2.8) to Loki 2.7 (chart 4.4.2). I have been running in the single-binary mode, writing to S3, and was disappointed when saw in the docs for the new Helm chart that this is no longer supported.

So I was pleasantly surprised when I brought up the new Helm release, and it came up in single binary mode. 🤔

```
❯ kubectl -n loki get pods
NAME     READY   STATUS    RESTARTS   AGE
loki-0   2/2     Running   0          39m
```

I checked the S3 bucket and it’s definitely writing there.

It seems I accidentally found a workaround by using `storage_config` to set this up, as that’s how I did things in the old Helm chart:

```yaml
storage:
  type: aws

storage_config:
  aws:
    s3: "s3://<region>/<bucket>"
  boltdb_shipper:
    shared_store: s3
```

[...]

PR #8284, opened 2023-01-25
[helm] Single binary object storage

### LINKED ISSUE #8269: [Helm] Enable the gateway for single binary mode

Currently the gateway does not work with single binary mode. The is problematic since the single binary service config is not exposed. This is also problematic for those that want to run single binary mode in an HA fashion: #8185 

Enable the gateway when using single binary mode.

Exposing config of the single binary service, as in #8147.

trevorwhitney:

The PR allows an operator to install Loki in Single Binary mode in an HA fashion, meaning more than 1 Single Binary replica. In order for this to work, it is not possible to specify an object storage backend when running in Single Binary mode. The gateway is also now always enabled, for both Single Binary and Scalable modes.

Fixes #8269 #8185

This PR builds off #8261, will rebase after that is merged and diff will be smaller.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/installation/helm/install-monolithic/index.md
+++ b/docs/sources/installation/helm/install-monolithic/index.md
@@ -12,9 +12,7 @@ keywords: []
 
 This Helm Chart installation runs the Grafana Loki *single binary* within a Kubernetes cluster.
 
-If the storage type is set to `filesystem`, this chart configures Loki to run the `all` target in a [monolithic mode]({{<relref "../../../fundamentals/architecture/deployment-modes#monolithic-mode">}}), designed to work with a filesystem storage. It will also configure meta-monitoring of metrics and logs.
-
-It is not possible to install the single binary with a different storage type.
+If the `singleBinary.replicas` value is set to something greater than 0, it will configure Loki to run the `all` target in a [monolithic mode]({{<relref "../../../fundamentals/architecture/deployment-modes#monolithic-mode">}}), designed to work with a filesystem storage. If using just 1 replica, this *single binary* deployment will be configured to use object storage. If you would like to run *single binary* in a replicated, highly available mode, you must supply an object storage configuration and set `singleBinary.replicas` to a value greater than or equal to 2.
 
 **Before you begin: Software Requirements**
 
@@ -35,9 +33,9 @@ It is not possible to install the single binary with a different storage type.
     helm repo update
     ```
 
-1. Configure the `filesystem` storage:
+1. Create the configuration file `values.yaml`:
 
-    - Create the configuration file `values.yaml`:
+    - If running a single replica of Loki, configure the `filesystem` storage:
 
       ```yaml
       loki:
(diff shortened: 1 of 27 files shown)


JStickler:
@trevorwhitney, you tagged me on this, but I don't see much in the way of documentation changes.

Not sure if we need to update this topic, which mentions using a shared object store?  https://grafana.com/docs/loki/latest/fundamentals/architecture/deployment-modes/

trevorwhitney:
@JStickler sorry the bot tagged you when I accidentally was removing changes to a docs file due to a botched rebase. No docs need on this one :smile:

jeschkies (commented):
Could you add some documentation on how this could be used?

jeschkies (approved):

docs/sources/installation/helm/install-monolithic/index.md
```diff
@@ -12,9 +12,7 @@ keywords: []
 
 This Helm Chart installation runs the Grafana Loki *single binary* within a Kubernetes cluster.
 
-If the storage type is set to `filesystem`, this chart configures Loki to run the `all` target in a [monolithic mode]({{<relref "../../../fundamentals/architecture/deployment-modes#monolithic-mode">}}), designed to work with a filesystem storage. It will also configure meta-monitoring of metrics and logs.
-
-It is not possible to install the single binary with a different storage type.
+If the `singleBinary.replicas` value is set to something greater than 0, it will configure Loki to run the `all` target in a [monolithic mode]({{<relref "../../../fundamentals/architecture/deployment-modes#monolithic-mode">}}), designed to work with a filesystem storage. If using just 1 replica, this *single binary* deployment will be configured to use object storage. If you would like to run *single binary* in a replicated, highly available mode, you must supply an object storage configuration and set `singleBinary.replicas` to a value greater than or equal to 2.
```
JStickler, line 15:
```suggestion
If you set the `singleBinary.replicas` value to 1, this chart configures Loki to run the `all` target in a [monolithic mode]({{<relref "../../../fundamentals/architecture/deployment-modes#monolithic-mode">}}), designed to work with a filesystem storage. It will also configure meta-monitoring of metrics and logs.
If you set the `singleBinary.replicas` value to 2 or more, this chart configures Loki to run a *single binary* in a replicated, highly available mode.  When running replicas of a single binary, you must configure object storage.
```

JStickler (approved):
[Docs squad]  One small suggestion for rewording to make it clearer that there are now two options here.

production/helm/loki/templates/validate.yaml
```diff
@@ -14,3 +14,11 @@
 {{- fail "Helm test requires a prometheusAddress for an instance scraping the Loki canary's metrics"}}
 {{- end }}
 
+{{- if and (eq (include "loki.isUsingObjectStorage" .) "false") (gt (int .Values.singleBinary.replicas) 1) }}
+{{- fail "Cannot run more than 1 Single Binary replica without an object storage backend."}}
+{{- end }}
+
+{{- if and (eq (include "loki.isUsingObjectStorage" .) "false") (or (gt (int .Values.backend.replicas) 0) (gt (int .Values.read.replicas) 0) (gt (int .Values.write.replicas) 0)) }}
```
sapslaj, line 21: One (potentially negative?) side effect of this check is that you have to explicitly scale down the `backend`, `read`, and `write` replicas if you're using single binary with filesystem storage. If you don't, then this check will fail.

I wonder if we need to also `and` in something like `(eq (int .Values.singleBinary.replicas) 0)` in here, like we do when defining `loki.deployment.isScalable`

sapslaj (commented):
This is great @trevorwhitney! I just had that one thing about the replica counts come up in my testing, otherwise it looks great.

production/helm/loki/templates/validate.yaml
trevorwhitney: Good catch, added that extra check to the condition.

PR #8285, opened 2023-01-25
retry+backoff querier->qf response

owen-d:
When the querier acquires a connection to the frontend from it's shared pool, but the connection is closed/no longer valid/errors, said querier logs the failure and moves on. The frontend, however, will wait for the querier's response until the entire query times out. This effectively blocks a subquery from being retried and prevents the entire query itself from succeeding. Furthermore, the query frontend will retain partial results in memory while waiting for the timeout.

I've added retry+backoff support to this operation in an attempt to minimize the likelihood of this happening.

### DIFF
--- a/pkg/querier/worker/scheduler_processor.go
+++ b/pkg/querier/worker/scheduler_processor.go
@@ -12,6 +12,7 @@ import (
 	"github.com/gogo/status"
 	"github.com/grafana/dskit/backoff"
 	"github.com/grafana/dskit/grpcclient"
+	"github.com/grafana/dskit/multierror"
 	"github.com/grafana/dskit/ring/client"
 	"github.com/grafana/dskit/services"
 	otgrpc "github.com/opentracing-contrib/go-grpc"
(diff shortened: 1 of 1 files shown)


MichelHollands (approved):
LGTM

pkg/querier/worker/scheduler_processor.go
```diff
@@ -189,17 +192,61 @@ func (sp *schedulerProcessor) runRequest(ctx context.Context, logger log.Logger,
...
+			level.Error(logger).Log("msg", "error acquiring client", "err", err)
+			errs.Add(err)
+			pool.RemoveClientFor(addr)
+			backoff.Wait()
+			continue
+		}
+
+		if err = f(c); err != nil {
+			errs.Add(err)
+			pool.RemoveClientFor(addr)
```
owen-d, line 245: Would lbe nice to only remove the connection from the pool in a certain error class, but I'm not sure what I'm looking for here.

MasslessParticle (approved):

owen-d:
Just tested this with success in one of our clusters 🎉

PR #8302, opened 2023-01-26
Promtail WAL support: Implement reader side

thepalbi:
This PR is the second in a series that will implement WAL support into Promtail. The main objective of this PR is implement the reader side of WAL. That is, using the previously instantiated `client.Manager` to somehow read entries from the WAL instead of receiving them from a channel.

**Pending work**:
- [x] Implement mechanism so that `WriteTo` targets of each `wal.Watcher` can cleanup un-used series
- [x] Write a full fledged test-case that tests a WAL enabled Promtail (maybe in a follow up)

Part of #8197 

The line count might be big since this PR implements two fully fledged test cases of Promtail with WAL enabled.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


clients/pkg/promtail/wal/watcher.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// WriteTo is responsible for doing the necessary work to process both series and entries while the Watcher
+// is reading / tailing segments. The implementor of this interface is not expected to implement thread safety if used
+// on a single watcher, since each callback is called synchronously.
+//
+// Based on https://github.com/prometheus/prometheus/blob/main/tsdb/wlog/watcher.go#L46
+type WriteTo interface {
+	StoreSeries(series record.RefSeries)
+	Append(entries RefEntries) error
```
cstyan, line 42: Can we name this `AppendEntry` just so if in the future we implement any other types we don't end up with the same `Append`, `AppendExemplar`, etc. I think it's nicer if they're all named with the type.

clients/pkg/promtail/wal/watcher.go
```diff
@@ -0,0 +1,287 @@
...
+}
```
cstyan, line 43: this interface will likely still need the cleanup functions from prometheus' `WriteTo` interface

clients/pkg/promtail/client/manager.go
```diff
@@ -34,15 +39,24 @@ func NewManager(metrics *Metrics, logger log.Logger, maxStreams, maxLineSize int
...
 	if len(clientCfgs) == 0 {
 		return nil, fmt.Errorf("at least one client config should be provided")
 	}
 	clientsCheck := make(map[string]struct{})
+
 	clients := make([]Client, 0, len(clientCfgs))
+	watchers := make([]Stoppable, 0, len(clientCfgs))
+	// Instead of each write to that connects a watcher to a client keeping it's own mapping of decoded series, using
+	// a shared one
+	seriesCache := newSeriesCache()
```
cstyan, line 58: we have to be careful about this, the reason prometheus queue managers each have a watcher and cache is so that one remote write falling behind doesn't mean we can't clean up series records from the cache

the obvious downside to that is that we duplicate the caching for each queue manager (client in this case)

if we can come up with a nice _eventual_ clean up mechanism then we should be fine

clients/pkg/promtail/wal/watcher.go
thepalbi: Sounds good!

clients/pkg/promtail/client/manager.go
thepalbi: Yep, agree. Switching for to using a series cache per writeTo. We can used a shared later once the cleanup mechanism is more clear.

clients/pkg/promtail/wal/watcher.go
thepalbi: Adding for cleanup just `SeriesReset(int)` for now. Since alongside with the segment deletion mechanism the WAL writer uses, we can use something similar that the watcher somehow notifies that a segment that was seen before is now gone, so the WriteTo can clean up unused series

thepalbi:
@cstyan this PR should be ready to review. First, I implemented what we last discussed and you've seen a bit previously, which is the Watcher + client orchestration that goes into the `client.Manager`. 

Then, to keep the implementation sound, I implemented a series cache cleanup mechanism based on the watching over segments that are deleted in the WAL. This is fully decoupled, the `wal.Writer` deletes old segments and it's own pace, and the `wal.Watcher` checks for newly deleted segments every so often, emitting the necessary `SeriesReset` to the target `wal.WriteTo`.

I tried to make everything as self explanatory as possible with comments. Let me know anything that's not clear 😄 .

clients/pkg/promtail/client/client_writeto.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/go-kit/log/level"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/tsdb/record"
+
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+	"github.com/grafana/loki/pkg/ingester/wal"
+	"github.com/grafana/loki/pkg/util"
+)
+
+type seriesCacheEntry struct {
```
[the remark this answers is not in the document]
cstyan: actually, the reason we have the labels cache and `lastSeen` map separate in prometheus is to avoid the lock contention possibility

I would tend to think the memory overhead of the 2nd map is worth it to split the two and avoid lock contention; this is especially true since at the moment we write a series record every time we see a log line since we're not tracking the series we already know about when writing to the WAL. Happy to have you prove otherwise if you would like to keep this struct and a single map. I think we could write benchmark that has a few go routines all writing lots of log entries and see which is able to write faster.

clients/pkg/promtail/client/manager.go
```diff
@@ -64,12 +82,8 @@ func (m *Manager) start() {
...
 	go func() {
 		defer m.wg.Done()
-		// keep reading received entries
-		for e := range m.entries {
-			// then fanout to every remote write client
-			for _, c := range m.clients {
-				c.Chan() <- e
-			}
+		// discard read entries
+		for range m.entries {
```
cstyan, line 86: what's the reason for throwing them away?

cstyan (commented):
some comments, but I haven't gotten to the wal subdir yet

clients/pkg/promtail/client/client_writeto.go
```diff
@@ -0,0 +1,90 @@
...
+			delete(c.series, ref)
+		}
+	}
+}
+
+func (c *clientWriteTo) AppendEntries(entries wal.RefEntries) error {
+	var entry api.Entry
+	c.cacheLock.RLock()
+	l, ok := c.series[uint64(entries.Ref)]
+	c.cacheLock.RUnlock()
```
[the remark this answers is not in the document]
thepalbi: I think they are, if you check the wal Entries record contains:
```go
type RefEntries struct {
	Counter int64
	Ref     chunks.HeadSeriesRef
	Entries []logproto.Entry
}
```
and each `logproto.Entry` just contains a timestamp and a line, meaning that the series referenced by that head apply to all log lines inside the entries record

clients/pkg/promtail/client/manager.go
thepalbi: Until we refactor all types of clients inside the manager, `client.Manager` is responsible just for the WAL enabled use case. Since in this PR we are still in the latter case (without refactoring), and we are implementing the wal reader side which would read all entries from the WAL and deliver them to each remote write client, we don't need the manager to also write them to the client.

That bit of code was in place for making the wal with just the writer side implemented work.

clients/pkg/promtail/client/client_writeto.go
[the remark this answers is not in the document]
thepalbi: I read over the queue manager implementation, and I like the idea of keeping entries across two maps, one with the labelsets, and once with the last seen segment. Regardless, I think it won't have any effect at this point of the implementation, since the main contention point would occurs when: 
- the watcher is running and appending entries/storing series
- a checkpoint is detected, and the [`UpdateSeriesSegment`](https://github.com/prometheus/prometheus/blob/8f3b2fc9aeaf1bd3f839e820e7f9ffc571006e89/storage/remote/queue_manager.go#L844) in the `WriteTo` is called, avoiding a possible contention if both were a single entry

And the Promtail WAL doesn't use checkpoints for the time being. Since the overhead is almost none, It'd be awesome to keep it so in the future we avoid possible contentions.


```
goos: darwin
goarch: arm64
pkg: github.com/grafana/loki/clients/pkg/promtail/client
                                                                     │ pr_single_map.txt │           pr_dual_map.txt            │
                                                                     │      sec/op       │   sec/op     vs base                 │
ClientWriteTo/num_writers_5,_with_1000_lines_written_in_total-10             1.433 ±  8%   1.362 ± 14%       ~ (p=0.710 n=10)
ClientWriteTo/num_writers_10,_with_1000_lines_written_in_total-10            1.427 ± 12%   1.485 ±  4%       ~ (p=0.201 n=10)
ClientWriteTo/num_writers_5,_with_10000_lines_written_in_total-10            1.970 ± 14%   2.007 ± 20%       ~ (p=0.503 n=10)
ClientWriteTo/num_writers_10,_with_10000_lines_written_in_total-10           1.762 ±  5%   1.722 ±  3%       ~ (p=0.541 n=10+8)
ClientWriteTo/num_writers_5,_with_100000_lines_written_in_total-10           6.628 ±  5%   6.892 ±  2%  +3.97% (p=0.037 n=10+8)
ClientWriteTo/num_writers_10,_with_100000_lines_written_in_total-10          4.033 ±  5%   4.000 ±  3%  -0.82% (p=0.037 n=10+8)
ClientWriteTo/num_writers_10,_with_1000000_lines_written_in_total-10         27.12 ±  0%   27.03 ±  1%       ~ (p=0.201 n=10)
geomean                                                                      3.390         3.396        +0.16%

                                                                     │ pr_single_map.txt │            pr_dual_map.txt             │
                                                                     │       B/op        │     B/op      vs base                  │
ClientWriteTo/num_writers_5,_with_1000_lines_written_in_total-10            35.39Ki ± 6%   35.30Ki ± 3%        ~ (p=0.503 n=10)
ClientWriteTo/num_writers_10,_with_1000_lines_written_in_total-10           43.96Ki ± 4%   48.91Ki ± 5%  +11.27% (p=0.000 n=10)
ClientWriteTo/num_writers_5,_with_10000_lines_written_in_total-10           312.3Ki ± 0%   312.8Ki ± 1%        ~ (p=0.056 n=10)
ClientWriteTo/num_writers_10,_with_10000_lines_written_in_total-10          316.2Ki ± 1%   317.1Ki ± 0%        ~ (p=0.090 n=10+8)
ClientWriteTo/num_writers_5,_with_100000_lines_written_in_total-10          3.055Mi ± 0%   3.055Mi ± 0%        ~ (p=0.794 n=10+8)
ClientWriteTo/num_writers_10,_with_100000_lines_written_in_total-10         3.053Mi ± 0%   3.051Mi ± 0%        ~ (p=0.334 n=10+8)
ClientWriteTo/num_writers_10,_with_1000000_lines_written_in_total-10        30.54Mi ± 0%   30.54Mi ± 0%   -0.01% (p=0.010 n=10)
geomean                                                                     646.1Ki        656.2Ki        +1.56%

                                                                     │ pr_single_map.txt │           pr_dual_map.txt            │
                                                                     │     allocs/op     │  allocs/op   vs base                 │
ClientWriteTo/num_writers_5,_with_1000_lines_written_in_total-10             1.169k ± 1%   1.163k ± 1%       ~ (p=0.909 n=10)
ClientWriteTo/num_writers_10,_with_1000_lines_written_in_total-10            1.236k ± 0%   1.248k ± 1%  +0.97% (p=0.000 n=10)
ClientWriteTo/num_writers_5,_with_10000_lines_written_in_total-10            18.88k ± 0%   18.88k ± 0%       ~ (p=0.392 n=10)
ClientWriteTo/num_writers_10,_with_10000_lines_written_in_total-10           17.69k ± 0%   17.68k ± 0%       ~ (p=0.405 n=10+8)
ClientWriteTo/num_writers_5,_with_100000_lines_written_in_total-10           198.9k ± 0%   198.9k ± 0%       ~ (p=0.636 n=10+8)
ClientWriteTo/num_writers_10,_with_100000_lines_written_in_total-10          197.7k ± 0%   197.7k ± 0%       ~ (p=0.117 n=10+8)
ClientWriteTo/num_writers_10,_with_1000000_lines_written_in_total-10         1.998M ± 0%   1.998M ± 0%  -0.00% (p=0.010 n=10)
geomean                                                                      32.45k        32.47k       +0.06%
```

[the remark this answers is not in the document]
cstyan: You're right, I was mis-remembering. The watcher is single threaded and can't concurrently append samples to `writeTo`and also call `seriesReset`.

clients/pkg/promtail/client/client_writeto.go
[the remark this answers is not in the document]
cstyan: Right again! My bad.

clients/pkg/promtail/client/manager.go
cstyan: we should maybe just remove the entries channel then? and then start the watcher here + stop it in the `manager.Stop` function?

cstyan: oh right, we need to keep this since the writer fanout writer has to take in things that satisfy the entry handler interface

cstyan (commented):
This is looking good, but I'm a bit hesitant about the diffset code. The comments are great, I understand exactly what's happening and it's unlikely we'd need to modify that code much in the future. But it feels like with that amount of comments we could potentially do something simpler? Cleanup will happen relatively infrequently so it being slightly less efficient isn't a problem (until it is, and until it is a simpler implementation that's easier to follow is maybe the way we should go)?

What do you think about either passing the `writeTo` interface to the writer code and letting it call `seriesReset` directly, or creating a new `notifier` interface that we use to do the same thing or get a handle to a channel for passing a single int value (the highest segment # deleted) to the watcher (and then on to the `seriesReset` function)?

clients/pkg/promtail/client/manager.go
thepalbi: Exactly! On the refactor PR (which will be the following), I can take another pass at this and see if there's some way to simplify it

clients/pkg/promtail/client/client_writeto.go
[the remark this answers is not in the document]
thepalbi: > The watcher is single threaded
Yes! But `AppendSeries` and `SeriesReset` can be called concurrently since one is called by the watcher main thread, when appending series, and the other by the "whomever" is polling for deleted segments: this can be the current impl, which is another thread that lives in the watcher, or the notifier you mention.
Regardless, both of those concurrent call can't be optimized that much since they lock the main cache map.

thepalbi:
> This is looking good, but I'm a bit hesitant about the diffset code. The comments are great, I understand exactly what's happening and it's unlikely we'd need to modify that code much in the future. But it feels like with that amount of comments we could potentially do something simpler? Cleanup will happen relatively infrequently so it being slightly less efficient isn't a problem (until it is, and until it is a simpler implementation that's easier to follow is maybe the way we should go)?
> 
> What do you think about either passing the `writeTo` interface to the writer code and letting it call `seriesReset` directly, or creating a new `notifier` interface that we use to do the same thing or get a handle to a channel for passing a single int value (the highest segment # deleted) to the watcher (and then on to the `seriesReset` function)?

I agree is not the simplest code, though I like the fact that the writer side and the reader side are completely decoupled. The main driver of doing this implementation, is that both have no coupling, and the passing of information is either through the WAL (writing entries), or side-channeling through the WAL (deleting a segment, and the other one noticing that).

> What do you think about either passing the `writeTo` interface to the writer code and letting it call `seriesReset` directly

To keep that decoupling I'd like to avoid this. But...

> creating a new `notifier` interface that we use to do the same thing or get a handle to a channel for passing a single int value (the highest segment # deleted) to the watcher (and then on to the `seriesReset` function)?

This one looks interesting. I'll take a stab later today to see what ideas I come up with!

thepalbi:
@cstyan , implemented the Notifier approach you mention. I really like how it looks now, you are right it's much simpler than the diffset implementation. 

For some reasons the e2e promtail tests I've added with the wal enabled are failing just because in `promtail.go` I'm instantiating the `wal.Writer` before the `client.Manager`. After I fix that the PR should be good.

clients/pkg/promtail/wal/writer.go
```diff
@@ -27,6 +27,14 @@ const (
...
 )
 
+// WriterEventSubscriber is an interface that objects that want to receive events from the wal Writer can implement. After
+// they can subscribe to events by adding themselves as subscribers on the Writer with writer.Subscribe.
+type WriterEventSubscriber interface {
+	// NotifySegmentsCleaned will notify the subscriber that all WAL segments with a number lower or equal to num, were
+	// reclaimed.
+	NotifySegmentsCleaned(num int)
+}
+
```
cstyan, line 37: couldn't this just be the `WriteTo` interface we already have? we could just call `subscribers.SeriesReset`

clients/pkg/promtail/utils/remotewrite_server.go
```diff
@@ -0,0 +1,41 @@
...
+
+// RemoteWriteRequest wraps the received logs remote write request that is received.
+type RemoteWriteRequest struct {
+	TenantID string
+	Request  logproto.PushRequest
+}
+
+// NewRemoteWriteServer creates and starts a new httpserver.Server that can handle remote write request. When a request is handled,
+// the received entries are written to receivedChan, and status is responded.
+func NewRemoteWriteServer(receivedChan chan RemoteWriteRequest, status int) *httptest.Server {
```
cstyan, line 20: if we're not using this anywhere else currently can we move it into a test file and name it more obviously that it's only used in tests?

clients/pkg/promtail/client/client_writeto.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (c *clientWriteTo) StoreSeries(series []record.RefSeries, segment int) {
+	c.seriesLock.Lock()
+	defer c.seriesLock.Unlock()
+	c.seriesSegmentLock.Lock()
+	defer c.seriesSegmentLock.Unlock()
+	for _, seriesRec := range series {
+		c.seriesSegment[seriesRec.Ref] = segment
+		// maybe we can avoid doing this below if already present
```
cstyan, line 53: I think we always want to store the labels just in case for some reason a ref ID has new labels. It shouldn't ever happen in regular usage, but we don't have control over what the rest of the code assigns as a ref ID for a stream.

clients/pkg/promtail/client/client_writeto.go
```diff
@@ -0,0 +1,89 @@
...
+		labels := util.MapToModelLabelSet(seriesRec.Labels.Map())
+		c.series[seriesRec.Ref] = labels
+	}
+}
+
+// SeriesReset will delete all cache entries that were last seen in segments numbered equal or lower than segmentNum
```
cstyan, line 59: I think this comment is currently incorrect, looks to me like we're just doing <

clients/pkg/promtail/wal/watcher.go
```diff
@@ -0,0 +1,305 @@
...
+	"github.com/grafana/loki/pkg/ingester/wal"
+)
+
+const (
+	readPeriod         = 10 * time.Millisecond
+	segmentCheckPeriod = 100 * time.Millisecond
+
+	// debug flag used for developing and testing the watcher code. Using this instead of level.Debug to avoid checking
+	// the log level inwards into the logger code, and just making the compiler omit this code.
+	debug = false
```
cstyan, line 26: not sure I understand the comment as to why we'd want to do this over `level.Debug`

cstyan (commented):
minor comments

cstyan (commented):
I think really the only thing we're missing is getting around reintroducing the same 10-100ms polling interval that eats a lot of CPU in the Agent when it's reading the WAL to grab metrics.

We can leave that for a follow up PR as this ones already getting pretty big, but we can use https://github.com/grafana/loki/pull/8874 and https://github.com/prometheus/prometheus/pull/11950, or alternatively https://github.com/prometheus/prometheus/pull/11949 which would mean introducing a new notification type on the writer/subscriber interface you have here

clients/pkg/promtail/client/client_writeto.go
thepalbi: Well thought! Deleting comment

clients/pkg/promtail/client/client_writeto.go
thepalbi: Changing the if condition to `<=` since when `SeriesReset(n)` is called, it means segment n has been deleted, and because all segments keep an order in which they are written, every other before that one should have been reclaimed as well.

clients/pkg/promtail/wal/watcher.go
cstyan: Good point. Thanks for the example of existing usage of the same pattern. Let keep this :+1:

clients/pkg/promtail/wal/watcher.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+// WriteTo is responsible for doing the necessary work to process both series and entries while the Watcher
+// is reading / tailing segments. Note that StoreSeries and SeriesReset might be called concurrently.
+//
+// Based on https://github.com/prometheus/prometheus/blob/main/tsdb/wlog/watcher.go#L46
+type WriteTo interface {
+	WriteCleanup
+
+	// StoreSeries is called when series are found in WAL entries by the watcher, alongside with the segmentNum they were
```
tpaschalis, line 55: Just one nit: have we considered going with the Promtail terminology (and call them streams maybe?) instead of keeping close to what Prometheus does?

cstyan (approved):

PR #8307, opened 2023-01-27
Ruler: add rule evaluation jitter

dannykopping:
Now that we have [rule-based sharding](https://github.com/grafana/loki/pull/8092/), all rules on a ruler will execute concurrently. Generally this should not be a problem, but with a large number of rules with the same interval (the default, for example) this could cause some contention and scheduler latency.

N/A

It would be better to apply the jitter to each rule and not evaluate a new random jitter with each evaluation, but we don't have access to the rules in the manager (we reuse Prometheus' rule engine).

I considered keeping a map to memoise jitters based on the query string, but this seemed inelegant.
I also considered using the query string, the only part of the rule definition we have access to in this context, hashing that and then computing a jitter so it's stable. This seems overly complex for what we need, given that we need to somehow compute a value from 0 to 1 from a `uint32` FNV hash to multiply by the defined maximum jitter to get a reasonable distribution of jitters. We'd also still have to this on every rule evaluation, or memoise the result...

In the end, I just went with the simplest option.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/ruler/compat.go
+++ b/pkg/ruler/compat.go
@@ -4,11 +4,13 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"math/rand"
 	"os"
 	"strings"
 	"time"
 
 	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
@@ -57,13 +59,15 @@ type RulesLimits interface {
 
 // engineQueryFunc returns a new query function using the rules.EngineQueryFunc function
 // and passing an altered timestamp.
-func engineQueryFunc(engine *logql.Engine, overrides RulesLimits, checker readyChecker, userID string) rules.QueryFunc {
+func engineQueryFunc(cfg Config, engine *logql.Engine, overrides RulesLimits, checker readyChecker, userID string, logger log.Logger, rng *rand.Rand) rules.QueryFunc {
 	return rules.QueryFunc(func(ctx context.Context, qs string, t time.Time) (promql.Vector, error) {
 		// check if storage instance is ready; if not, fail the rule evaluation;
 		// we do this to prevent an attempt to append new samples before the WAL appender is ready
 		if !checker.isReady(userID) {
 			return nil, errNotReady
 		}
+
+		applyJitter(rng, cfg.EvaluationJitter, log.With(logger, "query_hash", logql.HashQueryString(qs)))
 
 		adjusted := t.Add(-overrides.EvaluationDelay(userID))
 		params := logql.NewLiteralParams(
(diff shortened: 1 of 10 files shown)


pkg/ruler/compat.go
```diff
@@ -96,6 +100,19 @@ func engineQueryFunc(engine *logql.Engine, overrides RulesLimits, checker readyC
 	})
 }
 
+// applyJitter sleeps up to a maximum of EvaluationJitter seconds to spread out evaluations to prevent thundering herd
+func applyJitter(rng *rand.Rand, jitter time.Duration, logger log.Logger) float64 {
```
salvacorts, line 104: Nits on naming:
- The comment refers to the `jitter` argument as `EvaluationJitter`.
- I'd rename the `jitter` argument to `maxJitter`.
- I'd rename the `j` variable to `jitter`.

pkg/ruler/compat_test.go
```diff
@@ -304,13 +306,25 @@ func TestNonMetricQuery(t *testing.T) {
...
+	queryFunc := engineQueryFunc(Config{}, engine, overrides, fakeChecker{}, "fake", log.NewNopLogger(), rand.New(rand.NewSource(time.Now().UnixNano())))
 
 	_, err = queryFunc(context.TODO(), `{job="nginx"}`, time.Now())
 	require.Error(t, err, "rule result is not a vector or scalar")
 }
 
+func TestRuleEvalJitter(t *testing.T) {
+	const jitter = 2 * time.Second
+
+	rng := rand.New(rand.NewSource(1234))
```
salvacorts, line 319: 👍

salvacorts (approved):
LGTM. Just left a minor nit comment.

chaudum (approved):

JStickler (approved):
[Docs squad] LGTM

pkg/ruler/compat.go
```diff
@@ -57,14 +59,16 @@ type RulesLimits interface {
...
-func engineQueryFunc(engine *logql.Engine, overrides RulesLimits, checker readyChecker, userID string) rules.QueryFunc {
+func engineQueryFunc(cfg Config, engine *logql.Engine, overrides RulesLimits, checker readyChecker, userID string, logger log.Logger, rng *rand.Rand) rules.QueryFunc {
 	return rules.QueryFunc(func(ctx context.Context, qs string, t time.Time) (promql.Vector, error) {
 		// check if storage instance is ready; if not, fail the rule evaluation;
 		// we do this to prevent an attempt to append new samples before the WAL appender is ready
 		if !checker.isReady(userID) {
 			return nil, errNotReady
 		}
 
+		applyJitter(rng, cfg.EvaluationJitter, log.With(logger, "query_hash", logql.HashQueryString(qs)))
```
jeschkies, line 70: Is this called once for a query out each time it's executed? I'm asking because we'd just need some delay at the start.

dannykopping: I covered this in "Special notes for your reviewer"

dannykopping: It would, but only slightly; it's a tradeoff because executing all the rules concurrently would lead to the same thing but with a lot more contention.

https://github.com/grafana/loki/pull/8129 addresses this, and will not really require jitter unless at large scales.

JStickler (approved):
[Docs squad]  LGTM

dannykopping:
Closing in favour of https://github.com/grafana/loki/pull/8848

PR #8308, opened 2023-01-27
operator: Cleanup ruler resources when disabled

aminesnow:
This PR adds a the removal of ruler resources when it has been disabled.

[LOG-3507](https://issues.redhat.com/browse/LOG-3507)

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/operator/internal/handlers/lokistack_create_or_update.go
+++ b/operator/internal/handlers/lokistack_create_or_update.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/ViaQ/logerr/v2/kverrors"
 	"github.com/go-logr/logr"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -416,3 +430,49 @@ func isNamespaceScoped(obj client.Object) bool {
...
+
+	if err := c.Delete(ctx, &ruler, &client.DeleteOptions{}); err != nil {
+		return kverrors.Wrap(err, "failed to delete statefulset",
+			"name", ruler.Name,
+			"namespace", ruler.Namespace,
+		)
+	}
+
+	return nil
+}
```
periklis, line 478: To keep some sanity in this file, we usually put helpers like these into the `handlers/internal` package. In addition this ensures that nobody is going to use them anywhere else accidentaly. The effect is that we can keep our manifests package static w/o apiserver calls during code review time.

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -250,6 +251,19 @@ func CreateOrUpdateLokiStack(
...
+		if err != nil {
+			ll.Error(err, "failed to remove rules configmap")
+			return err
+		}
+
+		err = removeRuler(ctx, req, k)
+		if err != nil {
+			ll.Error(err, "failed to remove ruler statefulset")
+			return err
+		}
```
periklis, line 266: In addition to the comment below, we can simplify this down to one call remove both or nothing. WDYT?

periklis (changes requested):

operator/internal/handlers/lokistack_create_or_update.go
aminesnow: I'd rather keep them separate. Smaller functions with a clear purpose I find is better for many reasons

periklis (approved):

PR #8315, opened 2023-01-27
Re-license and export `pkg/ingester` WAL code to be used in Promtail's WAL

thepalbi:

In PRs mentioned in #8197 , we are adding WAL support for Promtail. Since under-the-cover structure of the WAL will be much like Loki's, we need to re-use some bits from Loki's code:
- `WALRecord` and all of it's encode/decode methods exported (https://github.com/grafana/loki/blob/main/pkg/ingester/encoding.go#L35)
- `resettingPool`, also exporter (https://github.com/grafana/loki/blob/855d2b74bc075341044130303d879a3a23fa2f64/pkg/ingester/wal.go#L202)

This two bits have the default Loki license applies (AGPL), which is not compatible with Promtail's , ALv2. This PR moves them to a new package, and re-licenses it.

Also, some other bits are needed which due to its package `pkg/util` , fall under ALv2 so we are ok with them:
- `NewWALReader` (https://github.com/grafana/loki/blob/855d2b74bc075341044130303d879a3a23fa2f64/pkg/util/wal/reader.go#L11)
- These encoding bits, under `pkg/util/encoding` (https://github.com/grafana/loki/blob/855d2b74bc075341044130303d879a3a23fa2f64/pkg/util/encoding/encoding.go)

They are also shown in [this](https://github.com/grafana/loki/pull/8267/commits/0bf0132f7305b1e5d53c8a8f738df0b1c3a99a96#diff-531ca41b33686449180902c11253d32e3ae56091680ee898eb9f79f3d06e4ad3) commit in of the related PRs.

Also, since the new package is called wal, the main `WALRecord` struct has been renamed to `wal.Record` to avoid repetition in the name itself.

Required for #8197 

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/clients/pkg/promtail/targets/gcplog/pull_target.go
+++ b/clients/pkg/promtail/targets/gcplog/pull_target.go
@@ -1,17 +1,18 @@
 package gcplog
 
 import (
-	"cloud.google.com/go/pubsub"
 	"context"
+	"io"
+	"sync"
+	"time"
+
+	"cloud.google.com/go/pubsub"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/grafana/dskit/backoff"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/relabel"
 	"google.golang.org/api/option"
-	"io"
-	"sync"
-	"time"
 
 	"github.com/grafana/loki/clients/pkg/promtail/api"
 	"github.com/grafana/loki/clients/pkg/promtail/scrapeconfig"
--- a/clients/pkg/promtail/targets/gcplog/pull_target.go
+++ b/clients/pkg/promtail/targets/gcplog/pull_target.go
@@ -1,18 +1,17 @@
 package gcplog
 
 import (
-	"context"
-	"io"
-	"sync"
-	"time"
-
 	"cloud.google.com/go/pubsub"
+	"context"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/grafana/dskit/backoff"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/relabel"
 	"google.golang.org/api/option"
+	"io"
+	"sync"
+	"time"
 
 	"github.com/grafana/loki/clients/pkg/promtail/api"
 	"github.com/grafana/loki/clients/pkg/promtail/scrapeconfig"
--- a/pkg/ingester/checkpoint.go
+++ b/pkg/ingester/checkpoint.go
@@ -4,12 +4,13 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"github.com/grafana/loki/pkg/ingester/wal"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strconv"
 	"time"
+
+	"github.com/grafana/loki/pkg/ingester/wal"
 
 	"github.com/dustin/go-humanize"
 	"github.com/go-kit/log"
--- a/pkg/ingester/checkpoint.go
+++ b/pkg/ingester/checkpoint.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/grafana/loki/pkg/ingester/wal"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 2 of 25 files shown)


thepalbi:
I'd like to, as part of this PR, [rename `WALRecord` to `Record`](https://github.com/grafana/loki/pull/8315/files#diff-1f70440a13bd620fe23bf8e7d0e2f181b435a5baa9d3655249d237f22daaf2fcR40), since due to the new package name, it's imported reference will be `wal.Record`

thepalbi:
> I'd like to, as part of this PR, [rename `WALRecord` to `Record`](https://github.com/grafana/loki/pull/8315/files#diff-1f70440a13bd620fe23bf8e7d0e2f181b435a5baa9d3655249d237f22daaf2fcR40), since due to the new package name, it's imported reference will be `wal.Record`. Also the linter doesn't seem to like having `wal.WALRecord`, but we can omit it if not.
> 
> ```
> pkg/ingester/wal/encoding.go:40:6: exported: type name will be used as wal.WALRecord by other packages, and that stutters; consider calling this Record (revive)
> type WALRecord struct {
>      ^
> ```
> 
> What do you think @cstyan ?

Agreed on yes

clients/pkg/promtail/targets/gcplog/pull_target.go
```diff
@@ -1,17 +1,18 @@
 package gcplog
 
 import (
-	"cloud.google.com/go/pubsub"
 	"context"
+	"io"
+	"sync"
+	"time"
+
+	"cloud.google.com/go/pubsub"
```
cstyan, line 9: nit: in general can we avoid adding changes like this in a feature PR, can make it harder to review

if there's a lint failure that made this necessary that's fine, otherwise it's easier to review features and whitespace/cleanup changes separately

pkg/ingester/wal/encoding.go
```diff
@@ -184,7 +189,7 @@ func decodeEntries(b []byte, version RecordType, rec *WALRecord) error {
 	return nil
 }
 
-func decodeWALRecord(b []byte, walRec *WALRecord) (err error) {
+func DecodeWALRecord(b []byte, walRec *Record) (err error) {
```
cstyan, line 192: We can call this just `DecodeRecord`

pkg/ingester/wal/encoding.go
```diff
@@ -11,6 +11,11 @@ import (
 	"github.com/grafana/loki/pkg/util/encoding"
 )
 
+var (
+	// shared pool for WALRecords and []logproto.Entries
+	recordPool = NewRecordPool()
+)
+
```
cstyan, line 18: I think we don't need this package global here? it was just used in the actual ingester package, so we can create one with `NewRecordPool` there if we really need a global

clients/pkg/promtail/targets/gcplog/pull_target.go
thepalbi: Uhh my bad, I've been running the `Makefiles` linter to avoid CI back and forth, and it seems it fixes some import orders if it founds them. I'll revert those changes!

pkg/ingester/wal/encoding.go
thepalbi: Yhea, just found the dependency why I created that in that file:
- It was used in encoding_test, so created one as global there for the tests themselves to use
- [Used in `Record#Reset`](https://github.com/grafana/loki/blob/main/pkg/ingester/encoding.go#L57). Inverted the dependency, and moved that to the pool itself, since it was the only caller of reset

Also, created one as global in `pkg/ingester/wal.go` since it's used a bunch there

pkg/ingester/encoding_test.go
```diff
@@ -5,205 +5,15 @@ import (
 	"testing"
 	"time"
 
+	"github.com/grafana/loki/pkg/ingester/wal"
```
cstyan, line 8: nit, can we keep all the loki dependencies together at the bottom of the list

pkg/ingester/flush_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/grafana/loki/pkg/ingester/wal"
```
cstyan, line 12: same here

pkg/ingester/checkpoint.go
```diff
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/grafana/loki/pkg/ingester/wal"
```
cstyan, line 13: here too

cstyan (commented):
lets just clean up the imports ordering for consistency's sake

after that it looks like you'll need to merge in main or rebase but after that we're good to go IMO

pkg/ingester/encoding_test.go
thepalbi: Sure thing, resolving all

thepalbi:
@cstyan all should be fixed now 😀

thepalbi:
@cstyan this PR should be good to go right?

cstyan (approved):

PR #8316, opened 2023-01-27
Update index metrics

owen-d:
* Renames a metric that was poorly named and suggested it was only available for the boltdb-shipper. This is breaking, but I didn't find any references in our included mixin/etc. 
* Increases histogram bucket ranges to allow measuring beyond 10s p99s for index queries.


DylanGuedes (approved):
Since this is a "breaking" change maybe you'll want to add a changelog entry for it. Anyway, LGTM!

slim-bean (approved):

PR #8319, opened 2023-01-27
fix dashboard helper to use regex

trevorwhitney:

We use a regex value for this container: https://github.com/trevorwhitney/loki/blob/751fd7984dcb585d125a63097c84d2fc56eb2c47/production/loki-mixin/dashboards/loki-writes-resources.libsonnet#L19


MichelHollands (approved):
LGTM

PR #8325, opened 2023-01-30
Use organization ISSUE_COMMANDS_TOKEN with reduced scope

jdbaldry:
The GH_BOT_ACCESS_TOKEN has permissions beyond those required for addToProject and other API commands that the issue_commands workflow uses.

The new token is set at an organization level so it does not require repository administrators to rotate the token. It also has the minimal classic PAT permissions to facilitate the workflow.

It has expiry but that expiry is reported via email to the engineering organization and the IT Helpdesk have permissions to regenerate the token when expiration is imminent.

### DIFF
--- a/.github/workflows/issue_commands.yml
+++ b/.github/workflows/issue_commands.yml
@@ -16,6 +16,6 @@ jobs:
         run: npm install --production --prefix ./actions
       - name: Run Commands
         uses: ./actions/commands
-        with:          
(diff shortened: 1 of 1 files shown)


dannykopping (approved):
LGTM, thanks!

jdbaldry:
Thanks for the review @dannykopping. Would you mind merging for me too? I don't have authorization.

jdbaldry:
Thank you!

PR #8329, opened 2023-01-30
Sizing Tool: Set the minimum replias to 3.

### LINKED ISSUE #8296: Sizing Tool: Set minimum replicas to 3 for HA
In order to be highly available Loki should be running with at least three replicas on three nodes. This should be the minimum of for the generate Helm chart.

jeschkies:
The sizing tool should set the minimum replicas to three so that high availability is enabled.

Fixes #8296

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


jeschkies:
@trevorwhitney do we need to update the number of nodes as well?

MichelHollands (approved):
LGTM

trevorwhitney:
@jeschkies I think we do need to update the nodes here. It looks like we calculate write nodes based on pods/replicas, and then read nodes based on write nodes. This is assuming a constant ratio between read/write nodes, but that ratio will change in basic vs super query configurations.

PR #8337, opened 2023-01-30
Loki: only log "executing query" once per query in the frontend

slim-bean:

This is a continuation of #7469

I had to add a way to disable logging of the "executing query" line from the Engine in the query frontend because this is called for every split by time and shard created by our roundtrippers leading to many "executing query" lines for a single query which is not desired.

Added the "executing query" line to the roundtripper and it now covers: range, instant, series, and labels queries.

It should now also only log one line per query making it possible to correlate this to the "metrics.go" line for a finished query to help debug queries which don't finish because of OOM crash.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/engine.go
+++ b/pkg/logql/engine.go
@@ -119,7 +119,7 @@ type EngineOpts struct {
 	// only used for instant log queries.
 	MaxLookBackPeriod time.Duration `yaml:"max_look_back_period"`
 
-	// LogExecutingQuery will control if we log the query name when Exec is called.
+	// LogExecutingQuery will control if we log the query when Exec is called.
 	LogExecutingQuery bool `yaml:"-"`
 }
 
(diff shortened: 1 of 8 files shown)


pkg/logql/engine.go
```diff
@@ -118,12 +118,17 @@ type EngineOpts struct {
 	// MaxLookBackPeriod is the maximum amount of time to look back for log lines.
 	// only used for instant log queries.
 	MaxLookBackPeriod time.Duration `yaml:"max_look_back_period"`
+
+	// LogExecutingQuery will control if we log the query name when Exec is called.
```
dannykopping, line 122: It's not just the name, right?

```suggestion
	// LogExecutingQuery will control if we log the query when Exec is called.
```

dannykopping (approved):
LGTM! One tiny nit

PR #8339, opened 2023-01-30
Only report test coverage when it changes

MasslessParticle:
Our test-coverage report is noisy and reports for every single commit even when coverage doesn't change or the files changed are in different packages. This PR sums the coverage diff and only posts to the PR if it is non zero.


dannykopping (approved):
❤️

PR #8340, opened 2023-01-30
Ensure a delete request interval of at least a second

MasslessParticle:
This PR enforces that a the duration of a delete request covers at least 1 second.

Delete requests are designed with second-precision in mind. Without this fix, users are able to submit requests where `startTime == endTime`. In addition to logically ambiguous, it causes a divide-by-zero panic.

### DIFF
--- a/pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
+++ b/pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
@@ -121,11 +121,7 @@ func shardDeleteRequestsByInterval(startTime, endTime model.Time, query, userID 
 func (dm *DeleteRequestHandler) interval(params url.Values, startTime, endTime model.Time) (time.Duration, error) {
 	qr := params.Get("max_interval")
 	if qr == "" {
-		if dm.maxInterval == 0 {
-			return endTime.Sub(startTime), nil
-		}
-
-		return min(endTime.Sub(startTime), dm.maxInterval), nil
+		return dm.intervalFromStartAndEnd(startTime, endTime)
 	}
 
 	interval, err := time.ParseDuration(qr)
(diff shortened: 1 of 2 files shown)


pkg/storage/stores/indexshipper/compactor/deletion/request_handler.go
```diff
@@ -144,6 +140,18 @@ func (dm *DeleteRequestHandler) interval(params url.Values, startTime, endTime m
 	return interval, nil
 }
 
+func (dm *DeleteRequestHandler) intervalFromStartAndEnd(startTime, endTime model.Time) (time.Duration, error) {
+	interval := endTime.Sub(startTime)
+	if interval < time.Second {
+		return 0, errors.New("difference between start time and end time must be at lest one second")
```
DylanGuedes, line 146:
```suggestion
		return 0, errors.New("difference between start time and end time must be at least one second")
```

pkg/storage/stores/indexshipper/compactor/deletion/request_handler_test.go
```diff
@@ -170,6 +170,7 @@ func TestAddDeleteRequestHandler(t *testing.T) {
 			{"org-id", `{foo="bar"}`, "0000000000", "0000000001", "1ms", "invalid max_interval: valid time units are 's', 'm', 'h'\n"},
 			{"org-id", `{foo="bar"}`, "0000000000", "0000000001", "1h", "max_interval can't be greater than 1m0s\n"},
 			{"org-id", `{foo="bar"}`, "0000000000", "0000000001", "30s", "max_interval can't be greater than the interval to be deleted (1s)\n"},
+			{"org-id", `{foo="bar"}`, "0000000000", "0000000000", "", "difference between start time and end time must be at lest one second\n"},
```
DylanGuedes, line 173:
```suggestion
			{"org-id", `{foo="bar"}`, "0000000000", "0000000000", "", "difference between start time and end time must be at least one second\n"},
```

DylanGuedes (approved):
lgtm, just 2 small nits

owen-d (approved):

PR #8343, opened 2023-01-30
Loki: Add a limit for the [range] value on range queries

### LINKED ISSUE #8746: Support limiting step or interval size in queries
A user query with a large [step or interval](https://grafana.com/docs/loki/latest/api/) can slow done a cluster. One should be able to limit the impact of such a query.

There are already `MaxQueryLookback` and `MaxQueryLength` tenant limits. It feels reasonable to add `MaxQueryStep` and `MaxQueryInterval` tenant limits to them.

slim-bean:

Loki does not currently split queries by time to a value smaller than what's in the [range] of a range query.

Example

```
sum(rate({job="foo"}[2d]))
```

Imagine now this query being executed over a longer window of a few days with a step of something like 30m.

Every step evaluation would query the last [2d] of data.

There are use cases where this is desired, specifically if you force the step to match the value in the range, however what is more common is someone accidentally uses `[$__range]` in here instead of `[$__interval]` within Grafana and then sets the query time selector to a large value like 7 days.

This PR adds a limit which will fail queries that set the [range] value higher than the configured limit.

It's disabled by default.

In the future it may be possible for Loki to perform splits within the [range] and remove the need for this limit, but until then this can be an important safeguard in clusters with a lot of data.

Fixes #8746

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


MasslessParticle (approved):

pkg/logql/engine_test.go
```diff
@@ -2410,6 +2410,36 @@ func TestEngine_MaxSeries(t *testing.T) {
...
 
+func TestEngine_MaxRangeInterval(t *testing.T) {
+	eng := NewEngine(EngineOpts{}, getLocalQuerier(100000), &fakeLimits{rangeLimit: 24 * time.Hour}, log.NewNopLogger())
+
+	for _, test := range []struct {
+		qs             string
+		direction      logproto.Direction
+		expectLimitErr bool
+	}{
+		{`topk(1,rate(({app=~"foo|bar"})[2d]))`, logproto.FORWARD, true},
```
vlad-diachenko, line 2421: I think we could also add one positive scenario and one additional negative scenario with range deeper , just to confirm that it checks all ranges in the query ..
```suggestion
		{`topk(1,rate(({app=~"foo|bar"})[2d]))`, logproto.FORWARD, true},
		{`topk(1,rate(({app=~"foo|bar"})[1d]))`, logproto.FORWARD, false},
		{`topk(1,rate({app=~"foo|bar"}[12h]) / (rate({app="baz"}[23h]) + rate({app="fiz"}[25h])))`, logproto.FORWARD, true},
```

pkg/logql/engine_test.go
```diff
@@ -2410,6 +2410,36 @@ func TestEngine_MaxSeries(t *testing.T) {
 	}
 }
 
+func TestEngine_MaxRangeInterval(t *testing.T) {
+	eng := NewEngine(EngineOpts{}, getLocalQuerier(100000), &fakeLimits{rangeLimit: 24 * time.Hour}, log.NewNopLogger())
```
vlad-diachenko, line 2414: for the second test case we need to adjust `maxSeries` limit to not receive an error that we reached the maximum of series ;) 
```suggestion
	eng := NewEngine(EngineOpts{}, getLocalQuerier(100000), &fakeLimits{rangeLimit: 24 * time.Hour, maxSeries: 100000}, log.NewNopLogger())
```

pkg/logql/engine_test.go
```diff
@@ -2410,6 +2410,36 @@ func TestEngine_MaxSeries(t *testing.T) {
...
+				direction: test.direction,
+				limit:     1000,
+			})
+			_, err := q.Exec(user.InjectOrgID(context.Background(), "fake"))
+			if test.expectLimitErr {
+				require.NotNil(t, err)
+				require.True(t, errors.Is(err, logqlmodel.ErrRangeLimit))
+			} else {
+				require.Nil(t, err)
+			}
```
vlad-diachenko, line 2438: These asserts provide better error messages in case the test is failed

```suggestion
			if test.expectLimitErr {
				require.Error(t, err)
				require.ErrorIs(t, err, logqlmodel.ErrRangeLimit)
			} else {
				require.NoError(t, err)
			}
```

pkg/logql/engine.go
```diff
@@ -402,6 +411,23 @@ func (q *query) evalSample(ctx context.Context, expr syntax.SampleExpr) (promql_
...
+		switch e.(type) {
+		case *syntax.RangeAggregationExpr:
+			l := e.(*syntax.RangeAggregationExpr).Left
+			if l != nil {
+				if l.Interval > limit {
+					err = fmt.Errorf("%w: [%s] > [%s]", logqlmodel.ErrRangeLimit, l.Interval, limit)
+				}
+			}
+			return
+		}
```
vlad-diachenko, line 426: could be simplified with :
```suggestion
		switch e := e.(type) {
		case *syntax.RangeAggregationExpr:
			if e.Left == nil || e.Left.Interval <= limit {
				return
			}
			err = fmt.Errorf("%w: [%s] > [%s]", logqlmodel.ErrRangeLimit, e.Left.Interval, limit)
		}
```

pkg/logql/limits.go
```diff
@@ -14,6 +14,7 @@ var (
 // Limits allow the engine to fetch limits for a given users.
 type Limits interface {
 	MaxQuerySeries(userID string) int
+	MaxQueryRange(userID string) time.Duration
```
vlad-diachenko, line 17: let's call it `MaxQueryInterval`.
otherwise, we add the same confusion as we have on Grafana UI . range <-> interval

pkg/logqlmodel/error.go
```diff
@@ -13,6 +13,7 @@ var (
 	ErrParse          = errors.New("failed to parse the log query")
 	ErrPipeline       = errors.New("failed execute pipeline")
 	ErrLimit          = errors.New("limit reached while evaluating the query")
+	ErrRangeLimit     = errors.New("[range] value exceeds limit")
```
vlad-diachenko, line 16:
```suggestion
	ErrIntervalLimit     = errors.New("[interval] value exceeds limit")
```

pkg/logql/engine.go
```diff
@@ -317,21 +317,30 @@ func (q *query) evalSample(ctx context.Context, expr syntax.SampleExpr) (promql_
...
-	expr, err := optimizeSampleExpr(expr)
+	tenantIDs, err := tenant.TenantIDs(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	stepEvaluator, err := q.evaluator.StepEvaluator(ctx, q.evaluator, expr, q.params)
+	lim := validation.SmallestPositiveNonZeroDurationPerTenant(tenantIDs, q.limits.MaxQueryRange)
+	if lim != 0 {
+		err = q.checkLimits(expr, lim)
```
vlad-diachenko, line 327: I would call this function `checkIntervalLimit`

pkg/logql/engine.go
```diff
@@ -317,21 +317,30 @@ func (q *query) evalSample(ctx context.Context, expr syntax.SampleExpr) (promql_
...
 	}
 
-	expr, err := optimizeSampleExpr(expr)
+	tenantIDs, err := tenant.TenantIDs(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	stepEvaluator, err := q.evaluator.StepEvaluator(ctx, q.evaluator, expr, q.params)
+	lim := validation.SmallestPositiveNonZeroDurationPerTenant(tenantIDs, q.limits.MaxQueryRange)
```
vlad-diachenko, line 325:
```suggestion
	maxQueryInterval := validation.SmallestPositiveNonZeroDurationPerTenant(tenantIDs, q.limits.maxQueryInterval)
```

vlad-diachenko (approved):
Looks great 🚀  just added some comments that could help us to improve the implementation a bit.

owen-d (approved):
Agree with all @vlad-diachenko 's comments 👍 then lgtm

jeschkies (changes requested):
This is fixing https://github.com/grafana/loki/issues/8746 right?'

Could you add support to `pkg/util/querylimits/limiter.go` so that this limit can be set per request?

JStickler (approved):
[docs team] LGTM

pkg/logql/limits.go
vlad-diachenko: Ok, I was wrong )))) this limit applies to `[range_value]` not to the whole query range .... then let's leave it as is ;)

slim-bean:
Thanks for the input and help everyone!

PR #8367, opened 2023-02-01
Bump new loki/promtail versions in ksonnet libs

chaudum:

Updating the ksonnet libs as part of the post-release tasks.


MichelHollands (approved):
LGTM

PR #8370, opened 2023-02-01
[RFR] fix vector() function produces wrong timestamp on instant query

### LINKED ISSUE #8240: vector() function produces wrong timestamp on remote write
**Describe the bug**
If `vector()` function is used to produce a default value for extracted metrics from logs, it's being sent to prometheus using `remote write` with **bad** timestamp. Related to https://github.com/grafana/loki/issues/6946

The major issue it brings, is that Prometheus fully stops accepting any new metrics after this event. The only way to resume Prometheus operations is to stop and completely clear TSDB.

**To Reproduce**
1. Started Loki 2.7.1 with the following config snippet for `ruler`
```
ruler:
  storage:
    type: local
    local:
      directory: /rules
  wal:
    dir: /loki/rules-wal
  rule_path: /tmp/rules
  enable_api: true
  remote_write:
    enabled: true
    client:
      url: http://prometheus.my.internal/api/v1/write

```
2. Started Promtail 2.7.1
3. Started Prometheus with remote write option enabled.
4. Rule: 
    ```logql
    groups: 
    - name: Traefik
      interval: 1m
     rules:
     - record: traefic_count
       expr: |
         sum(count_over_time({namespace="traefik"}[5m]))
         or
         vector(0)
    ```
**Expected behavior**
Exported metrics to prometheus endpoint using `remote write` should use correct timestamp.

**Environment:**
 - Infrastructure: bare-metal with docker
 - Deployment tool: ansible

Screenshot 2023-01-23 at 12 37 08

liguozhong:
fix vector() function produces wrong timestamp on instant query
Fixes #8240

**error snapshot**
error timestamp in grafana 7.x
image
error timestamp in grafana 9.x
image

**with this PR**
fix logql:
` vector(5)
`image

with this PR
fix logql:
```
sum(count_over_time({namespace="traefik"}[5m]))
     or
     vector(5)
```
image

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/engine.go
+++ b/pkg/logql/engine.go
@@ -431,7 +431,7 @@ func (q *query) evalVector(_ context.Context, expr *syntax.VectorExpr) (promql_p
 
 	if GetRangeType(q.params) == InstantType {
 		return promql.Vector{promql.Sample{
-			Point:  promql.Point{T: s.T, V: value},
+			Point:  promql.Point{T: q.params.Start().UnixNano() / int64(time.Millisecond), V: value},
 			Metric: labels.Labels{},
 		}}, nil
 	}
--- a/pkg/logql/engine.go
+++ b/pkg/logql/engine.go
@@ -430,7 +430,10 @@ func (q *query) evalVector(_ context.Context, expr *syntax.VectorExpr) (promql_p
 	}
 
 	if GetRangeType(q.params) == InstantType {
-		return s, nil
+		return promql.Vector{promql.Sample{
+			Point:  promql.Point{T: s.T, V: value},
+			Metric: labels.Labels{},
+		}}, nil
 	}
 
 	return PopulateMatrixFromScalar(s, q.params), nil
--- a/pkg/logql/engine.go
+++ b/pkg/logql/engine.go
@@ -431,7 +431,7 @@ func (q *query) evalVector(_ context.Context, expr *syntax.VectorExpr) (promql_p
 
 	if GetRangeType(q.params) == InstantType {
 		return promql.Vector{promql.Sample{
-			Point:  promql.Point{T: q.params.Start().UnixNano() / int64(time.Millisecond), V: value},
+			Point:  promql.Point{T: q.params.Start().UnixMilli(), V: value},
 			Metric: labels.Labels{},
 		}}, nil
 	}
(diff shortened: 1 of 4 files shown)


pkg/logql/engine.go
```diff
@@ -426,7 +426,10 @@ func (q *query) evalVector(_ context.Context, expr *syntax.VectorExpr) (promql_p
 	}
 
 	if GetRangeType(q.params) == InstantType {
-		return s, nil
+		return promql.Vector{promql.Sample{
+			Point:  promql.Point{T: q.params.Start().UnixNano() / int64(time.Millisecond), V: value},
```
chaudum, line 430: Instead of `t.UnixNano()/int64(time.Millisecond)` we should use `t.UnixMilli()`

From the docs:

> // UnixMilli returns t as a Unix time, the number of milliseconds elapsed since
> // January 1, 1970 UTC. The result is undefined if the Unix time in
> // milliseconds cannot be represented by an int64 (a date more than 292 million
> // years before or after 1970). The result does not depend on the
> // location associated with t.

pkg/logql/evaluator.go
```diff
@@ -232,7 +232,7 @@ func (ev *DefaultEvaluator) StepEvaluator(
 		if err != nil {
 			return nil, err
 		}
-		return newVectorIterator(val, q.Step().Nanoseconds(), q.Start().UnixNano(), q.End().UnixNano()), nil
+		return newVectorIterator(val, q.Step().Milliseconds(), q.Start().UnixNano()/int64(time.Millisecond), q.End().UnixNano()/int64(time.Millisecond)), nil
```
chaudum, line 235: Same as above.

chaudum (commented):
Otherwise lgtm. Thanks for the fix!

pkg/logql/engine.go
liguozhong: done.thanks👍

pkg/logql/evaluator.go
liguozhong: ok

chaudum (approved):
LGMT. Maybe a second pair of eyes from someone with a bit more experience with the LogQL engine would be good.

liguozhong:
thanks！

dannykopping (approved):
LGTM!

PR #8373, opened 2023-02-01
Document memcached configuration for caching.

jeschkies:
Memcached is our default options for caching. However, the options for Memcached are not well documented.

This change documents how Memcached should be configured and how Loki itself should be documented based on our ksonnet deployment.

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/operations/caching.md
+++ b/docs/sources/operations/caching.md
@@ -39,7 +39,7 @@ To enable and configure Memcached:
 1. Configure Loki to use the cache.
     1. If the Helm chart is used
 
-       Set `memcached.chunk_cache.host` to the Memecache address of for the chunk cache, `memcached.results_cache.host` to the Memecache address of for the query result cache, `memcached.chunk_cache.enabled=true` and `memcached.results_cache.enabled=true`. 
+       Set `memcached.chunk_cache.host` to the Memecache address for the chunk cache, `memcached.results_cache.host` to the Memecache address for the query result cache, `memcached.chunk_cache.enabled=true` and `memcached.results_cache.enabled=true`. 
        
        Ensure that the connection limit of Memcached is at least `number_of_clients * max_idle_conns`.
        
--- a/docs/sources/operations/caching.md
+++ b/docs/sources/operations/caching.md
@@ -1,0 +1,94 @@
+---
+title: Caching 
+menuTitle: Caching 
+description: Enable and configure memcached for caching. 
+weight: 100
+keywords:
+  - memcached
+  - caching
+---
+
+# Caching
+
+Loki supports caching of index writes and lookups, chunks and query results to
+speed up query performance. This sections describes the recommended Memcached
+configuration to enable caching for chunks and query results. The index lookup
+cache is configured to be in-memory by default.
+
+## Before you begin
+
+- It is recommended to deploy four dedicated Memcached clusters.
+- As of 2023-02-01, the `memcached:1.6.17-alpine` version of the library is recommended.
+- Consult the Loki ksonnet [memcached](https://github.com/grafana/loki/blob/main/production/ksonnet/loki/memcached.libsonnet) deployment and the ksonnet [memcached library](https://github.com/grafana/jsonnet-libs/tree/master/memcached).
+
+## Steps
+
+To enable and configure Memcached:
+
+1. Deploy each Memcached service with at least three replicas and configure
+   each as follows:
+    1. Chunk cache 
+       ```
+       --memory-limit=4096 --max-item-size=2m --conn-limit=1024
+       ```
+    1. Query result cache
+       ```
+       --memory-limit=1024 --max-item-size=5m --conn-limit=1024
+       ```
+
+1. Configure Loki to use the cache.
+    1. If the Helm chart is used
+
+       Set `memcached.chunk_cache.host` to the Memecache address of for the chunk cache, `memcached.results_cache.host` to the Memecache address of for the query result cache, `memcached.chunk_cache.enabled=true` and `memcached.results_cache.enabled=true`. 
+       
+       Ensure that the connection limit of Memcached is at least `number_of_clients * max_idle_conns`.
+       
+       The options `host` and `service` depend on the type of installation. For example using the `bitname/memcached` Helm Charts with the following commands the `service` values is always `memcache`.
+       ```
+       helm upgrade --install chunk-cache -n loki bitnami/memcached -f memcached-overrides.yaml
+       helm upgrade --install write-cache -n loki bitnami/memcached -f memcached-overrides.yaml
+       helm upgrade --install results-cache -n loki bitnami/memcached -f memcached-overrides.yaml
+       helm upgrade --install index-cache -n loki bitnami/memcached -f memcached-overrides.yaml
+       ```
+       In this case, the Loki configuration would be
+       ```yaml
+       loki:
+         memcached:
+           chunk_cache:
+             enabled: true
+             host: chunk-cache-memcached.loki.svc
+             service: memcache
+             batch_size: 256
+             parallelism: 10
+           results_cache:
+             enabled: true
+             host: results-cache-memcached.loki.svc
+             service: memcache
+             default_validity: 12h
+       ```
+    1. If the Loki configuration is used
+        1. Configure the chunk cache
+           ```yaml
+           chunk_store_config:
+             chunk_cache_config:
+               memcached:
+                 batch_size: 256
+                 parallelism: 10
+               memcached_client:
+                 host: <memcached host>
+                 service: <port name of memcached service>
+           ```
+        1. Configure the query result cache
+           ```yaml
+           query_range:
+             cache_results: true
+             results_cache:
+               cache:
+                 memcached_client:
+                   consistent_hash: true
+                   host: <memcached host>
+                   service: <port name of memcached service>
+                   max_idle_conns: 16
+                   timeout: 500ms
+                   update_interval: 1m
+           ```
--- a/docs/sources/operations/caching.md
+++ b/docs/sources/operations/caching.md
@@ -43,7 +43,7 @@ To enable and configure Memcached:
        
        Ensure that the connection limit of Memcached is at least `number_of_clients * max_idle_conns`.
        
-       The options `host` and `service` depend on the type of installation. For example using the `bitname/memcached` Helm Charts with the following commands the `service` values is always `memcache`.
+       The options `host` and `service` depend on the type of installation. For example, using the `bitname/memcached` Helm Charts with the following commands, the `service` values are always `memcached`.
        ```
        helm upgrade --install chunk-cache -n loki bitnami/memcached -f memcached-overrides.yaml
        helm upgrade --install write-cache -n loki bitnami/memcached -f memcached-overrides.yaml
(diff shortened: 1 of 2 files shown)


docs/sources/operations/caching.md
```diff
@@ -0,0 +1,91 @@
...
+   each as follows:
+  1. Chunk cache `--memory-limit=4096 --max-item-size=2m --conn-limit=1024`
+  1. Index lookup cache `--memory-limit=1024 --max-item-size=5m --conn-limit=1024`
+  1. Index write cache `--memory-limit=1024 --max-item-size=1m --conn-limit=1024`
+  1. Query result cache `--memory-limit=1024 --max-item-size=5m --conn-limit=1024`
+
+1. Configure Loki to use the cache.
+  1. If the Helm chart is used
+    1. set `memcached.chunk_cache.host` to the Memecache address of for the chunk cache.
+    1. set `memcached.results_cache.host` to the Memecache address of for the query result cache.
```
jeschkies, line 38: @trevorwhitney should we add support for the index caches as well?

docs/sources/operations/caching.md
```diff
@@ -0,0 +1,91 @@
...
+    1. Ensure that the connection limit of Memcached is at least `number_of_clients? * max_idle_conns`.
+  1. If the Loki configuration is used
```
jeschkies, line 40: The following defaults are from our ksonnet deployment. Should we use these in Helm, @trevorwhitney?

docs/sources/operations/caching.md
```diff
@@ -0,0 +1,91 @@
...
+## Steps
+
+To enable and configure Memcached:
+
+1. Deploy each Memcached service with at least three replicas and configure
+   each as follows:
+  1. Chunk cache `--memory-limit=4096 --max-item-size=2m --conn-limit=1024`
+  1. Index lookup cache `--memory-limit=1024 --max-item-size=5m --conn-limit=1024`
+  1. Index write cache `--memory-limit=1024 --max-item-size=1m --conn-limit=1024`
+  1. Query result cache `--memory-limit=1024 --max-item-size=5m --conn-limit=1024`
```
jeschkies, line 33: @salvacorts I saw that you edited the Jsonnet files last. Is there anything we are missing?

jeschkies (commented):
I wonder if we should deploy memcached with our Helm chart since the ksonnet deployement does this as well.

docs/sources/operations/caching.md
jeschkies: Seems the fifocache should suffice.

docs/sources/operations/caching.md
jeschkies: Seems the fifocache should suffice.

docs/sources/operations/caching.md
```diff
@@ -0,0 +1,91 @@
...
+1. Deploy each Memcached service with at least three replicas and configure
+   each as follows:
+    1. Chunk cache `--memory-limit=4096 --max-item-size=2m --conn-limit=1024`
+    1. Index lookup cache `--memory-limit=1024 --max-item-size=5m --conn-limit=1024`
+    1. Index write cache `--memory-limit=1024 --max-item-size=1m --conn-limit=1024`
+    1. Query result cache `--memory-limit=1024 --max-item-size=5m --conn-limit=1024`
+
+1. Configure Loki to use the cache.
+    1. If the Helm chart is used
+        1. set `memcached.chunk_cache.host` to the Memecache address of for the chunk cache.
```
trevorwhitney, line 37: we also need to set `memcached.chunk_cache.enabled` and `memcached.results_cache.enabled` both to `true.

docs/sources/operations/caching.md
trevorwhitney: @jeschkies we already have these same defaults for the chunks and results cache. Currently the helm chart does not set up a `write_dedupe_cache`, so we would have to add that section completely including the defaults.

As for the index query cache we might need to revisit that. For `write` pods there's no need, as it [defaults to an in memory cache](https://github.com/grafana/loki/blob/9920a27f06b9a808bcd0d7323a55041bc343a8bc/pkg/loki/modules.go#L523). However the store is also use by `read` pods, and I'm not sure I understand the significance of the index query cache in query flows. Is in memory sufficient? We may need to add a section for that to the helm chart.

trevorwhitney (approved):
This looks great, but I would leave out the sections on the `write_dedupe_cache_config` and `index_queries_cache_config`. Neither of those are currently supported by the helm chart. We may want to add support, but update this doc if/when we do.

docs/sources/operations/caching.md
```diff
@@ -0,0 +1,69 @@
...
+      ```yaml
+      chunk_store_config:
+        chunk_cache_config:
+          memcached:
+            batch_size: 256
+            parallelism: 10
+          memcached_client:
+            host: <memcached host>
+            addresses: ???
+            service: ???
```
jeschkies, line 51: @trevorwhitney what should addresses and service be?

docs/sources/operations/caching.md
```diff
@@ -0,0 +1,65 @@
...
+
+Loki supports caching of index writes and lookups, chunks and query results to
+speed up query performance. This sections describes the recommended Memcached
+configuration to enable caching for chunks and query results. The index lookup
+cache is configured to be in-memory by default.
+
+## Before you begin
+
+- It is recommended to deploy four dedicated Memcached clusters.
+- As of 2023-02-01 the `memcached:1.6.17-alpine` is recommended.
```
JStickler, line 22:
```suggestion
- As of 2023-02-01, the `memcached:1.6.17-alpine` version of the library is recommended.
```

docs/sources/operations/caching.md
```diff
@@ -0,0 +1,65 @@
...
+    1. Chunk cache `--memory-limit=4096 --max-item-size=2m --conn-limit=1024`
+    1. Query result cache `--memory-limit=1024 --max-item-size=5m --conn-limit=1024`
+
+1. Configure Loki to use the cache.
+    1. If the Helm chart is used
+        1. set `memcached.chunk_cache.host` to the Memecache address of for the chunk cache.
+        1. set `memcached.results_cache.host` to the Memecache address of for the query result cache.
+	1. set `memcached.chunk_cache.enabled=true` and `memcached.results_cache.enabled=true`.
+        1. Ensure that the connection limit of Memcached is at least `number_of_clients? * max_idle_conns`.
+  1. If the Loki configuration is used
```
JStickler, line 40: I think something is off with the numbering formatting here, I think this should be nested at the same level as "If the Helm chart is used" under step 2?

jeschkies: Yes! Thank you!

docs/sources/operations/caching.md
jeschkies: So both, `host` and `service` must be specified?

docs/sources/operations/caching.md
```diff
@@ -0,0 +1,94 @@
...
+       ```
+    1. Query result cache
+       ```
+       --memory-limit=1024 --max-item-size=5m --conn-limit=1024
+       ```
+
+1. Configure Loki to use the cache.
+    1. If the Helm chart is used
+
+       Set `memcached.chunk_cache.host` to the Memecache address of for the chunk cache, `memcached.results_cache.host` to the Memecache address of for the query result cache, `memcached.chunk_cache.enabled=true` and `memcached.results_cache.enabled=true`.
```
JStickler, line 42:
```suggestion
       Set `memcached.chunk_cache.host` to the Memecache address for the chunk cache, `memcached.results_cache.host` to the Memecache address for the query result cache, `memcached.chunk_cache.enabled=true` and `memcached.results_cache.enabled=true`. 
```

docs/sources/operations/caching.md
```diff
@@ -0,0 +1,94 @@
...
+       ```
+
+1. Configure Loki to use the cache.
+    1. If the Helm chart is used
+
+       Set `memcached.chunk_cache.host` to the Memecache address of for the chunk cache, `memcached.results_cache.host` to the Memecache address of for the query result cache, `memcached.chunk_cache.enabled=true` and `memcached.results_cache.enabled=true`. 
+       
+       Ensure that the connection limit of Memcached is at least `number_of_clients * max_idle_conns`.
+       
+       The options `host` and `service` depend on the type of installation. For example using the `bitname/memcached` Helm Charts with the following commands the `service` values is always `memcache`.
```
JStickler, line 46:
```suggestion
       The options `host` and `service` depend on the type of installation. For example, using the `bitname/memcached` Helm Charts with the following commands, the `service` values are always `memcached`.
```

JStickler (commented):
Caught a couple more small things.

jeschkies:
@JStickler thanks. Could you take another look?

JStickler (approved):
[Doc squad] LGTM

docs/sources/operations/caching.md
trevorwhitney: I believe so. We use `memcached-client` in cloud so I'd stick with that as the default.

docs/sources/operations/caching.md
```diff
@@ -0,0 +1,94 @@
...
+1. Deploy each Memcached service with at least three replicas and configure
+   each as follows:
+    1. Chunk cache 
+       ```
+       --memory-limit=4096 --max-item-size=2m --conn-limit=1024
+       ```
+    1. Query result cache
+       ```
+       --memory-limit=1024 --max-item-size=5m --conn-limit=1024
+       ```
```
dannykopping, line 37: If we refer to 4 recommended memcached clusters here, we should specify the configs for those others too

dannykopping (commented):
Missed this while on PTO. I appreciate your putting together!
I picked up on a few small consistency issues which I commented on. I also think it'd be valuable to describe _why_ we recommend the settings we do, to give users more insight.

docs/sources/operations/caching.md
jseiser: > From the docs addresses is experimental, I would suggest using the host and service fields.
> 
> host: I have pointed this to the k8s service that is deployed for memcached service: Is the port name on that service exposing memcached
> 
> I have deployed 4 x memcached
> 
> ```
> helm upgrade --install chunk-cache -n loki bitnami/memcached -f memcached-overrides.yaml
> helm upgrade --install write-cache -n loki bitnami/memcached -f memcached-overrides.yaml
> helm upgrade --install results-cache -n loki bitnami/memcached -f memcached-overrides.yaml
> helm upgrade --install index-cache -n loki bitnami/memcached -f memcached-overrides.yaml
> ```
> 
> ```
> loki:
>   memcached:
>     chunk_cache:
>       enabled: true
>       host: chunk-cache-memcached.loki.svc
>       service: memcache
>       batch_size: 256
>       parallelism: 10
>     results_cache:
>       enabled: true
>       host: results-cache-memcached.loki.svc
>       service: memcache
>       default_validity: 12h
> 
>   structuredConfig:
>     chunk_store_config:
>       write_dedupe_cache_config:
>         memcached:
>           batch_size: 100
>           parallelism: 100
>         memcached_client:
>           host: write-cache-memcached.loki.svc
>           service: memcache
>           consistent_hash: true
> 
>     storage_config:
>       index_queries_cache_config:
>         memcached:
>           batch_size: 100
>           parallelism: 100
>         memcached_client:
>           host: index-cache-memcached.loki.svc
>           service: memcache
>           consistent_hash: true
> ```

Any chance you can  document this? `memcached-overrides.yaml` 

The bitnami chart being referenced can run standalone, or run in HA mode.  But HA mode would appear to require being able to pass in multiple hosts values to Loki.  So I assume you really want it to run in standalone mode, but then we only have 1 replica, not 3.

Any additional insight would be great.

PR #8374, opened 2023-02-01
Re-stablish tailing websocket connection if it's closed unexpectedly.

salvacorts:
When the querier handling a tailing connection shuts down (e.g. due to node eviction in K8S), the tailing request will fail. We can recover from this by re-establishing the connection to a new querier.

Fixes https://github.com/grafana/loki-private/issues/589

The linked issue has more context about this issue.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logcli/query/tail.go
+++ b/pkg/logcli/query/tail.go
@@ -34,8 +34,6 @@ func (q *Query) TailQuery(delayFor time.Duration, c client.Client, out output.Lo
 		os.Exit(0)
 	}()
 
-	tailResponse := new(loghttp.TailResponse)
-
 	if len(q.IgnoreLabelsKey) > 0 && !q.Quiet {
 		log.Println("Ignoring labels key:", color.RedString(strings.Join(q.IgnoreLabelsKey, ",")))
 	}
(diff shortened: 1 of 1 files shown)


pkg/logcli/query/tail.go
```diff
@@ -44,9 +42,32 @@ func (q *Query) TailQuery(delayFor time.Duration, c client.Client, out output.Lo
...
+				log.Printf("Remote websocket connection closed unexpectedly (%+v). Connecting again.", err)
+
+				// Close previous connection
+				if err = conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")); err != nil {
+					log.Fatalf("Error closing websocket: %+v", err)
+				}
+
+				conn, err = c.LiveTailQueryConn(q.QueryString, delayFor, q.Limit, lastReceivedTimestamp, q.Quiet)
+				if err != nil {
+					log.Fatalf("Error recreating tailing connection after unexpected close: %+v", err)
```
jeschkies, line 65: Shouldn't we break here?

jeschkies (commented):
I wonder if this could be baked into the gateway. Btw I think tail connects to the ingester.

salvacorts:
> I wonder if this could be baked into the gateway

That was my first idea, but I couldn't make it work. See:
https://github.com/grafana/loki-private/issues/589#issuecomment-1412199295

> Btw I think tail connects to the ingester.

Does it? As far as I can see, it connects to the querier. 
https://github.com/grafana/deployment_tools/blob/ecf37a0958aae4a47f6d650b0f591ce36928a019/ksonnet/lib/gateway/config.libsonnet#L130

jeschkies:
Checkout querier.Tail. The connection is forwarded to the ingested. But yes. It goes through the querier.

salvacorts:
I tested that this does not interfere with `querier.tail_max_duration `. It looks like it's working fine:
image

pkg/logcli/query/tail.go
salvacorts: Done at [24260d1](https://github.com/grafana/loki/pull/8374/commits/24260d11a4d70753326d9c833f1102efb8cd5bb8)

jeschkies (approved):

pkg/logcli/query/tail.go
```diff
@@ -44,9 +42,43 @@ func (q *Query) TailQuery(delayFor time.Duration, c client.Client, out output.Lo
...
+			if websocket.IsCloseError(err, websocket.CloseAbnormalClosure) {
+				log.Printf("Remote websocket connection closed unexpectedly (%+v). Connecting again.", err)
+
+				// Close previous connection. If it fails to close the connection it should be fine as it is already broken.
+				if err = conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")); err != nil {
+					log.Printf("Error closing websocket: %+v", err)
+				}
+
+				// Try to re-establish the connection up to 5 times.
+				for retry := 4; retry >= 0; retry-- {
```
kavirajk, line 64: We could use `github.com/grafana/dskit/backoff` here for the retry I think.

For example how to use it, recently Travis added it on promtail gcp target.
https://github.com/grafana/loki/pull/8281

salvacorts: That's super cool! Changed to use it with [fe8017e](https://github.com/grafana/loki/pull/8374/commits/fe8017ee8f9557e5f3fada48fd366ceb3fceec5f). Tested and works like a charm. Thanks for the tip 💯

slim-bean (approved):
LGTM!

salvacorts:
Merging even though Drone fails in the `check-doc` stage as this PR doesn't change anything regarding docs.

PR #8379, opened 2023-02-01
add section to prom operator instructions about necessary label

trevorwhitney:

Add a section to the helm monitoring docs about the need for ServiceMonitors to have the correct label to be discovered by the Prometheus Operator.

### DIFF
--- a/docs/sources/installation/helm/monitor-and-alert/index.md
+++ b/docs/sources/installation/helm/monitor-and-alert/index.md
@@ -86,7 +86,17 @@ In order to make sure the Prometheus Operator discovers the `ServiceMonitor` res
 monitoring:
   serviceMonitor:
     labels:
-      release: "prometheus"
+      release: prometheus
+```
+
+This is also true for the `PrometheusRule` resource deployed by the Helm chart, which in addition to a label, need to be in the same namespace as the Prometheus Operator. For example, if you installed the Prometheus Operator in the `monitoring` namespace, you would need to also add the following to the `values.yaml` for you Loki helm chart to ensure recroding rules are properly discoverd:
+
+```yaml
+monitoring:
+  rules:
+    namespace: monitoring
+    labels:
+      release: prometheus
 ```
 
 The `kube-prometheus-stack` installs `ServicMonitor` and `PrometheusRule` resources for monitoring Kubernetes, and it depends on the `kube-state-metrics` and `prometheus-node-exporter` helm charts which also install `ServiceMonitor` resources for collecting `kubelet` and `node-exporter` metrics. The above values file adds the necessary additional labels required for these metrics to work with the included dashboards.
--- a/docs/sources/installation/helm/monitor-and-alert/index.md
+++ b/docs/sources/installation/helm/monitor-and-alert/index.md
@@ -80,6 +80,15 @@ prometheus:
         targetLabel: cluster
 ```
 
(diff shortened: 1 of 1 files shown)


docs/sources/installation/helm/monitor-and-alert/index.md
```diff
@@ -80,6 +80,25 @@ prometheus:
         targetLabel: cluster
 ```
 
+In order to make sure the Prometheus Operator discovers the `ServiceMonitor` resources deployed by the `loki` chart, you will need to make sure those resources have the correct labels the Prometheus Operator is configured to look for. By default this is the key value label pair `release: prometheus`. Make sure the Loki `ServiceMonitor`s have this label by adding the following to the `values.yaml` for your Loki helm chart:
+
+```yaml
+monitoring:
+  serviceMonitor:
+    labels:
+      release: prometheus
```
slim-bean, line 89: unfortunately the helm-chart I used to install the prom operator used a different labelset

```
release: prom-stack
```

Maybe we can include instructions on how to figure out what label is needed?

trevorwhitney: @slim-bean which helm chart did you use? did you not use `kube-prometheus-stack`?

trevorwhitney: I think we should keep the instructions on `kube-prometheus-stack` as that install `kube-state-metrics` with the correct recodring rules, `node-exporter`, and the Prometheus Operator. One stop shop for everything someone needs to enable self-monitoring.

docs/sources/installation/helm/monitor-and-alert/index.md
```diff
@@ -80,6 +80,25 @@ prometheus:
         targetLabel: cluster
 ```
 
+In order to make sure the Prometheus Operator discovers the `ServiceMonitor` resources deployed by the `loki` chart, you will need to make sure those resources have the correct labels the Prometheus Operator is configured to look for. By default this is the key value label pair `release: prometheus`. Make sure the Loki `ServiceMonitor`s have this label by adding the following to the `values.yaml` for your Loki helm chart:
```
JStickler, line 83:
```suggestion
In order for the Prometheus Operator to discover the `ServiceMonitor` resources deployed by the `loki` chart, those resources must have the labels the Prometheus Operator is configured to expect. By default this is the key value label pair `release: prometheus`. Make sure the Loki `ServiceMonitor`s have this label by adding the following to the `values.yaml` for your Loki Helm chart:
```

docs/sources/installation/helm/monitor-and-alert/index.md
```diff
@@ -80,6 +80,25 @@ prometheus:
...
+
+```yaml
+monitoring:
+  serviceMonitor:
+    labels:
+      release: prometheus
+```
+
+This is also true for the `PrometheusRule` resource deployed by the Helm chart, which in addition to a label, need to be in the same namespace as the Prometheus Operator. For example, if you installed the Prometheus Operator in the `monitoring` namespace, you would need to also add the following to the `values.yaml` for you Loki helm chart to ensure recroding rules are properly discoverd:
```
JStickler, line 92:
```suggestion
This is also true for the `PrometheusRule` resource deployed by the Helm chart, which in addition to a label, need to be in the same namespace as the Prometheus Operator. For example, if you installed the Prometheus Operator in the `monitoring` namespace, you would need to add the following to the `values.yaml` for your Loki Helm chart to ensure recording rules are properly discovered:
```

JStickler (approved):
[Docs squad]  Two small suggestions.

PR #8381, opened 2023-02-02
operator: Enhancement Proposal for Zone Aware Replication

shwetaap:
Signed-off-by: Shweta Padubidri <spadubid@redhat.com>

Enhancement Proposal for Zone Aware Replication 

[LOG-3601](https://issues.redhat.com/browse/LOG-3601)

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


periklis (changes requested):
Reviewed only 1/2 of this.

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,270 @@
...
+- Enable Zone-Aware Replication configuration in LokiStack CR so the components are deployed in different zones.
+- Configure `loki-config.yaml` to support zone-aware data replication.
+
+## Motivation
+
+Zone-aware replication is the replication of data across failure domains. Avoiding data loss during a domain outage is the motivation to introduce zone-aware component deployment and data replication. 
+
+### Goals
+
+* The user can enable zone aware replication for the read and the write path
```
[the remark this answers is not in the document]
periklis: @cstyan Thanks for shedding more light on this. FWIW the Loki Operator deploys the Loki components in all-or-nothing fashion. Means it simply patches all Deployments/StatefulSets without considering if the belong to the write or the read path. So far we kept off our hands from doing this because it simplifies the design of the operator being a simple manifest create-or-update loop (plus some PKI certs generation).

@shwetaap The key point raised by @cstyan here needs explicit mention on the design proposal imho. For the sake of sanity and later documentation efforts we need to highlight that:
1. Either the operator updates the paths in two step, where waiting the max-query-lookback period (currently `30s` IIRC). 
2. Or the users should expect for the same period of time missing queries in the data.

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,426 @@
...
+- Configure `loki-config.yaml` to support zone-aware data replication.
+
+## Motivation
+
+Zone-aware replication is the replication of data across failure domains. Avoiding data loss during a domain outage is the motivation to introduce zone-aware component deployment and data replication. 
+
+### Goals
+
+* The user can enable zone aware replication in the loki operator
+* List options on how to make a distinct list of availability zones for Loki components in the LokiComponentSpec
```
periklis, line 41:
```suggestion
* The LokiStack administrator can enable zone-aware data replication for a managed Loki cluster.
* The LokiStack administrator can choose at least one topology label to enable the cluster spread across domains.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,426 @@
...
+
+* 
+
+## Proposal
+
+The following enhancement proposal describes the required API additions and changes in the Loki Operator to add zone-aware replication support
+
+### API Extensions
+
+#### LokiStack Changes: Support for configuring log retention
```
periklis, line 54:
```suggestion
#### LokiStack Changes: Support for configuring zone-aware data replication
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,426 @@
...
+
+//  LokiStackSpec defines the desired state of LokiStack
+type LokiStackSpec struct {
+...
+    // Replication defines the configuration for Loki data replication
+    //
+    // +optional
+    // +kubebuilder:validation:Optional
+    Replication *ReplicationSpec `json:"replication,omitempty"`
+}
```
periklis, line 70:
```suggestion
type LokiStackSpec struct {
...
   // ReplicationFactor defines the policy for log stream replication. (Deprecated: Please use replication.factor instead. This field will be remove in future versions of this CRD)
   //
   // +optional
   // +kubebuilder:validation:Optional
   // +kubebuilder:validation:Minimum:=1
   // +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="Replication Factor"
   ReplicationFactor int32 `json:"replicationFactor,omitempty"`
	
    // Replication defines the configuration for Loki data replication
    //
    // +optional
    // +kubebuilder:validation:Optional
    Replication *ReplicationSpec `json:"replication,omitempty"`
}
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,426 @@
...
+
+### Implementation Details/Notes/Constraints
+
+```yaml
+apiVersion: loki.grafana.com/v1beta1
+kind: LokiStack
+metadata:
+  name: lokistack-dev
+spec:
+  size: 1x.extra-small
```
periklis, line 132:
```suggestion
  size: 1x.medium.multi-zone
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,426 @@
...
+   "target": {
+     "kind": "Node",
+     "name": "aws-node-0"
+   }
+ }
+}
+```
+
+Inject the Nodes labels into the Pods through some annotations
+```yaml
```
periklis, line 234:
```suggestion
```go
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,426 @@
...
+  - maxSkew: 1
+    topologyKey: topology.kubernetes.io/zone
+    whenUnsatisfiable: DoNotSchedule
+    labelSelector:
+      matchLabels:
+        ingester: pod
+  containers:
+  - name: ingester
+    env:
+    # specify that the annotation that was added on the Pod by the operator due to the `eck.k8s.elastic.co/downward-node-labels` annotation feature should be automatically transformed to an env var
```
periklis, line 293:
```suggestion
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,426 @@
...
+      },
+      Distributor: &lokiv1.LokiComponentSpec{
+        Replicas: 3,
+      },
+      Ingester: &lokiv1.LokiComponentSpec{
+				Replicas: 6,
+			},
+			Querier: &lokiv1.LokiComponentSpec{
+				Replicas: 6,
+			},
```
periklis, line 391:
```suggestion
      Ingester: &lokiv1.LokiComponentSpec{
        Replicas: 6,
      },
      Querier: &lokiv1.LokiComponentSpec{
        Replicas: 6,
      },
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,433 @@
...
+
+  Switching a `1x.small` LokiStack to being zone-aware: setting `.zoneAware.enabled` in the LokiStack CR to `true` should be sufficient to restart the pods and reschedule them on different zones (given that the requirements are met).
+
+* `1x.medium` has a replication factor of 3, with the `ingester` and `querier` having 3 replicas, while the `distributor`, `query-frontend`, `gateway`, `index-gateway`, and `ruler` having only 2 replicas, which means 1 zone will have only 1 `querier` and 1 `ingester` that will communicate with components in other zones with higher delay. In this case, losing one zone cannot be tolerated, and relying on having a 4th zone isn't always possible. In this case it is preferable to suggest a new sizing configuration.
```
periklis, line 344:
```suggestion
* `1x.medium` has a replication factor of 3, with the `ingester` and `querier` having 3 replicas, while the `distributor`, `query-frontend`, `gateway`, `index-gateway`, and `ruler` having only 2 replicas. This means 1 zone will have only 1 `querier` and 1 `ingester` that will communicate with components (e.g. distributors) in other zones . In this case, losing one zone cannot be tolerated, and relying on having a 4th zone isn't always possible. In this case it is preferable to suggest a new sizing configuration.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,433 @@
...
+      Ruler: &lokiv1.LokiComponentSpec{
+        Replicas: 3,
+      },
+    },
+  }
+...
+}
+```
+
+
```
periklis, line 417:
```suggestion

```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+2. [Motivation](#motivation)
+3. [Proposal](#proposal)
+    1. [API Extensions](#api-extensions)
+    2. [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+4. [Design Details](#design-details)
+5. [Impact on LokiStack Sizing Configurations](#sizing)
+
+## Summary
+
+By default, data is transparently replicated across the whole pool of service instances, regardless of whether these instances are all running within the same availability zone (or data center, or rack) or in different ones. Storing multiple replicas for a given data within the same availability zone poses a risk for data loss if there is an outage affecting various nodes within a zone or a full zone outage. There is support for Zone Aware Replication in Loki. When enabled, replicas for the given data are guaranteed to span across different availability zones.
```
periklis, line 28:
```suggestion
By default, data is transparently replicated across the whole pool of service instances, regardless of whether these instances are all running within the same availability zone (or data center, or rack) or in different ones. Storing multiple replicas for a given data within the same availability zone poses a risk for data loss if there is an outage affecting various nodes within a zone or a full zone outage. There is support for zone-aware data replication in Loki. When enabled, replicas for the given data are guaranteed to span across different availability zones.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+A zone represents a logical failure domain. It is common for Kubernetes clusters to span multiple zones for increased availability. While the exact definition of a zone is left to infrastructure implementations, common properties of a zone include very low network latency within a zone, no-cost network traffic within a zone, and failure independence from other zones.
+The following sections describe a set of APIs in form of Custom Resource Definitions (CRD) that enable users of `LokiStack` resources to support:
+- Enable Zone-Aware Replication configuration in LokiStack CR so the components are deployed in different zones.
+- Configure `loki-config.yaml` to support zone-aware data replication.
```
periklis, line 32:
```suggestion
- Configure Loki to span data replication across multiple zones.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+
+## Motivation
+
+Zone-aware replication is the replication of data across failure domains. Avoiding data loss during a domain outage is the motivation to introduce zone-aware component deployment and data replication.
```
periklis, line 36:
```suggestion
Zone-aware replication is the replication of data across failure domains. Avoiding data loss during a domain outage is the motivation to introduce a zone-aware component deployment and enable Loki's zone-aware data replication capabilities.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+## Motivation
+
+Zone-aware replication is the replication of data across failure domains. Avoiding data loss during a domain outage is the motivation to introduce zone-aware component deployment and data replication. 
+
+### Goals
+
+* The LokiStack administrator can enable zone-aware data replication for a managed Loki cluster.
+* The LokiStack administrator can choose at least one topology label to enable the cluster spread across domains.
+
+
```
periklis, line 43:
```suggestion

```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+### Non-Goals
+
+*
```
periklis, line 46:
```suggestion
* Minimize cross-zone bandwith costs.
* Introduce data replication across Loki clusters in different regions.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+* The LokiStack administrator can choose at least one topology label to enable the cluster spread across domains.
+
+
+### Non-Goals
+
+* 
+
+## Proposal
+
+The following enhancement proposal describes the required API additions and changes in the Loki Operator to add zone-aware replication support
```
periklis, line 50:
```suggestion
The following enhancement proposal describes the required API additions and changes in the Loki Operator to add zone-aware data replication support.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+
+### API Extensions
+
+#### LokiStack Changes: Support for configuring zone-aware data replication
+
+
```
periklis, line 56:
```suggestion

The following API changes introduce a new spec to enable configuration for Loki's data replication properties.

_Note:_ The new `replicationSpec` introduces in addition a field `factor` that is a replacement for the old `replicationFactor` field. Moving forward the old field is officially deprecated any will be removed in future CRD versions.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+//  LokiStackSpec defines the desired state of LokiStack
+type LokiStackSpec struct {
+...
+   // ReplicationFactor defines the policy for log stream replication. (Deprecated: Please use replication.factor instead. This field will be remove in future versions of this CRD)
+   //
+   // +optional
+   // +kubebuilder:validation:Optional
+   // +kubebuilder:validation:Minimum:=1
+   // +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="Replication Factor"
+   ReplicationFactor int32 `json:"replicationFactor,omitempty"`
```
periklis, line 71:
```suggestion
    // ReplicationFactor defines the policy for log stream replication. (Deprecated: Please use replication.factor instead. This field will be removed in future versions of this CRD)
    //
    // +optional
    // +kubebuilder:validation:Optional
    // +kubebuilder:validation:Minimum:=1
    // +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="Replication Factor"
    ReplicationFactor int32 `json:"replicationFactor,omitempty"`
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+	
+    // Replication defines the configuration for Loki data replication
+    //
+    // +optional
+    // +kubebuilder:validation:Optional
+    Replication *ReplicationSpec `json:"replication,omitempty"`
+}
+
+type ReplicationSpec struct {
+
```
periklis, line 81:
```suggestion
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+    // Toplogy provides the information on zones in the cluster.
+    //
+    // +optional
+    // +kubebuilder:validation:Optional
+    Topology *TopologySpec
+}
+
+type TopologySpec struct {
+
```
periklis, line 97:
```suggestion
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+    // +required
+    // +kubebuilder:validation:Required
+    Topologykey string `json:"topologyKey,omitempty"`
+
+}
+
+```
+
+### Implementation Details/Notes/Constraints
+
```
periklis, line 133:
```suggestion

The following manifests represents a full example on how zone-aware data replication be turned on using the `topology.kubernetes.io/zone` node label as a key to spread pods across zones:

```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+}
+```
+Some reference to write a pods/binding webhook is here 
+* https://github.com/elastic/cloud-on-k8s/issues/3933#issuecomment-950581345
+* https://github.com/kubemod/kubemod/commit/95c43bc7d532db05051a83df5ad97e36a0f5d7e2#diff-d63889fb133b18cb690d365c25f43bf0580f7a7aaa91ffa03b121b9b6b5f2d60
+2. The second solution is to use the operator to modify the pods once they are scheduled. The operator checks if the `pod.spec.topologySpreadConstraints.topologyKey` is set. If yes, then the operator extracts the annotation(key and value) from the node, and sets it as an annotation for the pod by patching the pod. The zone value can then be accessed from within the container as an environment variable just like in solution 1 by adding the env variable in the container
+* https://github.com/elastic/cloud-on-k8s/blob/85aa75b7a7765330475c0b92a0851939ac3d9578/pkg/controller/elasticsearch/driver/node_labels.go#L57 
+* https://github.com/elastic/cloud-on-k8s/issues/3933#issuecomment-962897004
+
+This is how the expected individual pod spec will look like after the webhook has added the topology key annotation in the pod
```
periklis, line 280:
```suggestion
This is how the expected individual pod spec will look like after the webhook has added the topology key annotation in the pod:

```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+  - name: ingester
+    env:
+    - name: ZONE
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.annotations['topology.kubernetes.io/zone']
+```
+The container can then be started by passing the configuration via the cli flag `instance_availability_zone: $ZONE`
+
+Another variation of accessing the zone value in the container is by mounting the pod annotation as a downwardapi volume.
```
periklis, line 306:
```suggestion

The container can then be started by passing the configuration via the cli flag `instance_availability_zone: $ZONE`

```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+
+Another variation of accessing the zone value in the container is by mounting the pod annotation as a downwardapi volume.  
+
+Questions:
+* Both these methods first modify the pod to add the annotation[topology.kubernetes.io/zone: zone-a] in the pod and then modify the container to add a new env var or a volume which picks the zone value from the pod annotation via the downwardapi
+* In the Webhook method, is it safe to assume both the pod modification, followed by the container modification can be handled within the webhook?
+* In the operator method, there is going to be two merge patches of the object, one for the pod to add the annotation, and then the container to add the env variable.
+* Would it be better to handle the multiple modifications in an init container. If yes, Is that worth the overhead of  maintainence of an init container?
+* If the operator can handle this without init container isnt that better than a webhook?
+## Implementation History
```
periklis, line 314:
```suggestion

## Implementation History
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+    - name: ZONE
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.annotations['topology.kubernetes.io/zone']
+```
+The container can then be started by passing the configuration via the cli flag `instance_availability_zone: $ZONE`
+
+Another variation of accessing the zone value in the container is by mounting the pod annotation as a downwardapi volume.  
+
+Questions:
```
periklis, line 308:
```suggestion
### Open Questions

```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+
+  Switching a `1x.small` LokiStack to being zone-aware: setting `replication.topology.enabled` in the LokiStack CR to `true` should be sufficient to restart the pods and reschedule them on different zones (given that the requirements are met).
+
+* `1x.medium` has a replication factor of 3, with the `ingester` and `querier` having 3 replicas, while the `distributor`, `query-frontend`, `gateway`, `index-gateway`, and `ruler` having only 2 replicas. This means 1 zone will have only 1 `querier` and 1 `ingester` that will communicate with components (e.g. distributors) in other zones . In this case, losing one zone cannot be tolerated, and relying on having a 4th zone isn't always possible. In this case it is preferable to suggest a new sizing configuration.
```
periklis, line 340:
```suggestion
* `1x.medium` has a replication factor of 3, with the `ingester` and `querier` having 3 replicas, while the `distributor`, `query-frontend`, `gateway`, `index-gateway`, and `ruler` having only 2 replicas. This means 1 zone will have only 1 `querier` and 1 `ingester` that will communicate with components (e.g. distributors) in other zones . In this case, losing one zone cannot be tolerated, and relying on having a 4th zone isn't always possible. In this case it is preferable to suggest using a replication factor of 2 instead of the default set to 3.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+
+### `1x.medium.multizone` as a new flavour of `1x.medium`
+To increase availability and reduce cross-zone bandwidth, `2x.medium.multizone` ensures having 2 instances of ingesters and queriers in each zone if and only if zone-aware replication is enabled.
+
+Details like the CPU/memory/disk requests can vary from `1x.medium` since the load is distributed among more replicas. This is something that should be benchmarcked.
+
+
```
periklis, line 347:
```suggestion
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+## Open Questions
+* `lokiconfig.yaml` should be updated to include the `zone_awareness_enabled=true`. The `instance_availability_zone` will be different in the pods of different zones. 
+* Should we use the existing rollout-operator? since this operator is able to handle pods to be deployed automatically in different zones - https://github.com/grafana/rollout-operator
+* How can we minimize cross-zone bandwidth? (how to make queriers and ingesters zone-aware, so that each querier will only query ingesters in the same zone?)
+* Persistent volumes are bound to the zone they are created in. How do we re-bind the restarted pod to the PV after a zone failure?
+* How do we simulate zone failure in an OCP cluster?
+
+
+
```
periklis, line 356:
```suggestion

```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,362 @@
...
+## References
+1. https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/
+2. https://docs.openshift.com/container-platform/4.12/nodes/scheduling/nodes-scheduler-pod-topology-spread-constraints.html
+3. https://cortexmetrics.io/docs/guides/zone-aware-replication/
+
+
```
periklis, line 362:
```suggestion
1. [Kubernetes: Topology Spread Constraint Concept](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/)
2. [Kubernetes: Pod Topology Spread Constraints Docs ](https://docs.openshift.com/container-platform/4.12/nodes/scheduling/nodes-scheduler-pod-topology-spread-constraints.html)
3. [Cortex Metrics: Zone-Aware Data Replication Docs](https://cortexmetrics.io/docs/guides/zone-aware-replication/)
4. [elastic/cloud-on-k8s: Zone-Aware Replication Discussion Thread](https://github.com/elastic/cloud-on-k8s/issues/3933#issuecomment-950581345)
5. [KubeMod: Mutating Webhook Implementation](https://github.com/kubemod/kubemod/commit/95c43bc7d532db05051a83df5ad97e36a0f5d7e2#diff-d63889fb133b18cb690d365c25f43bf0580f7a7aaa91ffa03b121b9b6b5f2d60)
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,358 @@
...
+```
+
+The container can then be started by passing the configuration via the cli flag `instance_availability_zone: $ZONE`
+
+
+### Open Questions
+
+* Both these methods first modify the pod to add the annotation[topology.kubernetes.io/zone: zone-a] in the pod and then modify the container to add a new env var or a volume which picks the zone value from the pod annotation via the downwardapi
+* In the Webhook method, is it safe to assume both the pod modification, followed by the container modification can be handled within the webhook?
+* In the operator method, there is going to be two merge patches of the object, one for the pod to add the annotation, and then the container to add the env variable.
```
xperimental, line 315: I don't see why the configuration of the environment variable needs to be added by the webhook (does not matter which of the variants). I don't think there's validation on whether the value provided by the `fieldRef` is set, which should mean that the environment variable configuration can be part of the staticly generated `podSpec`.

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,358 @@
...
+  containers:
+  - name: ingester
+    env:
+    - name: ZONE
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.annotations['topology.kubernetes.io/zone']
+```
+
+The container can then be started by passing the configuration via the cli flag `instance_availability_zone: $ZONE`
```
xperimental, line 308: One thing I missed in the discussion of both solutions is that there's only one `instance_availability_zone` setting, even if the user decides to use more than one label for the topology.

In the end we need to reduce all the information to a single "topology key" that we pass into Loki. It might be as simple as concatenating the different label values in a defined way, don't know if Loki has a maximum size on the configuration value, though.

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,358 @@
...
+* Check if the Topology key provided by the user matches a valid node topology domain label
+* Verify that the replication factor defined is less than the number of available zones
+If both these conditions are satisfied, deploy the lokistack components using the `PodTopologySpreadConstraint` definition in the deployments/statefulsets Pod template spec.
+
+Another important consideration is that each of the loki component pod needs the specific zone name to be set in their configuration. There is no easy way to implement this since the Kubernetes Downward-API does not support exposing node labels (See issue https://github.com/kubernetes/kubernetes/issues/40610)
+
+
+A couple of solutions to implement passing the node labels to the Loki pods are listed below:
+ 
+1. The plan is to introduce a Admission Mutating Webhook, that would watch the `pods/binding` sub-resource for each Loki pods. This webhook can update the pod annotations to add the topology key-value pair(s) when it is being scheduled on a node. This annotation (having the zone information) can then be used in the container as an environment variable using the Downward API and expanded in the Loki configuration per pod.
```
xperimental, line 183: I'm not sure why this approach uses a MutatingWebhook at all, as it is not mutating the resource the webhook is watching (the `Binding`) but the `Pod`. It should be possible to do this in a normal "resource watcher" on the Binding resource instead.

That change would bring this solution closer to the one listed below, so maybe these two solutions are actually the same.

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,358 @@
...
+      zones:
+      - key: topology.kubernetes.io/zone
+        maxSkew: 1
+```
+	
+#### General constraints
+
+### Risks and Mitigations
+
+## Design Details
```
xperimental, line 159: I think I'd start this section with a summary of the issues we need to solve for the implementation:

- Spreading the pods across failure domains using the Kubernetes scheduler (using the `PodTopologySpreadConstraint`)
- Identifying the failure domain a pod is currently scheduled in (the two solutions discussed in length)
- Relaying the failure-domain information to the Loki configuration (using the `instance_availability_zone` configuration option & dynamically set environment variables)

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,358 @@
...
+
+* Configure the availability zone for each ingester via the `-ingester.availability-zone` CLI flag (or its respective YAML config option)
+* Rollout ingesters to apply the configured zone
+* Enable time-series zone-aware replication via the `-distributor.zone-awareness-enabled` CLI flag (or its respective YAML config option). Please be aware this configuration option should be set to `distributors`, `queriers` and `rulers`.
+https://github.com/grafana/rollout-operator
+
+<div id="sizing"/>
+
+## Impact on LokiStack Sizing Configurations
+Without zone-aware replication, the LokiStack pods are scheduled on different nodes within the same or different availability zones. There are no hard requirements on the number of zones within the cluster. In case a zone fails, the pods are simply rescheduled to other nodes based on the current policies (taints/tolerations/affinity).
```
xperimental, line 336: I don't think the last sentence is true. The scheduler will try to reschedule pods to a different zone, if available, but the issue is with the PersistentVolumes. Usually the PVs created on public clouds are also bound to a zone and can not be attached to a virtual machine (= Kubernetes Node) running in a different zone.

This will effectively mean that the ingestion will stop during a zone outage as the ingester pod can not be rescheduled to a different zone without also moving the PV, which is usually not possible. Kubernetes will not do anything in this case, the Pod will simply stay "Pending". Manual intervention _can_ remove the PVC, in which case it will be recreated by the operator and then possibly lead to a new PV created in a different zone and the Pod starting. But this is a manual operation and will also lead to loss of the data contained in the removed PV.
It might also be impossible to remove the PV while the zone is unavailable.

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,358 @@
...
+
+* `1x.medium` has a replication factor of 3, with the `ingester` and `querier` having 3 replicas, while the `distributor`, `query-frontend`, `gateway`, `index-gateway`, and `ruler` having only 2 replicas. This means 1 zone will have only 1 `querier` and 1 `ingester` that will communicate with components (e.g. distributors) in other zones . In this case, losing one zone cannot be tolerated, and relying on having a 4th zone isn't always possible. In this case it is preferable to suggest using a replication factor of 2 instead of the default set to 3.
+
+## Open Questions
+* `lokiconfig.yaml` should be updated to include the `zone_awareness_enabled=true`. The `instance_availability_zone` will be different in the pods of different zones. 
+* Should we use the existing rollout-operator? since this operator is able to handle pods to be deployed automatically in different zones - https://github.com/grafana/rollout-operator
+* How can we minimize cross-zone bandwidth? (how to make queriers and ingesters zone-aware, so that each querier will only query ingesters in the same zone?)
```
xperimental, line 350: This is listed as a non-goal, so I don't think it makes sense to have it in the questions.

xperimental (commented):
Did a first pass of the document. I'm still not convinced that the zone-aware replication has an effect on the t-shirt sizes, especially on the read-path. Need to think about this again...

There are a few small rendering issues for the document. Maybe it makes sense to feed it through something like `markdownlint`?

operator/docs/enhancements/zone_aware_replication.md
periklis: Not anymore at lease. I amended the non-goals just this morning.

operator/docs/enhancements/zone_aware_replication.md
periklis: This is a point to save for later but a good catch. I am not sure if the following applies only to AWS, but I hardly remember GCP zones include the name of the region in them:
```
NAME                           STATUS   ROLES                  AGE   VERSION           REGION      ZONE
ip-10-0-139-124.ec2.internal   Ready    worker                 80m   v1.25.4+18eadca   us-east-1   us-east-1a
ip-10-0-139-24.ec2.internal    Ready    control-plane,master   89m   v1.25.4+18eadca   us-east-1   us-east-1a
ip-10-0-146-149.ec2.internal   Ready    worker                 81m   v1.25.4+18eadca   us-east-1   us-east-1b
ip-10-0-149-216.ec2.internal   Ready    control-plane,master   89m   v1.25.4+18eadca   us-east-1   us-east-1b
ip-10-0-166-102.ec2.internal   Ready    control-plane,master   88m   v1.25.4+18eadca   us-east-1   us-east-1c
ip-10-0-170-152.ec2.internal   Ready    worker                 81m   v1.25.4+18eadca   us-east-1   us-east-1c
ip-10-0-186-239.ec2.internal   Ready    worker                 81m   v1.25.4+18eadca   us-east-1   us-east-1d
ip-10-0-197-101.ec2.internal   Ready    worker                 82m   v1.25.4+18eadca   us-east-1   us-east-1e
ip-10-0-211-56.ec2.internal    Ready    worker                 81m   v1.25.4+18eadca   us-east-1   us-east-1f
```

Thus if we allow the users to use `topology.kubernetes.io/region` with `topology.kubernetes.io/zone` we can skip the first if the above observation is matching across public clouds at least. Of course this does not apply on custom labels or `kubernetes.io/hostname` as topology keys.

operator/docs/enhancements/zone_aware_replication.md
periklis: I agree with you but I am still concerned if the whole MutatingWebhook in the admission control is about patching the pod before the kubelet ever starts the container. A simple resource watcher would probably come in to late and a pod restart would be in need, at least this is my mental model about how it works. I might be wrong.

operator/docs/enhancements/zone_aware_replication.md
periklis: After playing around with their tanka config, their PVC looks like this:

operator/docs/enhancements/zone_aware_replication.md
xperimental: GCP nomenclature is almost the same as AWS ([docs](https://cloud.google.com/compute/docs/regions-zones/)); regions are prefixes of zones.

I was not thinking of "public cloud users" when writing the comment but people in other environments who have different sets of labels that do not follow this prefix approach. One example I can think of is people labeling their nodes with "floor" and "rack" where rack is a number that can repeat for each of the floors. In that case the user might opt for using both for the topology which means for us we need to combine the values.

If I am honest I would put as little assumptions as possible into what the values of the different topology keys are and how they are related.

operator/docs/enhancements/zone_aware_replication.md
xperimental: I'm not sure what the additional level of indirection changes regarding the situation with the PVs. "Something" needs to step-in and manually remove the PVs, so that they are recreated. In our case this would be a human (current solution) or the loki-operator (possible future solution). In the case of Tanka above this would be the "rollout-operator" then? But what does this operator know about the internal operation of Tanka, so that it knows it is safe to remove the PVs / redistribute to another zone?

operator/docs/enhancements/zone_aware_replication.md
xperimental: Using the MutatingWebhook to change the pod definition "before it's written" is how I originally understood the "solution with MutatingWebhook", but the example provided does not seem to do that but just issues a "normal update". This is why I argued that _currently_ the two solutions are almost identical and can be reduced to one.

Either I am misunderstanding the code at hand or there's still room for "a real MutatingWebhook solution" that adds the annotation before the Pod configuration is written, so that no "empty startup" can happen.

operator/docs/enhancements/zone_aware_replication.md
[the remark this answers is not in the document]
shwetaap: Made a note of this here  - https://github.com/grafana/loki/blob/48ae6628e1508ceac3bfd897cc1ec52c66d651b6/operator/docs/enhancements/zone_aware_replication.md#1-read-and-write-path

operator/docs/enhancements/zone_aware_replication.md
shwetaap: Mentioned it here - https://github.com/grafana/loki/blob/48ae6628e1508ceac3bfd897cc1ec52c66d651b6/operator/docs/enhancements/zone_aware_replication.md#impact-on-lokistack-sizing-configurations

periklis: As mentioned in a call, we have two options here:
1. Treat this proposal simply as it says it will provide namely zone-aware data replication only. This means that we need some sort of playbook to execute failover activities during a zone outage.
2. Expand the operator capabilities to active fail-over from one zone to the other.

In summary we have consensus to err on the side of a simpler feature going forward with option no.1. The latter option demands to re-visit the capabilities of the Loki Operator managing StatefulSets (e.g. ingesters) via a custom controller to enable fail-over across zones. This was identified to be a prerequisite to the zone-aware data replication feature as described in the proposal.

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
+---
+title: Zone Awareness Replication Support
```
xperimental, line 2:
```suggestion
title: Zone-Aware Replication Support
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+
+The following enhancement proposal describes the required API additions and changes in the Loki Operator to add zone-aware data replication support.
+
+### API Extensions
+
+#### LokiStack Changes: Support for configuring zone-aware data replication
+
+The following API changes introduce a new spec to enable configuration for Loki's data replication properties.
+
+_Note:_ The new `replicationSpec` introduces in addition a field `factor` that is a replacement for the old `replicationFactor` field. Moving forward the old field is officially deprecated any will be removed in future CRD versions.
```
xperimental, line 61:
```suggestion
_Note:_ The new `replicationSpec` introduces a `factor` field that is a replacement for the old `replicationFactor` field. Moving forward the old field is officially deprecated any will be removed in future CRD versions.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+
+```go
+
+import "github.com/prometheus/prometheus/model/labels"
+
+
+//  LokiStackSpec defines the desired state of LokiStack
```
xperimental, line 68:
```suggestion
// LokiStackSpec defines the desired state of LokiStack
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+type LokiStackSpec struct {
+...
+    // ReplicationFactor defines the policy for log stream replication. (Deprecated: Please use replication.factor instead. This field will be removed in future versions of this CRD)
+    //
+    // +optional
+    // +kubebuilder:validation:Optional
+    // +kubebuilder:validation:Minimum:=1
```
xperimental, line 75: I think "optional" and "minimum=1" will clash, because there is no possibility to set "no value".

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+    // +kubebuilder:validation:Optional
+    Replication *ReplicationSpec `json:"replication,omitempty"`
+}
+
+type ReplicationSpec struct {
+    // Factor defines the policy for log stream replication.
+    //
+    // +optional
+    // +kubebuilder:validation:Optional
+    // +kubebuilder:validation:Minimum:=1
```
xperimental, line 91: I think "optional" and "minimum=1" will clash, because there is no possibility to set "no value".

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+    Zones []ZoneSpec
+}
+
+// ZoneSpec defines the spec to support zone aware component deployments.
+type ZoneSpec struct {
+    // MaxSkew describes the maximum degree to which Pods can be unevenly distributed
+    //
+    // +optional
+    // +kubebuilder:validation:optional
+    // +kubebuilder:default:="1"
```
xperimental, line 107: I don't think it makes sense to define both "optional" and "default" as the default will always be set.

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+kind: LokiStack
+metadata:
+  name: lokistack-dev
+spec:
+  size: 1x.small
+  storage:
+    secret:
+      name: test
+      type: s3
+  storageClassName: gp2
```
xperimental, line 134: Should we use a newer `storageClassName` for the example?

```suggestion
  storageClassName: gp3-csi
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+
+#### General constraints
+
+### Risks and Mitigations
+
+## Design Details
+
+### 1. Spreading the pods across failure domains using the Kubernetes scheduler (using the PodTopologySpreadConstraint)
+
+The loki components can be divided into the Write Path (Distributor, Ingester) and the Read Path (Query frontend, Querier, Ingester). In the first pass of the proposal do not distinguish between the two paths. See the `Drawbacks` section for more details on this.
```
xperimental, line 150:
```suggestion
The Loki components can be divided into the Write Path (Distributor, Ingester) and the Read Path (Query frontend, Querier, Ingester). In the first pass of the proposal we do not distinguish between the two paths. See the `Drawbacks` section for more details on this.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+In this step, all componemts and their replicas are configured to spread over different zones. This can be done by making use of the `PodTopologySpreadConstraint` feature in Kubernetes. (See <https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/>)
```
xperimental, line 151:
```suggestion
In this step, all components and their replicas are configured to spread over different zones. This can be done by making use of the [`PodTopologySpreadConstraint` feature in Kubernetes](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/).
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+
+Commonly used node labels to identify domains are
```
xperimental, line 153:
```suggestion
Commonly used node labels to identify failure-domains are
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+Kubernetes makes a few assumptions about the structure of zones and regions:
+
+- regions and zones are hierarchical: zones are strict subsets of regions and no zone can be in 2 regions
+- zone names are unique across regions; for example region "africa-east-1" might be comprised of zones "africa-east-1a" and "africa-east-1b"
+
+The user needs to be aware of the node labels set to identify the different topology domains. Node read operations are generally made available to admin/developer users in OCP. This should be provided in the Lokistack CR topology key so that the `podTopologySpreadConstraint` can use this to schedule the pods accordingly.
+
+These are some of the checks needed before the deployment happens:
+
+- Check if the Topology key provided by the user matches a valid node topology domain label
```
xperimental, line 168:
```suggestion
- Check if the topology-key provided by the user matches a valid node topology domain label
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+- Verify that the replication factor defined is less than the number of available zones
+
+If both these conditions are satisfied, deploy the lokistack components using the `PodTopologySpreadConstraint` definition in the deployments/statefulsets Pod template spec.
+
+### 2. Identifying the failure domain a pod is currently scheduled in (the two solutions discussed in length)
```
xperimental, line 173:
```suggestion
### 2. Identifying the failure-domain a pod is currently scheduled in (the two solutions discussed in length)
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+
+The second step is to identify the domain the pod is scheduled on, to add the value into the loki configuration file(loki-config.yaml). There is no easy way to implement this since the Kubernetes Downward-API does not support exposing node labels within containers. (See issue <https://github.com/kubernetes/kubernetes/issues/40610>)
```
xperimental, line 175:
```suggestion
The second step is to identify the domain the pod is scheduled in and then add this value into the Loki configuration file (`loki-config.yaml`). There is no easy way to implement this since the Kubernetes Downward-API does not support exposing node labels within containers. (see [Kubernetes issue](https://github.com/kubernetes/kubernetes/issues/40610))
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+      downwardAPI:
+        items:
+          - path: "annotations"
+            fieldRef:
+              fieldPath: metadata.annotations['topology.kubernetes.io/zone']
+```
+
+##### Drawbacks of Solution b
+
+- Maintain an image for the  new Init Container. - Since the Init container's work is very basic. We could try to see if we can use an existing lightweight container image for our purpose. That way we can avoid the additional work of maintaining a new image.
```
xperimental, line 389:
```suggestion
- Maintain an image for the new init-container. Since the init-container's work is very basic, ee could try to see if we can use an existing lightweight container image for our purpose. That way we can avoid the additional work of maintaining a new image.
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+
+The container can then be started by passing the configuration via the cli flag `instance_availability_zone: $ZONE`
+
+### Cluster uses multiple topology keys
+
+PodTopologyCOnstraint in Kubernetes is able to handle multiple topology keys and can schedule the pods as expected. But, there is no documentation on how Cortex/Loki handles multiple topology keys. Until further information is found, a simple proposal is to contatenate the values of the different topology keys and create the $ZONE variable for the loki configuration. For eg:
```
xperimental, line 395:
```suggestion
PodTopologyConstraint in Kubernetes is able to handle multiple topology keys and can schedule the pods as expected. But, there is no documentation on how Cortex/Loki handles multiple topology keys. Until further information is found, a simple proposal is to concatenate the values of the different topology keys and create the $ZONE variable for the loki configuration. For example:
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+
+```
+topology.kubernetes.io/zone: "zone-a"
+kubernetes.io/hostname: "ip-172-20-114-199.ec2.internal"
+Concatenate the values
+ZONE: "zone-a-ip-172-20-114-199.ec2.internal"
+```
```
xperimental, line 402:
```suggestion
```plain
# node labels
topology.kubernetes.io/zone: "zone-a"
kubernetes.io/hostname: "ip-172-20-114-199.ec2.internal"
# possible concatenation of the values
ZONE: "zone-a-ip-172-20-114-199.ec2.internal"
```
```

File: operator/docs/enhancements/zone_aware_replication.md [hunk 52]
```diff
@@ -0,0 +1,460 @@
...
+
+### Open Questions
+
+- Both these methods first modify the pod to add the annotation[topology.kubernetes.io/zone: zone-a] in the pod and then modify the container to add a new env var or a volume which picks the zone value from the pod annotation via the downwardapi
```
xperimental, line 406:
```suggestion
- Both these methods first modify the pod to add the annotation `topology.kubernetes.io/zone: zone-a` in the pod and then modify the container to add a new env var or a volume which picks the zone value from the pod annotation via the downward-API
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+- In the Webhook method, is it safe to assume both the pod modification, followed by the container modification can be handled within the webhook?
+- In the operator method, there is going to be two merge patches of the object, one for the pod to add the annotation, and then the container to add the env variable.
+- Would it be better to handle the multiple modifications in an init container. If yes, Is that worth the overhead of  maintainence of an init container?
```
xperimental, line 409:
```suggestion
- Would it be better to handle the multiple modifications in an init-container. If yes, Is that worth the overhead of maintaining an init-container?
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+- If the operator can handle this without init container isnt that better than a webhook?
```
xperimental, line 410:
```suggestion
- If the operator can handle this without init-container isnt that better than a webhook?
```

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+
+### 2. Topology Key limitations
+
+By just using the Topologykey to deploy the replicas in the different zones, we only ensure that 2 pods of different zones are not in the same node. But we cant control 2 replica pods in the same zone getting deployed on the same node. To prevent this we might want to let the user input zone details that can be used as a NodeSelector
```
xperimental, line 422:
```suggestion
By just using the topology-key to deploy the replicas in the different zones, we only ensure that 2 pods of different zones are not in the same node. But we can't control 2 replica pods in the same zone getting deployed on the same node. To prevent this we might want to let the user input zone details that can be used as a NodeSelector.
```

xperimental (commented):
I read through the proposal again, this time for correcting typos, so most of the comments contain suggestions.

xperimental (commented):
I read through the proposal again, this time for correcting typos, so most of the comments contain suggestions.

JoaoBraveCoding (commented):
In general, the document looks great! I left mainly some speeling/formating suggestions + 1 topic for possible discussion.

cstyan:
Sorry, I haven't been keeping up with the conversation here after I chimed in initially. Please @ me directly if you need any input on how we configured or migrated zone aware ingesters.

operator/docs/enhancements/zone_aware_replication.md
periklis: @cstyan :point_up: thread is interesting for you maybe. Do you use the rollout-operator to do automatic failure due to the PVCs bound in one zone?

cstyan: Recap of our slack conversation for those who might be reading along here:

The short answer is no. Currently we only use "zone awareness" in the sense of logical zone awareness. We use the zone awareness config and separate node pools to replicate each piece of data to one ingester in each zone, which means we can roll out ingester pods much faster during deployments. In some cases you can roll an entire zone at once. The rollout operator, via some labels configuration, knows about the three different statefulsets, and ensures that all pods in a single zone are rolled before moving on to the next one.

To use actual phsyical zone awareness the implementation at the hardware or cloud provider level is likely what would be needed. Some options might be via things like GKE mutli-zone and/or regional clusters, google cloud regional persistent disks, and manually creating a node pool in each zone for each ingester statefulset each of which would have a taint, and then each ingester statefulset can be constrained to that set of nodepools via it's toleration config.

operator/docs/enhancements/zone_aware_replication.md
shwetaap: Changed it to required

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+        items:
+          - path: "annotations"
+            fieldRef:
+              fieldPath: metadata.annotations['topology.kubernetes.io/zone']
+```
+
+##### Drawbacks of Solution a
+
+- The status of the application pod during the waiting phase is unknown
+- The liveness&readiness probes will fail since the loki application would not have started
```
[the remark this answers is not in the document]
shwetaap: Added a note on startupprobes in the Open questions section

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+
+# Table of Contents
+
+1. [Summary](#summary)
+2. [Motivation](#motivation)
+3. [Proposal](#proposal)
+    1. [API Extensions](#api-extensions)
+    2. [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+4. [Design Details](#design-details)
+5. [Impact on LokiStack Sizing Configurations](#sizing)
```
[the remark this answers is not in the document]
shwetaap: Removed it

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+
+## Implementation History
+
+## Drawbacks
+
+### 1. Read and Write Path
+
+As suggested in the `Design` section there is no separate enabling of the Read and the Write path in the initial pass of the feature implementation. The benefits of enabling this separately is that, during live migrations we can first turn on write path zone awareness, leave it enabled for the max query lookback period, and then enable read path zone awareness. This means that at the time we turn on the read path zone awareness all ingesters that should have data for a stream will have all the data for that stream. If you enable the read path zone awareness earlier, then potentially data would be missing in queries. At this point the loki-operator deploys all the components in an all or nothing fashion. So the users should expect missing queries in the data for the max query lookback period.To avoid missing the queries, the operator needs to update the deployment process to two stages - first the components in the write path are deployed, wait the max query lookback period(currently 30s), followed by the deployment of the components in the read path.
```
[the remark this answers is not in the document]
shwetaap: Done

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,460 @@
...
+kubernetes.io/hostname: "ip-172-20-114-199.ec2.internal"
+Concatenate the values
+ZONE: "zone-a-ip-172-20-114-199.ec2.internal"
+```
+
+### Open Questions
+
+- Both these methods first modify the pod to add the annotation[topology.kubernetes.io/zone: zone-a] in the pod and then modify the container to add a new env var or a volume which picks the zone value from the pod annotation via the downwardapi
+- In the Webhook method, is it safe to assume both the pod modification, followed by the container modification can be handled within the webhook?
+- In the operator method, there is going to be two merge patches of the object, one for the pod to add the annotation, and then the container to add the env variable.
```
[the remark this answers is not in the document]
shwetaap: Removed it :)

operator/docs/enhancements/zone_aware_replication.md
```diff
@@ -0,0 +1,448 @@
...
+      downwardAPI:
+        items:
+          - path: "annotations"
+            fieldRef:
+              fieldPath: metadata.annotations['topology.kubernetes.io/zone']
+```
+
+##### Drawbacks of Solution b
+
+- Maintain an image for the new init-container. Since the init-container's work is very basic, ee could try to see if we can use an existing lightweight container image for our purpose. That way we can avoid the additional work of maintaining a new image.
```
bparees, line 375: is there a reason you can't use the loki image itself for this, just w/ a different startup command?  The node has to pull the loki image anyway, so there's no additional overhead.

PR #8382, opened 2023-02-02
Promtail: Add event log message stage

kelnage:
The Windows Event Log includes a field named `Message` that include a large amount of additional data, stored in a simple key-value format: 
* key-value pairs are separated with Windows-style `\r\n` newlines,
* the key and value are separated with `: ` (or just `:` if the key has no associated value).

The key-value data in the `Message` field is commonly utilised when investigating Windows events, e.g., in security use-cases, but the existing `windows_events` pipeline stage keeps the Message data as a string, leading to requiring significant effort at the query time to extract that data to be used in field filters.

This PR request introduces a new promtail stage that parses a specified Source (defaulting to `message` used by the `windows_events`) and extracts all the key-value pairs into the extracted fields, which can then be used in latter stages in the pipeline. An example configuration I am currently testing is as follows:

```yaml
- job_name: windows
  windows_events:
    use_incoming_timestamp: false
    bookmark_path: "./bookmark.xml"
    eventlog_name: "Microsoft-Windows-Sysmon/Operational"
    labels:
      job: windows
  pipeline_stages:
  - json:
      expressions:
        message:
  - eventlogmessage:
      source: message
      overwrite_existing: true
  - pack:
      labels:
        - CommandLine
        - Image
        - EventID
        - ParentImage
        - ScriptBlockText
        - ParentCommandLine
```

I have a few aspects that I wasn't sure about and would appreciate your feedback on:
* I noted a number of other stages included a map of fields to extract (as well as optionally renaming them). I decided against this approach as there is a large range of different fields that a Message might/might not contain and automatically renaming any fields that were not valid labels (replacing invalid characters with underscores)
* There is a possibility this stage will extract fields from the Message that already exist in the data. By default, the stage does not overwrite those existing fields (unless configured otherwise) - but I suppose it could take the approach used in Loki of adding `_extracted` to the field name as a default action - would that make sense?
* I decided against using the whole string as the input by default as the event log contains a large number of other fields that will generally be relevant (such as the event\_id).

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/clients/pkg/logentry/stages/eventlogmessage.go
+++ b/clients/pkg/logentry/stages/eventlogmessage.go
@@ -105,7 +105,7 @@ func (m *eventLogMessageStage) processEntry(extracted map[string]interface{}, ke
 	}
 	lines := strings.Split(s, "\r\n")
 	for _, line := range lines {
-		parts := RegexSplitKeyValue.Split(line, 2)
+		parts := strings.SplitN(line, ":", 2)
 		if len(parts) < 2 {
 			level.Warn(m.logger).Log("msg", "invalid line parsed from message", "line", line)
 			continue
@@ -126,6 +126,9 @@ func (m *eventLogMessageStage) processEntry(extracted map[string]interface{}, ke
 			mkey += "_extracted"
 		}
 		mval := parts[1]
+		if len(mval) > 0 {
+			mval = mval[1:]
+		}
 		if !model.LabelValue(mval).IsValid() {
 			if Debug {
 				level.Debug(m.logger).Log("msg", "invalid value parsed from message", "value", mval)
--- a/clients/pkg/logentry/stages/eventlogmessage.go
+++ b/clients/pkg/logentry/stages/eventlogmessage.go
@@ -152,13 +152,13 @@ func SanitizeLabelName(input string) string {
 	if len(input) == 0 {
 		return "_"
 	}
-	var valid_sb strings.Builder
+	var validSb strings.Builder
 	for i, b := range input {
 		if !((b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z') || b == '_' || (b >= '0' && b <= '9' && i > 0)) {
-			valid_sb.WriteRune('_')
+			validSb.WriteRune('_')
 		} else {
-			valid_sb.WriteRune(b)
+			validSb.WriteRune(b)
 		}
 	}
-	return valid_sb.String()
+	return validSb.String()
 }
--- a/clients/pkg/logentry/stages/eventlogmessage.go
+++ b/clients/pkg/logentry/stages/eventlogmessage.go
@@ -98,18 +98,14 @@ func (m *eventLogMessageStage) processEntry(extracted map[string]interface{}, ke
 	}
 	s, err := getString(value)
 	if err != nil {
-		if Debug {
-			level.Debug(m.logger).Log("msg", "invalid label value parsed", "value", value)
-		}
+		level.Warn(m.logger).Log("msg", "invalid label value parsed", "value", value)
 		return err
 	}
 	lines := strings.Split(s, "\r\n")
 	for _, line := range lines {
 		parts := regexp.MustCompile(": ?").Split(line, 2)
 		if len(parts) < 2 {
-			if Debug {
-				level.Debug(m.logger).Log("msg", "invalid line parsed from message", "line", line)
-			}
+			level.Warn(m.logger).Log("msg", "invalid line parsed from message", "line", line)
 			continue
 		}
 		mkey := parts[0]
--- a/clients/pkg/logentry/stages/eventlogmessage.go
+++ b/clients/pkg/logentry/stages/eventlogmessage.go
@@ -21,6 +21,7 @@ const (
 type EventLogMessageConfig struct {
 	Source            *string `mapstructure:"source"`
 	DropInvalidLabels bool    `mapstructure:"drop_invalid_labels"`
+	OverwriteExisting bool    `mapstructure:"overwrite_existing"`
 }
 
 type eventLogMessageStage struct {
@@ -116,15 +117,15 @@ func (m *eventLogMessageStage) processEntry(extracted map[string]interface{}, ke
 		if !model.LabelName(mkey).IsValid() {
 			if m.cfg.DropInvalidLabels {
 				if Debug {
-					level.Debug(m.logger).Log("msg", "invalid key parsed from message", "key", mkey)
+					level.Debug(m.logger).Log("msg", "invalid label parsed from message", "key", mkey)
 				}
 				continue
 			}
 			mkey = SanitizeLabelName(mkey)
 		}
-		if _, ok := extracted[mkey]; ok {
+		if _, ok := extracted[mkey]; ok && !m.cfg.OverwriteExisting {
 			if Debug {
-				level.Debug(m.logger).Log("msg", "extracted key that already existed, ignoring", "key", mkey)
+				level.Debug(m.logger).Log("msg", "extracted label that already existed, ignoring", "key", mkey)
 			}
 			continue
 		}
--- a/clients/pkg/logentry/stages/eventlogmessage.go
+++ b/clients/pkg/logentry/stages/eventlogmessage.go
@@ -17,6 +17,8 @@ const (
 	ErrEmptyEvtLogMsgStageConfig = "empty event log message stage configuration"
 )
 
+var RegexSplitKeyValue = regexp.MustCompile(": ?")
+
 type EventLogMessageConfig struct {
 	Source            *string `mapstructure:"source"`
 	DropInvalidLabels bool    `mapstructure:"drop_invalid_labels"`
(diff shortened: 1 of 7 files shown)


jeschkies:
> I suppose it could take the approach used in Loki of adding _extracted to the field name as a default action - would that make sense?

Yes 😊 I always forget what suffix we use but that's the way.

clients/pkg/logentry/stages/eventlogmessage.go
```diff
@@ -0,0 +1,164 @@
...
+	key := defaultSource
+	if m.cfg.Source != nil {
+		key = *m.cfg.Source
+	}
+	go func() {
+		defer close(out)
+		for e := range in {
+			err := m.processEntry(e.Extracted, key)
+			if err != nil {
+				continue
```
jeschkies, line 81: Could you report the error in the logs as a warning?

clients/pkg/logentry/stages/eventlogmessage.go
```diff
@@ -0,0 +1,164 @@
...
+	s, err := getString(value)
+	if err != nil {
+		if Debug {
+			level.Debug(m.logger).Log("msg", "invalid label value parsed", "value", value)
+		}
+		return err
+	}
+	lines := strings.Split(s, "\r\n")
+	for _, line := range lines {
+		parts := regexp.MustCompile(": ?").Split(line, 2)
```
jeschkies, line 108: For efficiency the regex soups be compiled only once. I'm wondering if regex is required at all though.

clients/pkg/logentry/stages/eventlogmessage.go
```diff
@@ -0,0 +1,164 @@
...
+		if len(parts) < 2 {
+			if Debug {
+				level.Debug(m.logger).Log("msg", "invalid line parsed from message", "line", line)
```
jeschkies, line 111: They should probably be a warning what do you think?

clients/pkg/logentry/stages/eventlogmessage.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return nil
+}
+
+func (m *eventLogMessageStage) Name() string {
+	return StageTypeEventLogMessage
+}
+
+// Sanitize a input string to convert it into a valid prometheus label
+func SanitizeLabelName(input string) string {
```
jeschkies, line 151: We should have this method somewhere already 😊

jeschkies (changes requested):

clients/pkg/logentry/stages/eventlogmessage.go
kelnage: I found [this function](https://github.com/prometheus/prometheus/blob/c70d85baed260f6013afd18d6cd0ffcac4339861/util/strutil/strconv.go#L41) in Prometheus - and I saw it is being used elsewhere in promtail - but it has a (admittedly subtle) bug: it allows `0key` as a label, yet the first character of a Prometheus label must be `[a-zA-Z_]`. Happy to push this fix upstream (though I can't see a way of doing it with regex replace unless we use two regexes) - but I'd prefer not to depend on it for the time being.

clients/pkg/logentry/stages/eventlogmessage.go
kelnage: My alternative approach was going to be splitting on `:` and using strings.Trim* to remove the leading space(s) (if applicable), e.g.:

```go
parts := strings.SplitN(line, ":", 2)
...
mval := strings.TrimLeft(parts[1], " ")
```

Alternatively, it could just take off the first character if the value is not empty:

```go
mval := parts[1]
if len(mval) > 0 {
    mval = mval[1:]
}
```

Any preferences? If pushed, I guess I'd say I prefer the latter as it preserves the value as much as possible.

clients/pkg/logentry/stages/eventlogmessage.go
kelnage: There's currently only a single context where err won't be nil (when getString returns an err) - I've now generated a warning when that occurs - so I think this should be covered (seems a little overkill to warn about it twice).

clients/pkg/logentry/stages/eventlogmessage.go
kelnage: I've [submitted a PR](https://github.com/prometheus/prometheus/pull/11936) for my fix upstream - once it is accepted, I'll remove this code and import it.

clients/pkg/logentry/stages/eventlogmessage.go
jeschkies: Could you add a small benchmark and move the regex compile out? I think this is fine. We could follow up with a different version if it's faster.

kelnage: I've [added (hopefully fair!) benchmarks](https://github.com/grafana/loki/blob/88d2c29d6ffa48d491fffb9a9499c909ac1ae332/clients/pkg/logentry/stages/eventlogmessage_test.go#L344) for all three options:

```
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/clients/pkg/logentry/stages
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
BenchmarkSplittingKeyValuesRegex-20          	 1000000	      1158 ns/op	     614 B/op	       8 allocs/op
BenchmarkSplittingKeyValuesSplitTrim-20      	 1823691	       663.4 ns/op	     339 B/op	       5 allocs/op
BenchmarkSplittingKeyValuesSplitSubstr-20    	 2126520	       562.4 ns/op	     339 B/op	       5 allocs/op
```

```
goos: windows (VM)
goarch: amd64
pkg: github.com/grafana/loki/clients/pkg/logentry/stages
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
BenchmarkSplittingKeyValuesRegex-4         	 1850302	       654.9 ns/op	     612 B/op	       8 allocs/op
BenchmarkSplittingKeyValuesSplitTrim-4     	 3055104	       389.8 ns/op	     338 B/op	       5 allocs/op
BenchmarkSplittingKeyValuesSplitSubstr-4   	 2888671	       374.5 ns/op	     338 B/op	       5 allocs/op
```

Looks to me like the two Split-based options are basically neck and neck - and given the string slicing option appears to be slightly more "correct", I'm going to go with it.

docs/sources/clients/promtail/stages/eventlogmessage.md
```diff
@@ -0,0 +1,69 @@
...
+  [drop_invalid_labels: <bool> | default = false]
+```
+
+The extracted data can hold non-string values and this stage does not do any
+type conversions; downstream stages will need to perform correct type
+conversion of these values as necessary. Please refer to the
+[the `template` stage]({{<relref "template">}}) for how to do this.
+
+
+## Example
```
JStickler, line 35: Empty heading (we can delete)

docs/sources/clients/promtail/stages/eventlogmessage.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Combined with json
```
JStickler, line 37:
```suggestion
## Example combined with json
```

JStickler (approved):
[Docs squad]  Just one small suggestion, otherwise LGTM.

clients/pkg/logentry/stages/eventlogmessage.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+		if _, ok := extracted[mkey]; ok && !m.cfg.OverwriteExisting {
+			level.Info(m.logger).Log("msg", "extracted key that already existed, appending _extracted to key",
+				"key", mkey)
+			mkey += "_extracted"
+		}
+		mval := parts[1]
+		if len(mval) > 0 {
+			mval = mval[1:]
+		}
```
chaudum, line 131: I think you could use `TrimSpace` to trim off whitespaces before and after.

```suggestion
		mval := strings.TrimSpace(parts[1])
```

chaudum (commented):
Otherwise LGTM

kelnage:
Thank you for all your feedback @jeschkies, @chaudum and @JStickler - I believe I've made all the requested changes - let me know if there's anything I might have missed!

jeschkies (approved):

chaudum:
> Thank you for all your feedback @jeschkies, @chaudum and @JStickler - I believe I've made all the requested changes - let me know if there's anything I might have missed!

Looks good! One thing that I almost forgot: Could you please also add a changelog entry underneath the `Promtail` section?

wz2b:
I have an issue with this, where it's apparently unable to parse the Message field.  The error is:
```
level=warn ts=2023-06-06T17:18:30.0763048Z caller=eventlogmessage.go:107 component=windows_pipeline component=stage
type=event_log_message msg="invalid line parsed from message" line="\t\t\t\t"
```

the log's message fields look more or less like this:
```
"message":"Cryptographic operation.\r\n\r\nSubject:\r\n\tSecurity ID:\t\tS-1-5-18\r\n\tAccount Name:\t\tGIS-SHARES$\r\n\tAccount Domain:\t\tAD\r\n\tLogon ID:\t\t0x3E7\r\n\r\nCryptographic Parameters:\r\n\tProvider Name:\tMicrosoft Software Key Storage Provider\r\n\tAlgorithm Name:\tRSA\r\n\tKey Name:\tConfigMgrPrimaryKey\r\n\tKey Type:\tMachine key.\r\n\r\nCryptographic Operation:\r\n\tOperation:\tOpen Key.\r\n\tReturn Code:\t0x0"}
```

I know that's messy (sorry).  Basically, the code wants to split on colon:

```
parts := strings.SplitN(line, ":", 2)
```

which by itself shouldn't be a problem, I think ... except that all these log messages seem to start with a string title (without a colon) so it's going to get confused by that.  I think that's really where the problem lies - in the security log there are all lines that aren't `key: value`

PR #8420, opened 2023-02-06
Add LID for improving Index Gateway sharding

chaudum:
This document tries to come up with a proposal on how to do a better sharding of data on the index gateways so we are able to scale the service horizontally to fulfill the increased need for metadata queries of big tenants.

### Update (2023-09-05)

I gave the LID an ordinal number (0004) and marked it as Rejected / Not Implemented so this PR can be merged.


docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
```diff
@@ -0,0 +1,99 @@
...
+The client can execute a request on any index gateway. This handler instance then identifies the index files that are involved, splits the query, and resolves the appropriate instances. Once it received the sub-queries it resembles the full response result and sends it back to the client.
+
+**Pros:**
+* Sharding is handled transparently to the client.
+* Clients can communicate with any instance of the index gateway ring.
+* Domain information about splitting and merging is kept within index gateway server implementation.
+
+Due to it's architectural advantages, option B is proposed.
+
+While this proposal suggests the biggest change to the index gateway's sharding mechanism in terms of effort and complexity, it also offers the biggest win in terms of scalability.
```
owen-d, line 93: I'm not sure this is that different. Either the querier|query-frontend act as a "client", scattering/gathering requests across idx-gw instances, or an individual idx-gw does. Theoretically it could reduce some gossip pressure if non idx-gw nodes don't need to be aware of idx-gw ring membership, but I'm not sure this is a current issue or if our memberlist library supports this easily.

owen-d (commented):
Thanks for putting this together, left some thoughts

docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
```diff
@@ -0,0 +1,99 @@
...
+
+The index gateway service can be run in "simple mode", where an index gateway instance is responsible for handling, storing and returning requests for all indices for all tenants, or in "ring mode", where an instance is responsible for a subset of tenants instead of all tenants.
+
+On top of that, in order to achieve redundancy as well as spreading load, the index gateway ring uses a replication factor of 3.
+
+This means, before an index gateway client makes a request to the index gateway server, it first hashes the tenant ID and then requests a replication set for that hash from the index gateway ring. Due to the fixed replication factor (RF), the replication set contains three server addresses. On every request, a random server from that list is picked to then execute the request on.
+
+## Problem Statement
+
+The current strategy of sharding by tenant ID and having a replication factor of 3 fails in the long run, because even when running lots of index gateways, only a maximum of three could be utilized by a single tenant.
```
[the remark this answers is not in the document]
chaudum: Updated the wording a bit to clarify that the RF is configurable.

docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
```diff
@@ -0,0 +1,99 @@
...
+
+Instead of using a fixed replication factor of 3, the RF can be derived from the amount of active members in the index gateway ring. That means that the RF would be a percentage of the available gateway instances. For example, a ring with 12 instances and 30% replication utilization would result in a RF of 3 (`floor(12*0.3)`). Scaling up to 18 instances would result in a RF of 5.
+
+This approach would solve the problem of horizontal scaling. However, it does not solve the problem of different tenant sizes. It also fails to ensure replication for availability when running a small number of instances, unless there is a fixed lower value for the RF. It also tends to over-shard data in large index gateway deployments.
+
+### Proposal 2: Fixed per-tenant replication factor
+
+Adding a random shard ID (e.g. `shard-0`, `shard-1`, ... `shard-n`) to the tenant ID allows to utilize a certain amount of `n` instances. The amount of shards can be implemented as a per-tenant override. However, this approach results in non-deterministic hash keys.
+
+### Proposal 3: Shard by index files
```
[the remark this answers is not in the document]
chaudum: By usage, do you mean the size of the index?

docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
```diff
@@ -0,0 +1,97 @@
...
+
+The index gateway service can be run in "simple mode", where an index gateway instance is responsible for handling, storing and returning requests for all indices for all tenants, or in "ring mode", where an instance is responsible for a subset of tenants instead of all tenants.
+
+On top of that, in order to achieve redundancy as well as spreading load, the index gateway ring uses by default a replication factor of 3.
+
+This means, before an index gateway client makes a request to the index gateway server, it first hashes the tenant ID and then requests a replication set for that hash from the index gateway ring. Due to the fixed replication factor (RF), the replication set contains three server addresses. On every request, a random server from that list is picked to then execute the request on.
+
+## Problem Statement
+
+The current strategy of sharding by tenant ID and having a replication factor fails in the long run, because even when running lots of index gateways, only a maximum of `n` instances could be utilized by a single tenant, where `n` is the value of the configured RF.
```
jeschkies, line 36: Is the problem that's some tenants see heavier traffic so a relocation factor of three doesn't scale? If so I think it should be mentioned here.

docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
```diff
@@ -0,0 +1,97 @@
...
+The goal of this document is to find a better sharding mechanism for the index gateway, so that there are no boundaries for scaling the service horizontally.
+
+* The sharding needs to account for the "size" of a tenant.
+* A single tenant needs to be able to utilize more than three index gateways.
+
+## Proposals
+
+### Proposal 0: Do nothing
+
+If we do not improve the sharding mechanism for the index gateways and leave it as it is, it will become more and more difficult to serve metadata queries for large tenant in a reasonable amount of time, proportionally to the demand for these queries.
```
jeschkies, line 51: Nit: `for large tenants`

docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Proposal 1: Dynamic replication factor
+
+Instead of using a fixed replication factor of 3, the RF can be derived from the amount of active members in the index gateway ring. That means that the RF would be a percentage of the available gateway instances. For example, a ring with 12 instances and 30% replication utilization would result in a RF of 3 (`floor(12*0.3)`). Scaling up to 18 instances would result in a RF of 5.
+
+This approach would solve the problem of horizontal scaling. However, it does not solve the problem of different tenant sizes. It also fails to ensure replication for availability when running a small number of instances, unless there is a fixed lower value for the RF. It also tends to over-shard data in large index gateway deployments.
```
jeschkies, line 57: Does this also imply a resharding?

jeschkies (commented):
I think proposal 3 is the most reasonable.

docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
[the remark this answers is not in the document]
chaudum: Ok, not size of the index, but how often the index is queried

docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
chaudum: That is right. A replication factor of 3 may not be sufficient for big tenants. It can be increased, however, that does not change the sharding itself. It just means that more instances can be utilized. Since RF is a ring property it affects all tenants alike.

docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
chaudum: Active re-sharding is not considered in any of the proposals. At the moment, index files are treated as transient/intermediate resources, because they can be re-downloaded for processing when the assignment changes.

docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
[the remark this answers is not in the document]
owen-d: Size plays a part, but I'm referring more to the underlying _work_ that needs to be done. For instance, tenant A may have 100TB/day of ingest and does a lot of querying. Even with sharding, we'd "hotspot" access to this daily index for said tenant.

docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Proposal 2: Fixed per-tenant replication factor
+
+Adding a random shard ID (e.g. `shard-0`, `shard-1`, ... `shard-n`) to the tenant ID allows to utilize a certain amount of `n` instances. The amount of shards can be implemented as a per-tenant override setting. This would allow to use different amount of instances for each tenant. However, this approach results in non-deterministic hash keys.
+
+### Proposal 3: Shard by index files
+
+In order to answer requests, the index gateway needs to download index files from the object storage, and since Loki builds a daily index file per tenant, these index files can be sharded evenly across all available index gateway instances. Each instance is then assigned a unique set of index files which it can answer metadata queries for.
+
+This means that the sharding key is the name of the file in object storage. While this name encodes both the tenant and the date, this is not strictly necessary. Such a sharding mechanism could shard any files from object storage across a set of instances of a ring.
```
chaudum, line 67: Sharding by file name solves the problem of distributing a single tenant's files across multiple instances. However, it does not solve the problem of sharding the workload of a single tenant across, because metadata queries for a single day are assigned to a single instance regardless of how frequent the data is queried and how big the index is.

chaudum: This could be solved by tracking the access rate to a single resource (index file) and based on that add a replica suffix to the filename for the hash function input.

Dummy example:

```
   RPS -> MAX_REPLICAS
  1rps -> 1
 10rps -> 2
100rps -> 4

key = HASH_FUNC(FILENAME + ";" + RAND_INT(0, MAX_REPLICAS))
```

This would not shard a single resource that is requested more often to multiple instances, essentially spreading the load of the requests, when every request picks a random replica.

Note, that this replica is independent of the ring's replication factor!

docs/sources/lids/_DRAFT-Index-Gateway-Sharding.md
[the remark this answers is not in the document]
sandeepsukhani: I like this proposal but have a couple of suggestions to avoid hot-spotting and make it less agnostic of underlying storage:
* Use table number/bucket instead of file names to avoid jumping around index gateways since each file change also changes the file names. This also keeps the index client agnostic of how the index is managed on the storage.
* Consider hashing with tenant id as well with the table number to avoid hot-spotting the same index gateways for all the tenants querying data in the same time interval.

This would help better distribute the load.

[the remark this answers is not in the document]
sandeepsukhani: this is how we calculate table number/buckets. You can ignore table name since it won't add any value here
https://github.com/grafana/loki/blob/2208ea7147d4ac99071cd9d468990d97e12bc447/pkg/storage/stores/tsdb/manager.go#L284

dannykopping (approved):

PR #8425, opened 2023-02-06
Skip image validation

jdbaldry:
In the future, all images will be stored in GCS and not in the repository and this linting rule will be removed.

Until then, skip insisting that all images are in page bundles.


chaudum (approved):
Thank you!

PR #8435, opened 2023-02-07
Update publishing workflows to use organization secret

jdbaldry:
The new tokens are managed centrally and have a longer expiry. Administrators of the grafanabot account will be
notified of the pending expiry and the secret can be rotated centrally without the need for a repository administrator to update their secrets.

The existing repository secrets can safely be removed. The tokens for those secrets will be removed by the end of this week.


osg-grafana (approved):

MichelHollands (approved):
LGTM

PR #8453, opened 2023-02-07
Fix ruler deadlock

MasslessParticle:
A circular dependency between the Prometheus `rules.Manager` and the `MemStore` causes occasional deadlock in the rulers. 

Fixing the circular dependency is non-trivial because we rely on the `rules.Manager` to be the source-of-truth for rules storage. Instead, this PR moves retrieving rules to outside the `memStoreQuerier`. Rules are now retrieved and cached as part of the `run` loop. Rules are infrequently updated so the 5 minute lag between updates should be acceptable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 
 ##### Fixes
 
+* [8453](https://github.com/grafana/loki/pull/8453) **masslessparticle** Fix deadlock in rulers.
 * [7784](https://github.com/grafana/loki/pull/7784) **isodude**: Fix default values of connect addresses for compactor and querier workers to work with IPv6.
 * [7880](https://github.com/grafana/loki/pull/7880) **sandeepsukhani**: consider range and offset in queries while looking for schema config for query sharding.
 * [7937](https://github.com/grafana/loki/pull/7937) **ssncferreira**: Deprecate CLI flag `-ruler.wal-cleaer.period` and replace it with `-ruler.wal-cleaner.period`.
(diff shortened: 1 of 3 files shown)


trevorwhitney (commented):
This looks good to me, but would like another review before we merge as this is the first time I've looked at this code.

owen-d (changes requested):
The memstore is only used to precalculate the historical state of the rule (for example, whether it should be considered “firing” upon startup). It’s only used when a rule is first added to a ruler.

The memstore only cares about finding the rule definitions (such as the query, `for` period, etc). We can probably cache the `LoadGroups` call and have the memstore use that for rule iteration instead. It’s called early in `manager.Update` [here](https://github.com/prometheus/prometheus/blob/e023d896f2d79b1a48bbd345429c6b496cf46cc8/rules/manager.go#L1007).

tl;dr: I think we need to cache the result of `LoadGroups` whenever it's called and have that result fulfill the `AlertingRules` interface. It's guaranteed to be populated by the time `memstoreQuerier.Select` is called. This is definitely still ugly, hard to refactor, etc, but it should fix the deadlock.

Your current solution isn't guaranteed to have the alerting rule definitions cached when the `memstoreQuerier.Select` is called, meaning we'll see errors [here](https://github.com/grafana/loki/blob/main/pkg/ruler/memstore.go#L213-L217) when the race condition doesn't work out in our favor.

MasslessParticle:
Closed in favor of https://github.com/grafana/loki/pull/8486

PR #8461, opened 2023-02-08
More precise sharding logic for `on|ignoring`

owen-d:
While adding some tests to compliment https://github.com/grafana/loki/pull/8448, I noticed some imprecision when determining whether a binary operation is shardable. Particularly,
* It is unshardable when using `on` grouping or when using `ignoring` with a non-zero set of labels to ignore as this can mutate labels.
* It is shardable when using `ignoring ()` is effectively the zero value which cannot mutate labels.


chaudum (approved):
Great find!

PR #8467, opened 2023-02-08
Helm enterprise-logs: update kubernetes secrets creation

ssncferreira:

Fix the Kubernetes secrets creation for the tokengen and provisioner jobs for multiple helm installations/uninstallations. 

The execution flow for the helm installation is the following:
1) `tokengen` job `loki` container generates a new admin token, stores it in the object storage and writes it to the `shared` mount.
2) `tokengen` job `create-secret` container creates a new Kubernetes secret for the generated admin token.
3) `provisioner` job `provisioner` container creates the new admin resources via the Admin API and writes the tokens to the `bootstrap` mount.
4) `provisioner` job `create-secret` container creates new Kubernetes secrets for each token of the additional tenants.

## Previous implementation 

Uninstall and re-install the enterprise-logs helm:
1) `tokengen` job `loki` container generates a new admin token, stores it in the object storage, and writes it to the `shared` mount. This is because this job uses an [emptyDir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) mount which is deleted once the job completes. When a new job is created, the directory will be again empty, forcing the `tokengen` to generate a new token.
2) `tokengen` job `create-secret` container fails, because an admin token secret already exists:
```
Defaulted container "create-secret" out of: create-secret, loki (init)
error: failed to create secret secrets "enterprise-logs-admin-token" already exists
```
3) The helm installation fails ❌ 

## Implementation in this PR

1) `tokengen` job `loki` container generates a new admin token, stores it in the object storage, and writes it to the `shared` mount. 
2) `tokengen` job `create-secret` container updates the previously created Kubernetes secret for the admin token.
3) `provisioner` job `provisioner` container checks that the admin resources already exist and doesn't write the token files to the `bootstrap` mount.
4) `provisioner` job `create-secret` container checks that the token files are not present on the `bootstrap` mount, and does not create new secrets.
5) The helm installation succeeds ✅ 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


production/helm/loki/templates/tokengen/job-tokengen.yaml
```diff
@@ -45,6 +45,10 @@ spec:
           image: {{ template "loki.image" . }}
           imagePullPolicy: {{ .Values.loki.image.pullPolicy }}
           args:
+            # The shared emptyDir exists only while the job is running, and deleted once the job is completed.
+            # The tokengen generates a new admin token in case the 'token-file' file doesn't exist.
+            # As a result, subsequent executions of this tokengen job will generate new admin tokens.
+            # Note that previously generated tokens remain valid.
```
ssncferreira, line 51: This means that each installation will generate a new admin token which will be stored in the object storage. The Kubernetes secret will be updated to the latest generated token. 

For instance, if 3 helm installations take place, there will be 3 valid admin tokens stored in the object storage. The Kubernetes secret will be updated with the latest generated token. All these tokens will remain valid.

production/helm/loki/templates/tokengen/job-tokengen.yaml
```diff
@@ -45,6 +45,10 @@ spec:
           image: {{ template "loki.image" . }}
           imagePullPolicy: {{ .Values.loki.image.pullPolicy }}
           args:
+            # The shared emptyDir exists only while the job is running, and is deleted once the job is completed.
+            # The tokengen generates a new admin token in case the 'token-file' file doesn't exist.
+            # As a result, subsequent executions of this tokengen job will generate new admin tokens.
+            # Note that previously generated tokens remain valid, as these remain present in the object storage.
```
ssncferreira, line 51: This means that each installation will generate a new admin token which will be stored in the object storage. The Kubernetes secret will be updated to the latest generated token. This happens, because of the `tokengen` implementation to generate a new token in each execution, independently if an admin token already exists in the object storage: https://github.com/grafana/backend-enterprise/blob/master/pkg/enterprise/admin/bootstrap/tokengen.go#L46

For instance, if 3 helm installations take place, there will be 3 valid admin tokens stored in the object storage. The Kubernetes secret will be updated with the latest generated token. All these tokens will remain valid.

jeschkies (approved):
Awesome work. I wonder if we should put the scripts into an external file at some point. But this would be out of scope of this change.

Waiting for @trevorwhitney  to take notice.

production/helm/loki/templates/provisioner/job-provisioner.yaml
```diff
@@ -92,21 +92,27 @@ spec:
...
             - -exuc
             - |
+              # In case, the admin resources have already been created, the provisioner job
+              # does not write the token files to the bootstrap mount.
+              # Therefore, secrets are only created if the respective token files exist.
               {{- range .Values.enterprise.provisioner.additionalTenants }}
-              kubectl --namespace "{{ .secretNamespace }}" create secret generic "{{ include "enterprise-logs.provisionedSecretPrefix" $ }}-{{ .name }}" \
-                --from-literal=token-write="$(cat /bootstrap/token-write-{{ .name }})" \
-                --from-literal=token-read="$(cat /bootstrap/token-read-{{ .name }})"
+              ! test -s /bootstrap/token-write-{{ .name }} || \
```
trevorwhitney, line 99: The negation in the logic took me a bit longer to process. Would `test -s /bootstrap/token-write-{{ .name }} &&` work instead? I personally find that logic a little easier to read.

production/helm/loki/templates/tokengen/job-tokengen.yaml
```diff
@@ -79,10 +83,17 @@ spec:
             - /bin/bash
             - -euc
             - |
-              kubectl create secret generic "{{ include "enterprise-logs.adminTokenSecret" . }}" --from-file=token=/shared/admin-token
+              # Create or update admin token secrets generated by tokengen job
+              kubectl create secret generic "{{ include "enterprise-logs.adminTokenSecret" . }}" \
+                --from-file=token=/shared/admin-token \
+                --dry-run=client -o yaml \
+                | kubectl apply -f -
```
trevorwhitney, line 90: clever piping through to an apply to get the create/update behavior, nice!

production/helm/loki/templates/tokengen/job-tokengen.yaml
trevorwhitney: This sounds like something we should change at a later point in the tokengen job. I wonder if there's a way to make that target itself idempotent?

trevorwhitney (approved):
Looks great! Just one small nit about the bash logic, otherwise LGTM!

production/helm/loki/templates/provisioner/job-provisioner.yaml
ssncferreira: Agree, initially wrote it like that however, we want the bash command to return a successful status code, otherwise (if an error status code is returned) the installation of the provision job is not considered successful. 
Added a comment to justify this https://github.com/grafana/loki/pull/8467/commits/987b8125d3906b998d78b86b0ecf135329dabb4c 👍

PR #8469, opened 2023-02-08
DynamoDB: Accept a KMS key for DynamoDB tables

### LINKED ISSUE #8466: Support KMS for DynamoDb
Currently there is no way to configure a KMS for Loki indexes.  My immediate usecase is to be able to specify an AWS KMS customer managed key (CMK) when using a DynamoDB index.  The TableDesc nor DynamoDbConfig do not have anything related to key providers in it, so the DynamoDB Table Client is unable to set this value.  This results in DynamoDB indexes only being able to use an Amazon Owned KMS key.

I'd like to see a `kms_key_id` added to the DynamoDb configuration.

* I could not use Loki's table creation mechanism, and instead have a Lambda that pre-creates tables for Loki and provides the correct KMS key
* I could have a Lambda that runs periodically and updates the table to use the proper KMS key if it is not set.
* I investigated adding this to to the TableDesc struct, but I don't think it belongs there as it doesn't apply to many of the other tables.  For example, this is a config parameter on the BigTable cluster that gets automatically applied to the tables in the cluster.  This mirrors how S3 is configured at the aws level instead of at a blob configuration level

None of the above solutions are Ideal - The first one recreates functionality already in Loki and requires tables to be configrued in a specific way which is a Loki concern.  The second is not ideal because there's a chance data is written to the table prior to the Lambda running, leaving it at risk until the key is rotated.  The third one would add a generic configuration value that does not appear to be broadly applicable.

**Additional context**
N/A

brettlounsbury:
This change adds a `kms_key_id` configuration to the `storage_config.aws.dynamodb` config.  This is used when creating DynamoDB tables to enable SSE using the specified KMS key.

Currently DynamoDB tables can only be encrypted using Amazon Owned KMS keys, and this has security and compliance drawbacks that make it not usable in some environments

Fixes #8466

**Checklist**
- [X] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/configuration/examples.md
+++ b/docs/sources/configuration/examples.md
@@ -4,9 +4,35 @@ description: Loki Configuration Examples
 ---
  # Examples
 
+## alibaba-cloud-storage-config.yaml
+
+```yaml
+
+# This partial configuration uses Alibaba for chunk storage
+
+schema_config:
+  configs:
+  - from: 2020-05-15
+    object_store: alibabacloud
+    schema: v11
+    index:
+      prefix: loki_index_
+      period: 168h
+
+storage_config:
+  alibabacloud:
+    bucket: <bucket>
+    endpoint: <endpoint>
+    access_key_id: <access_key_id>
+    secret_access_key: <secret_access_key>
+
+```
+
+
 ## 1-Local-Configuration-Example.yaml
 
 ```yaml
+
 auth_enabled: false
 
 server:
@@ -35,6 +61,7 @@ schema_config:
 ## 2-S3-Cluster-Example.yaml
 
 ```yaml
+
 # This is a complete configuration to deploy Loki backed by a s3-Comaptible API
 # like MinIO for storage. Loki components will use memberlist ring to shard and
 # the index will be shipped to storage via boltdb-shipper.
@@ -86,6 +113,7 @@ compactor:
 ## 3-S3-Without-Credentials-Snippet.yaml
 
 ```yaml
+
 # If you don't wish to hard-code S3 credentials you can also configure an EC2
 # instance role by changing the `storage_config` section
 
@@ -109,6 +137,7 @@ storage_config:
 ## 4-BOS-Example.yaml
 
 ```yaml
+
 schema_config:
   configs:
     - from: 2020-05-15
@@ -140,6 +169,7 @@ compactor:
 ## 5-S3-And-DynamoDB-Snippet.yaml
 
 ```yaml
+
 # This partial configuration uses S3 for chunk storage and uses DynamoDB for index storage
 
 schema_config:
@@ -162,6 +192,7 @@ storage_config:
 ## 6-Cassandra-Snippet.yaml
 
 ```yaml
+
 # This is a partial config that uses the local filesystem for chunk storage and Cassandra for index storage
 
 schema_config:
@@ -191,6 +222,7 @@ storage_config:
 ## 7-Schema-Migration-Snippet.yaml
 
 ```yaml
+
 schema_config:
   configs:
     # Starting from 2018-04-15 Loki should store indexes on Cassandra
@@ -219,6 +251,7 @@ schema_config:
 ## 8-GCS-Snippet.yaml
 
 ```yaml
+
 # This partial configuration uses GCS for chunk storage and uses BigTable for index storage
 
 schema_config:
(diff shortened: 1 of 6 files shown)


docs/sources/configuration/examples.md
```diff
@@ -32,6 +32,32 @@ schema_config:
 ```
 
 
+## 10-S3-And-DynamoDB-With-KMS-Snippet.yaml
```
JStickler, line 35: Since this is numbered as "10" is there some reason it is so early in the file and doesn't come after "9"?

brettlounsbury: This file is auto generated by `make generate-example-config-doc` and checked-in as-is.  There was a pipeline step that failed that provided instructions specifically to run that command and checkin the file.  My guess is the sort order isn't treating the first field as a number.

brettlounsbury:
@JStickler 

I'm having problems with the pipeline.  Its telling me to run `make doc` and add that file as well, but it throws out this error generating docs.  I've tried this on my macbook running go-lang 1.19 on on a windows machine using the linux subsystem for windows and go-lang 1.20 and I get the following errors on both.  Any idea whats triggering that?

```
blounsbury@LOUNSBURY:/mnt/c/Users/Lounsbury/Projects/loki$ make doc
go run ./tools/doc-generator docs/sources/configuration/index.template > docs/sources/configuration/_index.md
# github.com/grafana/loki/pkg/configs/userconfig
pkg/configs/userconfig/config.go:313:7: not enough arguments in call to rules.NewAlertingRule
        have (string, parser.Expr, time.Duration, labels.Labels, labels.Labels, nil, string, bool, "github.com/go-kit/log".Logger)
        want (string, parser.Expr, time.Duration, time.Duration, labels.Labels, labels.Labels, labels.Labels, string, bool, "github.com/go-kit/log".Logger)
# github.com/grafana/loki/pkg/querier/series
pkg/querier/series/series_set.go:193:27: cannot use &ConcreteSeries{…} (value of type *ConcreteSeries) as storage.Series value in argument to append: *ConcreteSeries does not implement storage.Series (wrong type for method Iterator)
                have Iterator() chunkenc.Iterator
                want Iterator(chunkenc.Iterator) chunkenc.Iterator
pkg/querier/series/series_set.go:205:27: cannot use &ConcreteSeries{…} (value of type *ConcreteSeries) as storage.Series value in argument to append: *ConcreteSeries does not implement storage.Series (wrong type for method Iterator)
                have Iterator() chunkenc.Iterator
                want Iterator(chunkenc.Iterator) chunkenc.Iterator
pkg/querier/series/series_set.go:239:9: cannot use &DeletedSeries{…} (value of type *DeletedSeries) as storage.Series value in return statement: *DeletedSeries does not implement storage.Series (wrong type for method Iterator)
                have Iterator() chunkenc.Iterator
                want Iterator(chunkenc.Iterator) chunkenc.Iterator
pkg/querier/series/series_set.go:250:34: not enough arguments in call to d.series.Iterator
        have ()
        want (chunkenc.Iterator)
pkg/querier/series/series_set.go:345:9: cannot use emptySeries{…} (value of type emptySeries) as storage.Series value in return statement: emptySeries does not implement storage.Series (wrong type for method Iterator)
                have Iterator() chunkenc.Iterator
                want Iterator(chunkenc.Iterator) chunkenc.Iterator
# github.com/grafana/loki/pkg/ruler/storage/wal
pkg/ruler/storage/wal/wal.go:541:26: cannot use (*appender)(nil) (value of type *appender) as storage.Appender value in variable declaration: *appender does not implement storage.Appender (wrong type for method AppendHistogram)
                have AppendHistogram(storage.SeriesRef, labels.Labels, int64, *histogram.Histogram) (storage.SeriesRef, error)
                want AppendHistogram(storage.SeriesRef, labels.Labels, int64, *histogram.Histogram, *histogram.FloatHistogram) (storage.SeriesRef, error)
make: *** [Makefile:745: doc] Error 1
```

JStickler:
@brettlounsbury I'm still learning about all the automation on this repo, so unfortunately I'm not much help with debugging your error messages.  I'm hoping that someone else on the Loki team can help with debugging.

docs/sources/configuration/examples.md
brettlounsbury: FYI - I cleaned up the `generate-example-config-doc` target in the Makefile.  This now sorts by numerical ordering, and I think its quite a bit more readable now too.

brettlounsbury:
Looks like one of the updates I pulled in resolved the doc generation problem.  I've pushed that file to the repo to let the CI pipeline continue.

MasslessParticle (approved):
Thanks for the contribution! 

Sorry about the tooling but I'm excited you got through it

PR #8483, opened 2023-02-09
Add support for or between pattern parsers

MichelHollands:

The `pattern` parsers can be combined with an `or` statement as in this example:

```
| pattern `=== FAIL <testname>` or pattern `PASS <testname>`
```

Only the captured fields of the first matching pattern are added as labels.

**Checklist**
- [X] Documentation added
- [X] Tests updated
- [x] `CHANGELOG.md` updated


JStickler (approved):
[Docs squad} Documentation LGTM

jeschkies (changes requested):
Thanks for the initiative but I'm working off this should have gone through a LID first. Eg I'd prefer `pattern "===FAIL..." or `===OK ...`.

Maybe some others prefer it to be inline. Furthermore, how many `or`s are we going to support.

jeschkies:
I think I know where this is coming from. LogQL [allows](https://grafana.com/docs/loki/latest/logql/log_queries/) chaining filter expressions. Could we use those instead? Basically extract an `outcome` label compare it. Btw is there a different outcome than FAIL and PASS?

MichelHollands:
Chaining them together will work.

PR #8486, opened 2023-02-09
Fix ruler deadlock

MasslessParticle:
A circular dependency between the Prometheus rules.Manager and the MemStore causes occasional deadlock in the rulers.

This change makes both the manager and the memstore rely on the `GroupLoader` interface to break the circular dependency. The manger holding the `GroupLoader` only ever adds rules to it so we add another layer to cleanup any leftover state after an update happens.

This supersedes https://github.com/grafana/loki/pull/8453


owen-d (approved):
LGTM

PR #8494, opened 2023-02-10
correctly calculate max shards

owen-d:
Fixes a bug introduced in https://github.com/grafana/loki/pull/8487 which had the effect of always enforcing shard factor of 1 when max-shards was intended to be disabled: https://github.com/grafana/loki/pull/8487

### DIFF
--- a/pkg/querier/queryrange/shard_resolver.go
+++ b/pkg/querier/queryrange/shard_resolver.go
@@ -124,10 +124,8 @@ func (r *dynamicShardResolver) Shards(e syntax.Expr) (int, error) {
 	}
 
 	combined := stats.MergeStats(results...)
-	factor := guessShardFactor(combined)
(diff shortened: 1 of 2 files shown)


kavirajk (approved):
LGTM :+1:

pkg/querier/queryrange/shard_resolver.go
```diff
@@ -166,8 +164,21 @@ func guessShardFactor(stats stats.Stats) int {
...
+
+	// shortcut: no need to run any sharding logic when factor=1
+	// as it's the same as no sharding
 	if factor == 1 {
 		factor = 0
 	}
 	return factor
 }
+
+func min(a, b int) int {
```
MichelHollands, line 179: There is already a Min function in pkg/util/math.

PR #8513, opened 2023-02-12
helm: azure config: don't set use_federated_token if false

### LINKED ISSUE #8450: Helm chart: latest versions of the chart (4.0 to 4.5) does not run on Azure with official images
**Describe the bug**

The helm chart defines `use_federated_token` in the config file for Azure in these places:
* https://github.com/grafana/loki/blob/helm-loki-4.5.0/production/helm/loki/templates/_helpers.tpl#L243
* https://github.com/grafana/loki/blob/helm-loki-4.5.0/production/helm/loki/templates/_helpers.tpl#L310

The problem is this parameter is not yet available in existing releases of loki.
It is in `main`, but not yet in `v2.7.3`:

```
[loki]$ git checkout main 
Already on 'main'
Your branch is up to date with 'origin/main'.
[loki]$ grep -r "use_federated_token" *
docs/sources/configuration/_index.md:[use_federated_token: <boolean> | default = false]
pkg/storage/chunk/client/azure/blob_storage_client.go:  UseFederatedToken   bool           `yaml:"use_federated_token"`
production/helm/loki/templates/_helpers.tpl:  use_federated_token: {{ .useFederatedToken }}
production/helm/loki/templates/_helpers.tpl:  use_federated_token: {{ .useFederatedToken }}
[loki]$ 
[loki]$ git checkout main 
Already on 'main'
Your branch is up to date with 'origin/main'.
[loki]$ git pull
Already up to date.
[loki]$ grep -r "use_federated_token" *
docs/sources/configuration/_index.md:[use_federated_token: <boolean> | default = false]
pkg/storage/chunk/client/azure/blob_storage_client.go:  UseFederatedToken   bool           `yaml:"use_federated_token"`
production/helm/loki/templates/_helpers.tpl:  use_federated_token: {{ .useFederatedToken }}
production/helm/loki/templates/_helpers.tpl:  use_federated_token: {{ .useFederatedToken }}
[loki]$ git checkout v2.7.3
Note: switching to 'v2.7.3'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at a880ea3bf Prepare 2.7.3 changelog and release notes (#8348)
[loki]$ grep -r "use_federated_token" *
[loki]$ 
```

So, for this to work, we have to use a custom tag, like `main-5e53303` as defined in [default values for chart 4.5.0](https://github.com/grafana/loki/blob/helm-loki-4.5.0/production/helm/loki/values.yaml#L50).
But if one uses an proper loki release, it won't run.

This has been discussed on [slack](https://grafana.slack.com/archives/CEPJRLQNL/p1675186600386189), and I open this issue as proposed in the discussion.

hervenicol:

Setting `use_federated_token` in azure config crashes with current Loki version (2.7.3). Support for this parameter is in `main` but not released yet.
As `use_federated_token`'s default value will be `false`, let's not set it in config when it's `false`.

This way, the helm chart works on Azure with current Loki versions (if people don't set this parameter), and is compatible with future versions when this parameter will be supported.

Fixes #8450

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/CHANGELOG.md
+++ b/production/helm/loki/CHANGELOG.md
@@ -14,6 +14,7 @@ Entries should include a reference to the pull request that introduced the chang
 [//]: # (<AUTOMATED_UPDATES_LOCATOR> : do not remove this line. Add your changelog bellow this line. This locator is used by CI pipeline to find the place where to put changelog entry.)
 
 - [FEATURE] add `gateway.nginxConfig.customReadUrl`, `gateway.nginxConfig.customWriteUrl` and `gateway.nginxConfig.customBackendUrl` to override read/write/backend paths.
+- [BUGFIX] Azure config: don't set `use_federated_token` when not needed. Avoids crashes when loki does not support this parameter.
 
 ## 4.5.1
 
--- a/production/helm/loki/templates/_helpers.tpl
+++ b/production/helm/loki/templates/_helpers.tpl
@@ -240,7 +240,9 @@ azure:
   {{- end }}
   container_name: {{ $.Values.loki.storage.bucketNames.chunks }}
   use_managed_identity: {{ .useManagedIdentity }}
+  {{- if .useFederatedToken }}
   use_federated_token: {{ .useFederatedToken }}
+  {{- end }}
   {{- with .userAssignedId }}
   user_assigned_id: {{ . }}
   {{- end }}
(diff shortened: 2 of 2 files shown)


chaudum (commented):
While this change fixes the Helm chart for default values, it still breaks if people change the `useFederatedToken` value to `true`, because they don't know that this setting has not been released yet. I don't think that this is a good solution. I would rather remove the setting from the `values.yaml` completely and add it back in, once it's supported by the Loki version that is used by the Helm chart.

PR #8518, opened 2023-02-13
logcli: Add parallel flags

SN9NV:
Requesting a large range, a day, for example, takes a very long time to download. You can download in parallel by starting many processes, but this just made my job so much easier.

Continuation of  #7543

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/cmd/logcli/main.go
+++ b/cmd/logcli/main.go
@@ -80,8 +80,8 @@ the "instant-query" command instead.
 
 Parallelization:
 
-You can download logs in parallel, there are a few flags which control this
-behaviour:
+You can download an unlimited number of logs in parallel, there are a few
+flags which control this behaviour:
 
 	--parallel-duration
 	--parallel-max-workers
@@ -107,7 +107,7 @@ Example:
 
 This example will create a queue of jobs to execute, each being 15 minutes in
 duration. In this case, that means, for the 10-hour total duration, there will
-be forty 15-minute jobs.
+be forty 15-minute jobs. The --limit flag is ignored.
 
 It will start four workers, and they will each take a job to work on from the
 queue until all the jobs have been completed.
@@ -263,6 +263,8 @@ func main() {
 		} else if rangeQuery.ParallelMaxWorkers == 1 {
 			rangeQuery.DoQuery(queryClient, out, *statistics)
 		} else {
+			// `--limit` doesn't make sense when using parallelism.
+			rangeQuery.Limit = 0
 			rangeQuery.DoQueryParallel(queryClient, out, *statistics)
 		}
 	case instantQueryCmd.FullCommand():
@@ -449,7 +451,7 @@ func newQuery(instant bool, cmd *kingpin.CmdClause) *query.Query {
 		cmd.Flag("interval", "Query interval, for log queries. Return entries at the specified interval, ignoring those between. **This parameter is experimental, please see Issue 1779**").DurationVar(&q.Interval)
 		cmd.Flag("batch", "Query batch size to use until 'limit' is reached").Default("1000").IntVar(&q.BatchSize)
 		cmd.Flag("parallel-duration", "Split the range into jobs of this length to download the logs in parallel. This will result in the logs being out of order. Use --part-path-prefix to create a file per job to maintain ordering.").Default("1h").DurationVar(&q.ParallelDuration)
-		cmd.Flag("parallel-max-workers", "Max number of workers to start up for parallel jobs. A value of 1 will not create any parallel workers.").Default("1").IntVar(&q.ParallelMaxWorkers)
+		cmd.Flag("parallel-max-workers", "Max number of workers to start up for parallel jobs. A value of 1 will not create any parallel workers. When using parallel workers, limit is ignored.").Default("1").IntVar(&q.ParallelMaxWorkers)
 		cmd.Flag("part-path-prefix", "When set, each server response will be saved to a file with this prefix. Creates files in the format: 'prefix-utc_start-utc_end.part'. Intended to be used with the parallel-* flags so that you can combine the files to maintain ordering based on the filename. Default is to write to stdout.").StringVar(&q.PartPathPrefix)
 		cmd.Flag("overwrite-completed-parts", "Overwrites completed part files. This will download the range again, and replace the original completed part file. Default will skip a range if it's part file is already downloaded.").Default("false").BoolVar(&q.OverwriteCompleted)
 		cmd.Flag("merge-parts", "Reads the part files in order and writes the output to stdout. Original part files will be deleted with this option.").Default("false").BoolVar(&q.MergeParts)
--- a/cmd/logcli/main.go
+++ b/cmd/logcli/main.go
@@ -428,6 +428,11 @@ func newQuery(instant bool, cmd *kingpin.CmdClause) *query.Query {
 			q.End = mustParse(to, defaultEnd)
 		}
 		q.Quiet = *quiet
+
+		if q.ParallelMaxWorkers < 1 {
+			return fmt.Errorf("parallel-max-workers must be greater than 0")
+		}
+
 		return nil
 	})
 
--- a/cmd/logcli/main.go
+++ b/cmd/logcli/main.go
@@ -376,8 +376,10 @@ func newQuery(instant bool, cmd *kingpin.CmdClause) *query.Query {
 		cmd.Flag("step", "Query resolution step width, for metric queries. Evaluate the query at the specified step over the time range.").DurationVar(&q.Step)
 		cmd.Flag("interval", "Query interval, for log queries. Return entries at the specified interval, ignoring those between. **This parameter is experimental, please see Issue 1779**").DurationVar(&q.Interval)
 		cmd.Flag("batch", "Query batch size to use until 'limit' is reached").Default("1000").IntVar(&q.BatchSize)
-		cmd.Flag("parallel-duration", "Split the range into jobs of this length to download the logs in parallel. This will result in the logs being out of order.").Default("1h").DurationVar(&q.ParallelDuration)
+		cmd.Flag("parallel-duration", "Split the range into jobs of this length to download the logs in parallel. This will result in the logs being out of order. Use --part-file-prefix to create a file per job to maintain ordering.").Default("1h").DurationVar(&q.ParallelDuration)
 		cmd.Flag("parallel-max-workers", "Max number of workers to start up for parallel jobs. A value of 1 will not create any parallel workers.").Default("1").IntVar(&q.ParallelMaxWorkers)
+		cmd.Flag("part-file-prefix", "When set, each server response will be saved to a file with this prefix. Creates files in the format: 'prefix-unix_start-unix_end.part'. Intended to be used with the parallel-* flags so that you can combine the files to maintain ordering based on the filename. Default is to write to stdout.").StringVar(&q.PartFilePrefix)
+		cmd.Flag("overwrite-completed-parts", "Overwrites completed part files. This will download the range again, and replace the original completed part file. Default will skip a range if it's part file is already downloaded.").Default("false").BoolVar(&q.OverwriteCompleted)
 	}
 
 	cmd.Flag("forward", "Scan forwards through logs.").Default("false").BoolVar(&q.Forward)
--- a/cmd/logcli/main.go
+++ b/cmd/logcli/main.go
@@ -76,7 +76,54 @@ data points between the start and end query time. This output is used to
 build graphs, similar to what is seen in the Grafana Explore graph view.
 If you are querying metrics and just want the most recent data point
 (like what is seen in the Grafana Explore table view), then you should use
-the "instant-query" command instead.`)
+the "instant-query" command instead.
+
+Parallelization:
+
+You can download logs in parallel, there are a few flags which control this
+behaviour:
+
+	--parallel-duration
+	--parallel-max-workers
+	--part-file-prefix
+	--overwrite-completed-parts
+	--merge-part-files
+	--keep-part-files
+
+Refer to the help of these specific flags to understand what each of them do.
+
+Example:
+
+	logcli query
+	   --timezone=UTC
+	   --from="2021-01-19T10:00:00Z"
+	   --to="2021-01-19T20:00:00Z"
+	   --output=jsonl
+	   --parallel-duration="15m"
+	   --parallel-max-workers="10"
+	   --part-file-prefix="/tmp/my_query"
+	   --merge-part-files
+	   'my-query'
+
+This will start 10 workers, and they will each start downloading 15 minute
+slices of the specified time range.
+
+Each worker will save a "part" file to the location specified in the prefix.
+Different prefixes can be used to run multiple queries at the same time.
+The timestamp of the start and end of the part is in the file name.
+While the part is being downloaded, the filename will end in ".part", when it
+is complete, the file will be renamed to remove this ".part" extension.
+By default, if a completed part file is found, that part will not be downloaded
+again. This can be overridden with the --overwrite-completed-parts flag.
+
+If you do not specify the --merge-part-files flag, the part files will be
+downloaded, and logcli will exit, and you can process the files as you wish.
+With the flag specified, the part files will be read in order, and the output
+printed to the terminal. The lines will be printed as soon as the next part is
+complete, you don't have to wait for all the parts to download before getting
+output. --merge-part-files will remove the part files when it is done reading
+each of them, to change this, you can add --keep-part-files and the part file
+wParallelizationill not be removed.`)
 	rangeQuery = newQuery(false, queryCmd)
 	tail       = queryCmd.Flag("tail", "Tail the logs").Short('t').Default("false").Bool()
 	follow     = queryCmd.Flag("follow", "Alias for --tail").Short('f').Default("false").Bool()
@@ -380,6 +427,8 @@ func newQuery(instant bool, cmd *kingpin.CmdClause) *query.Query {
 		cmd.Flag("parallel-max-workers", "Max number of workers to start up for parallel jobs. A value of 1 will not create any parallel workers.").Default("1").IntVar(&q.ParallelMaxWorkers)
 		cmd.Flag("part-file-prefix", "When set, each server response will be saved to a file with this prefix. Creates files in the format: 'prefix-unix_start-unix_end.part'. Intended to be used with the parallel-* flags so that you can combine the files to maintain ordering based on the filename. Default is to write to stdout.").StringVar(&q.PartFilePrefix)
 		cmd.Flag("overwrite-completed-parts", "Overwrites completed part files. This will download the range again, and replace the original completed part file. Default will skip a range if it's part file is already downloaded.").Default("false").BoolVar(&q.OverwriteCompleted)
+		cmd.Flag("merge-part-files", "Reads the part files in order and writes the output to stdout. Original part files will be deleted with this option.").Default("false").BoolVar(&q.MergePartFiles)
+		cmd.Flag("keep-part-files", "Overrides the default behaviour of --merge-part-files which will delete the part files once all the files have been read. This option will keep the part files.").Default("false").BoolVar(&q.KeepPartFiles)
 	}
 
 	cmd.Flag("forward", "Scan forwards through logs.").Default("false").BoolVar(&q.Forward)
--- a/cmd/logcli/main.go
+++ b/cmd/logcli/main.go
@@ -193,8 +193,10 @@ func main() {
 
 		if *tail || *follow {
 			rangeQuery.TailQuery(time.Duration(*delayFor)*time.Second, queryClient, out)
-		} else {
+		} else if rangeQuery.ParallelMaxWorkers == 1 {
 			rangeQuery.DoQuery(queryClient, out, *statistics)
+		} else {
+			rangeQuery.DoQueryParallel(queryClient, out, *statistics)
 		}
 	case instantQueryCmd.FullCommand():
 		location, err := time.LoadLocation(*timezone)
@@ -374,7 +376,8 @@ func newQuery(instant bool, cmd *kingpin.CmdClause) *query.Query {
 		cmd.Flag("step", "Query resolution step width, for metric queries. Evaluate the query at the specified step over the time range.").DurationVar(&q.Step)
 		cmd.Flag("interval", "Query interval, for log queries. Return entries at the specified interval, ignoring those between. **This parameter is experimental, please see Issue 1779**").DurationVar(&q.Interval)
 		cmd.Flag("batch", "Query batch size to use until 'limit' is reached").Default("1000").IntVar(&q.BatchSize)
-
+		cmd.Flag("parallel-duration", "Split the range into jobs of this length to download the logs in parallel. This will result in the logs being out of order.").Default("1h").DurationVar(&q.ParallelDuration)
+		cmd.Flag("parallel-max-workers", "Max number of workers to start up for parallel jobs. A value of 1 will not create any parallel workers.").Default("1").IntVar(&q.ParallelMaxWorkers)
 	}
 
 	cmd.Flag("forward", "Scan forwards through logs.").Default("false").BoolVar(&q.Forward)
--- a/cmd/logcli/main.go
+++ b/cmd/logcli/main.go
@@ -85,7 +85,7 @@ behaviour:
 
 	--parallel-duration
 	--parallel-max-workers
-	--part-prefix
+	--part-path-prefix
 	--overwrite-completed-parts
 	--merge-parts
 	--keep-parts
@@ -101,7 +101,7 @@ Example:
 	   --output=jsonl
 	   --parallel-duration="15m"
 	   --parallel-max-workers="4"
-	   --part-prefix="/tmp/my_query"
+	   --part-path-prefix="/tmp/my_query"
 	   --merge-parts
 	   'my-query'
 
@@ -443,10 +443,10 @@ func newQuery(instant bool, cmd *kingpin.CmdClause) *query.Query {
 		cmd.Flag("step", "Query resolution step width, for metric queries. Evaluate the query at the specified step over the time range.").DurationVar(&q.Step)
 		cmd.Flag("interval", "Query interval, for log queries. Return entries at the specified interval, ignoring those between. **This parameter is experimental, please see Issue 1779**").DurationVar(&q.Interval)
 		cmd.Flag("batch", "Query batch size to use until 'limit' is reached").Default("1000").IntVar(&q.BatchSize)
-		cmd.Flag("parallel-duration", "Split the range into jobs of this length to download the logs in parallel. This will result in the logs being out of order. Use --part-prefix to create a file per job to maintain ordering.").Default("1h").DurationVar(&q.ParallelDuration)
+		cmd.Flag("parallel-duration", "Split the range into jobs of this length to download the logs in parallel. This will result in the logs being out of order. Use --part-path-prefix to create a file per job to maintain ordering.").Default("1h").DurationVar(&q.ParallelDuration)
 		cmd.Flag("parallel-max-workers", "Max number of workers to start up for parallel jobs. A value of 1 will not create any parallel workers.").Default("1").IntVar(&q.ParallelMaxWorkers)
-		cmd.Flag("part-prefix", "When set, each server response will be saved to a file with this prefix. Creates files in the format: 'prefix-unix_start-unix_end.part'. Intended to be used with the parallel-* flags so that you can combine the files to maintain ordering based on the filename. Default is to write to stdout.").StringVar(&q.PartPrefix)
-		cmd.Flag("overwrite-completed-parts", "Overwrites completed part files. This will download the range again, and replace the original completed part file. Default will skip a range if its part file is already downloaded.").Default("false").BoolVar(&q.OverwriteCompleted)
+		cmd.Flag("part-path-prefix", "When set, each server response will be saved to a file with this prefix. Creates files in the format: 'prefix-utc_start-utc_end.part'. Intended to be used with the parallel-* flags so that you can combine the files to maintain ordering based on the filename. Default is to write to stdout.").StringVar(&q.PartPathPrefix)
+		cmd.Flag("overwrite-completed-parts", "Overwrites completed part files. This will download the range again, and replace the original completed part file. Default will skip a range if it's part file is already downloaded.").Default("false").BoolVar(&q.OverwriteCompleted)
 		cmd.Flag("merge-parts", "Reads the part files in order and writes the output to stdout. Original part files will be deleted with this option.").Default("false").BoolVar(&q.MergeParts)
 		cmd.Flag("keep-parts", "Overrides the default behaviour of --merge-parts which will delete the part files once all the files have been read. This option will keep the part files.").Default("false").BoolVar(&q.KeepParts)
 	}
--- a/cmd/logcli/main.go
+++ b/cmd/logcli/main.go
@@ -90,7 +90,7 @@ behaviour:
 	--merge-parts
 	--keep-parts
 
-Refer to the help of these specific flags to understand what each of them do.
+Refer to the help for each flag for details about what each of them do.
 
 Example:
 
@@ -100,15 +100,19 @@ Example:
 	   --to="2021-01-19T20:00:00Z"
 	   --output=jsonl
 	   --parallel-duration="15m"
-	   --parallel-max-workers="10"
+	   --parallel-max-workers="4"
 	   --part-prefix="/tmp/my_query"
 	   --merge-parts
 	   'my-query'
 
-This will start 10 workers, and they will each start downloading 15 minute
-slices of the specified time range.
+This example will create a queue of jobs to execute, each being 15 minutes in
+duration. In this case, that means, for the 10-hour total duration, there will
+be forty 15-minute jobs.
 
-Each worker will save a "part" file to the location specified in the prefix.
+It will start four workers, and they will each take a job to work on from the
+queue until all the jobs have been completed.
+
+Each job will save a "part" file to the location specified by the --part-path-prefix.
 Different prefixes can be used to run multiple queries at the same time.
 The timestamp of the start and end of the part is in the file name.
 While the part is being downloaded, the filename will end in ".part", when it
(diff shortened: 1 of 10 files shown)


SN9NV:
@jeschkies

Hi, I have updated the code and made a new PR.

Basically what is new here is that there's two new flags:

- `merge-parts`
- `keep-parts`

What this does is it will read the parts in order and write the output to stdout. And it will do this in order. So it will act similar to the non-parallel flags, and then after each file has been output, it will delete the file.

`keep-parts` overrides this behaviour, keeping the part files instead of deleting them in case the user wants to keep the files.

cmd/logcli/main.go
```diff
@@ -76,7 +76,54 @@ data points between the start and end query time. This output is used to
...
+	   --to="2021-01-19T20:00:00Z"
+	   --output=jsonl
+	   --parallel-duration="15m"
+	   --parallel-max-workers="10"
+	   --part-prefix="/tmp/my_query"
+	   --merge-parts
+	   'my-query'
+
+This will start 10 workers, and they will each start downloading 15 minute
+slices of the specified time range.
```
jeschkies, line 109: They won't finish after the first part though, right? I mean if we need more than ten 15m slices they would continue downloading.

cmd/logcli/main.go
```diff
@@ -76,7 +76,54 @@ data points between the start and end query time. This output is used to
...
+the "instant-query" command instead.
+
+Parallelization:
+
+You can download logs in parallel, there are a few flags which control this
+behaviour:
+
+	--parallel-duration
+	--parallel-max-workers
+	--part-prefix
```
jeschkies, line 88: Hm, I think it should be `part-path-prefix` or something that indicates that this is a path. Your documentation is great, though.

cmd/logcli/main.go
```diff
@@ -76,7 +76,54 @@ data points between the start and end query time. This output is used to
...
+	   --merge-parts
+	   'my-query'
+
+This will start 10 workers, and they will each start downloading 15 minute
+slices of the specified time range.
+
+Each worker will save a "part" file to the location specified in the prefix.
+Different prefixes can be used to run multiple queries at the same time.
+The timestamp of the start and end of the part is in the file name.
+While the part is being downloaded, the filename will end in ".part", when it
```
jeschkies, line 114: Could you add an example.

cmd/logcli/main.go
```diff
@@ -76,7 +76,54 @@ data points between the start and end query time. This output is used to
...
+again. This can be overridden with the --overwrite-completed-parts flag.
+
+If you do not specify the --merge-parts flag, the part files will be
+downloaded, and logcli will exit, and you can process the files as you wish.
+With the flag specified, the part files will be read in order, and the output
+printed to the terminal. The lines will be printed as soon as the next part is
+complete, you don't have to wait for all the parts to download before getting
+output. --merge-parts will remove the part files when it is done reading
+each of them, to change this, you can add --keep-parts and the part file
+wParallelizationill not be removed.`)
```
jeschkies, line 126:
```suggestion
will not be removed.`)
```

pkg/logcli/query/part_file.go
```diff
@@ -0,0 +1,106 @@
...
+	if err := f.fd.Close(); err != nil {
+		return fmt.Errorf("filed to close part file: %s: %s", filename, err)
+	}
+
+	f.fd = nil
+	return nil
+}
+
+// Complete closes the temporary file, and renames it to the complete file name.
+func (f *PartFile) Complete() error {
```
jeschkies, line 91: I think this should call `f.Close`.

pkg/logcli/query/query.go
```diff
@@ -157,9 +195,195 @@ func (q *Query) DoQuery(c client.Client, out output.LogOutput, statistics bool)
...
+	}
+}
+
+func (q *Query) outputFilename() string {
+	return fmt.Sprintf("%s-%d-%d.part", q.PartPrefix, q.Start.Unix(), q.End.Unix())
+}
+
+// createPartFile returns a PartFile if the PartFilePrefix is set.
+// The bool value shows if the part file already exists, and this range should be skipped.
+func (q *Query) createPartFile() (*PartFile, bool) {
```
jeschkies, line 214: It was a little confusing to me when I read that we always use a part file. What do you think about `createOptionalPartFile`?

pkg/logcli/query/query.go
```diff
@@ -157,9 +195,195 @@ func (q *Query) DoQuery(c client.Client, out output.LogOutput, statistics bool)
...
+	if err := partFile.CreateTemp(); err != nil {
+		log.Fatalf("Query failed: %s\n", err)
+	}
+
+	return partFile, false
+}
+
+// rounds up duration d by the multiple m, and then divides by m.
+func durationCeilDiv(d, m time.Duration) int64 {
+	return int64((d + m - 1) / m)
```
jeschkies, line 243: I'm not sure what it does. Would math.Ceil be better here?

pkg/logcli/query/query.go
```diff
@@ -157,9 +195,195 @@ func (q *Query) DoQuery(c client.Client, out output.LogOutput, statistics bool)
...
+func (q *Query) mergeJobs(jobs []*parallelJob) error {
+	if !q.MergeParts {
+		return nil
+	}
+
+	for _, job := range jobs {
+		// wait for the next job to finish
+		<-job.done
+
+		f, err := os.Open(job.q.outputFilename())
```
jeschkies, line 306: Do we have to close the file again?

pkg/logcli/query/query.go
```diff
@@ -157,9 +195,195 @@ func (q *Query) DoQuery(c client.Client, out output.LogOutput, statistics bool)
...
+	statistics bool,
+) *sync.WaitGroup {
+	wg := sync.WaitGroup{}
+	jobsChan := make(chan *parallelJob, len(jobs))
+
+	// Queue up the jobs
+	for _, job := range jobs {
+		jobsChan <- job
+	}
+	close(jobsChan)
```
jeschkies, line 342: This might be a micro optimization but I'd avoid allocating a queue here
```go
jobsChan := make(chan *parallelJob)
go func() {
    for _, job := range jobs {
		jobsChan <- job
	}
	close(jobsChan)
}()
```

pkg/logcli/query/query.go
```diff
@@ -157,9 +195,195 @@ func (q *Query) DoQuery(c client.Client, out output.LogOutput, statistics bool)
...
+
+func newParallelJob(q *Query) *parallelJob {
+	return &parallelJob{
+		q:    q,
+		done: make(chan struct{}),
+	}
+}
 
+func (j *parallelJob) run(c client.Client, out output.LogOutput, statistics bool) {
+	j.q.DoQuery(c, out, statistics)
```
jeschkies, line 270: I'm wondering if you can miss some log entries if the time range has more logs than the configured limit.

jeschkies: Ok. I've finally got the answer. Say the limit is 1000.
- if the first job fetches 1000 lines it's done and all remaining jobs can be canceled
- if the job gets less than 1000 lines the next need to continue

docs/sources/tools/logcli.md
```diff
@@ -250,26 +250,22 @@ usage: logcli query [<flags>] <query>
...
 
-By default we look over the last hour of data; use --since to modify or provide
-specific start and end times with --from and --to respectively.
+By default we look over the last hour of data; use --since to modify or provide specific start and end times with --from and --to
+respectively.
 
-Notice that when using --from and --to then ensure to use RFC3339Nano time
-format, but without timezone at the end. The local timezone will be added
-automatically or if using --timezone flag.
+Notice that when using --from and --to then ensure to use RFC3339Nano time format, but without timezone at the end. The local timezone will
```
JStickler, line 267:
```suggestion
Note that when using --from and --to you must use the RFC3339Nano time format, but without timezone at the end. The local timezone will
```

docs/sources/tools/logcli.md
```diff
@@ -250,26 +250,22 @@ usage: logcli query [<flags>] <query>
...
+be added automatically or if using --timezone flag.
```
JStickler, line 268:
```suggestion
be added automatically or when using the --timezone flag.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+You can download logs in parallel, there are a few flags which control this behaviour:
+
+  --parallel-duration
+  --parallel-max-workers
+  --part-prefix
+  --overwrite-completed-parts
+  --merge-parts
+  --keep-parts
+
+Refer to the help of these specific flags to understand what each of them do.
```
JStickler, line 296:
```suggestion
Refer to the help for each flag for details about what each of them do.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+     --from="2021-01-19T10:00:00Z"
+     --to="2021-01-19T20:00:00Z"
+     --output=jsonl
+     --parallel-duration="15m"
+     --parallel-max-workers="10"
+     --part-prefix="/tmp/my_query"
+     --merge-parts
+     'my-query'
+
+This will start 10 workers, and they will each start downloading 15 minute slices of the specified time range.
```
JStickler, line 311:
```suggestion
This example will start 10 workers, and they will each start downloading 15 minute slices of the specified time range.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+
+Each worker will save a "part" file to the location specified in the prefix. Different prefixes can be used to run multiple queries at the
+same time. The timestamp of the start and end of the part is in the file name. While the part is being downloaded, the filename will end
+in ".part", when it is complete, the file will be renamed to remove this ".part" extension. By default, if a completed part file is found,
+that part will not be downloaded again. This can be overridden with the --overwrite-completed-parts flag.
+
+If you do not specify the --merge-parts flag, the part files will be downloaded, and logcli will exit, and you can process the files as you
+wish. With the flag specified, the part files will be read in order, and the output printed to the terminal. The lines will be printed as
+soon as the next part is complete, you don't have to wait for all the parts to download before getting output. --merge-parts will remove
```
JStickler, line 320:
```suggestion
soon as the next part is complete, you don't have to wait for all the parts to download before getting output. the --merge-parts flag will remove
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+the part files when it is done reading each of them, to change this, you can add --keep-parts and the part file wParallelizationill not be
```
JStickler, line 321:
```suggestion
the part files when it is done reading each of them. To change this, you can use the  --keep-parts flag, and the part file with parallelization will not be
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
-                              retrieved using the configured storage in the
-                              given Loki configuration file.
-      --colored-output        Show output with colored labels
-  -t, --tail                  Tail the logs
-  -f, --follow                Alias for --tail
-      --delay-for=0           Delay in tailing by number of seconds to
-                              accumulate logs for re-ordering
+      --help                     Show context-sensitive help (also try --help-long and --help-man).
+      --version                  Show application version.
+  -q, --quiet                    Suppress query metadata
```
JStickler, line 327:
```suggestion
  -q, --quiet                    Suppress query metadata.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --stats                    Show query statistics
```
JStickler, line 328:
```suggestion
      --stats                    Show query statistics.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+  -o, --output=default           Specify output mode [default, raw, jsonl]. raw suppresses log labels and timestamp.
+  -z, --timezone=Local           Specify the timezone to use when formatting output timestamps [Local, UTC]
```
JStickler, line 330:
```suggestion
  -z, --timezone=Local           Specify the timezone to use when formatting output timestamps [Local, UTC].
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --cpuprofile=""            Specify the location for writing a CPU profile.
+      --memprofile=""            Specify the location for writing a memory profile.
+      --stdin                    Take input logs from stdin
```
JStickler, line 333:
```suggestion
      --stdin                    Take input logs from stdin.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --addr="http://localhost:3100"
+                                 Server address. Can also be set using LOKI_ADDR env var.
```
JStickler, line 335:
```suggestion
                                 Server address. Can also be set using LOKI_ADDR environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --username=""              Username for HTTP basic auth. Can also be set using LOKI_USERNAME env var.
```
JStickler, line 336:
```suggestion
      --username=""              Username for HTTP basic authentication. Can also be set using LOKI_USERNAME environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --password=""              Password for HTTP basic auth. Can also be set using LOKI_PASSWORD env var.
```
JStickler, line 337:
```suggestion
      --password=""              Password for HTTP basic authentication. Can also be set using LOKI_PASSWORD environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --ca-cert=""               Path to the server Certificate Authority. Can also be set using LOKI_CA_CERT_PATH env var.
```
JStickler, line 338:
```suggestion
      --ca-cert=""               Path to the server Certificate Authority. Can also be set using LOKI_CA_CERT_PATH environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --tls-skip-verify          Server certificate TLS skip verify. Can also be set using LOKI_TLS_SKIP_VERIFY env var.
```
JStickler, line 339:
```suggestion
      --tls-skip-verify          Server certificate TLS skip verify. Can also be set using LOKI_TLS_SKIP_VERIFY environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --cert=""                  Path to the client certificate. Can also be set using LOKI_CLIENT_CERT_PATH env var.
```
JStickler, line 340:
```suggestion
      --cert=""                  Path to the client certificate. Can also be set using LOKI_CLIENT_CERT_PATH environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --key=""                   Path to the client certificate key. Can also be set using LOKI_CLIENT_KEY_PATH env var.
```
JStickler, line 341:
```suggestion
      --key=""                   Path to the client certificate key. Can also be set using LOKI_CLIENT_KEY_PATH environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --org-id=""                adds X-Scope-OrgID to API requests for representing tenant ID. Useful for requesting tenant data when
+                                 bypassing an auth gateway. Can also be set using LOKI_ORG_ID env var.
```
JStickler, line 343:
```suggestion
                                 bypassing an authentication gateway. Can also be set using LOKI_ORG_ID environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --query-tags=""            adds X-Query-Tags http header to API requests. This header value will be part of `metrics.go` statistics.
+                                 Useful for tracking the query. Can also be set using LOKI_QUERY_TAGS env var.
```
JStickler, line 345:
```suggestion
                                 Useful for tracking the query. Can also be set using LOKI_QUERY_TAGS environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --bearer-token=""          adds the Authorization header to API requests for authentication purposes. Can also be set using
+                                 LOKI_BEARER_TOKEN env var.
```
JStickler, line 347:
```suggestion
                                 LOKI_BEARER_TOKEN environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --bearer-token-file=""     adds the Authorization header to API requests for authentication purposes. Can also be set using
+                                 LOKI_BEARER_TOKEN_FILE env var.
```
JStickler, line 349:
```suggestion
                                 LOKI_BEARER_TOKEN_FILE environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --retries=0                How many times to retry each query when getting an error response from Loki. Can also be set using
+                                 LOKI_CLIENT_RETRIES env var.
```
JStickler, line 351:
```suggestion
                                 LOKI_CLIENT_RETRIES environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --min-backoff=0            Minimum backoff time between retries. Can also be set using LOKI_CLIENT_MIN_BACKOFF env var.
```
JStickler, line 352:
```suggestion
      --min-backoff=0            Minimum backoff time between retries. Can also be set using LOKI_CLIENT_MIN_BACKOFF environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --max-backoff=0            Maximum backoff time between retries. Can also be set using LOKI_CLIENT_MAX_BACKOFF env var.
```
JStickler, line 353:
```suggestion
      --max-backoff=0            Maximum backoff time between retries. Can also be set using LOKI_CLIENT_MAX_BACKOFF environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --auth-header="Authorization"
+                                 The authorization header used. Can also be set using LOKI_AUTH_HEADER env var.
```
JStickler, line 355:
```suggestion
                                 The authorization header used. Can also be set using LOKI_AUTH_HEADER environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --proxy-url=""             The http or https proxy to use when making requests. Can also be set using LOKI_HTTP_PROXY_URL env var.
```
JStickler, line 356:
```suggestion
      --proxy-url=""             The http or https proxy to use when making requests. Can also be set using LOKI_HTTP_PROXY_URL environment variable.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --limit=30                 Limit on number of entries to print.
+      --since=1h                 Lookback window.
+      --from=FROM                Start looking for logs at this absolute time (inclusive)
```
JStickler, line 359:
```suggestion
      --from=FROM                Start looking for logs at this absolute time (inclusive).
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --to=TO                    Stop looking for logs at this absolute time (exclusive)
```
JStickler, line 360:
```suggestion
      --to=TO                    Stop looking for logs at this absolute time (exclusive).
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --step=STEP                Query resolution step width, for metric queries. Evaluate the query at the specified step over the time
+                                 range.
+      --interval=INTERVAL        Query interval, for log queries. Return entries at the specified interval, ignoring those between. **This
+                                 parameter is experimental, please see Issue 1779**
+      --batch=1000               Query batch size to use until 'limit' is reached
```
JStickler, line 365:
```suggestion
      --batch=1000               Query batch size to use until 'limit' is reached.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --parallel-duration=1h     Split the range into jobs of this length to download the logs in parallel. This will result in the logs
+                                 being out of order. Use --part-prefix to create a file per job to maintain ordering.
+      --parallel-max-workers=1   Max number of workers to start up for parallel jobs. A value of 1 will not create any parallel workers.
```
JStickler, line 368: Question - What happens with a value of 0?

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --part-prefix=PART-PREFIX  When set, each server response will be saved to a file with this prefix. Creates files in the format:
+                                 'prefix-unix_start-unix_end.part'. Intended to be used with the parallel-* flags so that you can combine
+                                 the files to maintain ordering based on the filename. Default is to write to stdout.
+      --overwrite-completed-parts
+                                 Overwrites completed part files. This will download the range again, and replace the original completed
+                                 part file. Default will skip a range if it's part file is already downloaded.
```
JStickler, line 374:
```suggestion
                                 part file. Default will skip a range if its part file is already downloaded.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --merge-parts              Reads the part files in order and writes the output to stdout. Original part files will be deleted with
+                                 this option.
+      --keep-parts               Overrides the default behaviour of --merge-parts which will delete the part files once all the files have
+                                 been read. This option will keep the part files.
+      --forward                  Scan forwards through logs.
+      --no-labels                Do not print any labels
```
JStickler, line 380:
```suggestion
      --no-labels                Do not print any labels.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --exclude-label=EXCLUDE-LABEL ...
+                                 Exclude labels given the provided key during output.
+      --include-label=INCLUDE-LABEL ...
+                                 Include labels given the provided key during output.
+      --labels-length=0          Set a fixed padding to labels
```
JStickler, line 385:
```suggestion
      --labels-length=0          Set a fixed padding to labels.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --store-config=""          Execute the current query using a configured storage from a given Loki configuration file.
+      --remote-schema            Execute the current query using a remote schema retrieved using the configured storage in the given Loki
+                                 configuration file.
+      --colored-output           Show output with colored labels
```
JStickler, line 389:
```suggestion
      --colored-output           Show output with colored labels.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+  -t, --tail                     Tail the logs
```
JStickler, line 390:
```suggestion
  -t, --tail                     Tail the logs.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+  -f, --follow                   Alias for --tail
```
JStickler, line 391:
```suggestion
  -f, --follow                   Alias for --tail.
```

docs/sources/tools/logcli.md
```diff
@@ -282,101 +278,118 @@ Example:
...
+      --delay-for=0              Delay in tailing by number of seconds to accumulate logs for re-ordering
```
JStickler, line 392:
```suggestion
      --delay-for=0              Delay in tailing by number of seconds to accumulate logs for re-ordering.
```

JStickler (commented):
[Docs squad]  Lots of missing punctuation and some rewording suggestions.

cmd/logcli/main.go
SN9NV: I'm not really sure what you are suggesting here. It's a pretty simple worker pool implementation. A worker will pick up the next job when it's done.

docs/sources/tools/logcli.md
SN9NV: @JStickler Hi, what you are commenting on here is generated, it's not changes that I made. It is output of the help text of the application. A few are from new stuff that I added and I will change that, do you want me to change all of these things? Maybe it would be better to make a PR yourself to change it? I feel it's a bit out of the scope of my contribution here. IDK what do you think?

SN9NV (commented):
@jeschkies I have made a few comments on some of your comments and will hopefully work on some of your suggestions tomorrow.

JStickler:
>what you are commenting on here is generated, it's not changes that I made. ... I feel it's a bit out of the scope of my contribution here.
@SN9NV I'll leave that up to you.  I can make a PR later if that's easier for you.

docs/sources/tools/logcli.md
SN9NV: It would not start up any workers, and exit. I guess it would be pretty stupid to set this value. I will see if there's any validation done on the parameters and add some error messages there if such a value is set.

pkg/logcli/query/query.go
jeschkies: Fair enough. Maybe leave a comment so that the after world will know 😂

cmd/logcli/main.go
jeschkies: I probably phrased it weird. I meant that the doc should sind the misconception that only ten works process then jobs. But maybe that's just me.

pkg/logcli/query/part_file.go
jeschkies: I've missed that 🙈

pkg/logcli/query/query.go
SN9NV: I have moved the conditional creation to the caller, so this is much clearer now.

pkg/logcli/query/query.go
SN9NV: I changed the name here, hopefully it's clearer.

SN9NV:
@JStickler I have added some more docs, can you please check the language again. I believe I have solved all your previous comments, but it's possible I missed one because there were so many. `cmd/logcli/main.go` would be the right file to check for the documentation comments. Thank you.

@jeschkies I have pushed a few commits, I believe everything should be good now.

I reworked the part file to use some better names, I didn't like `Complete`, it didn't really feel like a command to me, `Finalize` feels much better.

cmd/logcli/main.go
SN9NV: I reworked the docs here, so I explain the whole process now. Hopefully it's as clear as it can be.

pkg/logcli/query/query.go
jeschkies: `DoQuery` is always [limited](https://github.com/grafana/loki/blob/main/pkg/logcli/query/query.go#L52) by [--limit](https://github.com/grafana/loki/blob/main/cmd/logcli/main.go#L365). Say we have the limit 100 and a query for an hour and thus four 15 minutes jobs.

If the first job reaches 100 entries the processing should stop. Right now this implementation will return more entries a the other jobs will return up to 100 entries as well. They might also have gaps as the following example should illustrate.

Say we time ranges 12:00-12:15, 12:15-12:30, 12:30-12:45 and 12:45-13:00. End is exclusive. If 12:00-12:15 has actually 150 entries `DoQuery` will only process the first 100. The job for 12:15-12:30 will not includes these.

jeschkies: I suggest we abide to the limit and follow up with `--limit=-1` or something similar to disable it.

jeschkies:
@SN9NV thanks for your effort. We are close and just need to make a decision for https://github.com/grafana/loki/pull/8518#discussion_r1107077743.

pkg/logcli/query/query.go
SN9NV: Right. The reason this worked for me is because I was setting the limit to something really high. I did this long ago and forgot why I did that.

SN9NV: I've pushed a commit, it's using the same solution used [here](https://github.com/grafana/loki/blob/main/cmd/logcli/main.go#L172)

SN9NV: I think a proper way to illustrate an infinite limit would be to have a struct with an `Infinite/NegativeInfinite bool` parameters with `GreaterThan` and `LessThan` functions taking the infinite values into account. Kind of similar to how SQL handles this. But it means refactoring all the usages of `Limit`, and testing it. That could be a bit of work. Will someone have more than 2^32 log lines? I mean, maybe it's possible. No one thought ipv4 would fill up when they made it.

jeschkies: > I've pushed a commit, it's using the same solution used here

That is a great solution. I agree that the limit won't make much sense with parallel execution. Please add the documentation I've proposed.

cmd/logcli/main.go
```diff
@@ -76,7 +76,74 @@ data points between the start and end query time. This output is used to
...
+	   --output=jsonl
+	   --parallel-duration="15m"
+	   --parallel-max-workers="4"
+	   --part-path-prefix="/tmp/my_query"
+	   --merge-parts
+	   'my-query'
+
+This example will create a queue of jobs to execute, each being 15 minutes in
+duration. In this case, that means, for the 10-hour total duration, there will
+be forty 15-minute jobs.
```
jeschkies, line 110:
```suggestion
be forty 15-minute jobs. The limit option will be ignored.
```

cmd/logcli/main.go
```diff
@@ -76,7 +76,74 @@ data points between the start and end query time. This output is used to
 build graphs, similar to what is seen in the Grafana Explore graph view.
 If you are querying metrics and just want the most recent data point
 (like what is seen in the Grafana Explore table view), then you should use
-the "instant-query" command instead.`)
+the "instant-query" command instead.
+
+Parallelization:
+
+You can download logs in parallel, there are a few flags which control this
```
jeschkies, line 83:
```suggestion
You can download unlimited logs in parallel, there are a few flags which control this
```

jeschkies (changes requested):
Awesome work. Please address my last comment about documenting the limit and we are good to go from my side.

I'll follow up with a "proper" unlimited option as this will require more discussion.

pkg/logcli/query/query.go
SN9NV: Sure thing. I will update the docs tomorrow.

cmd/logcli/main.go
```diff
@@ -193,8 +260,12 @@ func main() {
 
 		if *tail || *follow {
 			rangeQuery.TailQuery(time.Duration(*delayFor)*time.Second, queryClient, out)
-		} else {
+		} else if rangeQuery.ParallelMaxWorkers == 1 {
 			rangeQuery.DoQuery(queryClient, out, *statistics)
+		} else {
+			// `--limit` doesn't make sense when using parallelism.
+			rangeQuery.Limit = math.MaxInt
```
jeschkies, line 267:
```suggestion
			rangeQuery.Limit = 0
```

Checkout the changes in https://github.com/grafana/loki/pull/8537

SN9NV:
Cool, I think everything should be done.

jeschkies (approved):
🎉

jeschkies:
@SN9NV I've just ran it on my machine and found the parallel execution to be slower than the normal execution. I'll take a closer look tomorrow.

SN9NV:
That is strange. For me it was definitely faster, a download which took many hours, like over night, took just tens of minutes with the parallelism. And I could see the time it took to download a single part is the time it took to download the same period sequentially. I did have autoscaling in Kubernetes. Maybe it is a problem if you just have one instance. IDK. It would add one or two more pods, but for sure, it was not scaling up to the same degree as the speedup if that makes sense.

SN9NV:
Hmm, I checked the last few rebase/commits and it seemed like it was still working as expected. I was starting 4 workers with 5-minute jobs in my tests, but I also downloaded about 25GB of logs on Tuesday and it worked well with 24 workers, 30-minute jobs, so I think it was working for me at least with my setup. I can test master again tomorrow.

pkg/logcli/query/query.go
```diff
@@ -160,11 +205,203 @@ func (q *Query) DoQuery(c client.Client, out output.LogOutput, statistics bool)
...
+func newParallelJob(q *Query) *parallelJob {
+	return &parallelJob{
+		q:    q,
+		done: make(chan struct{}),
+	}
+}
+
+func (j *parallelJob) run(c client.Client, out output.LogOutput, statistics bool) {
+	j.q.DoQuery(c, out, statistics)
+	j.done <- struct{}{}
```
jeschkies, line 282: @SN9NV this is the deadlock. An unbuffered channel will write into the target when read. That means it blocks here until the channel is read. However, we don't want to wait for that. See my fix https://github.com/grafana/loki/pull/8553.

SN9NV: I just came here to report the same thing, but you beat me to it. :joy:

I was testing with very large number of threads, so this was kind of less noticeable. Smaller numbers makes the problem much more visible. So yeah, good catch.

dannykopping:
@SN9NV @jeschkies this feature seems to only be implemented for range queries in logcli, and as a consequence has broken instant queries:

```bash
$ go run ./cmd/logcli/main.go --addr="http://localhost:3100" --org-id="docker" instant-query 'sum by (method) (count_over_time({job="generated-logs"} | json[1m]))'
main: error: parallel-max-workers must be greater than 0, try --help
exit status 1
```

Is there any reason why this should _not_ apply to instant queries?

SN9NV:
@dannykopping thanks for the report.

That indeed is not ideal that it's showing an error message for an unrelated command.

Why it was not created for instant queries, I only understand range query, but if it's useful, then we can have a look at adding the feature for this command as well.

I will have a fix soon. I think setting to 1 will be a better idea instead of an error stopping execution.

dannykopping:
@SN9NV thanks for the quick response.

As far as I understand, this feature mainly deals with downloading large volumes of logs returned by a query. Thinking about it more, if that is the case, we don't need to handle instant queries I _guess_ because it's very rare that one might have a large volume of logs for the same instant.

Apart from the instant query question above:
The functionality as implemented seems to break down when range aggregations are performed - with the responses not being returned in a coherent manner. The results are ostensibly not recombined in a way that mimicks the existing behaviour.

I was also a little disappointed to see that this was merged without tests which validate that the responses returned are identical with and without parallelization, which would be a reasonable expectation for a user to have; parallelization should not change the output IMHO since it's just a performance optimization.

SN9NV:
@dannykopping can you please provide a query which is not returning correct results? I tested quite a few queries for logs which I have had to download in the past and the parallel and standard methods produced exactly the same files in the output.

I didn't test any aggregation queries because this is not something I've done before, just simple filtering/mapping functions. I guess I can see that if there's an aggregation which spans many of the `parallel-duration`'s that it would not be able to return the correct result. I'm not sure such a case would be possible to compensate for.

dannykopping:
> @dannykopping can you please provide a query which is not returning correct results? I tested quite a few queries for logs which I have had to download in the past and the parallel and standard methods produced exactly the same files in the output.

Just run any query that has an aggregation, like `count_over_time`. When a query produces metrics this feature should be disabled IMHO.

Issue #7155, opened 2022-09-14
QueryFrontend panics when cloning a special query


DylanGuedes:
**Describe the bug**
One of our query-frontends had a panic on an internal cluster this week caused by a special query. I obfuscated the query but I think this is enough for us to track-down what happened and hopefully replicate this. Obfuscated query:
```
panic: error cloning query: 
(sum(count_over_time({a_id=~"x1|x2|x3|x4|x5|x6|x7|x8|x9|x10|x11|x12|x13|x14|x15|x16|x17|x18|x19|x20|x21|x22|x23|x24|x25|x26|x27", job=~"dev-nice-node_exporter", instance=~".*"} |= "AB"[1h])) / (sum(count_over_time({a_id=~"x1|x2|x3|x4|x5|x6|x7|x8|x9|x10|x11|x12|x13|x14|x15|x16|x17|x18|x19|x20|x21|x22|x23|x24|x25|x26|x27", job=~"dev-nice-node_exporter", instance=~".*"} != "AB"[1h])) / 100)): parse error at line 1, col 1025: syntax error: unexpected $end, expecting NUMBER or { or (
``` 
My hypothesis is that the division (/100, math operation) isn't expected somehow.

**Expected behavior**
No panic! :D

**Environment:**
Latest Loki main

**Screenshots, Promtail config, or terminal output**

PR #8531, opened 2023-02-15
fix panics when cloning a special query

garrettlish:
Sometimes we will get panics when parse the expression which is cloning a special query: `syntax error: unexpected $end, expecting NUMBER or { or (", line:1, col:1025}`.

Fixes [#7155](https://github.com/grafana/loki/issues/7155)

N/A

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


dannykopping (commented):
@garrettlish can you please explain why this change should fix the issue?
This breaks other tests btw: https://drone.grafana.net/grafana/loki/20419/3/5

I would appreciate if you could reduce those test cases down to more simple forms, which describe the actual problem at hand; these queries are complex and they will be difficult to interpret by future readers.

garrettlish:
@dannykopping sorry for break other tests in the first commit. I add that complex query is because the syntax parse will panic for these kinds of complex queries only, looks like `goyacc` cannot parse the complex queries properly if we don't have space between `by` and `(`, let me try to refine the query.

dannykopping (approved):
Thanks!

The fix looks good, but I think we should address the underlying `panic` as well.
I'll raise this with the team 👍

PR #8532, opened 2023-02-15
S3 Storage Class Configuration

### LINKED ISSUE #7678: Make S3 Storage Class configurable
Hey, 
Right now if I understand correctly Loki only works with the “Standard” storage class. Due to price concerns we’d want to run Loki with the “Standard-IA” Storage Class in the same way this ( https://github.com/grafana/tempo/pull/1714 ) pr adds storage class configuration for tempo

justcompile:
Operators may want to use alternative storage classes for S3 objects to; amongst other reasons, reduce cost. Whilst they could use bucket lifecycles to transition the objects into different storage classes, there is a transition cost associated with that

Fixes #7678

Upgrades github.com/thanos-io/objstore (which includes some new/updated dependencies) which allows the setting of StorageClass field from PutUserMetadadta internal fields

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/storage/bucket/s3/bucket_client.go
```diff
@@ -7,6 +7,11 @@ import (
 	"github.com/thanos-io/objstore/providers/s3"
 )
 
+const (
+	// Applied to PUT operations to denote the desired storage class for S3 Objects
+	awsStorageClassHeader = "X-Amz-Storage-Class"
```
vlad-diachenko, line 12: is it case-insensitive? 
AWS Docs says that it must be `x-amz-storage-class` https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html

pkg/storage/bucket/s3/config.go
```diff
@@ -88,6 +104,10 @@ func (cfg *Config) Validate() error {
 		return errUnsupportedSignatureVersion
 	}
 
+	if !util.StringsContain(supportedStorageClasses, cfg.StorageClass) {
+		return errUnsupportedStorageClass
```
vlad-diachenko, line 108: could you please change the error to return the current storage class and all supported storage classes? 
something like : 
`unsupported S3 storage class: users_storage_class . supported storage classes: [STANDARD, STANDARD_IA, ... resto of the classes]`

pkg/storage/chunk/client/aws/s3_storage_client.go
```diff
@@ -39,11 +39,25 @@ import (
...
+	// https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestSyntax
+	StorageClassGlacier                  = "GLACIER"
+	StorageClassDeepArchive              = "DEEP_ARCHIVE"
+	StorageClassGlacierInstantRetrieval  = "GLACIER_IR"
+	StorageClassIntelligentTiering       = "INTELLIGENT_TIERING"
+	StorageClassOneZoneInfrequentAccess  = "ONEZONE_IA"
+	StorageClassOutposts                 = "OUTPOSTS"
+	StorageClassReducedRedundancy        = "REDUCED_REDUNDANCY"
+	StorageClassStandard                 = "STANDARD"
+	StorageClassStandardInfrequentAccess = "STANDARD_IA"
```
vlad-diachenko, line 53: could you please remove code duplicates?  you added exactly the same variables to pkg/storage/bucket/s3/config.go . maybe create a new package with all common logic to a new package and re-use this new package in both packages... 
Does it make sense?

pkg/storage/bucket/s3/bucket_client.go
vlad-diachenko: minio lib confirms that it can be :
```golang
	// Storage class header.
	amzStorageClass = "X-Amz-Storage-Class"
```

so, ignore my comment

vlad-diachenko (commented):
Looks awesome. just think if we could extract code duplicates to a separate package to re-use it.

pkg/storage/chunk/client/aws/s3_storage_client.go
justcompile: Thanks Vlad, yes I agree it should be refactored - it did feel a bit dirty. Is there a package location you'd suggest?

vlad-diachenko: maybe create a package in `pkg/storage/common/aws`

justcompile: Duplicates have been refactored :)

vlad-diachenko (approved):
Looks awesome. thanks @justcompile

pkg/storage/chunk/client/aws/s3_storage_client.go
vlad-diachenko: thanks 🚀

pkg/storage/bucket/s3/config.go
vlad-diachenko: 💠

PR #8537, opened 2023-02-15
Fetch all entries with logcli if limit==0.

jeschkies:
So far logcli enforces a limit for fetching logs. However, sometimes users want to fetch all entries in a time range. This change allows to disable the limit by setting it to `0`. The entries are still fetched in batches. However, the batch loop is only stopped once no more results are returned.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logcli/query/query_test.go
+++ b/pkg/logcli/query/query_test.go
@@ -460,6 +460,46 @@ func Test_batch(t *testing.T) {
 				"line10", "line9", "line8", "line7", "line6b", "line6a", "line6", "line5", "line4", "line3", "line2",
 			},
 		},
+		{
+			name: "single stream backward batch identical timestamps without limit",
+			streams: []logproto.Stream{
+				{
+					Labels: "{test=\"simple\"}",
+					Entries: []logproto.Entry{
+						{Timestamp: time.Unix(1, 0), Line: "line1"},
+						{Timestamp: time.Unix(2, 0), Line: "line2"},
+						{Timestamp: time.Unix(3, 0), Line: "line3"},
+						{Timestamp: time.Unix(4, 0), Line: "line4"},
+						{Timestamp: time.Unix(5, 0), Line: "line5"},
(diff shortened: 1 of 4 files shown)


pkg/logcli/query/query_test.go
```diff
@@ -460,6 +460,46 @@ func Test_batch(t *testing.T) {
...
+			labelMatcher: "{test=\"simple\"}",
+			forward:      false,
+			// Our batchsize is 2 but each query will also return the overlapping last element from the
+			// previous batch, as such we only get one item per call so we make a lot of calls
+			// Call one:   line10 line9 line8 line7
+			// Call two:   line7 line6b line6a line6
+			// Call three: line6b line6a line6 line5
+			// Call four:  line5 line5 line3 line2
+			// Call five:  line1
+			// Call six:   -
```
jeschkies, line 497: This is a little annoying. We could break early if the returned results length is less than the batch size. However, the print stream API does not return that length.

dannykopping (commented):
Nice! LGTM
I'm concerned though that a user can set the batch size really high and cause excessive memory growth in the query-frontend when very large responses are returned. Is this a valid concern?

jeschkies:
> I'm concerned though that a user can set the batch size really high and cause excessive memory growth in the query-frontend when very large responses are returned. Is this a valid concern?

Yes but they can already do this. However, we are not going to run the queries in parallel. Maybe we should set a maximum for the batch size, aka limit in the Loki API.

dannykopping:
> > I'm concerned though that a user can set the batch size really high and cause excessive memory growth in the query-frontend when very large responses are returned. Is this a valid concern?
> 
> Yes but they can already do this. However, we are not going to run the queries in parallel. Maybe we should set a maximum for the batch size, aka limit in the Loki API.

You're right, and i think we're protected by `max_entries_limit_per_query` as well 👍

dannykopping (approved):

PR #8549, opened 2023-02-17
lambda-promtail: various terraform fixes

james-callahan:

Various fixes for the lambda-promtail terraform code

Each commit stands independently (but later ones depend on earlier ones)

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/tools/lambda-promtail/variables.tf
+++ b/tools/lambda-promtail/variables.tf
@@ -92,7 +92,7 @@ variable "lambda_vpc_security_groups" {
 
 variable "kms_key_arn" {
   type        = string
-  description = "kms key arn for encryp env vars."
+  description = "kms key arn for encrypting env vars."
   default     = ""
 }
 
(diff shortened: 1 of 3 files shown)


tools/lambda-promtail/variables.tf
```diff
@@ -67,9 +67,9 @@ variable "extra_labels" {
 }
 
 variable "omit_extra_labels_prefix" {
-  type        = string
+  type        = bool
   description = "Whether or not to omit the prefix `__extra_` from extra labels defined in the variable `extra_labels`."
-  default     = "false"
+  default     = false
```
chaudum, line 72: Should we do the same change for the `keep_stream` variable?

chaudum (commented):
LGTM
One minor question, whether we should also change other boolean variables from string to bool, see other comment.

tools/lambda-promtail/variables.tf
james-callahan: As I mentioned in the commit message: the only reason I changed this was that it hasn't been in a tagged release yet.

I avoided changing `keep_stream` as it has been around a while.

chaudum (approved):

PR #8578, opened 2023-02-21
operator: Refactor status update to reduce API calls

xperimental:
This changes the code used for updating the status of the `LokiStack` to use a single `Update` call to the Kubernetes API instead of doing one per "part" of the status. It also adds logic to retry the update on conflict. This change should reduce the number of API calls we need and make the status update more stable.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/internal/status/status.go
```diff
@@ -2,69 +2,81 @@ package status
...
+	err = k.Status().Update(ctx, &stack)
+	switch {
+	case err == nil:
+		return nil
+	case apierrors.IsConflict(err):
+		// break into retry-logic below on conflict
+		break
+	case err != nil:
+		// return non-conflict errors
+		return err
```
Red-GV, line 71: Maybe I've missed something, but do we even need this when we just call `retry.RetryOnConflict` anyway?

Red-GV (commented):
lgtm for the most part.

operator/internal/status/status.go
xperimental: The flow of this function is a bit unintuitive and that's something that still bothers me. The `retry.RetryOnConflict` will only be reached in the second case because of the `break`, the other two cases return from this function directly.

The upshot of this is that the retry-logic and with that the second `Get` call are only triggered if there was a conflict on the first try.

Red-GV: Oh, sorry. The difference a single word makes. 😆 

I meant: Maybe I've missed something, but do we even need this when we **can** just call `retry.RetryOnConflict` anyway?

periklis (approved):

periklis (approved):

PR #8581, opened 2023-02-21
Size Based Compaction

MasslessParticle:
This draft PR is intended as feedback to https://github.com/grafana/loki/pull/7927 about size-based retention. I'm also including @sandeepsukhani to get a second set of eyes.

This came from trying to reuse as much as possible for size-based compaction. We had all the pieces but they were coupled to the current retention implementation. I decoupled the `expirationChecker` from the compactor and that enabled us to run different kinds of compaction with differing expiration criteria.

Size Based Retention steps:
- Find all the chunks we want to delete by their size on disk
- create an expiration checker that takes the list of deleted chunks that will mark all chunks in the list expired
- Use the chunks' filenames to find the index tables which need to be compacted
- Compact the tables with retention set to `true`

With the custom expiration checker, the compaction should be a no-op and then `applyRetention` will mark all found chunks as deleted.

~1m after than, the normal process will come by and remove all the deleted chunks.

This PR is pretty huge, I'm going to also annotate relevant changes


pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -248,10 +262,10 @@ func (c *Compactor) init(objectClient client.ObjectClient, schemaConfig config.S
 			encoder = client.FSEncoder
 		}
 
-		chunkClient := client.NewClient(objectClient, encoder, schemaConfig)
+		c.chunkClient = client.NewClient(objectClient, encoder, schemaConfig)
```
MasslessParticle, line 265: These are attached to the compactor so we can create `tableMarker` instances for individual compactions

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -260,7 +274,9 @@ func (c *Compactor) init(objectClient client.ObjectClient, schemaConfig config.S
 			return err
 		}
 
-		c.tableMarker, err = retention.NewMarker(retentionWorkDir, c.expirationChecker, c.cfg.RetentionTableTimeout, chunkClient, r)
```
MasslessParticle, line 263: `tableMarker` and `expirationChecker` are now created per compaction rather than up-front

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -316,6 +331,18 @@ func (c *Compactor) starting(ctx context.Context) (err error) {
 		return errors.Wrap(err, "unable to start compactor subservices")
 	}
 
+	if c.sizeBasedRetention != nil {
+		c.sizeBasedRetention.RetentionLoop = services.NewTimerService(
+			c.sizeBasedRetention.RetentionLoopInterval,
+			nil,
+			c.sizeBasedCompaction,
```
MasslessParticle, line 338: This changes to be a function on this type so we have access to all the compaction goodness.

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -533,11 +581,20 @@ func (c *Compactor) RegisterIndexCompactor(indexType string, indexCompactor Inde
 }
 
 func (c *Compactor) RunCompaction(ctx context.Context, applyRetention bool) error {
+	c.compactionMtx.Lock()
```
MasslessParticle, line 584: Ensure `RunCompaction` and `RunSizeBasedCompaction` can't be run at the same time

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -571,17 +628,57 @@ func (c *Compactor) RunCompaction(ctx context.Context, applyRetention bool) erro
...
+	}
+
+	err = c.compactTables(ctx, expirationChecker, tableMarker, true, tables)
+	if err != nil {
+		status = statusFailure
+	}
+	return err
+}
+
+func (c *Compactor) compactTables(ctx context.Context, expirationChecker retention.ExpirationChecker, tableMarker *retention.Marker, applyRetention bool, tables []string) error {
```
MasslessParticle, line 674: Refactor out all the parallel table compaction so both stratagies can use it

pkg/storage/stores/indexshipper/compactor/util/index.go
```diff
@@ -1,14 +1,13 @@
-package compactor
+package util
```
MasslessParticle, line 1: There was already code that converted filenames to chunk refs but it was in a package that couldn't be imported in the compactor without creating a circular dependency.

I moved it here and pointed all usages at it to break the circular dependency.

MasslessParticle:
Looks like this was pulled directly into the branch in the PR?

Abuelodelanada:
> Looks like this was pulled directly into the branch in the PR?

Yes, with @rbarry82 we are working on this.

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -343,6 +370,27 @@ func (c *Compactor) starting(ctx context.Context) (err error) {
...
+func (c *Compactor) sizeBasedCompaction(ctx context.Context) error {
+	if c.cfg.SharedStoreType != config.StorageTypeFileSystem {
+		return nil
+	}
+
+	// TODO: Find the chunks you want to delete
+	// Use the stuff you wrote before
+	// Make sure you use the absolute path
+	// Turn your paths into chunk refs with index_util.ParseChunkRef(index_util.DecodeKey([]byte("absolute path")))
+	// Pass the list of to delete to RunSizeCompaction
```
Abuelodelanada, line 390: Hi @MasslessParticle 

Do you mean something like this?

```golang
var toDelete []retention.ChunkRef

...

fullPath := directory + "/" + file.Name()
chunkRef, _, err := compactor_util.ParseChunkRef(compactor_util.DecodeKey([]byte(fullPath)))
level.Info(util_log.Logger).Log("msg", "block size retention exceeded, removing file", "filepath", fullPath)
if err != nil {
    level.Error(util_log.Logger).Log("msg", err.Error())
    continue
}

toDelete = append(toDelete, chunkRef)
```

If so, with this example `ParseChunkRef()` returns an error....

MasslessParticle: What's `fullPath` look like and what's the error?

PR #8585, opened 2023-02-22
LID: Improve scheduler to control QoS across users within a tenant

chaudum:

Even though Loki is built as multi-tenant system by default, there are use-cases where a Loki installation only has a very large, single tenant, e.g. dedicated Loki cells for customers in Grafana Cloud.

However, there are potentially a lot of different users using the same tenant to query logs, such as users accessing Loki from Grafana or via CLI or HTTP API. This can lead to contention between queries of different users, because they all share the same tenant.

While the current implementation of the scheduler queues allows for QoS guarantees between tenants, it does not account for QoS guarantees across individual users within a single tenant.


docs/sources/lids/_DRAFT_QueryFairness.md
```diff
@@ -0,0 +1,319 @@
...
+T2 -->> Q1
+T2 -->> Q2
+Q2 -->> W2
+```
+
+![component diagram](../scheduler-component-diagram.png)
+
+**Sequence diagram:**
+
+```plantuml
```
jeschkies, line 74: PlantUML is not supported by GitHub but [Mermaid is](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams#creating-mermaid-diagrams). Maybe you could try it out :slightly_smiling_face:

chaudum: The docs will be rendered with Hugo before they go to the website. That said, our documentation setup does not support Mermaid nor PlantUML yet (there are requests to add support). That's why I added the rendered images as well.

jeschkies: Ah sorry. I thought we had it. Well, you could add it manually https://mermaid.js.org/intro/#mermaid-api. I think going with Mermaid would make reviewing in GitHub a little easier :slightly_smiling_face:

docs/sources/lids/_DRAFT_QueryFairness.md
```diff
@@ -0,0 +1,319 @@
...
+```go
+actorN := []string{"L0 Queue", "L1 Queue", "L2 Queue", ... "Ln Queue"}
+```
+
+The L0 queue (root queue) needs to be able to handle worker connections and therefore needs additional functionality compared to its leaf queues.
+
+The following code snippet is meant to show the simplified recursive structure of the queues.
+
+```go
+type RequestQueue struct {
```
jeschkies, line 293: Could you add an example of `GetNextRequestForQuerier`. Also, should there be an interface `Queue`? This way you avoid `interface`.

docs/sources/lids/_DRAFT_QueryFairness.md
```diff
@@ -0,0 +1,319 @@
...
+actorA := []string{"tenant_a", "user_1"}
+actorB := []string{"tenant_b", "user_2"}
+actorC := []string{"tenant_b", "user_3", "service_foo"}
+actorD := []string{"tenant_b", "user_3", "service_bar"}
+```
+
+More generally:
+
+```go
+actorN := []string{"L0 Queue", "L1 Queue", "L2 Queue", ... "Ln Queue"}
```
jeschkies, line 285: I have a tough time understanding how enqueuing and dequeuing would work.

docs/sources/lids/_DRAFT_QueryFairness.md
```diff
@@ -0,0 +1,319 @@
...
+However, this has the disadvantage that the concept of users (that does not exist yet in Loki) leaks into the scheduler domain.
+
+**Pros:**
+* Relatively simple to to implement
+
+**Cons:**
+* Not extensible
+* Leaks domain knowledge
+
+### Proposal 2: Fully hierarchical scheduler
```
jeschkies, line 201: This is easier to implement IMHO. It abstracts away the nesting neatly and we would be able to control it with a feature flag.

docs/sources/lids/_DRAFT_QueryFairness.md
chaudum: Yeah, I can add the interface. However, I don't want to be too specific in this document.

docs/sources/lids/_DRAFT_QueryFairness.md
```diff
@@ -0,0 +1,180 @@
...
+**Pros:**
+* Backwards compatible, because tenant can be identified as `[]string{"tenantID"}`
+* Queue hierarchy can be extended without changing the scheduler implementation
+* Implementation does not require knowledge outside of its domain
+
+**Cons:**
+* More complex to implement than fixed amount of levels
+* Each queue comes with memory overhead
+
+### Proposal 3: Multiple per-tenant sub-queues
```
jeschkies, line 156: This is kind of like our stream sharding.

jeschkies (approved):
I like the interface in Proposal 2.

PR #8591, opened 2023-02-22
Loki: Fix memchunk headblock filter

DylanGuedes:
Fix our memchunk headblock filter to not skip entries that have timestamp=maxT.

### DIFF
--- a/pkg/chunkenc/memchunk.go
+++ b/pkg/chunkenc/memchunk.go
@@ -1005,7 +1005,7 @@ func (hb *headBlock) Iterator(ctx context.Context, direction logproto.Direction,
 	baseHash := pipeline.BaseLabels().Hash()
 	process := func(e entry) {
 		// apply time filtering
-		if e.t < mint || e.t >= maxt {
+		if e.t < mint || e.t > maxt {
 			return
 		}
 		stats.AddHeadChunkBytes(int64(len(e.s)))
--- a/pkg/chunkenc/memchunk.go
+++ b/pkg/chunkenc/memchunk.go
@@ -1008,6 +1008,9 @@ func (hb *headBlock) Iterator(ctx context.Context, direction logproto.Direction,
 		if e.t < mint || e.t > maxt {
 			return
 		}
+		if e.t == maxt && mint != maxt {
+			return
+		}
 		stats.AddHeadChunkBytes(int64(len(e.s)))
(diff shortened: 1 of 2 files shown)


JordanRushing (approved):

pkg/chunkenc/memchunk.go
```diff
@@ -1005,7 +1005,7 @@ func (hb *headBlock) Iterator(ctx context.Context, direction logproto.Direction,
 	baseHash := pipeline.BaseLabels().Hash()
 	process := func(e entry) {
 		// apply time filtering
-		if e.t < mint || e.t >= maxt {
+		if e.t < mint || e.t > maxt {
```
slim-bean, line 1008: I think in all cases timestamp filtering Loki is inclusive of the start time and exclusive of the end time.

I'm not 100% sure what's driving this change but I think it is likely correct as is?

slim-bean: Lol, sorry should have slowed down a little, it looks like the change you are making would follow the convention I just outlined.

Is there more context for this change though?

DylanGuedes: context:
- we're releasing a new version that updates Go version to Go 1.20
- go 1.20 linter complained that this part of the code was shadowing the `tc` variable
- once I added the `tc := tc` assignment the tests started to break. Removing the parallel flag also exposed the problem

DylanGuedes: all that said: this isn't a major problem, I talked with Owen and it is a nanosecond precision, so not really important. however, it is blocking the new Loki version.

slim-bean: Looking at this again, I think I stand by my initial feedback, I think this change breaks important behavior. (it's actually testing if the timestamp is outside the bounds)

slim-bean: ts == maxt is technically outside of the bounds since end time is not inclusive.

DylanGuedes: but what if maxt == mint? ex: maxt=1, mint1= and you have an entry with ts=1.
since start time is inclusive I'd expect it to have precedence over end time not being inclusive.

owen-d: I think when `maxt == mint` (such as an instant query), we mutate the `endts = endts+1nanosecond` prior to running the query

PR #8568, opened 2023-02-21
[new feature] introduce loki Coprocessor querier pre query ,And provider a golang demo XRayCoprocessor.

liguozhong:
**What this PR does / why we need it**:
ref issue : [High cardinality labels](https://github.com/grafana/loki/issues/91) 
`{log_type="service_metrics"} |= "ee74f4ee-3059-4473-8ba6-94d8bfe03272"
`

We have counted the source distribution of our logql, and 85% of the grafana log explore queries are traceID queries.
Generally, the log time range of traceID is about 10 minutes（trace time= start～end）, but because users do not know traceID start time and end time , they usually search for 7 day log. In fact having a time range of "7d-10m" is an invalid search.

So we hope to introduce some auxiliary abilities to solve this "7d-10m" invalid search.

We have checked that in the database field, such feature have been implemented very maturely. 

And our team tried to implement the `preQuery Coprocessor`, and achieved great success. Through this feature, we solved the problem of "loki + traceID search is very slow".

image
Thanks Google’s BigTable coprocessor and HBase coprocessor.
HBase coprocessor_introduction link： https://blogs.apache.org/hbase/entry/coprocessor_introduction
The idea of HBase Coprocessors was inspired by Google’s BigTable coprocessors. Jeff Dean gave a talk at LADIS ’09 (http://www.scribd.com/doc/21631448/Dean-Keynote-Ladis2009, page 66-67)

```
HBase Coprocessor 
The RegionObserver interface provides callbacks for:

preOpen, postOpen: Called before and after the region is reported as online to the master.
preFlush, postFlush: Called before and after the memstore is flushed into a new store file.
preGet, postGet: Called before and after a client makes a Get request.
preExists, postExists: Called before and after the client tests for existence using a Get.
prePut and postPut: Called before and after the client stores a value.
preDelete and postDelete: Called before and after the client deletes a value.
etc.
```

**Describe the solution you'd like**
```
Loki Coprocessor 
The QuerierObserver interface provides callbacks for:

`**preQuery**`: Called before querier , Pass (logql, start, end) 3 parameters to the Coprocessor,
 and the Coprocessor judges whether it is necessary for the querier to actually execute this query.

 For example, for traceID search,   query range = 7d + `split_queries_by_interval: 2h`. 
This logql query will actually be divided into 84 query sub-requests, and here 83 are invalid, 
and only one 2h sub-request can find the log of traceID.
We try to implement two types of Coprocessors in this scenario.

traceID Coprocessor 1 simple text analysis : 
[...]

Issue #91, opened 2018-12-13
High cardinality labels


sandstrom:
Most of our services (nginx, application servers, etc) emit JSON lines[1] like this:

```
{ "timestamp" : "2018-12-24T18:00:00", "user-id" : "abc", "request-id" : "123", "message" : "Parsing new user data" }
{ "timestamp" : "2018-12-24T18:01:00", "user-id: " "def", "request-id" : "456", "message" : "Updating user record" }
{ "timestamp" : "2018-12-24T18:02:00", "user-id: " "abc", "request-id" : "789", "message" : "Keep alive ping from client" }
```

While I realize that high cardinality will affect indices, I think it's worth discussing whether this is something that Loki can support in the future. My guess is that you'll find a lot of logging use-cases (for example ELK) where people can easily ditch full-text search. But not having easy lookup on log metadata such as user-id or request-id may pose a problem. 

[1] http://jsonlines.org/

*(this is partly related to labeling streams at ingest time, but I've split it out into two separate issues, the other one is here: https://github.com/grafana/loki/issues/88)*

sandstrom:
A thought:

An alternative to high cardinality labels could be to introduce a complement to labels called *metadata*, that is allowed to be high cardinality. Having two different things could allow us to impose other restrictions on the metadata type of values.

If grep is quick enough, maybe it would work to not index the metadata key/value pairs. But still allow them to be filtered (at "grep speed"). That would allow you to give these high-cardinality key-value pairs *blessed* UI, for easy filtering, whilst still avoiding the cost of indexing high cardinality fields.

tomwilkie:
> While I realize that high cardinality will affect indices, I think it's worth discussing whether this is something that Loki can support in the future.

The metadata we index for each streams has the same restrictions as it does for Prometheus labels: I don't want to say never, but I'm not sure if this is something we're ever likely to support.

> An alternative to high cardinality labels could be to introduce a complement to labels called metadata, that is allowed to be high cardinality.
> But not having easy lookup on log metadata such as user-id or request-id may pose a problem.

Does the support for regexp filtering allow you to filter by a given user-id or request-id and achieve the same end as a second sets of non-index labels?

I'd agree its cumbersome, so perhaps adding some jq style language here to make this more natural for json logs would be better?

sandstrom:
@tomwilkie It's all about speed really. With our ELK here are some of the most common tasks:

[...]

PR #8616, opened 2023-02-24
LID: introduce Coprocessor

liguozhong:
[new feature] introduce loki Coprocessor querier pre query

Fixes https://github.com/grafana/loki/pull/8568 AND https://github.com/grafana/loki/pull/8568

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/lids/0003-Coprocessor.md
+++ b/docs/sources/lids/0003-Coprocessor.md
@@ -1,0 +1,113 @@
+---
+title: "0003: Coprocessor"
+description: "introduce coprocessor from Google’s BigTable coprocessor and HBase coprocessor"
+---
+
+# 0003: Coprocessor
+
+**Author:** liguozhong (fuling.lgz@alibaba-inc.com)
+
+**Date:** 02/2023
+
+**Sponsor(s):** @jeschkies
+
+**Type:** Feature
+
+**Status:** Draft
+
+**Related issues/PRs:** https://github.com/grafana/loki/issues/8559 AND https://github.com/grafana/loki/issues/91
+
+**Thread from [mailing list](https://groups.google.com/forum/#!forum/lokiproject):** N/A
+
+---
+
+## Background
+
+{log_type="service_metrics"} |= "ee74f4ee-3059-4473-8ba6-94d8bfe03272"
+
+We have counted the source distribution of our logql, and 85% of the grafana log explore queries are traceID queries.
+Generally, the log time range of traceID is about 10 minutes（trace time= start～end）
+
+## Problem Statement
+
+but because users do not know traceID start time and end time , they usually search for 7 day log. 
+In fact having a time range of "7d-10m" is an invalid search.
+
+## Goals
+
+So we hope to introduce some auxiliary abilities to solve this "7d-10m" invalid search.
+
+We have checked that in the database field, such feature have been implemented very maturely.
+
+And our team tried to implement the preQuery Coprocessor, and achieved great success. 
+
+Through this feature, we solved the problem of "loki + traceID search is very slow".
+
+## Non-Goals
+
+The problem of slow traceID search In the past six months, we tried to introduce kv system / reverse index text system/ bloomfilter to speed up logql return, but the machine cost was too high and finally gave up.
+So we don't want to introduce cost-intensive solutions to solve the problem of slow traceID search.
+
+## Proposals
+
+### Proposal 0: Do nothing
+
+Users cannot migrate from log systems like ELK and other indexing schemes to LOKI,if the user heavily uses traceID log search.
+
+### Proposal 1: Query Coprocessor
+
+Thanks Google’s BigTable coprocessor and HBase coprocessor.
+HBase coprocessor_introduction link： https://blogs.apache.org/hbase/entry/coprocessor_introduction
+The idea of HBase Coprocessors was inspired by Google’s BigTable coprocessors. Jeff Dean gave a talk at LADIS ’09 (http://www.scribd.com/doc/21631448/Dean-Keynote-Ladis2009, page 66-67)
+HBase Coprocessor
+The RegionObserver interface provides callbacks for:
+
+`**preOpen, postOpen**`: Called before and after the region is reported as online to the master.
+`**preFlush, postFlush**`: Called before and after the memstore is flushed into a new store file.
+`**preGet, postGet**`: Called before and after a client makes a Get request.
+`**preExists, postExists**`: Called before and after the client tests for existence using a Get.
+`**prePut and postPut**`: Called before and after the client stores a value.
+`**preDelete and postDelete**`: Called before and after the client deletes a value.
+etc.
+
+
+Loki Coprocessor
+The QuerierObserver interface provides callbacks for:
+
+`**preQuery**`: Called before querier , Pass (logql, start, end) 3 parameters to the Coprocessor,
+and the Coprocessor judges whether it is necessary for the querier to actually execute this query.
+
+For example, for traceID search,   query range = 7d + `split_queries_by_interval: 2h`.
+This logql query will actually be divided into 84 query sub-requests, and here 83 are invalid,
+and only one 2h sub-request can find the log of traceID.
+We try to implement two types of Coprocessors in this scenario.
+
+traceID Coprocessor 1 simple text analysis :
+if traceID is traceID from XRay or openTelemtry (《Change default trace-id format to be
+similar to AWS X-Ray (use timestamp )#1947》), this type of traceID information has a timestamp,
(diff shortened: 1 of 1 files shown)


docs/sources/lids/0003-Coprocessor.md
```diff
@@ -0,0 +1,113 @@
...
+
+For example, for traceID search,   query range = 7d + `split_queries_by_interval: 2h`.
+This logql query will actually be divided into 84 query sub-requests, and here 83 are invalid,
+and only one 2h sub-request can find the log of traceID.
+We try to implement two types of Coprocessors in this scenario.
+
+traceID Coprocessor 1 simple text analysis :
+if traceID is traceID from XRay or openTelemtry (《Change default trace-id format to be
+similar to AWS X-Ray (use timestamp )#1947》), this type of traceID information has a timestamp,
+and Coprocessor can specify a trace to execute the longest duration to cooperate
```
jeschkies, line 88: Ah. So the coprocessor parses the timestamp. Did you try doing it in LogQL?

docs/sources/lids/0003-Coprocessor.md
```diff
@@ -0,0 +1,113 @@
...
+
+We have counted the source distribution of our logql, and 85% of the grafana log explore queries are traceID queries.
+Generally, the log time range of traceID is about 10 minutes（trace time= start～end）
+
+## Problem Statement
+
+but because users do not know traceID start time and end time , they usually search for 7 day log. 
+In fact having a time range of "7d-10m" is an invalid search.
+
+## Goals
```
jeschkies, line 36: If I understood the HBade use case they basically created a plug-in system with some hooks. This seems to be the goal here. How would we implement it? How are we shipping the data? We could use Hashicorp's go-plug but there's probably s serialization overhead. As I understood the HBade design the coprocessor is started on the host where the data is an this can read the data directly. Not sure we can do it here.

jeschkies (commented):
Thanks for your contribution. I like the idea overall but have some open question. Also, this use case is very narrow. I wonder if there's another solution.

docs/sources/lids/0003-Coprocessor.md
liguozhong: > If I understood the HBade use case they basically created a plug-in system with some hooks.

yes.

> How would we implement it? How are we shipping the data? We could use Hashicorp's go-plug but there's probably s serialization overhead

My suggestion is that we should not restrict the language. Through http+protobuf, more developer can work together with loki. 
Because prometheus remote read and k8s hpa implement the plugin mechanism through http+protobuf, this  has achieved great success.
The implementation of HBase limits the jvm language so that the coprocessor can only be implemented in languages such as java or scala. I don't think it's a good idea.



> Hashicorp's go-plug

Will this limit the development language?

jeschkies: > Will this limit the development language?

No, they also use [gRPC](https://github.com/hashicorp/go-plugin#features) but make it a little simpler for plugins in Go.

liguozhong: It would be great if there is no restriction on the implementation language. 
Can we refer to open telemetry’s collector that supports both http and grpc protocols?
The SREs I know actually don’t know much about grpc, but they can easily build an http server .

jeschkies: I don't think the plug-in is restricted to Go. One could use gRPC.

liguozhong: great

docs/sources/lids/0003-Coprocessor.md
```diff
@@ -0,0 +1,113 @@
...
+`**preQuery**`: Called before querier , Pass (logql, start, end) 3 parameters to the Coprocessor,
+and the Coprocessor judges whether it is necessary for the querier to actually execute this query.
+
+For example, for traceID search,   query range = 7d + `split_queries_by_interval: 2h`.
+This logql query will actually be divided into 84 query sub-requests, and here 83 are invalid,
+and only one 2h sub-request can find the log of traceID.
+We try to implement two types of Coprocessors in this scenario.
+
+traceID Coprocessor 1 simple text analysis :
+if traceID is traceID from XRay or openTelemtry (《Change default trace-id format to be
```
cyriltovena, line 86: So this works only if we're using XRay format ?

owen-d (commented):
I think this is a pretty interesting idea, thanks for opening the issue. We've also been thinking about the "find a traceID over a large time range" use case, which is a common ask for people migrating to Loki from other, more heavily index based projects. I've been considering ways to solve it _within_ the project, but this is another very interesting approach. To be honest, I'm not sure how I feel yet. I see our current options as:
1) Do nothing
2) Implement key-value lookups within Loki
  * Benefits: Less complex for users who want key-value lookups in loki.
  * Costs: More complexity (code & configuration) in the base system. More surface area in the project that not all users may use. Not sure how to do it (yet).
3) Implement Coprocessor support in Loki.
  * Benefits: Loki doesn't need to solve each use case explicitly. Other use cases can be added. Loki can be extended in arbitrary ways.
  * Costs: External use cases require a lot more investment, both operationally (running plugins) and development (building plugins). Loki now has arbitrary dependencies. Reliability concerns with external plugins, etc.
4) Something else?

Sorry I don't have an immediate answer, I need to think about this more.

As for the method signature you proposed, what about something like:

```go
PreQuery(params logql.Params) (mutatedParams logql.Params, shouldQuery bool)
```

This would allow loki to choose how to _change_ the query that's executed (for instance by reducing the time range to just the `10m` that contains the trace) as well as choose to halt the query.

docs/sources/lids/0003-Coprocessor.md
liguozhong: Yes, our team's exploration is only for this purpose now.

liguozhong:
thanks，Very detailed answer, if loki wants to directly solve the traceID search problem in the future.

I suggest not to consider introducing a similar mechanism in the past two years.This will disperse the energy of the community. 

In the future, there may be duplication of feature, and it will be a very difficult time to choose one of the two feature to delete.

liguozhong:
I have seen tempo's traceID related code before, which is to build a bloomfilter for each block when writing, and quickly judge whether there is traceID in the block through bloomfilter when querying a traceID.

owen-d:
> I have seen tempo's traceID related code before, which is to build a bloomfilter for each block when writing, and quickly judge whether there is traceID in the block through bloomfilter when querying a traceID.

Ha, I've been thinking about this _exact_ same thing :)

slim-bean:
@liguozhong thank you for this idea and discussion!

There is something here, I don't know what it will look like though.

Solving for the X-Ray trace ID like this is very clever.

We have not given up on this idea yet.

sandstrom:
@slim-bean Any estimate on a time frame in which you are planning to tackle this.

We have this problem with our Loki setup, where basically 90% of queries are for a request-id (unique id shared by some ~10-100 lines that all were part of the same request) or similar.

It's a pain point and we're sometimes regretting that we didn't stick with ELK, even though Loki also have a lot of nice properties that we like.

## Wild thought

Maybe bloom filters could be used? They have false positives, but not false negatives. A request-id stored in a bloom filter would at worst cause loki to grab an extra chunk or two that it didn't have to look in. Maybe not so bad.

valyala:
FYI, [VictoriaLogs](https://docs.victoriametrics.com/VictoriaLogs/) supports high-cardinality labels - they just don't go into [stream labels](https://docs.victoriametrics.com/VictoriaLogs/keyConcepts.html#stream-fields), so they do not lead to high cardinality issues such as high memory usage, OOM or significant slowdown. It supports fast search over such labels with `label_name:value` query syntax - see [these docs](https://docs.victoriametrics.com/VictoriaLogs/LogsQL.html) for details.

It would be great if Loki will support high-cardinality labels in a similar fashion too. See more details on the underlying storage architecture of VictoriaLogs [here](https://docs.victoriametrics.com/VictoriaLogs/FAQ.html#how-does-victorialogs-work).

sandstrom:
@slim-bean @owen-d Congrats on the 2.9 release! 🎉 

Just wanted to check in on this issue and the related [High Cardinality labels](https://github.com/grafana/loki/issues/91) and see if you have any plans in that area?

We've migrated off ElasticSearch a while ago. Overall happy, but even with our modest data, finding these 'needle in the haystack' type of things (request ids and trace ids) is an area where Loki could improve.

liguozhong:
90% of our logs are traceID and requestID logs. We solved the problem by changing requestID to traceID.

PR #8637, opened 2023-02-26
Iterators: re-implement mergeEntryIterator using loser.Tree for performance

bboreham:

Building on #8351, this re-implements `mergeEntryIterator` using `loser.Tree`; the benchmark says it goes much faster but uses a bit more memory (while building the tree).

```
name                       old time/op    new time/op    delta
SortIterator/merge_sort-4    10.7ms ± 4%     2.9ms ± 2%  -72.74%  (p=0.008 n=5+5)

name                       old alloc/op   new alloc/op   delta
SortIterator/merge_sort-4    11.2kB ± 0%    21.7kB ± 0%  +93.45%  (p=0.008 n=5+5)

name                       old allocs/op  new allocs/op  delta
SortIterator/merge_sort-4      6.00 ± 0%      7.00 ± 0%  +16.67%  (p=0.008 n=5+5)
```

The implementation is very different: rather than relying on iterators supporting `Peek()`, `mergeEntryIterator` now pulls items into its buffer until it finds one with a different timestamp or stream, and always works off what is in the buffer.
The comment `"[we] pop the ones whose common value occurs most often."` did not appear to match the previous implementation, and no attempt was made to match this comment.

A `Push()` function was added to `loser.Tree` to support live-streaming. This works by finding or making an empty slot, then re-running the initialize function to find the new winner.  A consequence is that the previous "winner" value is lost after calling
`Push()`, and users must call `Next()` to see the next item.

A couple of tests had to be amended to avoid assuming particular behaviour of the implementation; I recommend that reviewers consider these closely.

**Checklist**
- NA Documentation added
- [x] Tests updated
- NA `CHANGELOG.md` updated
- NA Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

### DIFF
--- a/pkg/iter/entry_iterator.go
+++ b/pkg/iter/entry_iterator.go
@@ -1,7 +1,6 @@
 package iter
 
 import (
-	"container/heap"
 	"context"
 	"io"
 	"math"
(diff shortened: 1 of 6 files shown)


MasslessParticle (approved):
Awesome, thanks for this!

PR #8640, opened 2023-02-27
added missed binaries required to run `./tools/image-tag` script

vlad-diachenko:

[This drone step](https://drone.grafana.net/grafana/loki/20764/21/3) had to prepare a config for the `updater` to update loki helm chart with the latest image version automatically. However, it failed due to an error:
```
+ RELEASE_TAG=$(./tools/image-tag)
env: can't execute 'bash': No such file or directory
```
I forgot to install `bash` and `git` before running this script inside `alpine` container.

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -707,6 +707,8 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
         image: 'alpine',
         depends_on: ['check-version-is-latest'],
         commands: [
+          'apk add --no-cache bash git',
+          'git fetch origin --tags',
           'RELEASE_TAG=$(./tools/image-tag)',
           'echo $PLUGIN_CONFIG_TEMPLATE > %s' % configFileName,
           // replace placeholders with RELEASE TAG
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -1292,6 +1292,8 @@ steps:
     event:
     - tag
 - commands:
+  - apk add --no-cache bash git
+  - git fetch origin --tags
   - RELEASE_TAG=$(./tools/image-tag)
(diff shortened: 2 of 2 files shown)


kavirajk (approved):
LGTM.  :+1: 

Q: do we really need `bash` here? (I see only we use git)

kavirajk:
Ah. nvm. PR description make it clear.

.drone/drone.jsonnet
```diff
@@ -707,6 +707,8 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
         image: 'alpine',
         depends_on: ['check-version-is-latest'],
         commands: [
+          'apk add --no-cache bash git',
```
dannykopping, line 710: Out of interest, can we not just use `ash` instead?

kavirajk: You mean `sh` instead of `bash`?

```
+ RELEASE_TAG=$(./tools/image-tag)
env: can't execute 'bash': No such file or directory
```

I guess our `image-tag` requires `bash` looks like.

dannykopping: No, I mean `ash` which is the default shell of Alpine

dannykopping: We have `#!/usr/bin/env bash` in the header of `image-tag`, which is probably unnecessary.

vlad-diachenko: > We have #!/usr/bin/env bash in the header of image-tag, which is probably unnecessary.

yep, nothing bash specific is used in the script )) so I believe we can remove the header from the script or make it flexible to be able to run in any terminal...

anyway, not only bash was missed, `git` is also required, so let's merge this PR and update `./tools/image-tag` script when we have time and need  ))

dannykopping: SGTM 👍

PR #8644, opened 2023-02-27
Helm: Split rules into more groups

DylanGuedes:
Split our rules into more rule groups in case evaluation takes too long. This is similar to what Mimir does https://github.com/grafana/mimir/blob/main/operations/mimir-mixin-compiled/rules.yaml.


Jayclifford345:
Hey @DylanGuedes, I am closing this out as we are having a bit of a late spring clean. Happy to reopen if we need to.

PR #8646, opened 2023-02-27
Optimize .+ regex line filter

MasslessParticle:
Interpolated variables from Grafana dashboards may result in the regex matcher `.+`. Filters with `+` are already being rewritten to be non greedy but there is a further optimization to make replace the regex altogether with a filter that just matchers `len(line) > 0`

### DIFF
--- a/pkg/logql/log/filter.go
+++ b/pkg/logql/log/filter.go
@@ -33,6 +33,20 @@ func (trueFilter) ToStage() Stage       { return NoopStage }
 // TrueFilter is a filter that returns and matches all log lines whatever their content.
 var TrueFilter = trueFilter{}
 
+type existsFilter struct{}
+
+func (existsFilter) Filter(_ []byte) bool { return true }
+func (existsFilter) ToStage() Stage {
+	return StageFunc{
+		process: func(_ int64, line []byte, _ *LabelsBuilder) ([]byte, bool) {
+			return line, len(line) > 0
+		},
+	}
+}
+
+// ExistsFilter is a filter that returns and matches when a line has any characters.
+var ExistsFilter = existsFilter{}
+
 type notFilter struct {
 	Filterer
 }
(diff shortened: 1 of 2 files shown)


owen-d (approved):
LGTM, excited to see some benches

pkg/logql/log/filter.go
```diff
@@ -425,6 +439,10 @@ func simplify(reg *syntax.Regexp) (Filterer, bool) {
 		if reg.Sub[0].Op == syntax.OpAnyCharNotNL {
 			return TrueFilter, true
 		}
+	case syntax.OpPlus:
+		if len(reg.Sub) == 1 && reg.Sub[0].Op == syntax.OpAnyCharNotNL { // simplify ".+"
```
DylanGuedes, line 443:
```suggestion
		if len(reg.Sub) >= 1 && reg.Sub[0].Op == syntax.OpAnyCharNotNL { // simplify ".+"
```

no idea if the suggestion makes sense but maybe we can have multiple subexpressions and the simplification still be valid

jeschkies (approved):
Would you have a benchmark? I wonder how much faster this is.

pkg/logql/log/filter.go
MasslessParticle: There's definitely room for improvement but it gets more difficult to understand what the expression is doing when there are more sub expressions of `+` than just a single `.`

MasslessParticle:
@jeschkies @owen-d Looks like it improved the performance pretty substantially:
```
name                               old time/op    new time/op    delta
_LineFilter/simplified_true_.+-8     45.2ns ± 2%     2.0ns ± 7%  -95.56%  (p=0.000 n=8+10)
_LineFilter/simplified_false_.+-8    50.4ns ±16%     3.0ns ± 2%  -94.05%  (p=0.000 n=10+8)

name                               old alloc/op   new alloc/op   delta
_LineFilter/simplified_true_.+-8      0.00B          0.00B          ~     (all equal)
_LineFilter/simplified_false_.+-8     0.00B          0.00B          ~     (all equal)

name                               old allocs/op  new allocs/op  delta
_LineFilter/simplified_true_.+-8       0.00           0.00          ~     (all equal)
_LineFilter/simplified_false_.+-8      0.00           0.00          ~     (all equal)
```

PR #8662, opened 2023-03-01
LogQL: Introduce `distinct`

### LINKED ISSUE #8649: Feature Request: [LogQL ] should have a way to filter out duplicate label values
It's like this:
### Original log lines hava common label `app=order`
```
{"event": "access", "id": "1", "time": "2023-02-28 15:12:11"}
{"event": "access", "id": "1", "time": "2023-02-28 15:13:11"}
{"event": "access", "id": "2", "time": "2023-02-28 15:14:11"}
{"event": "access", "id": "2", "time": "2023-02-28 15:15:11"}
```

### `LogQL` to query logs: ` {app="order"} | json | distinct id `

```
{"event": "access", "id": "1", "time": "2023-02-28 15:13:11"}
{"event": "access", "id": "2", "time": "2023-02-28 15:15:11"}
```

### Context (why I need this feature)
I want to query a user's activity log, but don't want to see duplicate logs for the same user.

liguozhong:
Introduces `distinct` to LogQL. Usage:
`{job="varlogs"} | distinct filename`
It is similar to `distinct` in SQL and especially useful for sampling. Similar syntax exists in other log query languages.

**Example**
For the following log lines:
```json
{"event": "access", "id": "1", "time": "2023-02-28 15:12:11"}
{"event": "access", "id": "1", "time": "2023-02-28 15:13:11"}
{"event": "access", "id": "2", "time": "2023-02-28 15:14:11"}
{"event": "access", "id": "2", "time": "2023-02-28 15:15:11"}
```

The query below:
```logql
{app="order"} | json | distinct id
```

Will return:
```json
{"event": "access", `"id": "1",` "time": "2023-02-28 15:13:11"}
{"event": "access", `"id": "2", `"time": "2023-02-28 15:15:11"}
```

**Example with multiple labels**
```logql
{app="order"} | json | distinct id,time
```

Fixes #8649 

`{job="varlogs"} |distinct filename
`
The test passed in a simple local development environment
snapshot.
image

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


rkonfj:
@liguozhong Thank you very much for solving the problem that is bothering me, I just want to say cool!

Two questions:
1. Can multiple labels be specified?（Doesn't matter if this feature exists or not） `| distinct label1, label2`
2. Are the deduplicated log lines the latest? I don't know if `LogQL` needs to implement a deduplication strategy, at least the strategy I need is to keep only the latest log line.

docs/sources/logql/log_queries/_index.md
```diff
@@ -268,6 +268,16 @@ To evaluate the logical `and` first, use parenthesis, as in this example:
 
 Label filter expressions have support matching IP addresses. See [Matching IP addresses]({{<relref "../ip">}}) for details.
 
+### Distinct filter expression
+
+Distinct filter expression allows filtering log line using their original and extracted labels to filter out duplicate label values.
```
JStickler, line 273:
```suggestion
Distinct filter expression allows filtering log lines using their original and extracted labels to filter out duplicate label values.
```

JStickler (approved):
[Docs squad]  One small comment, otherwise the docs LGTM.

liguozhong:
> * Can multiple labels be specified?（Doesn't matter if this feature exists or not） `| distinct label1, label2`

nice, I submitted a commit to complete this feature.
```logql
{app="order"} | json | distinct id,time
```

liguozhong:
> 2\. Are the deduplicated log lines the latest? I don't know if `LogQL` needs to implement a deduplication strategy, at least the strategy I need is to keep only the latest log line.

No, currently the earliest log is kept. because the `log query` is stateless now, so aggregation cannot be achieved.
If we want to realize this feature, we need to make a lot of changes, and we need to introduce a state for the` log query`. According to my analysis, this requirement is difficult to achieve.

liguozhong:
The mysql's distinct was no keep the last one too, maybe we can keep the same logic with mysql's SQL now.

rkonfj:
@liguozhong 
> nice, I submitted a commit to complete this feature.

Thanks.

> No, currently the earliest log is kept. because the log query is stateless now, so aggregation cannot be achieved.
If we want to realize this feature, we need to make a lot of changes, and we need to introduce a state for the log query. According to my analysis, this requirement is difficult to achieve.

You are right, pipeline filtering is not possible, which conflicts with streaming log lines, since the latest log line is in the future.

However, when the date range and query conditions are determined, the log lines is no longer a stream, but a determined lines. Is it possible to reevaluate the log lines to be returned(if there is a suitable place to do that)?

rkonfj:
> The mysql's distinct was no keep the last one too, maybe we can keep the same logic with mysql's SQL now.

The result of the MySQL query is the column values, it doesn't matter whether it is earliest or latest. But for logs, the log content itself is important, the deduplication strategy become more important, right?

liguozhong:
`Metric query` can do agg and do a commit after all the data is processed. But there is no such callback func in `log query` now.

And there is an important question we have to consider.
The `log query `always has a limit, which means that we don't need to pull all the logs to end the query, but if we promise that `distinct` can return the last line of logs, we will have to pull all the s3 files to end this query.

I'm afraid it's hard for loki to pull the last log.But I also think that the last line of log is the most important log. Maybe you can modify the start time continuously.

liguozhong:
Anyway, your issue is very valuable, I asked our developers, they also want a feature in their daily oncall work.

pkg/logql/log/distinct.go
```diff
@@ -0,0 +1,37 @@
+package log
+
+type distinctFilter struct {
+	datas  map[string]map[string]int
```
jeschkies, line 4: How do you keep the state across multiple queriers?

liguozhong: Great point, I didn't notice this detail.
Maybe we can avoid logql to be executed in multiple queriers by making the `Shardable() bool` function of `DistinctFilterExpr` struct return false, just like calculating the 99th quantile.

There are too many details to consider when introducing a state
image

jeschkies: Great idea. Could add it?

jeschkies: You always did 😊

pkg/logql/log/distinct.go
```diff
@@ -0,0 +1,37 @@
...
+	}, nil
+}
+func (r *distinctFilter) Process(_ int64, line []byte, lbs *LabelsBuilder) ([]byte, bool) {
+	isPass := false
+	for _, label := range r.labels {
+		val, ok := lbs.Get(label)
+		if !ok {
+			return line, false
+		}
+		_, ok = r.datas[label][val]
```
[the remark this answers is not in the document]
liguozhong: If the label does not exist, it will be discarded, and the new label is based on the order of the pipeline. Such as the following example

log
`{"a":"1", "b":2}`

logql 1 :
`{app="buy2"} | json | distinct a
`
logql 2 :
`{app="buy2"} | distinct a | json`

logql 1 can get log. logql 2 will not get log.

docs/sources/logql/log_queries/_index.md
```diff
@@ -268,6 +268,16 @@ To evaluate the logical `and` first, use parenthesis, as in this example:
...
+### Distinct filter expression
+
+Distinct filter expression allows filtering log lines using their original and extracted labels to filter out duplicate label values.
+
+For example the following expression:
+
+```logql
+{job="varlogs"} |distinct filename
+```
+
```
DylanGuedes, line 280:
```suggestion
For example, for the following log lines:
```
{"event": "access", "id": "1", "time": "2023-02-28 15:12:11"}
{"event": "access", "id": "1", "time": "2023-02-28 15:13:11"}
{"event": "access", "id": "2", "time": "2023-02-28 15:14:11"}
{"event": "access", "id": "2", "time": "2023-02-28 15:15:11"}
```

The expression `{app="order"} | json | distinct id` will return the distinct occurrences of `id`:
```
{"event": "access", "id": "1", "time": "2023-02-28 15:13:11"}
{"event": "access", "id": "2", "time": "2023-02-28 15:15:11"}
```

File: docs/sources/logql/log_queries/_index.md [hunk 5]
```diff
@@ -268,6 +268,16 @@ To evaluate the logical `and` first, use parenthesis, as in this example:
 
 Label filter expressions have support matching IP addresses. See [Matching IP addresses]({{<relref "../ip">}}) for details.
 
+### Distinct filter expression
+
+Distinct filter expression allows filtering log lines using their original and extracted labels to filter out duplicate label values.
+
```
DylanGuedes, line 274:
```suggestion
Distinct filter expression allows filtering log lines using their original and extracted labels to filter out duplicate label values. The first line occurrence of a distinct value is returned, and the others are dropped.

```

pkg/logql/log/distinct.go
```diff
@@ -0,0 +1,37 @@
...
+		if ok {
+			return line, false
```
DylanGuedes, line 27: if `r.datas[label][val]` is 1 why don't you have pass=true here?

ex:
1. starting with lbs={"a": "bla"}, r.datas={"a": "bla"}
2.  lbs.Get("a") will return `"bla", true` for line 21
3. r.datas["a"]["bla"] will have `ok=true` for line 25
4. instead of dropping the log line, because you had already seen that label value, you actually return with pass=false

pkg/logql/log/distinct_test.go
```diff
@@ -0,0 +1,61 @@
...
+			{Name: logqlmodel.ErrorLabel, Value: errJSON},
+			{Name: "status", Value: "200"},
+			{Name: "method", Value: "POST"},
+		},
+		input: []string{
+			`{"event": "access", "id": "1", "time": "1"}`,
+			`{"event": "access", "id": "1", "time": "2"}`,
+			`{"event": "access", "id": "2", "time": "3"}`,
+			`{"event": "access", "id": "2", "time": "4"}`,
+			`{"event": "access", "id": "1", "time": "5"}`,
```
DylanGuedes, line 34: could you add one more case with "id": "1", "time":1", but with a different event?

pkg/logql/log/distinct_test.go
```diff
@@ -0,0 +1,61 @@
...
+		},
+	}
+
+	distinctFilter, err := NewDistinctFilter(c.label)
+	require.NoError(t, err)
+
+	total := 0
+	passLines := make([]string, 0)
+	for _, line := range c.input {
+		b := NewBaseLabelsBuilder().ForLabels(c.lbs, c.lbs.Hash())
```
DylanGuedes, line 49: in case this is valid:
```suggestion
	passLines := make([]string, 0)
	b := NewBaseLabelsBuilder().ForLabels(c.lbs, c.lbs.Hash())
	for _, line := range c.input {
```

pkg/logql/log/distinct_test.go
```diff
@@ -0,0 +1,61 @@
...
+		NewJSONParser().Process(1, []byte(line), b)
+		_, pass := distinctFilter.Process(1, []byte(line), b)
+		if pass {
+			total++
+			passLines = append(passLines, line)
```
DylanGuedes, line 54: does `pass` here mean `filter out`? if so, WDYT of naming `total` as `filteredLines` or `droppedLines` or something similar? I got a little confused if `pass` is "pass this line forwards" (i.e: do not drop it) or "pass/skip this line"

pkg/logql/log/distinct_test.go
```diff
@@ -0,0 +1,61 @@
...
+	c := struct {
+		name          string
+		label         []string
+		lbs           labels.Labels
+		input         []string
+		expectedCount int
+		expectedLines []string
+	}{
+		name:  "distinct test",
+		label: []string{"id", "time"},
```
DylanGuedes, line 23: in case it makes sense: could you add to this list a label that won't be present on the given input?

DylanGuedes (changes requested):
a few questions but tbh this looks good

pkg/logql/log/distinct.go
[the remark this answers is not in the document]
jeschkies: Hm, but you'd still have a nil pointer right?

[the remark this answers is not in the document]
liguozhong: good point.This code does look dangerous.
No, each label has been initialized in the `NewDistinctFilter` () function.
```go
func NewDistinctFilter(labels []string) (Stage, error) {
	datas := make(map[string]map[string]int, 0)
	for _, label := range labels {
		datas[label] = make(map[string]int, 0)
	}
	...
}
```
If `distinctFilter` struct initialized through `NewDistinctFilter`(), the `distinctFilter` struct will never get nil point. But if we directly build a `distinctFilter` struct with &`distinctFilter`{}, it is possible that the nil point you mentioned will appear.

another way. 
image
this for performance, it will not need to make a judgment like this for each log.

```go
 _, ok = r.datas[label]
if !ok {
r.datas[label] = make(map[string]int, 0)
}
```

pkg/logql/log/distinct_test.go
liguozhong: ok,done

pkg/logql/log/distinct_test.go
liguozhong: pass is "pass this line forwards"

I searched and most of the test cases use `ok`

image

pkg/logql/log/distinct.go
liguozhong: pass is "pass this line forwards"

It looks like this var name is misunderstood, can you help me with a better var name

pkg/logql/log/distinct_test.go
liguozhong: good point. After adding a label that does not exist, in order to make the test pass, I have to modify `false` to `true`, please help review whether this modification is correct

image
image

liguozhong:
thanks, ready for review.

pkg/logql/log/distinct.go
DylanGuedes: I think "keep" is a little more clear

pkg/logql/log/distinct_test.go
DylanGuedes: got it, thank you. "ok" is good for the consistency", "keep" or something similar is good for clarity; whatever you decide I'll be happy with :smile:

liguozhong: 👌, ☑️ "ok"

pkg/logql/log/distinct.go
liguozhong: done, thanks  😃

pkg/logql/log/distinct_test.go
liguozhong: done

pkg/logql/log/distinct.go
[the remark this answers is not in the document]
jeschkies: Is `NewDestinctFilter` called for each line? How is preserving the state?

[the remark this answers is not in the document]
liguozhong: `NewDestinctFilter` is generated for each `logql`, not for each line. So the `state` in `NewDestinctFilter` can be maintained continuously in one logql. The screenshot in the PR can also show that this `state` is valid

image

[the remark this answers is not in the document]
jeschkies: Thanks for your patience. I still don't understand how `data[label]` can be not nil if the label not known when `NewDestinctFilter` is called. I'll try out.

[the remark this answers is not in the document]
jeschkies: Ah, no I get it. The labels are not the labels seen in the line but the ones passed to `distinct` :see_no_evil:

pkg/logql/syntax/ast_test.go
```diff
@@ -358,6 +360,13 @@ func Test_FilterMatcher(t *testing.T) {
 			},
 			[]linecheck{{"duration=5m total_bytes=5kB", true}, {"duration=1s total_bytes=256B", false}, {"duration=0s", false}},
 		},
+		{
+			`{app="foo"} | logfmt | distinct id`,
+			[]*labels.Matcher{
+				mustNewMatcher(labels.MatchEqual, "app", "foo"),
+			},
+			[]linecheck{{"id=foo", true}, {"id=foo", false}, {"id=bar", true}},
+		},
```
jeschkies, line 369: @liguozhong I hope you don't mind me adding this test.

jeschkies (approved):
@DylanGuedes I'm fine with this change. What do you say?

liguozhong:
thanks

DylanGuedes:
> @DylanGuedes I'm fine with this change. What do you say?

same :smile: 

@liguozhong thank you again for another awesome contribution :heart:

xkkker:
Hi,what's the milestone of this feature?  The lastest version 2.8.4 don't support distinct for now.

PR #8665, opened 2023-03-01
deletion: fix issue in processing delete requests with tsdb index

sandeepsukhani:
When a delete request creates a new chunk from an existing chunk, we index it to the index store.
In TSDB, we [drop](https://github.com/grafana/loki/blob/944b488bd2506012d149fc363401f9b7d51af3e7/pkg/storage/stores/tsdb/head_manager.go#L235-L237) `__name__` label, which was being added to chunkrefs for historical reasons.
However, this was not done while indexing chunks from the deletion code, causing deletion to fail with the error [chunk insertion is only allowed on existing streams](https://github.com/grafana/loki/blob/619c2212f58c645661c7a004c084b839e91aec1b/pkg/storage/stores/tsdb/builder.go#L67).

This PR handles the issue by dropping the `__name__` label and adding relevant tests.

I will add an integration test for deletion with TSDB once we have PR https://github.com/grafana/loki/pull/7754 merged since it does some groundwork to run integration tests with both `boltdb-shipper` and `tsdb`.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated


MichelHollands (approved):
LGTM

PR #8672, opened 2023-03-01
operator: Add support for memberlist bind network configuration

periklis:
This PR adds support to configure the memberlist `bind_addr` field using the pod network IP range instead of the default private network range used (See ). In managed Kubernetes/OpenShift cluster environments as well as in special on-prem setup the private IP range might not be available for using them. With this change set the LokiStack administrator can choose as a bind address the current pod network IP assigned by the cluster's pod network. 

```yaml
apiVersion: loki.grafana.com/v1
kind: LokiStack
metadata:
  name: lokistack-dev
  namespace: openshift-logging
spec:
  size: 1x.small
  hashRing:
    type: memberlist
    memberlist:
      instanceAddrType: podIP
  storage:
    schemas:
    - version: v12
      effectiveDate: 2022-06-01
    secret:
      name: test
      type: s3
  storageClassName: gp2-csi
  tenants:
    mode: openshift-logging
```

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


Red-GV (commented):
lgtm

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -345,6 +345,58 @@ type ClusterProxy struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:memberlist"},displayName="Type"
+	// +kubebuilder:default:=memberlist
+	Type HashRingType `json:"type"`
+
+	// MemberList configuration spec
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Memberlist Config"
+	MemberList *MemberListSpec `json:"memberlist,omitempty"`
```
xperimental, line 397: `memberlist` or `memberList`? Loki configuration seems to use the first, so current code should be fine.

operator/internal/manifests/internal/config/loki-config.yaml
```diff
@@ -166,9 +166,15 @@ limits_config:
   split_queries_by_interval: 30m
 memberlist:
   abort_if_cluster_join_fails: true
-  bind_port: {{ .GossipRing.Port }}
+  {{- with .GossipRing.BindAddrs }}
+  bind_addr:
+    {{- range . }}
+    - {{ . }}
+    {{- end }}
+  {{- end }}
```
xperimental, line 174: Couldn't this be much simpler because we only have two different possible values (any/default or "pod IP")? (`BindPodIP` is `bool`)

```suggestion
  {{- with .GossipRing.BindPodIP }}
  bind_addr:
    - ${POD_IP}
  {{- end }}
```

operator/internal/manifests/internal/config/loki-config.yaml
```diff
@@ -166,9 +166,15 @@ limits_config:
...
+  bind_port: {{ .GossipRing.BindPort }}
   join_members:
-    - {{ .GossipRing.FQDN }}:{{ .GossipRing.Port }}
+    - {{ .GossipRing.MembersDiscoveryAddr }}:{{ .GossipRing.BindPort }}
```
xperimental, line 177: Does it make sense to put the whole "memberlist configuration block" into a `{{ With .GossipRing }}`?

operator/internal/manifests/var.go
```diff
@@ -19,6 +19,8 @@ const (
 	grpcPort         = 9095
 	protocolTCP      = "TCP"
 
+	gossipBindAddrEnvVarName = "POD_IP"
```
xperimental, line 22: I'd probably prefer the environment variable being named after the "purpose" and not the "content". I know that it contains the "pod IP", but we are using it to set a bind address (or maybe an advertised address). That should make it clearer in debug output.

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -674,6 +726,13 @@ type LokiStackSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:1x.extra-small","urn:alm:descriptor:com.tectonic.ui:select:1x.small","urn:alm:descriptor:com.tectonic.ui:select:1x.medium"},displayName="LokiStack Size"
 	Size LokiStackSizeType `json:"size"`
 
+	// HashRing defines the spec for the distributed hash ring configuration.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Hash Ring"
+	HashRing *HashRingSpec `json:"hashRing,omitempty"`
```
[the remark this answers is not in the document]
periklis: Members discovery is actually done by DNS which is an underlying detail of the hash ring. Here we concentrate on members forming a hash ring. In fact the Loki config has a hash_ring section which should be represented here

operator/apis/loki/v1/lokistack_types.go
periklis: Yes I used the non-camel case format for the yaml field here to keep consistency with Loki

operator/apis/loki/v1/lokistack_webhook.go
```diff
@@ -50,6 +50,27 @@ func (r *LokiStack) ValidateDelete() error {
...
+	var allErrs field.ErrorList
+
+	if hrs == nil {
+		return nil
+	}
+
+	switch hrs.Type {
+	case HashRingMemberList:
+		if hrs.MemberList == nil {
+			allErrs = append(allErrs, field.NotFound(field.NewPath("Spec").Child("HashRing").Child("Memberlist"), hrs.MemberList))
```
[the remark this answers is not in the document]
periklis: I agree will amend asap

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -345,6 +345,58 @@ type ClusterProxy struct {
...
+	// BindNetworkType defines the type of network to use to advertise IPs to the ring.
+	// Defaults to the private network available to the current pod. Alternatively the
+	// pod public network can be used in case private networks (RFC 1918 and RFC 6598)
+	// are not available.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:private","urn:alm:descriptor:com.tectonic.ui:select:public"},displayName="Hash Ring Type"
+	// +kubebuilder:default:=private
+	BindNetworkType MemberListNetworkType `json:"bindNetworkType"`
```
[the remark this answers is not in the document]
periklis: > Two questions come up because of this:
> 
>     * Do we actually have to change the bind? The memberlist complains of not finding an "advertise address", looking at the [Loki docs](https://grafana.com/docs/loki/latest/configuration/) the `instance_addr` configuration field seems like a better choice for specifically setting the address advertised in the ring. (Have not looked at the actual implementation yet though, so keep a grain of salt ready.)

I have to admit that `instance_addr` is new to me and in general the `common.ring` section is not obvious to me on how it interacts with the `memberlist_config`. Later is/was the no.1 section for the gossip ring configured once for all ring-related component sets: 1) distributor, ingester, 2) querier, ingester,  3) ruler and 4) compactor (last not used in the Loki Operator as per compactor single replica for all tenants applied). 

FWIW I need to dig into the Loki code base in how `instance_addr` and `bind_addr` interact with each other. Another disturbing fact about the `instance_addr` setting is the following comment on the Loki docs:

```
# If "instance_addr" under the common ring section is configured, this common "instance_addr"
# is only applied to the frontend, but not for ring related components (ex: distributor, ruler, etc).
```

> 
>     * If we want to keep setting the bind address: Does it make sense to make the setting "more explicit", for example by having `any` and `podIP` as possible settings instead of the current `private`/`public`?

I believe your suggestion is easier to comprehend and will consider switching to enum options `any` and `podIP` after investigating the interaction of `instance_addr` and `bind_addr`.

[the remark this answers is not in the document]
periklis: Leaving this as a note here, none conclusions yet, but it looks like `bind_addr` is the first fallback if we don't provide `instance_addr` in the ring configuration for `memberlist`: https://github.com/grafana/dskit/blob/main/kv/memberlist/tcp_transport.go#L347-L397

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -345,6 +345,58 @@ type ClusterProxy struct {
...
+// +kubebuilder:validation:Enum=memberlist
+type HashRingType string
+
+const (
+	// HashRingMemberList when using memberlist for the distributed hash ring.
+	HashRingMemberList HashRingType = "memberlist"
+)
+
+// MemberListInstanceAddrType defines the type of pod network to use for advertising IPs to the ring.
+type MemberListInstanceAddrType string
```
xperimental, line 359: Shorten the type name to `InstanceAddrType`, so that the full name can be used in the constants below?

operator/internal/manifests/config.go
```diff
@@ -247,6 +244,26 @@ func alertManagerConfig(spec *lokiv1beta1.AlertManagerSpec) *config.AlertManager
 	return conf
 }
 
+func gossipRingConfig(stackName, stackNs string, spec *lokiv1.HashRingSpec) config.GossipRing {
+	var instanceAddr string
+	if spec != nil && spec.Type == lokiv1.HashRingMemberList {
+		switch spec.MemberList.InstanceAddrType {
+		case lokiv1.MemberListPodIP:
+			instanceAddr = fmt.Sprintf("${%s}", gossipInstanceAddrEnvVarName)
```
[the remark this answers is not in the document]
periklis: The `gossipInstanceAddrEnvVarName` is used also in the `corev1.EnvVar` definitions that's why I kept this like this here.

periklis (commented):
Self-review

xperimental (commented):
I think I tried all the permutations of specifying options now... just a small improvement to the validation I came across and then we're good :slightly_smiling_face:

xperimental (approved):

PR #8682, opened 2023-03-02
Querier/Ruler: add histogram to track fetched chunk size distribution

dannykopping:
We are looking at making some changes to the chunk caching strategy, and we need this data to know what size chunks we're typically requesting. For example, if the overwhelming majority of chunks are small we may decide to cache only small chunks to decrease the number of requests to the object store; we could fit more chunks in cache this way, which may have a positive performance impact.

N/A

I picked buckets that I thought would be useful, and even included a bucket that is above the recommended chunk target size since some users may be setting it that high.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/storage/chunk/fetcher/fetcher.go
+++ b/pkg/storage/chunk/fetcher/fetcher.go
@@ -8,6 +8,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/promql"
 
 	"github.com/grafana/loki/pkg/logqlmodel/stats"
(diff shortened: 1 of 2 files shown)


pkg/storage/chunk/fetcher/fetcher.go
```diff
@@ -173,6 +183,11 @@ func (c *Fetcher) FetchChunks(ctx context.Context, chunks []chunk.Chunk, keys []
 	if err != nil {
 		level.Warn(log).Log("msg", "error fetching from cache", "err", err)
 	}
+
+	for _, buf := range cacheBufs {
+		chunkFetchedSize.With(labels.FromStrings("source", "cache").Map()).Observe(float64(len(buf)))
```
cyriltovena, line 188: `WithValues` should be more performant than creating a new labels set and turning it into a map.

cyriltovena (approved):
LGTM

Please consider my comment.

sandeepsukhani (approved):
left a minor non-blocking nit, but otherwise it LGTM

pkg/storage/chunk/fetcher/fetcher.go
sandeepsukhani: hah, you beat me

vlad-diachenko (approved):
LGTM

pkg/storage/chunk/fetcher/fetcher.go
dannykopping: Doh! Thanks for spotting this

PR #8683, opened 2023-03-02
operator: Add s3-virtual-hosted ObjectStorage type

### LINKED ISSUE #8638: operator: Allow disabling the s3forcepathstyle option for S3 Storage in CRD
The AlibabaCloud OSS service is not compatible s3 path like accessible, that is only domain style, so we need to set s3forcepathstyle is off. But the operator is not support s3forcepathstyle for S3 storage.

enable option(s3forcepathstyle) in CRD and Loki-config.yaml

current we just hardcode  Loki-config.yaml and repackage operator image for our cluster

**Additional context**

xuanyunhui:
Add a new Object Storage Type to loki-operator for domain base only s3-compatible storage
Fixes #8638

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


periklis (commented):
This is really awesome! Looks overall good to me!

periklis (changes requested):
There is one step missing when you edit the types namely `make bundle-all` and committing the generated files.

periklis (commented):
Looks good to me

periklis:
@xuanyunhui You can use `make fmt` to format your code changes before pushing. This will make the linter happy ;)

periklis:
@xuanyunhui PTAL some of the config tests are failing

operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -25,6 +25,9 @@ func ExtractSecret(s *corev1.Secret, secretType lokiv1.ObjectStorageSecretType)
 		storageOpts.S3, err = extractS3ConfigSecret(s)
 	case lokiv1.ObjectStorageSecretSwift:
 		storageOpts.Swift, err = extractSwiftConfigSecret(s)
+	case lokiv1.ObjectStorageSecretS3VirtualHosted:
+		storageOpts.S3VirtualHosted, err = extractS3ConfigSecret(s)
+		storageOpts.SharedStore = lokiv1.ObjectStorageSecretS3
```
periklis, line 30: IMHO I liked the previous mapping of the `s3-virtual-hosted` type into `s3` because it reuses the config rendering below.

periklis (changes requested):
Although we introduce a new object storage type `s3-virtual-hosted` on the CRD level, from an low-level implementation standpoint it is still just the `storage.Options.S3` type. I suggest to keep the mapping of `s3-virtual-hosted` into `s3` introduced in a previous revision of this PR.

operator/internal/handlers/internal/storage/secrets.go
xuanyunhui: Do you mean revert to previous commit？

periklis: Yes please

xuanyunhui: I'm happy to do this!

xperimental:
The style of S3 URLs seems like an implementation detail on the S3 storage. AWS was planning to move away from path-based buckets as well but seems to have post-poned this decision. I don't know if we want to introduce a whole now "storage type" just for this difference, I think the current setting of "path-based = true" is an artifact of how we tested the operator in the beginning.

So in summary: would adding an option to the existing S3 configuration which decides whether the vhost-based or the path-based style is used maybe be a simpler option?

xuanyunhui:
> The style of S3 URLs seems like an implementation detail on the S3 storage. AWS was planning to move away from path-based buckets as well but seems to have post-poned this decision. I don't know if we want to introduce a whole now "storage type" just for this difference, I think the current setting of "path-based = true" is an artifact of how we tested the operator in the beginning.
> 
> So in summary: would adding an option to the existing S3 configuration which decides whether the vhost-based or the path-based style is used maybe be a simpler option?

@periklis WDYT? If only need adding an option to the existing S3 configuration, I glad to change the PR.

periklis:
> > The style of S3 URLs seems like an implementation detail on the S3 storage. AWS was planning to move away from path-based buckets as well but seems to have post-poned this decision. I don't know if we want to introduce a whole now "storage type" just for this difference, I think the current setting of "path-based = true" is an artifact of how we tested the operator in the beginning.
> > So in summary: would adding an option to the existing S3 configuration which decides whether the vhost-based or the path-based style is used maybe be a simpler option?
> 
> @periklis WDYT? If only need adding an option to the existing S3 configuration, I glad to change the PR.

I trust @xperimental judgement here. If you don't mind just make it a extra option (default set to true) in the secret. Let's remove the additional changes in the CRD since you are going to use this temporarily until we get AlibabaCloud support released with Loki. Thanks in advance for all the effort put in here.

xperimental (commented):
Added a little more detail.

operator/internal/manifests/storage/options.go
```diff
@@ -34,11 +35,12 @@ type GCSStorageConfig struct {
...
-	Region          string
-	Buckets         string
-	AccessKeyID     string
-	AccessKeySecret string
+	Endpoint         string
+	Region           string
+	Buckets          string
+	AccessKeyID      string
+	AccessKeySecret  string
+	EnforcePathStyle bool
```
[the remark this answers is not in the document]
xuanyunhui: I will try it according to my understanding.

xuanyunhui:
I made a mistake on my branch, so I closed this PR and Opened a new #8766

Issue #8419, opened 2023-02-06
Write path lifecycle autoscaling


wiardvanrij:
As per: https://github.com/grafana/loki/blob/24accf6a5b1ee1706e3fc6adf82b4940ba7db5e1/docs/sources/operations/storage/wal.md#how-to-scale-updown

It would be nice to have the same option as was implemented in Cortex here: https://github.com/cortexproject/cortex-helm-chart/blob/53fcb8f35a0fe8b46bb064cd2ec25a678f401db7/values.yaml#L497 in the Loki helm chart.

This would allow us to be a bit more creative in scaling ingesters automatically

JStickler:
Closing this issue as already built. The Loki Helm chart now includes autoscaling support for the write path with configurable min/max replicas and automatic lifecycle hook management (including `/flush_shutdown` preStop hooks for ingesters). See the `write.autoscaling` section in `production/helm/loki/values.yaml`.

PR #8684, opened 2023-03-02
Helm: add hpa for read, write and backend

### LINKED ISSUE #7864: Helm chart of loki-read simple scalable lacks HPA (HorizontalPodAutoscaler)
**Describe the bug**
The HPA (HorizontalPodAutoscaler) attributes are mentioned in [values](./production/helm/loki/values.yaml) and [reference](./docs/sources/installation/helm/reference.md) docs, as `read.autoscaling.enabled` for example.

However there's no reference of any HorizontalPodAutoscaler kind nor those attributes in **read** part of chart. The gateway part is has HPA explicitly as an example.

**To Reproduce**
Look at the chart files or try to deploy HPA for **read** Loki pods.

**Expected behavior**
Set HorizontalPodAutoscaler kind or provide an alternative to scale `loki-read` StatefulSet.

Or maybe the HPA was there, bug got removed for some reason and only docs are not being cleaned up?

**Environment:**
 - Infrastructure: Kubernetes, simple-scalable deployment
 - Deployment tool: Helm via FluxCD in my case

N/A

oleksii-boiko-ua:
It adds missing HPA (HorizontalPodAutoscaler) templates to write, read, and backend components 
Fixes #7864

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/operations/storage/wal.md
+++ b/docs/sources/operations/storage/wal.md
@@ -95,6 +95,18 @@ Statefulsets are significantly more cumbersome to work with/upgrade/etc. Much of
 
 In this case, try `kubectl -n <namespace> delete sts ingester --cascade=false`. This will leave the pods alive but delete the statefulset. Then you may recreate the (updated) statefulset and one-by-one start deleting the `ingester-0` through `ingester-n` pods _in that order_, allowing the statefulset to spin up new pods to replace them.
 
+#### Scaling Down Using `/flush_shutdown` Endpoint and Lifecycle Hook
+
+1. **StatefulSets for Ordered Scaling Down**: Loki's ingesters should be scaled down one by one, which is efficiently handled by Kubernetes StatefulSets. This ensures an ordered and reliable scaling process, as described in the [Deployment and Scaling Guarantees](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#deployment-and-scaling-guarantees) documentation.
+
+2. **Using PreStop Lifecycle Hook**: During the pod scaling down process, the PreStop [lifecycle hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) triggers the `/flush_shutdown` endpoint on the ingester. This action flushes the chunks and removes the ingester from the ring, allowing it to register as unready and become eligible for deletion.
+
+3. **Using terminationGracePeriodSeconds**: Provides time for the ingester to flush its data before being deleted, if flushing data takes more than 30 minutes, you may need to increase it.
+
+4. **Cleaning Persistent Volumes**: Persistent volumes are automatically cleaned up by leveraging the [enableStatefulSetAutoDeletePVC](https://kubernetes.io/blog/2021/12/16/kubernetes-1-23-statefulset-pvc-auto-deletion/) feature in Kubernetes.
+
+By following the above steps, you can ensure a smooth scaling down process for Loki's ingesters while maintaining data integrity and minimizing potential disruptions.
+
 ### Non-Kubernetes or baremetal deployments
 
 * When the ingester restarts for any reason (upgrade, crash, etc), it should be able to attach to the same volume in order to recover back the WAL and tokens.
--- a/production/helm/loki/templates/write/hpa.yaml
+++ b/production/helm/loki/templates/write/hpa.yaml
@@ -1,5 +1,6 @@
+{{- $isSimpleScalable := eq (include "loki.deployment.isScalable" .) "true" -}}
 {{- $autoscalingv2 := .Capabilities.APIVersions.Has "autoscaling/v2" -}}
-{{- if .Values.gateway.autoscaling.enabled }}
+{{- if and $isSimpleScalable ( .Values.write.autoscaling.enabled ) }}
 {{- if $autoscalingv2 }}
 apiVersion: autoscaling/v2
 {{- else }}
@@ -7,23 +8,22 @@ apiVersion: autoscaling/v2beta1
 {{- end }}
 kind: HorizontalPodAutoscaler
 metadata:
-  name: {{ include "loki.gatewayFullname" . }}
-  namespace: {{ $.Release.Namespace }}
+  name: {{ include "loki.writeFullname" . }}
   labels:
-    {{- include "loki.gatewayLabels" . | nindent 4 }}
+    {{- include "loki.writeLabels" . | nindent 4 }}
 spec:
   scaleTargetRef:
     apiVersion: apps/v1
-    kind: Deployment
-    name: {{ include "loki.gatewayFullname" . }}
-  minReplicas: {{ .Values.gateway.autoscaling.minReplicas }}
-  maxReplicas: {{ .Values.gateway.autoscaling.maxReplicas }}
-  {{- with .Values.gateway.autoscaling.behavior }}
+    kind: StatefulSet
+    name: {{ include "loki.writeFullname" . }}
+  minReplicas: {{ .Values.write.autoscaling.minReplicas }}
+  maxReplicas: {{ .Values.write.autoscaling.maxReplicas }}
+  {{- with .Values.write.autoscaling.behavior }}
   behavior:
     {{- toYaml . | nindent 4 }}
   {{- end }}
   metrics:
-  {{- with .Values.gateway.autoscaling.targetMemoryUtilizationPercentage }}
+  {{- with .Values.write.autoscaling.targetMemoryUtilizationPercentage }}
     - type: Resource
       resource:
         name: memory
@@ -35,7 +35,7 @@ spec:
         targetAverageUtilization: {{ . }}
         {{- end }}
   {{- end }}
-  {{- with .Values.gateway.autoscaling.targetCPUUtilizationPercentage }}
+  {{- with .Values.write.autoscaling.targetCPUUtilizationPercentage }}
     - type: Resource
       resource:
         name: cpu
--- a/production/helm/loki/templates/write/statefulset-write.yaml
+++ b/production/helm/loki/templates/write/statefulset-write.yaml
@@ -9,9 +9,10 @@ metadata:
     {{- include "loki.writeLabels" . | nindent 4 }}
     app.kubernetes.io/part-of: memberlist
 spec:
+{{- if not .Values.write.autoscaling.enabled }}
   replicas: {{ .Values.write.replicas }}
-
-  podManagementPolicy: Parallel
+{{- end }}
+  podManagementPolicy: OrderedReady
   updateStrategy:
     rollingUpdate:
       partition: 0
@@ -101,9 +102,15 @@ spec:
             {{- toYaml .Values.loki.containerSecurityContext | nindent 12 }}
           readinessProbe:
             {{- toYaml .Values.loki.readinessProbe | nindent 12 }}
-          {{- with .Values.write.lifecycle }}
+          {{- if .Values.write.lifecycle }}
           lifecycle:
-            {{- toYaml . | nindent 12 }}
+            {{- toYaml .Values.write.lifecycle | nindent 12 }}
+          {{- else if .Values.write.autoscaling.enabled }}
+          lifecycle:
+            preStop:
+              httpGet:
+                path: "/ingester/flush_shutdown"
+                port: http-metrics
           {{- end }}
           volumeMounts:
             - name: config
--- a/production/helm/loki/values.yaml
+++ b/production/helm/loki/values.yaml
@@ -645,6 +645,31 @@ monitoring:
 write:
   # -- Number of replicas for the write
   replicas: 3
+  autoscaling:
+    # -- Enable autoscaling for the write.
+    enabled: false
+    # -- Minimum autoscaling replicas for the write.
+    minReplicas: 1
+    # -- Maximum autoscaling replicas for the write.
+    maxReplicas: 3
+    # -- Target CPU utilisation percentage for the write.
+    targetCPUUtilizationPercentage: 60
+    # -- Target memory utilization percentage for the write.
+    targetMemoryUtilizationPercentage:
+    # -- Behavior policies while scaling.
+    behavior:
+    # -- see https://github.com/grafana/loki/blob/main/docs/sources/operations/storage/wal.md#how-to-scale-updown for scaledown details
+      scaleUp:
+        policies:
+          - type: Pods
+            value: 1
+            periodSeconds: 900
+      scaleDown:
+        policies:
+          - type: Pods
+            value: 1
+            periodSeconds: 1800
+      stabilizationWindowSeconds: 3600
   image:
     # -- The Docker registry for the write image. Overrides `loki.image.registry`
     registry: null
@@ -672,6 +697,11 @@ write:
   extraEnvFrom: []
   # -- Lifecycle for the write container
   lifecycle: {}
+  # -- The default /flush_shutdown preStop hook is recommended as part of the ingester
+  # scaledown process so it's added to the template by default when autoscaling is enabled,
+  # but it's disabled to optimize rolling restarts in instances that will never be scaled
+  # down or when using chunks storage with WAL disabled.
+  # https://github.com/grafana/loki/blob/main/docs/sources/operations/storage/wal.md#how-to-scale-updown
   # -- Init containers to add to the write pods
   initContainers: []
   # -- Volume mounts to add to the write pods
@@ -784,6 +814,20 @@ read:
     targetCPUUtilizationPercentage: 60
     # -- Target memory utilisation percentage for the read
     targetMemoryUtilizationPercentage:
+    # -- Behavior policies while scaling.
+    behavior: {}
+      #  scaleUp:
+      #   stabilizationWindowSeconds: 300
+      #   policies:
+      #   - type: Pods
+      #     value: 1
+      #     periodSeconds: 60
+      #  scaleDown:
+      #   stabilizationWindowSeconds: 300
+      #   policies:
+      #   - type: Pods
+      #     value: 1
+      #     periodSeconds: 180
   image:
     # -- The Docker registry for the read image. Overrides `loki.image.registry`
     registry: null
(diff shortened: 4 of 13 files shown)


jeschkies:
Thanks. We had plans to add autoscaling but didn't knew when. @trevorwhitney can the backend be scaled? I thought it was stateful. @salvacorts  could you take a look as well?

oleksii-boiko-ua:
BTW QQ, when are we going to have `Three Scalable Targets` supported and merged into the main branch? I saw [this guide](https://github.com/grafana/loki/blob/main/docs/sources/installation/helm/migrate-to-three-scalable-targets/index.md) which  refers to `main-f5fbfab-amd64` docker tag which is 2 month old release

production/helm/loki/values.yaml
```diff
@@ -604,6 +604,17 @@ monitoring:
 write:
   # -- Number of replicas for the write
   replicas: 3
+  autoscaling:
+    # -- Enable autoscaling for the write
```
JStickler, line 608:
```suggestion
    # -- Enable autoscaling for the write.
```

production/helm/loki/values.yaml
```diff
@@ -604,6 +604,17 @@ monitoring:
...
+    enabled: false
+    # -- Minimum autoscaling replicas for the write
```
JStickler, line 610:
```suggestion
    # -- Minimum autoscaling replicas for the write.
```

production/helm/loki/values.yaml
```diff
@@ -604,6 +604,17 @@ monitoring:
...
+    minReplicas: 1
+    # -- Maximum autoscaling replicas for the write
```
JStickler, line 612:
```suggestion
    # -- Maximum autoscaling replicas for the write.
```

production/helm/loki/values.yaml
```diff
@@ -604,6 +604,17 @@ monitoring:
...
+    maxReplicas: 3
+    # -- Target CPU utilisation percentage for the write
```
JStickler, line 614:
```suggestion
    # -- Target CPU utilization percentage for the write.
```

production/helm/loki/values.yaml
```diff
@@ -604,6 +604,17 @@ monitoring:
...
+    targetCPUUtilizationPercentage: 60
+    # -- Target memory utilisation percentage for the write
```
JStickler, line 616:
```suggestion
    # -- Target memory utilization percentage for the write.
```

production/helm/loki/values.yaml
```diff
@@ -810,6 +821,17 @@ read:
 backend:
   # -- Number of replicas for the backend
   replicas: 3
+  autoscaling:
+    # -- Enable autoscaling for the backend`
```
JStickler, line 825:
```suggestion
    # -- Enable autoscaling for the backend.
```

production/helm/loki/values.yaml
```diff
@@ -810,6 +821,17 @@ read:
...
+    enabled: false
+    # -- Minimum autoscaling replicas for the backend
```
JStickler, line 827:
```suggestion
    # -- Minimum autoscaling replicas for the backend.
```

production/helm/loki/values.yaml
```diff
@@ -810,6 +821,17 @@ read:
...
+    minReplicas: 1
+    # -- Maximum autoscaling replicas for the backend
```
JStickler, line 829:
```suggestion
    # -- Maximum autoscaling replicas for the backend.
```

production/helm/loki/values.yaml
```diff
@@ -810,6 +821,17 @@ read:
...
+    maxReplicas: 3
+    # -- Target CPU utilisation percentage for the backend
+    targetCPUUtilizationPercentage: 60
+    # -- Target memory utilisation percentage for the backend
```
JStickler, line 833:
```suggestion
    # -- Target memory utilization percentage for the backend.
```

JStickler (commented):
[Docs squad]  Added missing punctuation and we've standardized on American spelling in the docs.

production/helm/loki/templates/write/hpa.yaml
```diff
@@ -0,0 +1,46 @@
...
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ include "loki.writeFullname" . }}
+  labels:
+    {{- include "loki.writeFullname" . | nindent 4 }}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: StatefulSet
+    name: {{ include "loki.writeFullname" . }}
```
kd7lxl, line 18: It would be helpful to support [`behavior:`](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#configurable-scaling-behavior) here, to control how dramatic the scaling is, since the writer is stateful. It could also be used to disable scale down.

oleksii-boiko-ua:
@JStickler @kd7lxl @jeschkies updated pr. thanks

oleksii-boiko-ua:
@kd7lxl should we add behavior to all(read, backend) hpa templates?

kd7lxl:
> @kd7lxl should we add behavior to all(read, backend) hpa templates?

Consistency is the least surprising option, so from that standpoint it would be good to add the same options to all. I don't have an immediate use case for behavior on read, but someone else might.

Thanks for working on this! I needed it today and was about to get started writing it when I found your PR.

production/helm/loki/values.yaml
```diff
@@ -604,6 +604,31 @@ monitoring:
...
   # -- Number of replicas for the write
   replicas: 3
+  autoscaling:
+    # -- Enable autoscaling for the write.
+    enabled: false
+    # -- Minimum autoscaling replicas for the write.
+    minReplicas: 1
+    # -- Maximum autoscaling replicas for the write.
+    maxReplicas: 3
+    # -- Target CPU utilisation percentage for the write
```
JStickler, line 614:
```suggestion
    # -- Target CPU utilisation percentage for the write.
```

JStickler (approved):
[Docs squad]  One more missing period, but other than that, LGTM

oleksii-boiko-ua:
updated, build failure seems not related

oleksii-boiko-ua:
@trevorwhitney  @jeschkies @salvacorts could anyone review it, please?

trevorwhitney (changes requested):
can you walk me through what it's like to scale the `write` and `backend` pods? I imagine automatically scaling up is easier than going down, due to their stateful nature. I love this for `read`, since that's a deployment now. I'm not against this for `write` and `backend`, but could you provide some documentation on how you would scale those components without any data-loss/downtime?

kd7lxl:
> can you walk me through what it's like to scale the `write` and `backend` pods?

Have you seen the doc? https://github.com/grafana/loki/blob/main/docs/sources/operations/storage/wal.md#how-to-scale-updown

production/helm/loki/templates/write/statefulset-write.yaml
```diff
@@ -23,7 +24,7 @@ spec:
     operation easier, and will rely on re-fetching data when needed.
   */}}
   persistentVolumeClaimRetentionPolicy:
-    whenDeleted: Delete
+    whenDeleted: Deleteq
```
trevorwhitney, line 27: typo
```suggestion
    whenDeleted: Delete
```

trevorwhitney:
@kd7lxl I have, but that doesn't really answer my question for how this is handled during autocaling. For exmaple, how do you ensure that ingesters flush to storage when they're scaling down? Forgive me if I missed something, but I don't see this PR modifying the `flush_on_shutdown` config, nor calling the flush endpoint (if that's even possible with HPAs?).

In addition to making sure we programmatically handle this, I also think it's important to call out in the Loki operational documentation that enabling auto-scaling will turn on this behavior and clarify how the WAL is handled, and how PVs and PVCs of scaled down ingesters can be treated.

oleksii-boiko-ua:
Hi there, we could use [lifecycle hooks](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/), for example pre stop, and hit endpoint ingester/flush_shutdown before container goes down, as for pvc it will be cleaned automatically

kd7lxl:
> Hi there, we could use [lifecycle hooks](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/), for example pre stop, and hit endpoint ingester/flush_shutdown before container goes down, as for pvc it will be cleaned automatically

There is a request in #8419 to add the lifecycle hook, like I did for the Cortex helm chart. This isn't a perfect solution because it runs every time the pod stops, even though it's unnecessary for typical pod restarts where the WAL is resumed. It's only useful for scaledown events. If you were to configure your HPA behavior to disable scaledown, the lifecycle hook would not be needed. However, if you want to support scaledown it works great in Cortex.

oleksii-boiko-ua:
you could always have conditional lifecycle hook right?

kd7lxl:
> you could always have conditional lifecycle hook right?

Yes, you could do it like the Cortex chart where it's all in values so the right combination of values overrides will disable scaledown and the lifecycle hook, or you could write an explicit condition (like `.write.autoscaling.scaledown.enabled`) for it that automatically injects all the corresponding settings.

oleksii-boiko-ua:
@trevorwhitney updated

production/helm/loki/values.yaml
```diff
@@ -629,6 +657,15 @@ write:
   extraEnvFrom: []
   # -- Lifecycle for the write container
   lifecycle: {}
+    # -- The /flush_shutdown preStop hook is recommended as part of the ingester
```
trevorwhitney, line 660: instead of just a comment, how would you feel about taking an opinionated stance here? 

I would like as a default to set the `lifecycle` on line 107 of the write template to have the `preStop` hook, but with an option to override via `values.yaml`. For example, maybe we make the default value here `null`, and have a template that, if `null` and autoscaling, we include the default `preStop`, but if not `null` and autoscaling, we include whatever was provided in the `values.yaml`.

similarly, if not `null` and not autoscaling, we can include whatever was provided, but if `null` and not autoscaling we don't include anything?

production/helm/loki/values.yaml
```diff
@@ -739,6 +776,20 @@ read:
     targetCPUUtilizationPercentage: 60
     # -- Target memory utilisation percentage for the read
     targetMemoryUtilizationPercentage:
+    # -- Behavior policies while scaling.
+    behavior: {}
```
trevorwhitney, line 780: again with behavior, how would you feel about taking a more opinionated stance with good defaults and the ability to override?

trevorwhitney (changes requested):
This looks great, thank you for making those changes! I'm much happier with this approach using the `preStop` hook and really appreciate the added documentation, thank you!

My last little nit is I'd like to provide more opinionated defaults. See my comments, but I'd like to inject more logic to automatically set the `lifecycle` and `behavior` values to good defaults when using autoscaling rather than hiding best practices in a comment. Of course we need to preserve the ability to change them or not set them at all, but if someone turns on "yes I want autoscaling" let's give them sensible defaults. WDYT?

production/helm/loki/values.yaml
oleksii-boiko-ua: sure, updated

production/helm/loki/values.yaml
oleksii-boiko-ua: TBH, while the default configuration for the write component is conservative to avoid frequent scale-down events, it's difficult to establish optimal default settings for other components, as their ideal scaling behavior depends on the specific environment and workload.

oleksii-boiko-ua:
@trevorwhitney updated

production/helm/loki/values.yaml
trevorwhitney: yeah, fair point, agreed. thanks for the feedback!

trevorwhitney (approved):
Thanks so much for all you effort on this, looks great! Just a few build failures to get straightened out.

oleksii-boiko-ua:
@kd7lxl could you take a look?

docs/sources/operations/storage/wal.md
```diff
@@ -95,6 +95,18 @@ Statefulsets are significantly more cumbersome to work with/upgrade/etc. Much of
...
 
+#### Scaling Down Using `/flush_shutdown` Endpoint and Lifecycle Hook
+
+1. **StatefulSets for Ordered Scaling Down**: Loki's ingesters should be scaled down one by one, which is efficiently handled by Kubernetes StatefulSets. This ensures an ordered and reliable scaling process, as described in the [Deployment and Scaling Guarantees](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#deployment-and-scaling-guarantees) documentation.
+
+2. **Using PreStop Lifecycle Hook**: During the pod scaling down process, the PreStop [lifecycle hook](https://pwittrock.github.io/docs/concepts/containers/container-lifecycle-hooks/) triggers the `/flush_shutdown` endpoint on the ingester. This action flushes the chunks and removes the ingester from the ring, allowing it to register as unready and become eligible for deletion.
+
+3. **ScaleDown Policy**: The ScaleDown policy is configured to permit the removal of one pod every 30 minutes. This duration provides time for the ingester to flush its data before being deleted.
```
kd7lxl, line 104: This is a little misleading. The `terminationGracePeriodSeconds` is what provides time for the ingester to flush its data before being deleted.

https://github.com/grafana/loki/blob/425736614c43f3435e52bc1e5a2295dfcbb6418a/production/helm/loki/values.yaml#L679-L682

If you think flushing data may take 30 minutes, you might recommend increasing the `terminationGracePeriodSeconds`.

The scaledown policy is more important to avoid thrashing and for queries when configured to only query the ingesters. If all replicas of a stream are flushed from ingesters due to multiple pods scaledown within the `max_chunk_age`, the query will rely on the store to return the data. My understanding is that default Loki configuration handles this gracefully, so it would only be a concern with an odd config like `query_ingester_only: true`.

docs/sources/operations/storage/wal.md
```diff
@@ -95,6 +95,18 @@ Statefulsets are significantly more cumbersome to work with/upgrade/etc. Much of
 
 In this case, try `kubectl -n <namespace> delete sts ingester --cascade=false`. This will leave the pods alive but delete the statefulset. Then you may recreate the (updated) statefulset and one-by-one start deleting the `ingester-0` through `ingester-n` pods _in that order_, allowing the statefulset to spin up new pods to replace them.
 
+#### Scaling Down Using `/flush_shutdown` Endpoint and Lifecycle Hook
+
+1. **StatefulSets for Ordered Scaling Down**: Loki's ingesters should be scaled down one by one, which is efficiently handled by Kubernetes StatefulSets. This ensures an ordered and reliable scaling process, as described in the [Deployment and Scaling Guarantees](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#deployment-and-scaling-guarantees) documentation.
```
kd7lxl, line 100: You describe the behavior of `OrderedReady` here, but the statefulset is configured for `Parallel`.

https://github.com/grafana/loki/blob/425736614c43f3435e52bc1e5a2295dfcbb6418a/production/helm/loki/templates/write/statefulset-write.yaml#L15

production/helm/loki/templates/write/statefulset-write.yaml
```diff
@@ -101,9 +102,17 @@ spec:
             {{- toYaml .Values.loki.containerSecurityContext | nindent 12 }}
           readinessProbe:
             {{- toYaml .Values.loki.readinessProbe | nindent 12 }}
-          {{- with .Values.write.lifecycle }}
+          {{- if .Values.write.lifecycle }}
           lifecycle:
-            {{- toYaml . | nindent 12 }}
+            {{- toYaml .Values.write.lifecycle | nindent 12 }}
+          {{- else }}
+            {{- if .Values.write.autoscaling.enabled }}
```
kd7lxl, line 109: Not a big deal, but this could be a single `else if`.
```suggestion
          {{- else if .Values.write.autoscaling.enabled }}
```

kd7lxl (commented):
@oleksii-boiko-ua, per your request, I've provided a review. Note that I am not a maintainer of this project.

The way you've organized the values is simple to use and easy to override. I like it. Just made a few comments on wording in the docs.

docs/sources/operations/storage/wal.md
oleksii-boiko-ua: Updated, thanos good catch

docs/sources/operations/storage/wal.md
```diff
@@ -95,6 +95,18 @@ Statefulsets are significantly more cumbersome to work with/upgrade/etc. Much of
...
+
+2. **Using PreStop Lifecycle Hook**: During the pod scaling down process, the PreStop [lifecycle hook](https://pwittrock.github.io/docs/concepts/containers/container-lifecycle-hooks/) triggers the `/flush_shutdown` endpoint on the ingester. This action flushes the chunks and removes the ingester from the ring, allowing it to register as unready and become eligible for deletion.
```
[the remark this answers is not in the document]
oleksii-boiko-ua: fixed

oleksii-boiko-ua:
@kd7lxl updated. thanks

oleksii-boiko-ua:
@trevorwhitney @jeschkies could anyone merge it, or what is the process?

aglees:
Is there any news on when this might be released?

HerbBoy:
Is there a policy preventing this from being merged? Or when can we expect this fix to be implemented?

MichelHollands:
@oleksii-boiko-ua Can you add a HELM changelog entry and update the Helm chart version to 5.5.6? An example can be found [here](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213). 
Also please rebase and fix the 2 small CI issues.
Sorry for the delay. There are a lot of Helm PRs coming in.

oleksii-boiko-ua:
@MichelHollands updated. thanks

MichelHollands (approved):
LGTM

MichelHollands:
@oleksii-boiko-ua There is a parse error when the chart is installed. See the failing CI step. Can you have a look at that?

PR #7451, opened 2022-10-18
operator: Add support for rules configmap sharding

btaani:
**What this PR does / why we need it**:
This PR adds the functionality of splitting the rules ConfigMap into shards in case the size of alerting/recording rules becomes larger than 1MB ([ConfigMap docs](https://kubernetes.io/docs/concepts/configuration/configmap/#motivation))

**Which issue(s) this PR fixes**:
[LOG-3140](https://issues.redhat.com/browse/LOG-3140)

**Special notes for your reviewer**:
Cleaning up the ConfigMaps that are no longer needed caused by deleting rules is not part of this PR.

**Checklist**
- [x] Reviewed the `CONTRIBUTING.md` guide
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #8687, opened 2023-03-02
[Draft] Make loki aware of nested rules directories

btaani:
Loki currently stores the rules in the following directory on the ruler pod:

`/base/userID/rules.yaml`

However, on the operator side, if rules size exceed 1MB, they are split across multiple config maps, which should mount to different directories on the pod, for example, the rules will be split into different nested directories for each userID:

`/base/configMap-x/userID/rules.yaml`, `/base/configMap-y/userID/rules.yaml`

This PR makes Loki aware of any nesting caused by the operator, and reads the rules in their new path that contains the config map name.

Prerequisite to #7451 

This PR is currently a draft. one unit test (`TestClient_LoadAllRuleGroups`) still fails in case there's no config map name in the rule path. Since loki reads the user IDs from the rule path, when a new directory is introduced to the path (CM name in this case), it is read as a userID, which is not correct.

The only workaround I am thinking of is introducing a fixed prefix to the config map name so that it is ignored if read by loki.

cc @periklis @xperimental 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/ruler/rulestore/local/local.go
+++ b/pkg/ruler/rulestore/local/local.go
@@ -70,7 +70,7 @@ func (l *Client) ListAllUsers(ctx context.Context) ([]string, error) {
 		if d.Type()&fs.FileMode(os.ModeSymlink) != 0 {
 			fi, err := os.Stat(p)
 			if err != nil {
-				return err
+				return errors.Wrapf(err, "unable to stat file %s", p)
 			}
 			isDir = fi.IsDir()
 		} else {
@@ -86,7 +86,7 @@ func (l *Client) ListAllUsers(ctx context.Context) ([]string, error) {
 		return nil
 	})
 	if err != nil {
-		println(err)
+		return nil, errors.Wrapf(err, "failed to walk the file tree located at %s", l.cfg.Directory)
 	}
 
 	return result, nil
--- a/pkg/ruler/rulestore/local/local.go
+++ b/pkg/ruler/rulestore/local/local.go
@@ -3,7 +3,9 @@ package local
 import (
 	"context"
 	"flag"
+	"io/fs"
 	"os"
+	"path"
 	"path/filepath"
 
 	"github.com/pkg/errors"
@@ -28,6 +30,12 @@ func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 // Client expects to load already existing rules located at:
 //
 //	cfg.Directory / userID / namespace
+//
+// When the Loki Operator stores rules in
+//
+//	multiple Config Map shards, they are located at:
+//
+//	cfg.Directory / configMapName / userID / namespace
 type Client struct {
 	cfg    Config
 	loader promRules.GroupLoader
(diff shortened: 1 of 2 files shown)


pkg/ruler/rulestore/local/local.go
```diff
@@ -44,35 +52,41 @@ func NewLocalRulesClient(cfg Config, loader promRules.GroupLoader) (*Client, err
...
 
-		if entry.Type()&os.ModeSymlink != 0 {
-			// os.ReadDir only returns result of LStat. Calling Stat resolves symlink.
-			fi, err := os.Stat(filepath.Join(root, entry.Name()))
+	err := filepath.WalkDir(l.cfg.Directory, func(p string, d fs.DirEntry, err error) error {
+		if d.Type()&fs.FileMode(os.ModeSymlink) != 0 {
+			fi, err := os.Stat(p)
 			if err != nil {
-				return nil, err
+				return err
```
dannykopping, line 73: Please wrap this error

pkg/ruler/rulestore/local/local.go
```diff
@@ -44,35 +52,41 @@ func NewLocalRulesClient(cfg Config, loader promRules.GroupLoader) (*Client, err
...
+			userDir := path.Base(path.Dir(p))
+			if !contains(result, userDir) {
+				result = append(result, userDir)
+			}
 		}
+		return nil
+	})
+	if err != nil {
+		println(err)
 	}
```
dannykopping, line 90: Please either log this with the standard logger or return it

pkg/ruler/rulestore/local/local.go
```diff
@@ -133,49 +147,60 @@ func (l *Client) DeleteNamespace(ctx context.Context, userID, namespace string)
...
+				if err != nil {
+					return nil, errors.Wrapf(err, "unable to stat rule file %s", path)
+				}
+				isDir = fi.IsDir()
+			} else {
+				isDir = entry.IsDir()
+			}
+			if isDir {
+				continue
+			}
```
dannykopping, line 190: This code looks almost identical to `ListAllUsers`. Can we reuse it somehow?

pkg/ruler/rulestore/local/local.go
```diff
@@ -185,7 +210,7 @@ func (l *Client) loadAllRulesGroupsForUserAndNamespace(_ context.Context, userID
 	var list rulespb.RuleGroupList
 
 	for _, group := range rulegroups.Groups {
-		desc := rulespb.ToProto(userID, namespace, group)
+		desc := rulespb.ToProto(path.Base(rootDir), namespace, group)
```
dannykopping, line 213: `ToProto` expects the first argument to be the userID; why are you using the path here?

pkg/ruler/rulestore/local/local_test.go
```diff
@@ -71,10 +71,16 @@ func TestClient_LoadAllRuleGroups(t *testing.T) {
 
 	ctx := context.Background()
 	userMap, err := client.ListAllRuleGroups(ctx) // Client loads rules in its List method.
+
+	for u, _ := range userMap {
+		println("user = ", u)
```
dannykopping, line 76: Please use a logger or use `t.Logf`

pkg/ruler/rulestore/local/local_test.go
```diff
@@ -83,3 +89,53 @@ func TestClient_LoadAllRuleGroups(t *testing.T) {
 		require.Equal(t, rulespb.ToProto(u, namespace2, ruleGroups.Groups[0]), actual[1])
 	}
 }
+
+func TestListRules_withConfigMapSharding(t *testing.T) {
+	user1 := "user1"
```
dannykopping, line 94: I don't think there should be any mention of configmaps in the code or the tests. The limitation from k8s should have no bearing on Loki, and instead we should express the intent that all files are loaded recursively.

Given the structure `/base/<user-id>/<any-amount-of-nesting>/file.yaml`, load all the files for each user.

dannykopping (commented):
Thank you for the contribution @btaani.

I think we should omit any mention of `ConfigMap`s in Loki itself, as this is a k8s concern. I do like that we can now setup arbitrarily organised rules dirs and load all files recursively, however I think the directory structure should be

```
/base
  /<user-id>
    / ...any amount of nesting, including none
      /rules.yaml
```

not `/tmp/rules/configMapName/userID` as you describe in the description.

The userID key should always be the top-most level below the base.

btaani:
This PR is closed for now since it's no longer needed by the operator.

PR #8691, opened 2023-03-03
Implement hierarchical queues for query scheduler

chaudum:

This PR is the first step towards hierarchical queues within the query scheduler.
Hierarchical queues can be used to ensure query fairness across multiple actors within a single tenant, such as individual (human) users using the same tenant.

The change is described in the LID ["Query fairness across users within tenants"](https://grafana.com/docs/loki/next/lids/0003-queryfairnessinscheduler/).

* Extract metrics for `RequestQueue` into their own `metrics` field, so that they can extended and passed more easily in the future.
* Move mapping for tenant queues into its own data structure (ordered map) so its functionality can be reused. It also simplifies the code inside the `tenantQueues` struct.
* Add the `LeafQueue` struct that implements hierarchical queues. Sub-queues are dequeued in a round-robin manner using the same implementation of `Mapping` that is also used by the tenant queues.

### Special notes for your reviewer

Part of https://github.com/grafana/loki/pull/8585
Contains the changes from PR https://github.com/grafana/loki/pull/8722

**This PR does not yet make use of the new hierarchical queues, it only adds the relevant data structures. Wiring up will be done in a follow up PR.**

While the new `Mapping` data structure adds an unsignificant overhead to the time/op I consider the simplification an overall better tradeoff.

```console
$ benchstat scheduler-bench-old.txt scheduler-bench-new.txt 
name              old time/op    new time/op    delta
GetNextRequest-8    18.0µs ± 2%    20.5µs ± 3%  +13.89%  (p=0.008 n=5+5)
QueueRequest-8      56.6µs ± 0%    53.5µs ± 3%   -5.48%  (p=0.008 n=5+5)

name              old alloc/op   new alloc/op   delta
GetNextRequest-8    1.60kB ± 0%    1.60kB ± 0%     ~     (all equal)
QueueRequest-8      35.6kB ± 0%    29.8kB ± 0%  -16.37%  (p=0.008 n=5+5)

name              old allocs/op  new allocs/op  delta
GetNextRequest-8       100 ± 0%       100 ± 0%     ~     (all equal)
QueueRequest-8         819 ± 0%       806 ± 0%   -1.59%  (p=0.008 n=5+5)
```

#### Complexity of operations

| Operation | Function | Complexity |
| --------- | -------- | ---------- |
| Insert | `Put(key string, value *tenantQueue) bool` | `O(1)`|
| Select | `GetNext(idx QueueIndex) *tenantQueue` | `O(1)` (best case) `O(n)` (worst case) |
| Remove | `Remove(key string) bool` | `O(1)` |

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/scheduler/queue/mapping.go
+++ b/pkg/scheduler/queue/mapping.go
@@ -1,5 +1,12 @@
 package queue
 
+type Mapable interface {
+	*tenantQueue | *LeafQueue
+	// https://github.com/golang/go/issues/48522#issuecomment-924348755
+	Pos() QueueIndex
+	SetPos(index QueueIndex)
+}
+
 var empty = string([]byte{byte(0)})
 
 // Mapping is a map-like data structure that allows accessing its items not
@@ -9,47 +16,46 @@ var empty = string([]byte{byte(0)})
 // changing the index of the remaining items after the removed key.
 // Mapping uses *tenantQueue as concrete value and keys of type string.
 // The data structure is not thread-safe.
-// TODO(chaudum): Implement QueueMapping using generics
-type Mapping struct {
-	m     map[string]*tenantQueue
+type Mapping[v Mapable] struct {
+	m     map[string]v
 	keys  []string
-	empty []int
+	empty []QueueIndex
 }
 
-func (om *Mapping) Init(size int) {
-	om.m = make(map[string]*tenantQueue, size)
-	om.keys = make([]string, 0, size)
-	om.empty = make([]int, 0, size)
+func (m *Mapping[v]) Init(size int) {
+	m.m = make(map[string]v, size)
+	m.keys = make([]string, 0, size)
+	m.empty = make([]QueueIndex, 0, size)
 }
 
-func (om *Mapping) Put(key string, value *tenantQueue) bool {
+func (m *Mapping[v]) Put(key string, value v) bool {
 	// do not allow empty string or 0 byte string as key
 	if key == "" || key == empty {
 		return false
 	}
-	if len(om.empty) == 0 {
-		value.index = len(om.keys)
-		om.keys = append(om.keys, key)
+	if len(m.empty) == 0 {
+		value.SetPos(QueueIndex(len(m.keys)))
+		m.keys = append(m.keys, key)
 	} else {
-		idx := om.empty[0]
-		om.empty = om.empty[1:]
-		om.keys[idx] = key
-		value.index = idx
+		idx := m.empty[0]
+		m.empty = m.empty[1:]
+		m.keys[idx] = key
+		value.SetPos(idx)
 	}
-	om.m[key] = value
+	m.m[key] = value
 	return true
 }
 
-func (om *Mapping) Get(idx QueueIndex) *tenantQueue { //nolint:revive
-	if len(om.keys) == 0 {
+func (m *Mapping[v]) Get(idx QueueIndex) v {
+	if len(m.keys) == 0 {
 		return nil
 	}
-	k := om.keys[idx]
-	return om.GetByKey(k)
+	k := m.keys[idx]
+	return m.GetByKey(k)
 }
 
-func (om *Mapping) GetNext(idx QueueIndex) *tenantQueue { //nolint:revive
-	if len(om.keys) == 0 {
+func (m *Mapping[v]) GetNext(idx QueueIndex) v {
+	if len(m.keys) == 0 {
 		return nil
 	}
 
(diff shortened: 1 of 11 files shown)


pkg/scheduler/queue/tenant_queues.go
```diff
@@ -154,31 +134,27 @@ func (q *tenantQueues) getNextQueueForQuerier(lastUserIndex QueueIndex, querierI
...
-		// for example when q.users has shrunk since last call.
-		if int(uid) >= len(q.tenants) {
+		if int(uid) >= q.mapping.Len() {
 			uid = 0
 		}
 
-		u := q.tenants[uid]
-		if u == "" {
-			continue
+		tq := q.mapping.Get(uid)
```
jeschkies, line 144: Just so I understand. The iteration order needs to be kept so that the scheduling is deterministic, ie the next queue is is fixed by the order, right?

chaudum: Right

pkg/scheduler/queue/mapping.go
```diff
@@ -0,0 +1,160 @@
+package queue
+
+// OrderedMap is a map-like data structure that allows accessing its items not
+// only by key but also by index.
+// It preserves the ingest order of times for iterating over the values.
+// Keys are only of type string, values are of any type.
+// The data structure is not thread-safe.
+// TODO(chaudum): Implement QueueMapping using generics
+type OrderedMap struct {
```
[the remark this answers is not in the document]
chaudum: I removed the `OrderedMap` completely, because it does not have the same semantics as the current implementation.

pkg/scheduler/queue/mapping.go
```diff
@@ -0,0 +1,160 @@
...
+func (om *OrderedMap) Put(key string, value any) int {
+	if key == "" {
+		return len(om.keys)
+	}
+	om.keys = append(om.keys, key)
+	om.m[key] = value
+	return len(om.keys)
+}
+
+func (om *OrderedMap) Get(idx int) any {
```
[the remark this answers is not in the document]
chaudum: Replaced by `GetNext()` I could rename it to `GetNextByIndex()`

pkg/scheduler/queue/tenant_queues_test.go
```diff
@@ -193,7 +193,7 @@ func TestQueuesConsistency(t *testing.T) {
 
 			conns := map[string]int{}
 
-			for i := 0; i < 10000; i++ {
+			for i := 0; i < 10; i++ {
```
chaudum, line 196: That is a mistake

pkg/scheduler/queue/queue.go
```diff
@@ -140,12 +142,12 @@ FindQueue:
 
 		// Pick next request from the queue.
 		for {
-			request := <-queue
-			if len(queue) == 0 {
+			request := <-queue.Chan()
+			if len(queue.Chan()) == 0 {
```
owen-d, line 146: I think this needs to use the new `queue.Len` since the new queue impl is nested and `len(ch)` won't test the recursive sub queues.

edit: you've already addressed in https://github.com/grafana/loki/pull/8722

owen-d (approved):
There's a lot of overlap between this and https://github.com/grafana/loki/pull/8722 which I've also reviewed. My only request is already fixed there, so LGTM

pkg/scheduler/queue/mapping.go
```diff
@@ -0,0 +1,111 @@
...
+var empty = string([]byte{byte(0)})
+
+// Mapping is a map-like data structure that allows accessing its items not
+// only by key but also by index.
+// When an item is removed, the iinternal key array is not resized, but the
+// removed place is marked as empty. This allows to remove keys without
+// changing the index of the remaining items after the removed key.
+// Mapping uses *tenantQueue as concrete value and keys of type string.
+// The data structure is not thread-safe.
+// TODO(chaudum): Implement QueueMapping using generics
```
kavirajk, line 12: nit: Better to add `O(n)` notation for inserts, scans and deletes for this data structure?. My feeling is that would communicate the efficiency of this datastructure bit better without much words?

kavirajk (approved):
LGTM :+1:

pkg/scheduler/queue/mapping.go
chaudum: | Operation | Function | Complexity |
| --------- | -------- | ---------- |
| Insert | `Put(key string, value *tenantQueue) bool` | `O(1)`|
| Select | `GetNext(idx QueueIndex) *tenantQueue` | `O(1)` (best case) `O(n)` (worst case) |
| Remove | `Remove(key string) bool` | `O(1)` |

PR #8693, opened 2023-03-03
fix bug: one line duplicate log after retry

### LINKED ISSUE #8692: one line duplicate log after retry
**Describe the bug**
When use loki-cli tail to query the real-time log, if server close the websocket connection the client will try to reconnect to the websocket server. And the parameter 'start' will use the last received timestamp. However, if we use the last received timestamp, the last log will duplicate after retry successfully.

infinityjay:
When use loki-cli tail to query the real-time log, if server close the websocket connection the client will try to reconnect to the websocket server. And the parameter 'start' will use the last received timestamp. However, if we use the last received timestamp, the last log will duplicate after retry successfully. So we should add 1 nanosecond to the last received timestamp as next time query parameter.

Fixes #8692

**Checklist**
- [ ] Reviewed the /pkg/logcli/query/tail.go(https://github.com/grafana/loki/blob/main/pkg/logcli/query/tail.go)

### DIFF
--- a/pkg/logcli/query/tail.go
+++ b/pkg/logcli/query/tail.go
@@ -115,7 +115,7 @@ func (q *Query) TailQuery(delayFor time.Duration, c client.Client, out output.Lo
 
 			for _, entry := range stream.Entries {
 				out.FormatAndPrintln(entry.Timestamp, labels, 0, entry.Line)
-				lastReceivedTimestamp = entry.Timestamp
+				lastReceivedTimestamp = entry.Timestamp.Add(1 * time.Nanosecond)
 			}
 
 		}


infinityjay:
Thanks for your question!
Let me explain the problem more concretely by taking an example.
First of all, we have to assume that two logs will not be created at the same timestamp(ns). In fact, we can use the unique timestamp(ns) to distinguish one log from another one in loki. 

If the logs stored in loki as follows:
`1677859200000000000    Log line 1`
`1677859260000000000    Log line 2`
`1677859260000000005    Log line 3`
`1677859380000000000    Log line 4`
We use logcli to query real-time logs by websocket connection. And the connection is interrupted on 1677859280000000000, which is between `Log line 2` and `Log line 3`. And the connection is rebuilt after 3 retries.

Before fix, the logcli will show as follows:
`1677859200000000000    Log line 1`
`1677859260000000000    Log line 2`
`1677859260000000000    Log line 2`
`1677859260000000005    Log line 3`
`1677859380000000000    Log line 4`
It is obvious that the logcli directly uses the last received timestamp `1677859260000000000` as the start parameter to ask for the next connection during retries. And the last received log will show up twice after retry successfully. 

So, we should add 1 ns to the last received timestamp which is `1677859260000000001` . And the next query request will query logs from `1677859260000000001` (including the log with timestamp `1677859260000000001`), and the logcli will show as follows:
`1677859200000000000    Log line 1`
`1677859260000000000    Log line 2`
`1677859260000000005    Log line 3`
`1677859380000000000    Log line 4`
And the user will also fail to notice the websocket connection was once interrupted and rebuilt.


> Hey thanks for your contribution ❤️
> 
> I didn't get it: adding the 1ns will make the retried line different than the other but logcli would still be duplicating it. Could you elaborate on why this isn't a duplication?

DylanGuedes:
ahá, got it now, thanks. Btw you probably meant to say the interruption occurred at `1677859260000000001` or something since `1677859280000000000` would be between lines 3 and 4 instead.

I don't think it is fair to ask you to implement unit tests for the change since looks like tests for the tailing are non-existent at this moment, but could you please test it (by running locally) what happens if the interruption occurs 1ns before the next line that wasn't received and one more with the same timestamp of the last log line received? 

Taking your example as an initial point, something like:
Test 1:
1677859200000000000    Log line 1
1677859260000000001    Log line 2
1677859260000000001    <interruption occurs>
1677859260000000002    Log line 3
*should print all 3 lines just fine once the connection gets fixed*

and Test 2:
1677859200000000000    Log line 1
1677859260000000001    Log line 2
1677859260000000002    Log line 3
1677859260000000002    <interruption occurs>
*shouldn't print anything after the connection gets fixed*

If testing this is impossible because you can't control the exact moment of the interrupt let me know.

infinityjay:
> ahá, got it now, thanks. Btw you probably meant to say the interruption occurred at `1677859260000000001` or something since `1677859280000000000` would be between lines 3 and 4 instead.
> 
> I don't think it is fair to ask you to implement unit tests for the change since looks like tests for the tailing are non-existent at this moment, but could you please test it (by running locally) what happens if the interruption occurs 1ns before the next line that wasn't received and one more with the same timestamp of the last log line received?
> 
> Taking your example as an initial point, something like: Test 1: 1677859200000000000 Log line 1 1677859260000000001 Log line 2 1677859260000000001 1677859260000000002 Log line 3 _should print all 3 lines just fine once the connection gets fixed_
> 
> and Test 2: 1677859200000000000 Log line 1 1677859260000000001 Log line 2 1677859260000000002 Log line 3 1677859260000000002 _shouldn't print anything after the connection gets fixed_
> 
> If testing this is impossible because you can't control the exact moment of the interrupt let me know.

Thanks for your time! 
After check the code, I'v found the last timestamp has been added one nanosecond when handling the 'start' parameter, which used the same method as mine. And the next query will not duplicate one log.
image

Issue #8054, opened 2023-01-06
Promtail: possible memory leak


MiGrandjean:
**Describe the bug**
We are seeing a slow but steady increase in memory usage for our Promtail pods. IMHO this looks very typical for a memory leak. Sooner or later we experience OOM kills for Promtail. 

**To Reproduce**
1. Running Loki (2.6.1), deployed on EKS via Helm Chart
2. Running Promtail (2.7.0), deployed on EKS via Helm Chart
3. Running some Pods (Ingress, Prometheus, Grafana, Thanos ... ) on the Node

**Expected behavior**
I would expect memory consumption to be somewhat linear under regular operation. Or, if there are spikes, that the memory is freed after increased demand.

**Environment:**
 - Infrastructure: Kubernetes 1.22 (EKS)
 - Deployment tool: Helm ([official Promtail chart](https://github.com/grafana/helm-charts/tree/main/charts/promtail))

**Screenshots, Promtail config, or terminal output**

![Auswahl_274](https://user-images.githubusercontent.com/8418423/210974404-6da7f499-b6b1-4951-bf72-0717acaab438.png)

We are using the default values of the Helm Chart (with exception of the Loki URL and some podAnnotations and tolerations).

I'm also happy to investigate more on this and e.g. track down what is actually driving the memory consumption, if someone can point me in the right direction how to do this.

DylanGuedes:
Could you share a memory dump? You can do it by:
1. Run promtail with `-server.profiling_enabled`
2. Access `http://localhost:9080/debug/pprof/heap?duration=15s`

grandich:
Also seeing a "leaky" pattern and oom kills across our promtail daemonsets

![image](https://github.com/grafana/loki/assets/1310215/b60a0869-3640-458d-af37-822b1ea5bc8c)

sum by (node) (container_memory_working_set_bytes{job="cadvisor/cadvisor", pod=~"promtail.+",container!=""})

grandich:
We could correlate the leak with extremely long lines or absolutely no line breaks in certain pods logs. We experimented adding line breaks and the leak disappeared. 

Using max_line_size = 0

Gakhramanzode:
> We could correlate the leak with extremely long lines or absolutely no line breaks in certain pods logs. We experimented adding line breaks and the leak disappeared.
> 
> Using max_line_size = 0

hi @grandich . you mean adding this lines ?
```
config:
  snippets:
    extraLimitsConfig: |
      max_line_size: 0
[...]

PR #8695, opened 2023-03-03
Promtail: Stop ticker to avoid leaking

DylanGuedes:
From [NewTicker docs](https://pkg.go.dev/time#NewTicker):
`Stop the ticker to release associated resources.`
So not calling `Stop` for Promtail positions is probably leading to leakage at some level. Not sure what is the impact of this, though.

### DIFF
--- a/clients/pkg/promtail/positions/positions.go
+++ b/clients/pkg/promtail/positions/positions.go
@@ -152,6 +152,7 @@ func (p *positions) run() {
 	}()
 
 	ticker := time.NewTicker(p.cfg.SyncPeriod)
+	defer ticker.Stop()
 	for {
 		select {
 		case <-p.quit:
--- a/clients/pkg/promtail/positions/positions_test.go
+++ b/clients/pkg/promtail/positions/positions_test.go
@@ -2,7 +2,6 @@ package positions
 
 import (
 	"os"
-	"runtime"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


dannykopping (commented):
I think this only applies to the `Tick()` function, no?
https://pkg.go.dev/time#Tick

In any case, seems worthwhile but I don't think we should approve this without a test.

DylanGuedes:
> I think this only applies to the `Tick()` function, no? https://pkg.go.dev/time#Tick
> 
> In any case, seems worthwhile but I don't think we should approve this without a test.

Oops, you're right. But still, the `NewTicker` [docs kinda confirm the leaking potential](https://pkg.go.dev/time#NewTicker) of not calling `Stop`:
```
Stop the ticker to release associated resources.
```

DylanGuedes:
> I think this only applies to the `Tick()` function, no? https://pkg.go.dev/time#Tick
> 
> In any case, seems worthwhile but I don't think we should approve this without a test.

I pushed a test that replicates the leak if you comment out the `.Stop()` call the test won't pass. WDYT?

dannykopping (approved):
Thanks for adding this!

I'd love to understand better _what_ is leaking and what impact it has. I'm thinking how we can test this in a way that expresses the intent a little more clearly. Maybe just leave off the `Stop` and test that the current and previous allocations are short by 1, indicating a leak... What you've got here is interesting but I think it's testing a Go runtime internal, which may end up being flimsy.

Either way it seems sensible to add this so let me not block it. Another question: is this the only ticker that is not stopped?

DylanGuedes:
> Thanks for adding this!
> 
> I'd love to understand better _what_ is leaking and what impact it has.

- Promtail will have a `Position` instance for every target. During a `Position` initialization, it allocates/starts a new ticker using `time.NewTicker`. The leaking comes from when a target has its scrapping stopped. The target in question would invoke `Position.Stop()` with the intention of freeing `Position` resources, but its `Position` won't release its allocated ticker, which will keep running in the background.
- The impact is probably low and depends on how many targets you stop and start during Promtail execution. I found this while investigating [this report](https://github.com/grafana/loki/issues/8054). But regardless of this having a low impact, I decided to open the PR anyway because it is a bug (in the sense you never want to start a ticker with `time.NewTicker` and never stops `Stop` for that ticker).

> I'm thinking how we can test this in a way that expresses the intent a little more clearly. Maybe just leave off the `Stop` and test that the current and previous allocations are short by 1, indicating a leak... What you've got here is interesting but I think it's testing a Go runtime internal, which may end up being flimsy.

Tbh I didn't like that test. I pushed it more to illustrate, so I decided to remove it anyway. Do you have a suggestion on how I could test this better?

> is this the only ticker that is not stopped?

Before opening the PR I searched inside `promtail` folder and yeah, that was the only place with a ticker not being stopped. After your comment I decided to check across Loki's whole code base and found other 2 places that have this problem. I'll open a separate PR for those.

PR #8704, opened 2023-03-06
Promtail: Fix the default value for a Mechanism field in SASL auth config

andriikushch:

This PR allows using a default value for a Mechanism field in the SASL auth config, as specified in the documentation.

Without this change, the error `error unsupported sasl mechanism:` is returned if the Mechanism field is not provided in a scrape config and Promtail doesn't start.

Not available

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/clients/pkg/promtail/targets/kafka/target_syncer.go
+++ b/clients/pkg/promtail/targets/kafka/target_syncer.go
@@ -159,8 +159,8 @@ func withSASLAuthentication(cfg sarama.Config, authCfg scrapeconfig.KafkaAuthent
 		sarama.SASLTypeSCRAMSHA256,
 		sarama.SASLTypePlaintext,
 	}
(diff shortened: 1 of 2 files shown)


cstyan:
@andriikushch Hello, thanks for your PR.

We're currently reevaluating promtails position as a project within Grafana Labs. Internally we're actually using the [Agent](https://github.com/grafana/agent) for both metrics and logs collection at this point. Additionally, the agent team is more likely to have time to dedicate to PR. 

We could probably consider this PR a bugfix and move forward anyways, but I'm quite understanding how checking the value of `cfg.Net.SASL.Mechanism` instead of `authCfg.SASLConfig.Mechanism` is the correct fix if we're wanting to ensure there's a default value. I think instead there's likely an issue with the parsing or the default values for the `authCfg.SASLConfig.Mechanism` in the struct definition.

PR #8724, opened 2023-03-07
Short circuit parsers

MasslessParticle:
In many cases we only need a few labels from a line. Because most of our parsers parse lines incrementally, we can stop parsing a line after we have all the labels we want from it.

This pr uses `ParserHints` to keep track of the number of extracted labels. It also provides a way for parsers to know when they should stop parsing.  

Notes:
- parsers had inconsistent ordering between the `ShouldExtract` call and adding the `_extracted` label to duplicates. This PR makes appending `_extracted` always happen before `ShouldExtract` to keep counts of what is extracted compared to expected labels consistent.

Next Steps:
- When the user specifies a query with a grouping containing the `_extracted` label but there is no duplicate between the passed line and labels, short circuiting will not work. I'll address this in a follow up PR.

Benchmarks:
To try and show a balanced view of what this buys us, this pr picks a label our of the middle of a line. In a best case this might be much better. In the worst case, we have to parse the whole line.

```
benchstat short_circuit_old.txt short_circuit_new.txt
name                               old time/op    new time/op    delta
KeyExtraction/json-8                  456ns ± 3%     256ns ± 2%  -43.84%  (p=0.000 n=9+10)
KeyExtraction/logfmt-8                347ns ± 4%     171ns ± 2%  -50.86%  (p=0.000 n=10+10)
KeyExtraction/logfmt-expression-8     552ns ± 2%     368ns ± 2%  -33.23%  (p=0.000 n=9+10)

name                               old alloc/op   new alloc/op   delta
KeyExtraction/json-8                  5.00B ± 0%     5.00B ± 0%     ~     (all equal)
KeyExtraction/logfmt-8                5.00B ± 0%     5.00B ± 0%     ~     (all equal)
KeyExtraction/logfmt-expression-8     16.0B ± 0%     16.0B ± 0%     ~     (all equal)

name                               old allocs/op  new allocs/op  delta
KeyExtraction/json-8                   1.00 ± 0%      1.00 ± 0%     ~     (all equal)
KeyExtraction/logfmt-8                 1.00 ± 0%      1.00 ± 0%     ~     (all equal)
KeyExtraction/logfmt-expression-8      2.00 ± 0%      2.00 ± 0%     ~     (all equal)
```

### DIFF
--- a/docs/sources/upgrading/_index.md
+++ b/docs/sources/upgrading/_index.md
@@ -35,6 +35,11 @@ The output is incredibly verbose as it shows the entire internal config struct u
 
 ### Loki
 
+#### Change in LogQL behavior
+
+When there are duplicate labels in a log line, only the first value will be kept. Previously only the last value
+was kept
+
 #### Default retention_period has changed
 
 This change will affect you if you have:
--- a/docs/sources/upgrading/_index.md
+++ b/docs/sources/upgrading/_index.md
@@ -38,7 +38,7 @@ The output is incredibly verbose as it shows the entire internal config struct u
 #### Change in LogQL behavior
 
 When there are duplicate labels in a log line, only the first value will be kept. Previously only the last value
-was kept
+was kept.
 
 #### Default retention_period has changed
 
--- a/pkg/logql/log/parser.go
+++ b/pkg/logql/log/parser.go
@@ -66,6 +66,7 @@ func (j *JSONParser) Process(_ int64, line []byte, lbs *LabelsBuilder) ([]byte, 
 
 	if err := jsonparser.ObjectEach(line, j.parseObject); err != nil {
 		if errors.Is(err, errFoundAllLabels) {
+			// Short-circuited
 			return line, true
 		}
 
@@ -95,6 +96,8 @@ func (j *JSONParser) parseObject(key, value []byte, dataType jsonparser.ValueTyp
 	}
 
 	if j.lbs.ParserLabelHints().AllRequiredExtracted() {
+		// Not actually an error. Parsing can be short-circuited
+		// and this tells jsonparser to stop parsing
 		return errFoundAllLabels
 	}
 
--- a/pkg/logql/log/parser.go
+++ b/pkg/logql/log/parser.go
@@ -362,10 +362,6 @@ func (l *PatternParser) Process(_ int64, line []byte, lbs *LabelsBuilder) ([]byt
 		}
 
 		lbs.Set(name, string(m))
-
-		if lbs.ParserLabelHints().AllRequiredExtracted() {
-			return line, true
-		}
 	}
 	return line, true
 }
--- a/pkg/logql/log/parser.go
+++ b/pkg/logql/log/parser.go
@@ -244,6 +244,10 @@ func (r *RegexpParser) Process(_ int64, line []byte, lbs *LabelsBuilder) ([]byte
 				if lbs.BaseHas(sanitize) {
 					sanitize = fmt.Sprintf("%s%s", sanitize, duplicateSuffix)
 				}
+				if !lbs.ParserLabelHints().ShouldExtract(sanitize) {
+					return "", false
+				}
+
 				return sanitize, true
 			})
 			if !ok {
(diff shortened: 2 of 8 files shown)


pkg/logql/log/parser.go
```diff
@@ -64,6 +65,10 @@ func (j *JSONParser) Process(_ int64, line []byte, lbs *LabelsBuilder) ([]byte,
 	j.lbs = lbs
 
 	if err := jsonparser.ObjectEach(line, j.parseObject); err != nil {
+		if errors.Is(err, errFoundAllLabels) {
+			return line, true
+		}
```
jeschkies, line 70: Ah, this is the old trick to use "exceptions". I think this needs a comment that it is not an actually error :slightly_smiling_face:

pkg/logql/log/parser_hints.go
```diff
@@ -57,30 +64,45 @@ func (p *parserHint) ShouldExtractPrefix(prefix string) bool {
...
+
+func (p *Hints) RecordExtracted() {
+	p.extracted++
+}
+
+func (p *Hints) AllRequiredExtracted() bool {
+	if len(p.requiredLabels) == 0 {
+		return false
+	}
+	return p.extracted == len(p.requiredLabels)
```
jeschkies, line 79: What if a label is twice in a JSON document?

jeschkies (changes requested):
Nice idea! I think we need a test for duplicated labels though.

pkg/logql/log/parser_hints.go
MasslessParticle: This is actually really interesting. I thought duplicate json keys were invalid json and it turns out that _might_ not be the case. It's ambiguous enough that this is a good thing to cover.

https://stackoverflow.com/a/23195243

MasslessParticle: lol, @jeschkies , this actually invalidates this whole approach. If duplicates can exist in the log -- as they can for `json` and `logfmt` -- there's no way to short-circuit because Loki currently handles duplicates by taking the `last seen` value.

MasslessParticle:
Closed due to the issue @jeschkies mentioned.

pkg/logql/log/parser_hints.go
jeschkies: 🙈 I'm sorry

MasslessParticle:
@jeschkies I chatted with @slim-bean and @owen-d  and decided that this is worth the breaking change for something that ought to be an edge case.

docs/sources/upgrading/_index.md
```diff
@@ -35,6 +35,11 @@ The output is incredibly verbose as it shows the entire internal config struct u
 
 ### Loki
 
+#### Change in LogQL behavior
+
+When there are duplicate labels in a log line, only the first value will be kept. Previously only the last value
+was kept
```
JStickler, line 41:
```suggestion
was kept.
```

JStickler (approved):
[docs team] LGTM, just one missing period.

pkg/logql/log/parser_hints.go
```diff
@@ -25,26 +25,41 @@ type ParserHint interface {
...
+	extracted      []string
 }
 
-func (p *parserHint) ShouldExtract(key string) bool {
+func (p *Hints) ShouldExtract(key string) bool {
 	if len(p.requiredLabels) == 0 {
 		return true
 	}
+
+	for _, l := range p.extracted {
```
jeschkies, line 47: You mean the already extracted labels, right?

pkg/logql/log/parser_hints.go
```diff
@@ -57,30 +72,52 @@ func (p *parserHint) ShouldExtractPrefix(prefix string) bool {
...
+		if l == key {
+			p.extracted = append(p.extracted, key)
+			return
+		}
+	}
+}
+
+func (p *Hints) AllRequiredExtracted() bool {
+	if len(p.requiredLabels) == 0 {
+		return false
```
jeschkies, line 90: Isn't this true? If not labels need to be extracted all are already extracted right?

pkg/logql/log/parser_test.go
```diff
@@ -111,6 +111,73 @@ func Test_jsonParser_Parse(t *testing.T) {
...
+      "name": "text1",
+      "name": "duplicate",
+      "hOffset": 250,
+      "vOffset": 100,
+      "alignment": "center",
+      "onMouseUp": "sun1.opacity = (sun1.opacity / 100) * 90;"
+    }`)
+	logFmt := []byte(`data="ClickHere" size=36 style=bold name=text1 name=duplicate hOffset=250 vOffset=100 alignment=center onMouseUp="sun1.opacity = (sun1.opacity / 100) * 90;"`)
+
+	hints := &fakeParseHints{label: "name"}
```
jeschkies, line 128: Why not used the actually ParserHints struct here?

jeschkies (commented):
Thanks. This is super clean. I just have some clarifying questions :slightly_smiling_face:

pkg/logql/log/parser_test.go
MasslessParticle: I try to use mocks in unit tests where possible so the test isn't couple to any of the behavior of the actual parser hints. As long as the interface behaves as it should, this test doesn't ever need to change because the way we implement ParserHints changes. It also means that we don't have to set this test up with anything specific to parser hints.

pkg/logql/log/parser_hints.go
MasslessParticle: yes

PR #8727, opened 2023-03-07
Add initial implementation of per-query limits

### LINKED ISSUE #8762: Support per-request limits and propage them to querier.
So far Loki supports global and tenant based querier limits. However, sometimes individual requests could harm the query performance for one tenant if a query is ill-formed. That is why we should support per-request query limits that are only enforce for an individual request.

Per-request limits are almost the same as the query limits already in place. They are encoded in JSON and sent vi the `X-Loki-Query-Limits` header. The head is extracted and passed via the context of the request to the querier where it is enforce.

cstyan:
Sometimes we want to limit the impact of a single query by imposing limits that are stricter than the current tenant limit. E.g. the maximum query length could be seven days but based on the query or an admins decision a query should just have a maximum length of one day. This is where per-request limits come into play. They are passed via the `X-Loki-Query-Limit` header and extracted into the requests context.

It is the responsibility of the operator or admin that the header is valid.

Fixes #8762

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -648,7 +648,7 @@ func (t *Loki) setupModuleManager() error {
 		Distributor:              {Ring, Server, Overrides, TenantConfigs, UsageReport},
 		Store:                    {Overrides, IndexGatewayRing},
 		Ingester:                 {Store, Server, MemberlistKV, TenantConfigs, UsageReport},
-		Querier:                  {Store, Ring, Server, IngesterQuerier, TenantConfigs, UsageReport, CacheGenerationLoader},
+		Querier:                  {Store, Ring, Server, IngesterQuerier, Overrides, UsageReport, CacheGenerationLoader},
 		QueryFrontendTripperware: {Server, Overrides, TenantConfigs},
 		QueryFrontend:            {QueryFrontendTripperware, UsageReport, CacheGenerationLoader},
 		QueryScheduler:           {Server, Overrides, MemberlistKV, UsageReport},
@@ -676,7 +676,7 @@ func (t *Loki) setupModuleManager() error {
 
 		deps[QueryLimiter] = []string{Overrides}
 		deps[QueryLimitsInterceptors] = []string{}
-		deps[QueryLimitsTripperware] = []string{}
+		deps[QueryLimitsTripperware] = []string{QueryFrontendTripperware}
 
 		if err := mm.AddDependency(Server, QueryLimitsInterceptors); err != nil {
 			return err
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -677,6 +677,10 @@ func (t *Loki) setupModuleManager() error {
 		deps[QueryLimiter] = []string{Overrides}
 		deps[QueryLimitsInterceptors] = []string{}
 		deps[QueryLimitsTripperware] = []string{}
+
+		if err := mm.AddDependency(Server, QueryLimitsInterceptors); err != nil {
+			return err
+		}
 	}
 
 	// Add IngesterQuerier as a dependency for store when target is either querier, ruler, read, or backend.
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -5,11 +5,12 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"github.com/grafana/loki/pkg/util/querylimits"
 	"net/http"
 	"os"
 	rt "runtime"
 	"time"
+
+	"github.com/grafana/loki/pkg/util/querylimits"
 
 	"go.uber.org/atomic"
 
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/grafana/loki/pkg/util/querylimits"
 	"net/http"
 	"os"
 	rt "runtime"
@@ -14,6 +15,7 @@ import (
 
 	"github.com/fatih/color"
 	"github.com/felixge/fgprof"
+	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/grafana/dskit/flagext"
 	"github.com/grafana/dskit/grpcutil"
@@ -57,6 +59,7 @@ import (
 	"github.com/grafana/loki/pkg/usagestats"
 	"github.com/grafana/loki/pkg/util"
 	"github.com/grafana/loki/pkg/util/fakeauth"
+	"github.com/grafana/loki/pkg/util/limiter"
 	util_log "github.com/grafana/loki/pkg/util/log"
 	serverutil "github.com/grafana/loki/pkg/util/server"
 	"github.com/grafana/loki/pkg/validation"
@@ -329,17 +332,6 @@ type Frontend interface {
 	CheckReady(_ context.Context) error
 }
 
-type CombinedLimits interface {
-	compactor.Limits
-	distributor.Limits
-	ingester.Limits
-	querier.Limits
-	queryrange.Limits
-	ruler.RulesLimits
-	scheduler.Limits
-	storage.StoreLimits
-}
-
 // Loki is the root datastructure for Loki.
 type Loki struct {
 	Cfg Config
@@ -353,7 +345,7 @@ type Loki struct {
 	Server                   *server.Server
 	InternalServer           *server.Server
 	ring                     *ring.Ring
-	Overrides                CombinedLimits
+	Overrides                limiter.CombinedLimits
 	tenantConfigs            *runtime.TenantConfigs
 	TenantLimits             validation.TenantLimits
 	distributor              *distributor.Distributor
@@ -415,6 +407,9 @@ func (t *Loki) setupAuthMiddleware() {
 			"/schedulerpb.SchedulerForQuerier/QuerierLoop",
 			"/schedulerpb.SchedulerForQuerier/NotifyQuerierShutdown",
 		})
+	middleware.Merge(t.HTTPAuthMiddleware, querylimits.NewQueryLimitsMiddleware(
+		log.With(util_log.Logger, "component", "query_limits_middleware"),
+	))
 }
 
 func (t *Loki) setupGRPCRecoveryMiddleware() {
@@ -634,6 +629,9 @@ func (t *Loki) setupModuleManager() error {
 	mm.RegisterModule(QueryFrontendTripperware, t.initQueryFrontendTripperware, modules.UserInvisibleModule)
 	mm.RegisterModule(QueryFrontend, t.initQueryFrontend)
 	mm.RegisterModule(RulerStorage, t.initRulerStorage, modules.UserInvisibleModule)
+	mm.RegisterModule(QueryLimiter, t.initQueryLimiter, modules.UserInvisibleModule)
+	mm.RegisterModule(QueryLimitsInterceptors, t.initQueryLimitsInterceptors, modules.UserInvisibleModule)
+	mm.RegisterModule(QueryLimitsTripperware, t.initQueryLimitsTripperware, modules.UserInvisibleModule)
 	mm.RegisterModule(Ruler, t.initRuler)
 	mm.RegisterModule(TableManager, t.initTableManager)
 	mm.RegisterModule(Compactor, t.initCompactor)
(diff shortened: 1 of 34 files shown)


pkg/util/querylimits/limiter.go
```diff
@@ -0,0 +1,100 @@
...
+		return 0, err
+	}
+	requestLimits := ExtractQueryLimitsContext(ctx)
+	if requestLimits == nil || requestLimits.QueryTimeout == 0 {
+		_ = level.Debug(logutil.WithContext(ctx, logutil.Logger)).Log("msg", "using original limit")
+		return original, nil
+	}
+	if time.Duration(requestLimits.QueryTimeout) > original {
+		_ = level.Error(logutil.WithContext(ctx, logutil.Logger)).Log("msg", "invalid QueryTimeout", "tenant", userID, "requestLimit", requestLimits.QueryTimeout, "tenantLimit", original)
+		return 0, fmt.Errorf("invalid QueryTimeout of %v, tenant limit is %v", requestLimits.QueryTimeout, original)
```
jeschkies, line 57: I don't think this should be here. `QueryTimeout` is called during query evaluation. End-users executing queries should not see this error. We should apply the more restricting limit instead like we've done in the prototype.

I think this should be in another API endpoint that is called by the admin that sets the policies.

jeschkies (changes requested):
Thanks for starting the work. I think there was some misunderstanding around the error messaging.

pkg/loki/loki.go
```diff
@@ -634,6 +629,9 @@ func (t *Loki) setupModuleManager() error {
 	mm.RegisterModule(QueryFrontendTripperware, t.initQueryFrontendTripperware, modules.UserInvisibleModule)
 	mm.RegisterModule(QueryFrontend, t.initQueryFrontend)
 	mm.RegisterModule(RulerStorage, t.initRulerStorage, modules.UserInvisibleModule)
+	mm.RegisterModule(QueryLimiter, t.initQueryLimiter, modules.UserInvisibleModule)
```
jeschkies, line 632: I think we should make the limiter configurable. What do you think?

pkg/querier/querier.go
```diff
@@ -507,6 +515,9 @@ func (q *SingleTenantQuerier) Series(ctx context.Context, req *logproto.SeriesRe
 
 	// Enforce the query timeout while querying backends
 	queryTimeout := q.limits.QueryTimeout(ctx, userID)
+	if err != nil {
+		return nil, err
+	}
```
jeschkies, line 520: This is not needed.

pkg/querier/queryrange/roundtrip.go
```diff
@@ -238,15 +238,30 @@ func transformRegexQuery(req *http.Request, expr syntax.LogSelectorExpr) (syntax
 	return expr, nil
 }
 
+func SmallestPositiveNonZeroIntPerTenant(tenantIDs []string, f func(string) int) int {
```
jeschkies, line 241: We should have sich a method already. In any case why not move it to the other methods in the validation package?

pkg/util/querylimits/limiter.go
```diff
@@ -0,0 +1,57 @@
...
+		return original
+	}
+	_ = level.Debug(logutil.WithContext(ctx, logutil.Logger)).Log("msg", "using request limit", "limit", requestLimits.MaxQueryLength)
+	return time.Duration(requestLimits.MaxQueryLength)
+}
+
+// MaxQueryLookback returns the max lookback period of queries.
+func (l *Limiter) MaxQueryLookback(ctx context.Context, userID string) time.Duration {
+	original := l.CombinedLimits.MaxQueryLookback(ctx, userID)
+	// in theory this error should never happen
```
jeschkies, line 38: This can go 🙂

pkg/util/querylimits/limiter.go
```diff
@@ -0,0 +1,57 @@
...
+		_ = level.Debug(logutil.WithContext(ctx, logutil.Logger)).Log("msg", "using original limit")
+		return original
+	}
+	return time.Duration(requestLimits.MaxQueryLookback)
+}
+
+// MaxEntriesLimitPerQuery returns the limit to number of entries the querier should return per query.
+func (l *Limiter) MaxEntriesLimitPerQuery(ctx context.Context, userID string) int {
+	original := l.CombinedLimits.MaxEntriesLimitPerQuery(ctx, userID)
+	// in theory this error should never happen
```
jeschkies, line 50: Same here.

pkg/util/querylimits/propagation.go
```diff
@@ -0,0 +1,80 @@
...
+	HTTPHeaderQueryLimitsKey = "X-Loki-Query-Limits"
+)
+
+// NOTE: we use custom `model.Duration` instead of standard `time.Duration` because,
+// to support user-friendly duration format (e.g: "1h30m45s") in JSON value.
+type QueryLimits struct {
+	MaxQueryLength   model.Duration `json:"maxQueryLength"`
+	MaxQueryLookback model.Duration `json:"maxQueryLookback"`
+	// this should be changed to max bytes
+	MaxEntriesLimitPerQuery int      `json:"maxEntriesLimitPerQuery"`
```
jeschkies, line 26: Why is that? I think they both are meaningful. If I'm not mistaken the entry limit is used for log queries.

pkg/util/querylimits/propagation.go
```diff
@@ -0,0 +1,80 @@
...
+
+const (
+	queryLimitsContextKey key = 1
+
+	HTTPHeaderQueryLimitsKey = "X-Loki-Query-Limits"
+)
+
+// NOTE: we use custom `model.Duration` instead of standard `time.Duration` because,
+// to support user-friendly duration format (e.g: "1h30m45s") in JSON value.
+type QueryLimits struct {
```
jeschkies, line 22: Are we dropping the query timeout?

pkg/validation/limits.go
```diff
@@ -95,6 +95,8 @@ type Limits struct {
 	MaxQueriersPerTenant       int            `yaml:"max_queriers_per_tenant" json:"max_queriers_per_tenant"`
 	QueryReadyIndexNumDays     int            `yaml:"query_ready_index_num_days" json:"query_ready_index_num_days"`
 	QueryTimeout               model.Duration `yaml:"query_timeout" json:"query_timeout"`
+	// todo: (callum) cli flag?
```
jeschkies, line 98: Yes. We always enable cli configuration as well.

pkg/validation/limits.go
```diff
@@ -650,6 +652,11 @@ func (o *Overrides) BlockedQueries(ctx context.Context, userID string) []*valida
 	return o.getOverridesForUser(userID).BlockedQueries
 }
 
+func (o *Overrides) RequiredLabels(ctx context.Context, userID string) []string {
+	// todo (callum), this probably needs to take a list of the labels present?
```
jeschkies, line 656: You need to join this list with the one from the context.

pkg/validation/limits_test.go
```diff
@@ -182,6 +183,8 @@ func TestLimitsDoesNotMutate(t *testing.T) {
 		},
 	}
 	SetDefaultLimitsForYAMLUnmarshalling(newDefaults)
+	fmt.Println("initial default limits required labels: ", newDefaults.RequiredLabels)
+	//os.Exit(1)
```
jeschkies, line 187: Was this for debugging?

jeschkies (changes requested):
Thanks. I think we are close. It's advice against introducing new limits in this pull request as this is scope creep. We should land this forest and add the other limits later.

pkg/util/querylimits/grpc.go
```diff
@@ -0,0 +1,107 @@
+package querylimits
```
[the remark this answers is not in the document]
cstyan: I think we should enforce the query limits in the frontend too, which is why I moved this to a separate package.

pkg/util/querylimits/propagation.go
cstyan: I misunderstood an earlier discussion, thinking we were just using MaxEntries as a placeholder for the future MaxBytes limit

cstyan:
> Thanks. I think we are close. It's advice against introducing new limits in this pull request as this is scope creep. We should land this forest and add the other limits later.

Agreed. I removed the addition of the `RequiredLabels`.

Sorry about the messy state for your last review, not sure what happened there. Lots of stuff that I had cleaned up locally somehow didn't make it into a commit. Maybe I messed up my `git commit add --patch` last night.

pkg/loki/loki.go
cstyan: I'm not sure what configurable means in this context.

pkg/validation/limits.go
cstyan: Yes, that would be done in the QueryLimits Limiter when we add this `RequiredLabels` limit (follow up PR)

jeschkies (commented):
Nice. This looks really good. What's missing

- [ ] An integration test.
- [ ] A configuration flag to enable and disable this feature.
- [x] Some clean up for the logger.

I'll see how far I come today.

jeschkies:
@cstyan I've added an integration test. There's a race condition with the modules. Sometimes the test works and sometimes it fails.

pkg/util/querylimits/grpc_test.go
```diff
@@ -0,0 +1,27 @@
...
+		MaxQueryLength:          model.Duration(2 * 24 * time.Hour),
+		MaxQueryLookback:        model.Duration(14 * 24 * time.Hour),
+		MaxEntriesLimitPerQuery: 100,
+	}
+	c1 := InjectQueryLimitsContext(context.Background(), limits)
+
+	c1, err = injectIntoGRPCRequest(c1)
+	require.NoError(t, err)
+
+	c2, err := extractFromGRPCRequest(c1)
```
chaudum, line 24: Can you add a test case where you extract from a gRPC request that hasn't the limits encoded into its metadata?

pkg/util/querylimits/limiter.go
```diff
@@ -0,0 +1,69 @@
...
+		CombinedLimits: original,
+	}
+}
+
+// MaxQueryLength returns the limit of the length (in time) of a query.
+func (l *Limiter) MaxQueryLength(ctx context.Context, userID string) time.Duration {
+	original := l.CombinedLimits.MaxQueryLength(ctx, userID)
+	requestLimits := ExtractQueryLimitsContext(ctx)
+	if requestLimits == nil || requestLimits.MaxQueryLength == 0 || time.Duration(requestLimits.MaxQueryLength) > original {
+		_ = level.Debug(logutil.WithContext(ctx, l.logger)).Log("msg", "using original limit")
```
chaudum, line 31: I don't think that we need to assign to blank identifier.We explicitly ignore that in the linter:

`.golangci.yml`:

```yaml
issues:
  exclude:
    - Error return value of .*log\.Logger\)\.Log\x60 is not checked
    - Error return value of .*.Log.* is not checked
    - Error return value of `` is not checked
```

pkg/util/querylimits/limiter.go
```diff
@@ -0,0 +1,69 @@
...
+
+func NewLimiter(log log.Logger, original limiter.CombinedLimits) *Limiter {
+	return &Limiter{
+		logger:         log,
+		CombinedLimits: original,
+	}
+}
+
+// MaxQueryLength returns the limit of the length (in time) of a query.
+func (l *Limiter) MaxQueryLength(ctx context.Context, userID string) time.Duration {
```
chaudum, line 27: Let's use `tenantID`. I know we already have a mixup of `user`, `userID`, `tenant`, `tenantID` and `orgID`. We should agree on one term, though ...

pkg/loki/loki.go
```diff
@@ -675,6 +665,23 @@ func (t *Loki) setupModuleManager() error {
...
+		deps[Querier] = append(deps[Querier], QueryLimiter)
+		deps[QueryFrontend] = append(deps[QueryFrontend], QueryLimitsTripperware)
+
+		deps[QueryLimiter] = []string{Overrides}
+		deps[QueryLimitsInterceptors] = []string{}
+		deps[QueryLimitsTripperware] = []string{QueryFrontendTripperware}
+
+		if err := mm.AddDependency(Server, QueryLimitsInterceptors); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
cstyan: Deps doesn't already have an entry for `Server`, not sure if that's intentional or not. We end up looping through `deps` anyways and calling `AddDependency`, so it's essentially the same as if we had  a `Server` entry in `deps`.

pkg/util/querylimits/limiter.go
cstyan: Agreed. I've been meaning to resurrect the PR that changes all uses of these terms to `tenantID`.

pkg/loki/loki.go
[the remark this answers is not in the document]
jeschkies: Yes. I didn't do it because Deos was missing it

jeschkies (approved):

PR #8734, opened 2023-03-07
Reimplement JsonExpressionParser in terms of jsonparser 

MasslessParticle:
This PR reimplements the functionality  of `JSONExpressionParser` using the `jsonparser` library rather than `jsoniter` 

Benchmarks:
```
benchstat json_expression_old.txt json_expression_new.txt 
name                                    old time/op    new time/op    delta
JsonExpressionParser/json-expression-8    4.15µs ± 4%    0.82µs ± 3%  -80.17%  (p=0.000 n=10+10)

name                                    old alloc/op   new alloc/op   delta
JsonExpressionParser/json-expression-8    1.07kB ± 0%    0.02kB ± 0%  -97.76%  (p=0.000 n=10+10)

name                                    old allocs/op  new allocs/op  delta
JsonExpressionParser/json-expression-8      72.0 ± 0%       2.0 ± 0%  -97.22%  (p=0.000 n=10+10)
```


pkg/logql/log/parser.go
```diff
@@ -460,39 +460,89 @@ func NewJSONExpressionParser(expressions []LabelExtractionExpr) (*JSONExpression
...
+		keys:  internedStringSet{},
 	}, nil
 }
 
+func pathsToString(paths []interface{}) []string {
+	stingPaths := make([]string, 0, len(paths))
+	for _, p := range paths {
+		switch v := p.(type) {
+		case int:
+			stingPaths = append(stingPaths, fmt.Sprintf("[%d]", v))
```
MasslessParticle, line 479: The index syntax for paths in `jsonparser` differs slightly from `jsoniter`. This function turns our representation into what is expected by the new implementation

pkg/logql/log/parser.go
```diff
@@ -460,39 +460,89 @@ func NewJSONExpressionParser(expressions []LabelExtractionExpr) (*JSONExpression
...
 		lbs.SetErr(errJSON)
 		return line, true
 	}
 
-	for identifier, paths := range j.expressions {
-		result := jsoniter.ConfigFastest.Get(line, paths...).ToString()
+	var matches int
+	jsonparser.EachKey(line, func(idx int, data []byte, typ jsonparser.ValueType, err error) {
+		if err != nil {
+			lbs.SetErr(errJSON)
```
dannykopping, line 503: I see we're not actually returning `err` here, so might that not make it difficult to troubleshoot, or identify exactly where the problem occurred?

dannykopping (approved):
Great work @MasslessParticle! Really excited for this. I'd be keen to see how this stacks up against the regular json parser; if it's now faster/more resource-efficient, we can start recommending its use when specific keys are needed.

Added a comment about error handling, but overall LGTM

pkg/logql/log/parser.go
```diff
@@ -460,39 +460,89 @@ func NewJSONExpressionParser(expressions []LabelExtractionExpr) (*JSONExpression
...
+			}
+		}
 	}
+
 	return line, true
 }
 
+func isValidJSONStart(data []byte) bool {
+	switch data[0] {
+	case '"', '{', '}', '[', ']':
```
chaudum, line 539: Are closing brackets really valid JSON start tokens?

MasslessParticle:
@dannykopping Here's @cyriltovena's original PR for JSON Parser: https://github.com/grafana/loki/pull/7723.

the `jsonparser` library helped somewhat. [this]() PR ought to buy us quite a bit -- at lease for metrics queries: https://github.com/grafana/loki/pull/8724

pkg/logql/log/parser.go
MasslessParticle: Good catch, thanks! This is copy paste from the parser internals

pkg/logql/log/parser.go
MasslessParticle: I'm not worried about this. As near as I can tell, errors that occur in here are due solely to malformed json. Putting the err on the labels and moving on is consistent with how we treat errors in other parsers.

dannykopping: OK cool 👍

MasslessParticle:
@dannykopping It's interesting to note that when the JSON payload is small and doesn't contain nested keys, the JSON Parser performs about as well as `logfmt`. [These](https://github.com/grafana/loki/pull/8724) benchmarks use a simpler JSON that @cyriltovena's

PR #8744, opened 2023-03-08
Ruler: remote rule evaluation

dannykopping:
Adds the ability to evaluate recording & alerting rules against a given `query-frontend`, allowing these queries to be executed with all the parallelisation & optimisation that regular adhoc queries have. This is important because with `local` evaluation all queries are single-threaded, and rules that evaluate a large range/volume of data may timeout or OOM the `ruler` itself, leading to missed metrics or alerts.

When `remote` evaluation mode is enabled, the `ruler` effectively just becomes a gRPC client for the `query-frontend`, which will dramatically improve the reliability of the `ruler` and also drastically reduce its resource requirements.

This PR implements the feature discussed in https://github.com/grafana/loki/pull/8129 (**LID 0002: Remote Rule Evaluation**).

This PR is the first of two; in this PR I have just implemented the happy path. In a subsequent PR I will address all the unhappy paths and add some configurability to this feature. I didn't want to complicate this (already large) PR with all of those failure cases, as a couple of these will require some deliberation. That subsequent PR will also contain documentation.

Also, I added a couple drive-by improvements along the way:

- refactored our integration tests a little so the ruler config could be controlled by tests in a cleaner way
  - also added the ability to provide extra config fragments which are merged into the main Loki config; this makes tests quite clean and clear now I think
- added a `loki-backend` component to our `production/docker` docker-compose setup, as well as configured the ruler and added a couple sample rules
  - also added remote-writing of metrics to the `prometheus` instance

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/integration/client/client.go
+++ b/integration/client/client.go
@@ -434,7 +434,7 @@ func (c *Client) GetRules(ctx context.Context) (*RulesResponse, error) {
 	resp := RulesResponse{}
 	err = json.Unmarshal(buf, &resp)
 	if err != nil {
-		return nil, fmt.Errorf("error parsing response data: %w", err)
+		return nil, fmt.Errorf("error parsing response data %q: %w", buf, err)
 	}
 
 	return &resp, err
--- a/integration/util/merger.go
+++ b/integration/util/merger.go
@@ -1,0 +1,54 @@
+package util
+
+import (
+	"fmt"
+
+	"github.com/imdario/mergo"
+	"gopkg.in/yaml.v2"
+)
+
+// YAMLMerger takes a set of given YAML fragments and merges them into a single YAML document.
+// The order in which these fragments is supplied is maintained, so subsequent fragments will override preceding ones.
+type YAMLMerger struct {
+	fragments [][]byte
+}
+
+func NewYAMLMerger() *YAMLMerger {
+	return &YAMLMerger{}
+}
+
+func (m *YAMLMerger) AddFragment(fragment []byte) {
+	m.fragments = append(m.fragments, fragment)
+}
+
+func (m *YAMLMerger) Merge() ([]byte, error) {
+	merged := make(map[interface{}]interface{})
+	for _, fragment := range m.fragments {
+		fragmentMap, err := yamlToMap(fragment)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal given fragment %q to map: %w", fragment, err)
+		}
+
+		if err = mergo.Merge(&merged, fragmentMap, mergo.WithOverride, mergo.WithTypeCheck); err != nil {
+			return nil, fmt.Errorf("failed to merge fragment %q with base: %w", fragment, err)
+		}
+	}
+
+	mergedYAML, err := yaml.Marshal(merged)
+	if err != nil {
+		return nil, err
+	}
+
+	return mergedYAML, nil
+}
+
+func yamlToMap(fragment []byte) (interface{}, error) {
+	var fragmentMap map[interface{}]interface{}
+
+	err := yaml.Unmarshal(fragment, &fragmentMap)
+	if err != nil {
+		return nil, err
+	}
+
+	return fragmentMap, nil
+}
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -356,6 +356,7 @@ type Loki struct {
 	tableManager             *index.TableManager
 	frontend                 Frontend
 	ruler                    *base_ruler.Ruler
+	ruleEvaluator            ruler.Evaluator
 	RulerStorage             rulestore.RuleStore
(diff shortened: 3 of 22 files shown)


pkg/querier/queryrange/codec.go
```diff
@@ -805,6 +817,19 @@ func toProtoVector(v loghttp.Vector) []queryrangebase.SampleStream {
 	return res
 }
 
+func toProtoScalar(v loghttp.Scalar) []queryrangebase.SampleStream {
+	res := make([]queryrangebase.SampleStream, 0, 1)
+
+	res = append(res, queryrangebase.SampleStream{
+		Samples: []logproto.LegacySample{{
+			Value:       float64(v.Value),
+			TimestampMs: v.Timestamp.UnixNano() / 1e6,
```
chaudum, line 826: nit: There is a `v.Timestamp.UnixMilli()` function for that.

integration/client/client.go
```diff
@@ -434,7 +434,7 @@ func (c *Client) GetRules(ctx context.Context) (*RulesResponse, error) {
 	resp := RulesResponse{}
 	err = json.Unmarshal(buf, &resp)
 	if err != nil {
-		return nil, fmt.Errorf("error parsing response data: %w", err)
+		return nil, fmt.Errorf("error parsing response data %q: %w", buf, err)
```
kavirajk, line 437: Q: given this buffer contains results of `GetRules`. can this be huge enough to cause trouble? (also can it make too verbose error message?)

pkg/loki/loki.go
```diff
@@ -356,6 +356,7 @@ type Loki struct {
 	tableManager             *index.TableManager
 	frontend                 Frontend
 	ruler                    *base_ruler.Ruler
+	ruleEvaluator            ruler.Evaluator
```
kavirajk, line 359: Q: `ruleEvaluator` runs as a part of `ruler` component/process correct?

pkg/querier/queryrange/codec.go
```diff
@@ -521,6 +521,18 @@ func (Codec) DecodeResponse(ctx context.Context, r *http.Response, req queryrang
 				},
 				Statistics: resp.Data.Statistics,
 			}, nil
+		case loghttp.ResultTypeScalar:
```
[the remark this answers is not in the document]
kavirajk: Nice!. Can we lock this in test as well if possible?

kavirajk (approved):
Looks super nice :+1:

pkg/ruler/evaluator_remote.go
```diff
@@ -0,0 +1,263 @@
...
+// Query performs a query for the given time.
+func (q *remoteQuerier) Query(ctx context.Context, qs string, t time.Time) (*logqlmodel.Result, error) {
+	logger, ctx := spanlogger.NewWithLogger(ctx, q.logger, "ruler.remoteEvaluation.Query")
+	defer logger.Span.Finish()
+
+	res, err := q.query(ctx, qs, t, logger)
+	if err != nil {
+		return nil, err
+	}
+	return res, err
```
chaudum, line 147: nit: should be identical to:

```suggestion
	return q.query(ctx, qs, t, logger)
```

chaudum (commented):
LGTM :rocket: 
I added a few comments of minor importance, but don't want to block merging.

integration/util/merger.go
```diff
@@ -0,0 +1,54 @@
+package util
```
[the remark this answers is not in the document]
chaudum: Would have been nice if that change would be a separate PR (would reduce size of this one)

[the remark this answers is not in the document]
dannykopping: I needed these changes to make the integration tests work; but yeah I suppose I could've left the integration tests for the second PR as well. I didn't know from the beginning I would need a subsequent PR...

pkg/querier/queryrange/codec.go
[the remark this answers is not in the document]
dannykopping: Yup, in the subsequent PR

pkg/loki/loki.go
dannykopping: Correct

pkg/ruler/evaluator_remote.go
dannykopping: You're right, I think I had some error wrapping there before but removed it.

pkg/querier/queryrange/codec.go
dannykopping: Ah, right, this is a `prometheus/common/model.Time` so there's no UnixMilli() available

weizhengde:
when this feature will be released

dannykopping:
We have a 2.9 release coming soon which will include this

PR #8750, opened 2023-03-08
refactor(lambda-promtail): apply terraform best practices

mhulscher:

I would like to offer this PR as a suggestion to improve the lambda-promtail terraform-module. I forked it to be able to deploy it more than once in an AWS account. I also applied terraform best-practices. I was hoping that perhaps these changes could be merged into upstream as well.

Unlike https://github.com/grafana/loki/pull/8549 , I unfortunately did not end up making a separate commit for each change. 

If you would like me to create one or more issue(s) to address the points below, I'd be happy to do that as well.

List of improvements:

1. Added `var.name` (defaults to lambda-promtail) so that this module can be deployed multiple times in the same AWS account. This allows us to define unique, non-conflicting names for:
    * the Lambda function
    * the CloudWatch log-group
    * the IAM role
2. Split IAM role policies per component; only assign permissions when required
3. Scope down permissions of the IAM role policies
4. During terraform-destroy, ensure CloudWatch log-group is removed **after** the lambda-function. An accidental invocation of the function could re-create an already destroyed log-group, leaving an orphaned log-group

List of style changes:

1. Rename resources to `this` when there is only one instance of this resource-type
2. Add newline after `count|before_each` and before `depends_on`
3. Group resources together and add a section comment
4. Add missing(?) statement-id to S3 AWS lambda permission

Misc.

1. I added a `moves.tf` file to facilitate moving renamed resources in existing terraform statefiles. This prevents some resources from recreated. Can also be removed.

These changes are backwards compatible, even though some resources will end up being re-created. A `terraform apply` should succeed (it did for me).

**Checklist**

### DIFF
--- a/tools/lambda-promtail/main.tf
+++ b/tools/lambda-promtail/main.tf
@@ -1,92 +1,157 @@
 data "aws_region" "current" {}
 
-resource "aws_iam_role" "iam_for_lambda" {
-  name = "iam_for_lambda"
+#-------------------------------------------------------------------------------
+# IAM role assigned to the lambda function
+#-------------------------------------------------------------------------------
 
-  assume_role_policy = jsonencode({
-    "Version" : "2012-10-17",
-    "Statement" : [
-      {
-        "Action" : "sts:AssumeRole",
-        "Principal" : {
-          "Service" : "lambda.amazonaws.com"
-        },
-        "Effect" : "Allow",
-      }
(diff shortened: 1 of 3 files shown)


MasslessParticle (approved):

MasslessParticle:
This looks great! Thanks for contributing!

PR #8753, opened 2023-03-08
Loki: Change the behavior of the 0 value for retention_period to disable retention.

### LINKED ISSUE #8359: Compactor deleted all the index structure from S3 after retention_enabled change
**Describe the bug**
During a compactor configuration change to be able to delete log entries, following this [procedure](https://grafana.com/docs/loki/latest/operations/storage/logs-deletion/), compactor deleted all our index folders from S3, so due this we are not able to search old data.

**To Reproduce**
1. Started Loki Distributed using version 2.6
2. Configure compactor settings as below, the previous value was `false`

`    compactor:
         retention_enabled: true
         deletion_mode: filter-and-delete
`

**Expected behavior**
Loki index folders remain as is on S3

**Environment:**
 - Infrastructure: EKS
 - Loki Version: loki-distributed-0.55.1	2.6.1

**Logs**

`
2023-01-31 10:13:25 | {"log":"level=info ts=2023-01-31T13:13:25.111455971Z caller=index_set.go:281 table-name=loki-distributed-index-19109 msg=\"removing source db files from storage\" count=1\n","stream":"stderr","time":"2023-01-31T13:13:25.111526417Z"}
`

slim-bean:

Issue #8359 exposed a very unfortunate inconsistency in a Loki config where the zero value did not disable a feature.

By convention in Loki the zero value for a config typically "disables" it.

This PR changes the retention_period to honor the zero value of `0` or `0s` to disable retention.

Fixes #8359

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -2333,9 +2333,10 @@ ruler_remote_write_sigv4_config:
 # CLI flag: -compactor.deletion-mode
 [deletion_mode: <string> | default = "filter-and-delete"]
 
-# Retention to apply for the store, if the retention is enabled on the compactor
-# side.
-# A zero value of 0 or 0s disables retention.
+# Retention period to apply to stored data, only applies if retention_enabled is
+# true in the compactor config. As of version 2.8.0, a zero value of 0 or 0s
+# disables retention. In previous releases, Loki did not properly honor a zero
+# value to disable retention and a really large value should be used instead.
 # CLI flag: -store.retention
 [retention_period: <duration> | default = 0s]
 
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -2335,8 +2335,9 @@ ruler_remote_write_sigv4_config:
 
 # Retention to apply for the store, if the retention is enabled on the compactor
 # side.
+# A zero value of 0 or 0s disables retention.
 # CLI flag: -store.retention
-[retention_period: <duration> | default = 31d]
+[retention_period: <duration> | default = 0s]
 
 # Per-stream retention to apply, if the retention is enable on the compactor
 # side.
--- a/docs/sources/operations/storage/logs-deletion.md
+++ b/docs/sources/operations/storage/logs-deletion.md
@@ -21,6 +21,8 @@ Log entry deletion relies on configuration of the custom logs retention workflow
 
 Enable log entry deletion by setting `retention_enabled` to true in the compactor's configuration and setting and `deletion_mode` to `filter-only` or `filter-and-delete` in the runtime config.
 
+> **Warning:** Please be very careful when enabling retention, it's strongly recommended to enable versioning on your objects in object storage to allow for recovery from accidental misconfiguration of a retention setting. If you wish to enable delete capability but do not want to enforce any retention, make sure you configure retention_period with a value of `0s`. 
+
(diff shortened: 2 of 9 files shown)


DylanGuedes (commented):
the new default is indeed better/more safe but would be cool to have the default change listed on the `docs/sources/upgrading/_index.md` too.

docs/sources/operations/storage/logs-deletion.md
```diff
@@ -21,6 +21,8 @@ Log entry deletion relies on configuration of the custom logs retention workflow
 
 Enable log entry deletion by setting `retention_enabled` to true in the compactor's configuration and setting and `deletion_mode` to `filter-only` or `filter-and-delete` in the runtime config.
 
+> **Warning:** Please be very careful when enabling retention, it's strongly recommended to enable versioning on your objects in object storage to allow for recovery from accidental misconfiguration of a retention setting. If you wish to enable delete capability but do not want to enforce any retention, make sure you configure retention_period with a value of `0s`.
```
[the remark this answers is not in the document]
JStickler: >I was debating if we should be more explicit here that previous versions of Loki did not properly honor the zero value. This would be somewhat out of character of having our versioned docs mention previous versions but I wonder if that's prudent here.

I would definitely call out this change of behavior in the Release Notes.  And it might be worth a note in the Configuration topic to flag the change.

docs/sources/configuration/_index.md
```diff
@@ -2343,8 +2343,9 @@ ruler_remote_write_sigv4_config:
 
 # Retention to apply for the store, if the retention is enabled on the compactor
 # side.
+# A zero value of 0 or 0s disables retention.
```
JStickler, line 2346:
```suggestion
# As of version 2.8.0, a zero value of 0 or 0s disables retention.  In previous releases, Loki did not properly honor a zero value to disable retention.
```

docs/sources/operations/storage/logs-deletion.md
JStickler, line 24:
```suggestion
> **Warning:** Be very careful when enabling retention.  It is strongly recommended that you also enable versioning on your objects in object storage to allow you to recover from accidental misconfiguration of a retention setting.  If you want to enable deletion but not not want to enforce retention, configure the `retention_period` setting with a value of `0s`.
```

docs/sources/upgrading/_index.md
```diff
@@ -33,6 +33,31 @@ The output is incredibly verbose as it shows the entire internal config struct u
...
+```yaml
+compactor:
+  retention_enabled: true
+```
+
+And did *not* define a `retention_period` in `limits_config`, thus relying on the previous default of `744h`
+
+In this release the default has been changed to `0s`.
+
+A value of `0s` is the same as "retain forever" or "disable retention".
```
kavirajk, line 50: > A value of `0s` is the same as "retain forever" or "disable retention".

I think it's worth adding this into our config doc as well.

kavirajk (approved):
LGTM :+1: with one minor comment

docs/sources/upgrading/_index.md
slim-bean: we do have ` As of version 2.8.0, a zero value of 0 or 0s disables retention.` in the config section already, seems like this should be good enough?

PR #8755, opened 2023-03-09
index-gateway: fix failure in initializing index gateway when boltdb-shipper is not being used

sandeepsukhani:
When only `tsdb` index is being used in the schema without configuring `boltdb-shipper` index store, we fail to initialize `index-gateway` since it also tries to build `boltdb-shipper` index client. This reference is used for serving `QueryIndex` calls which are for serving index from boltdb files.

This PR fixes the issue by adding a check for the existence of `boltdb-shipper` in the schema config before initializing an index client. On the index gateway side, it adds a `failingIndexClient`, which errors when someone tries to query it.

**Checklist**
- [ ] `CHANGELOG.md` updated


slim-bean:
I think this will need to be solved a bit differently.

We need the index client for TSDB also, the problem is we hard coded the creation of the index client to look for boltdb specific directories

CHANGELOG.md
```diff
@@ -42,7 +42,8 @@
 * [8251](https://github.com/grafana/loki/pull/8251) **sandeepsukhani** index-store: fix indexing of chunks overlapping multiple schemas.
 * [8151](https://github.com/grafana/loki/pull/8151) **sandeepsukhani** fix log deletion with line filters.
 * [8448](https://github.com/grafana/loki/pull/8448) **chaudum**: Fix bug in LogQL parser that caused certain queries that contain a vector expression to fail.
-* [8448](https://github.com/grafana/loki/pull/8665) **sandeepsukhani**: deletion: fix issue in processing delete requests with tsdb index
+* [8448](https://github.com/grafana/loki/pull/8665) **sandeepsukhani**: deletion: fix issue in processing delete requests with tsdb index.
+* [8448](https://github.com/grafana/loki/pull/8755) **sandeepsukhani**: index-gateway: fix failure in initializing index gateway when boltdb-shipper is not being used.
```
DylanGuedes, line 46:
```suggestion
* [8665](https://github.com/grafana/loki/pull/8665) **sandeepsukhani**: deletion: fix issue in processing delete requests with tsdb index.
* [8775](https://github.com/grafana/loki/pull/8755) **sandeepsukhani**: index-gateway: fix failure in initializing index gateway when boltdb-shipper is not being used.
```

sandeepsukhani:
> I think this will need to be solved a bit differently.
> 
> We need the index client for TSDB also, the problem is we hard coded the creation of the index client to look for boltdb specific directories

We initialize tsdb index client with store and the store instance is passed to [index gateway factory function](https://github.com/grafana/loki/pull/8755/files#diff-dfe68c6d3018ac004616d53414bdfc82d0bbb04eea91fdff7f63e0369b22b022R1049) .

slim-bean:
Sorry I didn't get back to this soon enough, but this PR doesn't fix the inital problem I was investigating.

I'm not sure if it fixes another problem? Or if it should be reverted and replaced with another fix?

The problem is:

Take the `loki-local-config.yaml` in `cmd/loki` and change the index type from `boltdb-shipper` to `tsdb-shipper` and try to start Loki, you will get an error.

There may be multiple issues here, but one of them is we hard code stuff to boltdb-shipper when initializing stores. This doesn't fail for us because we have existing boltdb-shipper configs which allow this to work.

However, if you are starting new and only want to run TSDB you currently can't (without defining some boltdb storage_configs to workaround this issue)

PR #8758, opened 2023-03-09
Modify debug loki build to work on any architecture, including ARM

dannykopping:
On my M1 Pro, `make loki-debug-image` produces the following output:

```bash
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
```

This is because we run `delve` as the entrypoint for the debug image, and `delve` is being copied from `loki-build-image` which is exclusively built as `amd64` (we should make this a multi-arch image... created https://github.com/grafana/loki/issues/8759).

This change will work with whichever host OS is building the image and download `delve` using that architecture instead of relying on the `loki-build-image` version.

N/A

If you're on an `amd64` machine, please run `make loki-debug-image` and run the resultant image to make sure that things still work as before.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/cmd/loki/Dockerfile.debug
+++ b/cmd/loki/Dockerfile.debug
@@ -1,16 +1,24 @@
+ARG BUILD_IMAGE=grafana/loki-build-image:0.28.1
 # Directories in this file are referenced from the root of the project not this folder
 # This file is intended to be called from the root like so:
 # docker build -t grafana/loki -f cmd/loki/Dockerfile.debug .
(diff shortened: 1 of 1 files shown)


MichelHollands (approved):
Built it on amd64 and it runs fine as well.

dannykopping:
Thanks a lot @MichelHollands!

PR #8761, opened 2023-03-09
Loki: remove `subqueries` from metrics.go logging and replace it with separate split and shard counters

slim-bean:

Currently the `metrics.go` log line emitted after every query includes a metric called "subqueries".  This currently tracks the number of queries created by the split_by_time operations done in Loki, but does not include any counts for subqueries created as a result of sharding.

It becomes difficult to make a single subqueries counter that gives useful information to determine how much a query is split by time and sharded by a shard factor, especially now that sharding in TSDB indexes is dynamic.

This PR removes and deprecates the `subqueries` stat and instead creates a `splits` and `shards` statistic which records how much a query was split_by_time and the total number of shards created as well.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/querier/queryrange/codec_test.go
```diff
@@ -1010,7 +1012,9 @@ var (
 			"execTime": 22,
 			"linesProcessedPerSecond": 23,
 			"queueTime": 21,
-			"subqueries": 1,
+			"shards": 0,
+            "splits": 0,
+			"subqueries": 0,
```
kavirajk, line 1017: 1. `splits` is not formatted correctly looks like.
2. I would assume there shouldn't be `subqueries` keys here.

pkg/querier/queryrange/prometheus_test.go
```diff
@@ -81,6 +81,8 @@ var emptyStats = `"stats": {
 		"execTime": 0,
 		"linesProcessedPerSecond": 0,
 		"queueTime": 0,
+        "splits": 0,
```
kavirajk, line 84: formatting again.

pkg/logql/metrics_test.go
```diff
@@ -111,7 +111,7 @@ func TestLogLabelsQuery(t *testing.T) {
 	})
 	require.Equal(t,
 		fmt.Sprintf(
-			"level=info org_id=foo traceID=%s latency=slow query_type=labels length=1h0m0s duration=25.25s status=200 label=foo throughput=100kB total_bytes=100kB total_entries=12\n",
+			"level=info org_id=foo traceID=%s latency=slow query_type=labels length=1h0m0s duration=25.25s status=200 label=foo splits=0 throughput=100kB total_bytes=100kB total_entries=12\n",
```
kavirajk, line 114: I see we have test only for LabelQuery and SeriesQuery here. Good to add a test for `InstantQuery` as well? to see both `splits` and `shard` appear?

kavirajk (commented):
Nice. thanks for adding this. LGTM with few minor comments.

kavirajk:
wondering this may break existing dashboards based on `subqueries`. maybe add an entry in upgrade guide as well?

pkg/logql/metrics_test.go
slim-bean: I'm not entirely sure what these tests are actually testing? I just updated it to it wasn't failing but it seems strange to me to have a test that's validating the output of a log line?

I can see some arguments for this but we don't really have any contracts on log lines so I'm not sure we really need these tests?

pkg/querier/queryrange/codec_test.go
slim-bean: subqueries still exists because it is deprecated but not removed.

slim-bean:
> wondering this may break existing dashboards based on `subqueries`. maybe add an entry in upgrade guide as well?

added!

pkg/logql/metrics_test.go
MasslessParticle: I'd say that we should probably never test log lines

MasslessParticle (approved):

PR #8706, opened 2023-03-06
Tools: add tool to determine which release (if any) a commit appears in

dannykopping:
I've often found myself needing to determine which release a particular PR appears in. This tool makes that easy, and it accepts the merge commit ref of the PR.

Example from https://github.com/grafana/loki/pull/7044, in which [a user is asking which version this feature appears in](https://github.com/grafana/loki/pull/7044#issuecomment-1455694087).

```bash
$ ./tools/which-release.sh 93a5a71                                 
Commit was found in the following releases:
  release-2.7.x
```

PR #8768, opened 2023-03-10
Expand which-release to cover weekly builds as well

dannykopping:
Expanding tool introduced in https://github.com/grafana/loki/pull/8706

Knowing which, if any, weekly builds the commit is in can also be useful.


kavirajk (approved):
LGTM :+1:

PR #8779, opened 2023-03-12
Loki: Add UsageStatsURL to usagestats Config

### LINKED ISSUE #6502: Allow setting an alternate URL for usage reporting
I want to enable usage reporting, but the network setup of my application doesn't allow it. Specifically, the host that I run Loki cannot make requests to the internet. Since we have *very* few places where our system needs to make requests out, we have not created any kind of generic web proxy for outgoing connections so I can't just set up a standard proxy.

I would like to have Loki make its requests to a configurable alternate URL. In my specific case, that URL would call an AWS lambda via HTTPS, and the lambda would then send the data on to stats.grafana.org. This gives me control and visibility over what data leaves our system and how it goes out.

I *could* implement this by adding an entry for stats.grafana.org in my private DNS and then adding my own certificate authority to the OS so that the TLS request continues to work. It will be difficult to justify spending that much time on this problem to my project manager.

I could also implement this by adding inspecting web proxy software, which would essentially be having a third party do the certificate authority and server certificate solution specified above. That would be expensive and time-consuming and, again, would only be used to send usage reports for Loki. While I believe that sending usage reports back is the right thing to do (and can benefit me directly with expanded features or continued support), it's difficult to justify the time and expense of that solution to my project manager as well.

eisandbar:

Reports used to be sent to a fixed URL usageStatsURL.
Now this url can be configurable while using the old url as a default.
Updated tests to use config for custom URLs instead of changing package variable.

Fixes #6502

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -2797,6 +2797,10 @@ Configuration for usage report.
 # Enable anonymous usage reporting.
 # CLI flag: -reporting.enabled
 [reporting_enabled: <boolean> | default = true]
+
+# URL to which reports are sent
+# CLI flag: -reporting.usage-stats-url
+[usage_stats_url: <string> | default = "https://stats.grafana.org/loki-usage-report"]
 ```
 
 ### common
--- a/pkg/usagestats/reporter.go
+++ b/pkg/usagestats/reporter.go
@@ -40,13 +40,15 @@ var (
 )
 
 type Config struct {
-	Enabled bool `yaml:"reporting_enabled"`
-	Leader  bool `yaml:"-"`
+	Enabled       bool   `yaml:"reporting_enabled"`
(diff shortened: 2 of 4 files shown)


JStickler (approved):
[docs team]  What you've got for docs LGTM.   But as a non-developer, I have a few questions.  

Does this need to be a fully qualified URL?  If yes, we should document that requirement.

Is there any validation written into the new code that Loki can reach the custom URL?  What happens if the URL is malformed or unreachable?  Does Loki fall back to the default or does the usage report not get sent anywhere?  (Sorry, I don't know Go, my limited programming experience is with Java and it was a long time ago...)

MasslessParticle (approved):

PR #8785, opened 2023-03-13
Ruler: remote rule evaluation hardening

dannykopping:
This PR is part 2 of 2 implementing remote rule evaluation.
See [part 1](https://github.com/grafana/loki/pull/8744) for more context.

N/A

I've added some config to control rule evaluation as well as some metrics.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


docs/sources/operations/scalability.md
```diff
@@ -27,3 +27,63 @@ The query scheduler process itself can be started via the `-target=query-schedul
 In compute-constrained environments, garbage collection can become a significant performance factor. Frequently-run garbage collection interferes with running the application by using CPU resources. The use of memory ballast can mitigate the issue. Memory ballast allocates extra, but unused virtual memory in order to inflate the quantity of live heap space. Garbage collection is triggered by the growth of heap space usage. The inflated quantity of heap space reduces the perceived growth, so garbage collection occurs less frequently.
 
 Configure memory ballast using the ballast_bytes configuration option.
+
+## Remote Rule Evaluation
+
+This feature was introduced as part of [`LID-0002`](https://github.com/grafana/loki/pull/8129).
```
JStickler, line 33: I don't think this line is necessary.  
It is also incorrect, the LID is just the design doc and discussion, it doesn't change the code or introduce new features.

docs/sources/operations/scalability.md
```diff
@@ -27,3 +27,63 @@ The query scheduler process itself can be started via the `-target=query-schedul
...
+
+### Background & Identification of Problem
```
JStickler, line 35: Your overview text is fine, but I would remove both of the headings:
"Background & Identification of Problem" 
"Recommended Solution"
You're documenting a feature here, not a bug fix.  All of the paragraph text from lines 37 - 51 can sit right under the "Remote rule evaluation" heading.

docs/sources/operations/scalability.md
```diff
@@ -27,3 +27,63 @@ The query scheduler process itself can be started via the `-target=query-schedul
 In compute-constrained environments, garbage collection can become a significant performance factor. Frequently-run garbage collection interferes with running the application by using CPU resources. The use of memory ballast can mitigate the issue. Memory ballast allocates extra, but unused virtual memory in order to inflate the quantity of live heap space. Garbage collection is triggered by the growth of heap space usage. The inflated quantity of heap space reduces the perceived growth, so garbage collection occurs less frequently.
 
 Configure memory ballast using the ballast_bytes configuration option.
+
+## Remote Rule Evaluation
```
JStickler, line 31:
```suggestion
## Remote rule evaluation
```

docs/sources/operations/scalability.md
```diff
@@ -27,3 +27,63 @@ The query scheduler process itself can be started via the `-target=query-schedul
...
+improve rule evaluation performance and lead to fewer missed iterations.
+
+It is generally recommended to create a separate `query-frontend` deployment and `querier` pool from your existing one - which handles adhoc
+queries via Grafana, `logcli`, or the API. Rules should be given priority over adhoc queries because they are used to produce
+metrics or alerts which may be crucial to the reliable operation of your service; if you use the same `query-frontend` and `querier` pool
+for both, your rules will be executed with the same priority as adhoc queries which could lead to unpredictable performance.
+
+### Implementation
+
+To enable remote rule evaluation, set the following configuration options:
```
JStickler, line 55: Where do you set this configuration?  In which file?

docs/sources/operations/scalability.md
```diff
@@ -27,3 +27,63 @@ The query scheduler process itself can be started via the `-target=query-schedul
...
+    mode: remote
+    query_frontend:
+      address: dns:///<query-frontend-service>:<grpc-port>
+```
+
+The `ruler` component will become simply a gRPC client to the `query-frontend` service (side-note: this will result in much lower
+resource usage of your `ruler`s!). Rules' LogQL queries will be executed against the given `query-frontend` service.
+Requests will be load-balanced across all `query-frontend` IPs if the `dns:///` prefix is used.
+
+Queries that fail to execute are _not_ retried at this time.
```
JStickler, line 69: I think this is worth calling out in a note.
```suggestion
> **Note:** Queries that fail to execute are _not_ retried.
```

docs/sources/operations/scalability.md
```diff
@@ -27,3 +27,63 @@ The query scheduler process itself can be started via the `-target=query-schedul
...
+
+```yaml
+ruler:
+  evaluation:
+    mode: remote
+    query_frontend:
+      address: dns:///<query-frontend-service>:<grpc-port>
+```
+
+The `ruler` component will become simply a gRPC client to the `query-frontend` service (side-note: this will result in much lower
```
JStickler, line 65:
```suggestion
When you enable remote rule evaluation, the `ruler` component becomes a gRPC client to the `query-frontend` service.  This may result in lower resource usage of your rulers.
```

docs/sources/operations/scalability.md
```diff
@@ -27,3 +27,63 @@ The query scheduler process itself can be started via the `-target=query-schedul
...
+resource usage of your `ruler`s!). Rules' LogQL queries will be executed against the given `query-frontend` service.
```
JStickler, line 66:
```suggestion
The LogQL queries coming from the Ruler will be executed against the given `query-frontend` service.
```

JStickler (commented):
[docs team]

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ##### Enhancements
 
+* [8785](https://github.com/grafana/loki/pull/8785) **dannykopping**: Ruler: remote rule evaluation hardening.
```
chaudum, line 11: I don't think this needs an extra changelog entry. It's part of the "remote rule evaluation" feature.

docs/sources/operations/scalability.md
```diff
@@ -27,3 +27,63 @@ The query scheduler process itself can be started via the `-target=query-schedul
...
+
+### Background & Identification of Problem
+
+By default, the `ruler` component embeds a query engine to evaluate rules. This generally works fine, except when rules
+are complex or have to process a large amount of data regularly. Poor performance of the `ruler` manifests as recording rules metrics
+with gaps or missed alerts. This situation can be detected by alerting on the `cortex_prometheus_rule_group_iterations_missed_total` metric
+when it has a non-zero value.
+
+### Recommended Solution
+A solution to this problem is to externalize rule evaluation from the `ruler` process. The `ruler` embedded query engine
```
chaudum, line 43: Do we have a guideline whether to have a newline between heading and paragraph? I see both variants.

docs/sources/configuration/_index.md
```diff
@@ -2411,6 +2411,14 @@ ruler_remote_write_sigv4_config:
 # remote client id as key.
 [ruler_remote_write_config: <map of string to RemoteWriteConfig>]
 
+# Timeout for a remote rule evaluation. Defaults to the value of
+# 'querier.query-timeout'.
+[ruler_remote_evaluation_timeout: <duration>]
+
+# Maximum size (in bytes) of the allowable response size from a remote rule
+# evaluation. Set to 0 to disable.
```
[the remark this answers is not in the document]
dannykopping: This doesn't actually disable remote rule evaluation; it disables the size limitation. I could make that more clear...

How's this?

```suggestion
# evaluation. Set to 0 to allow responses of any size.
```

pkg/validation/limits.go
```diff
@@ -143,6 +143,10 @@ type Limits struct {
 
 	RulerRemoteWriteConfig map[string]config.RemoteWriteConfig `yaml:"ruler_remote_write_config,omitempty" json:"ruler_remote_write_config,omitempty" doc:"description=Configures global and per-tenant limits for remote write clients. A map with remote client id as key."`
 
+	// TODO(dannyk): possible enhancement is to align this with rule group interval
+	RulerRemoteEvaluationTimeout         time.Duration `yaml:"ruler_remote_evaluation_timeout" json:"ruler_remote_evaluation_timeout" doc:"description=Timeout for a remote rule evaluation. Defaults to the value of 'querier.query-timeout'."`
+	RulerRemoteEvaluationMaxResponseSize int64         `yaml:"ruler_remote_evaluation_max_response_size" json:"ruler_remote_evaluation_max_response_size" doc:"description=Maximum size (in bytes) of the allowable response size from a remote rule evaluation. Set to 0 to disable."`
```
kavirajk, line 148: Q: I'm failing to find default value for this config (`RulerRemoteEvaluationMaxResponseSize`). Is that intentional? (if not provided, default value will be 0 and meaning ruler remote evaluation will be disabled?)

Looks bit confusing and contradicting with the doc page on how to enable it?

```yaml
ruler:
  evaluation:
    mode: remote
    query_frontend:
      address: dns:///<query-frontend-service>:<grpc-port>
```

This config never mentions anything about setting `RulerRemoteEvaluationMaxResponseSize` when enabling it? Am i missing anything here?

docs/sources/operations/scalability.md
dannykopping: You're right, I could've worded it better; I want to include the design doc so folks can read up on the design considerations if they want to.

How's this?

```suggestion
This feature was initial proposed in [`LID-0002`](https://github.com/grafana/loki/pull/8129).
```

docs/sources/operations/scalability.md
dannykopping: OK 👍

docs/sources/operations/scalability.md
dannykopping: Loki only supports one configuration file; the other files are for overrides, and it would be clear from context that this belongs in the config file. I can add a link to the config docs page, if that'll make it more clear?

docs/sources/operations/scalability.md
dannykopping: Agree with the change except `s/may/will` because the ruler is now doing significantly less work. I'll update it 👍

docs/sources/operations/scalability.md
dannykopping: Good call 👍

pkg/ruler/evaluator_remote.go
```diff
@@ -161,36 +227,59 @@ func (q *remoteQuerier) query(ctx context.Context, query string, ts time.Time, l
...
+func (r *RemoteEvaluator) decodeResponse(ctx context.Context, resp *httpgrpc.HTTPResponse, orgID string) (*logqlmodel.Result, error) {
 	var decoded loghttp.QueryResponse
 	if err := json.NewDecoder(bytes.NewReader(resp.Body)).Decode(&decoded); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unexpected body encoding, not valid JSON: %w", err)
 	}
-	if decoded.Status == "error" {
-		return nil, fmt.Errorf("query response error: %s", decoded.Status)
+	if decoded.Status != loghttp.QueryStatusSuccess {
+		return nil, fmt.Errorf("query response error - status %q: %s", decoded.Status, resp.Body)
```
chaudum, line 282: resp.Body will potentially yield a very long unreadable byte slice. Maybe convert it to `string` first?

pkg/ruler/evaluator_remote.go
```diff
@@ -161,36 +227,59 @@ func (q *remoteQuerier) query(ctx context.Context, query string, ts time.Time, l
...
+
+		level.Warn(log).Log("msg", "failed to evaluate rule", "err", err)
+		return nil, fmt.Errorf("rule evaluation failed: %w", err)
 	}
+
+	// TODO(dannyk): consider retrying if the rule has a very high interval, or the rule is very sensitive to missing samples
+	//   i.e. critical alerts or recording rules producing crucial RemoteEvaluatorMetrics series
 	if resp.Code/100 != 2 {
-		return nil, fmt.Errorf("unexpected response status code %d: %s", resp.Code, string(resp.Body))
+		r.metrics.failedEvals.WithLabelValues("upstream_error", orgID).Inc()
```
kavirajk, line 256: nit: Do you think better to put all these reasons (`error`, `upstream_error`, `max_size`, `timeout`) for failed evaluations in one place as package level variables? 

Also these reasons looks so generic? may be tweak a bit? particularly `error`.

kavirajk (approved):
LGTM! few minor nits;

chaudum (approved):
LGTM

CHANGELOG.md
dannykopping: That's fair 👍 added it reflexively

pkg/ruler/evaluator_remote.go
dannykopping: > nit: Do you think better to put all these reasons (error, upstream_error, max_size, timeout) for failed evaluations in one place as package level variables?

For what purpose? And I assume you mean constants?

> Also these reasons looks so generic? may be tweak a bit? particularly error.

Correct, we can't have too much cardinality in these labels and also there could be a very large number of possible failure cases. The logs are extensive, so the operator would use the logs to figure out what the actual issue was.

pkg/ruler/evaluator_remote.go
dannykopping: It's being interpolated as a string in the `fmt.Errorf`.

pkg/validation/limits.go
dannykopping: It's an `int64` so its default / zero-value will be `0`.
The wording is misleading here; `Set to 0 to disable` means to disable the limit.

The config is not mentioned in that section above because it's a limit.

pkg/ruler/evaluator_remote.go
kavirajk: Good point. Also I believe `resp.Body` will be of type `io.ReadCloser`, so first reading it with something like `ioutil.ReadAll` and converting to string before printing in logs is the correct way to go I think.

even better to use `io.LimitReader` with some upper limit (just in case for some reason body is huge and to avoid some crazy memory footprints).

pkg/ruler/evaluator_remote.go
kavirajk: > For what purpose? And I assume you mean constants?

Yep constants. sorry!.

dannykopping: I don't really see the point of having them as package-level constants since their use is localized to this file and there is no duplication.

docs/sources/operations/scalability.md
dannykopping: I don't think it matters; the markdown renderer will sort it out - but we could add a linter @jdbaldry?

dannykopping:
@JStickler I believe I've addressed all your comments, but since we're in different timezones I'm going to merge the PR; if you have any follow-up changes I'll address those in a separate PR as I want to get this code into testing ASAP. Thanks!

PR #8786, opened 2023-03-13
Loki: Add a `prepare-shutdown` endpoint

DylanGuedes:
Add a new `/ingester/prepare_shutdown` endpoint. This new endpoint configures the ingester to release resources on the next SIGTERM.

### DIFF
--- a/docs/sources/api/_index.md
+++ b/docs/sources/api/_index.md
@@ -627,8 +627,8 @@ In microservices mode, the `/flush` endpoint is exposed by the ingester.
 POST /ingester/prepare_shutdown
 ```
 
-`/ingester/prepare_shutdown` will prepare the ingester to release resources on the next SIGTERM,
-where releasing resources means flushing data and unregistering from the ring.
+`/ingester/prepare_shutdown` will prepare the ingester to release resources on the next SIGTERM signal,
+where releasing resources means flushing data and unregistering from the ingester ring.
 This endpoint supersedes any YAML configurations and isn't necessary if the ingester is already
 configured to unregister from the ring or to flush on shutdown.
 
--- a/docs/sources/api/_index.md
+++ b/docs/sources/api/_index.md
@@ -621,6 +621,17 @@ backing store. Mainly used for local testing.
 
 In microservices mode, the `/flush` endpoint is exposed by the ingester.
 
+### Tell ingester to release all resources on next SIGTERM
+
+```
+POST /ingester/prepare_shutdown
+```
+
+`/ingester/prepare_shutdown` will prepare the ingester to release resources on the next SIGTERM,
+where releasing resources means flushing data and unregistering from the ring.
+This endpoint supersede any YAML configurations and isn't necessary if the ingester is already
+configured to unregister from the ring or to flush on shutdown.
+
 ## Flush in-memory chunks and shut down
 
 ```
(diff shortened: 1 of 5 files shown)


docs/sources/api/_index.md
```diff
@@ -621,6 +621,17 @@ backing store. Mainly used for local testing.
 
 In microservices mode, the `/flush` endpoint is exposed by the ingester.
 
+### Tell ingester to release all resources on next SIGTERM
+
+```
+POST /ingester/prepare_shutdown
+```
+
+`/ingester/prepare_shutdown` will prepare the ingester to release resources on the next SIGTERM,
```
JStickler, line 630: I feel like we should define SIGTERM here.

docs/sources/api/_index.md
```diff
@@ -621,6 +621,17 @@ backing store. Mainly used for local testing.
...
+where releasing resources means flushing data and unregistering from the ring.
```
JStickler, line 631: "unregistering from the ring"  Which ring?  Our documentation says we have six.  If we mean the ingester ring, let's make that clear.

docs/sources/api/_index.md
```diff
@@ -621,6 +621,17 @@ backing store. Mainly used for local testing.
...
+This endpoint supersede any YAML configurations and isn't necessary if the ingester is already
```
JStickler, line 632:
```suggestion
This endpoint supersedes any YAML configurations and isn't necessary if the ingester is already
```

JStickler (commented):
[docs team] Couple of small tweaks requested.

docs/sources/api/_index.md
DylanGuedes: SIGTERM is a type of signal that a process can receive and is supported by all Unix-based operating systems. Do you think saying `SIGTERM signal` is enough here?

docs/sources/api/_index.md
DylanGuedes: Yeah, the ingester ring. I pushed a commit addressing this.

docs/sources/api/_index.md
JStickler: LOL, and I have just exposed my UNIX ignorance. But it's also possible that we might have customers in Windows-only environments who could use the information?  What do you think?

DylanGuedes: Sorry, with "could use the information" do you mean the definition of SIGTERM? If so: personally, I feel like this isn't the appropriate place for explaining a concept like SIGTERM, although saying "SIGTERM signal" indeed looked better than just saying "SIGTERM".

JStickler: OK, let's go with "SIGTERM signal" then.

MasslessParticle (approved):

pkg/loki/loki.go
```diff
@@ -372,6 +372,8 @@ type Loki struct {
 	deleteClientMetrics *deletion.DeleteRequestClientMetrics
 
 	HTTPAuthMiddleware middleware.Interface
+
+	ingesterRelease *atomic.Bool
```
MichelHollands, line 376: Sorry for the late review but is this variable set anywhere?

DylanGuedes: lol looks like something I was using in an older implementation but forgot to clean, thanks for catching it

DylanGuedes: yep I was using it in a former implementation

PR #8787, opened 2023-03-13
Promtail: Add a new target for the Azure Event Hub

andriikushch:

We want to allow receiving logs from the Azure Cloud. To solve this problem, we need to be able to configure Promtail to consume logs from the Azure Events Hub.

To achieve this goal, we need to add a new configuration option to the scrape configuration in Promtail and implement the component that, given this configuration, can connect to the Azure Event Hub, parse received data, and prepare it to be stored in Loki.

Fix [#8788](https://github.com/grafana/loki/issues/8788)

# **Special notes for your reviewer**:

With the current implementation, I would like to leverage Azure Events Hub's compatibility with a Kafka protocol and reuse code for a Kafka target.

### Why I decided to create a new target instead of just reusing the Kafka target:
1. Configuration for Kafka in this scenario requires fewer parameters. For example, version and auth should have some specific values.
2. Message handling is different for Azure Logs. In addition to what we are doing for the Kafka target, we want to fix JSON and split incoming messages into multiple log lines before propagating it further.

### From the implementation perspective, there are two goals:
- Extract interfaces that are not Kafka configuration specific from the Kafka target module. Splitting Kafka config parsing and target creation.

- To make the message handler/parser injectable or configurable: we would like to fix JSON and split a message into multiple messages.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


clients/pkg/promtail/targets/kafka/target.go
```diff
@@ -53,41 +58,18 @@ func NewTarget(
...
-				Line:      string(message.Value),
-				Timestamp: timestamp(t.useIncomingTimestamp, message.Timestamp),
-			},
-			Labels: out,
+		if t.messageHandler == nil {
+			messageHandler(message, t)
+			continue
 		}
-		t.session.MarkMessage(message, "")
+		t.messageHandler(message, t)
```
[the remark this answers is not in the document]
andriikushch: Good question 👍  This handler is injected in the `NewTarget`, but there is no validation that this is not nil. 

I agree. In this case, the same logic also applies to the rest of the internals (for example, `session`) of the `Target`. We do not check if they are nil or not. 

For now, I will remove `if`.

clients/pkg/promtail/targets/kafka/handler.go
```diff
@@ -0,0 +1,61 @@
...
+	// UseIncomingTimestamp tells to use a timestamp from the incoming message
+	UseIncomingTimestamp() bool
+	// Session returns sarama consumer group session
+	Session() sarama.ConsumerGroupSession
+	// Logger returns logger
+	Logger() log.Logger
+}
+
+// messageHandler is a default message handler for a kafka target
+func messageHandler(message *sarama.ConsumerMessage, t KafkaTarget) {
```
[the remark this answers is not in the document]
andriikushch: Good point 👍 

Initially, I wanted more control over message handling, which is why I decided to go this way.

Anyway, I tried to change the code and inject only the parser, it reduced the amount of the code, and I got rid of a few new interfaces and some copy-paste. Thanks, @kgeckhart for the idea 👍

clients/pkg/promtail/targets/azurelog/parser.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/Shopify/sarama"
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+type eventHubMessage struct {
+	Records []any `json:"records"`
+}
+
+func messageParser(logger log.Logger, message *sarama.ConsumerMessage) []string {
```
kgeckhart, line 15: WDYT about having the parser return the error vs needing the logger here? If we do keep it this way I think those debug logs further down need to be errors instead.

clients/pkg/promtail/targets/azurelog/targetmanager.go
```diff
@@ -0,0 +1,75 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func (tm *TargetManager) Stop() {
+	for _, t := range tm.targetSyncers {
+		if err := t.Stop(); err != nil {
+			level.Error(t.Logger()).Log("msg", "error stopping azurelog target", "err", err)
```
kgeckhart, line 56: Do we need to use the target's logger here? It doesn't look like there's anything special about it vs using `tm.logger`. This would also avoid the need for the `TargetSyncer` `Logger()` function all together.

clients/pkg/promtail/targets/kafka/target_syncer.go
```diff
@@ -262,17 +306,32 @@ func (ts *TargetSyncer) Stop() error {
...
 
+// ActiveTargets returns active targets from its consumer
+func (ts *TargetSyncer) ActiveTargets() []target.Target {
+	return ts.getActiveTargets()
+}
+
+// DroppedTargets returns dropped targets from its consumer
+func (ts *TargetSyncer) DroppedTargets() []target.Target {
+	return ts.getDroppedTargets()
+}
```
kgeckhart, line 317: Now that these exists, should we update the `kafka/targetmanager.go` to use these instead of the embedded consumer ones?

andriikushch: Nice catch 👍

clients/pkg/promtail/targets/azurelog/targetmanager.go
andriikushch: Good question @kgeckhart 👍. I think we do not. I did it the same way it was done in the Kafka target. Just to keep it consistent. Now I changed it in both palaces, as you suggested 😃

clients/pkg/promtail/targets/azurelog/parser.go
andriikushch: I like the idea 👍 My initial plan was that if there is an error during the parsing, there is not much we can do besides writing the log and ignoring this message in the queue. These errors may occur because, technically, customers can send any data to the event hub.

After your comment, I did think about it again. I like having fewer parameters in the function signature, especially considering there is no need. 

Also, we can introduce error handling: if the message can not be parsed, put the original payload into Loki as a single record. I added these changes to the PR. What do you think?

clients/pkg/promtail/targets/kafka/target.go
```diff
@@ -34,17 +38,21 @@ type Target struct {
...
 
 func NewTarget(
+	logger log.Logger,
 	session sarama.ConsumerGroupSession,
 	claim sarama.ConsumerGroupClaim,
 	discoveredLabels, lbs model.LabelSet,
 	relabelConfig []*relabel.Config,
 	client api.EntryHandler,
 	useIncomingTimestamp bool,
+	messageParser MessageParser,
```
thepalbi, line 52: for injecting a dependency, I think is always better to use an interface rather than a function type, just to cover cases in which one needs to use something other than a function. For example, you could do:
```go
type MessageParser interface {
  Parse(...)
}
```
and in the azure case just make the function type `func(message *sarama.ConsumerMessage) ([]string, error)` implement that interface

clients/pkg/promtail/targets/kafka/target.go
```diff
@@ -80,13 +89,28 @@ func (t *Target) run() {
...
-			Entry: logproto.Entry{
-				Line:      string(message.Value),
-				Timestamp: timestamp(t.useIncomingTimestamp, message.Timestamp),
-			},
-			Labels: out,
+
+		logLines := []string{string(message.Value)}
+
+		if t.messageParser != nil {
+			ll, err := t.messageParser(message)
```
thepalbi, line 96: rather than nil-checking, I think is best to always inject a message parser, but in the Kafka target case inject a "no-op" implementation, that just returns the sarama message value

clients/pkg/promtail/targets/kafka/target.go
```diff
@@ -80,13 +89,28 @@ func (t *Target) run() {
...
+			if err != nil {
+				level.Error(t.logger).Log("msg", "message parsing error", "err", err)
```
thepalbi, line 98: if there's an error parsing, I'd skip the current kafka message logging the error, that's the usual approach in promtail

clients/pkg/promtail/targets/kafka/handler.go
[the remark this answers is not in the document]
thepalbi: +1 on using an interface like `KafkaMessageParser` for injection purposes

clients/pkg/promtail/targets/azurelog/parser.go
```diff
@@ -0,0 +1,35 @@
+package azurelog
+
+import (
+	"bytes"
+	"encoding/json"
+	"github.com/Shopify/sarama"
+)
+
+type eventHubMessage struct {
+	Records []any `json:"records"`
```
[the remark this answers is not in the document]
andriikushch: Nice 👍

clients/pkg/promtail/targets/azurelog/parser.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func messageParser(message *sarama.ConsumerMessage) ([]string, error) {
```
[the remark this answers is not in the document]
andriikushch: Agree 👍

clients/pkg/promtail/targets/azurelog/parser.go
```diff
@@ -0,0 +1,35 @@
...
+	// https://learn.microsoft.com/en-us/answers/questions/1001797/invalid-json-logs-produced-for-function-apps?fbclid=IwAR3pK8Nj60GFBtKemqwfpiZyf3rerjowPH_j_qIuNrw_uLDesYvC4mTkfgs
+	body := bytes.ReplaceAll(message.Value, []byte(`'`), []byte(`"`))
+
+	data := &eventHubMessage{}
+	err := json.Unmarshal(body, data)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []string
```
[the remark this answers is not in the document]
andriikushch: Good catch 👍

clients/pkg/promtail/targets/kafka/target.go
andriikushch: Valid point 👍. Agree, it gives more flexibility. 

I changed it and introduced a noopParser for a Kafka target, which just returns the raw message payload as `[]string{string(message.Value)}`

clients/pkg/promtail/targets/kafka/target.go
andriikushch: Sounds good 👍 Changed.

clients/pkg/promtail/targets/kafka/target.go
andriikushch: 👍

clients/pkg/promtail/targets/kafka/handler.go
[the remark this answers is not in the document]
andriikushch: I added `MessageParser` to the Kafka module. So it will be`kafka.MessageParser` 😃 

```go
// MessageParser defines parsing for each incoming message
type MessageParser interface {
	Parse(message *sarama.ConsumerMessage) ([]string, error)
}
```

clients/pkg/promtail/targets/kafka/target.go
```diff
@@ -90,24 +92,18 @@ func (t *Target) run() {
...
-		for _, logLine := range logLines {
-			t.client.Chan() <- api.Entry{
-				Entry: logproto.Entry{
-					Line:      logLine,
-					Timestamp: timestamp(t.useIncomingTimestamp, message.Timestamp),
-				},
-				Labels: out,
+		logLines, err := t.messageParser.Parse(message)
+		if err != nil {
+			level.Error(t.logger).Log("msg", "message parsing error", "err", err)
```
thepalbi, line 97: nit: can we do a `continue` after the log in the if branch, and do the `for _, logLine := range logLines { ...` outside the else statement? Might look more idiomatic

clients/pkg/promtail/targets/kafka/target_syncer.go
```diff
@@ -108,7 +108,7 @@ func NewSyncerFromScrapeConfig(
 		GroupID:              cfg.KafkaConfig.GroupID,
 	}
 
-	t, err := NewSyncer(context.Background(), reg, logger, pushClient, pipeline, group, client, nil, cfg.KafkaConfig.Topics, targetSyncConfig)
+	t, err := NewSyncer(context.Background(), reg, logger, pushClient, pipeline, group, client, noopParser{}, cfg.KafkaConfig.Topics, targetSyncConfig)
```
thepalbi: 🥇

clients/pkg/promtail/targets/kafka/target.go
andriikushch: You are right. I was thinking of doing it as well.

I did it differently because I think, in this case, we will have to duplicate also `t.session.MarkMessage(message, "")` statement before `continue`. Because if we can not process an event, we want to move to the next one.

This is why I decided to go with `else`.

But I don't have any strong opinion here 😃 

Would you prefer to go with `continue` and a `t.session.MarkMessage(message, "")`?

thepalbi: Ohh you are right, I forgot about the message ACK. Then it's good like this!

clients/pkg/promtail/targets/azurelog/parser.go
```diff
@@ -0,0 +1,36 @@
...
+func parseMessage(message *sarama.ConsumerMessage) ([]string, error) {
+	// fix json as mentioned here:
+	// https://learn.microsoft.com/en-us/answers/questions/1001797/invalid-json-logs-produced-for-function-apps?fbclid=IwAR3pK8Nj60GFBtKemqwfpiZyf3rerjowPH_j_qIuNrw_uLDesYvC4mTkfgs
+	body := bytes.ReplaceAll(message.Value, []byte(`'`), []byte(`"`))
+
+	data := &eventHubMessage{}
+	err := json.Unmarshal(body, data)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
andriikushch: You are right, it depends on our focus: logs or EventHub in general. 

Let's try to add a configuration parameter to control this behavior 👍

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
```diff
@@ -41,6 +41,7 @@ type Config struct {
 	PushConfig        *PushTargetConfig          `mapstructure:"loki_push_api,omitempty" yaml:"loki_push_api,omitempty"`
 	WindowsConfig     *WindowsEventsTargetConfig `mapstructure:"windows_events,omitempty" yaml:"windows_events,omitempty"`
 	KafkaConfig       *KafkaTargetConfig         `mapstructure:"kafka,omitempty" yaml:"kafka,omitempty"`
+	AzurelogConfig    *AzurelogTargetConfig      `mapstructure:"azurelog,omitempty" yaml:"azurelog,omitempty"`
```
kgeckhart, line 44: Based on, https://github.com/grafana/loki/pull/8787#discussion_r1138548299, I'm wondering if this should be AzurelogConfig or AzureEventHubConfig. We are building something which works with Azure Event Hub and isn't constrained to just azure resource logs from event hub.

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
```diff
@@ -246,6 +247,30 @@ type WindowsEventsTargetConfig struct {
...
+
+	// The list of brokers to connect to Event Hub (Required).
+	Brokers []string `yaml:"brokers"`
+
+	// The consumer group id.
+	GroupID string `yaml:"group_id"`
+
+	// Do not ignore messages that doesn't match schema for Azure resource logs
+	AllowCustomPayload bool `yaml:"allow_custom_payload"`
+}
```
kgeckhart, line 272: I wonder if at this level we should try to use as much Event Hub terminology as possible to avoid potential confusion and limit changes in the event we swap out the kafka consumer for the go azure client. I think this would only be two renames, `Topics` -> `EventHubs` and `Brokers`- > `ServiceBusEndpoint`? (I don't actually remember what this one is

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
andriikushch: You are right 👍 I will change it. Also, I will rename a package and the docs section

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
```diff
@@ -246,6 +247,30 @@ type WindowsEventsTargetConfig struct {
...
+	ConnectionString string `yaml:"connection_string"`
+
+	// The list of brokers to connect to Event Hub (Required).
+	Brokers []string `yaml:"brokers"`
+
+	// The consumer group id.
+	GroupID string `yaml:"group_id"`
+
+	// Do not ignore messages that doesn't match schema for Azure resource logs
+	AllowCustomPayload bool `yaml:"allow_custom_payload"`
```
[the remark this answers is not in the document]
andriikushch: Let's do it 😃

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
kgeckhart: 👍 that makes sense!

andriikushch: Done 👍

clients/pkg/promtail/targets/azurelog/parser.go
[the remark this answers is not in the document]
andriikushch: Done 👍

docs/sources/clients/promtail/configuration.md
```diff
@@ -1016,6 +1019,56 @@ When Promtail receives GCP logs, various internal labels are made available for
...
+# Event Hubs to consume (Required).
+[event_hubs: <strings> | default = [""]]
+
+# The consumer group id.
+[group_id: <string> | default = "promtail"]
+
+# Event Hub ConnectionString for authentication on Azure Cloud (Required).
+[connection_string: <string> | default = "range"]
+
+# If Promtail should pass on the timestamp from the incoming log or not.
```
thepalbi, line 1043:
```suggestion
# If Promtail should pass on the timestamp from the incoming message or not.
```

docs/sources/clients/promtail/scraping.md
```diff
@@ -345,6 +345,31 @@ For sending messages via UDP:
...
+    labels:
+      job: azure_event_hub
+  relabel_configs:
+    - action: replace
+      source_labels:
+        - __azure_event_hub_category
+      target_label: category
+```
+
+Only the `fully_qualified_namespace`, `event_hubs` and `connection_string` are required.
```
thepalbi, line 370:
```suggestion
Only the `fully_qualified_namespace`, `connection_string` and `event_hubs` are required.
```

clients/pkg/promtail/targets/kafka/parser.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"github.com/Shopify/sarama"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/model/relabel"
+
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+	"github.com/grafana/loki/pkg/logproto"
+)
+
+type noopMessageParser struct{}
```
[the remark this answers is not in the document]
andriikushch: 👍

kgeckhart (approved):
Great work with this, 🥳 ! Just one more question

clients/pkg/promtail/targets/azureeventhub/parser.go
```diff
@@ -0,0 +1,198 @@
...
+	if err == nil {
+		err = data.validate()
+	}
+
+	if err != nil {
+		if e.disallowCustomMessages {
+			return []api.Entry{}, err
+		}
+
+		return []api.Entry{e.entryWithCustomPayload(message.Value, labelSet, messageTime)}, nil
```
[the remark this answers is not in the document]
andriikushch: Good question 👍 

My understanding is that relabeling is already once happening in the Kafka target (for Kafka-specific labels). https://github.com/grafana/loki/pull/8787/files#diff-6788766e507b2afbb9f053e15f2e5fb5b5decd910af3bc32ef5e7fe33107a63eR74-R105

In this package, we are trying to apply it only for a `__azure_event_hub_category`, and the value for this label can be extracted only if the message is parsed correctly. 

The `entryWithCustomPayload` method should be invoked only if we recognize that the incoming message doesn't match the Azure Logs schema. Therefore we can not extract a value for this label, and we do not need to do relabeling in this case.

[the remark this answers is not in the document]
kgeckhart: Ah didn't think about that, that makes sense thank you!

vlad-diachenko (approved):
Looks awesome. 👍 just one nit, in case we do not have an agreement to include it into 2.8 release. I belive it would be great to include it.

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@
 * [6675](https://github.com/grafana/loki/pull/6675) **btaani**: Add logfmt expression parser for selective extraction of labels from logfmt formatted logs
 * [8474](https://github.com/grafana/loki/pull/8474) **farodin91**: Add support for short-lived S3 session tokens
 * [8774](https://github.com/grafana/loki/pull/8774) **slim-bean**: Add new logql template functions `bytes`, `duration`, `unixEpochMillis`, `unixEpochNanos`, `toDateInZone`, `b64Enc`, and `b64Dec`
+* [8474](https://github.com/grafana/loki/pull/8787) **andriikushch**: Promtail: Add a new target for the Azure Event Hubs
```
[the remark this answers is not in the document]
andriikushch: Good point 👍 Done.

docs/sources/clients/promtail/configuration.md
```diff
@@ -1016,6 +1019,56 @@ When Promtail receives GCP logs, various internal labels are made available for
...
+[Here](https://learn.microsoft.com/en-us/azure/active-directory/reports-monitoring/tutorial-azure-monitor-stream-logs-to-event-hub) is a tutorial on streaming Azure logs to an Azure event hub.
+
+Apache Kafka endpoint is not available within the `Basic` pricing plan. More information about Event Hubs pricing [here](https://azure.microsoft.com/en-us/pricing/details/event-hubs/). 
+
+```yaml
+# Event Hub namespace host names (Required). Typically, it looks like <your-namespace>.servicebus.windows.net:9093.
+[fully_qualified_namespace: <string> | default = ""]
+
+# Event Hubs to consume (Required).
+[event_hubs: <strings> | default = [""]]
```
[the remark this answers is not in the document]
andriikushch: Looks good 👍 Merged 😃

clients/pkg/promtail/targets/manager.go
```diff
@@ -241,6 +244,12 @@ func NewTargetManagers(
 				return nil, errors.Wrap(err, "failed to make kafka target manager")
 			}
 			targetManagers = append(targetManagers, kafkaTargetManager)
+		case AzureEventHubScrapeConfigs:
+			kafkaTargetManager, err := azureeventhub.NewTargetManager(reg, logger, client, scrapeConfigs)
```
tpaschalis, line 248: super minor nit: I would change this name to avoid any issues when scanning the code.

docs/sources/clients/promtail/configuration.md
```diff
@@ -1016,6 +1019,56 @@ When Promtail receives GCP logs, various internal labels are made available for
...
+
+The `azure_event_hub` block configures how Promtail receives Azure Event Hubs messages. Promtail uses an Apache Kafka endpoint on Event Hubs to receive messages. More information can be found [here](https://learn.microsoft.com/en-us/azure/event-hubs/azure-event-hubs-kafka-overview).
+
+[Here](https://learn.microsoft.com/en-us/azure/active-directory/reports-monitoring/tutorial-azure-monitor-stream-logs-to-event-hub) is a tutorial on streaming Azure logs to an Azure event hub.
+
+Apache Kafka endpoint is not available within the `Basic` pricing plan. More information about Event Hubs pricing [here](https://azure.microsoft.com/en-us/pricing/details/event-hubs/). 
+
+```yaml
+# Event Hub namespace host names (Required). Typically, it looks like <your-namespace>.servicebus.windows.net:9093.
+[fully_qualified_namespace: <string> | default = ""]
```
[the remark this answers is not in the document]
andriikushch: Makes sense 👍 Updated.

clients/pkg/promtail/targets/azureeventhub/parser.go
tpaschalis, line 75: I'm not sure I follow exactly what we're trying to achieve here along the `disallowCustomMessages` condition.

If `tryUnmarshal` returns an error, then do we ever want to return something? In this case, what is `message.Value`, won't it be a nil byte slice? If so, is that a useful signal to have?

clients/pkg/promtail/targets/kafka/parser.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/Shopify/sarama"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/model/relabel"
+
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+	"github.com/grafana/loki/pkg/logproto"
+)
+
+// noopMessageParser implements MessageParser. It doesn't modify the content of the original `message.Value`.
+type noopMessageParser struct{}
```
tpaschalis, line 13: I wasn't expecting this by its name; I mainly was expecting it to be some kind of test helper or something. Do we want to make it into `kafkaMessageParser` or something?

clients/pkg/promtail/targets/azureeventhub/parser.go
tpaschalis: Is the other case to populate the message in case it is _not_ a valid Azure Event Hub message but a random JSON object? Should we care about what's in there or allow anything to pass through?

tpaschalis (approved):
LGTM, nothing of note to add!

(disclaimer: I have not actually run this, review was just from reading the code).

clients/pkg/promtail/targets/azureeventhub/target_syncer.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return nil
+}
+
+func getConfig(connection string) *sarama.Config {
+	config := sarama.NewConfig()
+	config.Net.DialTimeout = 30 * time.Second
+	config.Net.SASL.Enable = true
+	config.Net.SASL.User = "$ConnectionString"
+	config.Net.SASL.Password = connection
```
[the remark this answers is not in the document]
andriikushch: Yes, you are right 😃 

Here is a doc https://learn.microsoft.com/en-us/azure/event-hubs/event-hubs-kafka-connect-debezium#configure-kafka-connect-for-event-hubs

clients/pkg/promtail/targets/azureeventhub/parser.go
andriikushch: That is an excellent question 👍 

I think the title of this PR was misleading, and I already renamed it.

This target was created for Event Hubs. Therefore we can not guarantee or expect that there will be only structured logs in the messages. 

Customers can send messages(with custom logs or something else) to the Event Hub into any topic, even if cloud services use it for logs. Therefore, there might be a mix of structured and unstructured data.

Allowing custom message format gives customers more flexibility. Having a `disallowCustomMessages` flag will enable them to filter unstructured data if they would like.

docs/sources/clients/promtail/scraping.md
```diff
@@ -345,6 +345,31 @@ For sending messages via UDP:
...
+      job: azure_event_hub
+  relabel_configs:
+    - action: replace
+      source_labels:
+        - __azure_event_hub_category
+      target_label: category
+```
+
+Only the `fully_qualified_namespace`, `connection_string` and `event_hubs` are required.
+Read the [configuration]({{< relref "configuration/#azure-event-hub" >}}) section for more information.
```
JStickler, line 371:
```suggestion
Read the [configuration]({{< relref "../../configuration/#azure-event-hub" >}}) section for more information.
```

docs/sources/clients/promtail/configuration.md
```diff
@@ -353,6 +353,9 @@ job_name: <string>
 # Configuration describing how to pull/receive Google Cloud Platform (GCP) logs.
 [gcplog: <gcplog_config>]
 
+# Configuration describing how to get Azure Event Hubs messages.
```
JStickler, line 356: Is it Azure Event Hub (singular) or Azure Event Hubs (plural)?

docs/sources/clients/promtail/configuration.md
```diff
@@ -1016,6 +1019,57 @@ When Promtail receives GCP logs, various internal labels are made available for
 - `__gcp_resource_type`
 - `__gcp_resource_labels_<NAME>`
 
+### Azure Event Hub
```
JStickler, line 1022: Is it Azure Event Hub (singular) or Azure Event Hubs (plural)?  I'm seeing both forms in the docs update when it should be consistent if it's a product name.

docs/sources/clients/promtail/configuration.md
```diff
@@ -1016,6 +1019,57 @@ When Promtail receives GCP logs, various internal labels are made available for
...
+
+The `azure_event_hub` block configures how Promtail receives Azure Event Hubs messages. Promtail uses an Apache Kafka endpoint on Event Hubs to receive messages. More information can be found [here](https://learn.microsoft.com/en-us/azure/event-hubs/azure-event-hubs-kafka-overview).
+
+[Here](https://learn.microsoft.com/en-us/azure/active-directory/reports-monitoring/tutorial-azure-monitor-stream-logs-to-event-hub) is a tutorial on streaming Azure logs to an Azure event hub.
```
JStickler, line 1026:
```suggestion
To learn more about streaming Azure logs to an Azure event hub, you can see this [tutorial](https://learn.microsoft.com/en-us/azure/active-directory/reports-monitoring/tutorial-azure-monitor-stream-logs-to-event-hub).
```

docs/sources/clients/promtail/configuration.md
```diff
@@ -1016,6 +1019,57 @@ When Promtail receives GCP logs, various internal labels are made available for
...
+
+Apache Kafka endpoint is not available within the `Basic` pricing plan. More information about Event Hubs pricing [here](https://azure.microsoft.com/en-us/pricing/details/event-hubs/).
```
JStickler, line 1028:
```suggestion
Note that an Apache Kafka endpoint is not available within the `Basic` pricing plan. For more information, see the [Event Hubs pricing page](https://azure.microsoft.com/en-us/pricing/details/event-hubs/). 
```

JStickler (commented):
[docs team]  Couple of questions and wording questions.

docs/sources/clients/promtail/configuration.md
andriikushch: @JStickler nice catch 👍 

I did check Azure [catalog](https://azure.microsoft.com/en-us/products/), and the service name is `Event Hubs` (plural).

I will update the code and docs to use plural everywhere.

Thank you for pointing it out 👍

docs/sources/clients/promtail/configuration.md
andriikushch: Great question 👍 

You are right. Its naming is inconsistent in this PR.

The service is called Event Hubs (plural), and I renamed it everywhere now to use "hubs" instead of "hub". 

I added a link to the Azure catalog in this comment https://github.com/grafana/loki/pull/8787#discussion_r1145895076

docs/sources/clients/promtail/configuration.md
```diff
@@ -1016,6 +1019,57 @@ When Promtail receives GCP logs, various internal labels are made available for
...
+
+### Available Labels
+
+When Promtail receives Azure Event Hubs messages, various internal labels are made available for [relabeling](#relabel_configs).
+
+**Available Labels:**
+
+- `__azure_event_hub_category`: The log category of the message when a message is an application log.
+
+The list of labels below are discovered because of consuming Kafka interface of Event Hubs:
```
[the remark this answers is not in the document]
andriikushch: Yes 😃 

We use Kafka integration to interact with Event Hubs. Because of this, all Kafka labels are available for this target too.

Your suggestion describes it well. I will update it.

Thank you! 👍

docs/sources/clients/promtail/scraping.md
andriikushch: Hi @JStickler,

I have a question about this suggestion 😃 

This PR adds a new target to Promtail, so I decided to keep documentation within Promtail docs. Similar to the other targets in this file.

If I understand correctly, "../../configuration" is a Loki configuration, not the Promtail one.

Could you please elaborate a little bit more on your suggestion?

JStickler: Ugh, I'm sorry, that's my mistake.  Ignore this suggestion.

andriikushch: @JStickler, because it was already merged, I created a PR to revert it: https://github.com/grafana/loki/pull/8904. Please take a look when you have time.

andriikushch:
Hi @MasslessParticle 😃  

Thank you very much for merging this PR. 

There was a discussion in this PR https://github.com/grafana/loki/pull/8787#discussion_r1145298591, where we discussed a link in the documentation section and decided to change it.

Could you please take a look at https://github.com/grafana/loki/pull/8904 when you have time? 

Thank you in advance!

PR #8793, opened 2023-03-14
Integration Tests: allow loki logs to be optionally output in integration tests

dannykopping:
Debugging integration tests can be difficult without Loki's logs being output.

Changing the call from `cluster.New()` to `cluster.New(level.WarnValue())` now allows the author the option to temporarily or permanently enable logging for their test, as well as configure a level filter to control the noise.


DylanGuedes (approved):
ahá, this is really wanted, gj.

PR #8799, opened 2023-03-14
fix to-string for noop label filters

MasslessParticle:
Label Filter optimizations omit `.*` label filters from the execution pipeline but intruduced a bug where those stages were also ommitted from calls to `String` on the expression. This results in unparsable queries and panics when queries are cloned via `query.String()` 

This PR makes `noopLabelFilter` embed matchers and inherit their `String` method just like the `StringLabelFilters` do


trevorwhitney (approved):
LGTM, should we backport as well?

JordanRushing (approved):

PR #8809, opened 2023-03-15
Karsten/hackathon 2023 03

jeschkies:


periklis:
@jeschkies This sound pretty like a super awesome addition. Can you provide a couple of example ouputs of the `explain` function?

jeschkies:
@periklis we'll need to get a working first but I'm sure we'll share the findings :slightly_smiling_face:

PR #8820, opened 2023-03-16
Loki: Add `/loki/api/v1/retention_simulation` endpoint

DylanGuedes:
Adds a new `/loki/api/v1/simulate_retention` endpoint
* Endpoint expects a `since` parameter that represents a period used on the simulation
* It works by simulating a chunk with `creationTime=now`. This simulated chunk is then passed to our internal retention validation that will consider `now=creationTime+since`.
* Ex: in a scenario where the user retention configured is `30d` and the endpoint is called with `since=15d`:
  * A chunk will be created with `creationTime=now`
  * The retention checker will consider `now=creationTime+since`=`now+15d`, forcing the comparison to consider an interval equal to `15d`
  * Since the retention is `30d` and the interval is `15d`, the endpoint will return that the chunk wouldn't be deleted


periklis:
Sounds more like `/loki/api/v1/delete?dryRun=true` WDYT?

DylanGuedes:
> Sounds more like `/loki/api/v1/delete?dryRun=true` WDYT?

agreed, it looks similar to the delete endpoint with dryRun. However, it might cause confusion: deletions != retention (although both end up in deleting data :smile: )

periklis:
> > Sounds more like `/loki/api/v1/delete?dryRun=true` WDYT?
> 
> agreed, it looks similar to the delete endpoint with dryRun. However, it might cause confusion: deletions != retention (although both end up in deleting data smile )

`/loki/api/v1/retention/explain`

DylanGuedes:
> > > Sounds more like `/loki/api/v1/delete?dryRun=true` WDYT?
> > 
> > 
> > agreed, it looks similar to the delete endpoint with dryRun. However, it might cause confusion: deletions != retention (although both end up in deleting data smile )
> 
> `/loki/api/v1/retention/explain` muscle

ohh I like that

PR #7860, opened 2022-12-05
Fix broken canary loki alerts

xinbinhuang:
**What this PR does / why we need it**:

The original added canary alert rules (https://github.com/grafana/loki/pull/7286) are malformatted and break the generated `loki-alerts` . i.e. 
```
// stripped away some labels
apiVersion: monitoring.coreos.com/v1
kind: PrometheusRule
metadata:
  labels:
    helm.sh/chart: loki-3.6.0
  name: release-name-loki-alerts
spec:
  groups: << nothing here
```

PTAL @dannykopping @trevorwhitney

**Which issue(s) this PR fixes**:
Fixes a bug introduced in #7286

**Special notes for your reviewer**:

**Checklist**
- [ ] Reviewed the `CONTRIBUTING.md` guide
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #8822, opened 2023-03-16
Add yaml linting step to helm ci

trevorwhitney:

Adds a yaml linting step for the included rules in the loki helm chart. Only way I could think to enforce the comma line ending was to require qutoes around string, otherwise the linter cannot tell if the comma is just part of the string.

Fixes issue raised by @dannykopping in https://github.com/grafana/loki/pull/7860#issuecomment-1338784070.

**Note**:
First commit has an intentionally wrong `alerts.yaml`. Will fix before merging after making sure the CI works.

### DIFF
--- a/.github/workflows/helm-ci.yml
+++ b/.github/workflows/helm-ci.yml
@@ -24,6 +24,9 @@ jobs:
               exit 1
           fi
 
+      - name: Lint Yaml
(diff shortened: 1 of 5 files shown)


dannykopping (approved):
Nice!

PR #8823, opened 2023-03-16
Allow math on errors

MasslessParticle:
Currently metric expressions will fail if they contain the parser error label `__error__`. There are cases where you may want to count errors. For example,  `sum(count_over_time({foo="bar"} | json | __error__= "JSONParserErr" [$__interval]))` to see how many of your logs are unparsable JSON.

This pr uses `ParseHint` to determine whether the `__error__` label was requested in the query and, if it was, do math on it like any other query. Parse hints are only included in VectorAggregations so the label used to signal whether or not to keep errors is never seen by a user.

All other operations remain unaffected

### DIFF
--- a/pkg/logql/evaluator.go
+++ b/pkg/logql/evaluator.go
@@ -499,8 +499,8 @@ func (r *rangeVectorEvaluator) Next() (bool, int64, promql.Vector) {
 	}
 	ts, vec := r.iter.At()
 	for _, s := range vec {
-		// Errors are not allowed in metrics.
-		if s.Metric.Has(logqlmodel.ErrorLabel) {
+		// Errors are not allowed in metrics unless they've been specifically requested.
+		if s.Metric.Has(logqlmodel.ErrorLabel) && !s.Metric.Has(logqlmodel.PreserveErrorLabel) {
 			r.err = logqlmodel.NewPipelineErr(s.Metric)
 			return false, 0, promql.Vector{}
 		}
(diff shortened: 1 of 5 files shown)


pkg/logql/log/parser_hints.go
```diff
@@ -25,11 +26,14 @@ type ParserHint interface {
 	//		 sum(rate({app="foo"} | json [5m]))
 	// We don't need to extract any labels from the log line.
 	NoLabels() bool
+	// CountError returns true when parsing errors were specifically requested
```
DylanGuedes, line 29: nit
```suggestion
	// PreserveError returns true when parsing errors were specifically requested.
```

pkg/logql/evaluator.go
```diff
@@ -499,8 +499,8 @@ func (r *rangeVectorEvaluator) Next() (bool, int64, promql.Vector) {
 	}
 	ts, vec := r.iter.At()
 	for _, s := range vec {
-		// Errors are not allowed in metrics.
-		if s.Metric.Has(logqlmodel.ErrorLabel) {
+		// Errors are not allowed in metrics unless they've been specifically requested.
+		if s.Metric.Has(logqlmodel.ErrorLabel) && !s.Metric.Has(logqlmodel.PreserveErrorLabel) {
```
DylanGuedes, line 503: hey but what if I have `__preserve_error__=false`?

pkg/logql/evaluator.go
```diff
@@ -531,8 +531,8 @@ func (r *absentRangeVectorEvaluator) Next() (bool, int64, promql.Vector) {
...
```
DylanGuedes, line 535: ditto

DylanGuedes (approved):
not sure if the `__preserve_error__=false` is a problem or not but other than that, lgtm!

PR #8836, opened 2023-03-20
Removing backend target from SSD docker setup

dannykopping:
The `backend` target is only present in v2.8.

N/A

I was using this docker setup to test https://github.com/grafana/loki/pull/8744 and I obviously left some things around from that. Sorry about that!

Thanks @wardbekker for noticing the issue

### DIFF
--- a/production/docker/config/loki.yaml
+++ b/production/docker/config/loki.yaml
@@ -53,14 +53,10 @@ ruler:
   enable_api: true
   wal:
     dir: /tmp/ruler-wal
-  evaluation:
-    mode: remote
-    query_frontend:
-      address: dns:///loki-read:9095
   storage:
     type: local
     local:
-      directory: /etc/loki/rules
+      directory: /loki/rules
   rule_path: /tmp/prom-rules
   remote_write:
     enabled: true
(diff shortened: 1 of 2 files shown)


slim-bean:
instead maybe we could update the images to reference k141 which is essentially our RC for 2.8?

MichelHollands (approved):
LGTM

vlad-diachenko (approved):
LGTM

dannykopping:
> shouldn't we only backport this for 2.7 then?

Not sure what you mean?
This isn't included in any release

dannykopping:
> instead maybe we could update the images to reference k141 which is essentially our RC for 2.8?

For now there's no real need to run the `backend` target in this setup, but we can certainly change the image later 👍

DylanGuedes:
> > shouldn't we only backport this for 2.7 then?
> 
> Not sure what you mean? This isn't included in any release

sorry, my message was misleading. I mean that instead of reverting on the main branch, we could just add this to the 2.7 branch, since the main branch supports the backend target.

dannykopping:
> > > shouldn't we only backport this for 2.7 then?
> > 
> > 
> > Not sure what you mean? This isn't included in any release
> 
> sorry, my message was misleading. I mean that instead of reverting on the main branch, we could just add this to the 2.7 branch, since the main branch supports the backend target.

No, because we're using the `2.7.x` image in this setup - so it'll break anyway
Once we cut `2.8` (or change the image to a weekly release) we could do that

PR #8847, opened 2023-03-21
Docs: Improve time range documentation

gabor:

Describes how exactly the `start` and `end` timestamps operate on the time range boundaries. 

**Checklist**
- [x] Documentation added

### DIFF
--- a/docs/sources/api/_index.md
+++ b/docs/sources/api/_index.md
@@ -270,7 +270,7 @@ accepts the following query parameters in the URL:
 - `query`: The [LogQL]({{< relref "../logql/" >}}) query to perform
 - `limit`: The max number of entries to return. It defaults to `100`. Only applies to query types which produce a stream(log lines) response.
 - `start`: The start time for the query as a nanosecond Unix epoch or another [supported format](#timestamp-formats). Defaults to one hour ago. Loki returns results with timestamp greater or equal to this value.
-- `end`: The end time for the query as a nanosecond Unix epoch or another [supported format](#timestamp-formats). Defaults to now. (This is exclusive, Loki returns results with timestamp lower than this value)
+- `end`: The end time for the query as a nanosecond Unix epoch or another [supported format](#timestamp-formats). Defaults to now. Loki returns results with timestamp lower than this value.
 - `since`: A `duration` used to calculate `start` relative to `end`. If `end` is in the future, `start` is calculated as this duration before now. Any value specified for `start` supersedes this parameter.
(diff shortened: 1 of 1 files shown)


docs/sources/api/_index.md
```diff
@@ -269,8 +269,8 @@ accepts the following query parameters in the URL:
 
 - `query`: The [LogQL]({{< relref "../logql/" >}}) query to perform
 - `limit`: The max number of entries to return. It defaults to `100`. Only applies to query types which produce a stream(log lines) response.
-- `start`: The start time for the query as a nanosecond Unix epoch or another [supported format](#timestamp-formats). Defaults to one hour ago.
-- `end`: The end time for the query as a nanosecond Unix epoch or another [supported format](#timestamp-formats). Defaults to now.
+- `start`: The start time for the query as a nanosecond Unix epoch or another [supported format](#timestamp-formats). Defaults to one hour ago. (This is inclusive, Loki returns results with timestamp greater or equal to this value)
```
JStickler, line 272: I'm a bit confused by the "inclusive" part here.  What is being included here?  I would just remove that bit.

```suggestion
- `start`: The start time for the query as a nanosecond Unix epoch or another [supported format](#timestamp-formats). Defaults to one hour ago. Loki returns results with timestamp greater or equal to this value.
```

docs/sources/api/_index.md
```diff
@@ -269,8 +269,8 @@ accepts the following query parameters in the URL:
...
+- `end`: The end time for the query as a nanosecond Unix epoch or another [supported format](#timestamp-formats). Defaults to now. (This is exclusive, Loki returns results with timestamp lower than this value)
```
JStickler, line 273: I'm a bit confused by the "exclusive" part here.  What is being excluded here?  I would just remove that bit.

```suggestion
- `end`: The end time for the query as a nanosecond Unix epoch or another [supported format](#timestamp-formats). Defaults to now. Loki returns results with timestamp lower than this value.
```

JStickler (commented):
[docs team]

docs/sources/api/_index.md
gabor: sure, your change makes sense

(to explain the inclusive/exclusive, it describes whether the `start` timestamp itself is included in the "valid" timestamps. in this case, it is included. but, when talking about the `end` timestamp, the `end` timestamp itself is not included in the "valid" timestamps, because we only take timestamps that are lower than `end`, so that one is exclusive.. but anyway, it's not needed, i'm totally fine with removing it)

docs/sources/api/_index.md
gabor: sure.

JStickler (approved):
[docs team]  LGTM  Glad my suggestions made sense.  I know what you were trying to say with the "inclusive" and "exclusive" but I had to stop and really think about it.  So thanks for accepting my suggestion.

PR #8848, opened 2023-03-21
Ruler: rule evaluation jitter

dannykopping:
This PR introduces a configurable random sleep before each rule evaluation to prevent contention on the query engine should multiple runs execute concurrently. It's quite likely that rules will be evaluated ~concurrently if they have the same interval.

N/A

In https://github.com/grafana/loki/pull/8744 I introduced the `ruler.Evaluator` interface, and this implementation is much simpler/elegant than what I had previously implemented in https://github.com/grafana/loki/pull/8307.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ##### Enhancements
 
-* [8848](https://github.com/grafana/loki/pull/8848) **dannykopping**: Ruler: rule evaluation jittter.
+* [8848](https://github.com/grafana/loki/pull/8848) **dannykopping**: Ruler: add configurable rule evaluation jitter.
 * [8752](https://github.com/grafana/loki/pull/8752) **chaudum**: Add query fairness control across actors within a tenant to scheduler, which can be enabled by passing the `X-Loki-Actor-Path` header to the HTTP request of the query.
 * [8786](https://github.com/grafana/loki/pull/8786) **DylanGuedes**: Ingester: add new /ingester/prepare_shutdown endpoint.
 * [8744](https://github.com/grafana/loki/pull/8744) **dannykopping**: Ruler: remote rule evaluation.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ##### Enhancements
 
+* [8848](https://github.com/grafana/loki/pull/8848) **dannykopping**: Ruler: rule evaluation jittter.
 * [8752](https://github.com/grafana/loki/pull/8752) **chaudum**: Add query fairness control across actors within a tenant to scheduler, which can be enabled by passing the `X-Loki-Actor-Path` header to the HTTP request of the query.
 * [8786](https://github.com/grafana/loki/pull/8786) **DylanGuedes**: Ingester: add new /ingester/prepare_shutdown endpoint.
 * [8744](https://github.com/grafana/loki/pull/8744) **dannykopping**: Ruler: remote rule evaluation.
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -817,9 +817,8 @@ storage:
 # CLI flag: -ruler.rule-path
 [rule_path: <string> | default = "/rules"]
 
-# Upper bound of random duration to wait before rule evaluation to
-# avoid contention during concurrent execution of rules. Set 0 to disable
-# (default).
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ##### Enhancements
 
+* [8848](https://github.com/grafana/loki/pull/8848) **dannykopping**: Ruler: rule evaluation jittter.
```
DylanGuedes, line 11: nit:
(I think it is a better reading for end users)
```suggestion
* [8848](https://github.com/grafana/loki/pull/8848) **dannykopping**: Add new ruler `evaluation_jitter` configuration.
```

docs/sources/configuration/_index.md
```diff
@@ -817,6 +817,12 @@ storage:
 # CLI flag: -ruler.rule-path
 [rule_path: <string> | default = "/rules"]
 
+# Upper bound of random duration to wait before rule evaluation; this is to
+# avoid contention during concurrent execution of rules. Set 0 to disable
+# (default).
```
DylanGuedes, line 822: nit
```suggestion
# Upper bound of random duration to wait before rule evaluation to
# avoid contention during concurrent execution of rules. Set 0 to disable
# (default).
```

pkg/ruler/base/ruler.go
```diff
@@ -161,6 +163,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.Var(&cfg.ExternalURL, "ruler.external.url", "URL of alerts return path.")
 	f.DurationVar(&cfg.EvaluationInterval, "ruler.evaluation-interval", 1*time.Minute, "How frequently to evaluate rules.")
 	f.DurationVar(&cfg.PollInterval, "ruler.poll-interval", 1*time.Minute, "How frequently to poll for rule changes.")
+	f.DurationVar(&cfg.EvaluationJitter, "ruler.evaluation-jitter", 0, "Upper bound of random duration to wait before rule evaluation; this is to avoid contention during concurrent execution of rules. Set 0 to disable (default).")
```
DylanGuedes, line 166:
```suggestion
	f.DurationVar(&cfg.EvaluationJitter, "ruler.evaluation-jitter", 0, "Upper bound of random duration to wait before rule evaluation to avoid contention during concurrent execution of rules. Set 0 to disable (default).")
```

docs/sources/configuration/_index.md
```diff
@@ -817,6 +817,12 @@ storage:
 # CLI flag: -ruler.rule-path
 [rule_path: <string> | default = "/rules"]
 
+# Upper bound of random duration to wait before rule evaluation; this is to
```
DylanGuedes, line 820: nit: might be useful to say that this is "experimental" (in case you think it is good to avoid using it for now for very sensitive environments)

pkg/ruler/evaluator_jitter.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (e *EvaluatorWithJitter) Eval(ctx context.Context, qs string, now time.Time) (*logqlmodel.Result, error) {
+	if e.inner == nil {
+		return nil, fmt.Errorf("no inner evaluator defined")
+	}
+
+	if e.maxJitter > 0 {
+		jitter := time.Duration(e.rng.Int63n(e.maxJitter.Nanoseconds()))
+		time.Sleep(jitter)
+	}
```
DylanGuedes, line 36: WDYT of moving those validations to the `NewEvaluatorWithJitter` instead? So you only check them a single time instead of checking for every evaluation

DylanGuedes (approved):
lgtm, just a few nits/suggestions

CHANGELOG.md
dannykopping: We try and keep the subsystem name at the beginning to make it easier to identify which changes someone might be interested in

docs/sources/configuration/_index.md
dannykopping: Nope, I don't think there's any risk at all.
The same instant is used when evaluating the rule, only we wait slightly before evaluating

pkg/ruler/evaluator_jitter.go
dannykopping: Good call 👍 
If `maxJitter <= 0` then I can just return the `inner`.

PR #8232, opened 2023-01-22
Loki: Fix version info issue that shows wrong version (#7669, #8055)

HwangTaehyun:
**What this PR does / why we need it**:

* It should fetch all tags before extracting tag info to ldflag of the build.

  * This issue could happen if a build tag like v2.7.1 was not fetched properly and then the build process was running. 
  * The following promtail is https://github.com/grafana/loki/releases/download/v2.7.1/promtail-darwin-arm64.zip

     image

  * If there's no tag on the exact commit to publish, then <kbd>BRANCH</kbd> variable becomes HEAD, so the <kbd>VERSION</kbd> variable becomes like "HEAD-e0af1cc".

    * https://github.com/grafana/loki/blob/86dcc8231bed247fc29a5f5e25c6bb184d08fa71/tools/image-tag#L7-L21
    * https://github.com/grafana/loki/blob/86dcc8231bed247fc29a5f5e25c6bb184d08fa71/Makefile#L35

**Which issue(s) this PR fixes**:
* Fixes #7669, #8055 

**Special notes for your reviewer**:
* This can happen when the version tag was not fetched properly before the build process.

**Checklist**
- [ ] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #8849, opened 2023-03-21
Build: Remove ./tools/fetch-tags invocation from Makefile

chaudum:

This invocation causes problems when make targets are executed in the build image container (`BUILD_IN_CONTAINER=true`), e.g. the `yacc` target, because git tries to fetch tags from `origin` using ssh, causing error message like these:

```
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
error: Could not fetch origin
```

This is because local ssh keys are not mounted into the build container.

In order not to break CI builds
(https://github.com/grafana/loki/pull/8232) we need to manually fetch the tags as the first step in release pipelines. This is added to the Drone configuration.

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -52,9 +52,11 @@ local gpg_private_key = secret('gpg_private_key', 'infra/data/ci/packages-publis
 local updater_config_template = secret('updater_config_template', 'secret/data/common/loki_ci_autodeploy', 'updater-config-template.json');
 local helm_chart_auto_update_config_template = secret('helm-chart-update-config-template', 'secret/data/common/loki-helm-chart-auto-update', 'on-loki-release-config.json');
 
+local build_image_name = 'grafana/loki-build-image:%s' % build_image_version;
+
 local run(name, commands, env={}) = {
   name: name,
-  image: 'grafana/loki-build-image:%s' % build_image_version,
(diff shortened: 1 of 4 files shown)


MichelHollands (approved):
LGTM

dannykopping (approved):
LGTM, thanks!

.drone/drone.jsonnet
```diff
@@ -653,8 +666,6 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
           RELEASE_TAG_REGEXP: '^([0-9]+\\.[0-9]+\\.[0-9]+)$',
         },
         commands: [
-          'apk add --no-cache bash git',
```
vlad-diachenko, line 656: i will cause an issue that ./tools/image-tag won't be executed becuse it uses bash

vlad-diachenko (changes requested):
we can not remove installation steps like : `apk add --no-cache bash git`

vlad-diachenko (approved):
Lgtm

chaudum:
@vlad-diachenko @MichelHollands @dannykopping [Opened a new PR](https://github.com/grafana/loki/pull/8854) because I don't want to open the can of worms that I did with this PR.

PR #8854, opened 2023-03-21
Remove ./tools/fetch-tags invocation from Makefile

chaudum:

This invocation causes problems when make targets are executed in the build image container (`BUILD_IN_CONTAINER=true`), e.g. the `yacc` target, because git tries to fetch tags from `origin` using ssh, causing error message like these:

```
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
error: Could not fetch origin
```

This is because local ssh keys are not mounted into the build container.

In order not to break CI builds
(https://github.com/grafana/loki/pull/8232) we need to manually fetch the tags as the first step in release pipelines. This is added to the Drone configuration.

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -1411,6 +1411,11 @@ services:
     path: /sys/fs/cgroup
 steps:
 - commands:
+  - apk add --no-cache bash git
+  - git fetch origin --tags
+  image: alpine
+  name: fetch-tags
+- commands:
   - printf "%s" "$NFPM_SIGNING_KEY" > $NFPM_SIGNING_KEY_FILE
   environment:
     NFPM_SIGNING_KEY:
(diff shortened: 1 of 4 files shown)


.drone/drone.yml
```diff
@@ -1410,6 +1410,11 @@ services:
   - name: cgroup
     path: /sys/fs/cgroup
 steps:
+- commands:
+  - apk add --no-cache bash git
+  - git fetch origin --tags
+  image: alpine
+  name: fetch-tags
```
chaudum, line 1417: Adding `git fetch` to this pipeline assures that we do not revert the change that introduced `git fetch` in the Makefile [here](https://github.com/grafana/loki/pull/8232).

PR #8874, opened 2023-03-22
promtail: support configurable polling interval

rfratto:
Currently, Promtail always polls files every 250ms. This PR allows the polling period to be customized, starting at a minimum polling period and exponentially increasing the polling frequency to a bounded maximum if the file has no changes.

When file changes are detected, the polling frequency resets to the minimum.

The default behavior is the existing behavior; both the minimum and maximum polling intervals are set to 250ms.

I'm opening this in draft to get early feedback. There's a few open questions:

* Where can I add validations to ensure that the min/max polling intervals are configured properly? 
* Is this the right approach for implementation? Do I need to make any changes?
* Does this need tests? 

I'll update the CHANGELOG and Documentation after we're agreed on an approach.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 * [8918](https://github.com/grafana/loki/pull/8918) **salvacorts** Introduce limit to require at least a number label matchers on metric and log queries.
 * [8909](https://github.com/grafana/loki/pull/8909) **salvacorts** Requests to `/loki/api/v1/index/stats` are split in 24h intervals.
 * [8732](https://github.com/grafana/loki/pull/8732) **abaguas**: azure: respect retry config before cancelling the context
+* [8874](https://github.com/grafana/loki/pull/8874) **rfratto**: Support expoential backoff when polling unchanged files for logs.
 
 ##### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@
 * [8918](https://github.com/grafana/loki/pull/8918) **salvacorts** Introduce limit to require at least a number label matchers on metric and log queries.
 * [8909](https://github.com/grafana/loki/pull/8909) **salvacorts** Requests to `/loki/api/v1/index/stats` are split in 24h intervals.
 * [8732](https://github.com/grafana/loki/pull/8732) **abaguas**: azure: respect retry config before cancelling the context
-* [8874](https://github.com/grafana/loki/pull/8874) **rfratto**: Support expoential backoff when polling unchanged files for logs.
+* [8874](https://github.com/grafana/loki/pull/8874) **rfratto**: Promtail: Support expoential backoff when polling unchanged files for logs.
 
 ##### Fixes
 
--- a/clients/pkg/promtail/targets/file/filetarget.go
+++ b/clients/pkg/promtail/targets/file/filetarget.go
@@ -28,7 +28,6 @@ const (
 type Config struct {
 	SyncPeriod time.Duration `mapstructure:"sync_period" yaml:"sync_period"`
 	Stdin      bool          `mapstructure:"stdin" yaml:"stdin"`
-	Watch      WatchConfig   `mapstructure:"watch_config" yaml:"watch_config"`
 }
 
 // RegisterFlags with prefix registers flags where every name is prefixed by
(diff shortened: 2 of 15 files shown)


cstyan:
Overall this LGTM. The only comment I had was about the cli flags prefixes but then I scrolled down and noticed you are doing `prefix+"target.<name>"` which is what I was going to suggest.

To your questions:
> Where can I add validations to ensure that the min/max polling intervals are configured properly?

I'm not sure we need to, the tail library already is?

> Is this the right approach for implementation? Do I need to make any changes?

My only questions is whether we'd want to allow this to be configured per file or per file sd job, rather than just one config globally for all file tailers.

> Does this need tests?

I don't think so.

rfratto:
> My only questions is whether we'd want to allow this to be configured per file or per file sd job, rather than just one config globally for all file tailers.

I'm not sure how easy it would be to configure it per-file, but I can add it to `scrapeconfig.Config`. Do you want me to do that? 

(There doesn't seem to be a good way to add it to the scrape config in the context of just file reading, so setting it there would have to be a no-op if other target managers are used)

cstyan:
> I'm not sure how easy it would be to configure it per-file, but I can add it to scrapeconfig.Config. Do you want me to do that?

Are you sure, unless I'm misreading the config I think [static_configs](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#static_configs) and [file_sd](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#file_sd_config) could take a config field for the polling interval?

I think per job is good enough though, instead of per file.

rfratto:
> Are you sure, unless I'm misreading the config I think [static_configs](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#static_configs) and [file_sd](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#file_sd_config) could take a config field for the polling interval?

Ah, sorry, I missed those. Those come directly from Prometheus, but I'll see how easy it would be to extend them without an upstream PR. If it's proving to be too annoying I'll go for a per-job fix 👍

clients/pkg/promtail/targets/file/filetarget.go
```diff
@@ -26,20 +27,40 @@ const (
 type Config struct {
 	SyncPeriod time.Duration `mapstructure:"sync_period" yaml:"sync_period"`
 	Stdin      bool          `mapstructure:"stdin" yaml:"stdin"`
+	Watch      WatchConfig   `mapstructure:"watch_config" yaml:"watch_config"`
```
rfratto, line 30: Move this to a new `global` block, but call it `file_watch_config` to make it clearer that it only applies to file_sd.

rfratto (commented):
After talking with Callum, he had a few suggestions:

1. Add a new `global` block and put the global settings for `file_watch_config` inside of it. 
2. Wrap around `file.Config` from Prometheus SD and add a `watch_config` block inside of that which overrides the global setting.

clients/pkg/promtail/targets/file/filetarget.go
rfratto: Actually, this is way harder than it seems. file_sd-specific settings get lost by the time a file target is made. Either it would need to be set at the job level or pushed as a Grafana Agent Flow feature. 

I'd like to make this global-only for now and we can see if people ask for more granular control.

docs/sources/clients/promtail/configuration.md
```diff
@@ -1284,70 +1314,70 @@ All Cloudflare logs are in JSON. Here is an example:
...
+  "WAFFlags": "0",
+  "WAFMatchedVar": "",
+  "WAFProfile": "unknown",
+  "WAFRuleID": "",
+  "WAFRuleMessage": "",
+  "WorkerCPUTime": 0,
+  "WorkerStatus": "unknown",
+  "WorkerSubrequest": false,
+  "WorkerSubrequestCount": 0,
+  "ZoneID": 1234
```
cstyan, line 1380: not sure what happened with the json formatting here? other than that lgtm

rfratto: 🙃 That's my vim config trying to be helpful. Fixed.

knylander-grafana (approved):
Doc changes look good. Thank you for adding documentation.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@
 * [8918](https://github.com/grafana/loki/pull/8918) **salvacorts** Introduce limit to require at least a number label matchers on metric and log queries.
 * [8909](https://github.com/grafana/loki/pull/8909) **salvacorts** Requests to `/loki/api/v1/index/stats` are split in 24h intervals.
 * [8732](https://github.com/grafana/loki/pull/8732) **abaguas**: azure: respect retry config before cancelling the context
+* [8874](https://github.com/grafana/loki/pull/8874) **rfratto**: Support expoential backoff when polling unchanged files for logs.
```
cstyan, line 35: nit, but can we prepend this with `promtail`? this changelog is for the whole repo

rfratto: Are you sure? It's in the promtail header: 

image

rfratto: Yeah, good point. Should I still add the Promtail prefix?

cstyan: personally, I think so yes

rfratto: Done 👍

PR #8891, opened 2023-03-23
Extend scheduler queue metrics with enqueue/dequeue counters

chaudum:

Better o11y of the scheduler.

This change yields new metrics with potentially high cardinality on the scheduler.

A different approach for the `enqueueCount` would be to add the `level` label to the `queueLength` metrics, however, then we would need to know the level when we dequeue an item. 

**TODO**
- [x] Clean up inactive metrics

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/scheduler/queue/metrics.go
+++ b/pkg/scheduler/queue/metrics.go
@@ -6,22 +6,21 @@ import (
 )
 
 type Metrics struct {
-	QueueLength       *prometheus.GaugeVec   // Per tenant and reason.
-	DiscardedRequests *prometheus.CounterVec // Per tenant.
-	// unexported fields must only used by the RequestQueue which resides in the same package
-	enqueueCount *prometheus.CounterVec // Per tenant and level
-	dequeueCount *prometheus.CounterVec // Per tenant and querier
+	queueLength       *prometheus.GaugeVec   // Per tenant and reason.
+	discardedRequests *prometheus.CounterVec // Per tenant.
+	enqueueCount      *prometheus.CounterVec // Per tenant and level
+	dequeueCount      *prometheus.CounterVec // Per tenant and querier
 }
 
 func NewMetrics(subsystem string, registerer prometheus.Registerer) *Metrics {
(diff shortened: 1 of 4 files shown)


pkg/scheduler/queue/metrics.go
```diff
@@ -6,23 +6,43 @@ import (
...
+		dequeueCount: promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
+			Namespace: "cortex",
+			Subsystem: subsystem,
+			Name:      "dequeue_count",
+			Help:      "Total number of dequeued (sub-)queries.",
+		}, []string{"user", "querier"}),
 	}
 }
+
+func (m *Metrics) Cleanup(user string) {
```
ashwanthgoli, line 44: is the `enqueueCount` clean-up excluded for any specific reason?

pkg/scheduler/queue/metrics.go
```diff
@@ -6,23 +6,43 @@ import (
...
 			Help:      "Number of queries in the queue.",
 		}, []string{"user"}),
-		DiscardedRequests: promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
+		discardedRequests: promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
 			Namespace: "cortex",
 			Subsystem: subsystem,
 			Name:      "discarded_requests_total",
 			Help:      "Total number of query requests discarded.",
 		}, []string{"user"}),
+		enqueueCount: promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
```
ashwanthgoli, line 29: should we add the new metrics under loki namespace?

pkg/scheduler/queue/metrics.go
chaudum: No, that's a good point. Will add it to the cleanup as well.

ashwanthgoli (approved):
LGTM

PR #8894, opened 2023-03-24
Helm: Removed support for PodDisruptionBudget in policy/v1alpha1 and upgraded it to policy/v1

### LINKED ISSUE #8887: Helmchart uses removed policy/v1beta1 so PodDisruptionBudget (PDB) can not be applied on k8s 1.25+
**Describe the bug**
The Loki-Helmchart do not installs the PDB in a Kubernetes 1.25+ because the beta api is removed in Kubernetes 1.25. By a normal helm install this is not that big deal because the other resources will be applied and only the PDB is ignored. (even it is not perfect because you get errors.)
The main problem for us is the Helmchart in combination with ArgoCD. Because ArgoCD don't apply any resource and even if you sync the resources manually. The app will never go on status "healthy" because it can not apply the defined PDB. 

**To Reproduce**
1. Create a ArgoCD app which refs the helmchart
2. Apply the app
3. See all the red and orange lights in ArgoCD :( 

Instead you can also install the helmchart in a k8s 1.25+ and try to get the PDB (kubectl get PDB) 

**Expected behavior**
The helmchart should not use the removed policy/v1beta1 instead it should use policy/v1 so the PDB can applied and the app goes on the status healthy. 

**Environment:**
 - Kubernetes(OpenShift) "v1.25.4+18eadca
 - helm

image

3deep5me:
This PR removes the old policy/v1beta1 from pdb objects and removes with it the not needed helper function. This simplifies the helmchart and makes it more compatible with argocd and newer kubernetes versions.

Fixes #8887 

First PR 😅


MasslessParticle (approved):
This seems reasonable, thanks for contributing. 

We can't pull this in until you've accepted the CLA

3deep5me:
Thanks for approving the PR. I'm on the go will check the CLA in the next few days.

3deep5me:
Done :)

PR #8896, opened 2023-03-24
Implementing consistent rule evaluation jitter using 32-bit hash

dannykopping:
This PR replaces the previous random jitter with a consistent jitter. While both are random, having the random jitter be applied _consistently_ is essential for evaluating rules on a predictable cadence.

If a rule is supposed to evaluate every minute, whether it evaluates at (e.g.) `01:00` or `01:03.234` is irrelevant because the evaluation _instant_ is not adjusted, so it will produce the same result whether run at `01:00` or `01:03.234`. However, if 1000 rules are set to evaluate at `01:00`, this will create a resource contention issue.

N/A

This change is part of the wider [remote rule evaluation](https://github.com/grafana/loki/pull/8744) improvements we have been making lately.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/ruler/evaluator_jitter.go
+++ b/pkg/ruler/evaluator_jitter.go
@@ -2,21 +2,29 @@ package ruler
 
 import (
 	"context"
-	"math/rand"
+	"hash"
+	"math"
+	"sync"
 	"time"
 
 	"github.com/grafana/loki/pkg/logqlmodel"
 )
 
-// EvaluatorWithJitter wraps a given Evaluator. It applies a randomly-generated jitter (sleep) before each evaluation to
-// protect against thundering-herd scenarios where multiple rules are evaluated at the same time.
+// EvaluatorWithJitter wraps a given Evaluator. It applies a consistent jitter based on a rule's query string by hashing
+// the query string to produce a 32-bit unsigned integer. From this hash, we calculate a ratio between 0 and 1 and
+// multiply it by the configured max jitter. This ratio is used to delay evaluation by a consistent amount of random time.
+//
+// Consistent jitter is important because it allows rules to be evaluated on a regular, predictable cadence
+// while also ensuring that we spread evaluations across the configured jitter window to avoid resource contention scenarios.
 type EvaluatorWithJitter struct {
+	sync.Mutex
+
 	inner     Evaluator
 	maxJitter time.Duration
-	rng       *rand.Rand
+	hasher    hash.Hash32
 }
 
-func NewEvaluatorWithJitter(inner Evaluator, maxJitter time.Duration, rngSource rand.Source) Evaluator {
+func NewEvaluatorWithJitter(inner Evaluator, maxJitter time.Duration, hasher hash.Hash32) Evaluator {
 	if maxJitter <= 0 {
 		// jitter is disabled or invalid
 		return inner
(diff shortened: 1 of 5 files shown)


pkg/ruler/evaluator_jitter.go
```diff
@@ -2,21 +2,29 @@ package ruler
...
-// EvaluatorWithJitter wraps a given Evaluator. It applies a randomly-generated jitter (sleep) before each evaluation to
-// protect against thundering-herd scenarios where multiple rules are evaluated at the same time.
+// EvaluatorWithJitter wraps a given Evaluator. It applies a consistent jitter based on a rule's query string by hashing
+// the query string to produce a 32-bit unsigned integer. From this hash, we calculate a ratio between 0 and 1 and
+// multiply it by the configured max jitter. This ratio is used to delay evaluation by a consistent amount of random time.
+//
+// Consistent jitter is important because it allows rules to be evaluated on a regular, predictable cadence
+// while also ensuring that we spread evaluations across the configured jitter window to avoid resource contention scenarios.
 type EvaluatorWithJitter struct {
+	sync.Mutex
```
kavirajk, line 20: nit: Not a big fan of this pattern of embedding mutex directly into struct particularly here. Since `EvaluatorWithJitter` is a public struct, you will have a public api `.Lock()` and `.Unlock()` methods on this struct, Meaning anyone can do locking and unlocking outside this package.

From the code below, I understand you want this lock accesible only inside private function `caculateJitter`.

Prefer to have something like `mu sync.Mutex` explicitly.

pkg/ruler/evaluator_jitter.go
```diff
@@ -25,13 +33,28 @@ func NewEvaluatorWithJitter(inner Evaluator, maxJitter time.Duration, rngSource
...
+	e.Lock()
+	{
+		_, _ = e.hasher.Write([]byte(qs))
+		h = e.hasher.Sum32()
+		e.hasher.Reset()
+	}
+	e.Unlock()
+
+	ratio := float32(h) / math.MaxUint32
+	return time.Duration(ratio * float32(e.maxJitter.Nanoseconds()))
```
kavirajk, line 59: q: How do you make sure returned duration is within `e.maxJitter` here? I'm failing to understand. Am I missing something here?

kavirajk (commented):
LGTM. one clarification and one minor nit.

pkg/ruler/evaluator_jitter.go
dannykopping: The hasher (which implements [`hash.Hash32`](https://pkg.go.dev/hash#Hash32)) returns a `uint32`, so by dividing it by `math.MaxUint32` we get a ratio from 0 to 1. The ratio can not be >1 so the jitter will always be between `0` and `maxJitter`.

pkg/ruler/evaluator_jitter.go
```diff
@@ -25,13 +33,28 @@ func NewEvaluatorWithJitter(inner Evaluator, maxJitter time.Duration, rngSource
...
 	return e.inner.Eval(ctx, qs, now)
 }
+
+func (e *EvaluatorWithJitter) calculateJitter(qs string) time.Duration {
+	var h uint32
+
+	// rules can be evaluated concurrently, so we protect the hasher with a mutex
+	e.Lock()
+	{
+		_, _ = e.hasher.Write([]byte(qs))
```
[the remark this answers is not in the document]
dannykopping: Good call, address in 15d193c08c7803b0c0e97f7426c613b7fe5d026c

kavirajk (approved):

PR #8898, opened 2023-03-25
Promtail: Fix examples how to build it

darxriggs:
The build flags have to be provided before the package paths.

Otherwise the `build` command fails with this error:
> malformed import path "--tags=promtail_journal_enabled": leading dash

See `go help build`:
> usage: go build [-o output] [build flags] [packages]


dannykopping (approved):
LGTM, thanks @darxriggs

PR #8911, opened 2023-03-27
operator: Update LokiStack annotation on RulerConfig delete

aminesnow:
When `RulerConfig` is deleted, update the `LokiStack` annotations in order to update the Loki configurations.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -1,8 +1,10 @@
 ## Main
 
+
+- [8911](https://github.com/grafana/loki/pull/8911) **aminesnow**: Update LokiStack annotaion on RulerConfig delete
+
 ## 0.2.0 (2023-03-27)
 
-- [8911](https://github.com/grafana/loki/pull/8911) **aminesnow**: Update LokiStack annotation on RulerConfig delete
 - [8651](https://github.com/grafana/loki/pull/8651) **periklis**: Prepare Community Loki Operator release v0.2.0
 - [8881](https://github.com/grafana/loki/pull/8881) **periklis**: Provide community bundle for openshift community hub
 - [8863](https://github.com/grafana/loki/pull/8863) **periklis**: Break the API types out into their own module
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.2.0 (2023-03-27)
 
+- [8911](https://github.com/grafana/loki/pull/8911) **aminesnow**: Update LokiStack annotaion on RulerConfig delete
 - [8651](https://github.com/grafana/loki/pull/8651) **periklis**: Prepare Community Loki Operator release v0.2.0
 - [8881](https://github.com/grafana/loki/pull/8881) **periklis**: Provide community bundle for openshift community hub
 - [8863](https://github.com/grafana/loki/pull/8863) **periklis**: Break the API types out into their own module
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.2.0 (2023-03-27)
 
-- [8911](https://github.com/grafana/loki/pull/8911) **aminesnow**: Update LokiStack annotaion on RulerConfig delete
+- [8911](https://github.com/grafana/loki/pull/8911) **aminesnow**: Update LokiStack annotation on RulerConfig delete
 - [8651](https://github.com/grafana/loki/pull/8651) **periklis**: Prepare Community Loki Operator release v0.2.0
 - [8881](https://github.com/grafana/loki/pull/8881) **periklis**: Provide community bundle for openshift community hub
 - [8863](https://github.com/grafana/loki/pull/8863) **periklis**: Break the API types out into their own module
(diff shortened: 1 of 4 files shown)


operator/controllers/loki/internal/lokistack/ruler_config_discovery.go
```diff
@@ -19,23 +19,48 @@ const (
...
-		if apierrors.IsNotFound(err) {
-			// Do nothing
-			return nil
-		}
-
-		return kverrors.Wrap(err, "failed to get lokistack", "key", key)
+	ss, err := getLokiStack(ctx, k, key)
+	if err != nil {
+		return err
 	}
```
periklis, line 26: AFAIU if the returned lokistack is nil (i.e. because the api returns `IsNotFound`) the following code will still try to update the annotation. Shouldn't we add an early return when `ss == nil` to omit update to err?

operator/controllers/loki/internal/lokistack/update.go
```diff
@@ -41,3 +41,35 @@ func updateAnnotation(ctx context.Context, k k8s.Client, stack *lokiv1.LokiStack
...
+		return nil
+	case errors.IsConflict(err):
+		// break into retry logic below on conflict
+		break
+	case err != nil:
+		return err
+	}
+
+	objectKey := client.ObjectKeyFromObject(stack)
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
periklis, line 63: I don't quite understand on what conditions this code ever passes the switch case and executes `RetryOnConflict`. This seems never to be the case to me.

operator/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.2.0 (2023-03-27)
 
+- [8911](https://github.com/grafana/loki/pull/8911) **aminesnow**: Update LokiStack annotaion on RulerConfig delete
```
periklis, line 5:
```suggestion

## Main/Unreleased

- [8911](https://github.com/grafana/loki/pull/8911) **aminesnow**: Update LokiStack annotaion on RulerConfig delete

## 0.2.0 (2023-03-27)

```

periklis (commented):
Just for confirmation this PR removes the annotation only to re-trigger a LokiStack CR reconciliation but the actual config re-rendering on `RulerConfig` not found works here:
https://github.com/grafana/loki/blob/44f1d8d7f6b1fb9b7fac32e913bd6517d15bad3f/operator/internal/handlers/lokistack_create_or_update.go#L212-L215

aminesnow:
> Just for confirmation this PR removes the annotation only to re-trigger a LokiStack CR reconciliation but the actual config re-rendering on `RulerConfig` not found works here:
> 
> https://github.com/grafana/loki/blob/44f1d8d7f6b1fb9b7fac32e913bd6517d15bad3f/operator/internal/handlers/lokistack_create_or_update.go#L212-L215
> 
> Right?

Yes. RulerConfigController only updates the annotation. LokiController is where the config is built.

operator/controllers/loki/internal/lokistack/update.go
aminesnow: That case here is when there is a conflict (same as for `updateAnnotation`), specifically the case `errors.IsConflict(err)`.

periklis: That makes sense now that you are telling me... I looked at `break` and read `return` :facepalm:

operator/controllers/loki/rulerconfig_controller.go
```diff
@@ -32,6 +32,12 @@ func (r *RulerConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	key := client.ObjectKey{Name: req.Name, Namespace: req.Namespace}
 	if err := r.Get(ctx, key, &rc); err != nil {
 		if errors.IsNotFound(err) {
+			// RulerConfig not found, remove annotation from LokiStack.
+			err = lokistack.RemoveRulerConfigAnnotation(ctx, r.Client, req.Name, req.Namespace)
```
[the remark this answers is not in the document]
aminesnow: This would be an issue if we have `finalizers` defined for this object, but we don't. So in this case, the object is marked for deletion and is immediately deleted.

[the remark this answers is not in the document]
xperimental: I was under the impression that checking for `deletionTimestamp` also worked without finalizers and thus was the preferred way. But actually testing the current code, it seems to work properly. Sorry, seems I remembered wrong. :slightly_smiling_face:

operator/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.2.0 (2023-03-27)
 
+- [8911](https://github.com/grafana/loki/pull/8911) **aminesnow**: Update LokiStack annotation on RulerConfig delete
```
periklis, line 5:
```suggestion

- [8911](https://github.com/grafana/loki/pull/8911) **aminesnow**: Update LokiStack annotaion on RulerConfig delete

## 0.2.0 (2023-03-27)

```

periklis (approved):
One last suggestion for the changelog and we are good to go.

PR #8914, opened 2023-03-27
Automatically Reorder Pipeline Filters

MasslessParticle:
This PR moves Line Filters to be as early in a log pipeline as possible. 

For example:
`{app="foo"} | logfmt |="some stuff"` becomes `{app="foo"} |="some stuff" | logfmt`

Any LineFilter after a `LineFormat` stage will be moved to directly after the nearest `LineFormat` stage

benchmarks:
```
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/logql/syntax
                    │ reorder_old.txt │           reorder_new.txt           │
                    │     sec/op      │   sec/op     vs base                │
ReorderedPipeline-8      2104.5n ± 3%   173.9n ± 2%  -91.74% (p=0.000 n=10)

                    │ reorder_old.txt │          reorder_new.txt          │
                    │      B/op       │   B/op    vs base                 │
ReorderedPipeline-8        336.0 ± 0%   0.0 ± 0%  -100.00% (p=0.000 n=10)

                    │ reorder_old.txt │          reorder_new.txt           │
                    │    allocs/op    │ allocs/op  vs base                 │
ReorderedPipeline-8        16.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)
```

### DIFF
--- a/pkg/logql/syntax/ast.go
+++ b/pkg/logql/syntax/ast.go
@@ -79,7 +79,7 @@ func (m MultiStageExpr) Pipeline() (log.Pipeline, error) {
 
 func (m MultiStageExpr) stages() ([]log.Stage, error) {
 	c := make([]log.Stage, 0, len(m))
-	for _, e := range m {
+	for _, e := range m.reorderStages() {
 		p, err := e.Stage()
 		if err != nil {
 			return nil, logqlmodel.NewStageError(e.String(), err)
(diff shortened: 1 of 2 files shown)


pkg/logql/syntax/ast.go
```diff
@@ -92,6 +92,56 @@ func (m MultiStageExpr) stages() ([]log.Stage, error) {
...
+				result = append(result, combineFilters(filters))
+			}
+			result = append(result, rest...)
+
+			filters = filters[:0]
+			rest = rest[:0]
+		default:
+			rest = append(rest, f)
+		}
+	}
```
cstyan, line 121: I don't quite understand what we want to happen here. I do sort of because of the test, but it's not obvious to me from this why we'd want to put LineFilters at the front of each section of a query and not just all at the front in general. 

In the case of your test, it confirms my first guess that after certain parts of a logql expression we need to start the building of `filters` and `rest` again, but the reasoning is unclear.

Can we include a comment as to why the ordering needs to preserved to a certain extent? For example, why is the correct result
```
|= "foo" |= "next" |= "bar" |= "baz" | logfmt | line_format "{{.foo}}" |= "1" |= "2" |= "3" | logfmt
```

and not
```
|= "foo" |= "next" |= "bar" |= "baz" | logfmt |= "1" |= "2" |= "3" | line_format "{{.foo}}" |logfmt
```

MasslessParticle: a `LineFormat` stage changes the line so any line filters that happen after it have to look at the new line instead.

example:
with the log line `foo=bar msg=something` and the logql expression `{} | line_format "new line" |= "new"`, we expect `|= "new"` to match but it won't if we put it before a line filter

Happy to add a comment

cstyan: Thanks, that makes sense. My lack of `logql` knowledge is showing.

I think a comment
```
// A line format expr will modify the contents of the line, so any line filter expressions that were originally after the line format expr must still be after the line format expr.
```
will help future readers understand.

MasslessParticle: done 🙂

sandeepsukhani (approved):
LGTM! Nice work!

JohanLindvall:
Hello, I just want to bring to your attention that this change makes our queries time out and/or consume a huge amount of CPU resources. See https://github.com/grafana/loki/issues/13869

PR #8927, opened 2023-03-28
Loki: Improve spans usage

DylanGuedes:
- At different places, inherit the span/spanlogger from the given context instead of instantiating a new one from scratch, which fix spans being orphaned on a read/write operation.
- At different places, turn spans into events. Events are lighter than spans and by having fewer spans in the trace, trace visualization will be cleaner without losing any details.
- Adds new spans/events to places that might be a bottleneck for our writes/reads.

### DIFF
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -458,6 +458,10 @@ rate_store:
   # updating rates
   # CLI flag: -distributor.rate-store.ingester-request-timeout
   [ingester_request_timeout: <duration> | default = 500ms]
+
+  # If enabled, detailed logs and spans will be emitted.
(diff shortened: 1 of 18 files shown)


MasslessParticle (approved):

MasslessParticle:
Didn't see that this was a draft. It looks totally reasonable, though

DylanGuedes:
> Didn't see that this was a draft. It looks totally reasonable, though

Np, thanks for the review. I tested in on our infra today and it definitely helped the orphaned spans.

DylanGuedes:
On this same PR I'm mixing fixes with new things. If you'd like to see things separated in smaller PRs just let me know.

JStickler (approved):
[docs team] LGTM

PR #8944, opened 2023-03-29
Diff configs script

trevorwhitney:

I always forget this trick for diffing configs when preparing a new release so I added it to a script.


MichelHollands (approved):
Nice addition. I've confirmed it works on Macs as well. LGTM

PR #8752, opened 2023-03-08
Scheduler: Add query fairness control across multiple actors within a tenant

chaudum:
**What this PR does / why we need it**:

This PR wires up the scheduler with the hierarchical queues.
It is the last PR to implement https://github.com/grafana/loki/pull/8585.

When these changes are in place, the client performing query requests can control their QoS (query fairness) using the `X-Actor-Path` HTTP header. This header controls in which sub-queue of the tenant's scheduler queue the query request is enqueued. The place within the hierarchy where it is enqueued defines the probability with which the request gets dequeued.

A common use-case for this QoS control is giving each Grafana user within a tenant their fair share of query execution time.

Any documentation is still missing and will be provided by follow-up PRs.

**Special notes for your reviewer**:

```console
$ gotest -count=1 -v ./pkg/scheduler/queue/... -test.run=TestQueryFairness
=== RUN   TestQueryFairness
=== RUN   TestQueryFairness/use_hierarchical_queues_=_false
    dequeue_qos_test.go:109: duration actor a 2.007765568s
    dequeue_qos_test.go:109: duration actor b 2.209088331s
    dequeue_qos_test.go:112: total duration 2.209280772s
=== RUN   TestQueryFairness/use_hierarchical_queues_=_true
    dequeue_qos_test.go:109: duration actor b 605.283144ms
    dequeue_qos_test.go:109: duration actor a 2.270931324s
    dequeue_qos_test.go:112: total duration 2.271108551s
--- PASS: TestQueryFairness (4.48s)
    --- PASS: TestQueryFairness/use_hierarchical_queues_=_false (2.21s)
    --- PASS: TestQueryFairness/use_hierarchical_queues_=_true (2.27s)
PASS
ok  	github.com/grafana/loki/pkg/scheduler/queue	4.491s
```

```console
$ gotest -count=5 -v ./pkg/scheduler/queue/... -bench=Benchmark -test.run=^$ -benchtime=10000x -benchmem
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/scheduler/queue
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
BenchmarkGetNextRequest
BenchmarkGetNextRequest/without_sub-queues
BenchmarkGetNextRequest/without_sub-queues-8         	   10000	     29337 ns/op	    1600 B/op	     100 allocs/op
[...]

PR #8947, opened 2023-03-30
Enforce per tenant queue size

chaudum:

Prior to the changes in https://github.com/grafana/loki/pull/8752 the max queue size per tenant was enforced by the size of the buffered channel to which a request was enqueued.
However, since we have hierarchical queues, every subqueue has the same channel capacity as the root (tenant) queue. Therefore the total queue size per tenant needs to be tracked separately, so requests can be rejected when the max queue size is reached.

**Checklist**
- [n/a] Documentation added
- [x] Tests updated
- [n/a] `CHANGELOG.md` updated
- [n/a] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

### DIFF
--- a/pkg/scheduler/queue/queue.go
+++ b/pkg/scheduler/queue/queue.go
@@ -93,7 +93,9 @@ func (q *RequestQueue) Enqueue(tenant string, path []string, req Request, maxQue
 		return errors.New("no queue found")
 	}
 
-	// Optimistically increase queue counter for tenant.
+	// Optimistically increase queue counter for tenant instead of doing separate
+	// get and set operations, because _most_ of the time the increased value is
+	// smaller than the max queue length.
 	// We need to keep track of queue length separately because the size of the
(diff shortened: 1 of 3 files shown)


pkg/scheduler/queue/queue.go
```diff
@@ -93,6 +93,18 @@ func (q *RequestQueue) Enqueue(tenant string, path []string, req Request, maxQue
 		return errors.New("no queue found")
 	}
 
+	// Optimistically increase queue counter for tenant.
+	// We need to keep track of queue length separately because the size of the
+	// buffered channel is the same across all sub-queues which would allow
+	// enqueuing more items than there are allowed at tenant level.
```
salvacorts, line 99: Instead of doing an increase to decrease inside the if. Can't we have a `get` method for the int map and use it?
So we only call increase after the if statement.

salvacorts (approved):
LGTM. Just a nit

PR #8951, opened 2023-03-30
[opt] s3: add s3 client http timeout for GetObject or PutObject

liguozhong:
add s3 client http timeout for GetObject or PutObject,

why？
The flush queue has been increasing. This is a difficult problem. We have been investigating this problem for half a year. When the querier has a large query, the load on the s3 server is too high and the number of connections is full. It will cause the flush queue of the ingester to increase continuously and cannot be reduced. When this problem occurs, after we restart the s3 server, the flush queue will quickly become 0.

We checked with our s3 (object storage compatible with the s3 protocol) team and found that it was because we used the s3 v1 sdk. The v1 sdk did not configure timeout by default, and it would wait indefinitely without configuration. The s3 v2 sdk defaults to a timeout of 30s, which only causes write failures and does not cause loki to get stuck.

Fixes #<issue number>
At present, as long as there are large logql queries such as 30 days, I must restart the s3 server, otherwise our ingester queue will continue to increase. We hope that we can break the stuck go routine after the 5m timeout occurs.

ingester flush queue length up 🔝
image

big query, 30day logql or something happen
image

pprof goroutine list profile snapshot
image

**There are other alternatives**:
```
ingester:
  flush_op_timeout: 10m
```
This `flush_op_timeout` configuration is more about the timeout of loki, not the timeout of s3 client. We will also have various s3 operations such as delete and getObject. We should add an independent timeout for s3, just like the cassandra client has a separate The timeout period can be configured.

And the parameter `flush_op_timeout` in the early version of loki represents the timeout of writing `a batch of chunks `to s3, not a timeout of s3. Now it is indeed changed to write to s3 one by one, but it is not ruled out that in the future, it will be changed back to write a batch of chunks to s3 sequentially. In this way, `flush_op_timeout: 5m` is far from enough to write a batch of chunks to s3.

```
  cassandra:
    timeout: 5m
```

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


JStickler (approved):
[docs team] LGTM

liguozhong:
This PR has been running in my environment for a month, and it can solve the system unavailability caused by big queries on the s3 server, such as the continuous accumulation of flush queues and other problems. If it is a self-built s3 server minio or..., it will be very helpful when s3 is not particularly stable

pkg/storage/chunk/client/aws/s3_storage_client.go
```diff
@@ -117,6 +118,7 @@ func (cfg *S3Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	cfg.SSEConfig.RegisterFlagsWithPrefix(prefix+"s3.sse.", f)
 
 	f.DurationVar(&cfg.HTTPConfig.IdleConnTimeout, prefix+"s3.http.idle-conn-timeout", 90*time.Second, "The maximum amount of time an idle connection will be held open.")
+	f.DurationVar(&cfg.HTTPConfig.Timeout, prefix+"s3.http.timeout", 5*time.Minute, "Timeout specifies a time limit for requests made by s3 Client.")
```
MichelHollands, line 121: This is a nice addition but we shouldn't set it to be on by default. Can you make the default zero?

liguozhong: thanks. 👌,done.

PR #8953, opened 2023-03-30
Querier: block query by hash

dannykopping:
Using the [query blocker](https://grafana.com/docs/loki/next/operations/blocking-queries/) can be unergonomic since queries can be long, require escaping, or hard to copy from logs. This change enables an operator to block queries by their hash.

N/A

We should probably provide a tool for creating uint32 representations of 32-bit FNV-1 hashes; all the online tools I found produce a hex value. It's not strictly necessary though since we have the `query_hash` in the logs, but it would be nice.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/operations/blocking-queries.md
+++ b/docs/sources/operations/blocking-queries.md
@@ -29,6 +29,10 @@ overrides:
       - pattern: '.*prod.*'
         regex: true
         types: filter,limited
+
+      # block any query that matches this query hash
+      - hash: 2943214005          # hash of {stream="stdout",pod="loki-canary-9w49x"}
+        types: filter,limited
 ```
 
 The available query types are:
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 #### Promtail
 
 ##### Enhancements
+* [8953](https://github.com/grafana/loki/pull/8953) **dannykopping**: Querier: block queries by hash.
```
chaudum, line 14: That should go into the `Loki` section, right?

dannykopping: Yes! Well spotted, thanks

dannykopping:
> The query hash changes when the order of the labels in the matcher changes, right? At some point we should hash the AST of the query so that logically ident queries have the same hash.

Yes, it's a hash of the query string - the contents of the string are not considered at all.

chaudum (approved):

owen-d (approved):
❤️ lgtm

JStickler (approved):
[docs team] LGTM.  

Just one question, how would a tenant know if their query had been blocked?  I'm guessing that not every user will have access to the logs, some users might run a query in the UI and have no idea why it did not complete?

dannykopping:
> [docs team] LGTM.
> 
> Just one question, how would a tenant know if their query had been blocked? I'm guessing that not every user will have access to the logs, some users might run a query in the UI and have no idea why it did not complete?

They will currently get a simple error stating "query blocked by policy". They'll then likely make contact with support to ask why this is happening. We have to be a bit circumspect sometimes because a bad actor might be trying to disrupt the service, and we don't want to give away too much - but on the other hand we want to be user-friendly for the majority of folks who will see this who will not be bad actors.

Definitely a topic to discuss further, let's touch base on it tomorrow?
I'll merge this so long, and we can improve the user experience if necessary in another PR.

JStickler:
>They will currently get a simple error stating "query blocked by policy".
OK, good, there's at least some indication why the query didn't run as expected.  That's mainly what I was looking for.

PR #8962, opened 2023-03-30
Implement "or" statement in line filters

MasslessParticle:
It might be nice to not force a user to use `|~` when they need to express `or` in LogQL. 

Example:
`{app="foo"} |= "foo" or "bar"`

Edit:
A ton of this is generated code or refactors. I've highlighed the parts most interesting to a reviewer


pkg/logql/syntax/ast.go
```diff
@@ -229,6 +230,17 @@ func newLineFilterExpr(ty labels.MatchType, op, match string) *LineFilterExpr {
 	}
 }
 
+func newOrLineFilter(left, right *LineFilterExpr) *LineFilterExpr {
```
MasslessParticle, line 233: Make a new OrLineFilter from left and right

pkg/logql/syntax/ast.go
```diff
@@ -302,10 +314,17 @@ func (e *LineFilterExpr) Filter() (log.Filterer, error) {
 			}
 			acc = append(acc, next)
 		default:
-			next, err := log.NewFilter(curr.Match, curr.Ty)
+			var next log.Filterer
```
MasslessParticle, line 317: Actually generate the filters

pkg/logql/syntax/ast.go
```diff
@@ -323,6 +342,23 @@ func (e *LineFilterExpr) Filter() (log.Filterer, error) {
 	return log.NewAndFilters(acc), nil
 }
 
+func newOrFilter(f *LineFilterExpr) (log.Filterer, error) {
```
MasslessParticle, line 345: Actually make a single or filter

pkg/logql/syntax/ast_test.go
```diff
@@ -358,6 +358,34 @@ func Test_FilterMatcher(t *testing.T) {
...
+		},
+		{
+			`{app="foo"} != "foo" or "bar"`,
+			[]*labels.Matcher{
+				mustNewMatcher(labels.MatchEqual, "app", "foo"),
+			},
+			[]linecheck{{"foo", false}, {"bar", false}, {"none", true}},
+		},
+		{
+			`{app="foo"} |~ "foo" or "bar"`,
```
dannykopping, line 376: Can you also please add a test to verify that the filter type is copied to the `or` branch? i.e. if it's a regex matcher on the left, it should be a regex matcher on the right.

dannykopping (commented):
Looks great! Added a suggestion about another test.
Also, it would be interesting to have a benchmark to compare performance of this approach vs a regex "or" condition.

Do we currently support "and" clauses in line filters, too (being a bit lazy here by asking instead of looking)? Do we support constructions like `{app="foo"} |~ "foo" or (".*bar.*" and "baz.*buzz.*")`?

pkg/logql/syntax/ast.go
```diff
@@ -283,6 +284,17 @@ func newLineFilterExpr(ty labels.MatchType, op, match string) *LineFilterExpr {
...
 }
 
+func newOrLineFilter(left, right *LineFilterExpr) *LineFilterExpr {
+	if left.Ty == labels.MatchEqual || left.Ty == labels.MatchRegexp {
+		left.Or = right
+		return left
+	}
+
+	// !(left or right) == (!left and !right).
+	right.Ty = left.Ty
```
jeschkies, line 294: What is `Ty`?

pkg/logql/syntax/ast.go
jeschkies: I vaguely remember that the ordering was reversed. At least for the `ContainsAllFilter`.

pkg/logql/syntax/ast.go
jeschkies: Why is this required? Shouldn't the parsing already take care or creating the right filter? I'm a little confused by this?

pkg/logql/syntax/ast_test.go
jeschkies: Do we want to support this case? Regular expressions already come with alternatives.

pkg/logql/syntax/expr.y
```diff
@@ -272,13 +274,21 @@ filterOp:
   IP { $$ = OpFilterIP }
   ;
 
+orFilter:
```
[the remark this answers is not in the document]
jeschkies: Hm, why is this required? Shouldn't it be enough to add it to line filter?

pkg/logql/syntax/ast_test.go
```diff
@@ -358,6 +358,34 @@ func Test_FilterMatcher(t *testing.T) {
...
 		},
+		{
+			`{app="foo"} |= "foo" or "bar"`,
+			[]*labels.Matcher{
+				mustNewMatcher(labels.MatchEqual, "app", "foo"),
+			},
+			[]linecheck{{"foo", true}, {"bar", true}, {"none", false}},
+		},
+		{
+			`{app="foo"} != "foo" or "bar"`,
```
jeschkies, line 369: Thanks. I didn't even think about this case.

jeschkies (commented):
Thanks for picking this up!

pkg/logql/syntax/ast.go
MasslessParticle: There may be a better way to do this. Afaict, the parser is responsible for creating a `*LineFilterExpr` but we don't get actual filters until we call `LineFilterExpr.Filter()` If the parser creates our type, it won't be composable with other `LineFilterExprs` because that's a concrete type rather than and interface.

pkg/logql/syntax/ast_test.go
MasslessParticle: This is a good call. Supporting regex is a side effect of the way this is implemented and this test ensures it's correct. This actually found a bug where it wasn't always

pkg/logql/syntax/expr.y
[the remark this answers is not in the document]
MasslessParticle: The grammar for filter is: `filter STRING` which includes the filter token (something like `|=`). If we made and `OrFilter`

```
lineFilter:
    filter STRING                                                   { $$ = newLineFilterExpr($1, "", $2) }
  | filter filterOp OPEN_PARENTHESIS STRING CLOSE_PARENTHESIS       { $$ = newLineFilterExpr($1, $2, $4) }
  | filter STRING OR lineFilter                               { $$ = }
  ;
```

The corresponding syntax would be `|= something or |= something` which is clunky but also impossible because of the `lineFilters` grammar that creates a nested expression any time there are line filters:
```
lineFilters:
    lineFilter                { $$ = $1 }
  | lineFilters lineFilter    { $$ = newNestedLineFilterExpr($1, $2) }
  ;
```

If the grammar were, instead:
```
lineFilter:
    filter STRING                                                   { $$ = newLineFilterExpr($1, "", $2) }
  | filter filterOp OPEN_PARENTHESIS STRING CLOSE_PARENTHESIS       { $$ = newLineFilterExpr($1, $2, $4) }
  | filter STRING OR STRING                               { $$ = }
  ;
```
We would only be able to have one `or` expression: `|= "foo" or "bar"` but `|= "foo" or "bar" or "baz"` would be invalid

To get an arbitrary number of `or` statements that don't also require the `filter` element, we have to define our own recursive type that can then be added to the `lineFilter` grammar.

pkg/logql/syntax/ast.go
MasslessParticle: The matcher type `MatchEqual`, `MatchNotEqual` etc.

MasslessParticle:
@dannykopping 

> Also, it would be interesting to have a benchmark to compare performance of this approach vs a regex "or" condition.

There is actually no difference in the underlying filters that are generated by this and the regex because [this code](https://github.com/grafana/loki/blob/0030cafb167fd70375399599acd8568c9290746e/pkg/logql/log/filter.go#L491-L509) parses a regex `Alternate` expression and does everything it can to turn it into a set of non-regex `or` filters

> Do we currently support "and" clauses in line filters, too (being a bit lazy here by asking instead of looking)? 

We don't. Instead a chain of `|=` is how `and` is expressed

> Do we support constructions like {app="foo"} |~ "foo" or (".*bar.*" and "baz.*buzz.*")?

We don't but it's an interesting idea. Maybe for a follow up?

dannykopping:
> @dannykopping
> 
> > Also, it would be interesting to have a benchmark to compare performance of this approach vs a regex "or" condition.
> 
> There is actually no difference in the underlying filters that are generated by this and the regex because [this code](https://github.com/grafana/loki/blob/0030cafb167fd70375399599acd8568c9290746e/pkg/logql/log/filter.go#L491-L509) parses a regex `Alternate` expression and does everything it can to turn it into a set of non-regex `or` filters

Gotcha 👍 thanks

> > Do we currently support "and" clauses in line filters, too (being a bit lazy here by asking instead of looking)?
> 
> We don't. Instead a chain of `|=` is how `and` is expressed

As a user, I would find that this violates the [_Principle of least astonishment_](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). I think it would be completely reasonable for a user to expect an `and` option if we supply an `or`, just like we do for label filters.

```yacc
labelFilter:
    ...
    | labelFilter AND labelFilter                    { $$ = log.NewAndLabelFilter($1, $3 ) }
    | labelFilter OR labelFilter                     { $$ = log.NewOrLabelFilter($1, $3 ) }
```

> > Do we support constructions like {app="foo"} |~ "foo" or ("._bar._" and "baz._buzz._")?
> 
> We don't but it's an interesting idea. Maybe for a follow up? That said, this is a ton simpler

Sure, I don't mean to creep the scope here - I was just asking if this had been considered and implemented already.

pkg/logql/syntax/ast_test.go
jeschkies: I agree with Danny. `|~ "foo|bar" or "\d+baz"` Eg would match a string containing "foo", "bar" or "42baz". This makes sense to me.

jeschkies (approved):

jeschkies (commented):
Could you add documentation as well?

PR #8968, opened 2023-03-30
Eager stream merge

owen-d:
This PR introduces a specialized heap based datastructure to merge incoming log results in the frontend. Recently we've experienced an increase in OOMs on frontends due to logs queries which match lots of data. Sharded requests in loki split based on the amount of data we expect and some queries see thousands of sub requests. For log queries, we'll fetch up the `limit` from each shard, return them to the frontend, and merge. High shard counts * limit log lines, especially combined with large log lines (in byte terms) are accumulated on the frontend. Once they all are received, the frontend merges them. This creates opportunity for OOMs as it can hold up a lot of memory.

This PR addresses one of these problems by eagerly accumulating responses as they're received and only retaining a total `limit` number of entries. There's still OOM potential due to race conditions between sub requests returning to the query-frontend and the query-frontend merging other sub requests, but this definitely improves the situation. I've been able to consistently run large limited queries that touch TBs of data (i.e. `{cluster=~".+"} |= "a"`) that previously OOMed frontends.


pkg/querier/queryrange/downstreamer.go
```diff
@@ -245,3 +254,372 @@ func ResponseToResult(resp queryrangebase.Response) (logqlmodel.Result, error) {
...
+		room = acc.limit - acc.count
+		// pop if there's no room to make the heap small enough for an append;
+		// in the short path of Push() we know that there's room for at least one entry
+		if room == 0 {
+			acc.Pop()
+		}
+
+		cpy := *s
+		cpy.Entries = []logproto.Entry{s.Entries[i]}
+		acc.Push(&cpy)
```
chaudum, line 487: Instead of pushing each entry individually, could we batch entries and push them at once?

owen-d: We need to compare them individually because we may only need 1, or 2, or n entries from the incoming stream and we can only see the "worst" value on the heap. That means we're only able to detect "is the best incoming value better than the current worst value on the heap". This is why each time that returns `true`, we `Pop()` the oldest worst value and `Push()` the new value and reconfigure the stream orders within the heap on both operations. There's probably some optimizations that could be made around tracking that in the future, but I don't think it's worth the current complexity.

pkg/querier/queryrange/downstreamer.go
```diff
@@ -245,3 +254,372 @@ func ResponseToResult(resp queryrangebase.Response) (logqlmodel.Result, error) {
...
+// a logsAccumulator or a bufferedAccumulator.
+// Which variant is detected on the first call to Accumulate.
+// Metric queries, which are generally small payloads, are buffered
+// since the memory overhead is negligible.
+// Log queries, sharded thousands of times and each returning <limit>
+// results, can be _considerably_ larger.	In this case, we eagerly
+// accumulate the results into a logsAccumulator, discarding values
+// over the limit to keep memory pressure down while other subqueries
+// are executing.
+type downstreamAccumulator struct {
```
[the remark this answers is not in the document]
owen-d: I dislike this since
* it's pretty common for accumulators to merge
* the non-logs version notably doesn't merge, it just appends to a slice

pkg/querier/queryrange/downstreamer.go
```diff
@@ -245,3 +254,372 @@ func ResponseToResult(resp queryrangebase.Response) (logqlmodel.Result, error) {
...
+	*accumulatedStreams
+	*bufferedAccumulator
```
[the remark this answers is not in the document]
owen-d: Good call, done

pkg/querier/queryrange/downstreamer.go
```diff
@@ -249,3 +258,374 @@ func ResponseToResult(resp queryrangebase.Response) (logqlmodel.Result, error) {
...
+
+	} else {
+		a.acc = &bufferedAccumulator{
+			results: make([]logqlmodel.Result, a.n),
+		}
+
+	}
+}
+
+func (a *downstreamAccumulator) Accumulate(ctx context.Context, index int, acc logqlmodel.Result) error {
```
jeschkies, line 307: Is this more of an append or add?

pkg/querier/queryrange/downstreamer.go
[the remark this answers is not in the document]
jeschkies: Hm, so we have to wrap the actual accumulators because we don't know what kind we need?

pkg/querier/queryrange/downstreamer.go
```diff
@@ -249,3 +258,374 @@ func ResponseToResult(resp queryrangebase.Response) (logqlmodel.Result, error) {
...
+// This allows us to easily find the 'worst' value
+// and replace it with a better one.
+// Once we've fully processed all log lines,
+// we return the heap in opposite order and then reverse it
+// to get the correct order.
+// Heap implements container/heap.Interface
+// solely to use heap.Interface as a library.
+// It is not intended for the heap pkg functions
+// to otherwise call this type.
+type accumulatedStreams struct {
```
jeschkies, line 352: How do you feel about using Bryan's Loser Tree implementation https://github.com/grafana/loki/blob/main/pkg/util/loser/tree.go?

pkg/querier/queryrange/downstreamer.go
owen-d: it's unknown if we'll actually accept any samples at this point. I think accumulate is the right word.

pkg/querier/queryrange/downstreamer.go
chaudum: Got it!

chaudum (approved):

PR #8971, opened 2023-03-31
Fix fetched chunk from store size in metric

dannykopping:
The `Size()` function on `chunk.Chunk` calls out to `func (ChunkRef) Size() int` in the proto, which doesn't actually include the size of the chunk data! We need to call `chunk.Data.Size()` for that.

N/A

This will fix the `Cache.Chunk.BytesSent` stat which indicates how many chunk bytes we sent to the chunk cache. This should probably be renamed for clarity at some point; it's a little confusing.

The same fix is applies to the `loki_chunk_fetcher_fetched_size_bytes{source="store"}` histogram metric.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated


MichelHollands (approved):
LGTM. Nice catch!

PR #8972, opened 2023-03-31
Cache index stats requests

salvacorts:

As described in https://github.com/grafana/loki/issues/8973, we are substantially increasing the load of index stat requests we sent to our index gateways. Many of these requests should be easily re-used by caching them.

This PR adds caching for index stat requests by reusing the results cache. Here's a demo ([source][1]):

image

Fixes https://github.com/grafana/loki/issues/8973

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

[1]: https://ops.grafana-ops.net/d/afcaef21-e5ad-49e7-ab06-42a9d7d915eb/index-stats?orgId=1&var-datasource=dev-cortex&var-cluster=dev-eu-west-2&var-namespace=loki-dev-009&var-loki_datasource=Grafana%20Logging&from=1680259907288&to=1680260431814&var-operation=All


pkg/querier/queryrange/queryrangebase/results_cache.go
```diff
@@ -580,7 +580,7 @@ func (s resultsCache) partition(req Request, extents []Extent) ([]Request, []Res
 }
 
 func (s resultsCache) filterRecentExtents(req Request, maxCacheFreshness time.Duration, extents []Extent) ([]Extent, error) {
-	maxCacheTime := (int64(model.Now().Add(-maxCacheFreshness)) / req.GetStep()) * req.GetStep()
```
salvacorts, line 583: IIUC, here the multiply cancels the division. Also, the step for an index stat req is 0 so this yields a divide by 0 panic.

pkg/querier/queryrange/index_stats_cache.go
```diff
@@ -0,0 +1,56 @@
...
+
+// GenerateCacheKey generates a cache key based on the userID, Request and interval.
+func (i IndexStatsSplitter) GenerateCacheKey(_ context.Context, userID string, r queryrangebase.Request) string {
+	return fmt.Sprintf("indexStats:%s:%s:%d:%d:%d", userID, r.GetQuery(), r.GetStep(), r.GetStart(), r.GetEnd())
+}
+
+type IndexStatsExtractor struct{}
+
+func (p IndexStatsExtractor) Extract(start, end int64, from queryrangebase.Response) queryrangebase.Response {
+	return from.(*IndexStatsResponse)
```
owen-d, line 22: Extract expects to be able to extract a subset of a response based on start & end. I'd suggest guessing what this should be by taking a proportion of the stats request based on the what % of the total request time the subset range covers.

pkg/querier/queryrange/queryrangebase/results_cache.go
owen-d: It's important to leave this for other types as this aligns the cache time with a multiple of the `step` via integer division. This keeps entries in the cache identical to the steps used in the query. This is important for metric queries. Instead, if the step is zero, can we consider it `1`?

owen-d (commented):
Looking good so far

pkg/querier/queryrange/querysharding.go
```diff
@@ -75,26 +76,35 @@ func newASTMapperware(
 	metrics *logql.MapperMetrics,
 	limits Limits,
 	maxShards int,
+	statsHandler ...queryrangebase.Handler,
```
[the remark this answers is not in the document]
chaudum: Could we not just pass `nil` as statsHandler argument in `NewQueryShardMiddleware()`?

CHANGELOG.md
```diff
@@ -86,6 +86,10 @@
 * [6675](https://github.com/grafana/loki/pull/6675) **btaani**: Add logfmt expression parser for selective extraction of labels from logfmt formatted logs
 * [8474](https://github.com/grafana/loki/pull/8474) **farodin91**: Add support for short-lived S3 session tokens
 * [8774](https://github.com/grafana/loki/pull/8774) **slim-bean**: Add new logql template functions `bytes`, `duration`, `unixEpochMillis`, `unixEpochNanos`, `toDateInZone`, `b64Enc`, and `b64Dec`
+* [8670](https://github.com/grafana/loki/pull/8670) **salvacorts** Introduce two new limits to refuse log and metric queries that would read too much data.
+* [8918](https://github.com/grafana/loki/pull/8918) **salvacorts** Introduce limit to require at least a number label matchers on metric and log queries.
+* [8909](https://github.com/grafana/loki/pull/8909) **salvacorts** Requests to `/loki/api/v1/index/stats` are split in 24h intervals.
+* [8972](https://github.com/grafana/loki/pull/8972) **salvacorts** Index stat requests are now cached in the results cache.
```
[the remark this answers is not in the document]
salvacorts: Done

pkg/querier/queryrange/querysharding.go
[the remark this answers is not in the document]
salvacorts: Done

pkg/querier/queryrange/querysharding.go
```diff
@@ -75,26 +76,35 @@ func newASTMapperware(
...
+		next:         next,
+		statsHandler: next,
+		ng:           logql.NewDownstreamEngine(logql.EngineOpts{LogExecutingQuery: false}, DownstreamHandler{next: next, limits: limits}, limits, logger),
+		metrics:      metrics,
+		maxShards:    maxShards,
 	}
+
+	if len(statsHandler) > 0 {
+		ast.statsHandler = statsHandler[0]
+	}
```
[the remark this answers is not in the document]
salvacorts: Done

pkg/querier/queryrange/index_stats_cache.go
salvacorts: Done https://github.com/grafana/loki/pull/8972/commits/fffbc332702263b70eb5bbbbbbc2827744d81cc1

pkg/querier/queryrange/index_stats_cache.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/grafana/loki/pkg/storage/chunk/cache"
+	"github.com/grafana/loki/pkg/util"
+)
+
+type IndexStatsSplitter struct {
+	Limits
+}
+
+// GenerateCacheKey generates a cache key based on the userID, Request and interval.
+func (i IndexStatsSplitter) GenerateCacheKey(_ context.Context, userID string, r queryrangebase.Request) string {
```
owen-d, line 21: Can we reuse the following, which has the same implementation?
```go
type cacheKeyLimits struct {
	Limits
	transformer UserIDTransformer
}
```

pkg/querier/queryrange/index_stats_cache.go
```diff
@@ -0,0 +1,79 @@
...
+	if denominator := int64(split / time.Millisecond); denominator > 0 {
+		currentInterval = r.GetStart() / denominator
+	}
+
+	return fmt.Sprintf("indexStats:%s:%s:%d", userID, r.GetQuery(), currentInterval)
+}
+
+type IndexStatsExtractor struct{}
+
+func (p IndexStatsExtractor) Extract(start, end int64, res queryrangebase.Response, resStart, resEnd int64) queryrangebase.Response {
```
owen-d, line 34:
```suggestion
// Extract favors the ability to cache over exactness of results. It assumes a constant distribution
// of log volumes over a range and will extract subsets proportionally.
func (p IndexStatsExtractor) Extract(start, end int64, res queryrangebase.Response, resStart, resEnd int64) queryrangebase.Response {
```

owen-d (approved):
Ok, this looks good. Left a few comments to address and there's a test failing, but I think we can merge once we address the final bits. I'll give an approval to unblock when you're ready.

pkg/querier/queryrange/index_stats_cache.go
salvacorts: Done. Similarly to [what the log results cache does (uses `log` prefix)][1], the idea was to add the `indexStats` prefix to the key to distinguish them from other queries (e.g. series queries, or log queries (w/o filter)),. Looks like the `NewSeriesTripperware` nor the `NewLimitedTripperware` configure any cache, so we should be safe there. Still, I think it would be a good idea to add the `indexStats` prefix just in case we decide to implement these caches in the future.

So instead of using the `cacheKeyLimits` directly, I'm  doing the following:
```go
type IndexStatsSplitter struct {
	cacheKeyLimits
}

// GenerateCacheKey generates a cache key based on the userID, Request and interval.
func (i IndexStatsSplitter) GenerateCacheKey(ctx context.Context, userID string, r queryrangebase.Request) string {
	cacheKey := i.cacheKeyLimits.GenerateCacheKey(ctx, userID, r)
	return fmt.Sprintf("indexStats:%s", cacheKey)
}
```

[1]: https://github.com/grafana/loki/blob/9159c1dac3ad1e80b9ebf0fe256dbe6dc0b8e903/pkg/querier/queryrange/log_result_cache.go#L125

PR #8975, opened 2023-03-31
tsdb: use unique uploader name while building tsdb index filename, same as boltdb-shipper

sandeepsukhani:
While building file names from `boltdb-shipper` and `tsdb`, we add the name of the uploader to it.
To avoid collisions in filenames when running multiple ingesters with the same name, we also add a nanosecond precise timestamp to the ingester name, which is persistent as the uploader name in the active index directory.

This is not being done for the tsdb index built from ingesters which can cause collisions in filenames when running multiple ingesters with the same name. I have moved the code to manage the uploader name to the index shipper config and used the same to share it with tsdb.

### DIFF
--- a/pkg/storage/stores/indexshipper/shipper.go
+++ b/pkg/storage/stores/indexshipper/shipper.go
@@ -99,6 +99,8 @@ func (cfg *Config) Validate() error {
 	return storage.ValidateSharedStoreKeyPrefix(cfg.SharedStoreKeyPrefix)
 }
 
+// GetUniqueUploaderName builds a unique uploader name using IngesterName + `-` + <nanosecond-timestamp>.
+// The name is persisted in the configured ActiveIndexDirectory and reused when already exists.
 func (cfg *Config) GetUniqueUploaderName() (string, error) {
 	uploader := fmt.Sprintf("%s-%d", cfg.IngesterName, time.Now().UnixNano())
 
--- a/pkg/storage/stores/indexshipper/shipper.go
+++ b/pkg/storage/stores/indexshipper/shipper.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"os"
+	"path"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


pkg/storage/stores/shipper/shipper_index_client.go
```diff
@@ -90,7 +86,7 @@ func (i *indexClient) init(storageClient client.ObjectClient, limits downloads.L
 	}
 
 	if i.cfg.Mode != indexshipper.ModeReadOnly {
-		uploader, err := i.getUploaderName()
+		uploader, err := i.cfg.GetUniqueUploaderName()
```
kavirajk, line 89: Q: So this name is applied when only handling `tsdb` files. And boltdb index files are not impacted correct? just double checking.

pkg/storage/stores/indexshipper/shipper.go
```diff
@@ -96,6 +99,33 @@ func (cfg *Config) Validate() error {
 	return storage.ValidateSharedStoreKeyPrefix(cfg.SharedStoreKeyPrefix)
 }
 
+func (cfg *Config) GetUniqueUploaderName() (string, error) {
```
kavirajk, line 102: nit: Can we small example how does the file name would look like as a comment description for this function?

kavirajk (approved):
LGTM. Thanks @sandeepsukhani . Jut one clarification and one minor nit :+1:

pkg/storage/stores/indexshipper/shipper.go
sandeepsukhani: Filenames are handled by the underlying index builders i.e `boltdb-shipper` and `tsdb`. I will add some comments at appropriate places.

pkg/storage/stores/shipper/shipper_index_client.go
sandeepsukhani: I have just copied the code to build the same uploader name from both `boltdb-shipper` and `tsdb`. There should not be any impact to `boltdb-shipper` file names.

PR #8977, opened 2023-03-31
Redefine 1x.extra-small to have 0 requirements.

alanconway:
Remove resource requirements so lokistack can run on laptop-sized SNO or CRC clusters.
Originally I thought to add an even smaller size, but I see that
1x.extra-small is already documented as not-for-production, so it seems appropriate.

Doc comment:

// SizeOneXExtraSmall defines the size of a single Loki deployment
// with NO resource requirements and without HA support.
//
// This is ONLY for development, testing, or demos
// on limited single-node clusters.
// There are NO performance guarantees.
// LokiStack will use whatever resources are available,
// and WILL NOT FUNCTION CORRECTLY if there is not enough memory or CPU.
//
// DO NOT USE THIS IN PRODUCTION!

### DIFF
--- a/operator/apis/loki/v1/lokistack_types.go
+++ b/operator/apis/loki/v1/lokistack_types.go
@@ -30,11 +30,15 @@ type LokiStackSizeType string
 
 const (
 	// SizeOneXExtraSmall defines the size of a single Loki deployment
-	// with extra small resources/limits requirements and without HA support.
-	// This size is ultimately dedicated for development and demo purposes.
-	// DO NOT USE THIS IN PRODUCTION!
+	// with NO resource requirements and without HA support.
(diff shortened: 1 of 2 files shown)


MasslessParticle (approved):

periklis:
@MasslessParticle @alanconway This shouldn't have been merged as such. This is an operator only related PR. Usually we prefix PR titles with `operator: ...`. However first time contributors might not know this convention

MasslessParticle:
@periklis Ack, sorry about that. Should I make a new PR reverting this?

periklis:
@MasslessParticle No worries we will reconcile this with @alanconway next week. We both work for the same team

PR #8987, opened 2023-04-01
Promtail: Fix file descriptor leak

darxriggs:
This reduces the amount of open file descriptors per followed journal file from 3 to 2. It also releases file descriptors for removed files (journal vacuuming).

The `defaultJournalEntryFunc` is only called once upon startup. It creates a `Journal` instance to only read a single entry to determine the age of the entry referenced in the positions file. Right after, this instance is not required anymore.

The actually used `Journal` instance used to follow the journal is created inside `journalTargetWithReader`.

### DIFF
--- a/clients/pkg/promtail/targets/journal/journaltarget.go
+++ b/clients/pkg/promtail/targets/journal/journaltarget.go
@@ -56,21 +56,25 @@ var defaultJournalReaderFunc = func(c sdjournal.JournalReaderConfig) (journalRea
 	return sdjournal.NewJournalReader(c)
 }
 
-var defaultJournalEntryFunc = func(c sdjournal.JournalReaderConfig, cursor string) (*sdjournal.JournalEntry, error) {
-	var (
-		journal *sdjournal.Journal
-		err     error
-	)
+var defaultJournalEntryFunc = func(c sdjournal.JournalReaderConfig, cursor string) (entry *sdjournal.JournalEntry, err error) {
+	var journal *sdjournal.Journal
 
 	if c.Path != "" {
 		journal, err = sdjournal.NewJournalFromDir(c.Path)
 	} else {
 		journal, err = sdjournal.NewJournal()
 	}
-
 	if err != nil {
 		return nil, err
-	} else if err := journal.SeekCursor(cursor); err != nil {
+	}
+	defer func() {
+		if errClose := journal.Close(); err == nil {
+			err = errClose
+		}
+	}()
+
+	err = journal.SeekCursor(cursor)
+	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


darxriggs:
What is the preferred way to update `CHANGELOG.md`?
In the initial commit the PR number is not yet known.

darxriggs:
It looks like that the tests are not stable. There has been a blinker [here](https://drone.grafana.net/grafana/loki/22118/3/14).

clients/pkg/promtail/targets/journal/journaltarget.go
```diff
@@ -67,10 +68,13 @@ var defaultJournalEntryFunc = func(c sdjournal.JournalReaderConfig, cursor strin
 	} else {
 		journal, err = sdjournal.NewJournal()
 	}
-
 	if err != nil {
 		return nil, err
-	} else if err := journal.SeekCursor(cursor); err != nil {
+	}
+	defer journal.Close()
```
chaudum, line 74: This would call `Close()` a second time, if it was already closed on line 95. Would that cause any problems?

chaudum:
> What is the preferred way to update `CHANGELOG.md`? In the initial commit the PR number is not yet known. Add another commit or change the initial commit after creating the PR?

People usually add a separate commit with the changelog entry after opening the PR. It's far from ideal, I know ...

clients/pkg/promtail/targets/journal/journaltarget.go
darxriggs: I had a look at the repository and found different resource closing patterns.

Therefore I have updated the pull request to an anonymous defer function to get rid of the double close.

clients/pkg/promtail/targets/journal/journaltarget.go
```diff
@@ -56,21 +56,25 @@ var defaultJournalReaderFunc = func(c sdjournal.JournalReaderConfig) (journalRea
...
 		journal, err = sdjournal.NewJournalFromDir(c.Path)
 	} else {
 		journal, err = sdjournal.NewJournal()
 	}
-
 	if err != nil {
 		return nil, err
-	} else if err := journal.SeekCursor(cursor); err != nil {
+	}
+	defer func() {
```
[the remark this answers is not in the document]
darxriggs: No. This function is only used once upon startup to read a single entry (see the PR description). Its name might be a bit misleading.

darxriggs:
I was wondering, after the review questions have been answered and the implementation has been adapted, who can have another look at this.

darxriggs:
@dannykopping You have already reviewed another of my PRs regarding promtail. Could you also take care of this one?

dannykopping:
> @dannykopping You have already reviewed another of my PRs regarding promtail. Could you also take care of this one?

Looks like @chaudum and @jeschkies already started on this one; I'll let them complete it unless they'd prefer I do it.

MasslessParticle (approved):
Thanks for the contribution!

Issue #8784, opened 2023-03-13
Promtail crashes in compress/gzip.(*Reader)


juissi-t:
**Describe the bug**

We are running promtail version 2.7.4 and the pods sometimes crash with this error:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x280 pc=0x61570e]

goroutine 3822006 [running]:
compress/gzip.(*Reader).Read(0x2c?, {0xc001d8f000?, 0x4?, 0x1?})
        /usr/local/go/src/compress/gzip/gunzip.go:247 +0x2e
bufio.(*Scanner).Scan(0xc0033d9f28)
        /usr/local/go/src/bufio/scan.go:214 +0x876
github.com/grafana/loki/clients/pkg/promtail/targets/file.(*decompressor).readLines(0xc000868e60)
        /src/loki/clients/pkg/promtail/targets/file/decompresser.go:198 +0x838
created by github.com/grafana/loki/clients/pkg/promtail/targets/file.newDecompressor
        /src/loki/clients/pkg/promtail/targets/file/decompresser.go:96 +0x58a
```

These two log lines seem to happen before the crash, with always different files mentioned:

```
ts=2023-03-10T00:44:37.924440049Z caller=log.go:168 level=info msg="Re-opening moved/deleted file /var/log/pods/monitoring_datadog-jz55r_15df45c3-a7ef-41db-a6a9-eaf7f2d4ead8/agent/0.log ..."
ts=2023-03-10T00:44:37.92448151Z caller=log.go:168 level=info msg="Successfully reopened /var/log/pods/monitoring_datadog-jz55r_15df45c3-a7ef-41db-a6a9-eaf7f2d4ead8/agent/0.log"
```

**To Reproduce**
1. Run promtail version 2.7.4 on Kubernetes
2. Ship logs to a Loki instance running in another cluster
3. Wait a few days, some promtails crash every now and then

**Expected behavior**
No crashes.

**Environment:**
 - Infrastructure: Kubernetes, EKS 1.22 with Bottlerocket instances
 - Deployment tool: Helm

**Screenshots, Promtail config, or terminal output**
Promtail config:

```yaml
server:
  log_level: info
  http_listen_port: 3101
  
clients:
  - external_labels:
      cluster: abc-dev
    url: https://loki-gateway.internal.abc.xyz/loki/api/v1/push

positions:
  filename: /run/promtail/positions.yaml

[...]

PR #8994, opened 2023-04-03
Promtail: Add `decompression` configuration

DylanGuedes:
Adds a new `compression` configuration to Promtail to customize decompression behavior.
As of now, two options are supported:
`enabled`: false by default but if truthy, Promtail will infer a compression algorithm to use based on the file extension
`initialDelay`: 0 by default but if higher, Decompressor will sleep for the given duration before trying to decompress. This is to avoid scenarios where the Decompressor starts to work on a file that is still being compressed/generated.

Fixes https://github.com/grafana/loki/issues/8784

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/clients/promtail/_index.md
+++ b/docs/sources/clients/promtail/_index.md
@@ -39,10 +39,32 @@ drop, and the final metadata to attach to the log line. Refer to the docs for
 
 ### Support for compressed files
 
-Promtail now has native support for ingesting compressed files by a mechanism that
-relies on file extensions. If a discovered file has an expected compression file
-extension, Promtail will **lazily** decompress the compressed file and push the
-parsed data to Loki. Important details are:
+Promtail now has native support for ingesting compressed files by a mechanism
+that relies on file extensions. If a discovered target has an expected
+compression file extension and the label `__infer_decompression__: true`,
+Promtail will **lazily** decompress the compressed file and push the parsed data
+to Loki. Example of YAML configuration to decompress files:
+
+```yaml
+server:
+  http_listen_port: 9080
+  grpc_listen_port: 0
+positions:
+  filename: /var/lib/promtail/positions.yaml
+clients:
+  - url: http://localhost:3100/loki/api/v1/push
+scrape_configs:
+- job_name: system
+  static_configs:
+  - targets:
+      - localhost
+    labels:
+      job: varlogs
+      __path__: /var/log/**.gz
+      __infer_decompression__: true
+```
+
+Important details are:
 * It relies on the `\n` character to separate the data into different log lines.
 * The max expected log line is 2MB bytes within the compressed file.
 * The data is decompressed in blocks of 4096 bytes. i.e: it first fetches a block of 4096 bytes
--- a/docs/sources/clients/promtail/_index.md
+++ b/docs/sources/clients/promtail/_index.md
@@ -41,9 +41,9 @@ drop, and the final metadata to attach to the log line. Refer to the docs for
 
 Promtail now has native support for ingesting compressed files by a mechanism
 that relies on file extensions. If a discovered target has an expected
-compression file extension and the label `__infer_decompression__: true`,
-Promtail will **lazily** decompress the compressed file and push the parsed data
-to Loki. Example of YAML configuration to decompress files:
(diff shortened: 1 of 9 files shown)


dannykopping (commented):
Nice idea!
Haven't gone through the changes, but I'm skeptical about the autodetection in this stage. Having autodetect is nice, but if it cannot detect the format from the filetype, or the file was given an incorrect filename, there is no mechanism to override it.

For this reason, I think we should value explicitness over implicitness, and just have the operator specify which format(s) will be attempted.

DylanGuedes:
> Nice idea! Haven't gone through the changes, but I'm skeptical about the autodetection in this stage. Having autodetect is nice, but if it cannot detect the format from the filetype, or the file was given an incorrect filename, there is no mechanism to override it.
> 
> For this reason, I think we should value explicitness over implicitness, and just have the operator specify which format(s) will be attempted.

That's fair but just one thing, this new configuration actually makes the infer opt-out.
Like, before this, Promtail would by default uncompress and infer the decompression algorithm etc; with this change, it will only do it if "decompression" is explicitly configured and enabled (it is disabled by default).
That said: having the algorithm/format configurable is probably a good idea, so let me add it :smile:

docs/sources/clients/promtail/_index.md
```diff
@@ -39,10 +39,34 @@ drop, and the final metadata to attach to the log line. Refer to the docs for
...
 
-Promtail now has native support for ingesting compressed files by a mechanism that
-relies on file extensions. If a discovered file has an expected compression file
-extension, Promtail will **lazily** decompress the compressed file and push the
-parsed data to Loki. Important details are:
+Promtail now has native support for ingesting compressed files by a mechanism
+that relies on file extensions. If a discovered target has an expected
+compression file extension and decompression is enabled, Promtail will
+**lazily** decompress the compressed file and push the parsed data to Loki.
+Example of YAML configuration to decompress files:
```
JStickler, line 46: Is this a Loki configuration or a Promtail configuration file?  There are multiple yaml files floating around, and it's helpful to be specific when we're talking about editing them.

JStickler (approved):
[docs team]  I'm going to request that we specify which YAML file we're talking about, but other than that LGTM.

docs/sources/clients/promtail/_index.md
DylanGuedes: Thank you, I pushed a commit addressing it. After your comment I just noticed I missed to add the new configuration to Promtail config docs, let me do it too :smile:

docs/sources/clients/promtail/_index.md
```diff
@@ -73,6 +97,7 @@ parsed data to Loki. Important details are:
 * Log rotations **aren't supported as of now**, mostly because it requires us modifying Promtail to
   rely on file inodes instead of file names. If you'd like to see support for it, please create a new
   issue on Github asking for it and explaining your use case.
+* If you compress a file under a folder being scrapped, Promtail might try to ingest your file before you finish compressing it. To avoid it, pick a `initial_delay` that is enough to avoid it.
```
JStickler, line 100: I think "scrapped" should be "scraped" here?

docs/sources/clients/promtail/configuration.md
```diff
@@ -335,6 +335,18 @@ job_name: <string>
 # Describes how to transform logs from targets.
 [pipeline_stages: <pipeline_stages>]
 
+# Defines decompression behavior for the given scrape target.
+decompression:
+  # Whether decompression should be tried or not.
+  [enabled: <boolean> | default = false]
+
+  # Initial delay to wait before starting the decompression.
+  # Specially useful in scenarios where compressed files are found before the compression is finished.
```
JStickler, line 344:
```suggestion
  # Especially useful in scenarios where compressed files are found before the compression is finished.
```

docs/sources/clients/promtail/_index.md
DylanGuedes, line 100:
```suggestion
* If you compress a file under a folder being scraped, Promtail might try to ingest your file before you finish compressing it. To avoid it, pick a `initial_delay` that is enough to avoid it.
```

PR #9011, opened 2023-04-04
Rename per-request query limits.

jeschkies:
`maxQueryTime` is more in line with the other per-request limits we have.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/util/querylimits/propagation.go
+++ b/pkg/util/querylimits/propagation.go
@@ -22,11 +22,11 @@ const (
 // NOTE: we use custom `model.Duration` instead of standard `time.Duration` because,
 // to support user-friendly duration format (e.g: "1h30m45s") in JSON value.
 type QueryLimits struct {
-	MaxQueryLength          model.Duration   `json:"maxQueryLength,omitempty"`
+	MaxQueryLength          model.Duration   `json:"maxQueryTimeRange,omitempty"`
 	MaxQueryRange           model.Duration   `json:"maxQueryInterval,omitempty"`
 	MaxQueryLookback        model.Duration   `json:"maxQueryLookback,omitempty"`
 	MaxEntriesLimitPerQuery int              `json:"maxEntriesLimitPerQuery,omitempty"`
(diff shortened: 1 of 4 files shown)


pkg/util/querylimits/propagation.go
```diff
@@ -22,11 +22,11 @@ const (
 // NOTE: we use custom `model.Duration` instead of standard `time.Duration` because,
 // to support user-friendly duration format (e.g: "1h30m45s") in JSON value.
 type QueryLimits struct {
-	MaxQueryLength          model.Duration   `json:"maxQueryLength,omitempty"`
+	MaxQueryLength          model.Duration   `json:"maxQueryTimeRange,omitempty"`
 	MaxQueryRange           model.Duration   `json:"maxQueryInterval,omitempty"`
 	MaxQueryLookback        model.Duration   `json:"maxQueryLookback,omitempty"`
 	MaxEntriesLimitPerQuery int              `json:"maxEntriesLimitPerQuery,omitempty"`
-	QueryTimeout            model.Duration   `json:"queryTimeout,omitempty"`
+	QueryTimeout            model.Duration   `json:"maxQueryTime,omitempty"`
```
DylanGuedes, line 29: hmm that's a breaking change, I'd postpone it until 3.x but in case you think we should have this, make sure to have it on the upgrading guide and changelog

DylanGuedes (changes requested):

pkg/util/querylimits/propagation.go
jeschkies: It's not really a breaking change because we didn't release/advertise the per-request limit API.

DylanGuedes: aháaa there you go. thank you.

cstyan:
personally I think these names are more confusing than the current names since they're so similar

jeschkies:
> personally I think these names are more confusing than the current names since they're so similar

I had the same thoughts. What do you think about just renaming query timeout? Or maybe drop that limit because it's so different from the others.

PR #9022, opened 2023-04-04
Loki: Add max number of flush retries

### LINKED ISSUE #8821: abort flushing chunk after N retries
One of my chunks is stuck in the flushLoop() with the problem of failing to flush because of chunk_store returning an exception and as a result the chunk keeps getting readded to the queue to be flushed again in the future as a retry.

The chunk_store CANNOT handle the flush therefore this needs to be mitigated on loki's side. It is not acceptable to continously retry and fail, I need a way to delete that chunk. The catalyst of that failure to flush was fixed on the agent's side but now I'm stuck with a chunk continously retrying to flush.

A flag that describes the max retries for a flushLoop() until it gives up and stops readding it to the queue to try again. I don't see how the cycle of try flush -> fail -> add back to queue -> try to flush -> fail -> add back to queue without a fallback is manageable.

**Additional context**
The concern in question is here https://github.com/grafana/loki/blob/8b8d2ca382d29c0d5612dee956dfe48ae2b51d10/pkg/ingester/flush.go#L147-L151

DylanGuedes:
Adds a new configuration to set the max number of flush retries, disabled by default.
Especially useful for scenarios where object storage experiences an outage and it is desired to alleviate Ingesters load.

Fixes https://github.com/grafana/loki/issues/8821

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/ingester/flush.go
+++ b/pkg/ingester/flush.go
@@ -78,6 +78,7 @@ type flushOp struct {
 	userID    string
 	fp        model.Fingerprint
 	immediate bool
+	try       int
 }
 
 func (o *flushOp) Key() string {
@@ -122,7 +123,7 @@ func (i *Ingester) sweepStream(instance *instance, stream *stream, immediate boo
 	firstTime, _ := stream.chunks[0].chunk.Bounds()
 	i.flushQueues[flushQueueIndex].Enqueue(&flushOp{
 		model.TimeFromUnixNano(firstTime.UnixNano()), instance.instanceID,
-		stream.fp, immediate,
+		stream.fp, immediate, 0,
 	})
 }
 
@@ -147,6 +148,11 @@ func (i *Ingester) flushLoop(j int) {
 		// If we're exiting & we failed to flush, put the failed operation
 		// back in the queue at a later point.
 		if op.immediate && err != nil {
+			if i.cfg.MaxFlushRetries > 0 && op.try > i.cfg.MaxFlushRetries {
+				continue
+			}
+			op.try += 1
+
 			op.from = op.from.Add(flushBackoff)
 			i.flushQueues[j].Enqueue(op)
 		}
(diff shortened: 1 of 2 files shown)


pkg/ingester/flush.go
```diff
@@ -147,6 +148,11 @@ func (i *Ingester) flushLoop(j int) {
 		// If we're exiting & we failed to flush, put the failed operation
 		// back in the queue at a later point.
 		if op.immediate && err != nil {
+			if i.cfg.MaxFlushRetries > 0 && op.try > i.cfg.MaxFlushRetries {
+				continue
```
dannykopping, line 152: We should add a log line here.

dannykopping (commented):
The streams which cannot be flushed will remain in the WAL, right? I wonder if this will have downstream effects like not being able to checkpoint/truncate. I don't know the WAL code that well so my understanding could be off.

pkg/ingester/flush.go
DylanGuedes: good call

DylanGuedes:
> The streams which cannot be flushed will remain in the WAL, right? I wonder if this will have downstream effects like not being able to checkpoint/truncate. I don't know the WAL code that well so my understanding could be off.

that's a good call, I'll check that. I'm turning this into a draft to test it better before marking it ready for review, if you find other things to me to test/check let me know.

pkg/ingester/flush.go
liguozhong: This `continue` means that the data is lost? If yes, in addition to logs, we also need a prometheus metrics to send alarms. This metrics is used to notify our SRE to quickly check whether there is a problem with S3 and Cassandra server.

DylanGuedes: good observation. But just fyi, the idea was to have `maxflushretries` as 0 by default to make sure you'd only suffer from it in case you're already aware of its implications.

PR #9025, opened 2023-04-04
These features did not make the release

trevorwhitney:

These features did not actually make it into this release.


stevendungan:
Looking at this PR; did we determine **Query max bytes read limit** _did_ make the release, and that's why it is staying in the release notes? Otherwise we'd remove that also?

stevendungan:
and I think we should also update the changelog to move these PR's from 2.8 to `next`

JStickler (approved):
[docs team] LGTM

PR #9030, opened 2023-04-05
Updated to explicitly call our simple scalable deployment mode

Alex3k:
The docs prior to this change reference "A scalable deployment" which implies that some of our deployments do not scale which isn't a great message to have in high traffic docs like this page. Therefore I propose updating it to explicitly call our Loki's Simple Scalable deployment mode as we link to our deployment mode docs and all but the unsupported monolithic deployment scale very happily.


JStickler (approved):
[docs team]  I'm going to approve this for now, but will note that there are currently ongoing discussions around what specifically call each deployment model.

MasslessParticle:
@Alex3k Can you fix the merge conflict? Then we can merge

PR #9032, opened 2023-04-05
Updated Scalability benefit to call out decoupled read/write paths an…

Alex3k:
Updated Scalability benefit to call out decoupled read/write paths and that benefit - this is a huge competitive edge over tools like Splunk and Elastic and it makes sense to explicitly call it out under our benefits


JStickler (approved):
[doc team] LGTM

PR #9039, opened 2023-04-05
Loki: log all trace ID's but include `sampled=true` when the trace is logged

slim-bean:

Currently when we log "with context" we will only add the traceID if the particular trace is being sampled. This is nice because it makes it obvious that if you go search for a trace you'll find it.

However, traceID is also useful for selecting all the log lines from a single query within Loki itself and if the trace wasn't sampled we remove this useful piece of context for querying the log lines.

In this PR, instead of dropping the traceID from the logs, we instead prefix it with `NS-`

Where NS is short for Not Sampled

Example:
```
level=debug ts=2023-04-05T15:06:24.377067817Z caller=fetcher.go:292 org_id=145265 traceID=NS-13ed6c30228e221d chunks=1 decodeRequests=1 missing=0
```

Questions: 

* should we bikeshed a little on this prefix? (I wanted something concise but if it's concise it's probably confusing to folks who've never seen it before)
* Instead of a prefix, we could change the logfmt key? `ID=13ed6c30228e221d` (this is less confusing in some ways but maybe more difficult/confusing in others?)

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/util/log/wrappers.go
+++ b/pkg/util/log/wrappers.go
@@ -31,10 +31,13 @@ func WithContext(ctx context.Context, l log.Logger) log.Logger {
 		l = WithUserID(userID, l)
 	}
 
-	traceID, ok := tracing.ExtractSampledTraceID(ctx)
-	if !ok {
+	traceID, sampled := tracing.ExtractSampledTraceID(ctx)
+	if sampled {
+		return log.With(l, "traceID", traceID, "sampled", "true")
+	}
+	if traceID != "" {
 		return log.With(l, "traceID", traceID)
 	}
+	return l
 
-	return log.With(l, "traceID", traceID, "sampled", "true")
 }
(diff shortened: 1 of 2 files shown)


dannykopping (commented):
The prefix is fine, but only if we document it somewhere.

pkg/util/log/wrappers.go
```diff
@@ -33,7 +33,7 @@ func WithContext(ctx context.Context, l log.Logger) log.Logger {
 
 	traceID, ok := tracing.ExtractSampledTraceID(ctx)
 	if !ok {
-		return l
+		return log.With(l, "traceID", "NS-"+traceID)
```
owen-d, line 36:
```suggestion
		return log.With(l, "traceID", traceID, "sampled", "false")
```

owen-d: This would be my preference

owen-d (approved):

PR #9047, opened 2023-04-05
helm: add setting 'useFederatedToken' for azure config

charlrvd:

This is adding back some changes that have been removed a while back. Some more context [here](https://github.com/grafana/loki/issues/8450)

This adds back the possibility to use azure federated tokens when using the loki chart.

As far as the usability of the feature is, I have used the version that was using a build of loki from the main branch with the azure federated token feature and it works fine so it would be pretty beneficial to be able to use it while using the normal releases of Loki.

I only undid the changes that removed the flag

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/installation/helm/reference.md
+++ b/docs/sources/installation/helm/reference.md
@@ -1892,6 +1892,7 @@ null
     "accountKey": null,
     "accountName": null,
     "requestTimeout": null,
+    "useFederatedToken": false,
     "useManagedIdentity": false,
     "userAssignedId": null
   },
--- a/production/helm/loki/CHANGELOG.md
+++ b/production/helm/loki/CHANGELOG.md
@@ -15,6 +15,10 @@ Entries should include a reference to the pull request that introduced the chang
 
 - [BUGFIX] Fix dns port in network-policy
 
+## 5.1.1
+
+- [FEATURE] add setting `useFederatedToken` from Azure configuration block back.
+
 ## 4.10.0
 
 - [CHANGE] Changed version of Grafana Enterprise Logs to v1.6.3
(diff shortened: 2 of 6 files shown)


docs/sources/installation/helm/reference.md
```diff
@@ -1881,6 +1881,7 @@ null
     "accountKey": null,
     "accountName": null,
     "requestTimeout": null,
+    "useFederatedToken": false,
```
JStickler, line 1884: See the note at line 9:  This file is autogenerated. Modify ../production/helm/loki/reference.md.gotmpl instead.

JStickler (commented):
[docs team]  Please update the correct file for the reference documentation.

docs/sources/installation/helm/reference.md
charlrvd: My bad, thanks for pointing this out

JStickler: No worries, I just don't want doc updates to get overwritten and lost.  =)

charlrvd: I am slightly confused. If I remove that change the doc will be updated with it since the helm-docs command will generate with it, as it comes out of the values.yaml. Is there something I am missing to update or should I just rebase without that file change?
using `helm-docs -t reference.md.gotmpl -o ../../../docs/sources/installation/helm/reference.md` just ends up with the same change above so I am not sure now if I should change something else

jeschkies: It should be alright. We have a CI check for this.

JStickler (approved):
[docs team] LGTM

jeschkies (approved):

jeschkies:
I think that was covered by https://github.com/grafana/loki/pull/9168.

PR #9054, opened 2023-04-06
Fix CHANGELOG link

dannykopping:
https://github.com/grafana/loki/pull/9048 only updated the link in the `/next` version of the docs which reflects `main`. This will update the `/latest` link which reflects the latest version (2.8).


JStickler (approved):
[docs team] LGTM

PR #9063, opened 2023-04-06
Add release notes appender action

trevorwhitney:

This PR aims to remove the manual process required to create release notes during a release. It will allow contributors to add the label `add-to-release-notes` to any PR. When that PR is merged, this action will create another PR appending the original PR's # and title to the release notes for the next release. This second PR will give the author an opportunity to add a description to their addition, as well as give maintainers an opportunity to discuss it's relevance in the release notes.

If we move forward with this approach, I will follow up with an addition to the PR template to ask contributors if they think their PR is worthy of mention in the next releases' release notes, and if so to add the label.

Depends on new [github action](https://github.com/grafana/grafana-github-actions/pull/138). After both these PRs are merged I'll follow up with a PR to change the workflow to point to the version in the grafana repo.

### DIFF
--- a/.github/workflows/append-release-notes.yml
+++ b/.github/workflows/append-release-notes.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           repository: "trevorwhitney/grafana-github-actions"
           path: ./actions
-          ref: 01725b6aaaca42e31776a5d35d26a110346b4a15
+          ref: 0d27c31b3f0c39f6f1f4378aa11b18f9a9fa6c23
       - name: Install Actions
         run: npm install --production --prefix ./actions
       - name: Run release notes appender
--- a/.github/workflows/append-release-notes.yml
+++ b/.github/workflows/append-release-notes.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Actions
-        uses: actions/checkout@v3.3.0
+        uses: actions/checkout@v3
         with:
           repository: "trevorwhitney/grafana-github-actions"
           path: ./actions
--- a/.github/workflows/append-release-notes.yml
+++ b/.github/workflows/append-release-notes.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           repository: "trevorwhitney/grafana-github-actions"
           path: ./actions
-          ref: release-notes-action
+          ref: d2320e2d28dee87520455ed4f7ab7522e088ea11
       - name: Install Actions
         run: npm install --production --prefix ./actions
       - name: Run release notes appender
--- a/.github/workflows/append-release-notes.yml
+++ b/.github/workflows/append-release-notes.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           repository: "trevorwhitney/grafana-github-actions"
           path: ./actions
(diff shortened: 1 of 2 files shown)


trevorwhitney:
The ergonomics of adding a new GitHub action are very confusing to me. This action will only run when I try to PR with `release-notes-appender` as my base branch. This was a branch I created in `origin`, but can no longer push to as I'm told it's "protected". However, I think what's happening here is that when I use `release-notes-appender` as my base, it's running the action as defined there already, which is why the title template is not working. The rest of the action is working fine because if fetches it from source every run, whereas the overridden `title` is cached in the workflow definition.

The title should be fixed by using the [default](https://github.com/trevorwhitney/grafana-github-actions/commit/0d27c31b3f0c39f6f1f4378aa11b18f9a9fa6c23) from the action, which should hopefully happen once this is merged, since I've removed the custom value from the workflow definition.

.github/workflows/append-release-notes.yml
```diff
@@ -0,0 +1,27 @@
...
+      - labeled
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Actions
+        uses: actions/checkout@v3.3.0
+        with:
+          repository: "trevorwhitney/grafana-github-actions"
```
trevorwhitney, line 16: will update this to reference the upstream repo is a subsequent PR once the action is merged.

.github/workflows/append-release-notes.yml
```diff
@@ -0,0 +1,27 @@
...
+          ref: 0d27c31b3f0c39f6f1f4378aa11b18f9a9fa6c23
+      - name: Install Actions
+        run: npm install --production --prefix ./actions
+      - name: Run release notes appender
+        uses: ./actions/release-notes-appender
+        with:
+          metricsWriteAPIKey: ${{secrets.GRAFANA_MISC_STATS_API_KEY}}
+          token: ${{secrets.GH_BOT_ACCESS_TOKEN}}
+          labelsToAdd: "release-notes"
+          releaseNotesFile: docs/sources/release-notes/v2-8.md
```
trevorwhitney, line 27: will update this to `docs/sources/release-notes/next.md` once merged, but that won't exist until this is merged

docs/sources/release-notes/next.md
```diff
@@ -0,0 +1,13 @@
+---
+title: V?.?
+description: Version ?.? release notes
+weight: 100000
+---
+
+# V?.?
+Grafana Labs is excited to announce the release of Loki ?.?. Here's a summary of new enhancements and important fixes:
+
+:warning: This a placeholder for the nex release. Please clean up all features listed below
```
MichelHollands, line 10:
```suggestion
:warning: This a placeholder for the next release. Please clean up all features listed below
```

.github/workflows/append-release-notes.yml
```diff
@@ -0,0 +1,27 @@
+---
+name: Append to Release Notes
+on:
+  pull_request_target:
+    types:
+      - closed
```
dannykopping, line 6: Why on closed PRs? Is there a `merged` type perhaps?

dannykopping (approved):
Let's go!

PR #9074, opened 2023-04-07
Tools/index analyzer

owen-d:
Adds a tool that allows downloading and querying statistics across TSDB indices.

Example output:
```
analyzed 1046302 series and 7607495 chunks for an average of 7.270841 chunks per series. max chunks/series was 45695. number of series with over 1k chunks: 15
```


DylanGuedes (changes requested):
This looks awesome, gj.

Two suggestions:
- WDYT of after having the series with more than 1k chunks, the ones with less than 100 chunks or something? If you have 100 series and all of them have 1k+ chunks then the distribution of data is potentially good, but if you have some with >100k chunks and some with <100chunks then the distribution is horrible
- WDYT of adding a way of doing the analysis also on bytes instead of just line numbers? (because of cases where lines have high discrepancy of bytes)

owen-d:
Happy to let people extend this as they see fit, just wanted to build the skeleton for it so I could do some basic index analysis to make better informed decisions while building tsdb v3

DylanGuedes (approved):
got it. I'll approve it since it looks good but also to unblock your work and then as you said we can iterate if we feel like something is lacking.

PR #9075, opened 2023-04-07
Runs `make fmt` over the whole repo

owen-d:
Some format housekeeping


clients/pkg/promtail/server/ui/assets_vfsdata.go
```diff
@@ -1,5 +1,6 @@
 // Code generated by vfsgen; DO NOT EDIT.
 
+//go:build !dev
```
DylanGuedes, line 3: double checking that this is necessary

operator/cmd/size-calculator/main.go
```diff
@@ -6,9 +6,10 @@ import (
 	"os"
 	"time"
 
-	"github.com/grafana/loki/operator/internal/sizes"
 	"github.com/prometheus/common/model"
 
+	"github.com/grafana/loki/operator/internal/sizes"
+
```
DylanGuedes, line 12: for some reason gofmt didn't put this after the next package :thinking:

pkg/querier/queryrange/limits_test.go
```diff
@@ -17,6 +16,8 @@ import (
 	"go.uber.org/atomic"
 	"gopkg.in/yaml.v2"
 
+	"github.com/grafana/loki/pkg/util/math"
```
DylanGuedes, line 19: hmm this one is isolated from other packages below (I'm not adding a diff suggestion for you to just apply because the order would be wrong anyway)

pkg/querier/queryrange/split_by_interval.go
```diff
@@ -5,14 +5,15 @@ import (
...
 
-	"github.com/grafana/loki/pkg/util/math"
 	"github.com/opentracing/opentracing-go"
 	otlog "github.com/opentracing/opentracing-go/log"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/model"
 	"github.com/weaveworks/common/httpgrpc"
 
+	"github.com/grafana/loki/pkg/util/math"
```
DylanGuedes, line 15: ditto: this one after `dskit/tenant` package

DylanGuedes (approved):
lgtm with some nits

periklis:
@owen-d FYI this touches sources in the operator sub folder and it looks like we use different formatting styles

clients/pkg/promtail/server/ui/assets_vfsdata.go
owen-d: Looks like it's just using a better (newer?) form of [build constraint](https://pkg.go.dev/cmd/go#hdr-Build_constraints) compared to the line below

periklis:
> Hey @periklis is that ok? This all comes from the `goimports -local` configs.
> 
> ```
> $ goimports --help
> usage: goimports [flags] [path ...]
>   -local string
>         put imports beginning with this string after 3rd-party packages; comma-separated list
> ```
> 
> which is called by `make format`
> 
> ```
> format:
> 	find . $(DONT_FIND) -name '*.pb.go' -prune -o -name '*.y.go' -prune -o -name '*.rl.go' -prune -o \
> 		-type f -name '*.go' -exec gofmt -w -s {} \;
> 	find . $(DONT_FIND) -name '*.pb.go' -prune -o -name '*.y.go' -prune -o -name '*.rl.go' -prune -o \
> 		-type f -name '*.go' -exec goimports -w -local github.com/grafana/loki {} \;
> ```
> 
> We can exclude the operator dir from the `make format` command if you'd prefer, but I personally like the formatting changes. WDYT?

I would like to skip the operator dir for now. As per now it would make backport bugs harder. I suggest the SIG operator lifts the exclude once we have applied `goimports` on `main` and present release branches`.

jeschkies (approved):
Thanks. I was hesitating to do this because I feared it would mess up `git blame`. However, these are just imports. 

Once this land we can simplify the `make check-format` step.

spiridonov:
Closing as stale.

PR #9077, opened 2023-04-07
Helm: Ability to change loki-read and loki-write services to NodePort

karlkeppner1:
When deployed to AWS EKS, services must be of type `NodePort` if `ingressClassName: alb`. By adding the ability to specify service type for the loki-read and loki-write services, EKS users can now use the chart without loki-gateway. Added ability to specify a `clusterIP` if service `type: ClusterIP`. Added ability to specify annotations for each service. 

I never raised an issue for this. I needed it personally, so I wrote it.

Previous defaults were added explicitly in values.yaml to maintain backwards compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,3 +143,5 @@ If you would like to add your organization to the list, please open a PR to add 
 ## License
 
 Grafana Loki is distributed under [AGPL-3.0-only](LICENSE). For Apache-2.0 exceptions, see [LICENSING.md](LICENSING.md).
+
+# changes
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -143,3 +143,5 @@ If you would like to add your organization to the list, please open a PR to add
 ## License
 
 Grafana Loki is distributed under [AGPL-3.0-only](LICENSE). For Apache-2.0 exceptions, see [LICENSING.md](LICENSING.md).
+
+# changes
```
MichelHollands, line 147: This entry should not be here. Changes should be added to the CHANGELOG

MichelHollands:
@karlkeppner1 Can you add a HELM changelog entry and update the Helm chart version to 5.5.8? An example can be found [here](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213).
Please also merge from main.
Once that is done this can be merged.

PR #9080, opened 2023-04-08
Docs: Update Loki GitHub URLs

darxriggs:
* use `main` instead of `master` branch
* use a stable version in case lines are referenced
* fix outdated URLs

### DIFF
--- a/operator/docs/operator/storage_size_calculator.md
+++ b/operator/docs/operator/storage_size_calculator.md
@@ -15,7 +15,7 @@ Storage Size Calculator is used to have an idea on how to properly size a Loki c
 
 ## Installation
 
-* Deploy the [Loki Operator](https://github.com/grafana/loki/blob/master/operator/docs/hack_loki_operator.md#hacking-on-loki-operator-on-openshift) to the cluster.
+* Deploy the [Loki Operator](https://github.com/grafana/loki/blob/main/operator/docs/operator/hack_loki_operator.md#hacking-on-loki-operator-on-openshift) to the cluster.
 
 * Create the `openshift-logging` namespace in the cluster:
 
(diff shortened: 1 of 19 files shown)


operator/docs/operator/storage_size_calculator.md
```diff
@@ -63,21 +63,21 @@ This will cleanup the resources related to storage size calculator. However, the
...
+* All the files related to storage size calculator exists at [`config/overlays/openshift/size-calculator`](https://github.com/grafana/loki/tree/main/operator/config/overlays/openshift/size-calculator) and the main file is at [`cmd/size-calculator`](https://github.com/grafana/loki/tree/main/operator/cmd/size-calculator).
 
 * Update the code to fix a bug or add a new feature.
 
-* To test the changes made, build the image and push it to quay. Replace [here](https://github.com/grafana/loki/blob/master/operator/config/overlays/openshift/size-calculator/storage_size_calculator.yaml#L18) with your quay image to test the changes.
+* To test the changes made, build the image and push it to quay. Replace [here](https://github.com/grafana/loki/blob/v2.8.0/operator/config/overlays/openshift/size-calculator/storage_size_calculator.yaml#L18) with your quay image to test the changes.
```
JStickler, line 72: As a best practice, avoid putting release numbers in URLs because they become stale.   Since the latest code is in main, I would change this to main.
```suggestion
* To test the changes made, build the image and push it to quay. Replace [here](https://github.com/grafana/loki/blob/main/operator/config/overlays/openshift/size-calculator/storage_size_calculator.yaml#L18) with your quay image to test the changes.
```

darxriggs: Using a version here was done by intention (the 2nd point in the description):
> use a stable version in case lines are referenced

This is to make sure that the referenced line 18 uses a stable reference. Otherwise, if this file is changed and something is added/removed above line 18 in this file, the link would then point to something different than when created initially. I have seen this much too often.

If you don't like using a version, a git commit would do too, having the same effect.

If you would want to always refer to the `main` branch, the line reference should be removed and maybe only the `image` option mentioned.

JStickler: Good point, line 18 might move around.  Let's try this instead:

```suggestion
* To test the changes made, build the image and push it to quay. Replace [spec.template.spec.containers.image](https://github.com/grafana/loki/blob/main/operator/config/overlays/openshift/size-calculator/storage_size_calculator.yaml) with your quay image to test the changes.
```

darxriggs: Done. I opted for another wording as `spec.template.spec.containers.image` would not be correct because `containers` is a list.

JStickler (approved):
[docs team] LGTM

PR #9086, opened 2023-04-10
Loki: Fix span event logging on rate store

DylanGuedes:
Fix an event logging on rate store that is missing the `event` key. `LogKV` expects a key-value pair always.


dannykopping (approved):
LGTM

PR #9090, opened 2023-04-10
[helm] Updating loki-logs dashboard

### LINKED ISSUE #9089: [helm] Loki Logs Dashboard showing only No Data
**Describe the bug**
A default installation of the loki-logs.json Dashboard always shows No Data for most items. It looks like for most items `cluster` is not attached to metrics besides those that are go related.

**To Reproduce**
1. Started Loki 2.7.5
2. Started Promtail 2.7.1 to tail k8s logs

**Expected behavior**
loki-logs dashboard lists pods + containers and shows metrics plus logs

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: helm

Current Dashboard deployment:

loki-logs-current

rgarcia6520:
Updating loki-logs.json dashboard expressions to remove cluster entry when it's not needed.

Fixes #9089

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/src/dashboards/loki-logs.json
+++ b/production/helm/loki/src/dashboards/loki-logs.json
@@ -237,7 +237,7 @@
             "steppedLine": false,
             "targets": [
                {
-                  "expr": "sum(rate(container_cpu_usage_seconds_total{cluster=\"$cluster\", namespace=\"$namespace\", pod=~\"$deployment.*\", pod=~\"$pod\", container=~\"$container\"}[5m]))",
+                  "expr": "sum(rate(container_cpu_usage_seconds_total{namespace=\"$namespace\", pod=~\"$deployment.*\", pod=~\"$pod\", container=~\"$container\"}[5m]))",
                   "refId": "A"
                }
             ],
@@ -323,7 +323,7 @@
             "steppedLine": false,
             "targets": [
                {
-                  "expr": "sum(container_memory_working_set_bytes{cluster=\"$cluster\", namespace=\"$namespace\", pod=~\"$deployment.*\", pod=~\"$pod\", container=~\"$container\"})",
+                  "expr": "sum(container_memory_working_set_bytes{namespace=\"$namespace\", pod=~\"$deployment.*\", pod=~\"$pod\", container=~\"$container\"})",
                   "refId": "A"
                }
             ],
@@ -409,7 +409,7 @@
             "steppedLine": false,
             "targets": [
                {
-                  "expr": "sum(rate(container_network_transmit_bytes_total{cluster=\"$cluster\", namespace=\"$namespace\", pod=~\"$deployment.*\", pod=~\"$pod\"}[5m]))",
+                  "expr": "sum(rate(container_network_transmit_bytes_total{namespace=\"$namespace\", pod=~\"$deployment.*\", pod=~\"$pod\"}[5m]))",
                   "refId": "A"
                }
             ],
@@ -495,7 +495,7 @@
             "steppedLine": false,
             "targets": [
                {
-                  "expr": "sum(rate(container_network_receive_bytes_total{cluster=\"$cluster\", namespace=\"$namespace\", pod=~\"$deployment.*\", pod=~\"$pod\"}[5m]))",
+                  "expr": "sum(rate(container_network_receive_bytes_total{namespace=\"$namespace\", pod=~\"$deployment.*\", pod=~\"$pod\"}[5m]))",
                   "refId": "A"
                }
             ],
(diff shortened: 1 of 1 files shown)


rgarcia6520:
Here is Dashboard screenshot with changes deployed:

loki-logs-updated

production/helm/loki/src/dashboards/loki-logs.json
```diff
@@ -237,7 +237,7 @@
             "steppedLine": false,
             "targets": [
                {
-                  "expr": "sum(rate(container_cpu_usage_seconds_total{cluster=\"$cluster\", namespace=\"$namespace\", pod=~\"$deployment.*\", pod=~\"$pod\", container=~\"$container\"}[5m]))",
+                  "expr": "sum(rate(container_cpu_usage_seconds_total{namespace=\"$namespace\", pod=~\"$deployment.*\", pod=~\"$pod\", container=~\"$container\"}[5m]))",
```
jeschkies, line 240: @trevorwhitney you've explained this to me but I've forgot again. Why was the cluster label an issue and was it required?

rgarcia6520: It is a metric label/annotation on the goroutine or go specific metric info, but for the last several chart releases our version of this dashboard has been mostly empty because the pod metrics for CPU/MEM/etc don't have `cluter` label/annotations

trevorwhitney (changes requested):
Did you generate this from https://github.com/grafana/loki/tree/main/production/loki-mixin? If so do you mind including your command?

I see 2 problems with this. One is we want to keep these dashboards in line with the ones we use for monitoring our cloud SAAS so that they stay up to date. For that use case we need the cluster label. 

Furthermore, if you're using a cloud Grafana instance, the [kubernetes monitoring plugin](https://grafana.com/docs/grafana-cloud/kubernetes-monitoring/configuration/config-k8s-agent-guide/) will add a cluster label to your metrics so these work for that case too. This is another use case we need to continue to support, granted it's a bit more complicated because in that flow you would probably use the cloud Loki integration which installs the dashboards and doesn't use these directly, however both a generated from the mixin.

I think the better solution here is to have these dashboards generated by a job that compiles the mixin. This way, we could add more configuration values to the mixin (like whether or not you want a cluster) and expose them via the helm chart's values, making these more widely applicable.

Alternatively, we could remove the dashboards altogether, and instruct users to compile the mixin to get dashaboards that work for them.

As it stands, I don't think static dashboards are going to catch everyone's use case. Currently we're at least capturing what's needed to monitor the Loki instances run in Grafana Cloud,  and I don't see a compelling reason to exclude that case.

rgarcia6520:
> Did you generate this from https://github.com/grafana/loki/tree/main/production/loki-mixin? If so do you mind including your command?
> 
> I see 2 problems with this. One is we want to keep these dashboards in line with the ones we use for monitoring our cloud SAAS so that they stay up to date. For that use case we need the cluster label.
> 
> Furthermore, if you're using a cloud Grafana instance, the [kubernetes monitoring plugin](https://grafana.com/docs/grafana-cloud/kubernetes-monitoring/configuration/config-k8s-agent-guide/) will add a cluster label to your metrics so these work for that case too. This is another use case we need to continue to support, granted it's a bit more complicated because in that flow you would probably use the cloud Loki integration which installs the dashboards and doesn't use these directly, however both a generated from the mixin.
> 
> I think the better solution here is to have these dashboards generated by a job that compiles the mixin. This way, we could add more configuration values to the mixin (like whether or not you want a cluster) and expose them via the helm chart's values, making these more widely applicable.
> 
> Alternatively, we could remove the dashboards altogether, and instruct users to compile the mixin to get dashaboards that work for them.
> 
> As it stands, I don't think static dashboards are going to catch everyone's use case. Currently we're at least capturing what's needed to monitor the Loki instances run in Grafana Cloud, and I don't see a compelling reason to exclude that case.

They were deployed directly from `production/helm/loki/src`  as configmaps then imported by prometheus-operator from kube-prometheus-stack alongside the `loki` v5.0.0 chart in that parent directory. When deploying that loki chart the pods that display in prometheus don't have a `cluster` attribute:

```
container_cpu_usage_seconds_total{container="loki", cpu="total", endpoint="https-metrics", id="/kubepods/burstable/pod52e01f9d-fc60-435f-9a57-8f9aab7a184f/5775d074564fc75315928f1a10af9259b22736db169bfbc99e698b40b28d7f561", image="grafana/loki:2.8.0", instance="172.18.0.2:10250", job="kubelet", metrics_path="/metrics/cadvisor", name="XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX", namespace="logging", node="k3d-k3s-default-server-0", pod="logging-loki-backend-2", service="monitoring-monitoring-kube-kubelet"}
```

trevorwhitney:
@rgarcia6520 since `container_cpu_usage_seconds_total` comes from `kube-state-metrics`, the presence or absence of a cluster label is going to depend on your scrape config for `kube-state-metrics`, right? Could this problem also be solved by adding a cluster label to the `relabel_configs` of your `kube-state-metrics` deployment?

MasslessParticle:
Closing for inactivity

Issue #9060, opened 2023-04-06
The official Linux release binary of promtail 2.8.0 is built without systemd journal support (on at least amd64)


siebenmann:
**Describe the bug**
The official promtail binaries for Linux amd64 are built without support for reading from the systemd journal. Starting them with the journal configured as a log source gets you:

```
level=info ts=2023-04-06T15:19:32.082276461Z caller=promtail.go:133 msg="Reloading configuration file" md5sum=f77bc87e64d66cca88ef9c78b403eb79
level=warn ts=2023-04-06T15:19:32.082499874Z caller=journaltargetmanager.go:29 msg="WARNING!!! Journal target was configured but support for reading the systemd journal is not compiled into this build of promtail!"
level=info ts=2023-04-06T15:19:32.084100883Z caller=server.go:323 http=[::]:9080 grpc=[::]:9095 msg="server listening on addresses"
level=info ts=2023-04-06T15:19:32.084264747Z caller=main.go:174 msg="Starting Promtail" version="(version=2.8.0, branch=HEAD, revision=90888a0cc)"
level=warn ts=2023-04-06T15:19:32.084327616Z caller=promtail.go:265 msg="enable watchConfig"
```
I suspect that promtail binaries were not built with [the new build tag to include the journal](https://github.com/grafana/loki/blob/release-2.8.x/docs/sources/upgrading/_index.md#the-go-build-tag-promtail_journal_enabled-was-introduced) that's covered in the 2.8 upgrade notes.

m1keil:
just hit the same issue...

dannykopping:
Thank you for reporting this.

We will get it sorted out and update our build automation to prevent reoccurrences in the future.

jeschkies:
This fix is up for review. I'm a little confused when this regression was introduced though.

frittentheke:
@trevorwhitney are you pushing out that 2.8.1 with just journald support (re)added or are you waiting for more bugfixes?

trevorwhitney:
It will go out this week. There were a couple other fixes we wanted in there.

trevorwhitney:
fixed in the latest release! https://github.com/grafana/loki/releases/tag/v2.8.1

PR #8131, opened 2023-01-13
Cross build Promtail for ARM with journald support.

jeschkies:
**What this PR does / why we need it**:
This is the follow up to #8111 and adds journald support for Promatil ARM and ARM64 builds.

I've tested it locally with my Raspberry Pi and it works.

**Which issue(s) this PR fixes**:
Fixes #1459

**Special notes for your reviewer**:

**Checklist**
- [ ] Reviewed the [`CONTRIBUTING.md`](../CONTRIBUTING.md) guide (**required**)
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #9095, opened 2023-04-10
Add journald support to Promtail amd64 binary builds

JordanRushing:

Previously, in https://github.com/grafana/loki/pull/8131, we enabled journald support for Promtail arm/arm64 binary builds but not amd64.

- https://github.com/grafana/loki/issues/9060

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


.drone/drone.yml
```diff
@@ -1463,6 +1463,7 @@ steps:
     NFPM_PASSPHRASE:
       from_secret: gpg_passphrase
     NFPM_SIGNING_KEY_FILE: /drone/src/private-key.key
+    PROMTAIL_JOURNAL_ENABLED: "true"
```
DylanGuedes, line 1466: could you confirm whether "true" here should be in string or boolean ("true" vs true like at line 1454)?

DylanGuedes (changes requested):

.drone/drone.yml
JordanRushing: Good call; let me check

jeschkies (changes requested):
Please also change the `dist` target for ARM builds. We pass the promtail journald tag explicitly which we probably should not be doing.

.drone/drone.jsonnet
```diff
@@ -837,6 +837,7 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
             GITHUB_TOKEN: { from_secret: github_secret.name },
             NFPM_PASSPHRASE: { from_secret: gpg_passphrase.name },
             NFPM_SIGNING_KEY_FILE: '/drone/src/private-key.key',
+            PROMTAIL_JOURNAL_ENABLED: 'true',
```
jeschkies, line 840: I think we should add these to `test packaging` as well to ensure that the build image has all dev dependencies required for building promtail with journald enabled.

jeschkies (approved):

trevorwhitney (approved):
lgtm

PR #9103, opened 2023-04-11
Ignore broken query limits header.

### LINKED ISSUE #9045: Ensure broken per-request limits will not break an API call.
If the content of `X-Loki-Query-Limits` is invalid the request should not be rejected. Instead the header should be ignored.

jeschkies:
A broken or invalid `X-Loki-Query-Limits` header should be ignored because it might have been injected by a proxy. That means a user cannot correct the header and would be unable to perform a query.

Note that we ignore the header when its not properly encoded and when it contains invalid entries.

Fixes #9045

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/util/querylimits/middleware.go
+++ b/pkg/util/querylimits/middleware.go
@@ -27,11 +27,8 @@ func (l *queryLimitsMiddleware) Wrap(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		limits, err := ExtractQueryLimitsHTTP(r)
 		if err != nil {
-			w.WriteHeader(http.StatusBadRequest)
-			if _, err := w.Write([]byte(err.Error())); err != nil {
-				level.Error(util_log.Logger).Log("msg", "error in queryLimitsMiddleware Wrap", "err", err)
-			}
-			return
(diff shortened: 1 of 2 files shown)


pkg/util/querylimits/middleware.go
```diff
@@ -27,11 +27,8 @@ func (l *queryLimitsMiddleware) Wrap(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		limits, err := ExtractQueryLimitsHTTP(r)
 		if err != nil {
-			w.WriteHeader(http.StatusBadRequest)
-			if _, err := w.Write([]byte(err.Error())); err != nil {
-				level.Error(util_log.Logger).Log("msg", "error in queryLimitsMiddleware Wrap", "err", err)
-			}
-			return
```
cstyan, line 34: Is there a nice way we could still notify users that the query limits contents were broken?

cstyan (approved):

pkg/util/querylimits/middleware.go
jeschkies: I think cloud insights with some alerting.

PR #9108, opened 2023-04-11
Remove regex simplification that was causing panics

trevorwhitney:

We have some queries that were causing panics in the query-frontends in one of our cloud environments. We used one of those queries to figure out how to fix this, but cannot include that query as is, so we will need to follow up with a test that exposes the problem.

In the meantime, we'd like to get this merged so we can stop the panics.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


MasslessParticle:
Found a root cause for this issue closing this draft

PR #9128, opened 2023-04-13
Add filter on rules' labels to API query

aminesnow:
The goal of this PR is to introduce a filter on rule labels when querying them via the API.
The idea is to add a query parameter named `labels` to the `/loki/api/v1/rules` route. This parameter is a comma `,` separated list of label key/values formatted as `key:value`. 
By including this parameter in the query, only rules that have at least one of the labels key/value defined in the parameter will be fetched.
For instance: `/loki/api/v1/rules?labels=foo:bar,tik:tok` will only fetch rules that have one or both of the `foo: bar` and `tik: tok` labels.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/ruler/base/api.go
```diff
@@ -356,77 +356,151 @@ func parseGroupName(params map[string]string) (string, error) {
...
+		ls = make([]labels.Label, len(labelStrings))
+
+		for i, l := range labelStrings {
+			kv := strings.Split(l, ":")
+			if len(kv) != 0 {
+				ls[i] = labels.Label{
+					Name:  kv[0],
+					Value: kv[1],
+				}
+			}
```
dannykopping, line 409: Can you please split this out into its own function so that we can unit-test it independently?

pkg/ruler/base/api.go
```diff
@@ -356,77 +356,151 @@ func parseGroupName(params map[string]string) (string, error) {
...
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
-	level.Debug(logger).Log("msg", "retrieved rule groups from rule store", "userID", userID, "num_namespaces", len(rgs))
+	level.Debug(logger).Log("msg", "retrieved rule groups from rule store", "userID", pr.UserID, "num_namespaces", len(rgs))
+
+	if len(pr.Labels) > 0 {
+		level.Info(logger).Log("msg", "filtering rule groups with labels", "abels", pr.Labels, "userID", pr.UserID, "namespace", pr.Namespace)
```
dannykopping, line 487: Not quite sure why we need an additional log line for this; can we not add the filters to the previous log line?

```suggestion
		level.Info(logger).Log("msg", "filtering rule groups with labels", "labels", pr.Labels, "userID", pr.UserID, "namespace", pr.Namespace)
```

pkg/ruler/base/api_test.go
```diff
@@ -163,6 +164,53 @@ func TestRuler_alerts(t *testing.T) {
...
+
+	tc := []struct {
+		name     string
+		URLQuery string
+		output   string
+		err      error
+	}{
+		{
+			name:   "query with no filters",
+			output: "test:\n    - name: group1\n      interval: 1m\n      rules:\n        - record: UP_RULE\n          expr: up\n        - alert: UP_ALERT\n          expr: up < 1\n          labels:\n            foo: bar\n        - alert: DOWN_ALERT\n          expr: down < 1\n          labels:\n            namespace: delta\n",
```
dannykopping, line 183: This is hard to read, and will be hard to edit too.
Can we rather parse the responses and deep compare the result to a struct?

pkg/ruler/base/api_test.go
```diff
@@ -163,6 +164,53 @@ func TestRuler_alerts(t *testing.T) {
...
+		},
+		{
+			name:     "query with a valid filter found",
+			URLQuery: "labels=namespace:delta,foo:bar",
+			output:   "test:\n    - name: group1\n      interval: 1m\n      rules:\n        - alert: UP_ALERT\n          expr: up < 1\n          labels:\n            foo: bar\n        - alert: DOWN_ALERT\n          expr: down < 1\n          labels:\n            namespace: delta\n",
+		},
+		{
+			name:     "query with an invalid query param",
```
dannykopping, line 191: I'm not sure how to read this test-case. Should it just return the default list if the param is empty?

dannykopping (commented):
Overall looks like a nice feature to have, thanks!

Left mostly nits

pkg/ruler/base/api.go
aminesnow: Not sure I understand what you mean.. Are you suggesting to convert this into a DEBUG log line? The reason why I added the log line was because label filtering happens only when we have labels

pkg/ruler/base/api_test.go
aminesnow: Sure we can, that's what I did initially, but it made the test lengthier because of the steps it took to get the right data structure, but maybe my implementation wasn't the best. I'll re implement it so you can tell me if that's what you had in mind.

pkg/ruler/base/api_test.go
aminesnow: Here we have a query param that is not recognized/handled by the API, so it's just ignored, that's why it's returning the result as if no query param was specified.

aminesnow:
> Overall looks like a nice feature to have, thanks!
> 
> Left mostly nits

Thank you for the review @dannykopping! 
I updated the PR with your suggestions, let me know what you think.

pkg/ruler/base/api_test.go
dannykopping: I think if you defined a const with the default rule group which returns both for `query with no filters` and this test case, it'll read a lot more clearly.

dannykopping:
> > Overall looks like a nice feature to have, thanks!
> > Left mostly nits
> 
> Thank you for the review @dannykopping! I updated the PR with your suggestions, let me know what you think.

Thanks @aminesnow 

One last comment to make the tests more understandable, then this is good to go

aminesnow:
@dannykopping any updates on this?

dannykopping (approved):
LGTM, sorry I didn't see that you had push an update

PR #9130, opened 2023-04-13
Pass engine opts down to middlewares

salvacorts:

The following middlewares in the query frontend uses a downstream engine:
- `NewQuerySizeLimiterMiddleware` and `NewQuerierSizeLimiterMiddleware`
- `NewQueryShardMiddleware`
- `NewSplitByRangeMiddleware`

These were all creating the downstream engine as follows:

```go
logql.NewDownstreamEngine(logql.EngineOpts{LogExecutingQuery: false}, DownstreamHandler{next: next, limits: limits}, limits, logger),
```

As can be seen, the [engine options configured in Loki][1] were not being used at all. In the case of `NewQuerySizeLimiterMiddleware`, `NewQuerierSizeLimiterMiddleware` and `NewQueryShardMiddleware`, the downstream engine was created to get the `MaxLookBackPeriod`. When creating a new Downstream Engine as above, the `MaxLookBackPeriod` [would always be the default][2] (30 seconds).

This PR fixes this by passing down the engine config to these middlewares, so this config is used to create the new downstream engines.

Adresses some pending tasks from https://github.com/grafana/loki/pull/8670#issuecomment-1507031976.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

[1]: https://github.com/grafana/loki/blob/1bcf683513b5e11fb1dcc5625761b9760cfbd958/pkg/querier/querier.go#L52
[2]: https://github.com/grafana/loki/blob/edc6b0bff76714ff584c20d7ff0235461a4f4a88/pkg/logql/engine.go#L136-L140

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * [8971](https://github.com/grafana/loki/pull/8971) **dannykopping**: Stats: fix `Cache.Chunk.BytesSent` statistic and loki_chunk_fetcher_fetched_size_bytes metric with correct chunk size.
 * [8979](https://github.com/grafana/loki/pull/8979) **slim-bean**: Fix the case where a logs query with start time == end time was returning logs when none should be returned.
 * [9099](https://github.com/grafana/loki/pull/9099) **salvacorts**: Fix the estimated size of chunks when writing a new TSDB file during compaction.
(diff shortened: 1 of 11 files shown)


jeschkies (commented):
Great find! Could you add a test?

jeschkies (approved):

PR #9135, opened 2023-04-13
Notification based Promtail WAL Watcher polling approach

thepalbi:

This PR implements a new mechanism for the wal Watcher in Promtail, to know there are new records to be read. It uses a combination of:
- https://github.com/prometheus/prometheus/pull/11950
- https://github.com/prometheus/prometheus/pull/11949

The main idea is that the primary mechanism is a notification channel between the `wal.Writer` and `wal.Watcher`. The Watcher subscribes to write events the writer publishes, getting notified if the wal has been written. The same subscriptions design is used for cleanup events.

As a backup, the watcher has a timer that implements an exponential backoff strategy, which is constrained by a minimum and maximum that the user can configure.

Below the cpu difference is shown of running both main and this branch against the same scrape target.

image

The yellow line is the latest main build from where this branch started, and the green line is this branch. Both promtails tailing docker logs, and using the following metrics to get cpu usage from cadvisor:
```
avg by (name) (rate(container_cpu_usage_seconds_total{job=~".+", instance=~".+", name=~"promtail-wal-test_promtail.+"}[$__rate_interval]))
```

Part of https://github.com/grafana/loki/issues/8197

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


clients/pkg/promtail/wal/timer.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (bt *backoffTimer) reset() {
+	bt.curr = bt.min
+	bt.recycle()
+}
+
+// recycle stops and attempts to drain the time.Timer underlying channel, in order to fully recycle the instance.
+func (bt *backoffTimer) recycle() {
+	if !bt.timer.Stop() {
```
thepalbi, line 41: similar approach used here https://github.com/valyala/gorpc/blob/master/common.go#L85

clients/pkg/promtail/wal/watcher.go
```diff
@@ -191,53 +195,76 @@ func (w *Watcher) watch(segmentNum int) error {
...
+			level.Warn(w.logger).Log("msg", "Error reading segment inside readTicker", "segment", segmentNum, "read", reader.Offset(), "err", err)
 		}
+
+		// io.EOF error are non-fatal since we are tailing the wal
+		if errors.Cause(err) != io.EOF {
+			return err
+		}
+
+		if ok {
+			// TODO: we'll be resetting the readTimer for each notification we get. Should we do this just if the readSegment
```
thepalbi, line 216: only open doubt, should we differentiate this block when it was triggered by a notification from the writer, or the timer? Maybe not, since the timer is a backup that `attempts to read if a notification hasn't arrived for X time`

cstyan (commented):
I think we're missing a way of removing the client/manager from the list of subscribers when it goes away?

clients/pkg/promtail/wal/writer.go
```diff
@@ -215,16 +242,12 @@ func newEntryWriter() *entryWriter {
...
 	// Reset wal record slices
 	ew.reusableWALRecord.RefEntries = ew.reusableWALRecord.RefEntries[:0]
 	ew.reusableWALRecord.Series = ew.reusableWALRecord.Series[:0]
 
 	defer func() {
-		err := wl.Log(ew.reusableWALRecord)
-		if err != nil {
-			level.Error(logger).Log("msg", "Failed to write entry to wal", "err", err)
-		}
 	}()
```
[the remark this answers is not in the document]
thepalbi: deleting!

clients/pkg/promtail/wal/writer.go
```diff
@@ -27,14 +27,21 @@ const (
...
 )
 
-// WriterEventSubscriber is an interface that objects that want to receive events from the wal Writer can implement. After
-// they can subscribe to events by adding themselves as subscribers on the Writer with writer.Subscribe.
-type WriterEventSubscriber interface {
+// CleanupEventSubscriber is an interface that objects that want to receive events from the wal Writer can implement. After
+// they can subscribe to events by adding themselves as subscribers on the Writer with writer.SubscribeCleanup.
+type CleanupEventSubscriber interface {
 	// WriteCleanup is implemented by WriteTo, who can subscribe to these type of events for cleaning up the series cache
 	// entries that originated from deleted segments.
```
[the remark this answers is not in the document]
thepalbi: yep, agree the commenet should go on the implementor side. Also, the name write to was inspired in the one picked by prometheus. Since in here the WriteTo has become something more of bits and pieces of actions the watcher performs, I'm going with `WatcherActions`. wdyt?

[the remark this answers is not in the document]
thepalbi: I'll defer renaming all `WriteTo` stuff to `WatcherActions` once we agree on a name, since pushing that in this same PR will add a lot of noise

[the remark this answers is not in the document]
thepalbi: > I think we're missing a way of removing the client/manager from the list of subscribers when it goes away?

Do we need such subscriber removal action implemented? The only time a subscriber, be it a watcher or a write to, are stopped, is when Promtail is stopped. This could happen when the config is re-loaded, or it's actually stopped. Since when re-loading the whole `client.Manager` and `wal.Writer` are re-created from scratch, we don't care about removing the subscribers.

thepalbi:
> I think we're missing a way of removing the client/manager from the list of subscribers when it goes away?

Do we need such subscriber removal action implemented? The only time a subscriber, be it a watcher or a write to, are stopped, is when Promtail is stopped. This could happen when the config is re-loaded, or it's actually stopped. Since when re-loading the whole `client.Manager` and `wal.Writer` are re-created from scratch, we don't care about removing the subscribers.

cstyan:
> Since when re-loading the whole client.Manager and wal.Writer are re-created from scratch, we don't care about removing the subscribers.

Got it. I was incorrectly assuming that the Manager struct itself supported reloading, but you're right I just read the code and we create the Manager when reloading the config file.

clients/pkg/promtail/wal/writer.go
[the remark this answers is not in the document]
cstyan: to me `WatcherActions` implies actions you can perform against the Watcher, I think keeping the name as `WriteTo` is okay the point was more that the comment here could become out of date if the interface name changed or the interface was split

if you want a better name, what about something like `WALDataReciever` or just `DataReceiver`? something that makes it clear that it's the interface that's used by whatever is sending data to the Client to be processed and sent on to Loki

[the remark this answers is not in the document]
thepalbi: I prefer for now to stick with with `WriteTo`, since renaming that interface involves a LOT of renames over the `wal` package. Also, I like that is the same used in prometheus, which makes transfer of ideas from one project to the other easier.

clients/pkg/promtail/wal/watcher.go
thepalbi: The main question here is because if we don't differentiate, the timer would be "refreshed" every time we receive a write notification, which could happen often and could lead to a lot of unnecessary calls to the timer funcs that perform the refresh. Maybe worth benchmark + profiling?

wdyt @cstyan ?

cstyan: imo we should, if only just to include debug logging and/or a metric about timer reads

similarly it may also be useful to include a metric about dropped notifications in `NotifyWrite` (due to the channel already being full with it's 1 element)

thepalbi: Good idea! Added both log and a metrics on timer-triggered segment reads, and a metrics on dropped notifications

clients/pkg/promtail/wal/watcher_metrics.go
```diff
@@ -29,6 +31,24 @@ func NewWatcherMetrics(reg prometheus.Registerer) *WatcherMetrics {
...
+		),
+		timerSegmentReads: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: "loki",
+				Subsystem: "wal_watcher",
+				Name:      "read_segment_by_timer",
+				Help:      "Number of segment reads triggered by the backup timer firing.",
+			},
+			[]string{"id"},
+		),
```
cstyan, line 51: We should follow the naming conventions for counter and end the metric names in `_total`, what about `loki_wal_watcher_dropped_write_notifications_total` and `loki_wal_watcher_read__total` with a label `reason="timer"`, then if we want to we can track the same metric for `reason="notification"` in the future without having a breaking change due to metric renaming.

clients/pkg/promtail/wal/watcher.go
```diff
@@ -203,7 +203,7 @@ func (w *Watcher) watch(segmentNum int) error {
 		case <-readTimer.C:
 			w.metrics.timerSegmentReads.WithLabelValues(w.id).Inc()
 			if debug {
-				level.Info(w.logger).Log("msg", "Segment read triggered by backup timer", "segment", segmentNum, "read", reader.Offset())
```
cstyan: should this be `level.Debug`?

clients/pkg/promtail/wal/watcher.go
```diff
@@ -203,7 +203,7 @@ func (w *Watcher) watch(segmentNum int) error {
...
+				level.Info(w.logger).Log("msg", "Segment read triggered by backup timer", "segment", segmentNum)
```
cstyan, line 206: should this be `level.Debug`?

clients/pkg/promtail/wal/watcher_metrics.go
cstyan: sorry, lets add the same metric with `reason="notification"` just to avoid confusion as to why it doesn't exist

thepalbi: done!

cstyan (approved):

PR #9146, opened 2023-04-14
Revert "Use centralized make-docs script from Writers' Toolkit (#9126)"

cstyan:
This reverts commit df1b34898bea812b197ca7404b29094f740df26c.

Sorry @jdbaldry but something in your PR made the helm-docs check fail, and only fail in CI, I tried a few things to figure out what was happening locally but the same make target always passed locally.

cc @trevorwhitney

PR #9147, opened 2023-04-14
Bring back centralized docs toolkit (undo revert in #9146)

trevorwhitney:

This reverts the revert in https://github.com/grafana/loki/pull/9146 with a slightly less "nuclear" solution.

I think the problem was that CI was pulling `docs.mk` from `main`, and it was creating a diff between the version we had checked in. I changed `docs.mk` to point to a specific sha, which should prevent this in the future.

I also opened a [PR](https://github.com/grafana/writers-toolkit/pull/228) against the writers-toolkit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ pkg/loki/wal
 
 # nix
 result
+/docs/make-docs
--- a/.gitignore
+++ b/.gitignore
@@ -51,4 +51,3 @@ pkg/loki/wal
 
 # nix
 result
-/docs/make-docs
--- a/docs/docs.mk
+++ b/docs/docs.mk
@@ -1,0 +1,87 @@
+include variables.mk
+-include variables.mk.local
+
+.ONESHELL:
+.DELETE_ON_ERROR:
+export SHELL     := bash
+export SHELLOPTS := pipefail:errexit
+MAKEFLAGS += --warn-undefined-variables
+MAKEFLAGS += --no-builtin-rule
+
+.DEFAULT_GOAL: help
+
+# Adapted from https://www.thapaliya.com/en/writings/well-documented-makefiles/
+.PHONY: help
+help: ## Display this help.
+help:
+	@awk 'BEGIN {FS = ": ##"; printf "Usage:\n  make <target>\n\nTargets:\n"} /^[a-zA-Z0-9_\.\-\/%]+: ##/ { printf "  %-45s %s\n", $$1, $$2 }' $(MAKEFILE_LIST)
+
+GIT_ROOT := $(shell git rev-parse --show-toplevel)
+
+PODMAN := $(shell if command -v podman >/dev/null 2>&1; then echo podman; else echo docker; fi)
+
+ifeq ($(PROJECTS),)
+$(error "PROJECTS variable must be defined in variables.mk")
+endif
+
+# First project is considered the primary one used for doc-validator.
+PRIMARY_PROJECT := $(firstword $(subst /,-,$(PROJECTS)))
+
+# Name for the container.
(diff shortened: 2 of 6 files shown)


docs/docs.mk
```diff
@@ -0,0 +1,100 @@
...
+doc-validator: ## Run docs-validator on the entire docs folder.
+	DOCS_IMAGE=$(DOC_VALIDATOR_IMAGE) $(PWD)/make-docs $(PROJECTS)
+
+.PHONY: doc-validator/%
+doc-validator/%: ## Run doc-validator on a specific path. To lint the path /docs/sources/administration, run 'make doc-validator/administration'.
+doc-validator/%:
+	DOCS_IMAGE=$(DOC_VALIDATOR_IMAGE) DOC_VALIDATOR_INCLUDE=$(subst doc-validator/,,$@) $(PWD)/make-docs $(PROJECTS)
+
+docs.mk: ## Fetch the latest version of this Makefile from Writers' Toolkit.
+	curl -s -LO https://raw.githubusercontent.com/grafana/writers-toolkit/fd179ba615cbaf8cf738e12fe34a951b505ce5a1/docs/docs.mk
```
trevorwhitney, line 99: this is the key difference

trevorwhitney:
Depends on https://github.com/grafana/writers-toolkit/pull/228

docs/docs.mk
cstyan: I think this is the correct change, @jdbaldry can you confirm? thank you :)

jdbaldry:
Sorry for the inconvenience of the change. The idea was it was meant to be a straight drop in behavior-wise.

I guess it only failed in later runs because the file changed in Writers' Toolkit.
I'd like to keep a target that updates the `docs.mk` script conveniently but I can see that with `make -B` it's going to be a nuisance to call out to the network in CI every time the source file is updated.

I'd like to propose that the `docs.mk` target is renamed and made PHONY so that it is no longer implicitly invoked every time someone runs `make -B`. Something like `make update` would do what `make docs.mk` does currently.

Additionally, perhaps we make the `git diff` CI check more specific in https://github.com/grafana/loki/blob/main/Makefile#L626-L629? Shouldn't it only really care about the generated file `docs/sources/installation/helm/reference.md`?

JStickler:
@trevorwhitney are you aware of #9126?  Does it impact this PR?

jdbaldry:
> I'd like to propose that the docs.mk target is renamed and made PHONY so that it is no longer implicitly invoked every time someone runs make -B. Something like make update would do what make docs.mk does currently.

Unfortunately, this introduces a lot of complications since the `make-docs` script isn't committed itself. If we are happy with the script being committed too, I can introduce a PHONY target.

I still think the `git diff` should be restricted to just the generated file that is being checked.

jdbaldry:
I've opened https://github.com/grafana/writers-toolkit/pull/232

trevorwhitney:
@jdbaldry I've modified this PR to:
* temporarily rename `docs.mk` -> `update` and make phony until https://github.com/grafana/writers-toolkit/pull/232 is merged
* make the diff check in the `documentation-helm-reference-check` target more specific

I'm good with a phony update target moving forward. In the meantime this should undo the revert and keep us moving.

cc @cstyan

jdbaldry:
Thank you @trevorwhitney, that looks good to me!

.gitignore
```diff
@@ -51,3 +51,4 @@ pkg/loki/wal
 
 # nix
 result
+/docs/make-docs
```
jdbaldry, line 54: To avoid network calls in CI to fetch this file, I think we should remove the ignore and commit this file. WDYT?

MasslessParticle (approved):

PR #9156, opened 2023-04-17
expiration: do not drop index if period is a zero value

ashwanthgoli:

`DropFromIndex` should be return false when `retention_period` is set to a zero value.

If we take chunks that overlap period boundary and are indexed in two tables, we drop these from the table on the left side of the period boundary even if they have not expired - https://github.com/grafana/loki/blob/ashwanth/drop-from-index-zero-value/pkg/storage/stores/indexshipper/compactor/retention/retention.go#L218

**Checklist**
- [X] Tests updated
- [x] `CHANGELOG.md` updated


sandeepsukhani (approved):
thanks for finding and fixing the issue! LGTM

slim-bean:
Great catch @ashwanthgoli, I recently changed the zero value behavior for this setting in #8753, it seems like I missed this?

This seems like a pretty big error on my part, does this result in deleting data when the retention_period is set to zero?

PR #9175, opened 2023-04-18
Update prepare shutdown

MichelHollands:

This updates the PrepareShutdown method so it supports GET and DELETE methods as well. This makes it similar to Mimir: https://github.com/grafana/mimir/pull/4718. 
The status is now stored in a local file. A new config setting had to be added for this file as there is no obvious place to store it.

**Checklist**
- [X] Documentation added
- [X] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -1476,6 +1476,10 @@ wal:
 # Maximum number of dropped streams to keep in memory during tailing.
 # CLI flag: -ingester.tailer.max-dropped-streams
 [max_dropped_streams: <int> | default = 10]
+
+# Path where the shutdown marker file is stored
+# CLI flag: -ingester.shutdown_marker_path
+[shutdown_marker_path: <string> | default = ""]
 ```
 
 ### index_gateway
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -1477,7 +1477,8 @@ wal:
 # CLI flag: -ingester.tailer.max-dropped-streams
 [max_dropped_streams: <int> | default = 10]
 
-# Path where the shutdown marker file is stored
+# Path where the shutdown marker file is stored. If not set and
+# common.path_prefix is set then common.path_prefix will be used.
 # CLI flag: -ingester.shutdown-marker-path
 [shutdown_marker_path: <string> | default = ""]
 ```
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -1478,7 +1478,7 @@ wal:
 [max_dropped_streams: <int> | default = 10]
 
 # Path where the shutdown marker file is stored
-# CLI flag: -ingester.shutdown_marker_path
+# CLI flag: -ingester.shutdown-marker-path
 [shutdown_marker_path: <string> | default = ""]
 ```
 
--- a/docs/sources/upgrading/_index.md
+++ b/docs/sources/upgrading/_index.md
@@ -33,6 +33,14 @@ The output is incredibly verbose as it shows the entire internal config struct u
 
 ## Main / Unreleased
 
+### Loki
+
+#### Shutdown marker file
+
+A shutdown marker file can be written by the `/ingester/prepare_shutdown` endpoint.
+If the new `ingester.shutdown_marker_path` config setting has a value that value is used.
+If not the`common.path_prefix` config setting is used if it has a value. Otherwise an error is returned on startup.
+
 ## 2.8.0
 
 ### Loki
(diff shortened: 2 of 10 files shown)


JStickler (approved):
[docs team] LGTM

pkg/ingester/ingester.go
```diff
@@ -574,13 +591,140 @@ func (i *Ingester) LegacyShutdownHandler(w http.ResponseWriter, r *http.Request)
...
+func removeShutdownMarker(p string) error {
+	err := os.Remove(p)
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	dir, err := os.OpenFile(path.Dir(p), os.O_RDONLY, 0777)
+	if err != nil {
+		return err
+	}
```
DylanGuedes, line 708: could you explain why would you have to cleanup the whole dir the shutdown marker was already removed?

DylanGuedes (changes requested):
looks pretty solid, gj. I'm blocking the PR because of the flag suggestion (from using underscore to dashes) because if you merge this way and we decide to change it back it would be a breaking change.

JordanRushing (approved):
LGTM!

docs/sources/configuration/_index.md
```diff
@@ -1476,6 +1476,10 @@ wal:
 # Maximum number of dropped streams to keep in memory during tailing.
 # CLI flag: -ingester.tailer.max-dropped-streams
 [max_dropped_streams: <int> | default = 10]
+
+# Path where the shutdown marker file is stored
```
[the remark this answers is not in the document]
MichelHollands: Good idea. An extra line was added to the description.

pkg/ingester/ingester.go
MichelHollands: The variable p contains the full shutdownMarkerPath, including the filename. So this won't remove the dir, only the file.
The code starting on 705 opens the directory and syncs it. That way the we're sure the changes are on disk.
Note that this code was copied from Mimir.

pkg/loki/modules.go
```diff
@@ -450,6 +450,13 @@ func (t *Loki) initQuerier() (services.Service, error) {
 func (t *Loki) initIngester() (_ services.Service, err error) {
 	t.Cfg.Ingester.LifecyclerConfig.ListenPort = t.Cfg.Server.GRPCListenPort
 
+	if t.Cfg.Ingester.ShutdownMarkerPath == "" && t.Cfg.Common.PathPrefix != "" {
```
DylanGuedes, line 453: could you check if this works with SSD?

DylanGuedes (changes requested):
one more :smile:

pkg/loki/modules.go
```diff
@@ -450,6 +450,13 @@ func (t *Loki) initQuerier() (services.Service, error) {
...
+		t.Cfg.Ingester.ShutdownMarkerPath = t.Cfg.Common.PathPrefix
+	}
+	if t.Cfg.Ingester.ShutdownMarkerPath == "" {
```
DylanGuedes, line 456: By having this check we'll break all Loki installations that aren't specifying a `common_prefix`. Since this is a minor version we'd be giving a better experience to users by avoiding that.

Suggestion: Log a warning here and move this check to the `PrepareShutdown` endpoint. Panic'ing on `PrepareShutdown` is better imo because a user using it knows what to expect; won't be surprised by the new marker change.

DylanGuedes (changes requested):

pkg/loki/modules.go
MichelHollands: That is a very good suggestion. It has been implemented like that.

pkg/loki/modules.go
MichelHollands: Yes it works with the `write` path of SSD. I get the 500 error when the paths are not set. The warning also shows up in the logs in that case. With common.path_prefix set it works as expected.

DylanGuedes (dismissed):
lgtm

docs/sources/upgrading/_index.md
```diff
@@ -33,6 +33,14 @@ The output is incredibly verbose as it shows the entire internal config struct u
 
 ## Main / Unreleased
 
+### Loki
+
+#### Shutdown marker file
+
+A shutdown marker file can be written by the `/ingester/prepare_shutdown` endpoint.
+If the new `ingester.shutdown_marker_path` config setting has a value that value is used.
```
JordanRushing, line 41:
```suggestion
If the new `ingester.shutdown-marker-path` config setting has a value that value is used.
```

docs/sources/upgrading/_index.md
```diff
@@ -33,6 +33,14 @@ The output is incredibly verbose as it shows the entire internal config struct u
...
+If not the`common.path_prefix` config setting is used if it has a value. Otherwise an error is returned on startup.
```
JordanRushing, line 42:
```suggestion
If not the`common.path-prefix` config setting is used if it has a value. Otherwise an error is returned on startup.
```

docs/sources/upgrading/_index.md
DylanGuedes: the config setting is `shutdown_marker_path`, `shutdown-marker-path` is its flag counterpart

PR #9176, opened 2023-04-18
Loki: Fix wrong association of PSRL when sharding is enabled

DylanGuedes:
Obs: PSRL stands for `per-stream rate limit`
- Fixes the scenario where all entries errors are accounted as PSRL if sharding is enabled
- Fixes invalid/rejected entries being accounted as candidates to be inserted
Scenario to illustrate:
- Let's say a push request has two entries for the same stream, one entry being invalid/super old, and another one being a valid one. Two issues will occur: 1) the two will be accounted to check if they're hitting the psrl or not; 2) the two will be rejected with PSRL error

### DIFF
--- a/pkg/ingester/stream.go
+++ b/pkg/ingester/stream.go
@@ -394,13 +394,13 @@ func (s *stream) validateEntries(entries []logproto.Entry, isReplay, rateLimitWh
 	// ingestion, the limiter should only be advanced when the whole stream can be
 	// sent
 	now := time.Now()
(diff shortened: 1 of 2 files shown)


chaudum (approved):
This makes sense. I quickly checked that `toStore` is the slice of entries of the stream that will be stored in case the stream is not rate limited, right?

DylanGuedes:
> This makes sense. I quickly checked that `toStore` is the slice of entries of the stream that will be stored in case the stream is not rate limited, right?

exactly it, thank you.

Issue #9149, opened 2023-04-15
Redis client address configured as "localhost" can incorrectly set cluster mode


dannykopping:
**Describe the bug**
When multiple addresses are given to the redis client, it activates "cluster mode" whereby it will read/write to multiple instances.

The hostname `localhost` can be resolved to two addresses:
- `127.0.0.1` (IPv4)
- `::1` (IPv6)

**To Reproduce**

Configure `localhost:<port>` as the redis address, which results in the following confusing error message:

> `ERR This instance has cluster support disabled`

**Expected behavior**
`localhost` should be resolved to a single address, and for cluster mode to not be activated.

PR #9185, opened 2023-04-19
Prevent redis client from incorrectly choosing cluster mode with local address

dannykopping:
See referenced issue below for details

Fixes https://github.com/grafana/loki/issues/9149

This is the challenge of using magic in our code; it produces unexpected results sometimes which are hard to debug. In general we should prefer being explicit, but we cannot revert this functionality now because it may be in use. I think this (automatically deriving multiple endpoints from a single one) is a good candidate for a deprecation, and ultimately removal in v3. Thoughts?

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,22 @@
 * [8972](https://github.com/grafana/loki/pull/8972) **salvacorts** Index stat requests are now cached in the results cache.
 * [9177](https://github.com/grafana/loki/pull/9177) **salvacorts** Index stat cache can be enabled or disabled with the new `cache_index_stats_results` flag. Disabled by default.
 * [9096](https://github.com/grafana/loki/pull/9096) **salvacorts**: Compute proportional TSDB index stats for chunks that doesn't fit fully in the queried time range.
+
+##### Fixes
+
+* [8971](https://github.com/grafana/loki/pull/8971) **dannykopping**: Stats: fix `Cache.Chunk.BytesSent` statistic and loki_chunk_fetcher_fetched_size_bytes metric with correct chunk size.
+* [8979](https://github.com/grafana/loki/pull/8979) **slim-bean**: Fix the case where a logs query with start time == end time was returning logs when none should be returned.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,22 @@
 * [8972](https://github.com/grafana/loki/pull/8972) **salvacorts** Index stat requests are now cached in the results cache.
 * [9177](https://github.com/grafana/loki/pull/9177) **salvacorts** Index stat cache can be enabled or disabled with the new `cache_index_stats_results` flag. Disabled by default.
 * [9096](https://github.com/grafana/loki/pull/9096) **salvacorts**: Compute proportional TSDB index stats for chunks that doesn't fit fully in the queried time range.
+
+##### Fixes
```
ashwanthgoli, line 16: changelog is mixed up, I see the same commits further down in the file

pkg/storage/chunk/cache/redis_client.go
```diff
@@ -96,6 +83,49 @@ func NewRedisClient(cfg *RedisConfig) (*RedisClient, error) {
...
+func deriveEndpoints(endpoint string, lookup func(host string) ([]string, error)) ([]string, error) {
+	if lookup == nil {
+		return nil, fmt.Errorf("lookup function is nil")
+	}
+
+	endpoints := strings.Split(endpoint, ",")
+	if len(endpoints) < 1 {
+		return endpoints, nil
+	}
+
```
ashwanthgoli, line 95: should we also return early if `len(endpoints) > 1`
I am assuming the special case handling below is only when a single endpoint is provided?

CHANGELOG.md
dannykopping: Thanks! I'll fix

pkg/storage/chunk/cache/redis_client.go
dannykopping: That's a good call

dannykopping (commented):
@ashwanthgoli thanks for the comments. I've pushed changes

ashwanthgoli (approved):
LGTM

PR #9191, opened 2023-04-19
Logcli - Do not fail if tenant specific schema doesn't exist

salvacorts:
This is how we currently load a remote schema from the object store:
https://github.com/grafana/loki/blob/2f8d771c6b4fe5f659535c1f62df3dc6a7746e60/pkg/logcli/query/query.go#L441-L448

`LoadSchemaUsingObjectClient` ([here](https://github.com/grafana/loki/blob/2f8d771c6b4fe5f659535c1f62df3dc6a7746e60/pkg/logcli/query/query.go#L534)) will return an error if any of the objects trying to download doesn’t exist. Therefore, if the tenant-specific schema file doesn't exist, `LoadSchemaUsingObjectClient`will return an error and we won't try to use the global schema config even if it exists. In other words, the backward compatibility won't work.

This PR fixes this by first trying to download the tenant-specific schema config and if it doesn't exist, fallback to the global one for backwards compatibility.

This PR also improves the error messaging when an object cannot be downloaded from the store by printing the name of the object along with the error.

Internal support escalation

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logcli/query/query.go
+++ b/pkg/logcli/query/query.go
@@ -444,7 +444,7 @@ func (q *Query) DoLocalQuery(out output.LogOutput, statistics bool, orgID string
 		}
 		loadedSchema, err := LoadSchemaUsingObjectClient(client, objects...)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to load schema config: %w", err)
 		}
 
 		conf.SchemaConfig = *loadedSchema
@@ -538,7 +538,7 @@ func LoadSchemaUsingObjectClient(oc chunk.ObjectClient, names ...string) (*confi
 			defer cancel()
 			rdr, _, err := oc.GetObject(ctx, name)
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("failed to load object '%s': %w", name, err)
 			}
 			defer rdr.Close()
 
@@ -547,7 +547,7 @@ func LoadSchemaUsingObjectClient(oc chunk.ObjectClient, names ...string) (*confi
 			section := schemaConfigSection{}
 			err = decoder.Decode(&section)
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("failed to decode downloaded object '%s': %w", name, err)
 			}
 
 			return &section.SchemaConfig, nil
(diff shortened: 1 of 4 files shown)


pkg/storage/chunk/client/object_client.go
```diff
@@ -165,7 +166,7 @@ func (o *client) getChunk(ctx context.Context, decodeContext *chunk.DecodeContex
 
 	readCloser, size, err := o.store.GetObject(ctx, key)
 	if err != nil {
-		return chunk.Chunk{}, errors.WithStack(err)
+		return chunk.Chunk{}, errors.WithStack(fmt.Errorf("failed to load object '%s': %w", key, err))
```
MichelHollands, line 169: Can't you make this a separate error and check for it in `loadRemoteSchema`? Relying on the text of the error is not great.

pkg/logcli/query/query.go
```diff
@@ -420,6 +420,27 @@ func (q *Query) printResult(value loghttp.ResultValue, out output.LogOutput, las
...
+func loadRemoteSchema(client chunk.ObjectClient, tenant string) (*config.SchemaConfig, error) {
+	tenantSchemaConfigFilename := fmt.Sprintf("%s-%s.yaml", tenant, schemaConfigFilename)
+	globalSchemaConfigFilename := fmt.Sprintf("%s.yaml", schemaConfigFilename)
+	loadedSchema, err := LoadSchemaUsingObjectClient(client, tenantSchemaConfigFilename)
+	if err != nil {
+		if !strings.Contains(err.Error(), "object doesn't exist") {
+			return nil, fmt.Errorf("failed to load schema config: %w", err)
+		}
+
+		log.Printf("failed to load tenant specific schema config (%s). Falling back to load global schema", tenantSchemaConfigFilename)
```
MichelHollands, line 434:
```suggestion
		log.Printf("failed to load tenant specific schema config (%s). Falling back to the global schema", tenantSchemaConfigFilename)
```
It's probably better to print the global schema config file as well.

PR #9198, opened 2023-04-19
tsdb: add a tool for migrating tsdb index version

sandeepsukhani:
We are adding a new tsdb version. This tool would provide a way to recover from any issues if the new version has any problems.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


sandeepsukhani:
closing in favour of #9275

PR #9259, opened 2023-04-24
mark helm as the recommended installation method

MasslessParticle:
AFAIK, we're recommending folks use the helm chart to install these days. Make the installation documentation reflect that


MichelHollands (approved):
LGTM

PR #9265, opened 2023-04-24
Distributor: Modify `createShard` to create slice with more precise size

DylanGuedes:
Modify the `createShard` method to create the entries slice with the number of entries divided by the total number of shards instead of always creating it with len=1024.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -479,6 +479,11 @@ func (d *Distributor) createShards(stream logproto.Stream, totalShards int, tena
 		streamCount = streamCount(totalShards, stream)
 	)
 
+	if totalShards <= 0 {
+		level.Error(util_log.Logger).Log("msg", "attempt to create shard with zeroed total shards")
+		return derivedKeys, derivedStreams
+	}
+
 	startShard := d.shardTracker.LastShardNum(tenantID, stream.Hash)
 	for i := 0; i < streamCount; i++ {
 		shardNum := (startShard + i) % totalShards
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -480,7 +480,7 @@ func (d *Distributor) createShards(stream logproto.Stream, totalShards int, tena
 	)
 
 	if totalShards <= 0 {
-		level.Error(util_log.Logger).Log("msg", "attempt to create shard with zeroed total shards")
+		level.Error(util_log.Logger).Log("msg", "attempt to create shard with zeroed total shards", "org_id", tenantID, "stream", stream.Labels, "entries_len", len(stream.Entries))
 		return derivedKeys, derivedStreams
 	}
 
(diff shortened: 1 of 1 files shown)


DylanGuedes:
Profiling of the CPU usage inherited from creating this slice:
image

pkg/distributor/distributor.go
```diff
@@ -482,7 +482,7 @@ func (d *Distributor) createShards(stream logproto.Stream, totalShards int, tena
 	startShard := d.shardTracker.LastShardNum(tenantID, stream.Hash)
 	for i := 0; i < streamCount; i++ {
 		shardNum := (startShard + i) % totalShards
-		shard := d.createShard(streamLabels, streamPattern, shardNum)
+		shard := d.createShard(streamLabels, streamPattern, shardNum, len(stream.Entries)/totalShards)
```
dannykopping, line 485: Whenever dividing by a number, you should always check that `int > 0`

dannykopping: You're talking about a _loose_ check here because the check is done elsewhere before this code is run, right?
If we refactor the code later, that loose association might be broken.

In cases like this where a `panic` could result, it can't hurt to be extra careful.

DylanGuedes: I pushed a commit adding the check to `createShards` (to not check if for every call to `createShard`) and logging an error if the problem is spotted. WDYT?

DylanGuedes: I could also just do `max(totalShards, 1)` but I'd really like to log an error in such scenario because totalShards=0 shouldn't ever happen so logging it will be useful spotting a bigger problem

pkg/distributor/distributor.go
```diff
@@ -479,6 +479,11 @@ func (d *Distributor) createShards(stream logproto.Stream, totalShards int, tena
 		streamCount = streamCount(totalShards, stream)
 	)
 
+	if totalShards <= 0 {
+		level.Error(util_log.Logger).Log("msg", "attempt to create shard with zeroed total shards")
```
dannykopping, line 483: Can you walk me through how this log message would help you isolate the problem in production?

DylanGuedes: I'd know that somehow we're calling `createShards` with zero totalShards. By knowing that I could
- Isolate if it was introduced by a recent version (in case the log wasn't present before)
- Knowing that totalShards is zero is enough to start a deeper investigation of how it happened and I'd then enable verbose logs for this etc

dannykopping: > Isolate if it was introduced by a recent version (in case the log wasn't present before)
What if we spot this only in 3 months? I'm not sure the log line contains enough context. e.g. which tenant? which stream? other shard details?

> I'd then enable verbose logs for this etc
What do you mean by this?

dannykopping: OK cool. Last question: should this be an alert? Would we need to know that this happened? What's the impact?

dannykopping (approved):
LGTM, thanks for your patience with my grilling 🙂

DylanGuedes:
> LGTM, thanks for your patience with my grilling slightly_smiling_face

np, love them :smile:

PR #9270, opened 2023-04-24
Add Prometheus Link for scrape_configs

BlakeFernandez:
Adds some context to the scrape configs section in Promtail Config Docs to ensure clarity around the connection to Prometheus 
Here is a link to a slack thread where the customer was confused on this: https://raintank-corp.slack.com/archives/C04KY6NJSCE/p1681761335754199


docs/sources/clients/promtail/configuration.md
```diff
@@ -356,7 +356,7 @@ is restarted to allow it to continue from where it left off.
 ## scrape_configs
 
 The `scrape_configs` block configures how Promtail can scrape logs from a series
-of targets using a specified discovery method:
+of targets using a specified discovery method. Promtail uses the same [Prometheus scrape_configs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config). This means if you already own a Prometheus instance, the config will be very similar and easy to grasp:
```
JStickler, line 359: As a general rule, avoid saying that things are "easy" in documentation as this often frustrates users even more.  I think it's enough here to say that the config is similar.
```suggestion
of targets using a specified discovery method. Promtail uses the same [Prometheus scrape_configs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config). This means if you already own a Prometheus instance, the config will be very similar:
```

MichelHollands (approved):

PR #9287, opened 2023-04-25
Update centralized make-docs procedure

jdbaldry:
Supports Ubuntu Snap based Docker installations by using TMPDIR instead of an explicit tempfile. 

Because Ubuntu Snap packages have access to only the user's home directory, users must specify something like `TMPDIR="$(mktemp -d -q ~/.make-docs.XXXXX)"` in order to be able to run `make docs`.

### DIFF
--- a/docs/make-docs
+++ b/docs/make-docs
@@ -1,5 +1,5 @@
 #!/bin/sh
-# Source of this file is https://github.com/grafana/technical-documentation/scripts/make-docs.
+# The source of this file is https://raw.githubusercontent.com/grafana/writers-toolkit/main/docs/make-docs.
 
 set -ef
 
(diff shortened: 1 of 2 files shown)


JStickler (approved):
[docs team] LGTM

docs/make-docs
```diff
@@ -343,6 +352,7 @@ POSIX_HERESTRING
     --name "${DOCS_CONTAINER}" \
     --platform linux/amd64 \
     --publish "${DOCS_HOST_PORT}:3002" \
+    --publish "3003:3003" \
```
DylanGuedes, line 355: where is the "3003" from?

DylanGuedes (approved):
lgtm, just a question

docs/make-docs
jdbaldry: Good question,

I promise for future updates there will be a full CHANGELOG to link to now that the procedure is somewhat stable. 

The 3003 port is actually where Hugo listens in the `grafana/docs-base` container image. On 3002 there is an image proxy that handles the fetching of images from object storage (removing the need to keep them in the repo). More information is in https://grafana.com/docs/writers-toolkit/writing-guide/image-guidelines/. During local builds, Hugo injects some JavaScript that uses a websocket connection to the Hugo server process to auto-reload pages when changes are made to the underlying files. Exposing port 3003 was a quick fix to re-enable that behavior.

kavirajk (approved):
LGTM!

PR #9290, opened 2023-04-25
Add docs/migration guide for zone aware ingesters

cstyan:
This is migration guide is based on the internal migration guide we created when migrating our own environments to zone aware ingesters.

We should probably structure some config documentation similar to https://grafana.com/docs/mimir/latest/operators-guide/configure/configure-zone-aware-replication/ as well


cstyan:
@JStickler do you mind taking a quick look here? thank you :)

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
+---
+title: Zone Aware Ingesters
```
JStickler, line 2:
```suggestion
title: Zone aware ingesters
```

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
...
+description: Migration guide and additional details
+weight: 10
+---
+
+# Zone Aware Ingesters
```
JStickler, line 7: To conform with the Style Guide
```suggestion
# Zone aware ingesters
```

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
+---
+title: Zone Aware Ingesters
+description: Migration guide and additional details
```
JStickler, line 3: This is the social media/Google search preview text.  Could we be more specific about what we're migrating to help users know if this is the link they're searching for?

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
...
+weight: 10
+---
+
+# Zone Aware Ingesters
+Loki's zone aware ingesters are used by Grafana Labs in order to allow for easier rollouts of large Loki deployments. You can think of them as three logical zones, however with some extra k8s config you could deploy them in separate zones.
+
+By default, an incomming log stream's logs are replicated to 3 random (but the same 3, except in the case of some replica scaling) ingesters. This means that if one ingester is restarted, no data is lost, but two can mean data is lost and also impacts the systems ability to ingest logs because of an unhealthy ring status.
```
JStickler, line 10: Would saying "three randomly assigned ingesters" capture what you're saying in parentheses?  Also I feel like there's a word missing in the second sentence?

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
...
+
+With zone awareness enabled, an incomming log line will be replicated to one ingester in each zone. This means that we're no only concerned if ingesters in multiple zones restart at the same time. We can now rollout, or lose, an entire zone at once and not impact the system. This allows deployments with a large number of ingesters to be deployed too much more quickly.
```
JStickler, line 12: no only => no longer?

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
...
+
+We also make use of [rollout-operator](https://github.com/grafana/rollout-operator) to manage rollouts to the 3 statefulsets gracefully. The rollout-operator looks for labels on statefulsets to know which statefulsets are part of a certain rollout group, and coordinate rollouts of pods only from a single statefulset in the group at a time. See the README in the rollout-operator repo. for a more in depth explanation.
```
JStickler, line 14: FYI - K8 documentation uses StatefulSet (Pascal case)

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
...
+0. configure your existing ingesters to be part of a zone, for example `zone-default`, this will allow us to later exclude them from the write path while still allowing for graceful shutdown
+1. Prep for the increase in active streams (due to the way streams are split between ingesters) by increasing the # of active streams allowed for your tenants
+2. add and scale up your new zone aware ingester statefulsets such that each has 1/3 of the total # of replicas you want to run
+3. enable zone awareness on the write path by setting `distributor.zone-awareness-enabled` to true for distributors and rulers
+4. wait some time to ensure that the new zone aware ingesters have data for the time period they are queried for (`query_ingesters_within`)
+5. enable zone awareness on the read path by setting `distributor.zone-awareness-enabled` to true for queriers
+6. configure distributors and rulers to exclude ingesters in the `zone-default` so those ingesters no longer receive write traffic via `distributor.excluded-zones`
+7. use the shutdown endpoint to flush data from the default ingesters, then scale down and remove the associated statefulset
+8. clean up
```
JStickler, line 26: This list could use a few more specific details in the steps, and there's some cleanup to be done re: capitalization and punctuation.

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
...
+    // These last two lines are necessary now that we enable zone aware ingester by default
+    // so that newly created cells will not be migrated later on. If you miss them you will
+    // break writes in the cell.
+    multi_zone_ingester_replication_write_path_enabled: false,
+    multi_zone_ingester_replication_read_path_enabled: false,
+  },
+```
+
+If you're not using jsonnet the new ingester statefulsets should have a label with `rollout-group: ingester`, annotation `rollout-max-unavailable: x` (put a placeholder value in, later you should set the value of this to be some portion of the statefulsets total replicas, for example in jsonnet we template this so that each statefulset runs 1/3 of the total replicas and the max unavailable is 1/3 of each statefulsets replicas), and set the update strategy to `OnDelete`.
+2.  Diff ingester and ingester-zone-a StatefulSets and make sure all config matches
```
JStickler, line 50: Step 2 again?  (If you number every step as `1.` you avoid this problem....)

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
...
+    multi_zone_ingester_replicas: 0,
+    // These last two lines are necessary now that we enable zone aware ingester by default
+    // so that newly created cells will not be migrated later on. If you miss them you will
+    // break writes in the cell.
+    multi_zone_ingester_replication_write_path_enabled: false,
+    multi_zone_ingester_replication_read_path_enabled: false,
+  },
+```
+
+If you're not using jsonnet the new ingester statefulsets should have a label with `rollout-group: ingester`, annotation `rollout-max-unavailable: x` (put a placeholder value in, later you should set the value of this to be some portion of the statefulsets total replicas, for example in jsonnet we template this so that each statefulset runs 1/3 of the total replicas and the max unavailable is 1/3 of each statefulsets replicas), and set the update strategy to `OnDelete`.
```
JStickler, line 49: "If you're not using jsonnet"  but this is under the jsonnet heading?

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
...
+2.  Diff ingester and ingester-zone-a StatefulSets and make sure all config matches 
+```
+kubectl get statefulset -n loki-dev-008 ingester -o yaml > ingester.yaml
+kubectl get statefulset -n loki-dev-008 ingester-zone-a -o yaml > ingester-zone-a.yaml
+diff ingester.yaml ingester-zone-a.yaml
+```
+expected diffs are things like: creation time and revision #, the zone, fields used by rollout operator, # of replicas, anything related to kustomize/flux, and PVC for the wal since the containers don't exist yet
+3. Temporarily double max series limits for users that are using more than 50% of their current limit, the queries are as follows (add label selectors as appropriate):
```
JStickler, line 57: Step 3 is not on it's own line in the preview, I think because step 2 is missing a period at the end?

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
...
+
+```
+(sum (loki_ingester_memory_streams) by (cluster, namespace, tenant) / on (namespace) group_left max by(namespace) (loki_distributor_replication_factor)
+) / ignoring(tenant) group_left max by (cluster, namespace)(loki_overrides_defaults{limit_name="max_global_streams_per_user"}) > 0.4)
+unless on (tenant) (
+(label_replace(loki_overrides{limit_name="max_global_streams_per_user"},"tenant", "$1", "user", "(.+)")))
+```
+4. Scale up zone-aware StatefulSets until they have ⅓ of replicas each. In small cells you can do this all at once, in larger cells it might be safer to do it in chunks. The config value you need to change is `multi_zone_ingester_replicas: 6`, the value will be split across the three statefulsets. So in this case each statefulset would run 2 replicas.
+   
+   If you're not using jsonnet this is the step where you would also set the annotation `rollout-max-unavailable` to some value that is less than or equal to the # of replicas each statefulset is running.
```
JStickler, line 74: "If you're not using jsonnet" again, this is under the jsonnet heading?  Might need a separate procedure for non-jsonnet migration steps?

docs/sources/operations/zone-ingesters.md
```diff
@@ -0,0 +1,96 @@
...
+```
+   rule_evaluations_total
+   rule_evaluation_failures_total
+   rule_group_iterations_missed_total
+```
+
+8. Enable zone-aware replication on the read path `multi_zone_ingester_replication_read_path_enabled: true` or if you're not using jsonnet set `distributor.zone-awareness-enabled` to true for queriers.
+9. Check that queries are still executing correctly, for example look at `loki_logql_querystats_latency_seconds_count` to see that you don't have a big increase in latency or error count for a specific query type
+10. Configure distributor / ruler to exclude ingesters in the “zone-default” so those ingesters no longer receive write traffic `multi_zone_ingester_exclude_default: true` or if you're not using jsonnet set `distributor.excluded-zones` on distributors and rulers
+
```
JStickler, line 89: Steps after this point are displaying as a paragraph, so there's some kind of formatting error here.

JStickler (commented):
[docs team]  Just some quick comments on the WIP draft.  Feel free to ping me for a meeting if it would be faster to just have a quick meeting to discuss.

docs/sources/operations/zone-ingesters.md
cstyan: Here's what I was trying to say, in too many words, open to suggestions how to phrase it differently.

> By default, an incoming log stream's logs are replicated to 3 random ingesters. Except in the case of some replica scaling up or down, a given stream will always be replicated to the same 3 ingesters. This means that if one of those ingesters is restarted no data is lost, but two restarting can mean data is lost and also impacts the systems ability to ingest logs because of an unhealthy ring status.

docs/sources/operations/zone-ingesters.md
cstyan: why would we lowercase here?

docs/sources/operations/zone-ingesters.md
cstyan: How much detail should we put? Are search results not the combination of the title and the description?

`Zone aware ingester migration guide and additional details`?

docs/sources/operations/zone-ingesters.md
cstyan: should be `not only`, fixed

docs/sources/operations/zone-ingesters.md
cstyan: This is the "general steps" section, so that people can do a migration without using the deployment methodologies we provide code (jsonnet) for.

docs/sources/operations/zone-ingesters.md
cstyan: I've changed the heading

docs/sources/operations/zone-ingesters.md
cstyan: formatting code blocks in lists is always annoying, I think I fixed it though

docs/sources/operations/zone-ingesters.md
JStickler: Because we've standardized on sentence case as our house Style (and I haven't gotten a chance to go through and standardize all the headings yet).

https://grafana.com/docs/writers-toolkit/style-guide/capitalization-punctuation/#sentence-case-capitalization

docs/sources/operations/zone-ingesters.md
JStickler: Sadly I recently learned that our internal search on the Grafana docs site displays ONLY the description metadata in the search results.  This is another place where I need to take a pass through the documentation and add some helpful content.

GL_Search

docs/sources/operations/zone-ingesters.md
JStickler: Your phrasing in the comment is clearer (at least to me) than the original text.  I'd go with the text from the comment.

docs/sources/operations/zone-ingesters.md
JStickler: Ugh, sorry, that screen grab turned out to be huge.  The second line under the bold heading is pulled from the `description` in the metadata.

cstyan:
@JStickler thanks for the review! I'll open this PR as ready for review to anyone now, but I think I've addressed all your feedback :+1:

MasslessParticle (approved):

PR #9299, opened 2023-04-26
Expose signature version property for s3 storage configs

trevorwhitney:
This exposed the a `signatureVersion` property in the helm chart's storage config options that can be used to set the `signature_version` for a Loki S3 storage config.


JStickler (approved):
[docs team] LGTM

PR #9320, opened 2023-04-27
Add query info to query.Exec span

salvacorts:
It's kinda difficult to see the query that is being run on a trace.

The only place where I can see the query is in the `url` attribute of the `loki_api_v1_query_range` handler, but since it’s encoded it’s kinda difficult to read. If it's elsewhere it's not quite obvious.

It’d make sense to have the query in the `query.Exec` span along with the stats of the query.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/engine.go
+++ b/pkg/logql/engine.go
@@ -217,7 +217,7 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
 	spLogger := spanlogger.FromContext(ctx)
 	defer spLogger.Finish()
 
-	level.Debug(spLogger).Log(
+	sp.LogKV(
 		"type", GetRangeType(q.params),
 		"query", q.params.Query(),
 		"start", q.params.Start(),


pkg/logql/engine.go
```diff
@@ -214,8 +214,17 @@ func (q *query) resultLength(res promql_parser.Value) int {
 func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
 	sp, ctx := opentracing.StartSpanFromContext(ctx, "query.Exec")
 	defer sp.Finish()
-	log := spanlogger.FromContext(ctx)
-	defer log.Finish()
+	spLogger := spanlogger.FromContext(ctx)
+	defer spLogger.Finish()
+
+	level.Debug(spLogger).Log(
```
DylanGuedes, line 220: that would log a new line for every exec (which tbh isn't that bad). But what do you think of adding a new event to the span and adding those params to it?

salvacorts: Done. Looking at the opentracing docs, it's not super clear to me whether LogKV logs to stdout or not.
Even if it does, looks like it uses some buffered logging approach which should make it way faster ([see][1]).

I see this being used elsewhere within Loki. For example in the Distributor _push_ handler:
https://github.com/grafana/loki/blob/24d426ce4400d3197eef6698c1c058e12db652bb/pkg/distributor/distributor.go#L298

[1]: https://github.com/grafana/loki/blob/d765921735c971d5323f6950311c324f421ef7a9/vendor/github.com/uber/jaeger-client-go/span.go#L266

pkg/logql/engine.go
```diff
@@ -214,8 +214,17 @@ func (q *query) resultLength(res promql_parser.Value) int {
...
+	defer spLogger.Finish()
+
+	sp.LogKV(
+		"type", GetRangeType(q.params),
+		"query", q.params.Query(),
+		"start", q.params.Start(),
+		"end", q.params.End(),
+		"step", q.params.Step(),
+		"length", q.params.End().Sub(q.params.Start()),
+	)
```
DylanGuedes, line 227: `logKV` has this weird pattern where the first key-value pair is actually `event` and the event name.

One more thing: before you merge this, could you test it with a split query in a running cluster? This will work well if there's a span on top of every `Exec`, but if it is a single parent span for all `query.Exec` executions then you'd be emitting the same event multiple times for the same span; in such case, logging the line as you were doing previously is actually better (i.e: my suggestion was crap :rofl: )

```suggestion
        sp, ctx := opentracing.StartSpanFromContext(ctx, "query.Exec")
        defer sp.Finish()
        if sp != nil {
            sp.LogKV(
                "event", "started query exec",
		"type", GetRangeType(q.params),
		"query", q.params.Query(),
		"start", q.params.Start(),
		"end", q.params.End(),
		"step", q.params.Step(),
		"length", q.params.End().Sub(q.params.Start()),
	    )
	    defer sp.LogKV(
                "event", "finished query exec",
		"type", GetRangeType(q.params),
		"query", q.params.Query(),
		"start", q.params.Start(),
		"end", q.params.End(),
		"step", q.params.Step(),
		"length", q.params.End().Sub(q.params.Start()),
	    )
        }
	spLogger := spanlogger.FromContext(ctx)
	defer spLogger.Finish()

	
```

DylanGuedes (changes requested):
almost there

pkg/logql/engine.go
salvacorts: Looks like it's working fine ([source][1]).

image

> logKV has this weird pattern where the first key-value pair is actually event and the event name.

IIUC, this is just a convention, right? In this particular case, I find it kind of unnecessary as the same info gets printed twice plus we already record an event with the query stats once the query is finished executing.

Here's the working version w/o the finish event ([source][2]):
image



[1]: https://admin-dev-us-central-0.grafana.net/grafana/explore?left=%7B%22datasource%22:%22tempo-dev-01%22,%22queries%22:%5B%7B%22query%22:%2239f8ec2fe93e11a1%22,%22queryType%22:%22traceql%22,%22refId%22:%22A%22,%22limit%22:20%7D%5D,%22range%22:%7B%22from%22:%22now-5m%22,%22to%22:%22now%22%7D,%22panelsState%22:%7B%22trace%22:%7B%22spanId%22:%22623b15fbdbb46b74%22%7D%7D%7D&orgId=1
[2]: https://admin-dev-us-central-0.grafana.net/grafana/explore?left=%7B%22datasource%22:%22tempo-dev-01%22,%22queries%22:%5B%7B%22query%22:%225ac20ab385a80918%22,%22queryType%22:%22traceql%22,%22refId%22:%22A%22,%22limit%22:20%7D%5D,%22range%22:%7B%22from%22:%22now-5m%22,%22to%22:%22now%22%7D,%22panelsState%22:%7B%22trace%22:%7B%22spanId%22:%222b2fcbbed16c4584%22%7D%7D%7D&orgId=1

DylanGuedes: > IIUC, this is just a convention, right? In this particular case, I find it kind of unnecessary as the same info gets printed twice plus we already record an event with the query stats once the query is finished executing.

good point, but I think I read somewhere that these different trace libs (OTEL, tempo, jaeger, etc) somehow rely on this pattern. Could you try one more time without the `event` and check how it behaves?

salvacorts: > Could you try one more time without the event and check how it behaves?

Looks to be working fine.

image

DylanGuedes (approved):
lgtm! ty for the patience

PR #9328, opened 2023-04-27
fixes bug during chunk iteration using markers

owen-d:
Tricky one to find, but this does the following
* correctly subtracts remaining chunk count when skipping a page (this was previously using a zero-valued marker)
* correctly set previous maxt on pages that are skipped. This is important because chunks themselves are delta-encoded against the previous chunk.
* removes the markers pool from the chunk decoding which isn't needed.
* adds a test
* adds a bit more logging on which series resulted in a decoding error during chunk iteration

### DIFF
--- a/pkg/storage/stores/tsdb/index/index.go
+++ b/pkg/storage/stores/tsdb/index/index.go
@@ -2392,7 +2392,7 @@ func (dec *Decoder) Series(version int, b []byte, seriesRef storage.SeriesRef, f
 
 	// read chunks based on fmt
 	if err := dec.readChunks(version, d, seriesRef, from, through, chks); err != nil {
-		return 0, err
+		return 0, errors.Wrapf(err, "series %s", lbls.String())
 	}
 	return fprint, nil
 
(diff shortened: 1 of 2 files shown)


pkg/storage/stores/tsdb/index/index.go
```diff
@@ -2429,19 +2428,16 @@ func (dec *Decoder) readChunksV3(d *encoding.Decbuf, from int64, through int64,
 
 	nMarkers = d.Uvarint()
 
-	markers = chunkPageMarkersPool.Get(nMarkers)
-	markers = markers[:nMarkers]
-	defer chunkPageMarkersPool.Put(markers)
-	for i := range markers {
-		markers[i].decode(d)
+	for i := 0; i < nMarkers; i++ {
+		marker.decode(d)
```
DylanGuedes, line 2432: isn't marker null here?

DylanGuedes (approved):

MasslessParticle (approved):

MasslessParticle (approved):

PR #9341, opened 2023-04-28
Reuse string/InstanceDesc buffers in index gateway clients

chaudum:

When running the IndexGateway in ring mode, `ring.MakeBuffersForGet()` in the index gateway client allocated fixed size `[]string`/`[]InstanceDesc` slices with a capacity of 5 for each and every request.
By using a `sync.Pool` the client can reduce the amount of allocations. Additionally, since we run the index gateway ring with a high replication factor to be able to scale them better horizontally, an initial capacity of 5 is most of the time too low, leading to extra allocations when growing the capacity of the slice. 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/storage/stores/indexshipper/gatewayclient/gateway_client.go
+++ b/pkg/storage/stores/indexshipper/gatewayclient/gateway_client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
+	"sync"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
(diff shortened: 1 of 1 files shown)


pkg/storage/stores/indexshipper/gatewayclient/gateway_client.go
```diff
@@ -142,6 +146,20 @@ func NewGatewayClient(cfg IndexGatewayClientConfig, r prometheus.Registerer, log
 			return igPool, nil
 		}
 
+		// Replication factor plus additional room for JOINING/LEAVING instances
+		// See also ring.GetBufferSize
+		bufSize := cfg.Ring.ReplicationFactor() * 3 / 2
```
jeschkies, line 151: Could you use a constant with the same doc string?

chaudum: You don't want this to be a constant, because the RF is a setting that can change.

MasslessParticle (approved):

PR #9346, opened 2023-04-28
operator: Enable Route by default on OpenShift clusters

periklis:
With the availability of the `community-openshift` (See #8881) bundle we control all available hubs for OpenShift clusters. Thus we can assume that a user using either the `redhat-openshift-ecosystem/community-operators-prod` hub or the RedHat internal catalogs installs this Operator on an OpenShift clusters. Therefore we can always assume that the following facilities are available:
- OpenShift Route
- OpenShift Service Certificates Controller

This PR decouples reconciling a Route and a CABundle ConfigMap from the user selection of tenant mode. Formerly the user had to select either tenant mode `openshift-logging` or `openshift-network` to inform the operator that the hosting cluster is OpenShift. In the proposed change the cluster flavor is controlled automatically via the bundled feature gates configuration.

In addition this enables using the tenant modes `static` and `dynamic` with public OIDC providers via the route.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/operator/apis/config/v1/projectconfig_types.go
+++ b/operator/apis/config/v1/projectconfig_types.go
@@ -30,6 +30,7 @@ type BuiltInCertManagement struct {
 type OpenShiftFeatureGates struct {
 	// Enabled defines the flag to enable that these feature gates are used against OpenShift Container Platform releases.
 	Enabled bool `json:"enabled,omitempty"`
+
 	// ServingCertsService enables OpenShift service-ca annotations on the lokistack-gateway service only
 	// to use the in-platform CA and generate a TLS cert/key pair per service for
 	// in-cluster data-in-transit encryption.
--- a/operator/apis/config/v1/projectconfig_types.go
+++ b/operator/apis/config/v1/projectconfig_types.go
@@ -28,6 +28,8 @@ type BuiltInCertManagement struct {
 
 // OpenShiftFeatureGates is the supported set of all operator features gates on OpenShift.
 type OpenShiftFeatureGates struct {
+	// Enabled defines the flag to enable that these feature gates are used against OpenShift Container Platform releases.
+	Enabled bool `json:"enabled,omitempty"`
 	// ServingCertsService enables OpenShift service-ca annotations on the lokistack-gateway service only
 	// to use the in-platform CA and generate a TLS cert/key pair per service for
 	// in-cluster data-in-transit encryption.
--- a/operator/apis/config/v1/projectconfig_types.go
+++ b/operator/apis/config/v1/projectconfig_types.go
@@ -37,11 +37,6 @@ type OpenShiftFeatureGates struct {
 	// More details: https://docs.openshift.com/container-platform/latest/security/certificate_types_descriptions/service-ca-certificates.html
 	ServingCertsService bool `json:"servingCertsService,omitempty"`
 
-	// GatewayRoute enables creating an OpenShift Route for the LokiStack
-	// gateway to expose the service to public internet access.
-	// More details: https://docs.openshift.com/container-platform/latest/networking/understanding-networking.html
-	GatewayRoute bool `json:"gatewayRoute,omitempty"`
-
 	// ExtendedRuleValidation enables extended validation of AlertingRule and RecordingRule
 	// to enforce tenancy in an OpenShift context.
 	ExtendedRuleValidation bool `json:"ruleExtendedValidation,omitempty"`
--- a/operator/internal/manifests/openshift/options.go
+++ b/operator/internal/manifests/openshift/options.go
@@ -61,9 +61,11 @@ func NewOptions(
 	tenantConfigMap map[string]TenantData,
 	rulerName string,
 ) Options {
-	host := ingressHost(stackName, stackNamespace, gwBaseDomain)
-
-	var authn []AuthenticationSpec
+	var (
+		authn []AuthenticationSpec
+		authz AuthorizationSpec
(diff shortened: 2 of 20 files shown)


operator/internal/manifests/openshift/options.go
```diff
@@ -75,27 +89,22 @@ func NewOptions(
 		authn = append(authn, AuthenticationSpec{
```
aminesnow, line 89:
```suggestion
		o.Authentication = append(o.Authentication, AuthenticationSpec{
```

operator/internal/manifests/openshift/options.go
```diff
@@ -75,27 +89,22 @@ func NewOptions(
...
-			GatewayName:          gwName,
-			GatewaySvcName:       gwSvcName,
-			GatewaySvcTargetPort: gwPortName,
-			Labels:               gwLabels,
-			RulerName:            rulerName,
-		},
-		Authentication: authn,
-		Authorization: AuthorizationSpec{
+	if len(tenants) > 0 {
+		authz = AuthorizationSpec{
```
aminesnow, line 99:
```suggestion
		o.Authorization = AuthorizationSpec{
```

operator/internal/manifests/openshift/options.go
aminesnow: You can directly use `o.Authentication`

operator/internal/manifests/openshift/options.go
aminesnow: You can directly use `o.Authorization`.

aminesnow (approved):
Just a code improvement suggestion, otherwise LGTM.

operator/internal/manifests/openshift/options.go
```diff
@@ -54,16 +54,30 @@ type TenantData struct {
...
+		},
+	}
+}
 
-	var authn []AuthenticationSpec
+func (o *Options) WithTenantsForMode(mode lokiv1.ModeType, gwBaseDomain string, tenantConfigMap map[string]TenantData) *Options {
+	var (
+		authn []AuthenticationSpec
+		authz AuthorizationSpec
+		host  string = ingressHost(o.BuildOpts.LokiStackName, o.BuildOpts.LokiStackNamespace, gwBaseDomain)
```
xperimental, line 79:
```suggestion
		host  = ingressHost(o.BuildOpts.LokiStackName, o.BuildOpts.LokiStackNamespace, gwBaseDomain)
```

operator/internal/manifests/openshift/options.go
xperimental: Strictly speaking, this is only true if we know that `o.Authentication` is an empty slice before, otherwise this leads to a different behaviour. This function is used in a sort-of-fluent-interface kind of way. Usually when you do that you directly chain the function calls together, which we are currently not doing.

I think I'd rather leave the definitions of `authz`/`authn` the way they currently are, so that the "output" of this function is completely constructed inside of it.

operator/apis/config/v1/projectconfig_types.go
```diff
@@ -28,6 +28,8 @@ type BuiltInCertManagement struct {
 
 // OpenShiftFeatureGates is the supported set of all operator features gates on OpenShift.
 type OpenShiftFeatureGates struct {
+	// Enabled defines the flag to enable that these feature gates are used against OpenShift Container Platform releases.
+	Enabled bool `json:"enabled,omitempty"`
```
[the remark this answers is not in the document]
periklis: I totally forgot that we have `GatewayRoute` for this simple reason. Why deprecating and not removing at all?

operator/internal/manifests/openshift/options.go
aminesnow: That's right, I thought this was a call once and done function.

operator/apis/config/v1/projectconfig_types.go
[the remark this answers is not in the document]
xperimental: There are a few lines which still make use of `GatewayRoute` in the current code (mostly in the setup code in main). If you want to remove the `GatewayRoute` that code must be changed to now depend on the new flag. It's fine to do that, just did not know if you wanted to do it, so I stuck with "deprecate" for now.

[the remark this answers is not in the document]
periklis: Ok I went for the feature gate replacement option as per main intent of this PR is to have it always enabled on OpenShift clusters.

xperimental (commented):
I can still find two references of `gatewayRoute`:

- in `feature_gates.md` (needs a `make web-pre`)
- in `config.grafana.com_projectconfigs.yaml`

The second one seems to be an old reference that has not been updated in a while. I currently do not remember how to update that file... Is it possible that we generated that file once "by accident" and did not delete it afterwards?

periklis:
> I can still find two references of `gatewayRoute`:
> 
>     * in `feature_gates.md` (needs a `make web-pre`)
> 
>     * in `config.grafana.com_projectconfigs.yaml`
> 
> 
> The second one seems to be an old reference that has not been updated in a while. I currently do not remember how to update that file... Is it possible that we generated that file once "by accident" and did not delete it afterwards?

The generate YAML file is actually obsolete as per this CRD is not used for the API. However kubebuilder/operator-sdk generated this file initially.

xperimental (approved):
Needs a rebase, but otherwise lgtm :+1:

PR #9352, opened 2023-04-29
Promtail: Pull Kubernetes logs directly from API, rather than requiring host mounted volumes

### LINKED ISSUE #7205: [promtail] support scraping from Kubernetes API
Promtail supports generating scrape targets using the K8s API (`kubernetes_sd_configs`). It can get node/pod/container names from the K8s API and use these to generate `__path__` for logfiles to scrape. But AFAIK, it can't get **the logs themselves** from the K8s API.

> while Promtail can utilize the Kubernetes API to discover pods as targets, it can only read log files from pods that are running on the same node as the one Promtail is running on.

The usual pattern for k8s containers is stream logs to stdout, then k8s / the container engine will save these for you. The saved logs are then accessible via `kubectl logs`. It makes sense for Promtail to get logs the same way.

A new type of `scrape_config` which directs Promtail to scrape from the Kubernetes API. This would be analogous to e.g. scraping from the journal. This could be used effectively in combination with the existing `kubernetes_sd_configs`. Something like:

```yaml
scrape_configs:
  - job_name: k8s-logs
    kubernetes_sd_configs:
      - role: pod
    kubernetes_api:
      namespace: ...
      pod: __meta_kubernetes_pod_name
      container: __meta_kubernetes_pod_container_name
    relabel_configs:
      - action: replace
        source_labels:
          - __meta_kubernetes_pod_name
        target_label: pod
      - action: replace
        source_labels:
          - __meta_kubernetes_pod_container_name
        target_label: container
```

The alternative recommended in the [Promtail docs](https://grafana.com/docs/loki/latest/clients/promtail/installation/#sidecar) is to use a hostPath volume mount to the logs directory on the host (`/var/log/pods` and/or `/var/log/containers`). However, from the [k8s docs](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath):
> HostPath volumes present many security risks, and it is a best practice to avoid the use of HostPaths when possible.

It would be safer to use the Kube API, considering it already has a mechanism for capturing and serving logs. It also feels kinda hacky mounting `/var/log/pods` to our Promtail container.

**Additional context**
The Kubernetes API provides a [method to read logs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#read-log-pod-v1-core). This would be the best way to scrape the logs. I assume `kubectl log` internally uses this same API endpoint.

zlepper:
This PR makes Promtail pull logs in Kubernetes directly from the Kubernetes API. This enabled operators only to need 1 promtail pod running, rather than having Promtail run as a Daemonset. Additionally it enables Promtail to very easily work outside of the cluster if required. Lastly it makes it possible to scrabe logs from pods that doesn't run on normal nodes, such as AKS virtual nodes. 

Fixes #7205

Most of the code here is copy-pasted from the Docker target and changed to work with Kubernetes instead. Should I do some refactoring to allow sharing code between the two targets, since the structures is a lot like each other. 

This is also a massive breaking change for anyone running Promtail in kubernetes, since both their relabel_configs and pipeline_stages has to be changed to accommodate. 

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

**Outstanding issues**
* In the docs I can see Promtail supports scraping things like `endpoints`, `service`, `node` and `ingress`. I don't quite get how logging works with those types, that means my change probably breaks those things? I assume this is something left over from how the Prometheus integration works, but maybe isn't relevant for logs at all?
* Because everything can now go through the same running instance of Promtail, I'm not sure how well Promtails performance will be with, say 10000 pods being scraped simultaneously. At least with the old DaemonSet deployment model it was limited to each node, which kept the maximum number of pods at a reasonable level. So I'm not sure if I should keep the existing DaemonSet as a recommendation or use the Deployment example I updated it to. 
* How to deal with API disconnects while streaming the logs? Does Promtail have some internal machinery for automatically retrying, or is that up to the integration?


dannykopping (changes requested):
Really excited about this one!

> Most of the code here is copy-pasted from the Docker target and changed to work with Kubernetes instead. Should I do some refactoring to allow sharing code between the two targets, since the structures is a lot like each other.

Yes please.

> This is also a massive breaking change for anyone running Promtail in kubernetes, since both their relabel_configs and pipeline_stages has to be changed to accommodate.

This line concerns me. Can we introduce this feature in a backwards-compaible way?

I'm blocking this PR because of the BC-break; breaking promtail BC can be disastrous because folks may not realise it's broken until they need to access their logs, only to find that they are missing. Once we discuss this and come to some conclusion, I'll unblock it; I just want to make sure we don't merge this in the interim.

I can still see a few use-cases for running promtail as a DaemonSet even when this feature is implemented:
- to spread out the load
- if a k8s bug is introduced to the logging API
- if the k8s API version is not yet supported

zlepper:
> Really excited about this one!
> 
> > Most of the code here is copy-pasted from the Docker target and changed to work with Kubernetes instead. Should I do some refactoring to allow sharing code between the two targets, since the structures is a lot like each other.
> 
> Yes please.

I have attempted to do this, but my Go is apparently very rusty, and I don't find a combined version any easier to read, rather the opposite. Apparently, the changes compared to the docker version are there, it's just general code flow that looks like each other. 

> 
> > This is also a massive breaking change for anyone running Promtail in kubernetes, since both their relabel_configs and pipeline_stages has to be changed to accommodate.
> 
> This line concerns me. Can we introduce this feature in a backwards-compaible way?
> 
> I'm blocking this PR because of the BC-break; breaking promtail BC can be disastrous because folks may not realise it's broken until they need to access their logs, only to find that they are missing. Once we discuss this and come to some conclusion, I'll unblock it; I just want to make sure we don't merge this in the interim.

That is exactly why I was very careful to highlight that it was a breaking change. I have tried tweaking the code to make a version that shouldn't break backwards compatibility, though it might make things slightly confusing since there is now the same type of service discovery object 2 places in the configuration. 

> 
> I can still see a few use-cases for running promtail as a DaemonSet even when this feature is implemented:
> 
> * to spread out the load

Yup, that's also the main usecase I see.

> * if the k8s API version is not yet supported
I hope the log API is supported, otherwise `kubectl logs` doesn't work, since it pulls logs from the exact same API. But yes, fair point. Using the DaemonSet approach also allows scraping node-level logs while there, this can't do that. 

As for the DaemonSet vs Deployment talk in general, I think I can just describe both in the docs, including why you would want to go with one approach over the other. However before I do that, I will await some consensus if the current code approach is the way you will want to go :)

JStickler (changes requested):
[docs team] Once you figure out the implementation (and backwards compatibility) this still needs some documentation.

anjmao:
Would be interesting to see the overhead on control plane and kubelet. Fetching logs via kubernetes api involves calling control plane -> kubelet which looks like quite a big overhead for large logs volume compared to fetching from mounted volumes.

cstyan:
Hello, thanks for your PR.

We're currently reevaluating promtails position as a project within Grafana Labs. Internally we're actually using the [Agent](https://github.com/grafana/agent) for both metrics and logs collection at this point. Additionally, the agent team is more likely to have time to dedicate to PR.

PR #9358, opened 2023-05-01
Update replication factor ceveat sentence

AleksandrKovshariov:

This sentence "There’s also an edge case where we acknowledge a write if 2 of the three ingesters do which means that in the case where 2 writes succeed, we can only lose one ingester before suffering data loss." is really hard to understand.

I suggest "If a write is acknowledged by 2 out of 3 ingesters, we can tolerate the loss of one ingester but not two, as this would result in data loss."

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


JStickler (approved):
[docs team]  LGTM, I think this is much better phrasing. Thanks for the contribution!

MichelHollands (approved):

PR #9378, opened 2023-05-03
operator: add zone aware API spec validation

aminesnow:
Following #9315, this PR adds validation to the zones awareness spec of the `LokiStack` API. 

@periklis 

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/internal/validation/lokistack.go
```diff
@@ -23,6 +26,7 @@ var _ admission.CustomValidator = &LokiStackValidator{}
 // LokiStackValidator implements a custom validator for LokiStack resources.
 type LokiStackValidator struct {
 	ExtendedValidator func(context.Context, *lokiv1.LokiStack) field.ErrorList
+	Client            client.Client
```
xperimental, line 29: I think you should move `Client` above `ExtendedValidator` as the extended validation is optional, but the client is not.

xperimental (commented):
In general, I don't know if this validation is a good idea, because it might hinder the admin to make changes to a malfunctioning cluster as, depending on the failure, the list of nodes might not contain all the "normally available values" then.

In my opinion, the validation should only fail on configurations which "will absolutely never work" and not depend on the current state.

periklis (changes requested):
After digging a bit into the k8s docs, I suggest to take a different course of action here, since `PodTopologyConstraints` solving the issue for node labels internally offering the configuration option `minDomains` (See https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#spread-constraint-definition):

```
minDomains indicates a minimum number of eligible domains. This field is optional. A domain is a particular instance of a topology. An eligible domain is a domain whose nodes match the node selector.
```

WDYT?

aminesnow:
> After digging a bit into the k8s docs, I suggest to take a different course of action here, since `PodTopologyConstraints` solving the issue for node labels internally offering the configuration option `minDomains` (See https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#spread-constraint-definition):
> 
> ```
> minDomains indicates a minimum number of eligible domains. This field is optional. A domain is a particular instance of a topology. An eligible domain is a domain whose nodes match the node selector.
> ```
> 
> WDYT?

Interestintg find. Will give it a look and see how that goes.
Btw, should we be worried about this: `minDomains field is a beta field and disabled by default in 1.25.`?

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -821,6 +821,7 @@ type ReplicationSpec struct {
 	Factor int32 `json:"factor,omitempty"`
 
 	// Zones defines an array of ZoneSpec that the scheduler will try to satisfy.
+	// IMPORTANT: Make sure that the replication factor defined is less than the number of available zones.
```
periklis, line 824: Is it strictly required to be "less" or can it be also "less or equal"

periklis (approved):

JStickler (approved):
[Docs team]  Approving, since it looks like this is just adding the validation and the API is documented by #9315.

periklis (commented):
LGTM, let's update the branch to keep in sync with generated files.

PR #9387, opened 2023-05-03
Fix typo on lambda-promtail docs for keep_stream var

AluBhorta:
Change "log group" to "log stream" for `keep_stream` var.

### DIFF
--- a/docs/sources/clients/lambda-promtail/_index.md
+++ b/docs/sources/clients/lambda-promtail/_index.md
@@ -48,7 +48,7 @@ provider "aws" {
 }
 ```
 
-To keep the log group label add `-var "keep_stream=true"`.
+To keep the log stream label add `-var "keep_stream=true"`.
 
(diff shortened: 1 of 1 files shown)


JStickler (approved):
[doc team] LGTM

docs/sources/clients/lambda-promtail/_index.md
```diff
@@ -48,7 +48,7 @@ provider "aws" {
 }
 ```
 
-To keep the log group label add `-var "keep_stream=true"`.
+To keep the log stream label add `-var "keep_stream=true"`.
```
cstyan, line 51: can we make this a bit more verbose, and include ```(`__aws_cloudwatch_log_stream\`)``` so that users have an easy to know which label from cloudwatch this would preserve

MasslessParticle:
We haven't seen any response on this in about a month. I'm going to close for inactivity

PR #9400, opened 2023-05-04
Paul1r/track duplicate bytes

paul1r:
This PR is for counting the number of bytes of log lines that were marked as duplicates.  This will be utilized when we start investigating query pricing scenarios.

We previously were not tracking this data.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/iter/entry_iterator.go
```diff
@@ -144,6 +145,7 @@ func (i *mergeEntryIterator) fillBuffer() {
 		for _, t := range previous {
 			if t.Entry.Line == entry.Line {
 				i.stats.AddDuplicates(1)
+				i.stats.AddDuplicateBytes(int64(len(entry.Line)) + 2*binary.MaxVarintLen64)
```
dannykopping, line 148: What's the significance of `2*binary.MaxVarintLen64` here?
I see this repeated in a couple places; maybe a helper with an explanatory comment would be useful.

dannykopping (commented):
Overall LGTM! Added a question.

Nit: can you please give your PR a more descriptive title (and remove your username?)

pkg/iter/entry_iterator.go
paul1r: I swiped the 2*binary.MaxVarintLen64 from [memchunk.go](https://github.com/grafana/loki/blob/main/pkg/chunkenc/memchunk.go#L1163).  When adding the stats of decompressed bytes, that addition of ~20 bytes appears to account for the timestamp.

That being said, if not appropriate here, easy enough to remove.

dannykopping: I have no idea TBH. It just triggered my _wtf reflex_ 😛 
I'd check in with the author of that line to see if we need it here.

cyriltovena: Depends what we want to compute, but the byte processed stats was accounting for the line length and the timestamp for each line as those are encoded and read for memory chunks. The encoding used is varint, and I wanted to avoid encoding to know the size while reading so I went for the max number of bytes a varint 64 number can take.

I think after few years I realised we might want to be more precise and estimate better the size of those two. For instance we know the max line length (most likely below 10k) and timestamps are usually big (`2535372481000000000` in year 2050)

paul1r: With that said, I think I'll remove that computation for now.   Thank you for the input!

PR #9404, opened 2023-05-05
Fix case insensitive label filter

### LINKED ISSUE #9294: Case-insensitive flag line filter regex only matching uppercase
**Describe the bug**
Using the case-insensitive flag `(?i)`in a label regex filter `=~` causes the query to only match uppercase values, rather than any case.

**To Reproduce**
1. Using the [LogQL Analyzer](https://grafana.com/docs/loki/latest/logql/analyzer/?query=%7C+logfmt+%7C+level+%3D%7E+%60%28%3Fi%29info%60&line%5B%5D=level%3Dinfo&line%5B%5D=level%3DINFO&line%5B%5D=level%3DInFo) demonstrates this issue - only one of the sample log entries (the second one, `level=INFO`) is matched by this query
2. Sample data: 
```
level=info
level=INFO
level=InFo
```
3. Test query: `{job="analyze"} | logfmt | level =~ "(?i)info"`

**Expected behavior**
Every row in the sample data should have been matched. Using the case-insensitive flag on stream selectors does not appear to have the same outcome.

**Additional information**
Potentially related to #7837?

kelnage:
When comparing a case insensitive regular expression to a label value, Loki was only matching values that were in upper case. This was caused by compiling a case insensitive regular expression (such as `(?i)hello`), which produced an upper case `prefix` that was then used in `containsLower` as the `substr` value - but that function was written in such a way that only allowed matches of exact case or where the `line` value was upper case and the `substr` value was lower case.

This PR added a some additional test cases to demonstrate/understand the issue, and adds two additional checks to the `containsLower` function to enable it to work correctly when either input is in either case.

Fixes #9294

Given the importance of this code path, I used `BenchmarkLineLabelFilters` to validate that the performance of this code was not affected. Overall, [the differences appeared negligible](https://docs.google.com/spreadsheets/d/1BRKkCzO0i11AObC6ADguDNa7cfG93hB7JLtqYWmnfW0/edit?usp=sharing) on my device (Intel(R) Core(TM) i7-12700H).

**Checklist**
- [ ] Documentation added
- [X] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/log/filter.go
+++ b/pkg/logql/log/filter.go
@@ -315,8 +315,8 @@ func containsLower(line, substr []byte) bool {
 	j := 0
 	for len(line) > 0 {
 		// ascii fast case
-		if c := line[0]; c < utf8.RuneSelf {
-			if c == substr[j] || c+'a'-'A' == substr[j] {
+		if c := line[0]; c < utf8.RuneSelf && substr[j] < utf8.RuneSelf {
+			if c == substr[j] || c+'a'-'A' == substr[j] || c == substr[j]+'a'-'A' {
 				j++
 				if j == len(substr) {
 					return true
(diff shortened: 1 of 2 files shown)


MasslessParticle (approved):
Good catch. Thanks for the contribution!

PR #9408, opened 2023-05-05
operator: Add PodAntiAffinity overwrites per component

JoaoBraveCoding:
Issue: https://issues.redhat.com/browse/LOG-3854
Adds PodAffinity and PodAntiAffinity overwrites per component
Refactors Affinity/AntiAffinity tests to reduce code duplication

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Main
 
 - [9406](https://github.com/grafana/loki/pull/9406) **aminesnow**: Add label selector to zone awareness TopologySpreadConstraints
+- [9408](https://github.com/grafana/loki/pull/9408) **JoaoBraveCoding**: Add PodAntiAffinity overwrites per component
```
periklis, line 4:
```suggestion
- [9408](https://github.com/grafana/loki/pull/9408) **JoaoBraveCoding**: Add PodAntiAffinity overwrites per component
- [9406](https://github.com/grafana/loki/pull/9406) **aminesnow**: Add label selector to zone awareness TopologySpreadConstraints

```

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -269,6 +269,13 @@ type LokiComponentSpec struct {
 	// +optional
 	// +kubebuilder:validation:Optional
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// PodAntiAffinity defines the pod anti affinity scheduling rules to schedule pods
+	// of a component.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	PodAntiAffinity *corev1.PodAntiAffinity `json:"podAntiAffinity,omitempty"`
```
periklis, line 278: AFAIK, anti-affinity is only a pod-level concept in k8s. So WDYT to skip the `Pod` prefix here?

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -269,6 +269,13 @@ type LokiComponentSpec struct {
 	// +optional
 	// +kubebuilder:validation:Optional
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// PodAntiAffinity defines the pod anti affinity scheduling rules to schedule pods
+	// of a component.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
```
periklis, line 277: Consider to use additionaly the operator-sdk annotations for an enhanced console UX experience:
https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#13-podantiaffinity

operator/internal/manifests/var_test_lib.go
periklis, line 1: Why `var_test_lib.go` and not simply `var_test.go`? Furthermore I believe all these tests belong in `node_placement_test.go` anyway.

operator/internal/manifests/ruler_test.go
```diff
@@ -7,8 +7,9 @@ import (
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	v1 "github.com/grafana/loki/operator/apis/config/v1"
```
periklis, line 12: nit. Versioned k8s packages are always aliases as `packagenamev1`, e.g. `configv1`

operator/internal/manifests/var_test_lib.go
```diff
@@ -0,0 +1,135 @@
...
+	lokiv1 "github.com/grafana/loki/operator/apis/loki/v1"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestAffinity(t *testing.T, componentLabel string, makeObject func(cSpec *lokiv1.LokiComponentSpec, nAffinity bool) client.Object) {
```
periklis, line 16: This is the first time we introduce in this code base a Test Runnable Function with a builder function as a parameter. Since we already have two other styles to automate tests, I suggest **not** to introduce a third one:
1. Style 1: Use a table test, e.g. `service_test.go`
3. Style 2: Split each test in an individual function, e.g. `TestAffinity_RulerComponent_WhenSomethingGiven`

periklis (changes requested):

operator/apis/loki/v1/lokistack_types.go
JoaoBraveCoding: I don't 100% agree with this just because it assumes the user also knows about that "truth", while the full name makes it more explicit and clear. But I'm happy to oblige if it makes more sense to drop the `Pod` prefix

operator/internal/manifests/var_test_lib.go
JoaoBraveCoding: If defined the function in `node_placement_test.go` then 2 things would happen:
1. I could not export it because otherwise it would be run by the `go test` command and it would always fail
2. Since I could not export it, some components would not be able to access it since it would not be exported, because their tests are not on the package `manifest_test`.

However this discussion is not relevant given the comment https://github.com/grafana/loki/pull/9408#discussion_r1187305264

operator/config/manager/kustomization.yaml
```diff
@@ -5,5 +5,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openshift-logging/loki-operator
-  newTag: v0.1.0
+  newName: docker.io/grafana/loki-operator
+  newTag: main-ac1c1fd
```
periklis, line 9: Beware these changes need to get reverted after you test locally.

PR #9410, opened 2023-05-05
fix json expression parser

MasslessParticle:
When the value parsed by the `JsonExpressionParser` is an escaped json string, the result still contains escape character and may not be valid json. This causes subsequent json parsing steps to fail. For example:

`{app="foo"} | json message | line_format "{{.message}}" | json` fails when `message` remains escaped.

This PR unescapes the results of extracted `JsonExpressions` so they can be parsed by subsequent json parsers


dannykopping (approved):
LGTM

owen-d (approved):

PR #9417, opened 2023-05-08
Fix handling of () and add tests

MichelHollands:

A panic occurred when the LogQL query `count without () (rate({namespace="apps"}[15s]))`. This is due to the AST code removing the `()` after `without` while printing the expression if there are no values. The rangemapper then got an invalid query.

**Checklist**
- [ ] Documentation added
- [X] Tests updated
- [x] `CHANGELOG.md` updated


ssncferreira (commented):
Not sure we should add this logic to the `String()` method. AFAIK expressions with `by ()` and `without ()` are invalid, so should never even be possible to create a `SampleExpr`. I think it would be best to adjust this to the `ParseSampleExpr` https://github.com/grafana/loki/blob/main/pkg/logql/syntax/parser.go#L66 logic. 
This way, the parsing from string to SampleExpr will fail, e.g., https://github.com/grafana/loki/blob/main/pkg/logql/rangemapper.go#L82

Wdyt?

MichelHollands:
>  AFAIK expressions with `by ()` and `without ()` are invalid, so should never even be possible to create a `SampleExpr`. 

These are explicitly allowed in our grammar: see https://github.com/grafana/loki/blob/main/pkg/logql/syntax/expr.y#L547.
So I think we should handle them correctly.

I've updated the tests. These just needed a `()` after `without`. There are a lot of tests but they all had the same change.

pkg/logql/syntax/ast_test.go
```diff
@@ -681,3 +681,29 @@ func TestParseLargeQuery(t *testing.T) {
...
+		Groups:  []string{"a", "b"},
+		Without: false,
+	}
+	require.Equal(t, " by (a,b)", g.String())
+
+	g = Grouping{
+		Groups:  []string{},
+		Without: false,
+	}
+	require.Equal(t, "", g.String())
```
ssncferreira, line 696: I would assume this structure is equivalent to `by ()` 🕵️‍♀️ 
the empty grouping should be the one where `Grouping = nil`, no?

ssncferreira (commented):
> > AFAIK expressions with `by ()` and `without ()` are invalid, so should never even be possible to create a `SampleExpr`.
> 
> These are explicitly allowed in our grammar: see https://github.com/grafana/loki/blob/main/pkg/logql/syntax/expr.y#L547. So I think we should handle them correctly.
> 
> I've updated the tests. These just needed a `()` after `without`. There are a lot of tests but they all had the same change.

Ahh good catch 👀 
I think I was a bit confused with how the different (non-)grouping structure expression is represented in the code ([link](https://github.com/grafana/loki/blob/main/pkg/logql/syntax/ast.go#L1119-L1127))
```
type VectorAggregationExpr struct {
	Left SampleExpr

	Grouping  *Grouping
	Params    int
	Operation string
	err       error
	implicit
}
type Grouping struct {
	Groups  []string
	Without bool
}
```
I assume this is the representation of the Grouping structure:
* `count (<EXPR>)` represented as `Grouping = nil`
* `count by(<LABELS>) (<EXPR>)` represented as `Grouping = { Without: false, Groups: [<LABELS>] }`
* `count by() (<EXPR>)` represented as `Grouping = { Without: false, Groups: [] }`
* `count without(<LABELS>) (<EXPR>)` represented as `Grouping = { Without: true, Groups: [<LABELS>] }`
* `count without() (<EXPR>)` represented as `Grouping = { Without: true, Groups: [] }`

This is correct, right? If so, can you adjust the `TestGroupingString` test to include the expression with an empty `by()` and the one without any grouping?

pkg/logql/syntax/ast_test.go
MichelHollands: The code was changed to handle `Grouping = nil` and the tests updated.

pkg/logql/rangemapper_test.go
ssncferreira, line 1: We need to keep `without` here, as this a hack to preserve the stream labels. With the new stringer changes, this should be replaced with `without ()`.

See a more detailed explanation here: https://github.com/grafana/loki/blob/main/pkg/logql/rangemapper.go#L47-L48

pkg/logql/syntax/ast.go
```diff
@@ -1101,9 +1101,13 @@ type Grouping struct {
 // impls Stringer
 func (g Grouping) String() string {
 	var sb strings.Builder
+	if g.Groups == nil {
+		return ""
+	}
```
ssncferreira, line 1106: Sorry, maybe it was not clear in my previous [comment](https://github.com/grafana/loki/pull/9417#pullrequestreview-1418427105), but what states that an expression does not have any grouping is to set `Grouping` to nil. Therefore, in those cases, the `String()` method of the Grouping struct should not even happen.

ssncferreira (changes requested):
When an expression does not have any grouping, the `Grouping` struct should be `nil`, so in those cases, we should not invoke the `String()` method of `Grouping`. Let me know if my comments are clear 🙂

pkg/logql/syntax/ast.go
ssncferreira: Additionally, if `g.Groups` is `nil` it should be the same as if `len(g.Groups) == 0`

MichelHollands: @ssncferreira An empty Grouping struct is added to VectorAggregationExpr [here](https://raintank-corp.slack.com/archives/C029V4SSS9L/p1684769654536579). That causes a `by ()` clause to be added to every `avg`, `sum` and so on. Is this expected?

MichelHollands:
Superseded by https://github.com/grafana/loki/pull/9515

PR #9431, opened 2023-05-09
Adding more buckets to loki_memcache_request_duration_seconds histogram

dannykopping:
Users using extstore (with NVMe drives) may experience latencies higher than the ~250ms max as previously defined.


pkg/storage/chunk/cache/memcached.go
```diff
@@ -65,8 +65,13 @@ func NewMemcached(cfg MemcachedConfig, client MemcachedClient, name string, reg
 				Namespace: "loki",
 				Name:      "memcache_request_duration_seconds",
 				Help:      "Total time spent in seconds doing memcache requests.",
-				// Memcached requests are very quick: smallest bucket is 16us, biggest is 1s
```
dannykopping, line 68: This was wrong; the max was ~250ms

MichelHollands (approved):
LGTM

PR #9444, opened 2023-05-10
tsdb wal always includes labels with chunks for replay consistency

owen-d:
After debugging a WAL replay corruption, we found a situation where we were trying to replay chunks written but the labels wal record was missing. I haven't found out how this is possible, so in the meantime this adds a new record format which writes the labels alongside the chunks in the same record. This will result in some data duplication but should be more reliable since chunks can't be replayed into missing series. Given that chunk flushes are orders of magnitude less common than entry pushes, this shouldn't introduce a performance bottleneck in any meaningful way.

I've added a followup commit to be reverted later which continues to write the old multi-record format while being capable of reading both new and old formats. This means we can roll out the readonly version with this commit as a rollback target before we revert the commit and begin writing the new format.

### DIFF
--- a/pkg/storage/stores/tsdb/head_wal.go
+++ b/pkg/storage/stores/tsdb/head_wal.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/tsdb/chunks"
 	"github.com/prometheus/prometheus/tsdb/record"
 	"github.com/prometheus/prometheus/tsdb/wlog"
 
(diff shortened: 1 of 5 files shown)


pkg/storage/stores/tsdb/head_wal.go
```diff
@@ -220,24 +234,60 @@ func (w *headWAL) Stop() error {
...
 
-	// Always write series before chunks
-	if len(record.Series.Labels) > 0 {
-		buf = record.encodeSeriesWithFingerprint(buf[:0])
-		if err := w.wal.Log(buf); err != nil {
-			return err
+		if len(record.Chks.Chks) > 0 {
+			r := &WALRecord{
+				UserID: record.UserID,
+				Chks:   record.Chks,
```
DylanGuedes, line 282: double checking: don't you want Series here too? (in case they're >0)

chaudum:
Code is out of date

PR #9457, opened 2023-05-12
operator: Set seccomp profile to runtime default for all variants

Red-GV:
This PR sets the `seccompProfile` to be the `RuntimeDefault`. Now that the minimum OpenShift version is `4.12` of the operator, this setting can be safely applied to allow the operator to run in a `restricted` Kubernetes environment.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/config/manager/manager.yaml
```diff
@@ -47,3 +47,5 @@ spec:
       terminationGracePeriodSeconds: 10
       securityContext:
         runAsNonRoot: true
+        seccompProfile:
+           type: RuntimeDefault
```
periklis, line 51: This change should be addressed on the overlays for `community-openshift` and `openshift` to skip pulling this change into the `community` bundle. Latter is dedicated to vanillia k8s clusters and we cannot assume that they support this profile.

periklis (changes requested):

operator/config/manager/manager.yaml
Red-GV: Shouldn't then we also be moving `runAsNonRoot` into those bundles as well, @periklis ? Setting this value assumes the operator running with the Kubernetes's `restricted` pod security standard. Setting the `seccompProfile` is mandatory for that standard.

https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

periklis: I don't recall why we kept them though? Can you research a bit if we missed something back then?

Red-GV: I think this may have been overlooked with the [original implementation](https://github.com/grafana/loki/pull/6514). 

We smoke tested to ensure it would work in vanilla K8s, but I don't think we considered if needed to be `restricted` compliant there. OpenShift 4.11+ requires it, but don't see any reason why we need to enforce it on the vanilla K8s ones.

Red-GV:
@periklis Now only OpenShift builds should be automatically subjected to the `restricted` pod security policy. The `RestrictedPodSecurityStandard` runtime variable will be used instead of `RuntimeSeccompProfile` in order to enable these changes on builds.

I did leave `allowPrivilegeEscalation` off for all builds and containers as a default.

operator/apis/config/v1/projectconfig_types.go
```diff
@@ -100,8 +100,13 @@ type FeatureGates struct {
 
 	// RuntimeSeccompProfile enables the restricted seccomp profile on all
 	// Lokistack components.
+	// Deprecated: Please use RestrictedPodSecurityStandard instead.
 	RuntimeSeccompProfile bool `json:"runtimeSeccompProfile,omitempty"`
```
periklis, line 104: Since we are owning this configuration exclusively and maintain control over all available bundles, I suggest to delete this feature gate if not used anymore.

operator/config/overlays/community-openshift/manager_security_context_patch.yaml
```diff
@@ -0,0 +1,17 @@
...
+      containers:
+        - name: manager
+          securityContext:
+            capabilities:
+              drop:
+                - ALL
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+           type: RuntimeDefault
```
periklis, line 17: You don't need to add this patch in the `community-openshift` overlay if exactly the same exists in the `openshift` overlay. The first is pulling the latter, see kustomization.yaml.

operator/internal/manifests/distributor.go
```diff
@@ -117,10 +123,11 @@ func NewDistributorDeployment(opts Options) *appsv1.Deployment {
 				TerminationMessagePath:   "/dev/termination-log",
 				TerminationMessagePolicy: "File",
 				ImagePullPolicy:          "IfNotPresent",
-				SecurityContext:          containerSecurityContext(),
+				SecurityContext: &corev1.SecurityContext{
+					AllowPrivilegeEscalation: pointer.Bool(false),
+				},
```
periklis, line 128: Why not moving such repetition into `configurePodSpecForRestrictedStandard` ?

Red-GV: So, this is just ensuring that the containers will not allow privilege escalation at all times since it's an opt in to make restrictive. Not sure if we only want to toggle this on when building for `restricted` or not.

Red-GV: @periklis , I have changed it so the `allowPrivilegeEscalation: false` is only set when the restricted feature flag is on. I have also applied this logic to only apply to the OpenShift builds through the overlay.

operator/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Main
 
 - [9405](https://github.com/grafana/loki/pull/9405) **periklis**: Add support for configuring HTTP server timeouts
+- [9457](https://github.com/grafana/loki/pull/9457) **Red-GV**: Set seccomp profile to runtime default
```
periklis, line 4:
```suggestion
- [9457](https://github.com/grafana/loki/pull/9457) **Red-GV**: Set seccomp profile to runtime default
- [9405](https://github.com/grafana/loki/pull/9405) **periklis**: Add support for configuring HTTP server timeouts
```

periklis (approved):

PR #9466, opened 2023-05-16
Azure: Create blob container if not found.

### LINKED ISSUE #9467: Create Azure blob containers if they do not exist
Azure storage authentication can get fairly complicated when private access is required. Azure provides service endpoints and private endpoints as an alternative to firewall rules, but service endpoints stop working when an IP is added to the storage account firewall.

To simplify Loki deployments and reduce operational complexity, Loki should create blob containers if they do not exist. It has access to the storage account, and this measure ensures Loki will work in AKS scenarios.

Loki should create Azure storage blob containers if they do not exist.

It's possible to currently run init containers to achieve this functionality, but causes unnecessary overhead. It's also possible to stage the containers beforehand in an IAC pipeline, but this opens up networking issues depending on where the IAC runners are hosted. For Github actions, for instance, only a large runner with a public IP will be able to access Azure storage. In turn, enabling firewall rules breaks service endpoints for Loki and requires a private link.

**Additional context**

appkins:
This PR aligns with the fallback behavior in Thanos object store for creating azure blob containers when they do not exist. Loki depends on the containers to function and already configures the necessary credentials for creating a container. It follows that Loki should create any required containers to prevent running in a failed state.

This behavior matches the code for Thanos here:
https://github.com/thanos-io/objstore/blob/main/providers/azure/azure.go#L167

```go
ctx := context.Background()
	_, err = containerClient.GetProperties(ctx, &container.GetPropertiesOptions{})
	if err != nil {
		if !bloberror.HasCode(err, bloberror.ContainerNotFound) {
			return nil, err
		}
		_, err := containerClient.Create(ctx, nil)
		if err != nil {
			return nil, errors.Wrapf(err, "error creating Azure blob container: %s", conf.ContainerName)
		}
		level.Info(logger).Log("msg", "Azure blob container successfully created", "address", conf.ContainerName)
	}
```

I tried to match the behavior as closely as possible while using the existing `azblob` implementation in Loki (vs the azure sdk library in Thanos object store).

Fixes #9467

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/storage/chunk/client/azure/blob_storage_client.go
+++ b/pkg/storage/chunk/client/azure/blob_storage_client.go
@@ -209,6 +209,17 @@ func NewBlobStorage(cfg *BlobStorageConfig, metrics BlobStorageMetrics, hedgingC
 		return nil, err
 	}
 
+	ctx := context.Background()
+	_, err = blobStorage.containerURL.GetProperties(ctx, azblob.LeaseAccessConditions{})
+	if err != nil {
+		if bloberror, ok := err.(azblob.StorageError); ok && bloberror.ServiceCode() == azblob.ServiceCodeContainerNotFound {
+			_, err = blobStorage.containerURL.Create(ctx, azblob.Metadata{}, azblob.PublicAccessNone)
+			if err != nil {
+				return nil, fmt.Errorf("error creating Azure blob container: %s", blobStorage.cfg.ContainerName)
+			}
+		}
+	}
+
 	return blobStorage, nil
 }
 
--- a/pkg/storage/chunk/client/azure/blob_storage_client_test.go
+++ b/pkg/storage/chunk/client/azure/blob_storage_client_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/grafana/dskit/flagext"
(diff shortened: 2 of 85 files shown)


MasslessParticle:
This looks reasonable. Could you add a test for the new behavior?

pkg/storage/chunk/client/azure/blob_storage_client_test.go
```diff
@@ -90,7 +90,7 @@ func Test_Hedging(t *testing.T) {
 	}{
 		{
 			"delete/put/list are not hedged",
-			3,
+			4,
```
chaudum, line 93: Can we add a comment that the one additional request comes from the `GetProperties()` call when initialising the client?

pkg/storage/chunk/client/azure/blob_storage_client.go
```diff
@@ -209,6 +209,17 @@ func NewBlobStorage(cfg *BlobStorageConfig, metrics BlobStorageMetrics, hedgingC
 		return nil, err
 	}
 
+	ctx := context.Background()
+	_, err = blobStorage.containerURL.GetProperties(ctx, azblob.LeaseAccessConditions{})
+	if err != nil {
+		if bloberror, ok := err.(azblob.StorageError); ok && bloberror.ServiceCode() == azblob.ServiceCodeContainerNotFound {
+			_, err = blobStorage.containerURL.Create(ctx, azblob.Metadata{}, azblob.PublicAccessNone)
+			if err != nil {
+				return nil, fmt.Errorf("error creating Azure blob container: %s", blobStorage.cfg.ContainerName)
```
chaudum, line 218: Would it make sense to wrap the original error?

```suggestion
				return nil, fmt.Errorf("failed to create create Azure blob container '%s': %w", blobStorage.cfg.ContainerName, err)
```

PR #9471, opened 2023-05-17
query-scheduler: fix query distribution in SSD mode

### LINKED ISSUE #9195: Loki queries not split across queriers
TL;DR: How do I split queries across all of the available loki-read pods (queriers)? Do I need to setup the query-frontend as a separate deployment per the [docs](https://grafana.com/docs/loki/latest/configuration/query-frontend/) or is there something I'm missing in our config?

---

Hello all! I've recently setup Loki in our environment, we ship logs from 200K+ assets to Loki every 30 minutes or so via the Loki push API. We have two labels, and Loki seems pretty happy right now in regards to writes (our backend is S3.) The problem is querying logs. When I search a few hours back, queries are pretty snappy. However, we need to look back as far as 30 days and this is where we're seeing issues. I believe queries are not being split up and load balanced across the available `loki-read` pods (queriers).

We're using Loki in the single binary mode with `loki-read` and `loki-write` pods after reading this [blog post](https://grafana.com/blog/2022/12/19/the-only-helm-chart-you-need-for-grafana-loki-is-here/). There are quite a few Helm charts, but it seems the developers are saying to use the official `grafana/loki` chart. What I'm confused about is that even though the recommendation seems to be to use the `grafana/loki` chart which creates two statefulsets and a deployment for the gateway, I still see the [query-frontend](https://grafana.com/docs/loki/latest/configuration/query-frontend/) being mentioned in the official docs, which makes it seem like we should be setting up a separate instance of the `query-frontend`. However, when I hit the `/services` endpoint on the `loki-read` pods I can see that the `query-frontend` service is already running, 🤔 

```
ring => Running
query-scheduler => Running
index-gateway-ring => Running
store => Running
querier => Running
index-gateway => Running
compactor => Running
cache-generation-loader => Running
server => Running
query-frontend-tripperware => Running
usage-report => Running
ingester-querier => Running
query-frontend => Running                                <==== 👋 
ruler => Running
memberlist-kv => Running
```

Here is our config currently:

```
auth_enabled: false
common:
  path_prefix: /var/loki
  replication_factor: 3
  storage:
    s3: 
      bucketnames: <bucket>
      region: us-east-2
limits_config:
  unordered_writes: true
  max_streams_per_user: 0
  max_global_streams_per_user: 0
  enforce_metric_name: false
  reject_old_samples: false
  split_queries_by_interval: 15m
  max_chunks_per_query: 20000000
  max_query_length: 0
  max_query_series: 1000
  cardinality_limit: 1000000
  max_streams_matchers_per_query: 10000
  max_concurrent_tail_requests: 100
  max_entries_limit_per_query: 10000
ingester:
  wal:
    replay_memory_ceiling: 8GB
memberlist:
  join_members:
  - loki-memberlist
schema_config:
  configs:
  - from: "2022-01-11"
    index:
      period: 24h
      prefix: loki_index_
    object_store: s3
    schema: v12
    store: boltdb-shipper
server:
  grpc_listen_port: 9095
  http_listen_port: 3100
  http_server_read_timeout: 300s
  http_server_write_timeout: 60s
  http_server_idle_timeout: 240s
  grpc_server_max_recv_msg_size: 8388608 # default is 4194304 bytes
  grpc_server_max_send_msg_size: 8388608 # default is 4194304 bytes
frontend:
  log_queries_longer_than: 5s
```

When we query for two days of logs, we get this error:
```
Get "http://loki-gateway.loki.svc:80/loki/api/v1/query_range?direction=backward&end=1681912422566000000&limit=10000000&query=%7Blog%3D%22system%22%7D+%7C%3D+%22asset_tag%3Da1e2f8%22&start=1681739622566000000&step=60000ms": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

During this query ☝️ some of the queriers weren't doing anything at all:

image

In addition, when I look through the logs, I would expect each querier to perform a small query different from other queriers, i.e. I should see different values for `start_delta` and `end_delta` in each query, but I see the same exact queries being performed by the queriers.

Finally, here are our `loki-read` pod resources:

```
- op: replace
  path: /spec/replicas
  value: 9
- op: replace
  path: /spec/template/spec/containers/0/resources
  value:
    requests:
      cpu: 3000m
      memory: 6Gi
    limits:
      cpu: 3000m
      memory: 8Gi
```

sandeepsukhani:
When we run the `query-scheduler` in `ring` mode, `queriers` and `query-frontend` discover the available `query-scheduler` instances using the ring. However, we have a problem when `query-schedulers` are not running in the same process as queriers and query-frontend since [we try to get the ring client interface from the scheduler instance](https://github.com/grafana/loki/blob/abd6131bba18db7f3575241c5e6dc4eed879fbc0/pkg/loki/modules.go#L358).

This causes queries not to be spread across all the available queriers when running in SSD mode because [we point querier workers to query frontend when there is no ring client and scheduler address configured](https://github.com/grafana/loki/blob/b05f4fced305800b32641ae84e3bed5f1794fa7d/pkg/querier/worker_service.go#L115).

I have fixed this issue by adding a new hidden target to initialize the ring client in `reader`/`member` mode based on which service is initializing it. `reader` mode will be used by `queriers` and `query-frontend` for discovering `query-scheduler` instances from the ring. `member` mode will be used by `query-schedulers` for registering themselves in the ring.

I have also made a couple of changes not directly related to the issue but it fixes some problems:
* [reset metric registry for each integration test](https://github.com/grafana/loki/commit/18c4fe59078b649ad6a788a48765b101d0b97618) - Previously we were reusing the same registry for all the tests and just [ignored the attempts to register same metrics](https://github.com/grafana/loki/blob/01f0ded7fcb57e3a7b26ffc1e8e3abf04a403825/integration/cluster/cluster.go#L113). This causes the registry to have metrics registered only from the first test so any updates from subsequent tests won't reflect in the metrics. metrics was the only reliable way for me to verify that `query-schedulers` were connected to `queriers` and `query-frontend` when running in ring mode in the integration test that I added to test my changes. This should also help with other tests where earlier it was hard to reliably check the metrics.
* [load config from cli as well before applying dynamic config](https://github.com/grafana/loki/commit/f9e2448fc7e718db107165cd908054c806b84337) - Previously we were applying dynamic config considering just the config from config file. This results in unexpected config changes, for example, [this config change](https://github.com/grafana/loki/blob/4148dd2c51cb827ec3889298508b95ec7731e7fd/integration/loki_micro_services_test.go#L66) was getting ignored and [dynamic config tuning was unexpectedly turning on ring mode](https://github.com/grafana/loki/blob/52cd0a39b8266564352c61ab9b845ab597008770/pkg/loki/config_wrapper.go#L94) in the config. It is better to do any config tuning based on both file and cli args configs.

Fixes #9195 

I have copied most of the ring manager code from [indexgateway RingManager](https://github.com/grafana/loki/blob/440fb7d74cad481bebb222b21fc063fa56c2f5ac/pkg/storage/stores/shipper/indexgateway/ringmanager.go). I will open a follow-up PR to refactor and share the code between the two since most of the code is the same.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/scheduler/lifecycle.go
```diff
@@ -0,0 +1,28 @@
+package scheduler
+
+import (
+	"github.com/grafana/dskit/ring"
+)
+
+func (rm *RingManager) OnRingInstanceRegister(_ *ring.BasicLifecycler, ringDesc ring.Desc, instanceExists bool, instanceID string, instanceDesc ring.InstanceDesc) (ring.InstanceState, ring.Tokens) {
+	// When we initialize the index gateway instance in the ring we want to start from
```
trevorwhitney, line 8: I think this comment should say scheduler?

pkg/loki/modules.go
```diff
@@ -1222,24 +1223,45 @@ func (t *Loki) initIndexGatewayRing() (_ services.Service, err error) {
...
+func (t *Loki) initQuerySchedulerRing() (_ services.Service, err error) {
+	if !t.Cfg.QueryScheduler.UseSchedulerRing {
+		return
+	}
+
+	// Set some config sections from other config sections in the config struct
+	t.Cfg.QueryScheduler.SchedulerRing.ListenPort = t.Cfg.Server.GRPCListenPort
+
+	managerMode := scheduler.RingManagerModeReader
+	if t.Cfg.isModuleEnabled(QueryScheduler) || t.Cfg.isModuleEnabled(Backend) || t.Cfg.isModuleEnabled(All) {
```
trevorwhitney, line 1247: we are missing a check for legacy read mode here. in legacy read mode (before `backend` was introduced), the scheduler was part of the read target, so we should be a `member` when `t.Cfg.LegacyReadTarget && t.Cfg.isModuleEnabled(Read)`

pkg/scheduler/ringmanager.go
```diff
@@ -0,0 +1,244 @@
...
+		kv.RegistererWithKVName(prometheus.WrapRegistererWithPrefix("loki_", registerer), "scheduler"),
+		rm.log,
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "scheduler ring manager create KV store client")
+	}
+
+	// instantiate ring for both mode modes.
+	ringCfg := rm.cfg.SchedulerRing.ToRingConfig(ringReplicationFactor)
+	rm.Ring, err = ring.NewWithStoreClientAndStrategy(ringCfg, ringNameForServer, ringKey, ringStore, ring.NewIgnoreUnhealthyInstancesReplicationStrategy(), prometheus.WrapRegistererWithPrefix("cortex_", registerer), rm.log)
```
trevorwhitney, line 100: nit: can we break this long function call up over multiple lines?

pkg/scheduler/ringmanager.go
```diff
@@ -0,0 +1,244 @@
...
+	svcs := []services.Service{rm.Ring}
+	rm.subservices, err = services.NewManager(svcs...)
+	if err != nil {
+		return errors.Wrap(err, "failed to create services manager for scheduler ring manager in reader mode")
+	}
+
+	rm.subservicesWatcher = services.NewFailureWatcher()
+	rm.subservicesWatcher.WatchManager(rm.subservices)
+
+	rm.Service = services.NewIdleService(func(ctx context.Context) error {
```
trevorwhitney, line 159: sorry for the lack of understanding, but I'm not quite following how this IdleService works, and how it is able to read the ring without adding itself to it? Is it because it only has a `Ring` service and not a `RingLifecycler` service?

trevorwhitney (changes requested):
This looks great, thanks for fixing this! I think we're just missing one small check for legacy read mode, and then LGTM!

pkg/scheduler/ringmanager.go
sandeepsukhani: When running `RingManager` in reader mode, we only use the ring client for reading the ring. We do not have to register the service in the ring to be able to read the ring. `RingLifecycler` is required only when we want to register tokens in the ring.

trevorwhitney (approved):
LGTM!

PR #9474, opened 2023-05-17
Better combines stats in ingesters

owen-d:
* Reduces double counting streams when possible
* proportionally add stats based on chunk overlap range
* code reuse

### DIFF
--- a/pkg/ingester/instance.go
+++ b/pkg/ingester/instance.go
@@ -575,7 +575,7 @@ func (i *instance) GetStats(ctx context.Context, req *logproto.IndexStatsRequest
 		// Consider streams which overlap our time range
 		if shouldConsiderStream(s, from, through) {
 			s.chunkMtx.RLock()
-			res.Streams++
+			var hasChunkOverlap bool
 			for _, chk := range s.chunks {
 				// Consider chunks which overlap our time range
 				// and haven't been flushed.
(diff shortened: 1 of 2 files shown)


pkg/ingester/instance.go
```diff
@@ -584,12 +584,17 @@ func (i *instance) GetStats(ctx context.Context, req *logproto.IndexStatsRequest
 				chkFrom, chkThrough := chk.chunk.Bounds()
 
 				if !chk.flushed.Equal(zeroValueTime) && from.Before(chkThrough) && through.After(chkFrom) {
```
DylanGuedes, line 586: double checking: shouldn't here be `from.Before(chkThrough) || through.After(chkFrom)` as it is a partial overlap?
Ex:
```
	//          From           chkThrough       Through  chkFrom
	//       ┌────────┬───────────────┬────────┐
	//       │                      |  *      Chunk          *        │
	//       └────────┴──────────────   ┴────────┘
```

owen-d: `from.Before(chkThrough) && through.After(chkFrom)` can only be true when there's some degree of overlap with the time range and chunk. In your ascii example that holds true. If we used `||` instead of `&&`, it would include potentially non-overlapping ranges.

salvacorts (approved):
LGTM. Approving to unblock but I think we should add a test for this if possible.

slim-bean (approved):

PR #9476, opened 2023-05-17
Fix query behavior in Single Binary and SSD deployments

MasslessParticle:
To keep ingesters from calling themselves, the AsyncStore was only being used when the module was `Querier`, `Ruler`, or `Read`. This means that, for some operations, ingesters were never queries when the deployment mode was single-binary or `Backend` (where the ruler now lives).

This change scopes the async store to just the querier rather than all storage access and always includes in it queriers. 

Because the `AsyncStore` isn't automatically instantiated, now, I've also moved `IngesterQuerier` to be a dependency to the store. That way anything that wants to make an `AsyncStore` is guaranteed to have what it needs.


pkg/loki/modules.go
```diff
@@ -654,15 +654,12 @@ func (t *Loki) initStore() (_ services.Service, err error) {
...
-			t.Cfg.StorageConfig.EnableAsyncStore = true
-			t.Cfg.StorageConfig.AsyncStoreConfig = storage.AsyncStoreCfg{
-				IngesterQuerier: t.ingesterQuerier,
-				QueryIngestersWithin: calculateAsyncStoreQueryIngestersWithin(
-					t.Cfg.Querier.QueryIngestersWithin,
-					minIngesterQueryStoreDuration,
-				),
-			}
+		t.Cfg.StorageConfig.AsyncStoreConfig = storage.AsyncStoreCfg{
+			IngesterQuerier: t.ingesterQuerier,
```
ashwanthgoli, line 658: wondering how this assignment works given that the store has no dependency on ingesterQuerier. It would be possible for initStore to get called before ingesterQuerier is initialised

I am sure i must be missing something given that this code already works

MasslessParticle: This is a great question. It looks like it's been added in a couple of places.  

It works because it's a dependency of [`Querier`](https://github.com/grafana/loki/blob/358b8965c4fe5e6d95d09aa3166a02213974bd64/pkg/loki/loki.go#L655) and [`RuleEvaluator`](https://github.com/grafana/loki/blob/358b8965c4fe5e6d95d09aa3166a02213974bd64/pkg/loki/loki.go#L660) which are both dependencies of `All`, `Read`, `Ruler`, and `Backend`

It's also added as a dependency [here](https://github.com/grafana/loki/blob/358b8965c4fe5e6d95d09aa3166a02213974bd64/pkg/loki/loki.go#L706-L709). I think this is redundant as it's in the above graph, now.

MasslessParticle: After looking at it some more, I just made `IngesterQuerier` a dependency of `Store`. Before, they were peers so `IngesterQuerier` might not exist when `initStore` was called. Now the dependency is explicit.

pkg/storage/store.go
```diff
@@ -170,9 +170,6 @@ func (s *store) init() error {
 		s.composite.AddStore(p.From.Time, f, idx, w, stop)
 	}
 
-	if s.cfg.EnableAsyncStore {
```
trevorwhitney, line 173: can we get rid of the `EnableAsyncStore` property from the `Config` struct in `pkg/storage/factory.go` now? I think this was the only place it was being used.

trevorwhitney (approved):
My understanding of the change here is that `storage.NewAsyncStore` is only being called in the init functions for the `RuleEvaluator` and the `Querier`, and there is no longer being called in the init function on a `Store` anymore. As a result, it is safe to make `IngesterQuerier` a dependency of `Store`, because it's fine to always have a `t.ingesterQuerier` instantiated when you have store, because only the querier and rule evaluator will actually use it to setup an `AsyncStore`.

If I understood correctly, LGTM!

pkg/storage/store.go
MasslessParticle: done!

PR #9479, opened 2023-05-18
Count unflushed chunks in instance stats

MasslessParticle:
The instance wasn't counting unflushed chunks in the stats. This change checks that the chunk's `flushed time` is 0 which means it hasn't been flushed yet.


owen-d (approved):
LGTM, fixes my old bug

PR #9502, opened 2023-05-23
Fix panic in `pkg/analytics`

simonswine:
We discovered a bug in Phlare https://github.com/grafana/phlare/pull/706. There can be a race occuring between two different `New<Counter|Int>` calls for the same `name`.


MichelHollands (approved):
Nice use of generics!

MichelHollands:
@simonswine The build is failing.

PR #9503, opened 2023-05-23
operator: Add Pod annotations with node topology labels to support zone aware scheduling

shwetaap:
This PR adds support to zone aware scheduling by adding zone specific node topology labels to loki component pods
Fixes #[LOG-3834](https://issues.redhat.com/browse/LOG-3834)

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


operator/controllers/loki/lokistack_zone_labeling_controller.go
```diff
@@ -0,0 +1,122 @@
...
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, kverrors.Wrap(err, "failed to lookup lokistack", "name", req.NamespacedName)
+	}
+
+	if len(stack.Spec.Replication.Zones) > 0 {
+		var pods corev1.PodList
+		if err := r.Client.List(ctx, &pods); err != nil {
+			return ctrl.Result{}, err
+		}
```
periklis, line 68: Instead of listing all pods on the system you can use the `ListOptions` to limit on the `ctrl.Request` namespace and using its name as `app.kubernetes.io/component=lokistack-dev` label.

operator/go.mod
```diff
@@ -28,6 +28,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/baidubce/bce-sdk-go v0.9.141
+
```
periklis, line 32: We don't need this import because the import reference below is wrong.

operator/internal/handlers/lokistack_podannotation.go
```diff
@@ -0,0 +1,124 @@
+package handlers
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/ViaQ/logerr/v2/kverrors"
+	"github.com/baidubce/bce-sdk-go/util/log"
```
periklis, line 8:
```suggestion
	"github.com/go-logr/logr"
```

operator/internal/handlers/lokistack_podannotation.go
periklis, line 1: Can we give the handler a proper name for the purpose it serves? All handlers have this naming pattern `lokistack_VERB_OBJECT`, suggest something like `lokistack_enable_zone_awareness.go`. But please let your imagination flow.

operator/internal/manifests/distributor.go
```diff
@@ -134,6 +134,9 @@ func NewDistributorDeployment(opts Options) *appsv1.Deployment {
 
 	if opts.Stack.Replication != nil {
 		podSpec.TopologySpreadConstraints = append(podSpec.TopologySpreadConstraints, topologySpreadConstraints(*opts.Stack.Replication, LabelDistributorComponent, opts.Name)...)
+		if len(opts.Stack.Replication.Zones) > 0 {
+			podSpec.Containers[0].Env = append(podSpec.Containers[0].Env, getZoneEnvVar())
+		}
```
periklis, line 139: Is appending the `INSTANCE_AVAILABILITY_ZONE` env var safe on updates? I guess on every update of the pod spec, we will get the same variable appended each time. This will lead to excessive audit logs on duplicate environment variables. Do you recall adding this for the proxy variables?

https://github.com/grafana/loki/blob/f43dd58459239057cf8ca0dbe53b5e7ef89c7ae9/operator/internal/manifests/proxy_env.go#L50-L57

operator/internal/manifests/internal/config/options.go
```diff
@@ -68,6 +68,14 @@ type GossipRing struct {
...
 	// MembersDiscoveryAddr is required
 	MembersDiscoveryAddr string
+	// Zone is optional
+	Zone *ZoneConfig
+}
+
+// ZoneConfig  defines the zone aware configuration for Loki
+type ZoneConfig struct {
+	// InstanceAvailabilityZone is required, informs loki where this instance lives
+	InstanceAvailabilityZone string
```
periklis, line 78: I suggest to skip adding an extra struct and keep this simple with a field `InstanceAvailabilityZone string` in in the `GossipRing` struct.

operator/internal/manifests/var.go
```diff
@@ -619,3 +621,26 @@ func lokiReadinessProbe() *corev1.Probe {
...
+
+// isZoneawareComponent returns true if the Component pod is either in the read/write path or gateway.
+func CheckZoneawareComponent(pod *corev1.Pod) bool {
+	podLabels := pod.Labels
+
+	if podLabels[kubernetesComponentLabel] == LabelDistributorComponent || podLabels[kubernetesComponentLabel] == LabelIndexGatewayComponent ||
+		podLabels[kubernetesComponentLabel] == LabelIngesterComponent || podLabels[kubernetesComponentLabel] == LabelQuerierComponent ||
+		podLabels[kubernetesComponentLabel] == LabelQueryFrontendComponent {
+		return true
+	}
```
periklis, line 633: This is purely screaming for a switch statement to improve readability.

operator/internal/manifests/config.go
```diff
@@ -102,6 +103,12 @@ func ConfigOptions(opt Options) config.Options {
 		}
 	}
 
+	gossippRingconfig := gossipRingConfig(opt.Name, opt.Namespace, opt.Stack.HashRing)
+
+	if zoneAwareEnabled {
+		gossippRingconfig.Zone = zoneAwareConfig(opt.Stack.Replication)
+	}
```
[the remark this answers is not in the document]
shwetaap: Sure, I did consider it. The reason I kept it separate is, I thought adding Zone related configs directly into gossipRing, might be confusing (even though that's how it is finally added in the loki-config.yaml). Also a different struct would be good, if at some point in the future additional zone related configs might get added. But basically I am open to blending it into the gossipring struct too since a different struct is not an immediate requirement.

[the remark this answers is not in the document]
periklis: IMHO future additional zone configs are a thing of the future, thus we can expand to a separate struct if this future materializes somehow.

operator/controllers/loki/lokistack_zone_labeling_controller.go
shwetaap: I am assuming you meant - "app.kubernetes.io/instance": lokistack-dev  and not component. I made the change

operator/go.mod
shwetaap: Done

operator/internal/manifests/var.go
shwetaap: Done

operator/internal/manifests/distributor.go
shwetaap: Sorry I wanted to check what was done in proxy env before I respond. I had indeed forgotten we had added the reset Env function. It makes sense to use the same here as well. I will add that here as well.

operator/controllers/loki/lokistack_zone_labeling_controller.go
```diff
@@ -0,0 +1,130 @@
...
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		for _, pod := range pods.Items {
+			err := handlers.AnnotatePodWithNodeLabels(ctx, r.Log, r.Client, stack.Name, stack.Namespace, pod, stack.Spec.Replication.Zones)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+		}
```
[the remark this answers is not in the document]
shwetaap: Done

operator/internal/handlers/lokistack_enable_zone_awareness.go
```diff
@@ -0,0 +1,158 @@
...
+				continue
+			default:
+				if podAnnotations[LokiInstanceAZ] != "" {
+					podAnnotations[LokiInstanceAZ] = podAnnotations[LokiInstanceAZ] + "_" + podAnnotations[expectedAnnotation]
+				} else {
+					podAnnotations[LokiInstanceAZ] = podAnnotations[expectedAnnotation]
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
shwetaap: Moved the repetitive code into a different function, but continued the setting of the different  podannations in this block itself. Let me know if this seems good.

operator/internal/handlers/lokistack_enable_zone_awareness.go
```diff
@@ -0,0 +1,171 @@
...
+		return kverrors.Wrap(err, "failed to get pod list in lokistack", "name", stackName)
+	}
+
+	for _, pod := range pods.Items {
+		ll := log.WithValues("lokistack-pod-zone-annotation event", "createOrUpdatePred", "pod", pod.Name)
+		componentPod := manifests.CheckZoneawareComponent(&pod)
+
+		if !componentPod {
+			return nil
+		}
```
periklis, line 57:
```suggestion
		if !manifests.CheckZoneawareComponent(&pod) {
			return nil
		}
```

periklis (changes requested):

JoaoBraveCoding (commented):
First pass, looking really good! Just some minor comments

operator/internal/handlers/lokistack_enable_zone_awareness.go
```diff
@@ -0,0 +1,171 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	lokiv1 "github.com/grafana/loki/operator/apis/loki/v1"
+	"github.com/grafana/loki/operator/internal/external/k8s"
+	"github.com/grafana/loki/operator/internal/manifests"
+)
+
+const (
+	LokiInstanceAZ = "loki_instance_availability_zone"
```
periklis, line 21: After testing this PR a bit on a playground cluster, I noticed that annotations managed by the Loki Operator follow this pattern: `loki.grafana.com/label-name`. I suggest to rename this `loki.grafana.com/availability-zone`. 

nit. Also let's follow the var naming pattern for Annotation variables like in `manifests/var.go`: `annotationLokiInstanceAZ`

periklis (changes requested):

operator/internal/manifests/distributor.go
```diff
@@ -134,6 +134,10 @@ func NewDistributorDeployment(opts Options) *appsv1.Deployment {
 
 	if opts.Stack.Replication != nil {
 		podSpec.TopologySpreadConstraints = append(podSpec.TopologySpreadConstraints, topologySpreadConstraints(*opts.Stack.Replication, LabelDistributorComponent, opts.Name)...)
+		if len(opts.Stack.Replication.Zones) > 0 {
+			resetEnvVar(&podSpec, availibilityZoneEnvVarName)
+			podSpec.Containers[0].Env = append(podSpec.Containers[0].Env, getInstanceAvailabilityZoneEnvVar())
```
xperimental, line 139: `podSpec` is produced by this function. I think it would be much more straight-forward to produce the environment variable configuration directly in the beginning of the function and use that when creating `podSpec` instead of creating it with empty env variables first and then, possibly, updating the value later.

periklis:
After doing some testing on this using an OpenShift cluster with worker nodes on three different AWS zones on the same AWS region, I noticed that our patching scheduled pods strategy does not always work as expected. All not so often but very likely to happen I see pods missing the annotations that the `lokistack_zone_labeling_controller` is supposed to enforced:

```
❯ kubectl get pod -l app.kubernetes.io/instance=lokistack-dev -ocustom-columns='NAME:.metadata.name,NODE:.spec.nodeName,AVAILABILITY-ZONE:.metadata.annotations.loki\.grafana\.com\/availability\-zone' --sort-by=".metadata.labels.app\.kubernetes\.io\/component"

NAME                                            NODE                                            AVAILABILITY-ZONE
lokistack-dev-distributor-695c6d77fc-m9jd5      ip-10-0-169-238.eu-central-1.compute.internal   eu-central-1_eu-central-1b
lokistack-dev-distributor-695c6d77fc-w5wmj      ip-10-0-155-166.eu-central-1.compute.internal   eu-central-1_eu-central-1a
lokistack-dev-index-gateway-0                   ip-10-0-205-76.eu-central-1.compute.internal    eu-central-1_eu-central-1c
lokistack-dev-index-gateway-1                   ip-10-0-169-238.eu-central-1.compute.internal   <none>
lokistack-dev-ingester-0                        ip-10-0-205-76.eu-central-1.compute.internal    <none>
lokistack-dev-ingester-1                        ip-10-0-169-238.eu-central-1.compute.internal   eu-central-1_eu-central-1b
lokistack-dev-querier-76f87c9d95-gz6qx          ip-10-0-155-166.eu-central-1.compute.internal   eu-central-1_eu-central-1a
lokistack-dev-querier-76f87c9d95-z85kw          ip-10-0-169-238.eu-central-1.compute.internal   eu-central-1_eu-central-1b
lokistack-dev-query-frontend-59d794c575-htgsr   ip-10-0-155-166.eu-central-1.compute.internal   eu-central-1_eu-central-1a
lokistack-dev-query-frontend-59d794c575-v6vlg   ip-10-0-169-238.eu-central-1.compute.internal   eu-central-1_eu-central-1b
``` 

For the sake of completeness, here is how my worker nodes look like after installing the cluster on AWS:

```
❯ kubectl get node -L topology.kubernetes.io/region -L topology.kubernetes.io/zone
NAME                                            STATUS   ROLES                  AGE     VERSION           REGION         ZONE
ip-10-0-155-166.eu-central-1.compute.internal   Ready    worker                 7h12m   v1.26.5+7d22122   eu-central-1   eu-central-1a
ip-10-0-169-238.eu-central-1.compute.internal   Ready    worker                 7h9m    v1.26.5+7d22122   eu-central-1   eu-central-1b
ip-10-0-205-76.eu-central-1.compute.internal    Ready    worker                 7h12m   v1.26.5+7d22122   eu-central-1   eu-central-1c
```

PTAL (cc @xperimental )

xperimental (commented):
I have prepared [a branch](https://github.com/shwetaap/loki/compare/Log3834...xperimental:loki:zone-awareness) containing all the changes I have commented on. Have a look.

The main changes are:
- Filter Pod reconciliations by label
- Move Pod changes to a single place (unfortunately needed to modify the existing code a bit, pulling out `podTemplate`)
- Move constants for annotations and labels into the API package

Unfortunately it took a while to prepare this, I thought there was a timing issue in there when setting the annotation on the Pod, but I was not able to reproduce this today anymore.

operator/internal/handlers/lokistack_enable_zone_awareness.go
```diff
@@ -0,0 +1,83 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	lokiv1 "github.com/grafana/loki/operator/apis/loki/v1"
+	"github.com/grafana/loki/operator/internal/external/k8s"
+)
+
+func AnnotatePodWithAvailabilityZone(ctx context.Context, log logr.Logger, c k8s.Client, pod *corev1.Pod) error {
```
xperimental, line 19: Does it make sense for this function to have unit-tests?

operator/internal/manifests/config.go
```diff
@@ -102,6 +102,8 @@ func ConfigOptions(opt Options) config.Options {
 		}
 	}
 
+	gossipRingconfig := gossipRingConfig(opt.Name, opt.Namespace, opt.Stack.HashRing, opt.Stack.Replication)
```
xperimental, line 105: This variable seems to be only used once. I think we can inline it.

operator/main.go
```diff
@@ -180,6 +180,14 @@ func main() {
 			os.Exit(1)
 		}
 	}
+
+	if err = (&lokictrl.LokiStackZoneAwarePodReconciler{
+		Client: mgr.GetClient(),
+		Log:    logger.WithName("controllers").WithName("lokistack-zoneaware-pod"),
+	}).SetupWithManager(mgr); err != nil {
```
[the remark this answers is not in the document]
JoaoBraveCoding: Yes, it doesn't apply. I was under the impression that you had to pass the scheme but maybe it was a blip in my memory

operator/internal/manifests/zone_aware_config.go
```diff
@@ -0,0 +1,52 @@
...
+var availabilityZoneEnvVar = corev1.EnvVar{
+	Name: availabilityZoneEnvVarName,
+	ValueFrom: &corev1.EnvVarSource{
+		FieldRef: &corev1.ObjectFieldSelector{
+			FieldPath: availabilityZoneFieldPath,
+		},
+	},
+}
+
+func configureReplication(podTemplate *corev1.PodTemplateSpec, replication *lokiv1.ReplicationSpec, component string, stackName string) {
```
JoaoBraveCoding, line 26: nit: added this comment to allow people to more easily understand what is the goal of this function
```suggestion
// configureReplication is responsible for setting up the TopologySpreadContraints of the podTemplate and adding
// the metadata (LabelZoneAwarePod and AnnotationAvailabilityZoneLabels) required by the
// ZoneAwarenessController to enable zone awareness behaviour.
func configureReplication(podTemplate *corev1.PodTemplateSpec, replication *lokiv1.ReplicationSpec, component string, stackName string) {
```

JoaoBraveCoding: WDYT of adding a unit test for this function?

JoaoBraveCoding (commented):
lgtm, just some unit test missing. Still have to re-run it on my side

operator/internal/handlers/lokistack_enable_zone_awareness.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, key := range labelKeys {
+		value, ok := nodeLabels[key]
+		if !ok {
+			return "", kverrors.New("scheduled node is missing label", "label", key)
+		}
+
+		labelValues = append(labelValues, value)
+	}
+
+	return strings.Join(labelValues, "_"), nil
```
shwetaap, line 82: Don't we want to maintain an order on how the most commonly used topology keys values get concatenated - region_zone_hostname?

xperimental: I don't think it's a good idea to have a pre-defined order in the operator, because we can not anticipate which keys will be used by the users. Region and zone at most of the public cloud providers are already hierarchical, so users will probably only want to use `zone` in that case. For non-public providers we do not know what labels will be used or what structure they will have.

The current code will use the order of the zones in the LokiStack as basis for the order of the instance-availability zone generated for the pods, this puts the user in control on what should happen. The thing that I forgot to add is the documentation in the API mentioning that the order of the zones in the array is significant.

operator/internal/handlers/lokistack_enable_zone_awareness_test.go
```diff
@@ -0,0 +1,115 @@
...
+	// apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	// "k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	// ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	defaultZones = []lokiv1.ZoneSpec{
```
xperimental, line 23: This variable is unused.

operator/internal/handlers/lokistack_enable_zone_awareness_test.go
```diff
@@ -0,0 +1,115 @@
...
+				lokiv1.AnnotationAvailabilityZoneLabels: corev1.LabelTopologyZone,
+				lokiv1.AnnotationAvailabilityZone:       "us-east-zone-2c",
+			},
+		},
+		Spec: corev1.PodSpec{
+			NodeName: "test-node",
+		},
+	}
+
+	defaultNode = corev1.Node{
```
xperimental, line 46: This variable is unused.

operator/internal/handlers/lokistack_enable_zone_awareness_test.go
```diff
@@ -0,0 +1,115 @@
...
+	badRequestErr := kverrors.New("failed to lookup node")
+	k.GetStub = func(ctx context.Context, name types.NamespacedName, object client.Object, _ ...client.GetOption) error {
+		return badRequestErr
+	}
+
+	err := handlers.AnnotatePodWithAvailabilityZone(context.TODO(), logger, k, &defaultPod)
+	require.Equal(t, badRequestErr, errors.Unwrap(err))
+
+	// make sure create was NOT called because the Get failed
+	require.Zero(t, k.CreateCallCount())
```
xperimental, line 70: `Create` will never be called by `AnnotatePodWithAvailabilityZone`:

```suggestion
	// make sure patch was NOT called because the Get failed
	require.Zero(t, k.PatchCallCount())
```

operator/internal/handlers/lokistack_enable_zone_awareness_test.go
```diff
@@ -0,0 +1,115 @@
...
+	k.GetStub = func(ctx context.Context, name types.NamespacedName, object client.Object, _ ...client.GetOption) error {
+		k.SetClientObject(object, &testNode)
+		return nil
+	}
+
+	err := handlers.AnnotatePodWithAvailabilityZone(context.TODO(), logger, k, &testPod)
+	require.NoError(t, err)
+
+	// make sure create was NOT called because the Get failed
+	require.Zero(t, k.CreateCallCount())
```
xperimental, line 114: This assertion does not add any value to the test. What should actually be tested is that Patch is called with the correct parameters.

operator/internal/manifests/zone_aware_config.go
shwetaap: Each component tests this function in tests similar to this - https://github.com/grafana/loki/pull/9503/files#diff-ead163e4b6e1d91440ff4f5bc70a1d6050832a328d299bff703cf1459766581dR110

periklis (commented):
Here is a branch to capture a couple of improvements for the comments below.
- Using `mergo.Merge` for `configureReplication`
- Add test for rendering the Loki Config with `instance_availability_zone`
- Add Changelog entry

https://github.com/shwetaap/loki/compare/Log3834...periklis:loki:zone_awareness

operator/internal/manifests/node_placement.go
```diff
@@ -0,0 +1,78 @@
...
+
+	topologyKey := strings.Join(zoneKeys, ",")
+	template.Annotations[lokiv1.AnnotationAvailabilityZoneLabels] = topologyKey
+
+	src := corev1.Container{
+		Env: []corev1.EnvVar{availabilityZoneEnvVar},
+	}
+
+	for i, dst := range podTemplate.Spec.Containers {
+		if err := mergo.Merge(&dst, src, mergo.WithAppendSlice); err != nil {
```
xperimental, line 67: I don't know why we're going with `mergo` again, especially when we're just adding one slice element. I can let it slide though ...

operator/internal/manifests/node_placement_test.go
```diff
@@ -524,3 +525,125 @@ func TestCustomPodAntiAffinity(t *testing.T) {
...
+	spec := &lokiv1.ReplicationSpec{
+		Factor: 2,
+		Zones: []lokiv1.ZoneSpec{
+			{
+				MaxSkew:     1,
+				TopologyKey: "region",
+			},
+			{
+				MaxSkew:     1,
+				TopologyKey: "zone",
```
xperimental, line 576: I think I would like to avoid using "region" and "zone" in our examples / tests when using two replication zones in the config. Just to avoid confusion with the correct configuration in public cloud offerings (where region often is already a prefix of zone).

My suggestion would be to pick "datacenter" and "rack" instead to have something "relatable but not real".

xperimental (approved):

PR #9504, opened 2023-05-23
revert count unflused chunks

MasslessParticle:
revert this commit: https://github.com/grafana/loki/commit/aeba51ab0fe785ab251eaf8d7005f203ece0986f

We seem to be dramatically overcounting bytes now which is causing oversharding

This is the same query before and after:

Before: ~20 GB
image (2)

After: ~8 EB
image (1)

### DIFF
--- a/pkg/ingester/instance.go
+++ b/pkg/ingester/instance.go
@@ -569,6 +569,9 @@ func (i *instance) GetStats(ctx context.Context, req *logproto.IndexStatsRequest
 	from, through := req.From.Time(), req.Through.Time()
 
 	if err = i.forMatchingStreams(ctx, from, matchers, nil, func(s *stream) error {
+		// checks for equality against chunk flush fields
+		var zeroValueTime time.Time
+
 		// Consider streams which overlap our time range
 		if shouldConsiderStream(s, from, through) {
 			s.chunkMtx.RLock()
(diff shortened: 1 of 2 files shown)


DylanGuedes (approved):

pkg/storage/stores/tsdb/index/chunk.go
```diff
@@ -245,8 +245,27 @@ func (cs *ChunkStats) addRaw(chunks int, kb, entries uint32) {
...
+	// We want to get the percentage of time that fits into C
+	// to use it as a factor to get the amount of bytes and entries
+	// factor = C = (T - (A + B)) / T = (chunkTime - (leadingTime + trailingTime)) / chunkTime
+	chunkTime := chk.MaxTime - chk.MinTime
+	leadingTime := math.Max64(0, from-chk.MinTime)
+	trailingTime := math.Max64(0, chk.MaxTime-through)
+	factor := float32(chunkTime-(leadingTime+trailingTime)) / float32(chunkTime)
+
+	kb := uint32(float32(chk.KB) * factor)
+	entries := uint32(float32(chk.Entries) * factor)
```
DylanGuedes, line 269: are you sure this part has to be reverted?

DylanGuedes (changes requested):

DylanGuedes (approved):

PR #9518, opened 2023-05-25
Fix wrong error check

simonswine:
I am afraid I broke main with #9502. It seem to have auto-merged with failing tests :cry:


MichelHollands (approved):
LGTM

PR #9535, opened 2023-05-26
Stats cache can be configured independently and reports metrics/stats for the cache

salvacorts:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

**Checklist**
- [ ] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`
- [ ] For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)

PR #9537, opened 2023-05-26
Add config to avoid caching stats for recent data

salvacorts:

When we query the stats for recent data, we query both the ingesters and the index gateways for the stats.
https://github.com/grafana/loki/blob/ebdb2b18007e024d56105afc5230383165ca1650/pkg/storage/async_store.go#L112-L114

https://github.com/grafana/loki/blob/ebdb2b18007e024d56105afc5230383165ca1650/pkg/storage/async_store.go#L126-L127

Then we merge all the responses, which means summing up all the stats

https://github.com/grafana/loki/blob/ebdb2b18007e024d56105afc5230383165ca1650/pkg/storage/async_store.go#L157-L158

https://github.com/grafana/loki/blob/ebdb2b18007e024d56105afc5230383165ca1650/pkg/storage/stores/index/stats/stats.go#L23-L26

Because we have a replication factor of 3, this means that we will get the stats from the ingesters repeated up to 3 times, hence inflating the stats.

In the stats cache, we store the stats for a given matcher set for the whole day, then we extract the stats from the cache by the factor of time from the request that is stored in the cache:
https://github.com/grafana/loki/blob/336283acadb34f5fda9abce4e6fcef1dca9965d8/pkg/querier/queryrange/index_stats_cache.go#L33
https://github.com/grafana/loki/blob/336283acadb34f5fda9abce4e6fcef1dca9965d8/pkg/querier/queryrange/index_stats_cache.go#L40

Inflated stats for recent data will be cached, so subsequent stats extracted from the cache will be inflated regardless of the time.

This PR adds a new per-tenant limit `max_stats_cache_freshness` to not cache requests with an end time that falls within Now minus this duration.

Here's a scenario illustrating this. The graphs below show the bytes stats queried in the sharding middleware. We are running a log filter query that won't match any log, every 5 seconds with a length of 3h.

image

As can be seen, after enabling the stats cache and configuring`do_not_cache_request_within` to not cache stats for requests within 30m, the bytes stats used in the sharding middleware stopped increasing.

In both cases the stats cache hit ration was 100%.
image

- Blocked by https://github.com/grafana/loki/pull/9535
- Note that this PR doesn't fix the root issue of inflated stats form the ingesters, but rather buys us some time to work on that.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/querier/queryrange/index_stats_cache.go
+++ b/pkg/querier/queryrange/index_stats_cache.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"time"
 
 	"github.com/go-kit/log"
 
@@ -11,6 +12,7 @@ import (
 	"github.com/grafana/loki/pkg/querier/queryrange/queryrangebase"
 	"github.com/grafana/loki/pkg/storage/chunk/cache"
 	"github.com/grafana/loki/pkg/util"
+	"github.com/prometheus/common/model"
 )
 
 type IndexStatsSplitter struct {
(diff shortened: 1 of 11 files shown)


pkg/querier/queryrange/index_stats_cache.go
```diff
@@ -50,18 +52,35 @@ func (p IndexStatsExtractor) ResponseWithoutHeaders(resp queryrangebase.Response
 
 type IndexStatsCacheConfig struct {
 	queryrangebase.ResultsCacheConfig `yaml:",inline"`
+	// NOTE: We cannot use something similar to the per-tenant `max_cache_freshness_per_query` limit
+	// because resultsCache.filterRecentExtents would extract already inflated stats.
+	// Instead, we need to filter out whole requests that ask for stats in recent data.
+	DoNotCacheRequestWithin time.Duration `yaml:"do_not_cache_request_within"`
```
owen-d, line 58: Can we make this a tenant config, `max_index_cache_freshness`?

salvacorts: Done

owen-d (approved):

salvacorts:
Rebased the branch to main. If CI is happy we should be ready to merge once the docs team gives their thumb up.

JStickler (approved):
[docs team]  LGTM

PR #9560, opened 2023-05-30
Set unique timestamp in lambda-promtail.

jeschkies:
Loki runs into performance issues when too many log entries have the same timestamp. Ironically our own lambda-promtail can be a cause for such data.

In this change we modify the logic a little in case lambda-promtail does not find a timestamp in the source data. Instead of adding the same for all entries it sets a new one.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/tools/lambda-promtail/lambda-promtail/s3.go
+++ b/tools/lambda-promtail/lambda-promtail/s3.go
@@ -12,8 +12,9 @@ import (
 	"github.com/aws/aws-lambda-go/events"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
-	"github.com/grafana/loki/pkg/logproto"
 	"github.com/prometheus/common/model"
+
+	"github.com/grafana/loki/pkg/logproto"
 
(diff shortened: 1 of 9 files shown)


JordanRushing (approved):

tools/lambda-promtail/lambda-promtail/s3.go
```diff
@@ -131,11 +131,14 @@ func parseS3Log(ctx context.Context, b *batch, labels map[string]string, obj io.
...
 
 		match := timestampRegex.FindStringSubmatch(log_line)
+		var timestamp time.Time
 		if len(match) > 0 {
 			timestamp, err = time.Parse(time.RFC3339, match[1])
 			if err != nil {
 				return err
 			}
+		} else {
+			timestamp = time.Now()
```
cstyan, line 141: we could simplify further by just having `timestamp = time.Now()` as the first line of the loop, and declare `var timestamp time.Time` immediately before the loop

removes the need for an `else`

cstyan (approved):

PR #9561, opened 2023-05-30
feat: add lokitool

theSuess:
This PR adds `lokitool` - a command line interface for the ruler API. It is lifted from `cortextool` (github.com/grafana/cortex-tools) while removing everything cortex-related and focusing on Loki.
The purpose behind this change is to avoid confusion due to:
+ loki not being cortext
+ mimir replacing cortex
+ mimirtool not working with loki

Having `lokitool` in this repository will also simplify adding loki specific features without interfering with mimir/cortex functionality.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


slim-bean:
Sort of a fundamentals question:

Should this be a separate tool or should we make it part of `logcli`

I think I would personally vote to make this part of `logcli` so we don't have to manage another binary as well as all the build requirements for it.

dannykopping:
> Sort of a fundamentals question:
> 
> Should this be a separate tool or should we make it part of `logcli`
> 
> I think I would personally vote to make this part of `logcli` so we don't have to manage another binary as well as all the build requirements for it.

That an interesting point @slim-bean, I think I agree with you
I think it'll complicate the PR though; should we start by importing the cortextool code (PS: are there any licensing issues we need to be sensitive to?) and in a subsequent PR do the refactor?

theSuess:
Since the linter wants to force me to modernize the code anyway, I'll see how hard integrating this with `logcli` is

theSuess:
> Since the linter wants to force me to modernize the code anyway, I'll see how hard integrating this with `logcli` is

Update: It's not that easy, as it would require a complete refactor of the `logcli` code. The current `logcli` code layout makes heavy use of global flags which do not make sense in the rules context, but are reused across the different query types (e.g. `stdin`, `retries` and `quiet`).

A "clean" command and code layout (e.g. `logcli query [--instant|--series|--labels]`) would be a breaking change.

In my humble opinion, having two tools makes sense: `logcli` is for dealing with log data while `lokitool` is for loki administration.

dannykopping:
> In my humble opinion, having two tools makes sense: `logcli` is for dealing with log data while `lokitool` is for loki administration.

Seems like a reasonable conclusion based on the above.
I'm not strongly against having another tool.

@slim-bean what do you think?

dannykopping:
@MasslessParticle was closing this intentional?

MasslessParticle:
Was cleaning up old drafts and didn't realize this one still had active conversation. Sorry!

theSuess:
Implemented in #12166

Issue #9581, opened 2023-05-31
Input query within size limits rewritten to SampleExpr exceeding size limits


ssncferreira:
**Describe the bug**

Upon receiving instant queries, the query string is parsed and converted to a SampleExpr: https://github.com/grafana/loki/blob/main/pkg/logql/syntax/parser.go#L141-L152 
Afterward, this SampleExpr is again converted to a string for cloning: https://github.com/grafana/loki/blob/main/pkg/logql/rangemapper.go#L458-L466 
However, the resulting query string can in some cases be different from the initial query string. This is the case when the query contains regex expressions. For performance improvements, the LogQL parser rewrites regex expressions to be non-greedy.

The query process flow `string -> expr -> string` is commonly used in query processing throughout the code. 

As a result, the initial query string may be within the size limits defined in the LogQL parser https://github.com/grafana/loki/blob/main/pkg/logql/syntax/parser.go#L34, but the resultant string conversion from SampleExpr can have a different size, and ultimately exceeding this limit.

This is the case reported in the query example below.

**To Reproduce**
1. Started the latest version of Loki
2. Running the following query as an instant query:
```
[...]

PR #9584, opened 2023-05-31
Regex line/label filters return original regex when `String` is called

MasslessParticle:
To optimize regex line and label filters, Loki rewrites them to be non-greedy. This can cause extra characters to appear in the regex when `String()` is called. In addition to confusing logging, this can also cause internal panics when a previously-parsable query becomes too long.

This PR stores the original regex in the label/line filter and uses that in calls to `String`.

This PR also removes regex optimizations from label selectors. Making the optimization at parse time caused the original regex to be lost.

### DIFF
--- a/pkg/logql/matchers_test.go
+++ b/pkg/logql/matchers_test.go
@@ -32,7 +32,7 @@ func Test_match(t *testing.T) {
 				{mustMatcher(labels.MatchEqual, "a", "1")},
 				{
 					mustMatcher(labels.MatchEqual, "b", "2"),
-					mustMatcher(labels.MatchEqual, "c", "3"),
+					mustMatcher(labels.MatchRegexp, "c", "3"),
 					mustMatcher(labels.MatchNotEqual, "d", "4"),
 				},
 			},
--- a/pkg/logql/syntax/ast.go
+++ b/pkg/logql/syntax/ast.go
@@ -705,30 +705,10 @@ func (l *LogfmtExpressionParser) String() string {
 }
 
 func mustNewMatcher(t labels.MatchType, n, v string) *labels.Matcher {
-	if t == labels.MatchRegexp || t == labels.MatchNotRegexp {
-		return simplifyRegexMatcher(t, n, v)
-	}
-
 	m, err := labels.NewMatcher(t, n, v)
 	if err != nil {
 		panic(logqlmodel.NewParseError(err.Error(), 0, 0))
 	}
-	return m
-}
-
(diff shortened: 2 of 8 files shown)


pkg/logql/syntax/ast.go
```diff
@@ -705,33 +705,13 @@ func (l *LogfmtExpressionParser) String() string {
...
-	}
-
 	m, err := labels.NewMatcher(t, n, v)
 	if err != nil {
 		panic(logqlmodel.NewParseError(err.Error(), 0, 0))
 	}
 	return m
 }
 
-func simplifyRegexMatcher(typ labels.MatchType, name, value string) *labels.Matcher {
```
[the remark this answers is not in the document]
MasslessParticle: It was happening in two places. This is called whenever the grammar recognized a `matcher` https://github.com/grafana/loki/blob/033b504aa70efe7abdebcce9e376ab6c0c12e7cf/pkg/logql/syntax/expr.y#L251-L256 and was too soon.

The second place it's called is when a filter is created [here](https://github.com/grafana/loki/blob/033b504aa70efe7abdebcce9e376ab6c0c12e7cf/pkg/logql/log/filter.go#L407). That does the optimization [here](https://github.com/grafana/loki/blob/033b504aa70efe7abdebcce9e376ab6c0c12e7cf/pkg/logql/log/filter.go#L437) which is where we want to keep the original.

pkg/logql/matchers_test.go
```diff
@@ -32,7 +32,7 @@ func Test_match(t *testing.T) {
 				{mustMatcher(labels.MatchEqual, "a", "1")},
 				{
 					mustMatcher(labels.MatchEqual, "b", "2"),
-					mustMatcher(labels.MatchEqual, "c", "3"),
+					mustMatcher(labels.MatchRegexp, "c", "3"),
```
ssncferreira, line 35: Curious about how this was passing before 🤔 Is it because the matcher was being rewritten to `c="3"`?

ssncferreira (approved):
LGTM 🥇 thanks for looking into this

ssncferreira:
Related to issue: https://github.com/grafana/loki/issues/9581

pkg/logql/matchers_test.go
MasslessParticle: It is. When the value for a label is `=~<literal>`, that's equivalent to `=<literal>` because matchers anchor the regex to the beginning and end of the line with `^()$`

PR #9616, opened 2023-06-02
Loki ECS Client: Update Name in Options config

sarah-spang:
Configuration options show  "Name": "grafana-loki", however, fluentbit specifies this should just be "loki"

A few users have reported receiving an error when using the documented "Name" option, but it works as expected when using the Name specified in the fluentbit docs:

Reference: https://docs.fluentbit.io/manual/pipeline/outputs/loki#fluent-bit-+-grafana-cloud


JStickler (approved):
[docs team] LGTM

PR #9621, opened 2023-06-04
Introduces cache to TSDB postings

DylanGuedes:
Introduces a new interface for fetching TSDB postings and implements a cached TSDB postings.
- The new postings interface is named `PostingsReader` and contains two implementations: a simple one, which is identical to the former Postings behavior, and a new one, that caches postings using the existing `index_queries_cache_config`. By default, the simple/former implementation is used. The new cached implementation can be configured in the following way:
```yaml
storage_config:
  tsdb_shipper:
    cache_postings: true
```
In the future we'll remove the `cache_postings` flag/configuration and change the behavior to cache postings by default.

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/storage/chunk/cache/cache.go
```diff
@@ -145,6 +146,18 @@ func New(cfg Config, reg prometheus.Registerer, logger log.Logger, cacheType sta
 		}
 	}
 
+	if cfg.LRUCache.Enabled {
+		cache, err := NewLRUCache(cfg.Prefix+"embedded-cache", cfg.LRUCache, reg, logger, cacheType)
```
chaudum, line 150: I assume "embedded cache" is used for the FIFO implementation.

```suggestion
		cache, err := NewLRUCache(cfg.Prefix+"inmemory-lru-cache", cfg.LRUCache, reg, logger, cacheType)
```

pkg/storage/chunk/cache/cache.go
```diff
@@ -145,6 +146,18 @@ func New(cfg Config, reg prometheus.Registerer, logger log.Logger, cacheType sta
...
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to initialize LRU cache", "err", err)
+			return nil, err
+		}
+
+		if cache != nil {
+			caches = append(caches, CollectStats(Instrument(cfg.Prefix+"embedded-cache", cache, reg)))
```
chaudum, line 157: Same here

pkg/storage/chunk/cache/lru_cache.go
```diff
@@ -0,0 +1,249 @@
...
+const (
+	codecHeaderSnappy             codec = "dvs" // As in "diff+varint+snappy".
+	codecHeaderSnappyWithMatchers codec = "dm"  // As in "dvs+matchers"
+)
+
+var DefaultLRUCacheConfig = LRUCacheConfig{
+	MaxSizeBytes: "250MB",
+}
+
+const maxInt = int(^uint(0) >> 1)
```
chaudum, line 31: Why not use `math.MaxInt`?

pkg/storage/chunk/cache/lru_cache.go
```diff
@@ -0,0 +1,249 @@
...
+
+type LRUCacheConfig struct {
+	MaxSizeBytes string `yaml:"max_size_bytes"`
+
+	Enabled bool `yaml:"enabled"`
+}
+
+// RegisterFlagsWithPrefix adds the flags required to config this to the given FlagSet
+func (cfg *LRUCacheConfig) RegisterFlagsWithPrefix(prefix, description string, f *flag.FlagSet) {
+	f.StringVar(&cfg.MaxSizeBytes, prefix+"fifocache.max-size-bytes", "500MB", description+"Maximum memory size of the cache in bytes. A unit suffix (KB, MB, GB) may be applied.")
```
chaudum, line 48:
```suggestion
	f.StringVar(&cfg.MaxSizeBytes, prefix+"lrucache.max-size-bytes", "500MB", description+"Maximum memory size of the cache in bytes. A unit suffix (KB, MB, GB) may be applied.")
```

pkg/storage/chunk/cache/lru_cache.go
```diff
@@ -0,0 +1,249 @@
...
+type codec string
+
+const (
+	codecHeaderSnappy             codec = "dvs" // As in "diff+varint+snappy".
+	codecHeaderSnappyWithMatchers codec = "dm"  // As in "dvs+matchers"
+)
+
+var DefaultLRUCacheConfig = LRUCacheConfig{
+	MaxSizeBytes: "250MB",
+}
```
chaudum, line 29: Is this anywhere used?

pkg/storage/chunk/cache/lru_cache.go
```diff
@@ -0,0 +1,249 @@
...
+	maxItemSizeBytes uint64
+
+	evicted     *prometheus.CounterVec
+	requests    *prometheus.CounterVec
+	hits        *prometheus.CounterVec
+	totalMisses prometheus.Counter
+	added       *prometheus.CounterVec
+	current     *prometheus.GaugeVec
+	bytesInUse  prometheus.Gauge
+	overflow    *prometheus.CounterVec
```
chaudum, line 75: I assume these metrics are used by all different cache implementations. Would it make sense to move them into a separate struct so they can be re-used?

pkg/storage/chunk/cache/lru_cache.go
```diff
@@ -0,0 +1,249 @@
...
+		maxSizeBytes: maxSizeBytes,
+		logger:       logger,
+
+		done: make(chan struct{}),
+	}
+
+	c.totalMisses = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Namespace:   "querier",
+		Subsystem:   "cache",
+		Name:        "misses_total",
```
chaudum, line 95: Please check the metrics instantiation for consistent usage of Namespace, Subsystem, Name field usages.

pkg/storage/stores/tsdb/cached_postings_index.go
```diff
@@ -0,0 +1,186 @@
...
+		matchersLen += varint.UvarintSize(uint64(len(m.Name)))
+		matchersLen += len(m.Name)
+		matchersLen++ // 1 byte for the type
+		matchersLen += varint.UvarintSize(uint64(len(m.Value)))
+		matchersLen += len(m.Value)
+	}
+	return matchersLen
+}
+
+func (c *cachedPostingsClient) storePostings(postings index.Postings, matchers []*labels.Matcher) error {
```
chaudum, line 117: Could you pass the canonical label matchers key directly?

pkg/storage/stores/tsdb/cached_postings_index.go
```diff
@@ -0,0 +1,186 @@
...
+	dataToCache, err := diffVarintEncodeNoHeader(postings, 0)
+	if err != nil {
+		level.Warn(c.log).Log("msg", "couldn't encode postings", "err", err, "matchers", CanonicalLabelMatchersKey(matchers))
+	}
+
+	return c.cacheClient.Store(context.TODO(), []string{CanonicalLabelMatchersKey(matchers)}, [][]byte{dataToCache})
```
chaudum, line 123: Not sure if we want to use the `TODO()` context ;-)

pkg/storage/stores/tsdb/single_file_index.go
```diff
@@ -114,26 +116,60 @@ func (f *TSDBFile) Reader() (io.ReadSeeker, error) {
...
 // method but is helpful for building an io.reader for the index shipper
-func NewTSDBIndexFromFile(location string) (*TSDBIndex, GetRawFileReaderFunc, error) {
+func NewTSDBIndexFromFile(location string) (Index, GetRawFileReaderFunc, error) {
 	reader, err := index.NewFileReader(location)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	return NewTSDBIndex(reader), func() (io.ReadSeeker, error) {
+	tsdbIdx := NewTSDBIndex(reader, getPostingsClient(reader))
```
chaudum, line 132: I'm not sure whether passing a postingsClient to the TSDBIndex is the best approach.

```suggestion
	tsdbIdx := NewTSDBIndex(reader)
	tsdbIndex = NewCachedTSDBIndex(tsdbIndex)
```

I would have rather thought about inlineing the existing *TSDBIndex in a cached version, since `*TSDBIndex` is just an interface implementation of `IndexReader` and `RequestChunkFilterer`. Extend it with `PostingsClient`:

```go
type CachedTSDBIndex struct {
	*TSDBIndex
    PostingsClient
	cache Cache
}

func NewCachedTSDBIndex(tsdbIndex *TSDBIndex, backend Cache) *CachedTSDBIndex {
	return &CachedTSDBIndex{
		TSDBIndex: tsdbIndex,
		cache: backend,
	}
}
```

Then you can implement `ForPosting` from the `PostingsClient` interface.

I haven't completely thought it through, though.

pkg/storage/stores/tsdb/single_file_index.go
```diff
@@ -187,20 +223,6 @@ func (i *TSDBIndex) ForSeries(ctx context.Context, shard *index.ShardAnnotation,
...
-	_, _ model.Time,
-	matchers []*labels.Matcher,
-	fn func(index.Postings) error,
-) error {
-	p, err := PostingsForMatchers(i.reader, shard, matchers...)
-	if err != nil {
-		return err
-	}
-	return fn(p)
-}
```
chaudum, line 202: I would extract this function into an interface (you already have the `PostingsClient` interface).

pkg/storage/stores/tsdb/cached_postings_index.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, p.Err()
+	}
+
+	return buf.B, nil
+}
+
+func decodeToPostings(b []byte) index.Postings {
+	decoder := encoding.DecWrap(promEncoding.Decbuf{B: b})
+
+	refs := []storage.SeriesRef{}
```
chaudum, line 94: This is kinda expensive if we don't know the size of the refs upfront. Maybe using a pool?

chaudum (commented):
So far, I haven't looked into the LRU implementation too much in detail, but rather focussed on the overall integration.

pkg/storage/chunk/cache/lru_cache.go
DylanGuedes: took that code from Thanos/Mimir. Not sure why they didn't use math.MaxInt.

pkg/storage/chunk/cache/lru_cache.go
DylanGuedes: Not really. I think it was one of those parts I took from Mimir/Thanos that were necessary for something that Loki doesn't need.

pkg/storage/chunk/cache/lru_cache.go
DylanGuedes: Agreed. I checked after your suggestion and I made it more consistent with the existing FIFO cache, but FYI: it isn't super consistent to other caches (like redis/memcached).

I also removed all references to "index" and "querier" from this implementation since this LRU cache can be used by any Loki subsystem, not necessarily the index.

pkg/storage/stores/tsdb/cached_postings_index.go
DylanGuedes: I found a workaround doing the following:
- When encoding to store on the cache, first add the number of seriesRefs
- When decoding, fetch the first uint64, it should be the number of series in the postings. This way we can alloc the right number of refs straightaway

pkg/storage/stores/tsdb/cached_postings_index.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/storage"
+	promEncoding "github.com/prometheus/prometheus/tsdb/encoding"
+
+	"github.com/grafana/loki/pkg/storage/chunk/cache"
+	"github.com/grafana/loki/pkg/storage/stores/tsdb/index"
+	"github.com/grafana/loki/pkg/util/encoding"
+)
+
+type PostingsClient interface {
```
chaudum, line 21: Suggestion: Maybe name it `PostingsReader` or `PostingsIterable`? Wdyt?

pkg/storage/stores/tsdb/single_file_index.go
```diff
@@ -18,6 +18,8 @@ import (
 	util_log "github.com/grafana/loki/pkg/util/log"
 )
 
+var shouldCachePostings = false
```
chaudum, line 21: I don't think that we should use a module scope variable to define what postings client/reader should be used.

chaudum: I understand that it's pretty nasty to pass down the config to the constructor, though.

DylanGuedes: It is an ugly workaround, but the other option is changing the signature everywhere to expect the cacheClient (but notice that the cacheClient would only be used here in the single_file_index). I'm fine with either of them tbh.

pkg/storage/stores/tsdb/cached_postings_index.go
DylanGuedes: I like `PostingsReader`, let me change it.

pkg/storage/stores/tsdb/single_file_index.go
chaudum: I had a quick look at an approach to pass the config down to the constructor, and it seems it could be done using the opening function. Full diff:


```diff
diff --git a/pkg/storage/stores/tsdb/cached_postings_index.go b/pkg/storage/stores/tsdb/cached_postings_index.go
index 769f0a1b0..dc5c3baf6 100644
--- a/pkg/storage/stores/tsdb/cached_postings_index.go
+++ b/pkg/storage/stores/tsdb/cached_postings_index.go
@@ -24,15 +24,15 @@ type PostingsReader interface {
 
 var sharedCacheClient cache.Cache
 
-func NewCachedPostingsClient(reader IndexReader, logger log.Logger, cacheClient cache.Cache) PostingsReader {
-	return &cachedPostingsClient{
+func NewCachedPostingsReader(reader IndexReader, logger log.Logger, cacheClient cache.Cache) PostingsReader {
+	return &cachedPostingsReader{
 		reader:      reader,
 		cacheClient: cacheClient,
 		log:         logger,
 	}
 }
 
-type cachedPostingsClient struct {
+type cachedPostingsReader struct {
 	reader IndexReader
 
 	cacheClient cache.Cache
@@ -40,7 +40,7 @@ type cachedPostingsClient struct {
 	log log.Logger
 }
 
-func (c *cachedPostingsClient) ForPostings(ctx context.Context, matchers []*labels.Matcher, fn func(index.Postings) error) error {
+func (c *cachedPostingsReader) ForPostings(ctx context.Context, matchers []*labels.Matcher, fn func(index.Postings) error) error {
 	key := CanonicalLabelMatchersKey(matchers)
 	if postings, got := c.fetchPostings(ctx, key); got {
 		return fn(postings)
@@ -123,7 +123,7 @@ func encodedMatchersLen(matchers []*labels.Matcher) int {
 	return matchersLen
 }
 
-func (c *cachedPostingsClient) storePostings(ctx context.Context, postings index.Postings, canonicalMatchers string) error {
+func (c *cachedPostingsReader) storePostings(ctx context.Context, postings index.Postings, canonicalMatchers string) error {
 	dataToCache, err := diffVarintEncodeNoHeader(postings, 0)
 	if err != nil {
 		level.Warn(c.log).Log("msg", "couldn't encode postings", "err", err, "matchers", canonicalMatchers)
@@ -132,7 +132,7 @@ func (c *cachedPostingsClient) storePostings(ctx context.Context, postings index
 	return c.cacheClient.Store(ctx, []string{canonicalMatchers}, [][]byte{dataToCache})
 }
 
-func (c *cachedPostingsClient) fetchPostings(ctx context.Context, key string) (index.Postings, bool) {
+func (c *cachedPostingsReader) fetchPostings(ctx context.Context, key string) (index.Postings, bool) {
 	found, bufs, _, err := c.cacheClient.Fetch(ctx, []string{key})
 
 	if err != nil {
diff --git a/pkg/storage/stores/tsdb/cached_postings_index_test.go b/pkg/storage/stores/tsdb/cached_postings_index_test.go
index 85b8a6885..3bc2d97c8 100644
--- a/pkg/storage/stores/tsdb/cached_postings_index_test.go
+++ b/pkg/storage/stores/tsdb/cached_postings_index_test.go
@@ -78,7 +78,7 @@ func TestSingleIdxCached(t *testing.T) {
 					_, _ = head.Append(x.Labels, x.Labels.Hash(), x.Chunks)
 				}
 				reader := head.Index()
-				return NewTSDBIndex(reader, DefaultPostingsClient(reader))
+				return NewTSDBIndex(reader, NewPostingsReader(reader))
 			},
 		},
 	} {
diff --git a/pkg/storage/stores/tsdb/compactor.go b/pkg/storage/stores/tsdb/compactor.go
index 928252e02..1e9c1ff02 100644
--- a/pkg/storage/stores/tsdb/compactor.go
+++ b/pkg/storage/stores/tsdb/compactor.go
@@ -36,7 +36,8 @@ func (i indexProcessor) NewTableCompactor(ctx context.Context, commonIndexSet co
 }
 
 func (i indexProcessor) OpenCompactedIndexFile(ctx context.Context, path, tableName, userID, workingDir string, periodConfig config.PeriodConfig, logger log.Logger) (compactor.CompactedIndex, error) {
-	indexFile, err := OpenShippableTSDB(path)
+	opts := TSDBIndexOpts{UsePostingsCache: false}
+	indexFile, err := OpenShippableTSDB(path, opts)
 	if err != nil {
 		return nil, err
 	}
@@ -92,6 +93,8 @@ func (t *tableCompactor) CompactTable() error {
 	multiTenantIndices := make([]Index, len(multiTenantIndexes))
 	downloadPaths := make([]string, len(multiTenantIndexes))
 
+	opts := TSDBIndexOpts{UsePostingsCache: false}
+
 	// concurrently download and open all the multi-tenant indexes
 	err := concurrency.ForEachJob(t.ctx, len(multiTenantIndexes), readDBsConcurrency, func(ctx context.Context, job int) error {
 		downloadedAt, err := t.commonIndexSet.GetSourceFile(multiTenantIndexes[job])
@@ -100,7 +103,7 @@ func (t *tableCompactor) CompactTable() error {
 		}
 
 		downloadPaths[job] = downloadedAt
-		idx, err := OpenShippableTSDB(downloadedAt)
+		idx, err := OpenShippableTSDB(downloadedAt, opts)
 		if err != nil {
 			return err
 		}
@@ -205,6 +208,7 @@ func setupBuilder(ctx context.Context, userID string, sourceIndexSet compactor.I
 		}
 	}
 
+	opts := TSDBIndexOpts{UsePostingsCache: false}
 	// download all the existing compacted indexes and add them to the builder
 	for _, sourceIndex := range sourceIndexes {
 		path, err := sourceIndexSet.GetSourceFile(sourceIndex)
@@ -218,7 +222,7 @@ func setupBuilder(ctx context.Context, userID string, sourceIndexSet compactor.I
 			}
 		}()
 
-		indexFile, err := OpenShippableTSDB(path)
+		indexFile, err := OpenShippableTSDB(path, opts)
 		if err != nil {
 			return nil, err
 		}
@@ -387,7 +391,7 @@ func (c *compactedIndex) ToIndexFile() (index_shipper.Index, error) {
 		return nil, err
 	}
 
-	return NewShippableTSDBFile(id)
+	return NewShippableTSDBFile(id, TSDBIndexOpts{UsePostingsCache: false})
 }
 
 func getUnsafeBytes(s string) []byte {
diff --git a/pkg/storage/stores/tsdb/head_manager.go b/pkg/storage/stores/tsdb/head_manager.go
index c725cee4e..135fb573f 100644
--- a/pkg/storage/stores/tsdb/head_manager.go
+++ b/pkg/storage/stores/tsdb/head_manager.go
@@ -731,7 +731,7 @@ func (t *tenantHeads) tenantIndex(userID string, from, through model.Time) (idx
 	}
 
 	reader := tenant.indexRange(int64(from), int64(through))
-	idx = NewTSDBIndex(reader, &simplePostingsClient{reader: reader})
+	idx = NewTSDBIndex(reader, &simplePostingsReader{reader: reader})
 	if t.chunkFilter != nil {
 		idx.SetChunkFilterer(t.chunkFilter)
 	}
diff --git a/pkg/storage/stores/tsdb/manager.go b/pkg/storage/stores/tsdb/manager.go
index c5b970e46..7c685077e 100644
--- a/pkg/storage/stores/tsdb/manager.go
+++ b/pkg/storage/stores/tsdb/manager.go
@@ -129,7 +129,7 @@ func (m *tsdbManager) Start() (err error) {
 			indices++
 
 			prefixed := NewPrefixedIdentifier(id, filepath.Join(mulitenantDir, bucket), "")
-			loaded, err := NewShippableTSDBFile(prefixed)
+			loaded, err := NewShippableTSDBFile(prefixed, TSDBIndexOpts{UsePostingsCache: false})
 
 			if err != nil {
 				level.Warn(m.log).Log(
@@ -221,7 +221,7 @@ func (m *tsdbManager) buildFromHead(heads *tenantHeads, shipper indexshipper.Ind
 
 		level.Debug(m.log).Log("msg", "finished building tsdb for period", "pd", p, "dst", dst.Path(), "duration", time.Since(start))
 
-		loaded, err := NewShippableTSDBFile(dst)
+		loaded, err := NewShippableTSDBFile(dst, TSDBIndexOpts{UsePostingsCache: false})
 		if err != nil {
 			return err
 		}
diff --git a/pkg/storage/stores/tsdb/single_file_index.go b/pkg/storage/stores/tsdb/single_file_index.go
index 3d1e5b6bf..26412ffe5 100644
--- a/pkg/storage/stores/tsdb/single_file_index.go
+++ b/pkg/storage/stores/tsdb/single_file_index.go
@@ -27,17 +27,22 @@ var ErrAlreadyOnDesiredVersion = errors.New("tsdb file already on desired versio
 // GetRawFileReaderFunc returns an io.ReadSeeker for reading raw tsdb file from disk
 type GetRawFileReaderFunc func() (io.ReadSeeker, error)
 
-func OpenShippableTSDB(p string) (index_shipper.Index, error) {
+type TSDBIndexOpts struct {
+	UsePostingsCache bool
+}
+
+func OpenShippableTSDB(p string, opts TSDBIndexOpts) (index_shipper.Index, error) {
 	id, err := identifierFromPath(p)
 	if err != nil {
 		return nil, err
 	}
 
-	return NewShippableTSDBFile(id)
+	return NewShippableTSDBFile(id, opts)
 }
 
 func RebuildWithVersion(ctx context.Context, path string, desiredVer int) (index_shipper.Index, error) {
-	indexFile, err := OpenShippableTSDB(path)
+	opts := TSDBIndexOpts{UsePostingsCache: false}
+	indexFile, err := OpenShippableTSDB(path, opts)
 	if err != nil {
 		return nil, err
 	}
@@ -76,7 +81,7 @@ func RebuildWithVersion(ctx context.Context, path string, desiredVer int) (index
 	if err != nil {
 		return nil, err
 	}
-	return NewShippableTSDBFile(id)
+	return NewShippableTSDBFile(id, opts)
 }
 
 // nolint
@@ -92,8 +97,8 @@ type TSDBFile struct {
 	getRawFileReader GetRawFileReaderFunc
 }
 
-func NewShippableTSDBFile(id Identifier) (*TSDBFile, error) {
-	idx, getRawFileReader, err := NewTSDBIndexFromFile(id.Path())
+func NewShippableTSDBFile(id Identifier, opts TSDBIndexOpts) (*TSDBFile, error) {
+	idx, getRawFileReader, err := NewTSDBIndexFromFile(id.Path(), opts)
 	if err != nil {
 		return nil, err
 	}
@@ -125,42 +130,43 @@ type TSDBIndex struct {
 
 // Return the index as well as the underlying raw file reader which isn't exposed as an index
 // method but is helpful for building an io.reader for the index shipper
-func NewTSDBIndexFromFile(location string) (Index, GetRawFileReaderFunc, error) {
+func NewTSDBIndexFromFile(location string, opts TSDBIndexOpts) (Index, GetRawFileReaderFunc, error) {
 	reader, err := index.NewFileReader(location)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	tsdbIdx := NewTSDBIndex(reader, getPostingsClient(reader))
+	postingsReader := getPostingsReader(reader, opts.UsePostingsCache)
+	tsdbIdx := NewTSDBIndex(reader, postingsReader)
 
 	return tsdbIdx, func() (io.ReadSeeker, error) {
 		return reader.RawFileReader()
 	}, nil
 }
 
-func getPostingsClient(reader IndexReader) PostingsReader {
-	var postingsClient PostingsReader
+func getPostingsReader(reader IndexReader, usePostingsCache bool) PostingsReader {
+	var pr PostingsReader
 
-	if shouldCachePostings && sharedCacheClient != nil {
-		postingsClient = NewCachedPostingsClient(reader, util_log.Logger, sharedCacheClient)
+	if usePostingsCache && sharedCacheClient != nil {
+		pr = NewCachedPostingsReader(reader, util_log.Logger, sharedCacheClient)
 	}
 
-	if postingsClient == nil {
-		postingsClient = DefaultPostingsClient(reader)
+	if pr == nil {
+		pr = NewPostingsReader(reader)
 	}
 
-	return postingsClient
+	return pr
 }
 
-func DefaultPostingsClient(reader IndexReader) PostingsReader {
-	return &simplePostingsClient{reader: reader}
+func NewPostingsReader(reader IndexReader) PostingsReader {
+	return &simplePostingsReader{reader: reader}
 }
 
-type simplePostingsClient struct {
+type simplePostingsReader struct {
 	reader IndexReader
 }
 
-func (s *simplePostingsClient) ForPostings(_ context.Context, matchers []*labels.Matcher, fn func(index.Postings) error) error {
+func (s *simplePostingsReader) ForPostings(_ context.Context, matchers []*labels.Matcher, fn func(index.Postings) error) error {
 	p, err := PostingsForMatchers(s.reader, nil, matchers...)
 	if err != nil {
 		return err
diff --git a/pkg/storage/stores/tsdb/single_file_index_test.go b/pkg/storage/stores/tsdb/single_file_index_test.go
index f9c5988f3..ed8b4ae26 100644
--- a/pkg/storage/stores/tsdb/single_file_index_test.go
+++ b/pkg/storage/stores/tsdb/single_file_index_test.go
@@ -82,7 +82,7 @@ func TestSingleIdx(t *testing.T) {
 					_, _ = head.Append(x.Labels, x.Labels.Hash(), x.Chunks)
 				}
 				reader := head.Index()
-				return NewTSDBIndex(reader, DefaultPostingsClient(reader))
+				return NewTSDBIndex(reader, NewPostingsReader(reader))
 			},
 		},
 	} {
diff --git a/pkg/storage/stores/tsdb/store.go b/pkg/storage/stores/tsdb/store.go
index d848216c9..079a77b41 100644
--- a/pkg/storage/stores/tsdb/store.go
+++ b/pkg/storage/stores/tsdb/store.go
@@ -21,6 +21,7 @@ import (
 	"github.com/grafana/loki/pkg/storage/stores/index"
 	"github.com/grafana/loki/pkg/storage/stores/indexshipper"
 	"github.com/grafana/loki/pkg/storage/stores/indexshipper/downloads"
+	indexshipper_index "github.com/grafana/loki/pkg/storage/stores/indexshipper/index"
 	tsdb_index "github.com/grafana/loki/pkg/storage/stores/tsdb/index"
 )
 
@@ -32,6 +33,7 @@ type store struct {
 	index.Reader
 	indexShipper      indexshipper.IndexShipper
 	indexWriter       IndexWriter
+	indexReadCache    cache.Cache
 	backupIndexWriter index.Writer
 	logger            log.Logger
 	stopOnce          sync.Once
@@ -62,6 +64,7 @@ func NewStore(
 	storeInstance := &store{
 		backupIndexWriter: backupIndexWriter,
 		logger:            logger,
+		indexReadCache:    idxCache,
 	}
 
 	if err := storeInstance.init(name, indexShipperCfg, schemaCfg, objectClient, limits, tableRange, reg, idxCache); err != nil {
@@ -74,10 +77,9 @@ func NewStore(
 func (s *store) init(name string, indexCfg IndexCfg, schemaCfg config.SchemaConfig, objectClient client.ObjectClient,
 	limits downloads.Limits, tableRange config.TableRange, reg prometheus.Registerer, idxCache cache.Cache) error {
 
-	sharedCacheClient = idxCache
-
-	if indexCfg.CachePostings {
-		shouldCachePostings = true
+	usePostingsCache := indexCfg.Mode == indexshipper.ModeReadOnly && idxCache != nil
+	openFn := func(p string) (indexshipper_index.Index, error) {
+		return OpenShippableTSDB(p, TSDBIndexOpts{UsePostingsCache: usePostingsCache})
 	}
 
 	var err error
@@ -86,7 +88,7 @@ func (s *store) init(name string, indexCfg IndexCfg, schemaCfg config.SchemaConf
 		objectClient,
 		limits,
 		nil,
-		OpenShippableTSDB,
+		openFn,
 		tableRange,
 		prometheus.WrapRegistererWithPrefix("loki_tsdb_shipper_", reg),
 		s.logger,
diff --git a/tools/tsdb/tsdb-map/main_test.go b/tools/tsdb/tsdb-map/main_test.go
index ee12cf720..4d4908529 100644
--- a/tools/tsdb/tsdb-map/main_test.go
+++ b/tools/tsdb/tsdb-map/main_test.go
@@ -93,7 +93,7 @@ func BenchmarkQuery_GetChunkRefs(b *testing.B) {
 		if err != nil {
 			panic(err)
 		}
-		idx := tsdb.NewTSDBIndex(reader, tsdb.DefaultPostingsClient(reader))
+		idx := tsdb.NewTSDBIndex(reader, tsdb.NewPostingsReader(reader))
 		b.Run(bm.name, func(b *testing.B) {
 			refs := tsdb.ChunkRefsPool.Get()
 			for i := 0; i < b.N; i++ {
@@ -118,7 +118,7 @@ func BenchmarkQuery_GetChunkRefsSharded(b *testing.B) {
 		if err != nil {
 			panic(err)
 		}
-		idx := tsdb.NewTSDBIndex(reader, tsdb.DefaultPostingsClient(reader))
+		idx := tsdb.NewTSDBIndex(reader, tsdb.NewPostingsReader(reader))
 		shardFactor := 16
 
 		b.Run(bm.name, func(b *testing.B) {
```

DylanGuedes:
Benchmark results:
Non-cached:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -tags integration,requires_docker -bench ^BenchmarkTSDBIndex_GetChunkRefs$ github.com/grafana/loki/pkg/storage/stores/tsdb

goos: linux
goarch: amd64
BenchmarkTSDBIndex_GetChunkRefs-24    	     468	   2487220 ns/op	 6396621 B/op	      45 allocs/op
PASS
ok  	github.com/grafana/loki/pkg/storage/stores/tsdb	1.530s
```

Cached:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -tags integration,requires_docker -bench ^BenchmarkCacheableTSDBIndex_GetChunkRefs$ github.com/grafana/loki/pkg/storage/stores/tsdb

goos: linux
goarch: amd64
BenchmarkCacheableTSDBIndex_GetChunkRefs-24    	     704	   1706066 ns/op	 1410071 B/op	      31 allocs/op
PASS
ok  	github.com/grafana/loki/pkg/storage/stores/tsdb	1.481s

```

DylanGuedes:
Using `benchstat`:
```
benchstat old.txt new.txt
goos: linux
goarch: amd64
                                   │   old.txt    │       new.txt       │
                                   │    sec/op    │    sec/op     vs base   │
TSDBIndex_GetChunkRefs-24            2.468m ± ∞ ¹
CacheableTSDBIndex_GetChunkRefs-24                  1.586m ± ∞ ¹
geomean                              2.468m         1.586m        ? ² ³
¹ need >= 6 samples for confidence interval at level 0.95
² benchmark set differs from baseline; geomeans may not be comparable
³ ratios must be >0 to compute geomean

                                   │    old.txt    │       new.txt        │
                                   │     B/op      │     B/op       vs base   │
TSDBIndex_GetChunkRefs-24            6.100Mi ± ∞ ¹
CacheableTSDBIndex_GetChunkRefs-24                   1.344Mi ± ∞ ¹
geomean                              6.100Mi         1.344Mi        ? ² ³
¹ need >= 6 samples for confidence interval at level 0.95
² benchmark set differs from baseline; geomeans may not be comparable
³ ratios must be >0 to compute geomean

                                   │   old.txt   │      new.txt       │
                                   │  allocs/op  │  allocs/op   vs base   │
TSDBIndex_GetChunkRefs-24            45.00 ± ∞ ¹
CacheableTSDBIndex_GetChunkRefs-24                 31.00 ± ∞ ¹
geomean                              45.00         31.00        ? ² ³
¹ need >= 6 samples for confidence interval at level 0.95
² benchmark set differs from baseline; geomeans may not be comparable
³ ratios must be >0 to compute geomean

```

docs/sources/configuration/_index.md
```diff
@@ -3939,6 +3943,25 @@ embedded_cache:
...
+lru_cache:
+  # Maximum memory size of the whole cache. A unit suffix (KB, MB, GB) may be
+  # applied. Default: 1GB.
+  # CLI flag: -<prefix>.lrucache.max-size-bytes
+  [max_size_bytes: <int> | default = 0B]
+
+  # Maximum memory size of a single item in the cache. A unit suffix (KB, MB,
+  # GB) may be applied. Default: 100MB.
+  # CLI flag: -<prefix>.lrucache.max-item-size-bytes
+  [max_item_size_bytes: <int> | default = 0B]
```
JStickler, line 3955: Is the default 100BM (line 3953) or 0B (line 3955) ?

docs/sources/configuration/_index.md
```diff
@@ -3939,6 +3943,25 @@ embedded_cache:
   # CLI flag: -<prefix>.embedded-cache.ttl
   [ttl: <duration> | default = 1h]
 
+lru_cache:
```
JStickler, line 3946: Does LRU stand for anything?  Would it help people understand if it was defined somewhere?

JStickler (commented):
[docs team]

docs/sources/configuration/_index.md
DylanGuedes: It should be 100MB but the `make docs` task is incorrectly setting this to 0b :(

docs/sources/configuration/_index.md
DylanGuedes: Yep, it stands for last-recently used. I'll add a note about that.

owen-d (commented):
We should support headers in the same way mimir does, explained [here](https://github.com/grafana/mimir/blob/main/pkg/storegateway/postings_codec.go#L24-L38). Using no headers directly in our cache is dangerous because it makes adding new versions much more difficult.

pkg/storage/stores/tsdb/cached_postings_index.go
```diff
@@ -0,0 +1,194 @@
...
+	expandedPosts, err := index.ExpandPostings(p)
+	if err != nil {
+		return fmt.Errorf("failed to expand postings: %w", err)
+	}
+
+	if err := c.storePostings(ctx, expandedPosts, key); err != nil {
+		level.Error(c.log).Log("msg", "failed to cache postings", "err", err, "matchers", key)
+	}
+
+	// because `index.ExpandedPostings` walks with the iterator, we have to reset it current index by instantiating a new ListPostings.
```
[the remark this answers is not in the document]
DylanGuedes: `ExpandPostings` will walk the postings iterator.

pkg/storage/stores/tsdb/cached_postings_index.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to evaluate postings for matchers: %w", err)
+	}
+
+	expandedPosts, err := index.ExpandPostings(p)
+	if err != nil {
+		return fmt.Errorf("failed to expand postings: %w", err)
+	}
+
+	if err := c.storePostings(ctx, expandedPosts, key); err != nil {
```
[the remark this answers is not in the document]
DylanGuedes: good call, thank you.

pkg/storage/stores/tsdb/cached_postings_index.go
```diff
@@ -0,0 +1,194 @@
...
+	buf := encoding.Encbuf{}
+	buf.PutUvarint32(uint32(length))
+
+	// This encoding uses around ~1 bytes per posting, but let's use
+	// conservative 1.25 bytes per posting to avoid extra allocations.
+	if length > 0 {
+		buf.B = make([]byte, 0, binary.MaxVarintLen32+5*length/4)
+	}
+
+	buf.PutUvarint32(uint32(length)) // first we put the postings length used when decoding.
```
chaudum, line 94:
```suggestion
	if length == 0 {
	    return []byte{0}, nil
	}
	buf := encoding.EncWith(make([]byte, 0, binary.MaxVarintLen32+5*length/4))
	buf.PutUvarint32(uint32(length)) // first we put the postings length used when decoding.
```

pkg/storage/stores/tsdb/cached_postings_index.go
```diff
@@ -0,0 +1,194 @@
...
+	return buf.Get(), nil
+}
+
+func decodeToPostings(b []byte) index.Postings {
+	if len(b) <= 0 {
+		return index.EmptyPostings()
+	}
+
+	decoder := encoding.DecWrap(promEncoding.Decbuf{B: b})
+	postingsLen := decoder.Uvarint32()
```
chaudum, line 116: What if postingsLen == 0 ?

chaudum, line 116:
```suggestion
	postingsLen := decoder.Uvarint32()
	if postingsLen == 0 {
	    return index.EmptyPostings()
	}
```

pkg/storage/stores/tsdb/index.go
```diff
@@ -22,6 +24,18 @@ type ChunkRef struct {
 	Checksum    uint32
 }
 
+type IndexCfg struct {
+	indexshipper.Config `yaml:",inline"`
+
+	CachePostings bool `yaml:"cache_postings" category:"experimental"`
```
chaudum, line 30:
```suggestion
	EnablePostingsCache bool `yaml:"enable_postings_cache" category:"experimental"`
```

pkg/storage/stores/tsdb/index.go
```diff
@@ -22,6 +24,18 @@ type ChunkRef struct {
 	Checksum    uint32
 }
 
+type IndexCfg struct {
```
chaudum, line 27: Can we just name it `Config`?

pkg/storage/stores/tsdb/store.go
```diff
@@ -62,35 +65,43 @@ func NewStore(
...
-	limits downloads.Limits, tableRange config.TableRange, reg prometheus.Registerer) error {
+func (s *store) init(name string, indexCfg IndexCfg, schemaCfg config.SchemaConfig, objectClient client.ObjectClient,
+	limits downloads.Limits, tableRange config.TableRange, reg prometheus.Registerer, idxCache cache.Cache) error {
+
+	sharedCacheClient = idxCache
+
+	usePostingsCache := indexCfg.CachePostings && indexCfg.Mode == indexshipper.ModeReadOnly && idxCache != nil
+	openFn := func(p string) (indexshipper_index.Index, error) {
+		return OpenShippableTSDB(p, IndexOpts{UsePostingsCache: usePostingsCache})
+	}
```
chaudum, line 83: Another idea to avoid the module scoped cache instance: Could we pass the `idxCache` instead of `usePostingsCache`?

```suggestion
func (s *store) init(name string, indexCfg IndexCfg, schemaCfg config.SchemaConfig, objectClient client.ObjectClient,
	limits downloads.Limits, tableRange config.TableRange, reg prometheus.Registerer, idxCache cache.Cache) error {

	var cache.Cache sharedCache
	if indexCfg.CachePostings && indexCfg.Mode == indexshipper.ModeReadOnly && idxCache != nil {
	    sharedCache = idxCache
	}
	openFn := func(p string) (indexshipper_index.Index, error) {
		return OpenShippableTSDB(p, IndexOpts{UsePostingsCache: sharedCache})
	}
```

docs/sources/configuration/_index.md
```diff
@@ -2032,6 +2032,10 @@ tsdb_shipper:
   [mode: <string> | default = ""]
 
   [ingesterdbretainperiod: <duration>]
+
+  # Experimental. Whether TSDB should cache postings or not.
+  # CLI flag: -tsdb.cache-postings
+  [cache_postings: <boolean> | default = false]
```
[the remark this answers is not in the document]
DylanGuedes: pushed a commit addressing it, thank you.

pkg/storage/stores/tsdb/index.go
DylanGuedes: I think we can, but wouldn't it mislead you into thinking it is a general TSDB configuration? While this one is only relevant in the context of the index.

docs/sources/configure/_index.md
```diff
@@ -3946,6 +3950,25 @@ embedded_cache:
   # CLI flag: -<prefix>.embedded-cache.ttl
   [ttl: <duration> | default = 1h]
 
+lru_cache:
+  # Maximum memory size of the whole cache. A unit suffix (KB, MB, GB) may be
+  # applied. Default: 1GB.
+  # CLI flag: -<prefix>.lrucache.max-size-bytes
+  [max_size_bytes: <int> | default = 0B]
```
JStickler, line 3957: The defaults at line 3955 and line 3957 still don't match.  =(

Same issue with line 3960 and 3962.

Is there some other file that needs to be edited to update the defaults?

pkg/storage/stores/tsdb/single_file_index.go
```diff
@@ -114,26 +116,60 @@ func (f *TSDBFile) Reader() (io.ReadSeeker, error) {
...
+func DefaultPostingsClient(reader IndexReader) PostingsClient {
+	return &simplePostingsClient{reader: reader}
+}
+
+type simplePostingsClient struct {
+	reader IndexReader
+}
+
+func (s *simplePostingsClient) ForPostings(ctx context.Context, matchers []*labels.Matcher, fn func(index.Postings) error) error {
+	p, err := PostingsForMatchers(s.reader, nil, matchers...)
```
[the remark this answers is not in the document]
DylanGuedes: the underlying code basically ignores it (for the single_index_file at least), i think it is used by the head_index though (used by the ingester).

docs/sources/configure/_index.md
DylanGuedes: I found a way of fixing it, sorry for the delay :smile:

pkg/storage/stores/tsdb/cached_postings_index.go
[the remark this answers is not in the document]
DylanGuedes: added.

pkg/storage/stores/tsdb/cached_postings_index.go
DylanGuedes: applied your suggestion, thank you.

DylanGuedes:
> I think we should remove the LRUCache. We already have the FIFOCache which can be configured to evict when a size is reached, which gives us a similar enough functionality. Also, we already make extensive use of memcached which is also an LRU and likely more efficient than anything we can add (if the expense of a little network latency is tolerable).

That's fair and the PR was quite big. I pushed a commit stepping back from LRU; if we decide that it is necessary I can add it back.

docs/sources/configure/_index.md
```diff
@@ -2037,6 +2037,11 @@ tsdb_shipper:
   [mode: <string> | default = ""]
 
   [ingesterdbretainperiod: <duration>]
+
+  # Experimental. Whether TSDB should cache postings or not. The
+  # index-read-cache will be used as the backend.
+  # CLI flag: -tsdb.enable-cache-postings
+  [enable_cache_postings: <boolean> | default = false]
```
dannykopping, line 2044: Nit: this should be named `enable_postings_cache`, or simply `cache_postings`. In its current form it doesn't really make sense.

dannykopping (commented):
Thanks for removing it @DylanGuedes!
I'm not familiar with this part of the code so I won't be able to give it a meaningful review, but I did leave one naming nit as my gift to you 😆

docs/sources/configure/_index.md
DylanGuedes: hah thank you. pushed a commit addressing it.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ##### Enhancements
 
+* [9621](https://github.com/grafana/loki/pull/9621) **DylanGuedes**: Introduces TSDB postings cache.
```
chaudum, line 9: IMO imperative should be used for changelog entries, such as "Introduce", "Add", "Fix", ...

```suggestion
* [9621](https://github.com/grafana/loki/pull/9621) **DylanGuedes**: Introduce TSDB postings cache.
```

pkg/storage/factory.go
```diff
@@ -283,9 +284,9 @@ type Config struct {
 	DisableBroadIndexQueries bool         `yaml:"disable_broad_index_queries"`
 	MaxParallelGetChunk      int          `yaml:"max_parallel_get_chunk"`
 
-	MaxChunkBatchSize   int                 `yaml:"max_chunk_batch_size"`
-	BoltDBShipperConfig shipper.Config      `yaml:"boltdb_shipper" doc:"description=Configures storing index in an Object Store (GCS/S3/Azure/Swift/COS/Filesystem) in the form of boltdb files. Required fields only required when boltdb-shipper is defined in config."`
-	TSDBShipperConfig   indexshipper.Config `yaml:"tsdb_shipper"`
+	MaxChunkBatchSize   int            `yaml:"max_chunk_batch_size"`
+	BoltDBShipperConfig shipper.Config `yaml:"boltdb_shipper" doc:"description=Configures storing index in an Object Store (GCS/S3/Azure/Swift/COS/Filesystem) in the form of boltdb files. Required fields only required when boltdb-shipper is defined in config."`
+	TSDBShipperConfig   tsdb.IndexCfg  `yaml:"tsdb_shipper"`
```
chaudum, line 289: Please also add a `doc`, like the line above does.

pkg/storage/stores/tsdb/single_file_index.go
```diff
@@ -118,26 +125,61 @@ func (f *TSDBFile) Reader() (io.ReadSeeker, error) {
...
+
+	if postingsCache != nil {
+		pr = NewCachedPostingsReader(reader, util_log.Logger, postingsCache)
+	}
+
+	if pr == nil {
+		pr = NewPostingsReader(reader)
+	}
+
+	return pr
```
chaudum, line 160: Can be written shorter in more idiomatic Go:

```suggestion
	if postingsCache != nil {
		return NewCachedPostingsReader(reader, util_log.Logger, postingsCache)
	}
	return NewPostingsReader(reader)
```

pkg/storage/stores/tsdb/single_file_index.go
```diff
@@ -118,26 +125,61 @@ func (f *TSDBFile) Reader() (io.ReadSeeker, error) {
...
+}
+
+func NewPostingsReader(reader IndexReader) PostingsReader {
+	return &simplePostingsReader{reader: reader}
+}
+
+type simplePostingsReader struct {
```
chaudum, line 167: nit: Personally I don't like to use the word "simple" for objects. I'd use "default", since it's the default implementation of the `PostingsReader` interface.

```suggestion
type defaultPostingsReader struct {
```

pkg/storage/factory.go
DylanGuedes: added

integration/loki_micro_services_test.go
```diff
@@ -567,3 +567,120 @@ func TestSchedulerRing(t *testing.T) {
...
+
+	// end of setup. Ingestion and querying below.
+
+	t.Run("ingest-logs", func(t *testing.T) {
+		require.NoError(t, cliDistributor.PushLogLineWithTimestamp("lineA", time.Now().Add(-72*time.Hour), map[string]string{"job": "fake"}))
+		require.NoError(t, cliDistributor.PushLogLineWithTimestamp("lineB", time.Now().Add(-48*time.Hour), map[string]string{"job": "fake"}))
+	})
+
+	// restart ingester which should flush the chunks and index
+	require.NoError(t, tIngester.Restart())
```
trevorwhitney, line 651: we could also hit `/flush` here to be even more deliberate

integration/loki_micro_services_test.go
```diff
@@ -567,3 +567,120 @@ func TestSchedulerRing(t *testing.T) {
...
+
+	var (
+		tDistributor = clu.AddComponent(
+			"distributor",
+			"-target=distributor",
+		)
+		tIndexGateway = clu.AddComponent(
+			"index-gateway",
+			"-target=index-gateway",
+			"-tsdb.enable-postings-cache=true",
```
trevorwhitney, line 586: this test passes with this flag set to `false`. is there any way to this test to fail when caching is not happening? where are the objects coming from (local filesystem?), could we change the data at rest to prove the data we're getting is from cache?

pkg/storage/stores/tsdb/cached_postings_index.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/storage"
+	promEncoding "github.com/prometheus/prometheus/tsdb/encoding"
+
+	"github.com/grafana/loki/pkg/storage/chunk/cache"
+	"github.com/grafana/loki/pkg/storage/stores/tsdb/index"
+	"github.com/grafana/loki/pkg/util/encoding"
+)
+
+type PostingsReader interface {
```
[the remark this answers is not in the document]
trevorwhitney: `PostingsIterator`?

trevorwhitney (commented):
This looks great, I'm excited to see it land!

integration/loki_micro_services_test.go
DylanGuedes: I added an assertion that checks the cache added/misses/gets metrics results. they guarantee we're covering the cache behavior using the FIFO cache.

pkg/storage/stores/tsdb/cached_postings_index.go
```diff
@@ -0,0 +1,157 @@
...
+	return p, nil
+}
+
+// CanonicalLabelMatchersKey creates a canonical version of LabelMatchersKey
+func CanonicalLabelMatchersKey(ms []*labels.Matcher) string {
+	sorted := make([]labels.Matcher, len(ms))
+	for i := range ms {
+		sorted[i] = labels.Matcher{Type: ms[i].Type, Name: ms[i].Name, Value: ms[i].Value}
+	}
+	sort.Sort(sortedLabelMatchers(sorted))
```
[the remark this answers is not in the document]
DylanGuedes: `sorteableLabelMatchers` maybe?

integration/loki_micro_services_test.go
DylanGuedes: this part was just copied from the other tests

pkg/storage/stores/tsdb/cached_postings_index.go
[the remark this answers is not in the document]
DylanGuedes: hmm I think Iterator is misleading; it doesn't support the Iterator interface/pattern (.Next(), .At(), etc)

trevorwhitney (approved):
approving to unblock pending the few small changes left. great work Dylan!

PR #9626, opened 2023-06-05
logfmt: add --strict flag to enable strict parsing, perform nostrict parsing by default

ashwanthgoli:
Introduces `--strict` flag for logfmt parser to enable strict parsing.
With strict parsing enabled, logfmt parser stops scanning the log line and returns early with an error if it encounters any poorly formatted key=value pair.

This pr updates the parser to perform non-strict parsing by default where it skips over the invalid tokens. This offers the flexibility to parse semi-structured log lines.

Also adds `--keep-empty` flag to retain standalone keys(with empty value) as labels with value set to empty string

**Checklist**
- [X] Documentation added
- [X] Tests updated
- [x] `CHANGELOG.md` updated


docs/sources/query/log_queries/_index.md
```diff
@@ -463,6 +463,27 @@ The **logfmt** parser can operate in two modes:
     "fwd_ip" => "124.133.124.161"
     ```
 
+logfmt parser also supports `--strict` flag to enable strict parsing.
```
JStickler, line 466:
```suggestion
The logfmt parser also supports using the `--strict` flag to enable strict parsing.
```

docs/sources/query/log_queries/_index.md
```diff
@@ -463,6 +463,27 @@ The **logfmt** parser can operate in two modes:
...
+
+```
+| logfmt --strict
+| logfmt --strict host, fwd_ip="fwd"
+```
+
+With strict parsing enabled, logfmt parser stops scanning the log line and returns early with an error if it encounters any poorly formatted key=value pair.
```
JStickler, line 473:
```suggestion
With strict parsing enabled, the logfmt parser immediately stops scanning the log line and returns an error when it encounters any poorly formatted key=value pair.
```

docs/sources/query/log_queries/_index.md
```diff
@@ -463,6 +463,27 @@ The **logfmt** parser can operate in two modes:
...
+// accepted
+key=value key="value in double quotes" standalone_key
+
+// invalid
+=value // no key
+foo=bar=buzz
+fo"o=bar
+```
+
+By default, the parser skips the invalid tokens and continues parsing the rest of the log line.
```
JStickler, line 484: "By default" is unclear here, do you mean "when using the default settings" or more that skipping the invalid tokens and continuing to parse is the default behavior?

docs/sources/query/log_queries/_index.md
```diff
@@ -463,6 +463,27 @@ The **logfmt** parser can operate in two modes:
...
+This offers the flexibility to parse semi-structed log lines, though note that this is only best-effort and the resulting label set might not always contain what you expect.
```
JStickler, line 485: "This" is unclear, I'm not sure what you're referring to (the default behavior?)

JStickler (commented):
[docs team]  Couple of suggestions and questions.

docs/sources/query/log_queries/_index.md
ashwanthgoli: @JStickler I was referring to the latter. I have updated the doc to make this clear.

JStickler: The revision is much clearer.  Thanks!

JStickler (approved):
[docs team] LGTM

chaudum (dismissed):
I really would like to see an implementation that preserves the current behaviour.

As a user I do not want to write `| logfmt --strict` every time.

`--strict` means that the default is now "not strict", which could cause problems with people that do not have a strict `logfmt` logging format throughout. This is a breaking change and would also need to be handled as such.

This is not a super huge problem for users that have strict logfmt formatted logs everywhere, because the behaviour would not change, but for everyone else it poses a risk.

slim-bean:
> As a user I do not want to write `| logfmt --strict` every time.

The argument I'm making is exactly the opposite.

What happens if you get an error in logql 99% of the time? You write `| __error__ = ""` and run the query again.

I feel the behavior here would be similar: error parsing your log lines with logfmt? Re-run the query with `--no-strict`

Imagine the reverse now:

Your query returns without error if your log lines are properly structured or not, without having to modify and re-run your query.

As time goes on I find myself more of the opinion that Loki should do it's best without failure as the default mode, and if you would like it to be more strict you can opt into it. 

I believe the approach of making it default to non-strict will require less typing/shorter queries and this leads into the second topic about breaking changes.

I struggle to come up with uses cases where someone would actually prefer strict parsing of our parsers, I really can't think of why anyone would want to ignore lines completely because they don't parse without any errors. If anyone out there has such a use case please share!!! And if you do have that use case this is why we are adding the option to do strict parsing. But I don't think it makes sense for strict to be the default.

On the breaking part: my thoughts on this are: what scenario would making this change cause the most problems. 

It's been difficult for me to come up with many, the only one I've really been able to come up with is a user running a metric query like this `sum(rate({app="foo"} | logfmt | __error__=""[1m]))`  The result of this query could very likely change if there were lines being dropped from the result because they contained logfmt parse errors.

This is perhaps where I share another unpopular opinion: if we are going to chose to make this breaking change because we think it leads to a better user experience, I don't think it matters if we do it in a minor or major release so long as we clearly document it.

docs/sources/upgrading/_index.md
```diff
@@ -66,6 +66,13 @@ A new config option `-boltdb.shipper.compactor.delete-request-store` decides whe
 
 In the case where neither of these options are set, the `object_store` configured in the latest `period_config` that uses either a tsdb or boltdb-shipper index is used for storing delete requests to ensure pending requests are processed.
 
+#### logfmt parser non-strict parsing
+logfmt parser now performs non-strict parsing which helps scan semi-structured log lines.
+It skips invalid tokens and tries to extract as many key/value pairs as possible from the rest of the log line.
+If you have a use-case that relies on strict parsing where you expect the parser to throw an error, we suggest you use `| logfmt --strict` to enable strict mode.
```
MichelHollands, line 72:
```suggestion
If you have a use-case that relies on strict parsing where you expect the parser to throw an error, use `| logfmt --strict` to enable strict mode.
```
It's the only thing you can do.

MichelHollands (approved):
LGTM. Nice work!

PR #9650, opened 2023-06-07
config: ensure storage config defaults apply to named stores

ashwanthgoli:
Since named store config does not register any flags, storage configs defined under it do not get the defaults.
For example [aws_storage_config](https://grafana.com/docs/loki/latest/configuration/#aws_storage_config) sets the default `storage_class` to `STANDARD`, but the same doesn't get applied by default when using named stores.

This PR ensures that named storage configs are always assigned default values when they are unmarshalled by implementing `yaml.Unmarshaler` interface

**Checklist**
- [X] Tests updated
- [x] `CHANGELOG.md` updated


dannykopping (approved):
LGTM!

PR #9665, opened 2023-06-08
Adding background cache (en|de)queue counters

dannykopping:
The background writeback cache exposes gauge metric currently for the current queue size. Gauges can be useful, but they are susceptible to sample errors because they only represent the point in time as the time of the scrape.

Exposing counters for the bytes (en|de)queued to/from the cache will be more useful because they can be aggregated.


sandeepsukhani (approved):
LGTM

PR #9666, opened 2023-06-08
Update doc-validator to v3.0.0

jdbaldry:
### Added

- Structured output for use with [`reviewdog`](https://github.com/reviewdog/reviewdog).
  You can achieve the original error output by piping the output to the following `jq` expression: `jq -r '"ERROR: \(.location.path):\(.location.range.start.line):\(.location.range.start.column): \(.message)"'`.
- Suggestions for simple link fixes. In GitHub, [`reviewdog`](https://github.com/reviewdog/reviewdog) comments these suggestions for convenient replacement.
- Support for anchors referring to repeated headings.
  In the case that a page has multiple headings that share the same text, the renderer appends a zero indexed, numbered suffix to the identifier.
  For example, when there are two headings that are both "Heading text", the first anchor identifier is `heading-text` and the second is `heading-text-1`.
- Error when running `doc-validator` on no files.

**Notes for reviewer:**

This also updates the `make-docs` procedure. Please ensure you can run `make docs` from the `docs/` directory before approving and merging this PR.

### DIFF
--- a/.github/workflows/doc-validator.yml
+++ b/.github/workflows/doc-validator.yml
@@ -16,7 +16,6 @@ jobs:
       - name: "Run doc-validator tool"
         run: >
           doc-validator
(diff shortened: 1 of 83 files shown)


JStickler (approved):
[docs team] LGTM

PR #9668, opened 2023-06-08
Tpatterson/size based compaction with latest

MasslessParticle:
Another pr merged with main to get a better sense of the diff


pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -149,18 +151,19 @@ type Compactor struct {
...
 	DeleteRequestsHandler     *deletion.DeleteRequestHandler
 	DeleteRequestsGRPCHandler *deletion.GRPCRequestHandler
 	deleteRequestsManager     *deletion.DeleteRequestsManager
-	expirationChecker         retention.ExpirationChecker
-	metrics                   *metrics
-	running                   bool
-	wg                        sync.WaitGroup
-	indexCompactors           map[string]IndexCompactor
-	schemaConfig              config.SchemaConfig
+	//expirationChecker         retention.ExpirationChecker
```
MasslessParticle, line 159: Probably should delete this

pkg/loki/modules.go
```diff
@@ -1081,6 +1081,11 @@ func (t *Loki) initCompactor() (services.Service, error) {
 	// Set some config sections from other config sections in the config struct
 	t.Cfg.CompactorConfig.CompactorRing.ListenPort = t.Cfg.Server.GRPCListenPort
 
+	if !(config.UsingObjectStorageIndex(t.Cfg.SchemaConfig.Configs) || config.UsingBoltDBtorageIndex(t.Cfg.SchemaConfig.Configs)) {
```
MasslessParticle, line 1084: This is concerning. Why are we turning off index compaction when using the filesystem? It still provides a performance benefit, even if the chunks are local.

pkg/storage/stores/indexshipper/compactor/compactor.go
```diff
@@ -401,9 +443,114 @@ func (c *Compactor) starting(ctx context.Context) (err error) {
...
 
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
+func (c *Compactor) sizeBasedCompactionInterval(ctx context.Context) error {
+	if exceeded, err := c.sizeBasedRetention.ThresholdExceeded(); !exceeded {
```
MasslessParticle, line 454: This doesn't handle the error when `exceeded == true`

MasslessParticle (commented):
Sorry again for the delay to look at this. I can't run it because things have drifted but I have done a review. 

This is looking good and makes sense. I've added comments about a couple of concerns. I'm also concerned there there's a lot of logic here and very few tests. I notice that you merged my original suggestions. That's fair to avoid rework but also those were illustrative and should probably be tested

MasslessParticle:
I reviewed my own PR 🤦

PR #1145, opened 2019-10-11
Update `/loki/api/v1/push` to use the v1 json format

joe-elliott:
**What this PR does / why we need it**:
- Updates the json format accepted at `/loki/api/v1/push` to match other v1 endpoints
- Adds supporting unmarshalling structure and tests
- Updates documentation to the new format
- Reverts changes made to 

**Which issue(s) this PR fixes**:
Fixes #1142

**Special notes for your reviewer**:
Promtail, docker driver and fluentbit clients should all continue to operate normally as they all use protobuf which has not been touched.  An issue (https://github.com/grafana/loki/issues/1144) was filed to update the fluentd client.

**Checklist**
- [x] Documentation added
- [x] Tests updated

PR #9694, opened 2023-06-13
Add metadata to push payload

salvacorts:
We are adding support for attaching labels to each log line. This is one of the series of the PRs broken up to make it easier to review changes.

This PR updates the push payload to send labels with each log entry optionally. The log labels are supposed to be in the same format as the stream labels. Just to put it out, here is how it would look for proto and json push payload with same data:

**proto(`endpoint`: `(/loki/api/v1/push|/api/prom/push)`, `Content-Type`: `application/x-protobuf`)**(payload built using [push.Stream](https://github.com/grafana/loki/blob/4cd1246b8830ccc241fa4afff85d208dc6ae2129/pkg/push/types.go#L12)):
```
push.Stream{
	Entries: []logproto.Entry{
			{
				Timestamp: time.Unix(0, 1688515200000000000),
				Line:      "log line",
				Labels:    `{foo="bar"}`,
			},
	},
	Labels: `{app="test"}`,
}
```

**v1(`endpoint`: `/loki/api/v1/push`, `Content-Type`: `application/json`)**:
```json
{
    "streams": [{
        "stream": {
            "app": "test"
        },
        "values": [
            ["1688515200000000000", "log line", {
                "foo": "bar"
            }]
        ]
    }]
}
```
**legacy-json(`/api/prom/push`, `Content-Type`: `application/json`)**:
```json
{
    "streams": [{
        "labels": "{app=\"test\"}",
        "entries": [{
                "ts": "2023-07-05T00:00:00.000000000Z",
                "line": "log line",
                "labels": "{foo=\"bar\"}"
            }]
    }]
}
```

We may need to add more thoughtful tests.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


integration/client/client.go
```diff
@@ -117,6 +117,11 @@ type stream struct {
 func (c *Client) pushLogLine(line string, timestamp time.Time, metadata labels.Labels, extraLabelList ...map[string]string) error {
 	apiEndpoint := fmt.Sprintf("%s/loki/api/v1/push", c.baseURL)
 
+	metadataJson, err := metadata.MarshalJSON()
```
salvacorts: Can you elaborate on why we need this to have json format?

pkg/loghttp/entry.go
```diff
@@ -18,6 +18,7 @@ func init() {
 type Entry struct {
 	Timestamp time.Time
 	Line      string
+	Labels    string
```
owen-d, line 21: Making this a string feels counterintuitive compared to using `map[string]string`, which is what we use via the [`LabelSet`](https://github.com/grafana/loki/blob/main/pkg/loghttp/labels.go#L21-L22) struct used in our v1 push payloads in [`loghttp/push`](https://github.com/grafana/loki/blob/main/pkg/loghttp/push/push.go#L104). Point being, I think we should use JSON directly (not json with embedded stringified json fields) when `application/json` is the `Content-Type`. When snappy-encoding is used, `logproto` can use the string variants for consistency with how it treats labels.

owen-d (commented):
In the future, please add a description about the PR to help reviewers. A bit of history + the design choices taken and the implications is sufficient and makes it much easier to know what I'm looking at.

As for the PR, I think JSON content types should support json sub-structure like we do for stream labels in https://grafana.com/docs/loki/latest/api/#push-log-entries-to-loki

```json
{
  "streams": [
    {
      "stream": {
        "label": "value"
      },
      "values": [
          [ "<unix epoch in nanoseconds>", "<log line>", {"traceID": "foo"} ],
          [ "<unix epoch in nanoseconds>", "<log line>", {"order_id": "bar"} ]
      ]
    }
  ]
}
```

vendor/github.com/grafana/loki/pkg/push/push.proto
```diff
@@ -37,4 +37,8 @@ message EntryAdapter {
     (gogoproto.jsontag) = "ts"
   ];
   string line = 2 [(gogoproto.jsontag) = "line"];
+  // TODO: maybe we can use the labels.Labels type here?
+  //       So we can avoid the conversion to and from string.
+  //       Leaving it as string for now to be consistent with `labels` field in `StreamAdapter`.
+  string labels = 3 [(gogoproto.jsontag) = "labels,omitempty"];
```
vlad-diachenko, line 43: let's use labels here instead of string

vlad-diachenko (commented):
Overall looks great )

pkg/loghttp/entry.go
```diff
@@ -126,6 +143,8 @@ func (EntryEncoder) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
 	stream.WriteRaw(`"`)
 	stream.WriteMore()
 	stream.WriteStringWithHTMLEscaped(e.Line)
+	stream.WriteMore()
+	stream.WriteString(e.Labels)
```
owen-d, line 147: On a second pass, I notice this _decodes_ a json object of labels, but _encodes_  a string. This feels wrong (it will be unusual to see stringified labels in a json response but send them as an object). If we want to avoid doing the parsing until we're ready to send a response (for logproto compatibility), we could keep the representation in memory as a string, but we should definitely be able to do recursively `encode -> decode -> encode -> decode -> etc` the same data.

pkg/loghttp/entry.go
salvacorts: The problem I see with making this a map, and keeping as a string the labels for an entry in logproto, is that converting from `loghttp.Entry` to `logproto.Entry` is going to slower.

Right now (both with string labels), we just cast the pointer for the entries array of a stream since they have the same memory layout.
https://github.com/grafana/loki/blob/44b0932247909e5d196087b88ec15ae00ad7254d/pkg/loghttp/query.go#L236

If we decide to make one a map and another one a string, we won't be able to do that cast anymore (different memory layouts). Hence, we'll need to go entry by entry converting it from `loghttp.Entry` to `logproto.Entry`.

IMO, we should make both use the same layout.

salvacorts: I implemented this in https://github.com/grafana/loki/pull/9694/commits/f43dfab7febf8784e6624ea3459826ac0ec0309b

This is what we need to do instead of the cast:
https://github.com/grafana/loki/blob/f43dfab7febf8784e6624ea3459826ac0ec0309b/pkg/loghttp/query.go#L236-L239

If we decide to keep this as a string, we can revert the commit. If we decide to make logproto.Entry to use a map for labels, we can build on top of it.

pkg/loghttp/query.go
```diff
@@ -55,9 +55,122 @@ func (q *QueryResponse) UnmarshalJSON(data []byte) error {
...
+		if err != nil {
+			parseError = err
+			return
+		}
+		entries = append(entries, e)
+	})
+	if parseError != nil {
+		return nil, parseError
+	}
+	if err != nil {
```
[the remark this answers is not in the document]
salvacorts: The error handling of the implementation for `jsonparser.ArrayEach` doesn't seems correct to me. The error being passed to the callback is weird as it'll always be nil. Then, after calling the callback, the library checks for the value of the error, which will always be nil since it's passed by value to the callback. There's an issue open for this: 
https://github.com/buger/jsonparser/issues/255

IIUC, `parseError` can be nil while `err` can be non-nil as [it can return for example a MalformedJsonError][1].
Having said that, if `err` is not nil, we should return `err`, not `parseError`.


[1]: https://github.com/grafana/loki/blob/fed9115627973beb8394c7bfd7c0333d5e3c0502/vendor/github.com/buger/jsonparser/parser.go#L971

pkg/loghttp/query.go
```diff
@@ -55,9 +55,122 @@ func (q *QueryResponse) UnmarshalJSON(data []byte) error {
...
+			v, err := jsonparser.ParseString(value)
+			if err != nil {
+				parseError = err
+				return
+			}
+			e.Line = v
+		case 2: // labels
+			labels := make(LabelSet)
+			err := jsonparser.ObjectEach(value, func(key, val []byte, _ jsonparser.ValueType, _ int) error {
+				labels[yoloString(key)] = yoloString(val)
```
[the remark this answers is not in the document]
salvacorts: Done with 384d556

pkg/loghttp/query.go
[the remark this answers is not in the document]
salvacorts: Done with 384d556

vendor/github.com/grafana/loki/pkg/push/push.proto
salvacorts: Domne with 4088112

pkg/loghttp/entry.go
```diff
@@ -143,8 +177,15 @@ func (EntryEncoder) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
...
 	stream.WriteMore()
 	stream.WriteStringWithHTMLEscaped(e.Line)
-	stream.WriteMore()
-	stream.WriteString(e.Labels)
+	if len(e.Labels) > 0 {
+		stream.WriteMore()
+		stream.WriteObjectStart()
+		for lName, lValue := range e.Labels {
+			stream.WriteObjectField(lName)
+			stream.WriteString(lValue)
```
vlad-diachenko, line 185: we need to add `stream.WriteMore()` after this line to add coma between label<>values. except for the last label<>value.

vlad-diachenko (commented):
please update this error message https://github.com/grafana/loki/pull/9694/files/617e8641e9bfe8a94f9acb71abecd93902ccfc4b..4088112f418372a5ae459084165a8c98f43a25e1#diff-467d1278563679e03528ad3f9c2b52dfccc49243ec2039be8e1a205969e131f1R135

pkg/loghttp/entry.go
salvacorts: Thank you! Fixed and tested with 0b0af55dc0eacf86e583416e5e802d5b883d2f99

vlad-diachenko: thanks

pkg/loghttp/entry.go
vlad-diachenko: according to our API docs, we use JSON object for stream labels )https://grafana.com/docs/loki/latest/api/#push-log-entries-to-loki
so, it makes sense to use JSON object for line metadata also… but you are right @salvacorts that there is no need to convert it to the string when we convert `loghttp.Entry` to `logproto.Entry` ….

pkg/loghttp/query.go
```diff
@@ -55,9 +55,122 @@ func (q *QueryResponse) UnmarshalJSON(data []byte) error {
 	})
 }
 
-// PushRequest models a log stream push
+// PushRequest models a log stream push but is unmarshalled to proto push format.
 type PushRequest struct {
-	Streams []*Stream `json:"streams"`
+	Streams []LogProtoStream `json:"streams"`
```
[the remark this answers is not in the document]
salvacorts: This comes from this PR:
https://github.com/grafana/loki/pull/1145

The only reason why I think this is a pointer is so we don't make a copy of the stream when calling NewStream which receives a ptr to `loghttp.Stream`
https://github.com/grafana/loki/pull/1145/files#diff-71619e1c80a73b34eade235a55d012a0ddbb3375b8d4ac89c1f4fd672145b915R34

With Sandeep's changes, we no longer use that since now we decode from json to `loghttp.PushRequest` to then cast it to `logproto.PushRequest`
https://github.com/grafana/loki/blob/44b0932247909e5d196087b88ec15ae00ad7254d/pkg/util/unmarshal/unmarshal.go#L22

So having that as a ptr or not shouldn't make much difference.

pkg/loghttp/query.go
```diff
@@ -55,9 +55,122 @@ func (q *QueryResponse) UnmarshalJSON(data []byte) error {
...
+}
+
+// LogProtoStream helps with unmarshalling of each log stream for push request.
+// This might look un-necessary but without it the CPU usage in benchmarks was increasing by ~25% :shrug:
+type LogProtoStream struct {
```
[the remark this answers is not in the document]
salvacorts: I agree!

pkg/loghttp/query.go
```diff
@@ -55,9 +55,128 @@ func (q *QueryResponse) UnmarshalJSON(data []byte) error {
...
```
salvacorts, line 65: nit: I think the following definition is more idiomatic
```go
type LogProtoStream logproto.Stream
```

sandeepsukhani (approved):
LGTM

vendor/github.com/grafana/loki/pkg/push/push.proto
```diff
@@ -37,4 +37,8 @@ message EntryAdapter {
     (gogoproto.jsontag) = "ts"
   ];
   string line = 2 [(gogoproto.jsontag) = "line"];
+  // TODO: maybe we can use the labels.Labels type here?
```
vlad-diachenko, line 40: let's remove this todo

vendor/github.com/grafana/loki/pkg/push/push.proto
```diff
@@ -37,4 +37,8 @@ message EntryAdapter {
...
+  //       So we can avoid the conversion to and from string.
+  //       Leaving it as string for now to be consistent with `labels` field in `StreamAdapter`.
```
vlad-diachenko, line 42:
```suggestion
```

pkg/push/push.proto
```diff
@@ -37,4 +37,8 @@ message EntryAdapter {
     (gogoproto.jsontag) = "ts"
   ];
   string line = 2 [(gogoproto.jsontag) = "line"];
+  // TODO: maybe we can use the labels.Labels type here?
+  //       So we can avoid the conversion to and from string.
+  //       Leaving it as string for now to be consistent with `labels` field in `StreamAdapter`.
```
vlad-diachenko, line 42:
```suggestion
```

vlad-diachenko (approved):
Looks brilliant

slim-bean:
Two things I want to chat about before we merge this. One is naming, I don't think we should name the new value `labels`

Having `labels` multiple times in a payload where they end up doing and being stored in very different ways I think is confusing.

I think we should consider naming it `metadata` or perhaps our own spin on this `metalabels` or `MetaLabels` ??

The second thing is in the API's themselves.

In the proto:
```
Entries: []logproto.Entry{
			{
				Timestamp: time.Unix(0, 1688515200000000000),
				Line:      "log line",
				Labels:    `{foo="bar"}`,
			},
	},
	Labels: `{app="test"}`,
```

I don't know that I love forcing the string type with our labels syntax here? would it be better to use a map type?

I need to dig into the json a bit, I'm surprised to see we have a different format for those endpoints?

pkg/loghttp/entry.go
owen-d: @salvacorts You're right, making this anything but a `string` here will result in performance losses because this it would need to parse labels potentially on _every entry_ coming back from subqueries. Since this happens before we filter out unnecessary entries (over `limit`, etc), this can quickly result in higher resource utilization, OOMs, etc. 

Now, we have a few options:
1) Keep this as a `string` in memory, but turn it into a `map` when return the `json` response from the http api. This will maintain performance (no conversion from protos) & correctly return a map via the json api.
2) Use `LabelSet` for this field (stored as a `map`). This requires parsing strings -> json for potentially _every_ log line returned from queriers (yikes!). Even when combined with (4), this would require transforming `slice -> map`.
3) Use `labels.Labels` (stored as a `slice`). When combined with (4) this incurs no transformation cost and the resulting conversion to the json map for the external API is simple & relatively inexpensive.
4) Due to the performance implications here, I've changed my opinion on making [`EntryAdapter.labels`](https://github.com/grafana/loki/blob/main/pkg/push/push.proto#L33) a `string` for parity with [`StreamAdapter.labels`](https://github.com/grafana/loki/blob/main/pkg/push/push.proto#L24). Instead, I think we should use the [`LabelAdapter`](https://github.com/grafana/loki/blob/main/pkg/logproto/logproto.proto#L337) pattern to send the labels in the protos as a slice and combine it with (3) to not need to pay conversion costs.

Does this make sense?

salvacorts: Agree. Implemented it in https://github.com/grafana/loki/pull/9694/commits/40287588c7fb79c93e45d17e6ca4dedb69119d52. We now use `labels.Labels` in both `loghttp.Entry` and `logproto.Entry`.

I wonder if we should take care of duplicated labels. I think may not (?) provided the push payload in json is an object so it should have duplicated keys (is that somehow enforced?).

Note that [proto3 supports maps][1]. In the wire, the map is the same as an array so it needs extra logic to have the map API. I ran a benchmark and looks like maps compared to arrays are ~x3 slower to marshal and ~x2 slower to unmarshal (happy to share/discuss the benchmark). Therefore I agree the way to go is using arrays in protobuf.

[1]: https://protobuf.dev/programming-guides/proto3/#maps

salvacorts:
The CI is failing in the check-mod stage. I think it's due to now requiring the prometheus libs in the vendor pkg/push. If I run `go mod vendor`, it updates a ton of dependencies, making this PR huge. Any ideas how to avoid this?

vendor/github.com/grafana/loki/pkg/push/types.go
```diff
@@ -17,8 +21,48 @@ type Stream struct {
 
 // Entry is a log entry with a timestamp.
 type Entry struct {
-	Timestamp time.Time `protobuf:"bytes,1,opt,name=timestamp,proto3,stdtime" json:"ts"`
-	Line      string    `protobuf:"bytes,2,opt,name=line,proto3" json:"line"`
+	Timestamp        time.Time     `protobuf:"bytes,1,opt,name=timestamp,proto3,stdtime" json:"ts"`
+	Line             string        `protobuf:"bytes,2,opt,name=line,proto3" json:"line"`
+	NonIndexedLabels labels.Labels `protobuf:"bytes,3,opt,name=labels,proto3" json:"labels,omitempty"`
```
sandeepsukhani, line 26: I think it would be better to keep it consistent
```suggestion
	NonIndexedLabels labels.Labels `protobuf:"bytes,3,opt,name=nonIndexedLabels,proto3" json:"nonIndexedLabels,omitempty"`
```

pkg/loghttp/query.go
```diff
@@ -441,3 +561,7 @@ func labelVolumeLimit(r *http.Request) error {
 
 	return nil
 }
+
+func yoloString(buf []byte) string {
+	return *((*string)(unsafe.Pointer(&buf)))
+}
```
vlad-diachenko, line 567: the scariest thing in the PR ) we should use it only if we check that the buffer is not updated, however, a lot of functions in stack trace mean that somebody can do some optimization to reuse the slice and it will break all our implementation. Why do we need it here?

vlad-diachenko (approved):
LGTM

pkg/loghttp/entry.go
```diff
@@ -27,25 +29,49 @@ func (e *Entry) UnmarshalJSON(data []byte) error {
...
+				return
+			}
+			var nonIndexedLabels labels.Labels
+			if err := jsonparser.ObjectEach(value, func(key []byte, value []byte, dataType jsonparser.ValueType, _ int) error {
+				if dataType != jsonparser.String {
+					return jsonparser.MalformedStringError
+				}
+				nonIndexedLabels = append(nonIndexedLabels, labels.Label{
+					Name:  yoloString(key),
+					Value: yoloString(value),
```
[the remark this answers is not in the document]
sandeepsukhani: I think we should remove `yoloString` here. I had used it earlier because the labels were anyways converted to string immediately. The references to string are now being retained longer so it is not safe to use `yoloString` here.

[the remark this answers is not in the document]
salvacorts: As discussed in Slack, removed.

salvacorts:
I ran the changes in this PR for a couple of hours in a dev cluster and seems to be working fine:

**Write path**
image
image

**Read path**
image

Manually ran some queries for recent data that succeeded

owen-d (approved):
address @sandeepsukhani 's comment about `yolostring` and LGTM

PR #9703, opened 2023-06-13
Convert SeriesVolume response to prometheus-style

trevorwhitney:

This changes the response type of the label volume stats endpoint to return volumes as prometheus-style timeseries metrics. It currently only support instant queries, but is a necessary step to eventually supporting range queries.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/querier/queryrange/codec.go
```diff
@@ -744,23 +745,103 @@ func (Codec) MergeResponse(responses ...queryrangebase.Response) (queryrangebase
...
+			Samples: samples,
+		})
+	}
+
+	if limit < len(result) {
+		result = result[:limit]
+	}
+
+	return queryrangebase.PrometheusData{
+		ResultType: "vector",
```
trevorwhitney, line 840: I actually think there is always potential for this to be a `matrix` due to the way we're splitting index queries by interval, and using the end of the interval as the timestamp. Do we need to be able to support `vector` aka instant queries, or can we just decided to only support `matrix` aka range queries?

pkg/querier/queryrange/codec.go
```diff
@@ -744,23 +745,104 @@ func (Codec) MergeResponse(responses ...queryrangebase.Response) (queryrangebase
...
+			return samples[i].TimestampMs < samples[j].TimestampMs
+		})
+
+		result = append(result, queryrangebase.SampleStream{
+			Labels:  []logproto.LabelAdapter{stream},
+			Samples: samples,
+		})
+	}
+
+	if limit < len(result) {
```
trevorwhitney, line 836: should we do 1 more sort based on stream selector string/hash before applying the limit to make sure results are deterministic?

pkg/storage/stores/index/labelvolume/labelvolume.go
```diff
@@ -45,53 +63,92 @@ func (acc *Accumulator) Volumes() *logproto.LabelVolumeResponse {
...
+			return volumes[i].Volume > volumes[j].Volume
 		}
-
-		return volumes[i].Volume > volumes[j].Volume
+		return volumes[i].Timestamp < volumes[j].Timestamp
 	})
 
-	if limit < len(volumes) {
+  // Apply the limit
+	if len(volumes) > limit {
```
trevorwhitney, line 151: would be curious on your feedback here. based on the sort, this is going to limit to the first N by timestamp, whereas we probably want the last N, but then also the biggest. Maybe we should change the sort to be biggest timestamp first? Or maybe we don't care about the timestamp here?

pkg/querier/queryrange/codec.go
```diff
@@ -744,23 +741,104 @@ func (Codec) MergeResponse(responses ...queryrangebase.Response) (queryrangebase
...
+			Response:   &promResponse,
+			Statistics: stats.Result{},
+		}, nil
 	default:
 		return nil, errors.New("unknown response in merging responses")
 	}
 }
 
+func MergeToPrometheusResponse(responses []*logproto.LabelVolumeResponse, limit int32) queryrangebase.PrometheusData {
+	mergedVolumes := labelvolume.MergeVolumes(responses...)
```
MasslessParticle, line 771: I think we could turn this into a sorted list of `Volume` to simplify the next step

pkg/storage/stores/tsdb/index_client_test.go
```diff
@@ -272,27 +272,55 @@ func TestIndexClient_LabelVolume(t *testing.T) {
 
 	indexClient := NewIndexClient(idx, IndexClientOptions{UseBloomFilters: true})
 
-	t.Run("it returns label volumes from the whole index", func(t *testing.T) {
-		vol, err := indexClient.LabelVolume(context.Background(), "", indexStartYesterday, indexStartToday+1000, 10, nil...)
+	through := indexStartToday + 1000
+	intervals := getIntervalEndpoints(config.ObjectStorageIndexRequiredPeriod, indexStartYesterday.Time(), through.Time(), true)
+
+	t.Run("it returns label volumes from the whole index, separated by ts of interval end", func(t *testing.T) {
```
MasslessParticle, line 278: I would still expect volumes to be aggregated by last timestamp or something here. The `interval` here is an internal detail to how this call is made and shouldn't leak in the response type.

pkg/querier/queryrange/codec.go
```diff
@@ -744,23 +745,72 @@ func (Codec) MergeResponse(responses ...queryrangebase.Response) (queryrangebase
 		}, nil
 	case *VolumeResponse:
 		resp0 := responses[0].(*VolumeResponse)
-		headers := resp0.Headers
+		headers := make([]*definitions.PrometheusResponseHeader, len(resp0.Headers))
+		for i, header := range resp0.Headers {
```
MasslessParticle, line 749: Why copy all the headers? Does something change them later on?

pkg/querier/queryrange/codec.go
```diff
@@ -744,23 +745,72 @@ func (Codec) MergeResponse(responses ...queryrangebase.Response) (queryrangebase
...
+				Value:       float64(volume.Volume),
+			}
+		} else {
+			sample.Value += float64(volume.Volume)
+		}
+	}
+
+	result := make([]queryrangebase.SampleStream, 0, len(samplesByStream))
+	for stream, sample := range samplesByStream {
+		if lbls, err := syntax.ParseLabels(stream); err == nil {
```
MasslessParticle, line 800: In general we probably want to check the condition `err != nil` to avoid surprises. [This page ](https://www.digitalocean.com/community/tutorials/handling-errors-in-go)covers it succinctly: 

`Wherever a function could produce an error, it’s important to use an if statement to check whether one occurred. In this way, idiomatic Go code naturally has its [“happy path”](https://en.wikipedia.org/wiki/Happy_path) logic at the first indent level, and all the “sad path” logic at the second indent level.`

pkg/querier/queryrange/codec.go
```diff
@@ -744,23 +745,72 @@ func (Codec) MergeResponse(responses ...queryrangebase.Response) (queryrangebase
...
+func mapToPrometheusResponse(mergedResponse *logproto.VolumeResponse) queryrangebase.PrometheusData {
+	samplesByStream := map[string]*logproto.LegacySample{}
+
+	// since this is an instant response, we're only interested in a single "bucket", which is
+	// bounded by the latest timestamp we've seen (which should correspond to the end of the query range)
+	tsMs := mergedResponse.Through.UnixNano() / 1e6 //convert ns to ms
+
+	// Aggregate samples into single sample with latest timestamp
+	for _, volume := range mergedResponse.Volumes {
+		if sample, ok := samplesByStream[volume.Name]; !ok {
```
MasslessParticle, line 788: Extreme nit: omit the else statement:

```go
for _, volume := range mergedResponse.Volumes {
  if _, ok := samplesByStream[volume.Name]; !ok {
	  samplesByStream[volume.Name] = &logproto.LegacySample{
		  TimestampMs: tsMs,
	  }
  }
  samplesByStream[volume.Name].Value += float64(volume.Volume)
}
```

MasslessParticle (approved):
A few stylistic things but LGTM

pkg/querier/queryrange/codec.go
trevorwhitney: Not sure I follow. This is doing the same thing as the old code, just turning the slice into a slice of pointers since that's what the `PrometheusReponse` wants. Is the suggestions here just to take `resp0.Headers[0]`?

pkg/querier/queryrange/codec.go
trevorwhitney: The `0` value we finally click with this old Java developer some day...

PR #9704, opened 2023-06-13
Series volume endpoint

MasslessParticle:
This changes the `label_volume` endpoint to the `series_volume` endpoint. The new endpoint still returns volumes but now it does it for the requested streams defined by the selector names passed rather than individual labels. All relevant non-requested labels are aggregated into the returned results

ex: Assume we have the following streams:
```
{cluster="prod", team="A", component="foo"}
{cluster="prod", team="B", component="foo"}
{cluster="dev", team="A", component="foo"}
{cluster="dev", team="B", component="foo"}
```

- requesting `{cluster="prod"}` returns one result for all streams containing `{cluster="prod"}`
- requesting `{cluster=~".+"}` returns two results for the streams containing `{cluster="prod"}` and `{cluster="dev"}`
- requesting `{cluster=~".+", team=".+"}` returns four results for the streams containing:
```
{cluster="prod", team="A"}
{cluster="prod", team="B"}
{cluster="dev", team="A"}
{cluster="dev", team="B"}
```


pkg/storage/stores/tsdb/index_client_test.go
```diff
@@ -272,26 +272,27 @@ func TestIndexClient_SeriesVolume(t *testing.T) {
...
 
-	t.Run("it returns largest label from the index", func(t *testing.T) {
+	t.Run("it returns largest series from the index", func(t *testing.T) {
 		vol, err := indexClient.SeriesVolume(context.Background(), "", indexStartYesterday, indexStartToday+1000, 1, nil...)
 		require.NoError(t, err)
 
 		require.Equal(t, &logproto.VolumeResponse{
 			Volumes: []logproto.Volume{
-				{Name: "foo", Value: "bar", Volume: 300 * 1024},
+				{Name: `{foo="bar"}`, Value: "", Volume: 200 * 1024},
```
trevorwhitney, line 295: what caused the volume change here?

pkg/storage/stores/tsdb/single_file_index.go
```diff
@@ -335,12 +369,21 @@ func (i *TSDBIndex) SeriesVolume(ctx context.Context, _ string, from, through mo
 			}
 
 			if stats.Entries > 0 {
-				for _, l := range []labels.Label(ls) {
-					if _, ok := volumes[l.Name]; !ok {
-						volumes[l.Name] = make(map[string]uint64)
+				seriesLabels = seriesLabels[:0]
+				for _, l := range ls {
+					if _, ok := labelsToMatch[l.Name]; l.Name != TenantLabel && matchAll || ok {
```
trevorwhitney, line 374: I think adding parens here would make this quicker to read. I understand this as `(l.Name != TenantLabel && matchAll) || ok`, is that correct?

trevorwhitney (approved):
Looks great. A few small nits, nothing blocking. I'm going to push a commit after this approval with a few extra tests.

pkg/storage/stores/tsdb/index_client_test.go
MasslessParticle: Because we're returning streams now and matching everything doesn't narrow the labels returned, the result set differentiates between `{foo="bar"}` and `{fizz="buzz", foo="bar"}`. If the matcher was `{foo="bar"}` the result would still be `300*1024`

pkg/storage/stores/tsdb/single_file_index.go
MasslessParticle: That makes sense.

MasslessParticle: In my mind I was actually considering it to be more like `l.Name != TenantLabel && (matchAll || ok)`. Because things are resolved left-to-right, they all need to hold. The order here doesn't actually matter

PR #9710, opened 2023-06-15
Add shuffle sharding to index gateway

chaudum:

This PR introduces shuffle sharding of the index files across index gateways where the shuffle sharding key is the tenant ID.

Instead of having a global replication factor on the index gateways which affects all tenants the same, whether they are small or large tenants, the shuffle sharding allows to specify a shard factor per tenant (on top of the replication factor). This a) reduces the amount of indexes that are downloaded on individual index gateways in case of a high replication factor (which is currently the de-factor standard to deal with high load) and b) allows to scale tenants individually.

The shuffle sharding algorithm is part of dskit's ring implementation and is not part of this commit.
If have set a high RF for the index gateways, it is required to reduce the RF and instead set the shard factor for the tenants.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


docs/sources/configuration/_index.md
```diff
@@ -2612,6 +2612,10 @@ shard_streams:
 
 # Minimum number of label matchers a query should contain.
 [minimum_labels_number: <int>]
+
+# The shard size defines how many index gateways should be used for querying.
+# CLI flag: -index-gateway.shard-size
```
DylanGuedes, line 2617: 1) isn't this per-tenant? if so, maybe the CLI flag shouldn't be listed or you can add a clarification saying that the CLI flag usage will only define the default gateway shard size
2) the flag is different from the YAML config. WDYT of renaming it to be consistent with the config?

docs/sources/configuration/_index.md
```diff
@@ -2612,6 +2612,10 @@ shard_streams:
 
 # Minimum number of label matchers a query should contain.
 [minimum_labels_number: <int>]
+
+# The shard size defines how many index gateways should be used for querying.
```
DylanGuedes, line 2616: this description is a little misleading; if RF=5 and shard-size=1 I'd be using 5 gateways for querying, wouldn't I?

docs/sources/upgrading/_index.md
```diff
@@ -35,6 +35,16 @@ The output is incredibly verbose as it shows the entire internal config struct u
 
 ### Loki
 
+#### Index gateway shuffle sharding
+
+The index gateway now support shuffle sharding of index data when run in "ring"
```
DylanGuedes, line 40:
```suggestion
The index gateway now supports shuffle sharding of index data when running in "ring"
```

pkg/loki/modules.go
```diff
@@ -1158,6 +1158,15 @@ func (t *Loki) addCompactorMiddleware(h http.HandlerFunc) http.Handler {
 func (t *Loki) initIndexGateway() (services.Service, error) {
 	t.Cfg.IndexGateway.Ring.ListenPort = t.Cfg.Server.GRPCListenPort
 
+	var shardingStrategy indexgateway.ShardingStrategy
+	if t.Cfg.IndexGateway.Mode != indexgateway.RingMode || t.indexGatewayRingManager.Mode == indexgateway.ClientMode {
+		shardingStrategy = indexgateway.NewMatchAllStrategy()
+	} else {
+		instanceAddr := t.indexGatewayRingManager.RingLifecycler.GetInstanceAddr()
+		instanceID := t.indexGatewayRingManager.RingLifecycler.GetInstanceID()
+		shardingStrategy = indexgateway.NewShuffleShardingStrategy(t.indexGatewayRingManager.Ring, t.Overrides, instanceAddr, instanceID)
```
DylanGuedes, line 1167: nitpicking: WDYT of just passing `t.indexGatewayRingManager` to `NewShuffleShardingStrategy` and extracting the desired attrs there?  otherwise for every new attr that you might be interested you'd be adding a new parameter to this method

pkg/storage/factory.go
```diff
@@ -401,8 +403,11 @@ func NewIndexClient(periodCfg config.PeriodConfig, tableRange config.TableRange,
 			return nil, err
 		}
 
-		shipper, err := shipper.NewShipper(cfg.BoltDBShipperConfig, objectClient, limits,
-			ownsTenantFn, tableRange, registerer, logger)
+		var filterFn downloads.TenantFilter
```
DylanGuedes, line 406:
```suggestion
		filterFn := ownsTenantFn
```
(if shardingStrategy is nil you'll be passing a nil function while in the current implementation it would pass `ownsTenantFn`)

pkg/storage/stores/indexshipper/downloads/table_manager.go
```diff
@@ -31,10 +31,7 @@ type Limits interface {
 	DefaultLimits() *validation.Limits
 }
 
-// IndexGatewayOwnsTenant is invoked by an IndexGateway instance and answers whether if the given tenant is assigned to this instance or not.
-//
-// It is only relevant by an IndexGateway in the ring mode and if it returns false for a given tenant, that tenant will be ignored by this IndexGateway during query readiness.
```
DylanGuedes, line 36: you dropped the whole comment btw

pkg/validation/limits.go
```diff
@@ -173,6 +173,8 @@ type Limits struct {
 
 	RequiredLabels       []string `yaml:"required_labels,omitempty" json:"required_labels,omitempty" doc:"description=Define a list of required selector labels."`
 	RequiredNumberLabels int      `yaml:"minimum_labels_number,omitempty" json:"minimum_labels_number,omitempty" doc:"description=Minimum number of label matchers a query should contain."`
+
+	IndexGatewayShardSize int `yaml:"gateway_shard_size" json:"gateway_shard_size"`
```
DylanGuedes, line 177: ditto: to be more consistent, maybe you can use "index_gateway_shard_size" or change the cli flag from "index-gateway.shard-size" to "gateway.shard-size".

pkg/storage/stores/shipper/indexgateway/shufflesharding.go
```diff
@@ -0,0 +1,104 @@
...
+		r:            r,
+		limits:       l,
+		instanceAddr: instanceAddr,
+		instanceID:   instanceID,
+	}
+}
+
+// FilterUsers implements ShardingStrategy.
+func (s *ShuffleShardingStrategy) FilterUsers(tenantIDs []string) ([]string, error) {
+	// As a protection, ensure the store-gateway instance is healthy in the ring. It could also be missing
```
DylanGuedes, line 53:
```suggestion
	// As a protection, ensure the index-gateway instance is healthy in the ring. It could also be missing
```

pkg/storage/stores/shipper/indexgateway/shufflesharding.go
```diff
@@ -0,0 +1,104 @@
...
+	// in the ring if it was failing to heartbeat the ring and it got remove from another healthy store-gateway
```
DylanGuedes, line 54:
```suggestion
	// in the ring if it was failing to heartbeat the ring and it got remove from another healthy index-gateway
```

pkg/storage/stores/shipper/indexgateway/shufflesharding.go
```diff
@@ -0,0 +1,104 @@
...
+	// instance, because of the auto-forget feature.
+	if set, err := s.r.GetAllHealthy(IndexesSync); err != nil {
+		return nil, err
```
DylanGuedes, line 57: if the gateway is unhealthy, isn't better to just not filter anything (but to log the error)?

pkg/storage/stores/shipper/indexgateway/shufflesharding.go
```diff
@@ -0,0 +1,104 @@
...
+	} else if !set.Includes(s.instanceAddr) {
+		return nil, errGatewayUnhealthy
+	}
+
+	var filteredIDs []string
+
+	for _, tenantID := range tenantIDs {
+		subRing := GetShuffleShardingSubring(s.r, tenantID, s.limits)
+
+		// Include the user only if it belongs to this store-gateway shard.
```
DylanGuedes, line 67:
```suggestion
		// Include the user only if it belongs to this index-gateway shard.
```

DylanGuedes (commented):
this looks good. I added a few suggestions but they're mostly nitpicking.

chaudum:
I think I got it wrong how the replication factor ties into shuffle sharding. Calling `subring.GetReplicationSetForOperation(op)` will give you all matching instances for the operation _op_. 

`ring.ShuffleShard("tenant", shardSize)` will return a sub-ring with desired _shardSize_ instances. It does not have the notion of replicas at this point. The sub-ring can be used to fetch data based on a hash key using `subring.Get()`. That data then has replicas spread across the instances of _subring_.

pkg/loki/modules.go
chaudum: Hm, not sure if a ShardingStrategy should necessarily know about a RingManager. Maybe a new constructor `GetShardingStrategyForRing()`

pkg/storage/stores/indexshipper/downloads/table_manager.go
chaudum: Yeah, I should rather update

pkg/storage/stores/shipper/indexgateway/shufflesharding.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+type Limits interface {
+	IndexGatewayShardSize(tenantID string) int
+}
+
+type ShardingStrategy interface {
+	// FilterUsers whose indexes should be loaded by the index gateway.
+	// Returns the list of user IDs that should be synced by the index gateway.
+	FilterUsers(tenantID []string) ([]string, error)
```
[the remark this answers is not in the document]
chaudum: I rather don't want `indexgateway` depend on `pkg/storage/stores/indexshipper/downloads` that's why I have the same signature in two places

docs/sources/configuration/_index.md
DylanGuedes: ohh that makes sense.

pkg/storage/stores/shipper/indexgateway/shufflesharding.go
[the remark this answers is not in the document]
DylanGuedes: that makes sense, thank you.

pkg/storage/factory.go
chaudum: The current implementation also passes `nil` for the `ownsTenantFn` type

docs/sources/upgrading/_index.md
```diff
@@ -35,7 +35,26 @@ The output is incredibly verbose as it shows the entire internal config struct u
 
 ### Loki
 
+#### Index gateway shuffle sharding
+
+The index gateway now supports shuffle sharding of index data when running in
+"ring" mode. The index data is sharded by tenant where each tenant gets
+assigned a sub-set of all available instances of the gateway ring.
+
+In case you configured a high replication factor to accommodate for load, since
```
JStickler, line 44:
```suggestion
If you configured a high replication factor to accommodate for load, since
```

docs/sources/upgrading/_index.md
```diff
@@ -35,7 +35,26 @@ The output is incredibly verbose as it shows the entire internal config struct u
...
+in the past this was the only option to give a tenant more instances for
+querying, you should consider reducing the replication factor to meaningful
```
JStickler, line 46:
```suggestion
querying, you should consider reducing the replication factor to a meaningful
```

docs/sources/upgrading/_index.md
```diff
@@ -35,7 +35,26 @@ The output is incredibly verbose as it shows the entire internal config struct u
...
+value for replication (e.g. from 12 to 3) and instead set the shard factor for
```
JStickler, line 47:
```suggestion
value for replication (for example, from 12 to 3) and instead set the shard factor for
```

docs/sources/upgrading/_index.md
```diff
@@ -35,7 +35,26 @@ The output is incredibly verbose as it shows the entire internal config struct u
...
+individual tenants as required.
+
+If the global shard factor (no per-tenant) is 0 (default value), the global
+shard factor of is set to replication factor. It can still be overwritten per
```
JStickler, line 51: "shard factor of is set to replication "  Same comment as above, either there is a word missing here or delete the word "of".

JStickler (commented):
[docs team]  Couple of suggested revisions, and a possible missing word?

pkg/storage/stores/indexshipper/downloads/table_manager.go
```diff
@@ -349,6 +362,7 @@ func (tm *tableManager) ensureQueryReadiness(ctx context.Context) error {
 		level.Info(tm.logger).Log(
 			"msg", "index pre-download for query readiness completed",
 			"users_len", len(usersToBeQueryReadyFor),
+			"users", strings.Join(usersToBeQueryReadyFor, ","),
```
DylanGuedes, line 365: isn't the final log line going to be too long? you'd be logging it for every queryReadiness call (the log level is info)

pkg/storage/stores/shipper/indexgateway/shufflesharding.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/pkg/errors"
+)
+
+var (
+	// IndexesSync is the operation used to check the authoritative owners of an index
+	// (replicas included).
+	IndexesSync = ring.NewOp([]ring.InstanceState{ring.JOINING, ring.ACTIVE, ring.LEAVING}, nil)
+
+	// IndexesRead is the operation run by the querier to query indexes via the index-gateway.
+	IndexesRead = ring.NewOp([]ring.InstanceState{ring.ACTIVE}, func(s ring.InstanceState) bool {
```
sandeepsukhani, line 14: Please correct me if I am wrong but I think `shouldExtendReplicaSet` won't have any effect since `GetReplicationSetForOperation` is being called on a subring which would only have limited index gateways which owns the tenant.

sandeepsukhani (approved):
I left some minor non-blocking comments but other than that, it LGTM!

pkg/storage/stores/shipper/indexgateway/shufflesharding.go
chaudum: You are right, it does not have any effect. I better remove it.

JStickler (approved):
[docs team] LGTM

PR #9730, opened 2023-06-19
POC: Asynchronous queries

periklis:
The following PR provides a proof-of-concept feature implementation for asynchronous queries based on nats.io. Its main purpose is to explore how to integrate a built-in pub/sub infrastructure in Loki to support asynchronous query requests. It is based on a previous experiment with nats.io (See #9055). The current approach provides a simple two NATS streams infrastructure:
1. `query.{tenantID}`: A per-tenant subject to record the full instant query from the request queue.
2. `processed`: An inbox for query responses once processed by the querier worker.

Messages on each NATS stream share a simple UUID as a message id to relate the user's query request to a response. This id can be returned to the user as a potential task ID to return back and ask for the response once processed by the querier workers.

Fixes #<issue number>

This POC is to be treated as a companion to the corresponding LID: 

- https://github.com/grafana/loki/pull/9737

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


cyriltovena:
How does it works when the query is splitted/sharded ?

periklis:
> How does it works when the query is splitted/sharded ?

This is an excellent point that I should have mentioned in the description. I deliberately left this out in this proof-of-concept for the sake of simplicity. I tend to think that instant queries are a simplification of split/sharded queries whereas we just run for a user request a series of smaller queries. From this aspect we have a couple of options on how to address split/sharded asynchronous queries. Some food for thought and further discussion before we put this on the LID:

1. We can keep using the instant asynchronous queries JetStreams for incoming and processed queries. We could add another header to express which part of the response corresponds to a query split/shard (e.g. extra headers expressing: ` Nats-Msg-Chunk: 1/n`). An additional asynchronous routine (e.g. in the query-frontend/request-queue) can consume the response parts as they arrive merge and put them on the `processed` JetStream back.
2. In contrast we can keep the original unsplit/unsharded query request on the incoming JetStream. We add an additional set of headers expressing how many splits/shards it represents (e.g. something like a counter). Then each time a querier processes this request it decrements the counter and puts the message back on the incoming queue until the counter reaches zero. In that case the processed merged query response is compiled and put on the processed stream.

WDYT?

periklis:
@cyriltovena Please consider this diagram as an initial idea to spin the query range topic further, feedback welcome: https://github.com/grafana/loki/blob/c8a42fb5be88d01110f674e1f4c1ccdb21604b51/docs/sources/community/lids/_DRAFT_AsynchronousQueries.md#overview-on-using-asynchronous-streams-for-range-queries

PR #9737, opened 2023-06-19
Add LID for asynchronous queries

periklis:
As per #9730 this is the PR providing a LID for asynchronous queries and further discussion

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


docs/sources/community/lids/_DRAFT_AsynchronousQueries.md
```diff
@@ -0,0 +1,97 @@
...
+
+* Users need to tune their queries manually (e.g. distinct label selectors, batching time frames).
+* Users need to request per-tenant and/or per-request limits by their respecting Loki administrators to overcome limits for certain query demand over time.
+* Client applications and networking infrastructure needs to be tuned for Loki response latency scenarios (e.g. Grafana datasource dataproxy, Server Write Timeouts on Gateways)
+
+### Proposal 1: Built-in asynchronous streaming component for asynchronous queries
+
+The main approach of a built-in asynchronous streaming component as devised in the [proof-of-concept PR](https://github.com/grafana/loki/pull/9730) is based on [nats.io's jetstream technology](https://nats.io/) and consists of:
+
+- A new Loki component called the `nats-server` that is hosting the persistence layer for any NATS JetStreams.
```
cyriltovena, line 66: I think we'll need some sort of abstraction. If we run the `all` target I expect all of this to happen in-memory or on temp file.

docs/sources/community/lids/_DRAFT_AsynchronousQueries.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Proposal 0: Do nothing
+
+The current state represents the do-nothing scenario very well, as per trade-offs currently to consider:
+
+* Users need to tune their queries manually (e.g. distinct label selectors, batching time frames).
+* Users need to request per-tenant and/or per-request limits by their respecting Loki administrators to overcome limits for certain query demand over time.
+* Client applications and networking infrastructure needs to be tuned for Loki response latency scenarios (e.g. Grafana datasource dataproxy, Server Write Timeouts on Gateways)
+
+### Proposal 1: Built-in asynchronous streaming component for asynchronous queries
```
owen-d, line 62: Thanks for putting this together. A few notes:

* I think we'll want to provide storage (likely object storage) + discovery for query requests and status as an API.
* I'm not sure what the added benefit `nats` brings here over object storage + a component to move async queries towards completion (basically a `ruler` that reads requests from object storage and writes responses back to object storage, then combines partial responses into total ones, such as individual steps in a range query).
* We could get fancier and try to optimize running queries when there is low system load, update status based on partial completion (75% -> 80%, etc)

docs/sources/community/lids/_DRAFT_AsynchronousQueries.md
```diff
@@ -0,0 +1,105 @@
...
+- Allows full-control on stream characteristics (i.e. retention, message replication, ack policy)
+- Enables handling queries as simple pub/sub messages using built-in JetStream primitives (i.e. pub/sub provider as a native Loki subservice).
+
+The tradeoffs using this approach are:
+
+- A direct dependency to the NATS.io client and server code to be maintained.
+- A whole new set of settings to control and maintain JetStream streams on the server-side.
+- Object storage support for persisting streams currently experimental.
+
+### Proposal 2: Client support for external asynchronous streaming systems
```
owen-d, line 81: I'm less attracted to this option because it seems more complex in terms of external dependencies and thus less likely to see use.

owen-d (commented):
Really cool idea, thanks for the writeup. Left a few notes, let me know what you think.

docs/sources/community/lids/_DRAFT_AsynchronousQueries.md
periklis: > Thanks for putting this together. A few notes:
> 
>     * I think we'll want to provide storage (likely object storage) + discovery for query requests and status as an API.
> 

This sounds like another worthy proposal section in this LID. It sound like a self-baked solution that persists query request and could possible re-use the remote-ruler evaluation like pipeline to execute the queries.

>     * I'm not sure what the added benefit `nats` brings here over object storage + a component to move async queries towards completion (basically a `ruler` that reads requests from object storage and writes responses back to object storage, then combines partial responses into total ones, such as individual steps in a range query).
> 

Although `nats` can store the streams in object storage (currently experimental support but for sure coming), I believe asynchronous request can benefit from the message handling mechanism implemented, i.e.:
- It supports replication and retention out of the box regardless storage on disk or object storage.
- It supports message delivery acks/nacks/etc. so that in case of failure a re-process is an easy piece.

There are a couple of examples and not meant to be comprehensive. I think it is worth having a conversation to evaluate if we need such message queue handling. These usually are there for failure scenarios.

>     * We could get fancier and try to optimize running queries when there is low system load, update status based on partial completion (75% -> 80%, etc)

Sure a status API should be mandatory in the final implementation. The users want to experience that progress is made over time.

docs/sources/community/lids/_DRAFT_AsynchronousQueries.md
periklis: I feel similar about having just client-like implementations. However for the sake of completion and respecting the legacy of Loki (probably mostly based on how cortex evolved), I didn't want to leave this proposal out.

docs/sources/community/lids/_DRAFT_AsynchronousQueries.md
owen-d: Fair points. I'm personally leaning away from nats as my gut says the benefits aren't worth the additional complexity here, but open to the idea. Overall I like the idea of natively supporting async queries. If there aren't any detractors, I think the next step would be a more detailed design doc. Nice work 👍

PR #9754, opened 2023-06-21
store: fix incorrect usage of period config in table range

ashwanthgoli:
We are incorrectly taking reference to a loop variable in store init, I overlooked this part as the reference was being taken in the call to `GetIndexTableNumberRange`

As a result of this, `TableRange` of each period (which is used by index-shipper) points to the latest `periodConfig`. This in itself is not a problem if the index prefix doesn't change as we use the `periodConfig` only to generate the `tableName`. But if the index prefix changes across periods, we end up losing access to the indexes using the old prefix.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


dannykopping (changes requested):
Please add tests that would expose the original bug, and then provide it's fixed with this change

ashwanthgoli:
`TestMicroServicesIngestQueryWithSchemaChange` fails on main without the fix.
It doesn't return lines `lineA, lineB` from the previous period(which uses a different index prefix) in the query results.

```
loki on main [$⇡] via 🐹 v1.20.3 on ☁️  ashwanth.goli@grafana.com
❯ go test -v -count=1 ./integration -run ^TestMicroServicesIngestQueryWithSchemaChange
=== RUN   TestMicroServicesIngestQueryWithSchemaChange
....
=== RUN   TestMicroServicesIngestQueryWithSchemaChange/query-both-periods-with-defaul-lookback
time="2023-06-22T13:22:24+05:30" level=info msg="=== Handler.Stop()'d ==="
time="2023-06-22T13:22:25+05:30" level=info msg="server listening on addresses" grpc="[::]:62434" http="[::]:62433"
    loki_micro_services_test.go:295:
                Error Trace:    /Users/grafana/Workspace/loki/integration/loki_micro_services_test.go:295
                Error:          elements differ

                                extra elements in list A:
                                ([]interface {}) (len=2) {
                                 (string) (len=5) "lineA",
                                 (string) (len=5) "lineB"
                                }
....
=== RUN   TestMicroServicesIngestQueryWithSchemaChange/query_both_periods_to_verify_logs_being_served_from_storage
    loki_micro_services_test.go:320:
                Error Trace:    /Users/grafana/Workspace/loki/integration/loki_micro_services_test.go:320
                Error:          elements differ

                                extra elements in list A:
                                ([]interface {}) (len=2) {
                                 (string) (len=5) "lineA",
                                 (string) (len=5) "lineB"
                                }
```

dannykopping (approved):
Very nice clear test! 🏅 
LGTM

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
 * [9495](https://github.com/grafana/loki/pull/9495) **thampiotr**: Promtail: Fix potential goroutine leak in file tailer.
 * [9650](https://github.com/grafana/loki/pull/9650) **ashwanthgoli**: Config: ensure storage config defaults apply to named stores.
 * [9629](https://github.com/grafana/loki/pull/9629) **periklis**: Fix duplicate label values from ingester streams.
+* [9754](https://github.com/grafana/loki/pull/9754) **ashwanthgoli**: store: fix incorrect usage of period config in table range
```
chaudum, line 56: nit: Please phrase the changelog in a way that the user can identify it as a fix to their problem.

```
Store: Fix an issue where different index prefixes in the schemas period configs leads to missing logs from the previous schema periods when querying logs that have been ingested during a previous schema period.

chaudum: It could be condensed a little bit, though.

ashwanthgoli: how does this sound? ```Fixes an issue where indexes using a prefix different from the one in current period become unqueriable```

ashwanthgoli:
updated with a unit-test

chaudum (approved):
lgtm

PR #9774, opened 2023-06-22
Remove unused Value field

MasslessParticle:
We didn't end up needing the `Value` field because we can express everything we need to as selectors


trevorwhitney (approved):
This PR has no *value*
:wink:

JStickler:
Was this ever documented?  (Do we need to remove it from the docs anywhere?)

MasslessParticle:
@JStickler Nope :slightly_smiling_face: It didn't last long enough to warrant documentation

PR #9776, opened 2023-06-22
Log Volume: Do the prometheus-format conversion in dedicated middleware

MasslessParticle:
This PR moves the Prometheus-format conversion to it's own middleware. This keeps it all in one place and consolidates logic that was in many places. It also sets us up to make range queries as a next step.


pkg/logproto/logproto.proto
```diff
@@ -397,14 +397,6 @@ message VolumeRequest {
...
   repeated Volume volumes = 1 [(gogoproto.nullable) = false];
   int32 limit = 2;
-  int64 from = 3 [
-    (gogoproto.customtype) = "github.com/prometheus/common/model.Time",
-    (gogoproto.nullable) = false
-  ];
-  int64 through = 4 [
-    (gogoproto.customtype) = "github.com/prometheus/common/model.Time",
-    (gogoproto.nullable) = false
-  ];
```
trevorwhitney, line 407: This cleans up so much, looking much better now!

trevorwhitney (approved):
Looks great!

PR #9778, opened 2023-06-23
operator: add startupProbe to pods of zone aware components

JoaoBraveCoding:

Problem: when a user configures zones in spec.replication we want to
make sure that when loki pods are started that they have access to the
zone that they are scheduled in. The zone_labeling_controller is
responsible for annotating the pods of zone aware componets, once these
are scheduled an annotation containing the zone where the pod was
scheduled is added. Furthermore, the controller will also configure the container
env vars to have this annotation available as an ENV variable.
Loki for loki to operate propperly this ENV var must be set when the
container starts, otherwise loki will not be configured propperly.

Issue: https://issues.redhat.com/browse/LOG-3835

Solution: in this PR we have added a startupProbe to each pod of zone
aware components. This startup probe will validate if the ENV var is set
or not, if the ENV var is not set the container will be restarted.
So if a container starts at T and the controller still hasn't added the
annotation then nor loki nor the startupProbe will have access to the
annotation which will eventually lead to the restart of the container at
T + 2. If at T + 1 the controller adds the annotation then at T + 2 when
the container is restarted both loki and the startupProbe will see the
annotation

Validation: To validate this solution I tested the PR on a AWS cluster
and validated the if the ENV is set for the loki processes with:
cat /proc/1/environ | tr '\0' '\n'

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


JoaoBraveCoding:
initContainer alternative https://github.com/grafana/loki/pull/9795

PR #9780, opened 2023-06-23
Migrate individual chunks as opposed to all chunks.  

paul1r:
This way, if an error occurs, we get as much as we can.

In some situations, not all blob data may be accessible.  For instance, in the case where we have started migration between object storage buckets, some data may be in the old bucket.  If we wanted to migrate data from the new bucket, those chunks will error with a 404, as they are in the old bucket.

This PR is so that we continue on and get all of the chunks that we can, rather than stopping with whatever error may arise.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/cmd/migrate/main.go
+++ b/cmd/migrate/main.go
@@ -339,27 +339,38 @@ func (m *chunkMover) moveChunks(ctx context.Context, threadID int, syncRangeCh <
 						keys = append(keys, key)
 						chks = append(chks, chk)
 					}
-					for retry := 10; retry >= 0; retry-- {
-						chks, err = f.FetchChunks(m.ctx, chks, keys)
-						if err != nil {
-							if retry == 0 {
-								log.Println(threadID, "Final error retrieving chunks, giving up:", err)
-								errCh <- err
-								return
+					finalChks := make([]chunk.Chunk, 0, len(chunks))
+					for i := range chks {
+						onechunk := []chunk.Chunk{chunks[i]}
+						onekey := []string{keys[i]}
+						var retry int
+						for retry = 10; retry >= 0; retry-- {
+							onechunk, err = f.FetchChunks(m.ctx, onechunk, onekey)
+							if err != nil {
+								if retry == 0 {
+									log.Println(threadID, "Final error retrieving chunks, giving up:", err)
+								}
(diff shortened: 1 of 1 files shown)


cmd/migrate/main.go
```diff
@@ -339,27 +339,38 @@ func (m *chunkMover) moveChunks(ctx context.Context, threadID int, syncRangeCh <
...
-								log.Println(threadID, "Final error retrieving chunks, giving up:", err)
-								errCh <- err
-								return
+					finalChks := make([]chunk.Chunk, 0, len(chunks))
+					for i := range chks {
+						onechunk := []chunk.Chunk{chunks[i]}
+						onekey := []string{keys[i]}
+						var retry int
+						for retry = 10; retry >= 0; retry-- {
+							onechunk, err = f.FetchChunks(m.ctx, onechunk, onekey)
```
kavirajk, line 348: Q: I have a feeling copying once chunk at a time may be super slow for the migration process? (additionally, now even single chunk failure would retry upto 10 times)

wondering does `f.FetchChunks` returns partial chunks and an non-nil error in that case you are describing?. 
[Looks like it though.](https://github.com/kavirajk/loki/blob/kavirajk%2Fx-rule-uid-grafana-managed-alerts/pkg/storage/chunk/client/util/parallel_chunk_fetch.go#L75-L76).

in that case, can we still use `batch` and use the partial chunks returned from the store? instead of ignoring those.

paul1r: Regarding speed, there's a disclaimer somewhere in here saying the migrate tool is not optimized, so yes, definitely a known issue, and this isn't helping.

However, regarding the one by one, I did digging into this, and this line winds up throwing out the partial fetch:
https://github.com/grafana/loki/blob/main/pkg/storage/chunk/fetcher/fetcher.go#L225

So, while I'd like to do what you are proposing, it requires more rework within Loki itself.

kavirajk: Gotcha! thanks for the more info. Sounds good to me :+1:

kavirajk (approved):
LGTM :+1:

PR #9795, opened 2023-06-26
operator: add initContainer to pods of zone aware components

JoaoBraveCoding:
Issue: https://issues.redhat.com/browse/LOG-3835

Problem: when a user configures zones in spec.replication we want to
make sure that when loki pods are started that they have access to the
zone that they are scheduled in. The zone_labeling_controller is
responsible for annotating the pods of zone-aware components, once these
are scheduled an annotation containing the zone where the pod was
scheduled is added. Furthermore, the controller will also configure the container
env vars to have this annotation available as an ENV variable.
Loki for loki to operate properly this ENV var must be set when the
the container starts, otherwise, loki will not be configured properly.

Solution: in this PR we have added an initContainer to each pod of zone
aware components. This initContainer will validate if the annotation was
set, by mounting the annotation as a volume and validating that the file
inside that volume is not empty
So if a container starts and the controller still hasn't added the
annotation then initContainer will enter a loop waiting for the
annotation to be added. Once the annotation is added since we use a volume
mount to access tha annotation the file inside the volume will be
updated, hence breaking the loop and allowing the loki components to
start.

Validation: To validate this solution I tested the PR on a AWS cluster
and validated the if the ENV is set for the loki processes with:

`oc get pods -o json | jq --arg q "'\0' '\n'" -r ' .items[] | select( .metadata.name | test("distributor|index|ingester|querier|query")) | "echo \(.metadata.name) $(oc exec \(.metadata.name) -c \(.spec.containers[0].name) -- cat /proc/1/environ | tr \($q) | grep INSTANCE_AVAILIBILITY_ZONE)"' | bash`

image

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/operator/internal/manifests/node_placement.go
+++ b/operator/internal/manifests/node_placement.go
@@ -11,11 +11,11 @@ import (
 )
 
 const (
-	availabilityZoneEnvVarName = "INSTANCE_AVAILABILITY_ZONE"
-	availabilityZoneFieldPath  = "metadata.annotations['" + lokiv1.AnnotationAvailabilityZone + "']"
-	availabilityZoneInitVolumeName = "az-annotation"
+	availabilityZoneEnvVarName          = "INSTANCE_AVAILABILITY_ZONE"
+	availabilityZoneFieldPath           = "metadata.annotations['" + lokiv1.AnnotationAvailabilityZone + "']"
+	availabilityZoneInitVolumeName      = "az-annotation"
 	availabilityZoneInitVolumeMountPath = "/etc/az-annotation"
-	availabilityZoneInitVolumeFileName = "az"
+	availabilityZoneInitVolumeFileName  = "az"
 )
 
 var availabilityZoneEnvVar = corev1.EnvVar{
--- a/operator/internal/manifests/node_placement.go
+++ b/operator/internal/manifests/node_placement.go
@@ -44,9 +44,7 @@ func configureReplication(podTemplate *corev1.PodTemplateSpec, replication *loki
 			Annotations: map[string]string{},
 		},
 		Spec: corev1.PodSpec{
-			InitContainers: []corev1.Container{initContainerAZAnnotationCheck(podTemplate.Spec.Containers[0].Image)},
-			Containers:     make([]corev1.Container, len(podTemplate.Spec.Containers)),
-			Volumes:        []corev1.Volume{azAnnotationVolume()},
+			Containers: make([]corev1.Container, len(podTemplate.Spec.Containers)),
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


JoaoBraveCoding:
startupProve alternative https://github.com/grafana/loki/pull/9778

operator/internal/manifests/var.go
```diff
@@ -664,3 +664,39 @@ func getInstanceAvailabilityZoneEnvVar() corev1.EnvVar {
...
+func zoneAnnotationVolumeMount() corev1.Volume {
+	return corev1.Volume{
+		Name: "zone-annotation",
+		VolumeSource: corev1.VolumeSource{
+			DownwardAPI: &corev1.DownwardAPIVolumeSource{
+				Items: []corev1.DownwardAPIVolumeFile{
+					{
+						Path: "annotation",
+						FieldRef: &corev1.ObjectFieldSelector{
+							FieldPath: "metadata.annotations['loki_instance_availability_zone']",
```
periklis, line 695: Let's a keep a note to move the value of this field to a constant once the parent PR is merged if not done there already.

periklis (commented):
LGTM

operator/internal/manifests/node_placement.go
```diff
@@ -12,6 +13,13 @@ import (
 const (
 	availabilityZoneEnvVarName = "INSTANCE_AVAILABILITY_ZONE"
 	availabilityZoneFieldPath  = "metadata.annotations['" + lokiv1.AnnotationAvailabilityZone + "']"
+	// availabilityZoneVolumeName is the name of the volume that will contain the
+	// availability zone annotation we get from DownwardAPI
+	availabilityZoneVolumeName = "az-annotation"
+	// availabilityZoneVolumeMountPath path where the volume will be mounted on the init container
+	availabilityZoneVolumeMountPath = "/etc/az-annotation"
+	// availabilityZoneVolumeFileName name of the file containing the availability zone annotation
+	availabilityZoneVolumeFileName = "az"
```
xperimental, line 22: If we only need these in the init container then adding an `Init` to the constant name might be an idea?

```suggestion
	availabilityZoneInitVolumeName      = "az-annotation"
	availabilityZoneInitVolumeMountPath = "/etc/az-annotation"
	availabilityZoneInitVolumeFileName  = "az"
```

operator/internal/manifests/node_placement.go
```diff
@@ -78,3 +95,40 @@ func configureReplication(podTemplate *corev1.PodTemplateSpec, replication *loki
...
+
+func initContainerAZAnnotationCheck(image string) corev1.Container {
+	azPath := fmt.Sprintf("%s/%s", availabilityZoneVolumeMountPath, availabilityZoneVolumeFileName)
+	return corev1.Container{
+		Name:  "az-annotation-check",
+		Image: image,
+		Command: []string{
+			"sh",
+			"-c",
+			fmt.Sprintf("while ! [ -s %s ]; do echo Waiting for availability zone annotation to be set; sleep 2; done; echo availability zone annotation is set; cat %s", azPath, azPath),
```
xperimental, line 107: As far as I can see, the generated file for the annotation will (correctly) not contain a newline at the end.

This might look weird in output and mess with logging systems if the last thing the init container emits is a line with no EOL. I suggest we add another newline after outputting the AZ.

xperimental (commented):
Two comments inline. The changelog also needs an update, I think.

periklis (approved):

PR #9797, opened 2023-06-26
Add per-tenant request counter metric to index gateway server

chaudum:
This commit add a counter metric with labels `operation`, `tenant`, `status` for gRPC requests that are served by the index gateway.

**What for?**

The per-tenant RPS on the index gateway is used to derive the per-tenant shard factor.

**Why tracking on the server?**

Unlike tracking index gateway RPS on the client side, tracking on the server side does not yield that many series, even in multi-tenant installations with a lot of tenants, because the amout of index gateway instances is relatively small compared to the amount of queriers and frontends.

The previous approach of tracking requests on the client https://github.com/grafana/loki/pull/9781 has been abandoned.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -66,6 +66,7 @@ import (
 	boltdb_shipper_compactor "github.com/grafana/loki/pkg/storage/stores/shipper/index/compactor"
 	"github.com/grafana/loki/pkg/storage/stores/shipper/indexgateway"
 	"github.com/grafana/loki/pkg/storage/stores/tsdb"
+	"github.com/grafana/loki/pkg/util"
 	"github.com/grafana/loki/pkg/util/httpreq"
 	"github.com/grafana/loki/pkg/util/limiter"
 	util_log "github.com/grafana/loki/pkg/util/log"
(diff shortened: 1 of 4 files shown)


DylanGuedes (approved):

pkg/loki/modules.go
```diff
@@ -1242,6 +1244,15 @@ func (t *Loki) initIndexGatewayRing() (_ services.Service, err error) {
 	return t.indexGatewayRingManager, nil
 }
 
+func (t *Loki) initIndexGatewayInterceptors() (services.Service, error) {
+	// Only expose per-tenant metric if index gateway runs as standalone service
+	if util.StringsContain(t.Cfg.Target, IndexGateway) {
```
sandeepsukhani, line 1249: We actually use this to check if the module was set in the targets. Although it does the same thing under the hood, it would be good to keep things consistent.
```suggestion
	if t.Cfg.isModuleEnabled(IndexGateway) {
```

sandeepsukhani (approved):
left a minor nit but other than that it LGTM

pkg/loki/modules.go
chaudum: I did not check the function implementation and thought it that it would also return true if the module is a dependency of the target. I think the function name is slightly misleading.

Of course, I will apply your suggestion.

sandeepsukhani: yeah, the naming is confusing. This one checks if the user explicitly enabled the module in the targets flag. There is also `isModuleActive` which checks if the module would be initialized.

PR #9798, opened 2023-06-26
index: allow refreshing index file names cache for specific tables

sandeepsukhani:
We have a cache for index files available in the object storage, which helps reduce the number of list calls we make to the object storage. Currently, we only support refreshing the whole index cache, which includes all the files for all tables and users. However, in a large cell with many tables and tenants, it is less likely that all the tables are being referenced. Also, most iterative operations would only need a fresh listing of just a few tables at a time so it is wasteful to always refresh the whole index cache.

This PR adds support for refreshing the index cache for only specific tables. I am doing it as part of some compactor improvements to run compaction and retention in separate goroutines to avoid blocking compaction when some expensive delete requests are being processed. To make those changes, I needed a way to cost-effectively refresh the index files in the table being picked up for processing.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/indexshipper/storage/cached_client.go
+++ b/pkg/storage/stores/indexshipper/storage/cached_client.go
@@ -20,29 +20,46 @@ const (
 )
 
 type table struct {
+	name          string
+	mtx           sync.RWMutex
 	commonObjects []client.StorageObject
 	userIDs       []client.StorageCommonPrefix
 	userObjects   map[string][]client.StorageObject
-}
 
-type cachedObjectClient struct {
-	client.ObjectClient
-
-	tables       map[string]*table
-	tableNames   []client.StorageCommonPrefix
-	tablesMtx    sync.RWMutex
-	cacheBuiltAt time.Time
-
+	cacheBuiltAt   time.Time
 	buildCacheChan chan struct{}
 	buildCacheWg   sync.WaitGroup
 	err            error
 }
 
+func newTable(tableName string) *table {
+	return &table{
+		name:           tableName,
+		buildCacheChan: make(chan struct{}, 1),
+		userIDs:        []client.StorageCommonPrefix{},
+		userObjects:    map[string][]client.StorageObject{},
+		commonObjects:  []client.StorageObject{},
+	}
+}
+
+type cachedObjectClient struct {
+	client.ObjectClient
+
+	tables                 map[string]*table
+	tableNames             []client.StorageCommonPrefix
+	tablesMtx              sync.RWMutex
+	tableNamesCacheBuiltAt time.Time
+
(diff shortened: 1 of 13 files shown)


pkg/storage/stores/indexshipper/storage/cached_client.go
```diff
@@ -20,137 +20,279 @@ const (
...
+		})
+	}
+
+	// wait for cache build operation to finish, if running
+	tbl.buildCacheWg.Wait()
+
+	if tbl.err != nil {
+		return nil, c.err
+	}
+
```
ashwanthgoli, line 212: objects within the table are protected using the mtx from what I understand, it's missing here and in `listTable`
```
t.mtx.Lock()
defer t.mtx.Unlock()
```

sandeepsukhani: I deliberately didn't keep it locked to avoid contention. The table objects can change when refreshing the table names cache, however, that code does not clear/modify any data on the table objects. However, I think it would be better to have a read lock to avoid any possible bugs with future changes. I will see if I can change it. Thanks for the comment!

pkg/storage/stores/indexshipper/storage/cached_client.go
```diff
@@ -20,137 +20,279 @@ const (
...
 	}
 }
 
-func (c *cachedObjectClient) RefreshIndexListCache(ctx context.Context) {
-	c.buildCacheOnce(ctx, true)
-	c.buildCacheWg.Wait()
+func (c *cachedObjectClient) RefreshIndexTableNamesCache(ctx context.Context) {
+	buildCacheOnce(&c.buildTableNamesCacheWg, c.buildTableNamesCacheChan, func() {
+		c.err = nil
+		c.err = c.buildTableNamesCache(ctx, true)
```
ashwanthgoli, line 87: nit: should we log the error here

ashwanthgoli (approved):
lgtm

pkg/storage/stores/indexshipper/storage/cached_client.go
ashwanthgoli: just to be on the same page :) I am referring to the mtx in `table` which we are [locking](https://github.com/grafana/loki/pull/9798/files#diff-3a3ba0cf92979e89dab84da8c9e882ff5c38d7b467c5ed1cbeefd434fcb3c98cR290) when updating the common and user objects in `buildCache`

not taking the lock on the table names cache in this path might be alright I think

sandeepsukhani: ohh, sorry, my bad. I missed it. Good catch!

PR #9808, opened 2023-06-27
Include entry metadata bytes in metrics tracking ingestion stats

salvacorts:
In https://github.com/grafana/loki/pull/9694 we added support for adding metadata labels to each entry in a push payload. In this PR we update the following metrics to take into account the bytes needed to store those labels:

-  `loki_distributor_bytes_received_total`
- `distributor_bytes_received`

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


vlad-diachenko (approved):
looks awesome )

vlad-diachenko (commented):
could you please also check if we have the tests and add the test-case that covers the new logic . 
also, could you please update debug log that we have at the bottom of the method to print the size of non-indexed metadata separately.... it might be super useful during the investigation

salvacorts:
> could you please update debug log that we have at the bottom of the method to print the size of non-indexed metadata separately.... it might be super useful during the investigation

@vlad-diachenko Done

> could you please also check if we have the tests and add the test-case that covers the new logic .

As far as I can see, we are not testing the `bytesIngested` and `bytesReceivedStats` metrics anyway. Unfortunately, I don't see a way to test out those metrics. Happy to learn about how to do so! Do we have any examples? I couldn't find any.

vlad-diachenko:
@salvacorts 
> As far as I can see, we are not testing the `bytesIngested` and `bytesReceivedStats` metrics anyway. Unfortunately, I don't see a way to test out those metrics. Happy to learn about how to do so! Do we have any examples? I couldn't find any.

In general we use `testutil.GatherAndCompare()` to check the metrics, so, you can use unique tenant id in this test and check the metric from prometheus.DefaultRegistry to check the size is calculated correctly.

sandeepsukhani:
> As far as I can see, we are not testing the `bytesIngested` and `bytesReceivedStats` metrics anyway. Unfortunately, I don't see a way to test out those metrics. Happy to learn about how to do so! Do we have any examples? I couldn't find any.

You can also use `testutil` from Prometheus to get the current value from registered metrics. You can find a couple of examples in the [tests here](https://github.com/grafana/loki/blob/ad0c54385be7e93b9c7909ab6ec4ff6e8de3d9d1/pkg/storage/stores/indexshipper/compactor/compactor_test.go#L376-L397).

pkg/loghttp/push/push.go
```diff
@@ -132,9 +133,15 @@ func ParseRequest(logger log.Logger, userID string, r *http.Request, tenantsRete
 		}
 		for _, e := range s.Entries {
 			totalEntries++
-			entriesSize += int64(len(e.Line))
-			bytesIngested.WithLabelValues(userID, retentionHours).Add(float64(int64(len(e.Line))))
-			bytesReceivedStats.Inc(int64(len(e.Line)))
+			var entryLabelsSize int64
+			for _, l := range e.NonIndexedLabels {
+				entryLabelsSize += int64(len(l.Name)) + int64(len(l.Value))
```
sandeepsukhani, line 138: you can do the conversion just once
```suggestion
				entryLabelsSize += int64(len(l.Name) + len(l.Value))
```

sandeepsukhani (commented):
The changes look good to me, left just a minor nit. Let us add some tests as suggested by Vlad.

salvacorts:
Thank you @sandeepsukhani , I've added a test for the metrics :)

sandeepsukhani (approved):
LGTM

PR #9809, opened 2023-06-27
disabled indexGateway client for boltdbShipper in migration tool.

vlad-diachenko:
Without this fix, the migration tool tries to initialize the ingexGateway client, which is unreachable. 
This leads to the error:
`Failed to create source store: error creating index client: index gateway grpc dial: failed to build resolver: passthrough: received empty target in Build()`


ashwanthgoli (approved):
lgtm

slim-bean (approved):

PR #9811, opened 2023-06-27
adds tsdb-max-bytes-per-shard limit

owen-d:
Adds the per tenant limit `tsdb-max-bytes-per-shard` which is used in configuring the shard size for tsdb subqueries. This effectively gives control over how big subqueries should be (assuming they're shardable). The default is no different (`600MB`) than the previously hardcoded initial value.

This should help us iterate to find optimal shard sizes to improve operations in the long term.


slim-bean (approved):
LGTM

MasslessParticle (approved):
lgtm

PR #9812, opened 2023-06-27
Implement series volume range queries

trevorwhitney:
This PR adds a `series_volume_range` endpoint which allows for series volume queries over time with a specified step, returning timeseries data in the form of a Prometheus matrix response.

The existing `series_volume` endpoint still returns Prometheus vector responses, and hardcodes the step to 0.


pkg/loghttp/query.go
```diff
@@ -347,33 +348,63 @@ func ParseIndexStatsQuery(r *http.Request) (*RangeQuery, error) {
 }
 
 func ParseSeriesVolumeQuery(r *http.Request) (*RangeQuery, error) {
```
MasslessParticle, line 350: I'm not sure we have to change anything here. Calling `ParseRangeQuery` with a custom limit should still be sufficient because we can handle forcing the step to 0 in the Request encoder/decoder.

pkg/querier/queryrange/roundtrip.go
```diff
@@ -321,13 +320,45 @@ func (r roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		level.Info(logger).Log(
+			"msg", "executing query",
+			"type", "series_volume",
+			"query", volReq.Matchers,
+			"length", volReq.Through.Sub(volReq.From),
+			"step", volReq.Step,
+			"limit", volReq.Limit)
+
+		instantReq, err := LokiCodec.EncodeRequest(req.Context(), volReq)
+		if !ok {
```
MasslessParticle, line 343: Should this be `if err != nil`?

pkg/querier/queryrange/roundtrip.go
```diff
@@ -321,13 +320,45 @@ func (r roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 
 		return r.indexStats.RoundTrip(req)
 	case SeriesVolumeOp:
+		queryReq, err := LokiCodec.DecodeRequest(req.Context(), req, nil)
```
MasslessParticle, line 323: We talked about putting this here but I'm concerned that nothing else is using the codec in `RoundTrip`. Would it be better to copy the instant and range examples and have separate parse functions? wdyt?

pkg/querier/queryrange/series_volume.go
```diff
@@ -0,0 +1,216 @@
...
+				return nil, err
+			}
+
+			promResp := toPrometheusResponse(collector.responses)
+			return promResp, nil
+		})
+	})
+}
+
+type concurrentVolumeResponseCollector struct {
```
MasslessParticle, line 111: If we're going to use a type for collecting responses like this, we need to be sure and protect `responses` on read, as well. I know we never call `collector.responses` before we're done but, as is, this a foot-gun for someone in the future.

Another approach you could take is a channel of 
```go
struct {
  bucket ...
  response ...
}
```

This would leave the `toPrometheusResponse` code generally unchanged and might be simpler?

pkg/querier/queryrange/series_volume.go
```diff
@@ -0,0 +1,216 @@
...
+		})
+
+		sortableResult = append(sortableResult, sortableSampleStream{
+			labels:  lbls,
+			samples: samples,
+		})
+	}
+
+	sort.Slice(sortableResult, func(i, j int) bool {
+		return sortableResult[i].labels.String() < sortableResult[j].labels.String()
```
MasslessParticle, line 201: It might be good to store the label string on the sortable type too. `.String()` allocates so this is surprisingly expensive if there are a lot of results.

pkg/storage/stores/tsdb/single_file_index.go
```diff
@@ -384,7 +385,17 @@ func (i *TSDBIndex) SeriesVolume(ctx context.Context, _ string, from, through mo
 					seriesNames[hash] = seriesLabels.String()
 				}
 
-				volumes[seriesNames[hash]] += stats.KB << 10 // Return bytes
+				size := stats.KB << 10 // Return bytes
```
MasslessParticle, line 388: I don't think we want this here. This will cause the results from `stats` and this function to differ for the same inputs. 

_we do_ want it in the ingester instance.

I think this is handled by the `v3 index reader` already but we should validate and add it as a separate PR.

pkg/loghttp/query.go
trevorwhitney: `ParseRangeQuery` does not work because of the step logic used in 2 places.

1. There are cases where this is called and we already have a 0 step, which causes an error to be thrown here:
```
	if result.Step <= 0 {
		return nil, errZeroOrNegativeStep
	}
```

2. There are other cases where this is called before step has been set. Step does not need to be present in the request as we hardcode it based on the route, however this also fails because of how `ParseRangeQuery` sets a default step:
```
func step(r *http.Request, start, end time.Time) (time.Duration, error) {
	value := r.Form.Get("step")
	if value == "" {
		return time.Duration(defaultQueryRangeStep(start, end)) * time.Second, nil
	}
	return parseSecondsOrDuration(value)
}
```

pkg/storage/stores/tsdb/single_file_index.go
trevorwhitney: removed

pkg/loki/modules.go
```diff
@@ -415,9 +415,10 @@ func (t *Loki) initQuerier() (services.Service, error) {
...
 		"/loki/api/v1/label/{name}/values": labelsHTTPMiddleware.Wrap(http.HandlerFunc(t.querierAPI.LabelHandler)),
 
-		"/loki/api/v1/series":              querier.WrapQuerySpanAndTimeout("query.Series", t.querierAPI).Wrap(http.HandlerFunc(t.querierAPI.SeriesHandler)),
-		"/loki/api/v1/index/stats":         indexStatsHTTPMiddleware.Wrap(http.HandlerFunc(t.querierAPI.IndexStatsHandler)),
-		"/loki/api/v1/index/series_volume": querier.WrapQuerySpanAndTimeout("query.SeriesVolume", t.querierAPI).Wrap(http.HandlerFunc(t.querierAPI.SeriesVolumeHandler)),
+		"/loki/api/v1/series":                    querier.WrapQuerySpanAndTimeout("query.Series", t.querierAPI).Wrap(http.HandlerFunc(t.querierAPI.SeriesHandler)),
+		"/loki/api/v1/index/stats":               indexStatsHTTPMiddleware.Wrap(http.HandlerFunc(t.querierAPI.IndexStatsHandler)),
+		"/loki/api/v1/index/series_volume":       querier.WrapQuerySpanAndTimeout("query.SeriesVolume", t.querierAPI).Wrap(http.HandlerFunc(t.querierAPI.SeriesVolumeInstantHandler)),
+		"/loki/api/v1/index/series_volume_range": querier.WrapQuerySpanAndTimeout("query.SeriesVolume", t.querierAPI).Wrap(http.HandlerFunc(t.querierAPI.SeriesVolumeRangeHandler)),
```
MasslessParticle, line 421: nit: We probably want to differentiate instant and range queries in our spans

MasslessParticle (approved):
One tiny nit but otherwise looks great

PR #9813, opened 2023-06-27
Support content negotiation between query frontend and querier.

jeschkies:
Currently, the querier sends results to the query frontend in JSON which is then decoded to Protobuf. It is more efficient to send the results as Protobuf.

This will also allow to extend the results with custom data structures.

The change is backwards compatible through content negotiation.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/querier/http.go
```diff
@@ -98,9 +97,17 @@ func (q *QuerierAPI) RangeQueryHandler(w http.ResponseWriter, r *http.Request) {
 		serverutil.WriteError(err, w)
 		return
 	}
-	if err := marshal.WriteQueryResponseJSON(result, w); err != nil {
+
+	if r.Header.Get("Accept") == queryrange.ProtobufType {
+		// TODO: should this rather be: application/vnd.google.protobuf; proto=queryrange.QueryResponse
```
jeschkies, line 102: I vote to leave it as `application/vnd.google.protobuf`

dannykopping: Agreed; `Accept` headers should be for content types

pkg/querier/queryrange/codec.go
```diff
@@ -606,11 +678,13 @@ func (Codec) EncodeResponse(ctx context.Context, res queryrangebase.Response) (*
 			Data:       logqlmodel.Streams(streams),
 			Statistics: response.Statistics,
 		}
-		if loghttp.Version(response.Version) == loghttp.VersionLegacy {
+		if version == loghttp.VersionLegacy {
+			// NOTE: This is called with proto buf flag :/
```
dannykopping, line 682: Can you explain this pls?

pkg/querier/queryrange/codec_test.go
```diff
@@ -284,9 +289,456 @@ func Test_codec_DecodeResponse(t *testing.T) {
...
+					"filename": "\/var\/hostlog\/syslog",
+					"job": "varlogs"
+				  },
+				  "value": [
+					1568404331.324,
+					"3.45"
+				  ]
+				}
+			  ],
+			  "stats" : {
```
dannykopping, line 461: Nit: I wonder if it's not worth moving this out and replacing with a variable for readability

pkg/querier/queryrange/marshal.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/grafana/loki/pkg/querier/queryrange/queryrangebase"
+	"github.com/grafana/loki/pkg/storage/stores/index/stats"
+)
+
+const (
+	JSONType     = `application/json; charset=utf-8`
+	ProtobufType = `application/vnd.google.protobuf`
+)
+
+// WriteQueryResponseProtobuf marshals the promql.Value to v1 loghttp JSON and then
```
dannykopping, line 24: JSON?

pkg/querier/queryrange/queryrange.proto
```diff
@@ -164,3 +164,14 @@ message VolumeResponse {
     (gogoproto.customtype) = "github.com/grafana/loki/pkg/querier/queryrange/queryrangebase/definitions.PrometheusResponseHeader"
   ];
 }
+
+message QueryResponse {
```
dannykopping, line 168: ❤️❤️❤️❤️

pkg/querier/queryrange/queryrangebase/roundtrip.go
```diff
@@ -46,8 +46,14 @@ type Config struct {
...
 	MaxRetries           int                `yaml:"max_retries"`
 	ShardedQueries       bool               `yaml:"parallelise_shardable_queries"`
+
+	// TODO(karsten): remove unsed optoon ForwardHeaders with Loki 3.0 since
+	// it's a breaking change.
 	// List of headers which query_range middleware chain would forward to downstream querier.
 	ForwardHeaders flagext.StringSlice `yaml:"forward_headers_list"`
+
+	// Accepted format for querier responses
+	AcceptedQueryResponseFormat string `yaml:"accepted_query_response_format"`
```
dannykopping, line 56: Why not both?

pkg/querier/queryrange/roundtrip.go
```diff
@@ -132,46 +132,51 @@ func NewTripperware(
 		}
 	}
 
-	indexStatsTripperware, err := NewIndexStatsTripperware(cfg, log, limits, schema, LokiCodec, statsCache,
+	var codec queryrangebase.Codec = DefaultCodec
+	if cfg.AcceptedQueryResponseFormat == "protobuf" {
```
dannykopping, line 136: Can we not detect this automatically?

dannykopping (commented):
Looking great! Left mostly nits

docs/sources/configuration/_index.md
```diff
@@ -791,6 +791,11 @@ results_cache:
 # CLI flag: -frontend.forward-headers-list
 [forward_headers_list: <list of strings> | default = []]
 
+# The downstream querier is requested to answer in the accepted format. Can be
+# 'json' or 'protobuf'. Note: Both will still be routed over GRPC.
+# CLI flag: -frontend.accepted-query-response-format
+[accepted_query_response_format: <string> | default = "json"]
```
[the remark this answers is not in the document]
jeschkies: We handle both based on the `Accept` header. However, this option tells the frontend to ask for the Protobuf encoded response instead of the JSON encoded response. Maybe the option is not clear enough.

pkg/querier/queryrange/codec.go
```diff
@@ -35,10 +35,14 @@ import (
 	marshal_legacy "github.com/grafana/loki/pkg/util/marshal/legacy"
 )
 
-var LokiCodec = &Codec{}
+var DefaultCodec = &Codec{}
 
 type Codec struct{}
 
+type RequestProtobufCodec struct {
```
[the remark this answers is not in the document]
jeschkies: It's funny. That was the naming scheme I've used originally. However, `Codec` is doing encoding and decoding based on the `Accept` and `Content-Type` headers. So it's doing both. All that `RequestProtobufCodec` does is set a different `Accept` header to request Protobuf encoded responses.

pkg/querier/queryrange/codec.go
```diff
@@ -585,7 +609,55 @@ func (Codec) DecodeResponse(_ context.Context, r *http.Response, req queryrangeb
 	}
 }
 
-func (Codec) EncodeResponse(ctx context.Context, res queryrangebase.Response) (*http.Response, error) {
+func decodeResponseProtobuf(r *http.Response, req queryrangebase.Request) (queryrangebase.Response, error) {
+	var buf []byte
```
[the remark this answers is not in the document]
jeschkies: Ok, I'll preallocate the body length for starters. The original code has the same issue.

pkg/querier/queryrange/codec.go
jeschkies: Debugging. Thanks for catching it.

pkg/querier/queryrange/codec.go
```diff
@@ -649,14 +723,50 @@ func (Codec) EncodeResponse(ctx context.Context, res queryrangebase.Response) (*
 
 	resp := http.Response{
 		Header: http.Header{
-			"Content-Type": []string{"application/json"},
+			"Content-Type": []string{"application/json; charset=UTF-8"},
```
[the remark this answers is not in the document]
jeschkies: The [`ResponseJSONMiddleware`](https://github.com/grafana/loki/blob/main/pkg/util/server/middleware.go#L26) is adding this header for all. So I'm doing it here now. I'm happy to remove it.

pkg/querier/queryrange/roundtrip.go
jeschkies: We do. This is for requesting the protobuf from the querier by the frontend even if the frontend received a request for JSON.

pkg/querier/http.go
```diff
@@ -98,9 +96,17 @@ func (q *QuerierAPI) RangeQueryHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
-	if err := marshal.WriteQueryResponseJSON(result, w); err != nil {
+
+	if r.Header.Get("Accept") == queryrange.ProtobufType {
+		w.Header().Add("Content-Type", queryrange.ProtobufType)
+		err = queryrange.WriteQueryResponseProtobuf(params, result, w)
+	} else {
+		w.Header().Add("Content-Type", queryrange.JSONType)
+		err = marshal.WriteQueryResponseJSON(result, w)
+	}
```
cstyan, line 106: should these be switch statements with a default case that logs an error?

pkg/querier/queryrange/roundtrip.go
cstyan: Am I missing something, won't queries from external sources always be returned a JSON response? So this line is required regardless of what the external results format should be, this sets the format we want to use internally between services.

docs/sources/configuration/_index.md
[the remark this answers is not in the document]
cstyan: maybe the name just needs to change? `requested_query_response_format`, the comment could then be `The downstream querier is requested to send responses to the query-frontend in the following format.`

pkg/querier/http.go
```diff
@@ -408,7 +434,13 @@ func (q *QuerierAPI) SeriesHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	err = marshal.WriteSeriesResponseJSON(*resp, w)
+	if r.Header.Get("Accept") == queryrange.ProtobufType {
```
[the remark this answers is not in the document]
cstyan: Do we have a set of types already defined for the request type? could we have a generic `WriteJSONResponse` and `WriteProtoResponse` that both take in a parameter for the request type switch on that parameter to call the correct `queryrange.WriteXResponseProto` or `queryrange.WriteXResponseJSON`?

pkg/querier/queryrange/roundtrip.go
jeschkies: > So this line is required regardless of what the external results format should be, this sets the format we want to use internally between services.

Yes. However, we don't have to set it if we want JSON because we default to JSON.

pkg/querier/http.go
jeschkies: That would be a breaking change. If no `Accept` header is present we default to JSON.

pkg/querier/http.go
```diff
@@ -189,9 +187,16 @@ func (q *QuerierAPI) LogQueryHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	if err := marshal_legacy.WriteQueryResponseJSON(result, w); err != nil {
+	if r.Header.Get("Accept") == queryrange.ProtobufType {
+		w.Header().Add("Content-Type", queryrange.ProtobufType)
+		err = queryrange.WriteQueryResponseProtobuf(&params, result, w)
+	} else {
+		w.Header().Add("Content-Type", queryrange.JSONType)
+		err = marshal_legacy.WriteQueryResponseJSON(result, w)
+	}
```
jeschkies, line 196: I kept this instead of `queryrange.WriteResponse` because it's not easy to see when to use `marshal_legacy`.

jeschkies (commented):
It's funny that we don't have tests for the querier HTTP handlers. I'll add one just for serialization.

pkg/querier/queryrange/queryrangebase/roundtrip.go
```diff
@@ -47,13 +47,13 @@ type Config struct {
 	MaxRetries           int                `yaml:"max_retries"`
 	ShardedQueries       bool               `yaml:"parallelise_shardable_queries"`
 
-	// TODO(karsten): remove unsed optoon ForwardHeaders with Loki 3.0 since
+	// TODO(karsten): remove used optoon ForwardHeaders with Loki 3.0 since
```
cstyan, line 50: `optoon` -> `option`

cstyan (approved):

JStickler (approved):
[docs team] LGTM

pkg/querier/http.go
```diff
@@ -206,7 +189,7 @@ func (q *QuerierAPI) LogQueryHandler(w http.ResponseWriter, r *http.Request) {
 
 	if r.Header.Get("Accept") == queryrange.ProtobufType {
 		w.Header().Add("Content-Type", queryrange.ProtobufType)
-		err = queryrange.WriteQueryResponseProtobuf(params, result, w)
+		err = queryrange.WriteQueryResponseProtobuf(&params, result, w)
 	} else {
 		w.Header().Add("Content-Type", queryrange.JSONType)
 		err = marshal_legacy.WriteQueryResponseJSON(result, w)
```
dannykopping: Should this not be replaced with `WriteResponse`?

pkg/querier/queryrange/marshal.go
```diff
@@ -14,16 +15,45 @@ import (
...
+	"github.com/grafana/loki/pkg/util/marshal"
 )
 
 const (
 	JSONType     = `application/json; charset=utf-8`
 	ProtobufType = `application/vnd.google.protobuf`
 )
 
-// WriteQueryResponseProtobuf marshals the promql.Value to v1 loghttp JSON and then
+func WriteResponse(req *http.Request, params *logql.LiteralParams, v interface{}, w http.ResponseWriter) error {
```
dannykopping, line 26: 👌 nice!

pkg/querier/queryrange/queryrangebase/roundtrip.go
```diff
@@ -64,7 +64,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.ShardedQueries, "querier.parallelise-shardable-queries", true, "Perform query parallelisations based on storage sharding configuration and query ASTs. This feature is supported only by the chunks storage engine.")
 	f.Var(&cfg.ForwardHeaders, "frontend.forward-headers-list", "List of headers forwarded by the query Frontend to downstream querier.")
 
-	f.StringVar(&cfg.AcceptedQueryResponseFormat, "frontend.accepted-query-response-format", "json", "The downstream querier is requested to answer in the accepted format. Can be 'json' or 'protobuf'. Note: Both will still be routed over GRPC.")
+	f.StringVar(&cfg.RequestedQueryResponseFormat, "frontend.requested-query-response-format", "json", "The downstream querier is requested to answer in the accepted format. Can be 'json' or 'protobuf'. Note: Both will still be routed over GRPC.")
```
dannykopping, line 67: Nit: I think "requested" here is a confusing name.
IMHO should be `required`; this is not a request that the querier can choose to ignore, right?

pkg/querier/queryrange/codec.go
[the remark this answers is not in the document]
dannykopping: Right, but we're allocating memory for every `io.ReadAll` which is going to be discarded after each request. I'm not sure if this will escape to the heap, but it could because the size of the body is not known apriori i think?

Can you add a benchmark for this?

pkg/querier/queryrange/marshal.go
dannykopping: Why was this marked resolved? Looks like a typo in the comment, no?

pkg/querier/http.go
jeschkies: My comment was lost. Yes, however, it's hard to decide whether to use `marshal_legacy.WriteQueryResponseJSON` or `marshal.WriteQueryResponseJSON` based on the result type. It can only be determined here by the handler itself.

dannykopping (approved):
Let's gooooo!

PR #9833, opened 2023-06-29
Add limit to the series volume endpoint

MasslessParticle:
To protect index gateways, this PR adds a per-tenant limit to the `series_volume` endpoint so the accumulation of series can't grow without bound.


JStickler (approved):
[docs team] LGTM

MichelHollands (approved):
LGTM

docs/sources/configuration/_index.md
```diff
@@ -2467,6 +2467,10 @@ The `limits_config` block configures global and per-tenant limits in Loki.
 # Enable log-volume endpoints.
 [volume_enabled: <boolean>]
 
+# The maximum number of aggregated series in a log-volume response
+# CLI flag: -limits.volume-max-series
+[volume_max_series: <int> | default = 1000]
```
trevorwhitney, line 2472: could we use `max_query_series` as the limit instead? I like all the additional wiring in this PR to treat this as a different limit, but I'm curious if the underlying value should just come from `max_query_series` for now to avoid the additional configuration, and we can always easily change that later since it's all wired up as a separate limit?

pkg/storage/stores/index/seriesvolume/volume.go
```diff
@@ -10,29 +11,36 @@ import (
...
+	limit           int32
+	volumeMaxSeries int
 }
 
-func NewAccumulator(limit int32) *Accumulator {
+func NewAccumulator(limit int32, maxSize int) *Accumulator {
 	return &Accumulator{
-		volumes: make(map[string]uint64),
-		limit:   limit,
+		volumes:         make(map[string]uint64),
```
trevorwhitney, line 28: would there be a benefit to allocating `volumes` to a capacity of `volumeMaxSeries`?

trevorwhitney (changes requested):
Looks good, only thing I want to consider before moving forward is if we want a new actual limit, or if we just want a new set of wrapper/getter functions around `max_query_series`?

trevorwhitney (approved):
we chatted out-of-band about using `max_query_series`, and I'm ok with moving forward with a new config. while validating the concerns of additional configuration service area, @MasslessParticle made a convincing argument for adding the new config for a) operation clarity, and b) because the trade off in protection against OOMing query frontends was worth the additional config value. Our hope is once we gain experience with this new config value we can tune it appropriately so that most operators never need to touch it.

PR #9841, opened 2023-07-01
added ability to specify head block format version

vlad-diachenko:
Currently, the first byte of head block data identifies the type of the head block(ordered/unordered). However, using this approach, it's not possible to distinguish the version of the data that is stored in this head block. 
In the next version of the chunk format (V4), we add additional data and when we deserialize the data, it's necessary to know the version of the data format to use different deserialization strategies.

To make it possible, the first byte of the head block now specifies the version of the head block, not the type(ordered/unordered). 

### Special notes for your reviewer:
1. This PR does not enable the new format of the head block format and it will be enabled once chunk format V4 is added in [this PR](https://github.com/grafana/loki/pull/9700).
2. `VersionedHeadBlockFmtV1 byte = 128` - as long as previously the first byte of the head block was the chunk format version, I decided to use 128 as a baseline for head block format versioning. it allows to avoid the collision with chunk formats v1,v2 if somebody decides to upgrade to the latest version of Loki from some legacy version.

### Checklist
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/chunkenc/unordered.go
+++ b/pkg/chunkenc/unordered.go
@@ -49,9 +49,11 @@ type HeadBlock interface {
 		extractor log.StreamSampleExtractor,
 	) iter.SampleIterator
 	Format() HeadBlockFmt
+	ChunkDataFormat() byte
 }
 
 type unorderedHeadBlock struct {
+	chunkDataFormat byte
 	// Opted for range tree over skiplist for space reduction.
 	// Inserts: O(log(n))
 	// Scans: (O(k+log(n))) where k=num_scanned_entries & n=total_entries
@@ -62,13 +64,18 @@ type unorderedHeadBlock struct {
 	mint, maxt int64 // upper and lower bounds
 }
 
-func newUnorderedHeadBlock() *unorderedHeadBlock {
+func newUnorderedHeadBlock(chunkDataFormat byte) *unorderedHeadBlock {
 	return &unorderedHeadBlock{
-		rt: rangetree.New(1),
+		chunkDataFormat: chunkDataFormat,
+		rt:              rangetree.New(1),
 	}
 }
 
 func (hb *unorderedHeadBlock) Format() HeadBlockFmt { return UnorderedHeadBlockFmt }
+
+func (hb *unorderedHeadBlock) ChunkDataFormat() byte {
+	return hb.chunkDataFormat
+}
 
 func (hb *unorderedHeadBlock) IsEmpty() bool {
 	return hb.size == 0
@@ -87,7 +94,7 @@ func (hb *unorderedHeadBlock) UncompressedSize() int {
 }
 
 func (hb *unorderedHeadBlock) Reset() {
-	x := newUnorderedHeadBlock()
+	x := newUnorderedHeadBlock(hb.chunkDataFormat)
 	*hb = *x
 }
 
@@ -369,10 +376,10 @@ func (hb *unorderedHeadBlock) Serialise(pool WriterPool) ([]byte, error) {
 }
 
 func (hb *unorderedHeadBlock) Convert(version HeadBlockFmt) (HeadBlock, error) {
-	if version > OrderedHeadBlockFmt {
+	if version == UnorderedHeadBlockFmt {
 		return hb, nil
 	}
-	out := version.NewBlock()
+	out := version.NewBlock(hb.chunkDataFormat)
 
 	err := hb.forEntries(
 		context.Background(),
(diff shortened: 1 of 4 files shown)


sandeepsukhani (commented):
The changes look good to me. I just left some minor suggestions.

owen-d (commented):
> To make it possible, the first byte of the head block now specifies the version of the head block, not the type(ordered/unordered).

Conceptually, these are the same thing. Ordered vs Unordered are just the two versions we support (with clearer names than something like HeadV1 vs HeadV2). Even if we do go this new route you suggest, I don't see a reason to start with `VersionedHeadBlockFmtV1 = 128` (we could just use the next incremented number and add a second bytes encoded after that).

### Can we go about this in a different way? 

I think it's much simpler to add a new head format version (`UnorderedWithMetadataFmt`) that will (1) accept unordered entries and (2) encode non-indexed metadata. It's not intended that we'll need to mix-and-match different headblock vs chunk formats. Asides the issue of some of this being conceptually impossible (we can't have a head block accept non-indexed-metadata but then write to an older chunk format that doens't support it), it's intended that loki starts writing new versions while being able to read old versions and not the other way around.

In summary, I think it's clearer and simpler to add a new head format that accepts unordered entries with optional non-indexed-metadata: `UnorderedWithMetadataFmt`. WDYT?

vlad-diachenko:
> > To make it possible, the first byte of the head block now specifies the version of the head block, not the type(ordered/unordered).
> 
> Conceptually, these are the same thing. Ordered vs Unordered are just the two versions we support (with clearer names than something like HeadV1 vs HeadV2). Even if we do go this new route you suggest, I don't see a reason to start with `VersionedHeadBlockFmtV1 = 128` (we could just use the next incremented number and add a second bytes encoded after that).
> 
> ### Can we go about this in a different way?
> I think it's much simpler to add a new head format version (`UnorderedWithMetadataFmt`) that will (1) accept unordered entries and (2) encode non-indexed metadata. It's not intended that we'll need to mix-and-match different headblock vs chunk formats. Asides the issue of some of this being conceptually impossible (we can't have a head block accept non-indexed-metadata but then write to an older chunk format that doens't support it), it's intended that loki starts writing new versions while being able to read old versions and not the other way around.
> 
> In summary, I think it's clearer and simpler to add a new head format that accepts unordered entries with optional non-indexed-metadata: `UnorderedWithMetadataFmt`. WDYT?


Hey @owen-d . yes, if we are going to remove/deprecate ordered writes, we can just add a new format of a head like `UnorderedWithMetadataFmt`. However, currently, each user can disable/enable `unordered-writes` and Loki decides what format of head block to use: `OrderedHeadBlockFmt`, `UnorderedHeadBlockFmt`. So, with the current implementation of Loki, it's necessary to add 2 new formats `OrderedWithMetadataFmt` and `UnorderedWithMetadataFmt` because some users can have `unordered-writes` disabled but still want to use non-indexed metadata. 

however, even with current implementation, it does not stop us from removing/deprecating ordered writes in future.

pkg/chunkenc/unordered.go
```diff
@@ -369,10 +376,10 @@ func (hb *unorderedHeadBlock) Serialise(pool WriterPool) ([]byte, error) {
 }
 
 func (hb *unorderedHeadBlock) Convert(version HeadBlockFmt) (HeadBlock, error) {
-	if version > OrderedHeadBlockFmt {
+	if version == UnorderedHeadBlockFmt {
```
owen-d, line 379: thanks for changing this, I'm not sure what I was thinking :)

owen-d (commented):
Regardless of the route we chose, the main thing we're going to need to implement is the ability to write entries with non-indexed-metadata in both the unordered head block as well as in the blocks. If we choose to do this with the old ordered-only head as well instead of deprecating+removing ordered-only writes, that's fine too :).

The problem with using your `(head_version, block_version)` tuple is that it forces us to support the matrix of _all_ `head_versions` and _all_ `block_versions` in the future. In reality, we may want to add a new `head_version` or `block_version` without needing to make it compatible with all formats of the other.

Does that make sense?

sandeepsukhani:
> However, currently, each user can disable/enable unordered-writes and Loki decides what format of head block to use: OrderedHeadBlockFmt, UnorderedHeadBlockFmt. 

We default to enabling unordered writes and it is highly unlikely someone would go out of the way and disable unordered writes. Less code is always better so I think it would be better to keep it simple i.e always create unordered blocks and use chunk format version going forward.

pkg/chunkenc/unordered.go
vlad-diachenko: I believe you wanted to make this logic flexible to use unorderedHeadBlock for all the versions after OrderedHeadBlockFmt ;) like 4, 5,6,7,etc...

vlad-diachenko:
closed in favor of https://github.com/grafana/loki/pull/9700/commits/529ad9dd943b81193dab36ddd2aa651b7e2c885f

PR #9844, opened 2023-07-01
Helm: Add tsdb schema config

faustodavid:
To take advantage of the new tsdb schema.

I have also updated the `query_scheduler.max_outstanding_requests_per_tenant` and `querier. max_concurrent ` following the recommendations from this post:
https://grafana.com/docs/loki/latest/operations/storage/tsdb

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/docs/sources/installation/helm/reference.md
+++ b/docs/sources/installation/helm/reference.md
@@ -2001,7 +2001,9 @@ null
 			<td>object</td>
 			<td>Optional querier configuration</td>
 			<td><pre lang="json">
-{}
+{
+  "max_concurrent": 16
+}
 </pre>
 </td>
 		</tr>
@@ -2010,7 +2012,9 @@ null
 			<td>object</td>
 			<td>Additional query scheduler config</td>
 			<td><pre lang="json">
-{}
+{
+  "max_outstanding_requests_per_tenant": 32768
+}
 </pre>
 </td>
 		</tr>
@@ -2082,7 +2086,30 @@ null
 			<td>object</td>
 			<td>Check https://grafana.com/docs/loki/latest/configuration/#schema_config for more info on how to configure schemas</td>
 			<td><pre lang="json">
-{}
+{
+  "configs": [
+    {
+      "from": "2022-01-11",
+      "index": {
+        "period": "24h",
+        "prefix": "loki_index_"
+      },
+      "object_store": "{{ .Values.loki.storage.type }}",
+      "schema": "v12",
+      "store": "boltdb-shipper"
+    },
+    {
+      "from": "2023-07-01",
+      "index": {
+        "period": "24h",
+        "prefix": "tsdb_index_"
+      },
+      "object_store": "{{ .Values.loki.storage.type }}",
+      "schema": "v12",
+      "store": "tsdb"
+    }
+  ]
+}
 </pre>
 </td>
 		</tr>
--- a/production/helm/loki/CHANGELOG.md
+++ b/production/helm/loki/CHANGELOG.md
@@ -13,6 +13,10 @@ Entries should include a reference to the pull request that introduced the chang
 
 [//]: # (<AUTOMATED_UPDATES_LOCATOR> : do not remove this line. This locator is used by the CI pipeline to automatically create a changelog entry for each new Loki release. Add other chart versions and respective changelog entries bellow this line.)
 
+## 6.0.0
+
+- [CHANGE] **BREAKING** Add [tsdb](https://grafana.com/docs/loki/latest/operations/storage/tsdb/) schema config and apply recommendations.
+
 ## 5.8.9
 
 - [BUGFIX] Fix loki/logs dashboard: allow querying multiple log level at once
--- a/production/helm/loki/Chart.yaml
+++ b/production/helm/loki/Chart.yaml
@@ -3,7 +3,7 @@ name: loki
 description: Helm chart for Grafana Loki in simple, scalable mode
 type: application
 appVersion: 2.8.2
-version: 5.8.9
+version: 6.0.0
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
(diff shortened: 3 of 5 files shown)


MichelHollands:
@faustodavid Thanks for this contribution. This will have an impact on people upgrading an existing Helm installation. Can you make the date for the tsdb_index configurable so it can be in the future? The migration impact should also be documented. Given the big change this could be 5.9 or even a 6.0 release.

faustodavid:
> @faustodavid Thanks for this contribution. This will have an impact on people upgrading an existing Helm installation. Can you make the date for the tsdb_index configurable so it can be in the future? The migration impact should also be documented. Given the big change this could be 5.9 or even a 6.0 release.

I moved the configuration to `schemaConfig` and added a comment with a link to the TSDB documentation.

Should I add something more to explain the impact?

MichelHollands:
@faustodavid Can you add **BREAKING** to the CHANGELOG entry. It should also be 6.0 as this is a breaking change. Also please add instruction on how to upgrade (ie set the TSDB start date in the future). Did you test this upgrade?

faustodavid:
> @faustodavid Can you add **BREAKING** to the CHANGELOG entry. It should also be 6.0 as this is a breaking change. Also please add instruction on how to upgrade (ie set the TSDB start date in the future). Did you test this upgrade?
> 
> Here's an example of a new version in the changelog with breaking entries: https://github.com/grafana/loki/blob/main/production/helm/loki/CHANGELOG.md#40

Hey @MichelHollands, I ran a test by deploying the helm version 5.8.6 to AKS and then applied the changes from this PR. All went well and no Loki error logs popped out except when the pods were restarting.

Once the updated version was rolled out, the TSDB index got added and Loki could read both schemas with ease.

image

I sent out 5 messages when the `boltdb` schema was turned on and then sent 5 more when we switched to `tsdb`.

image

Here's what I've figured: the date in the TSDB schema doesn't need to be future-dated. It just needs to come after the boltdb date. So, if folks aren't ready to use the new schema right away, they can just set this date to the future.

sandeepsukhani:
> > @faustodavid Can you add **BREAKING** to the CHANGELOG entry. It should also be 6.0 as this is a breaking change. Also please add instruction on how to upgrade (ie set the TSDB start date in the future). Did you test this upgrade?
> > Here's an example of a new version in the changelog with breaking entries: https://github.com/grafana/loki/blob/main/production/helm/loki/CHANGELOG.md#40
> 
> Hey @MichelHollands, I ran a test by deploying the helm version 5.8.6 to AKS and then applied the changes from this PR. All went well and no Loki error logs popped out except when the pods were restarting.
> 
> Once the updated version was rolled out, the TSDB index got added and Loki could read both schemas with ease.
> 
> image
> I sent out 5 messages when the `boltdb` schema was turned on and then sent 5 more when we switched to `tsdb`.
> 
> image
> Here's what I've figured: the date in the TSDB schema doesn't need to be future-dated. It just needs to come after the boltdb date. So, if folks aren't ready to use the new schema right away, they can just set this date to the future.

I do not think it would work because we use schema config at query time to find the relevant tables. You need to check if the logs are really coming from both tables or just the tsdb. You can verify that by renaming(adding any letter at the beginning) or removing either of the tables in the object storage and running the queries again after restarting Loki.

faustodavid:
> > > @faustodavid Can you add **BREAKING** to the CHANGELOG entry. It should also be 6.0 as this is a breaking change. Also please add instruction on how to upgrade (ie set the TSDB start date in the future). Did you test this upgrade?
> > > Here's an example of a new version in the changelog with breaking entries: https://github.com/grafana/loki/blob/main/production/helm/loki/CHANGELOG.md#40
> > 
> > 
> > Hey @MichelHollands, I ran a test by deploying the helm version 5.8.6 to AKS and then applied the changes from this PR. All went well and no Loki error logs popped out except when the pods were restarting.
> > Once the updated version was rolled out, the TSDB index got added and Loki could read both schemas with ease.
> > image
> > I sent out 5 messages when the `boltdb` schema was turned on and then sent 5 more when we switched to `tsdb`.
> > image
> > Here's what I've figured: the date in the TSDB schema doesn't need to be future-dated. It just needs to come after the boltdb date. So, if folks aren't ready to use the new schema right away, they can just set this date to the future.
> 
> I do not think it would work because we use schema config at query time to find the relevant tables. You need to check if the logs are really coming from both tables or just the tsdb. You can verify that by renaming(adding any letter at the beginning) or removing either of the tables in the object storage and running the queries again after restarting Loki.

@sandeepsukhani, I did a few tests and everything worked smoothly. When I deleted the `boltdb` index folder and restarted the pods, I couldn't find the logs I previously made in `boltdb`. I only found those I logged with `tsdb`. But, after putting the index folder back and restarting, the logs were visible again.

faustodavid:
I have also tried locally using the example https://github.com/grafana/loki/tree/main/examples/getting-started.

I did the following steps:
* Run the docker-compose
* Wait some minutes to have some sample data
* Do POST on the writer `http://localhost:3102/ingester/shutdown`
* Shutdown the reader
* Add tsdb schema config with a date older than today.
* Turn on the writer pod, then the reader
* Query the reader and validate that old data is there.
* Just to double check Do POST on the writer `http://localhost:3102/ingester/shutdown`
* Validate that both index folders are in Minio
* Turn on the writer and reader and test the data again

MichelHollands:
@faustodavid I'm afraid I can't approve this as it is. Adding the TSDB there with a hardcoded start date will mess things up when people upgrade.  Note that https://grafana.com/docs/loki/latest/operations/storage/tsdb/ says the TSDB date has to be in the future.

However if you remove the TSDB part from the schemaConfig setting then it can be merged. The other settings are useful and the having the schemaConfig configurable will help with the migration to TSDB. The TSDB part can be mentioned in the CHANGELOG as something that can be used for new users or added with a start date in the future for existing users.

PR #9853, opened 2023-07-04
cache: fix a mistake in returning cached error in the index list cache

sandeepsukhani:
I did a mistake in returning a cached error in the code pushed in PR #9798. Fixing it in the PR with some tests.

**Checklist**
- [x] Tests updated


ashwanthgoli (approved):
lgtm

PR #9854, opened 2023-07-04
Test querier handlers reponse format.

jeschkies:
This is a follow up to #9813 that introduced a handler test to verify the response format for the legacy and v1 APIs.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/logql/engine.go
```diff
@@ -140,23 +140,29 @@ func (opts *EngineOpts) applyDefault() {
 	}
 }
 
-// Engine is the LogQL engine.
-type Engine struct {
+type Engine interface {
```
dannykopping, line 143: Would it not be simpler to name this something else, to avoid all the refactoring?
`NewDefaultEngine` doesn't really mean anything.

pkg/querier/http_test.go
```diff
@@ -342,3 +468,119 @@ func TestSeriesVolumeHandler(t *testing.T) {
...
+								Timestamp: time.Unix(0, 123456789012345),
+								Line:      "super line",
+							},
+						},
+						Labels: `{foo="bar"}`,
+					},
+				},
+				Statistics: statsResult,
+			},
+			expectedRespone: `{
```
dannykopping, line 533: Are we only checking that JSON works?

jeschkies: Yes. Protobuf is already tested in the parity check. Here we test that the wiring works for the legacy format. Come to think we should probably remove the old endpoints with the next mayor release.

pkg/logql/engine.go
jeschkies: Fair point but what would you name it? What do you think about `QueryFactory`?

dannykopping: It's not a factory, though...
Idiomatic Go would suggest `Querier`, which seems kinda reasonable here?

jeschkies: It is a factory. The method `Query()` constructs a `Query` that can be executed via `Exec()`. It is a little confusing.

dannykopping: What I meant is: the interface is not a factory.
Anyway I don't mind what it's called; but let's rather rename this than change the engine name everywhere I think

jeschkies: I beg to differ (https://en.wikipedia.org/wiki/Factory_(object-oriented_programming)). Although, I'm wondering if the implementations are factories and this is a factory interface (https://www.baeldung.com/java-abstract-factory-pattern). Java...

Anyhow, I'll come up with something else.

pkg/querier/http.go
```diff
@@ -47,12 +47,16 @@ type QueryResponse struct {
 	Result     parser.Value     `json:"result"`
 }
 
+type Engine interface {
+	Query(logql.Params) logql.Query
+}
```
jeschkies, line 52: This seemed cleaner. The querier defines the interface it wants to receive. What do you think @dannykopping

dannykopping (approved):
LGTM 👍

PR #9857, opened 2023-07-04
Queriers: Stop emitting spans for every objectstorage call

DylanGuedes:
Stop emitting spans for every call to our object storages (like S3 or Azure); we don't emit them for GCP already. It causes every chunk fetching to emit a span, which can end with a huge amount of unnecessary spans. It will still report request metrics just fine.

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/storage/chunk/client/aws/s3_storage_client.go
+++ b/pkg/storage/chunk/client/aws/s3_storage_client.go
@@ -381,7 +381,7 @@ func (a *S3ObjectClient) GetObject(ctx context.Context, objectKey string) (io.Re
 			return nil, 0, errors.Wrap(ctx.Err(), "ctx related error during s3 getObject")
 		}
 
-		lastErr = loki_instrument.ObserveRequest(ctx, "S3.GetObject", s3RequestDuration, instrument.ErrorCode, func(ctx context.Context) error {
+		lastErr = loki_instrument.TimeRequest(ctx, "S3.GetObject", s3RequestDuration, instrument.ErrorCode, func(ctx context.Context) error {
 			var requestErr error
 			resp, requestErr = a.hedgedS3.GetObjectWithContext(ctx, &s3.GetObjectInput{
 				Bucket: aws.String(bucket),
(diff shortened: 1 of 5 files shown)


pkg/storage/chunk/client/aws/s3_storage_client.go
```diff
@@ -367,25 +368,30 @@ func (a *S3ObjectClient) bucketFromKey(key string) string {
 
 // GetObject returns a reader and the size for the specified object key from the configured S3 bucket.
 func (a *S3ObjectClient) GetObject(ctx context.Context, objectKey string) (io.ReadCloser, int64, error) {
-	var resp *s3.GetObjectOutput
-
 	// Map the key into a bucket
 	bucket := a.bucketFromKey(objectKey)
 
-	retries := backoff.New(ctx, a.cfg.BackoffConfig)
 	err := ctx.Err()
```
MichelHollands, line 374: Do you still need to set this to ctx.Err()? It is not used until another value is set for this field.

DylanGuedes: it is used at line 398, but not sure if it is really possible to have a ctx.Err without ever entering the `retries.Ongoing` block. Maybe if a deadline is exceeded in the moment this function is invoked? :thinking:

MichelHollands: In line 393 the error is set. It isn't checked before that. So there's no point to set it to something.

pkg/util/instrument/instrument.go
```diff
@@ -0,0 +1,24 @@
...
+	"context"
+	"time"
+
+	"github.com/weaveworks/common/instrument"
+)
+
+// ObserveRequest reports how much time was spent on the given function  `f`.
+//
+// It is a thinner version of weaveworks/common/instrument.CollectedRequest that doesn't emit spans.
+func ObserveRequest(ctx context.Context, method string, col instrument.Collector, toStatusCode func(error) string, f func(context.Context) error) error {
```
MichelHollands, line 13: Maybe you can all this function `TimeRequest` to make it clearer what it is observing.

MichelHollands (approved):

pkg/util/instrument/instrument.go
DylanGuedes: that's fair. naming is too hard :laughing:

PR #9860, opened 2023-07-04
operator: Do not use mergo for updating existing resources

xperimental:

While testing the zone-awareness PR I noticed that the labels and annotations on the pods did not disappear when I deactivated the zone-awareness again. This turned out to be an issue with how we generally update existing Deployments/StateFulSets: We merge the existing PodTemplateSpec with the desired one.

I don't think we need the merge at that point, because the operator should be the complete owner of that attribute, so we should be able to completely replace the values instead of merging with the existing one.

The situation with the annotations and labels of our resources is a bit confusing. On one hand we have `mergeWithOverride` code at the beginning of the `MutateFuncFor` function but on the other hand almost every one of the per-type `mutate*` functions contains code that overwrites existing labels/annotations. I think it should be either one or the other.

On a related note, maybe it makes sense to keep existing labels/annotations on the resources ("merging" them with the existing ones, but explicitly keep track of _our_ labels/annotations, so that we remove our annotations/labels when they are not needed anymore but don't touch "other people's annotations/labels". This would make the merge code a bit longer though.

This PR remains a draft until the question about how to proceed with labels/annotations is cleared up.

**Checklist**
- [x] Tests updated

### DIFF
--- a/operator/internal/manifests/mutate.go
+++ b/operator/internal/manifests/mutate.go
@@ -16,14 +16,24 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-// MutateFuncFor returns a mutate function based on the
-// existing resource's concrete type. It supports currently
-// only the following types or else panics:
-// - ConfigMap
-// - Service
-// - Deployment
-// - StatefulSet
-// - ServiceMonitor
+// MutateFuncFor returns a mutate function based on the existing resource's concrete type.
+// It currently supports the following types and will return an error for other types:
+//
+//   - ConfigMap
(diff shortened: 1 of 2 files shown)


JoaoBraveCoding (commented):
Well caught! Apart from thinking that we might hit Hyrum's Law I'm on the side of just setting the labels/annotations. I checked on prom-operator and there we also ["just" set labels](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/server/statefulset.go#L141-L152) (at least on the Prometheus StatefulSet).

periklis (approved):

PR #9717, opened 2023-06-15
TOC Sync: Move files under Get Started.

JStickler:
**Which issue(s) this PR fixes**:
Fixes #8711 

**Special notes for your reviewer**:
Part of the Database products table of contents alignment project.
Moves Getting Started and Fundamentals files under Get Started in TOC.
Updates front matter for the topic (headings, page weights, descriptions, aliases to redirect from old URL to new URL)
Fixes links broken by the move.

PR #9881, opened 2023-07-06
Update redirects to use new alias format. (part 1 of 2)

JStickler:
@jdbaldry commented on #9717 that the alias format had changed, but we [agreed](https://github.com/grafana/loki/pull/9717#issuecomment-1594960563) that I could update it in a follow up PR. This is that follow up.  Also updating previously merged file moves that used the old alias format.

Breaking the alias updates into two PRs to make for smaller reviews.

### DIFF
--- a/docs/sources/visualize/grafana.md
+++ b/docs/sources/visualize/grafana.md
@@ -4,6 +4,7 @@ menuTitle: Visualize data
 description: Visualize your log data with Grafana
 
 aliases:
+   - ../getting-started/grafana/
    - ../operations/grafana/
    - ../visualize/grafana/
 weight: 825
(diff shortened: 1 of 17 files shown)


docs/sources/visualize/grafana.md
```diff
@@ -4,9 +4,8 @@ menuTitle: Visualize data
 description: Visualize your log data with Grafana
 
 aliases:
-   - /docs/loki/latest/operations/grafana/
-   - /docs/loki/latest/visualize/grafana/
-   - /docs/loki/latest/getting-started/grafana/
+   - ../operations/grafana/
```
DylanGuedes, line 7: double checking: don't you want "getting-started" on this list?

DylanGuedes (approved):
lgtm just a nit

docs/sources/visualize/grafana.md
JStickler: Thanks for checking, this file was under "getting started" way back in version 2.4, I was thinking that was a long time ago, but it looks like it wasn't even two years ago.  I'll add that back in.

PR #9882, opened 2023-07-07
Implement probabilistic topk via count min sketch and min heap

cstyan:
****Note:** the sketchbf and heavykeeper implementations will be removed from this branch before merging into main, they're just here for context at the moment for those who might be interested in seeing what alternative structures we could have used.**

This PR currently implements the following structures:
- **Count Min Sketch**: http://dimacs.rutgers.edu/~graham/pubs/papers/cmencyc.pdf
- **sketch-bf**: https://ietresearch.onlinelibrary.wiley.com/doi/epdf/10.1049/ell2.12482 improvements to the heap usage for Count Min Sketch
- HeavyKeeper: https://www.usenix.org/system/files/conference/atc18/atc18-gong.pdf

I also read and evaluated the following:
- topkapi: https://www.cs.rice.edu/~as143/Papers/topkapi.pdf
- space saving algorithm: https://www.cse.ust.hk/~raywong/comp5331/References/EfficientComputationOfFrequentAndTop-kElementsInDataStreams.pdf

**Description**:

As part of our work to ensure every query type is parallelizable we started exploring probabilistic data structures as a way to reduce memory consumption and allow merging of partial results after sharding of queries. In this PR I have implemented a few sketch structures as part of investigating structures that are applicable to topk queries. 

Unfortunately while many papers made claims of improved performance over the original count min sketch paper, topkapi and heavy keeper either didn't behave as expected or were not mergeable while keeping the same degree of accuracy. Count min sketch when using conservative updates (which vastly reduces error over counting, especially for low flows), also can't keep the same degree of accuracy, but it is still more accurate after merging than heavy keeper. I've also applied some of the principles outlined in the sketchbf paper related to reducing the # of heap operations required when observing events into the sketch, which sped up processing in my benchmark of 100k unique streams each with 1-2k randomized # of occurrences by nearly 3x.


pkg/util/sketch/cms.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+	"math"
+)
+
+type CountMinSketch struct {
+	depth, width int
+	counters     [][]uint32
+}
+
+// NewCountMinSketch creates a new CMS for a given width and depth.
```
jeschkies, line 13: Could you add documentation on the parameters?

pkg/util/sketch/cms.go
```diff
@@ -0,0 +1,117 @@
...
+	return ret
+}
+
+func (s *CountMinSketch) getPos(h1, h2 uint32, row int) uint32 {
+	pos := (h1 + uint32(row)*h2) % uint32(s.width)
+	return pos
+}
+
+// Add 'count' occurences of the given input.
+func (s *CountMinSketch) Add(event string, count int) {
```
jeschkies, line 40: I wonder if we should make the event type generic. Maybe that's something for the next iteration.

pkg/util/sketch/cms_test.go
```diff
@@ -0,0 +1,52 @@
+//go:build sketch
```
jeschkies, line 1: That's fine for starters but we need functional testing that always runs.

pkg/util/sketch/cms_test.go
```diff
@@ -0,0 +1,52 @@
...
+	for z := numStreams - k; z < numStreams; z++ {
+		n := int64(rand.Int63n(int64(maxPerStream)) + 1 + max)
+		for x := 0; x < int(n); x++ {
+			events = append(events, event{name: strconv.Itoa(z), count: 1})
+		}
+		//total += n
+		//oTotal += int(n)
+	}
+
+	rand.Seed(time.Now().UnixNano())
```
jeschkies, line 44: The seed should be a constant in order to compare different runs.

pkg/util/sketch/hash.go
```diff
@@ -0,0 +1,52 @@
+package sketch
+
+import "hash/fnv"
+
+// Copyright (c) 2012-2017  Dustin Sallings <dustin@spy.net>
```
jeschkies, line 5: Out of curiosity why did you copy the function here?

pkg/util/sketch/heap.go
```diff
@@ -0,0 +1,72 @@
...
+
+type node struct {
+	event string
+	count uint32
+	// used for the container heap Fix function
+	index           uint16
+	sketchPositions []uint32
+}
+
+type MinHeap []*node
```
jeschkies, line 15: Should we reuse an existing heap implementation?

pkg/util/sketch/heap_test.go
```diff
@@ -0,0 +1,34 @@
+//go:build sketch
```
jeschkies, line 1: I think this should always run.

pkg/util/sketch/heavykeeper.go
```diff
@@ -0,0 +1,220 @@
...
+	"container/heap"
+	"github.com/axiomhq/hyperloglog"
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"math"
+	"math/rand"
+	"sort"
+)
+
+type heavyKeeper struct {
```
jeschkies, line 13: Are we going to use heavy keeper?

pkg/util/sketch/sketchbf.go
```diff
@@ -0,0 +1,220 @@
...
+	"math"
+	"sort"
+)
+
+// SketchBF is an extension of CMS + heap topk implementations
+// it uses an additional "bloom filter" of bits to know whether
+// a given thing is already part of the heap
+type SketchBF struct {
+	max int
+	//currentTop          map[string]uint32
```
jeschkies, line 16: Should this be removed?

pkg/util/sketch/sketchbf.go
```diff
@@ -0,0 +1,220 @@
...
+func NewSketchBF(k, w, d int) (*SketchBF, error) {
+	if d < 1 || w < 1 {
+		return nil, fmt.Errorf("sketch dimensions must be positive, w: %d, d: %d", w, d)
+	}
+	cms, _ := NewCountMinSketch(w, d)
+	//return &SketchBF{
+	//	depth: d,
+	//	width: w,
+	//
+	//}, nil
```
jeschkies, line 52: Same here 😊

pkg/util/sketch/topk.go
```diff
@@ -0,0 +1,248 @@
...
+		width = 5120
+	case c >= 1000:
+		width = 640
+	case c >= 100:
+		width = 48
+	}
+	return width
+}
+
+func NewCMSTopkForCardinality(l log.Logger, k, c int) (*Topk, error) {
```
jeschkies, line 49: Could you document the parameters?

pkg/util/sketch/topk.go
```diff
@@ -0,0 +1,248 @@
...
+
+	sk, err := NewCMSTopK(k, w, d)
+	if err != nil {
+		return &Topk{}, err
+	}
+	sk.expectedCardinality = c
+	return sk, nil
+}
+
+func NewCMSTopK(k, w, d int) (*Topk, error) {
```
jeschkies, line 63: Same here 😊

pkg/util/sketch/topk.go
```diff
@@ -0,0 +1,248 @@
...
+	if t.InTopk(h1, h2) {
+		return
+	}
+
+	if len(*t.heap) < t.max {
+		t.heapPush(event, estimate, h1, h2)
+		return
+	}
+
+	if estimate > t.heap.Peek().(*node).count {
```
jeschkies, line 127: I'm not sure I understand this logic. Why are we looking at the minimum of the heap?

pkg/util/sketch/topk_test.go
```diff
@@ -0,0 +1,725 @@
+//go:build sketch
```
jeschkies, line 1: I think some Tests soups always run.

pkg/util/sketch/topk_test.go
```diff
@@ -0,0 +1,725 @@
...
+
+	require.LessOrEqualf(t, singleMissing, 2, "more than acceptable misses: %d > %d", singleMissing, 2)
+	// this condition is never actually true
+	//require.LessOrEqualf(t, mergedMissing, singleMissing, "merged sketch should be at least as accurate as a single sketch")
+}
+
+// compare the accuracy of cms topk and hk to the real topk
+func TestRealTopK(t *testing.T) {
+	// the HLL cardinality estimate for this page is ~72000
+	link := "https://www.gutenberg.org/cache/epub/100/pg100.txt"
```
jeschkies, line 349: I think we should check these fixtures in. If the resource is removed it breaks our tests.

pkg/util/sketch/topk_test.go
```diff
@@ -0,0 +1,725 @@
...
+
+	resp, err := http.Get(link)
+	require.NoError(t, err)
+	scanner := bufio.NewScanner(resp.Body)
+
+	m := make(map[string]uint32)
+	h := MinHeap{}
```
jeschkies, line 356: I'm not quite sure why we are not using a hash map instead to get the actual count.

jeschkies (changes requested):
This is really awesome. I think there's not too much to fix to merge it. Since it is its own package it's fine to improve it iteratively.

pkg/util/sketch/cms.go
cstyan: I'm not sure we do, is there a use case you're thinking of? 

The string of labels to identify each thing we're doing a topk on are what we'll want to use to insert things into the sketch. This could be the entire labelset for a stream, or partial, or with additional labels from line parsing. So we can't just use for example the already computed stream fingerprint.

pkg/util/sketch/cms_test.go
cstyan: Yes I'll split things out into separate files.

pkg/util/sketch/cms_test.go
cstyan: The amount of randomized events is already non-deterministic, and randomized # of events plus different sorting is a more realistic representation of data we might have to run a topk query against. We can have a separate test that has a static dataset and sorting.

pkg/util/sketch/hash.go
cstyan: It's an optimization for the hasing requirement of count min sketch, which is that for each row of the sketch (`d` depth), you need a separate hash function. These hash functions need to be pairwise independent, meaning that for each row of the sketch, passing the same input to each rows hash function should yield a different row, meaning that for each row we end up in a different column.

It's linked in our design doc and also in a comment in this file, but [this](http://www.eecs.harvard.edu/~michaelm/postscripts/tr-02-05.pdf) paper outlines that you can achieve the same pairwise independence using only two hash functions and for each row using the two hashes for the `event` and an extra little bit of math (hopefully [this](https://github.com/grafana/loki/pull/9882/files#diff-4df9b704f7d9ac192074c6b300f408db456bc8c96bcb4bef15d1bf058bee8af5R34-R37) link works, otherwise look for the `getPos` function).

The tl; dr is that by using only 2 hash functions rather than `d` hash functions, only hashing twice rather than once for each row of the sketch, we do less work when inserting something into a sketch. The hash computation is the majority of CPU time for inserting something into the sketch, and we just cut that in half if not better.

pkg/util/sketch/heap.go
cstyan: We could but I'd have to look at the ones we have so far to see if we can use it. I don't think it's too important for us to try to dedupe this particular bit of the code.

pkg/util/sketch/heavykeeper.go
cstyan: Not unless we can come up with a better, more accurate merging mechanism.

At the moment our current issue is that merging of both CMS (with conservative updates) and HK result in a loss of accuracy. It's not end of the world bad, we might go from 99/100 topk correct to 90/100, slightly less for HK.

pkg/util/sketch/topk.go
cstyan: I'll add a [link](http://www.eecs.harvard.edu/~michaelm/postscripts/tr-02-05.pdf) to a comment above the `Observe` function to explain the source of the optimization here, but the short(ish) version is this:

By default when we observe an event, if it's already in the current topk we would update it's value in the heap structure with the new count min sketch estimate and then rebalance the heap. This is potentially a lot of heap balancing operations that, at the end of the day, aren't really important. What information do we care about from the heap when we're actually still observing events and tracking the topk? The minimum value that's stored in the heap. If we observe an event and it's new count is greater than the minimum value in the heap, that event should go into the heap and the event with the minimum value should come out.

The optimization here is that we don't need to update the values for each item in the heap each time we observe an event that's in the heap, we only need to update them when we observe an event that's **not** in the heap, and it's new estimate is **greater** than the thing that's the current minimum value heap element. At that point, we update the values for each node in the heap and rebalance the heap, and then if the event we're observing has an estimate that is **still** greater than the minimum heap element count, we should put this event into the heap and remove the other one. This is the optimization that sped up the benchmark by ~3x.

pkg/util/sketch/topk_test.go
cstyan: we could, iirc the combination of the 3 is something like 7-10mb, but these are probably safe to keep imo all public domain books so unlikely to be taken down (similar to if we used links to wikipedia pages imo)

pkg/util/sketch/topk_test.go
cstyan: We are, the map `m` tracks the real counts for things, and the heap `h` tracks the current topk. Same logic as the cms implementation, if we observe an event into the heap and it's new count is greater than the heap min we insert that thing into the heap. This should be faster than doing a sort of a potentially massive cardinality set (think 1 million plus unique streams) at the end of observing all the events.

pkg/util/sketch/cms.go
jeschkies: I think the sample stream would give us the fingerprints but it's not urgent now. Let's roll with strings.

pkg/util/sketch/cms_test.go
jeschkies: Which events do you mean? They are so random but the same for each run.

pkg/util/sketch/hash.go
jeschkies: Thanks for the lengthy reply. What I meant is why is the hash function not imported from a package since the license at the top suggests is third party code.

pkg/util/sketch/heap.go
jeschkies: It's not that much but we would avoid bugs. On the other hand we can change and optimize this code as we see fit for our case.

pkg/util/sketch/heavykeeper.go
jeschkies: Hm. Should we check the code into main then? I'm on the fence

pkg/util/sketch/topk.go
jeschkies: Nice. Maybe we could add that as a comment. I've seen algorithm implementations with lengthy comments. I think it's totally valid to put your explanation into the code base.

pkg/util/sketch/topk_test.go
jeschkies: We could zip them. Even if they are public they could be down for a few hours breaking our CI.

pkg/util/sketch/topk_test.go
jeschkies: Well, you are sorting after all wir haben been scanned or am I missing something?

pkg/util/sketch/cms_test.go
cstyan: no, for each run of this test each stream (by string identifier `0 - numStreams-1` has a random amount of event occurrences between 1 and 100, essentially 10 streams each with 1-100 log line entries

pkg/util/sketch/hash.go
cstyan: we only need the one hashing optimization, the rest of our structures are slightly different and we may want to make modifications to them, such as the underlying type for each counter in a count min sketch

for reference, the hash function is from here: https://github.com/dustin/go-probably/blob/master/count.go#L40-L69

pkg/util/sketch/heap.go
cstyan: the heap is probably a "one and done" implementation anyways, the one I have here is basically straight from the go source examples for using the `container.Heap` package where they implement a priority queue, except reversed slightly so that it's a min heap: https://pkg.go.dev/container/heap#example-package-PriorityQueue

pkg/util/sketch/heavykeeper.go
cstyan: For sure, I'm fine to remove it (and sketchbf), I just wanted everyone whose interested to have the opportunity to see the other structure types.

pkg/util/sketch/topk_test.go
cstyan: I'm sorting the list I generate from the heap, since iterating over the heap elements it won't be in `top100-top1` order, the only guarantee is that the smallest count element will be first.

pkg/util/sketch/topk_test.go
cstyan: yeah that's fair :+1: I'll copy paste them for now

pkg/util/sketch/topk.go
cstyan: I've added a comment to the topk structure as well as above the `Observe` function about the optimizations, let me know if you think they explain it well enough.

pkg/util/sketch/cms_test.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+	}
+	// then another set of things more than the max of the previous entries
+	for z := numStreams - k; z < numStreams; z++ {
+		n := int64(rand.Int63n(int64(maxPerStream)) + 1 + max)
+		for x := 0; x < int(n); x++ {
+			events = append(events, event{name: strconv.Itoa(z), count: 1})
+		}
+		//total += n
+		//oTotal += int(n)
```
jeschkies, line 41: Is this needed?

pkg/util/sketch/topk.go
```diff
@@ -0,0 +1,313 @@
...
+// wrapper to bundle together updating of the bf portion of the sketch and pushing of a new element
+// to the heap
+func (t *Topk) heapPush(h *MinHeap, event string, estimate, h1, h2 uint32) {
+	var pos uint32
+	for i := range t.bf {
+		pos = t.sketch.getPos(h1, h2, i)
+		t.bf[i][pos] = true
+	}
+	heap.Push(h, &node{event: event, count: estimate})
+	// temp for testing
```
jeschkies, line 111: I'm not sure what you mean here.

pkg/util/sketch/topk.go
```diff
@@ -0,0 +1,313 @@
...
+func unsafeGetBytes(s string) []byte {
+	if s == "" {
+		return nil // or []byte{}
+	}
+	return (*[0x7fff0000]byte)(unsafe.Pointer(
+		(*reflect.StringHeader)(unsafe.Pointer(&s)).Data),
+	)[:len(s):len(s)]
+}
+
+// Observe is our sketch event observation function, which is a bit more complex than the original count min sketch + heap TopK
```
jeschkies, line 161: :+1:

pkg/util/sketch/topk.go
```diff
@@ -0,0 +1,313 @@
...
+// We only need to update the count for each event in the heap when we observe an event that's not in the heap, and it's
+// new estimate is greater than the thing that's the current minimum value heap element. At that point, we update the values
+// for each node in the heap and rebalance the heap, and then if the event we're observing has an estimate that is still
+// greater than the minimum heap element count, we should put this event into the heap and remove the other one.
+func (t *Topk) Observe(event string) {
+	estimate, h1, h2 := t.sketch.ConservativeIncrement(event)
+	//t.hll
+	//fmt.Println("hll hash for event: ", uint64(h1)<<32+uint64(h2))
+	//t.hll.InsertHash(uint64(h1<<32 + h2))
+	//t.eventBytes = []byte(event)
```
jeschkies, line 182: This can probably go.

pkg/util/sketch/topk_test.go
```diff
@@ -0,0 +1,369 @@
...
+func TestTopK_Merge(t *testing.T) {
+	nStreams := 1000
+	k := 100
+	maxPerStream := 1000
+	events := make([]event, 0)
+	max := int64(0)
+
+	for i := 0; i < nStreams-k; i++ {
+		num := int64(maxPerStream)
+		n := rand.Int63n(num) + 1
```
jeschkies, line 54: This test is not repeatable and it would be a pain to debug. Please use a rand instance with a constant seed.

jeschkies (changes requested):
Nice work. I have only two blockers:

1. Consider switching to `uint32` for sizes so you don't have to check for negative dimensions.
2. Use constant seeds for the randomizer in tests.

Once that's done I'm fine with merging this :slightly_smiling_face:

cstyan:
@jeschkies lets wait until sometime tomorrow my time before merging, want others to have a chance to take a look if they are interested. I'll also move the code (probably) to a package inside `logql` and remove the unused structures.

jeschkies (approved):
Please see my comment on rand :slightly_smiling_face:

PR #9888, opened 2023-07-07
Query ingesters for SeriesVolume in single binary

trevorwhitney:
This change is necessary to get recent SeriesVolume data from the ingesters when running without an `AsyncStore` in single binary. I'm not sure exactly how it affects SeriesVolume responses when using an `AsynceStore` in distributed mode and am a little worried it might result in double-counting results from ingesters.

To recreate the bug this PR fixes:
1. Remove everything from the local folder you use as `filesystem` backend for local single binary Loki (ie `rm -rf /tmp/loki`)
2. Start Loki in single binary mode with the following config:
```
auth_enabled: false

server:
  http_listen_port: 3100
  grpc_listen_port: 9096

common:
  instance_addr: 127.0.0.1
  path_prefix: /tmp/loki
  storage:
    filesystem:
      chunks_directory: /tmp/loki/chunks
      rules_directory: /tmp/loki/rules
  replication_factor: 1
  ring:
    kvstore:
      store: inmemory

query_range:
  results_cache:
    cache:
      embedded_cache:
        enabled: true
        max_size_mb: 100

schema_config:
  configs:
    - from: 2020-10-24
      store: tsdb
      object_store: filesystem
      schema: v12
      index:
        prefix: index_
        period: 24h

ruler:
  alertmanager_url: http://localhost:9093

limits_config:
  volume_enabled: true
```
3. Push logs to loki and try to query `series_volume`. The volumes will not be available. If you manually `/flush`, then they will be available.


trevorwhitney:
I am unclear if this conflicts with using an `AsyncStore`? Is this something we only want to do when not using an `AsyncStore`, or does it not matter?

trevorwhitney:
So I remember on Monday...

I got this running with a debugger in a scalable (read, write, backend) configuration that was utilizing an AsyncStore and confirmed this does indeed cause the ingesters to get queried twice. As a result I see 2 paths forward:

1) keep this change and remove `SeriesVolume` from `AsyncStore`, forcing the ingester querying to always happen in the querier.
2) remove this change and figure out a way to use the async store in single binary mode.

I'm worried this bug may exist for stats calls as well.

MasslessParticle:
This PR explicitly breaks the query times into two parts `ingester` and `store`. Because the store also calls ingesters, we'd expect ingesters be be called twice but for different time ranges.

MasslessParticle (approved):

PR #9890, opened 2023-07-07
Update redirects to use new alias format. (part 2 of 2)

JStickler:
Breaking the alias updates into two PRs to make for smaller reviews.

@jdbaldry commented on #9717 that the alias format had changed, but we [agreed](https://github.com/grafana/loki/pull/9717#issuecomment-1594960563) that I could update it in a follow up PR. This is that follow up.  Also updating previously merged file moves that used the old alias format.

<strike>I could not get relative redirects to work for the LIDS or Maintaining files, so had to use the older format, which did work.</strike>
Updated weights for some pages.
Updated a couple of `menuTitles` that I missed updating on previous PRs.


docs/sources/community/design-documents/2020-02-Promtail-Push-API.md
```diff
@@ -2,8 +2,8 @@
 title: Promtail Push API
 description: Promtail Push API
 aliases: 
-- /docs/loki/latest/design-documents/
-- /docs/loki/latest/community/design-documents/
+- ../design-documents/2020-02-promtail-push-api/
```
jdbaldry, line 5: The alias in this patch does not redirect me from http://localhost:3002/docs/loki/latest/design-documents/2020-02-promtail-push-api/ to http://localhost:3002/docs/loki/latest/design-documents/2020-02-promtail-push-api/.

Instead I believe it should be:
```suggestion
- ../../design-documents/2020-02-promtail-push-api/
```

docs/sources/community/design-documents/2020-02-Promtail-Push-API.md
```diff
@@ -2,8 +2,8 @@
...
+- ../community/design-documents/2020-02-promtail-push-api/
```
jdbaldry, line 6: This alias redirects http://localhost:3002/docs/loki/latest/community/community/design-documents/2020-02-promtail-push-api/ to this page. I don't believe that is intended behavior

```suggestion
```

docs/sources/community/lids/0001-Introduction.md
```diff
@@ -2,8 +2,8 @@
 title: "0001: Introducing LIDs"
 description: "0001: Introducing LIDs"
 aliases: 
-- /docs/loki/latest/lids
-- /docs/loki/latest/community/lids
+- /docs/loki/latest/lids/0001-introduction/
```
jdbaldry, line 5: We must not have absolute aliases in versioned documentation unless you want to have to update the version every time a new release/version branch is cut. Old release branches that still have "latest" in the aliases steal redirects if the page is moved later. This takes readers to old documentation unexpectedly when they follow indexed search engine results.

The correct alias to redirect http://localhost:3002/docs/loki/latest/lids/0001-introduction/ to this page is:

```suggestion
- ../../lids/0001-introduction/
```

jdbaldry (changes requested):
I think these aliases need to be revisited.

- Absolute aliases must be removed to avoid confusing user behavior in the future.
- All relative aliases need to be checked because I'm not seeing the intended redirect behavior.

Relative aliases that refer to the current page location won't be effective and should be removed because they can cause confusion and might result in mistaken redirects in the future if pages are moved again. In the past, it was considered a convenience to keep the absolute version of the current alias so that you would automatically have a redirect ready when the page moves. However, relative aliases won't work the same way.

docs/sources/community/design-documents/2020-02-Promtail-Push-API.md
JStickler: This is the intended behavior, the design docs have moved under community.

jdbaldry: But the URL in my comment has "community" twice.

docs/sources/community/lids/0001-Introduction.md
JStickler: I had noted in the description that "I could not get relative redirects to work for the LIDS or Maintaining files, so had to use the older format, which did work."  I don't know why it wasn't working last week (I tested repeatedly and couldn't get the redirects to work), because it's working fine now.  ¯\_(ツ)_/¯   Updated the five files that still used absolute references to use the new format.

docs/sources/community/design-documents/2020-02-Promtail-Push-API.md
JStickler: Why would it be doing that?

jdbaldry: If you follow the relative path `../community/design-documents/2020-02-promtail-push-api/` from the page `/docs/loki/latest/community/design-documents/2020-02-promtail-push-api/` you end up at `/docs/loki/latest/community/community/design-documents/2020-02-promtail-push-api/`.


For the page `/docs/loki/latest/community/design-documents/2020-02-promtail-push-api/`:
- `.` is `/docs/loki/latest/community/design-documents/`.
-`..` is `/docs/loki/latest/community/`
- `../community` is `/docs/loki/latest/community/community/design-documents/2020-02-promtail-push-api/`
- `../community/design-documents/` is `/docs/loki/latest/community/community/design-documents/`
- `../community/design-documents/2020-02-promtail-push-api/` is `/docs/loki/latest/community/community/design-documents/2020-02-promtail-push-api/`

jdbaldry: I think the Hugo behavior can be confusing because `.` might not be what you expect it to be.

jdbaldry: I'm sorry, Hugo has a lot of weird edges like this. The most valuable test is using the browser or curl to verify that the redirects are in place. 

It's hard to do a broad PR like this without having seen an example one first. If you'd like, I can open a replacement PR that adds the appropriate aliases and which can serve as a reference for the future?

JStickler: If you do the work for me then I don't learn how to do it for myself.

docs/sources/community/lids/_index.md
```diff
@@ -2,9 +2,8 @@
 title: Loki Improvement Documents (LIDs)
 description: Loki Improvement Documents (LIDs)
 aliases: 
-- /docs/loki/latest/lids
-- /docs/loki/latest/community/lids
-weight: 1400
+- ../../lids/
```
jdbaldry, line 5: I believe the intention is to redirect the page /docs/loki/latest/lids/, so the alias is: 
```suggestion
- ../lids/
```

docs/sources/community/maintaining/_index.md
```diff
@@ -2,9 +2,8 @@
 title: Maintaining
 description: Grafana Loki Maintainers' Guide
 aliases: 
-- /docs/loki/latest/maintaining
-- /docs/loki/latest/community/maintaining
-weight: 1200
+- ../../maintaining/
```
jdbaldry, line 5: I believe the intention is to redirect the page /docs/loki/latest/maintaining/, so the alias is: 
```suggestion
- ../maintaining/
```

docs/sources/reference/api.md
```diff
@@ -3,8 +3,7 @@ title: Grafana Loki HTTP API
 menuTitle: "HTTP API"
 description: Loki exposes REST endpoints for operating on a Loki cluster. This section details the REST endpoints.
 aliases:
-- /docs/loki/latest/api
-- /docs/loki/latest/reference
+- ../api
```
jdbaldry, line 6: nit: Aliases tend to have the trailing slash to match the page URL. It's not strictly necessary, it's only to be more consistent.
```suggestion
- ../api/
```

jdbaldry (commented):
Nice, it's looking a lot better.

I've got a few suggestions and one nit that you can feel free to ignore.

I only concerned myself with the aliases and not any other part of the patch.
I also only concerned myself with the aliases in this patch and not the aliases that have not been modified in this patch, even if the file has been modified in this patch.

PR #9898, opened 2023-07-10
Helm: Fix loki helm chart helper function for loki.host to explicitly include gateway port

MattSurabian:
This PR fixes an issue I ran into while installing enterprise Loki in non-single binary mode via Helm. We override the port value for the gateway here `gateway.service.port` and when the provisioner launches and tries to connect to the gateway it was not aware of the overriden port and instead would try to connect over port 80 which would fail, causing the whole helm install to fail as well.

I didn't do a comprehensive search but a quick search did not reveal an existing issue.

I think single binary mode is fine to assume the port as `3100` because it does not appear that can be overriden. It has been some time since I've used single binary mode though so maybe I'm missing something?

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/CHANGELOG.md
+++ b/production/helm/loki/CHANGELOG.md
@@ -21,7 +21,6 @@ Entries should include a reference to the pull request that introduced the chang
 
 - [CHANGE] Changed version of Loki to 2.8.3
 
-
 ## 5.8.11
 
 - [BUGFIX] Fix gateway: Add `/config` proxy_pass to nginx configuration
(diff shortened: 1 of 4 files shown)


MichelHollands:
@MattSurabian Thanks for this contribution. Can you update the version number? Once that is done I'll merge it. The CI issue seems a transient failure.

MattSurabian:
@MichelHollands Rebased the latest main and updated this branch, cheers!

production/helm/loki/CHANGELOG.md
```diff
@@ -25,6 +29,9 @@ Entries should include a reference to the pull request that introduced the chang
 ## 5.8.10
 
 - [ENHANCEMENT] Canary labelname can now be configured via monitoring.lokiCanary.labelname
+## 5.8.10
```
MichelHollands, line 32: @MattSurabian You left your previous 5.8.10 header in. This is now a duplicate and needs to be removed.

MattSurabian: Sorry about that!

MattSurabian:
Should be good to go now :crossed_fingers:

MichelHollands (approved):

PR #9899, opened 2023-07-10
Add 3rd backend target to docker-compose stack

trevorwhitney:
* Adds 3rd backend target to `production/docker` stack
* Uses a anchor for the image to make local overrides easier
* Updates the config to use TSDB

@dannykopping you seem to be the SME here, mind taking a look?

### DIFF
--- a/production/docker/config/loki.yaml
+++ b/production/docker/config/loki.yaml
@@ -1,7 +1,5 @@
 auth_enabled: true
 
-http_prefix:
-
 server:
   http_listen_address: 0.0.0.0
   grpc_listen_address: 0.0.0.0
@@ -10,6 +8,7 @@ server:
   log_level: info
 
 common:
+  path_prefix: /loki
   storage:
     s3:
       endpoint: minio:9000
@@ -52,12 +51,12 @@ ingester:
 ruler:
   enable_api: true
   wal:
-    dir: /tmp/ruler-wal
+    dir: /loki/ruler-wal
   storage:
     type: local
     local:
       directory: /loki/rules
-  rule_path: /tmp/prom-rules
+  rule_path: /loki/prom-rules
   remote_write:
     enabled: true
     clients:
@@ -71,18 +70,12 @@ ruler:
 schema_config:
   configs:
   - from: 2020-08-01
-    store: boltdb-shipper
+    store: tsdb
     object_store: s3
-    schema: v11
+    schema: v12
     index:
       prefix: index_
       period: 24h
-
-storage_config:
-  boltdb_shipper:
-    shared_store: s3
-    active_index_directory: /tmp/index
-    cache_location: /tmp/boltdb-cache
 
 
 limits_config:
@@ -94,6 +87,7 @@ limits_config:
   ingestion_burst_size_mb: 20
   # parallelize queries in 15min intervals
   split_queries_by_interval: 15m
+  volume_enabled: true
 
 chunk_store_config:
   max_look_back_period: 336h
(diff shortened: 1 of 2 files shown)


production/docker/config/loki.yaml
```diff
@@ -71,19 +70,13 @@ ruler:
 schema_config:
   configs:
   - from: 2020-08-01
-    store: boltdb-shipper
```
dannykopping, line 74: I don't think we can change this in this way; folks may already be using this compose setup, and then their indexes won't work. We'll need to add a new period config from a new date.

production/docker/docker-compose.yaml
```diff
@@ -3,20 +3,18 @@ version: "3.8"
...
+  loki:
 
 services:
-
   # Since the Loki containers are running as user 10001 and the mounted data volume is owned by root,
   # Loki would not have permissions to create the directories.
   # Therefore the init container changes permissions of the mounted directory.
   init:
-    image: grafana/loki:2.8.2
+    image: &lokiImage grafana/loki:2.8.2
```
dannykopping, line 17: Nice!

production/docker/docker-compose.yaml
```diff
@@ -115,19 +113,19 @@ services:
...
       - "3100"
       - "7946"
       # uncomment to use interactive debugging
-      # - "40000-40002:40000" # # makes the replicas available on ports 40000, 40001, 40002
-      #cap_add:
-      #  - SYS_PTRACE
-      #security_opt:
-      #  - apparmor=unconfined
+      # - "40000-40002:40000" # makes the replicas available on ports 40000, 40001, 40002
+    # cap_add:
```
dannykopping, line 125: Weird indentation here

production/docker/docker-compose.yaml
```diff
@@ -138,18 +136,18 @@ services:
...
       - "7946"
       # uncomment to use interactive debugging
-      # - "50000-50002:40000" # makes the replicas available on ports 50000, 50001, 50002
-      # cap_add:
-      #   - SYS_PTRACE
-      # security_opt:
-      #   - apparmor=unconfined
+      # - "50000-50002:40000" #  makes the replicas available on ports 50000, 50001, 50002
+    # cap_add:
+    # - SYS_PTRACE
```
dannykopping, line 148: Ditto

production/docker/config/loki.yaml
trevorwhitney: I can add as a new config, but honestly nothing we provide is going to be guaranteed to work out of the box. If I provide a date the in the future, it's going to require the stack to be running when the changeover happens. If I provided a date in the past, then chunks/indexes already written past that time will no longer work.

Maybe the safest thing to do is add with the approximate date of this PR being merged, but it's still likely to break anyone using this who tries to use our config out of the box.

dannykopping: > but it's still likely to break anyone using this who tries to use our config out of the box

I don't understand why that would be the case?

dannykopping: Re-reading your comment, I think you're right actually @trevorwhitney.
I still think we should add another period config, and then folks will just be able to change the date to get their storage config working again; replacing the store config will make that harder.

production/docker/docker-compose.yaml
trevorwhitney: yeah, it's messy because `cap_add` is on the same level as `ports`. I reorganized the file so the indentation is cleaner/more obvious.

production/docker/config/loki.yaml
trevorwhitney: sounds good, created a new schema starting on `2023-07-11`.

dannykopping (approved):
LGTM!

Issue #8710, opened 2023-03-06
[Epic] Create consistent information architecture across database products


JStickler:
The doc squads for Loki, Tempo, and Mimir have been meeting regularly for two months to create a common information architecture for the three projects.  This will make it easier for users to find the information that they’re looking for across the stack.

References
- (Internal) https://github.com/grafana/technical-documentation/issues/538
- Mimir Epic https://github.com/grafana/mimir/issues/4274 
- Tempo Epic https://github.com/grafana/tempo/issues/2223

```[tasklist]
### Tasks
- [ ] https://github.com/grafana/loki/issues/8711
- [ ] https://github.com/grafana/loki/issues/8712
- [ ] https://github.com/grafana/loki/issues/8713
- [ ] https://github.com/grafana/loki/issues/8714
- [ ] https://github.com/grafana/loki/issues/8715
- [ ] https://github.com/grafana/loki/issues/8716
- [ ] https://github.com/grafana/loki/issues/8717
- [ ] https://github.com/grafana/loki/issues/8718
- [ ] https://github.com/grafana/loki/issues/8719
- [ ] https://github.com/grafana/loki/issues/8720
- [ ] https://github.com/grafana/loki/issues/9226
- [ ] https://github.com/grafana/loki/issues/10152
```
Note that Phase One involves moving entire topics.
Phase Two (#8721) will involve refactoring and moving content within files.

marinnedea:
Suggestion as topic or maybe something to include in every doc: "On premises vs Grafana Cloud differences", if any.

knylander-grafana:
Looks like you've completed the work for this issue. Might be a good time to close it.

JStickler:
@knylander-grafana Thanks, I thought I already had closed it!

PR #9910, opened 2023-07-11
Fix image links in moved files.

JStickler:
Noticed the other day that images in files that have been moved as part of #8710 were not displaying in `/next/`.  This PR updates the image references.


jdbaldry (approved):
Can confirm I see no broken images in the pages touched by this PR.

I haven't checked any other pages to see if there are other broken images.

Note, at some point, we should move the images to the media storage bucket following the guidelines in https://grafana.com/docs/writers-toolkit/write/image-guidelines/.

PR #9911, opened 2023-07-11
Remove aliases to current location.

JStickler:
Re: [this comment ](https://github.com/grafana/loki/pull/9890#pullrequestreview-1521562982)from @jdbaldry  on #9890 that "Relative aliases that refer to the current page location won't be effective and should be removed because they can cause confusion and might result in mistaken redirects in the future if pages are moved again."

Removing aliases that refer to the current location of the pages.

### DIFF
--- a/docs/sources/alert/_index.md
+++ b/docs/sources/alert/_index.md
@@ -4,8 +4,6 @@ menuTitle: Alert
 description: Learn how the rule evaluates queries for alerting.
 aliases:
   - ./rules/
-  - ./alert/
-  - ./alerting/
 weight: 850
 keywords:
   - loki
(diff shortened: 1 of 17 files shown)


docs/sources/alert/_index.md
```diff
@@ -4,8 +4,6 @@ menuTitle: Alert
 description: Learn how the rule evaluates queries for alerting.
 aliases:
   - ./rules/
-  - ./alert/
-  - ./alerting/
```
jdbaldry, line 8: This could be a potentially valid alias. It would redirect /docs/loki/latest/alerting/ to this page.

jdbaldry (approved):
I think this patch looks good to me. I have added a note about one specific alias that might need keeping.

Issue #8711, opened 2023-03-06
[Docs] Create top-level Get started section


JStickler:
Move the following files under Getting started:
./sources/getting-started/_index.md	
./sources/fundamentals/_index.md	
./sources/fundamentals/architecture/_index.md	
./sources/fundamentals/architecture/components/_index.md	
./sources/fundamentals/architecture/deployment-modes/_index.md	
./sources/fundamentals/architecture/rings/_index.md	
./sources/fundamentals/labels.md	
./sources/fundamentals/overview/_index.md

PR #9912, opened 2023-07-11
Break up and move Best Practices topic.

JStickler:
I realized that I should have moved the Best Practices topic as part of #8711, since most of the content is best practices for using labels.
This PR:
* Splits the file into two parts (Best practices for labels and best practices for configuration)
* Moves the two files to Get Started and Configure, respectively.
* Updates page weights for both sections as needed.


docs/sources/configure/bp-configure.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Configuration best practices
+menuTitle:  Best practices
+description: Grafana Loki configuration best practices
+weight:  100
+---
+# Configuration best practices
+
+Grafana Loki is under active development, and we are constantly working to improve performance. But here are some of the most current best practices for configuration that will give you the best experience with Loki.
```
trevorwhitney, line 9: should we qualify "current" with a version?

docs/sources/configure/bp-configure.md
```diff
@@ -0,0 +1,71 @@
...
+{job="syslog", instance="host2"} 00:00:02 i'm a syslog!  <- Accepted, still in order for stream 2
+```
+
+But what if the application itself generated logs that were out of order? Well, I'm afraid this is a problem. If you are extracting the timestamp from the log line with something like [the Promtail pipeline stage](/docs/loki/latest/clients/promtail/stages/timestamp/), you could instead _not_ do this and let Promtail assign a timestamp to the log lines. Or you can hopefully fix it in the application itself.
+
+It's also worth noting that the batching nature of the Loki push API can lead to some instances of out of order errors being received which are really false positives. (Perhaps a batch partially succeeded and was present; or anything that previously succeeded would return an out of order entry; or anything new would be accepted.)
+
+## Use `chunk_target_size`
+
+Using `chunk_target_size` instructs Loki to try to fill all chunks to a target _compressed_ size of 1.5MB. These larger chunks are more efficient for Loki to process.
```
trevorwhitney, line 55: the 1.5MB is based on what the size is configured to. This value is in bytes.

Are we just saying 1.5MB is the default?

docs/sources/get-started/bp-labels.md
```diff
@@ -1,9 +1,12 @@
...
 description: Grafana Loki label best practices
-weight: 400
+aliases: 
+- ../best-practices/
+weight: 700
 ---
-# Best practices
+# Label best practices
 
 Grafana Loki is under active development, and we are constantly working to improve performance. But here are some of the most current best practices for labels that will give you the best experience with Loki.
```
trevorwhitney, line 11: this paragraph is repeated verbatim in the added file above

trevorwhitney (approved):
approving to unblock, but I have a few bits of feedback. thanks!

MichelHollands (approved):

docs/sources/configure/bp-configure.md
JStickler: Do we have any idea what version of Loki these best practices were written against?

docs/sources/get-started/bp-labels.md
JStickler: That's by design.

docs/sources/configure/bp-configure.md
JStickler: ¯\_(ツ)_/¯  I don't know, I didn't write this content, I'm just moving it around in this PR.

PR #9934, opened 2023-07-13
add sharded to querystats bytes processed

trevorwhitney:
This adds a shards label to the `loki_logql_querystats_bytes_processed_per_second` metric that we use for determining our `LokiQueryRangeReadPerformanceBudgetBurn` metric so we can exclude slow queries with 0 shards from burning our SLO budget.

WDYT?


chaudum:
> WDYT?

What is the potential cardinality of the `shards` label?

trevorwhitney:
@chaudum that's a good point, it's bounded by the number of shards which could be in the thousands. What if we converted this to a boolean, as we really only care about `0 shards` or `> 0 shards`?

trevorwhitney:
I'm thinking `shardable=[true|false]`

pkg/logql/metrics.go
```diff
@@ -260,7 +264,8 @@ func RecordSeriesQueryMetrics(
 		"total_entries", stats.Summary.TotalEntriesReturned,
 	)
 
-	bytesPerSecond.WithLabelValues(status, queryType, "", latencyType).
```
DylanGuedes, line 263: doesn't that mean that this metric can have infinite cardinality?

DylanGuedes (changes requested):

owen-d:
Make it `sharded=true|false` because shardable queries that touch less than 1 shard's maximum size will be executed unsharded.

trevorwhitney:
I have changed the label to a boolean, with the logic being 0 or 1 shards is `sharded=false`, while > 1 shards is `sharded=true`. There are a few places in the code we check if shards are 0 and if so set them to 1, so it seemed like we'd what to capture both cases for "non-sharded".

pkg/logql/metrics.go
```diff
@@ -133,7 +134,7 @@ func RecordRangeAndInstantQueryMetrics(
 		"store_chunks_download_time", stats.ChunksDownloadTime(),
 		"queue_time", logql_stats.ConvertSecondsToNanoseconds(stats.Summary.QueueTime),
 		"splits", stats.Summary.Splits,
-		"shards", stats.Summary.Shards,
+		"sharded", stats.Summary.Shards,
```
DylanGuedes, line 137:
```suggestion
		"shards", stats.Summary.Shards,
```

DylanGuedes (changes requested):

pkg/logql/metrics.go
trevorwhitney: sloppy rename... thank you!

DylanGuedes (approved):

PR #9935, opened 2023-07-13
Call FetchChunks in parallel from series index store

trevorwhitney:
This PR changes the invocation of `FetchChunks` in the series index store to happen in parallel. It seems like this was the original intention due to the `// bound concurrency` comment when allocating the `groups` slice. This came out of an investigation of some slow queries, and the traces showed many serial calls to `FetchChunks`.

image (1)

### DIFF
--- a/pkg/storage/stores/series/series_index_store.go
+++ b/pkg/storage/stores/series/series_index_store.go
@@ -244,10 +244,10 @@ func (c *indexReaderWriter) chunksToSeries(ctx context.Context, in []logproto.Ch
 	}
 
 	type f func() ([]labels.Labels, error)
-	var jobs []f
+	jobs := make([]f, 0, len(groups))
 
 	for _, g := range groups {
-    group := g
+		group := g
 		jobs = append(jobs, f(func() ([]labels.Labels, error) {
 			sort.Sort(group)
 			chunks, err := c.fetcher.FetchChunks(ctx, group.chunks, group.keys)
@@ -255,7 +255,7 @@ func (c *indexReaderWriter) chunksToSeries(ctx context.Context, in []logproto.Ch
 				return nil, err
 			}
 
-		results := make([]labels.Labels, 0, len(chunks))
+			lbls := make([]labels.Labels, 0, len(chunks))
 		outer:
 			for _, chk := range chunks {
 				for _, matcher := range matchers {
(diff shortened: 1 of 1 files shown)


pkg/storage/stores/series/series_index_store.go
```diff
@@ -244,31 +243,57 @@ func (c *indexReaderWriter) chunksToSeries(ctx context.Context, in []logproto.Ch
...
+	for _, g := range groups {
+    group := g
+		jobs = append(jobs, f(func() ([]labels.Labels, error) {
+			sort.Sort(group)
+			chunks, err := c.fetcher.FetchChunks(ctx, group.chunks, group.keys)
+			if err != nil {
+				return nil, err
+			}
+
+		results := make([]labels.Labels, 0, len(chunks))
```
DylanGuedes, line 258: you're defining results here and at line 281

pkg/storage/stores/series/series_index_store.go
```diff
@@ -244,31 +243,57 @@ func (c *indexReaderWriter) chunksToSeries(ctx context.Context, in []logproto.Ch
...
-			results = append(results, labels.NewBuilder(chk.Metric).Del(labels.MetricName).Labels())
-		}
+			return results, nil
+		}))
+	}
+
+	results := make([]labels.Labels, 0, len(chunksBySeries))
+	if err := concurrency.ForEachJob(
+		ctx,
+		len(jobs),
```
DylanGuedes, line 284: wdyt of putting a upper bound to the concurrency number? like: min(len(jobs), 20)

DylanGuedes (changes requested):

pkg/storage/stores/series/series_index_store.go
trevorwhitney: it might help to rename one for clarity nevertheless, will do!

pkg/storage/stores/series/series_index_store.go
```diff
@@ -244,31 +243,57 @@ func (c *indexReaderWriter) chunksToSeries(ctx context.Context, in []logproto.Ch
...
 	}
 
-	for _, group := range groups {
-		sort.Sort(group)
-		chunks, err := c.fetcher.FetchChunks(ctx, group.chunks, group.keys)
-		if err != nil {
-			return nil, err
-		}
+	type f func() ([]labels.Labels, error)
+	var jobs []f
```
MasslessParticle, line 247: You could preallocate this to avoid future allocs

DylanGuedes (approved):

pkg/storage/stores/series/series_index_store.go
```diff
@@ -244,31 +244,64 @@ func (c *indexReaderWriter) chunksToSeries(ctx context.Context, in []logproto.Ch
...
 	}
+
+	if err := concurrency.ForEachJob(
+		ctx,
+		len(jobs),
+		numConcurrent,
+		func(_ context.Context, idx int) error {
+			res, err := jobs[idx]()
+			if res != nil {
+				results = append(results, res...)
```
jeschkies, line 297: Is append thread safe?

PR #9936, opened 2023-07-13
correctly handle unpack when reordering stages

MasslessParticle:
The optimization to reorder stages doesn't take into account the `unpack` stage. When line filters come after unpack stages, they shouldn't be moved.


chaudum (commented):
Generally LGTM. Could you please add a changelog entry for the fix?

MasslessParticle:
@chaudum This didn't make it into `2.8` so no need to backport

trevorwhitney (approved):
lgtm

owen-d (approved):

PR #9942, opened 2023-07-14
operator: Use a condition to warn when there are no nodes with matching labels for zone-awareness

btaani:
When zone-awareness replication is enabled, the LokiStack goes into a custom degraded status when the pods are in pending state due to node's labels not matching the topologyKey

Fixes  [LOG-4075](https://issues.redhat.com/browse/LOG-4075)

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/controllers/loki/lokistack_controller.go
```diff
@@ -329,3 +333,58 @@ func (r *LokiStackReconciler) enqueueForStorageSecret() handler.EventHandler {
...
+
+				if nodeLabels != labelsAnnotation {
+					err := kverrors.New("Topology key pod annotation does not match its node's labels")
+					r.Log.Error(err, "Topology key pod annotation does not match its node's labels")
+					return nil
+					// still don't know how to handle it from here
+
+					/*return status.DegradedError{
+						Reason: lokiv1.ReasonZoneLabelsMismatch,
+					}*/
```
periklis, line 378: This function should be limited only for enqueueing the LokiStack resources when it detects label changes. A secondary helper the internal helpers should re-apply the check and then update the degraded status.

btaani: I see. This part was a bit confusing to me. thanks!

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -279,6 +280,42 @@ func CreateOrUpdateLokiStack(
...
+				"app.kubernetes.io/instance": stack.Name,
+			}),
+		}); err != nil {
+			ll.Error(err, "Error getting pod resources")
+			return nil
+		}
+
+		for _, pod := range podList.Items {
+			assignedNode := &corev1.Node{}
+			key := client.ObjectKey{Name: pod.Spec.NodeName}
```
xperimental, line 297: I had a question about how Kubernetes treats the `topologySpreadConstraints` myself yesterday and tested out the error condition that I see the condition as a "debug information" for: picking a label for the `topologyKey` that does not actually exist on any Node.

What I wanted to find out was: does Kubernetes automatically modify the scheduling so that it only schedules on Nodes that have the required label(s). And the answer is: yes, it does.

While this is good news for our operator in general, as we do not need to add another `nodeAffinity` or something like that it means that this code will never be triggered, because the Pods will actually never be `scheduled` to a Node which doesn't have the matching labels. This means that `pod.Spec.NodeName` will be empty and the Pod will perpetually be in the `Pending` state.

xperimental (commented):
Thought about this story a bit today. I think my suggestion would be to add a check to the code that currently creates the `.status` instead:

- Check if there are Pods in `Pending` state
- If there are pending Pods, check if there's at least _one Node_ that has all the required labels mentioned as `topologyKey`
- If there is no Node matching the labels, add a degraded condition. (Do we have "warning conditions" yet or something like that?)

This should make this a lot more bearable for LokiStacks that are correctly configured and this should also mean that we do not need any additional `Watches` calls, as the updates should already be triggered by the watch on `Pod`s.

operator/internal/handlers/lokistack_create_or_update.go
btaani: > If there are pending Pods, check if there's at least one Node that has all the required labels mentioned as topologyKey

If there are nodes that have matching labels, other pods will have already been scheduled to them until `maxSkew` is reached between the nodes.
My understanding is that a pending pod (reason: unschedulable) means that it cannot be scheduled even when there are nodes that have the topologyKey label

operator/controllers/loki/lokistack_controller.go
```diff
@@ -328,3 +330,62 @@ func (r *LokiStackReconciler) enqueueForStorageSecret() handler.EventHandler {
...
+						r.Log.Error(err, "failed to get pod availability zone from assigned node", "name", pod.Name)
+						return nil
+					}
+					if availabilityZone != pod.Annotations[corev1.LabelTopologyZone] {
+						err := kverrors.New("Topology key pod annotation does not match its node's labels")
+						r.Log.Error(err, "Topology key pod annotation does not match its node's labels")
+
+						if err = handlers.AnnotatePodWithAvailabilityZone(ctx, r.Log, r.Client, &pod); err != nil {
+							r.Log.Error(err, "Error annotating pod with availability zone", "pod", availabilityZone)
+						}
```
periklis, line 376: Why do we patch here the node changes into the Pod inside a watch handler? These handlers are not meant to change anything. They simply filter the reconciliation events.

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -279,6 +280,42 @@ func CreateOrUpdateLokiStack(
...
+
+			if availabilityZone != pod.Annotations[corev1.LabelTopologyZone] {
+				return &status.DegradedError{
+					Message: "Availability zone pod annotation does not match it's node's labels",
+					Reason:  lokiv1.ReasonAvailabilityZoneLabelsMismatch,
+					Requeue: false,
+				}
+			}
+		}
+	}
```
periklis, line 317: Can we make this whole block a function instead of adding it inline here. This handler is already bloated.

periklis (changes requested):

operator/controllers/loki/lokistack_controller.go
btaani: Yeah this was a mistake, will fix it right away

operator/controllers/loki/lokistack_controller.go
```diff
@@ -328,3 +330,56 @@ func (r *LokiStackReconciler) enqueueForStorageSecret() handler.EventHandler {
...
+							NamespacedName: types.NamespacedName{
+								Namespace: stack.Namespace,
+								Name:      stack.Name,
+							},
+						})
+						r.Log.Info("Enqueued requests for LokiStack because of Node labels change", "LokiStack", stack.Name, "Node", obj.GetName())
+						return requests
+					}
+				}
+			}
```
periklis, line 381: nit. You can always save levels of nesting by inverting the checks:

```suggestion
			if stack.Spec.Replication == nil || len(stack.Spec.Replication.Zones)  == 0 {
				// Skip this LokiStack as no replication is enabled or none zones are provided
				continue
			}
			podList := &corev1.PodList{}
			if err := r.Client.List(ctx, podList, &client.ListOptions{
				Namespace: stack.Namespace,
				LabelSelector: labels.SelectorFromSet(labels.Set{
					"app.kubernetes.io/instance": stack.Name,
				}),
			}); err != nil {
				r.Log.Error(err, "Error getting pod resources in event handler")
				return nil
			}

			for _, pod := range podList.Items {
				assignedNode := &corev1.Node{}
				key := client.ObjectKey{Name: pod.Spec.NodeName}
				if err := r.Client.Get(ctx, key, assignedNode); err != nil {
					r.Log.Error(err, "Error getting node resources in event handler")
					return nil
				}

				availabilityZone, err := handlers.GetAvailabilityZone([]string{corev1.LabelTopologyZone}, assignedNode.Labels)
				if err != nil {
					r.Log.Error(err, "failed to get pod availability zone from assigned node", "name", pod.Name)
					return nil
				}
				if availabilityZone != pod.Annotations[corev1.LabelTopologyZone] {
					requests = append(requests, reconcile.Request{
						NamespacedName: types.NamespacedName{
							Namespace: stack.Namespace,
							Name:      stack.Name,
						},
					})
					r.Log.Info("Enqueued requests for LokiStack because of Node labels change", "LokiStack", stack.Name, "Node", obj.GetName())
					return requests
				}
			}
```

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -440,3 +448,40 @@ func isNamespaceScoped(obj client.Object) bool {
 		return true
 	}
 }
+
+func checkNodeLabelChange(ll logr.Logger, k client.Client, ctx context.Context, stack lokiv1.LokiStack) error {
```
periklis, line 452: Function name should somehow highlight the degraded error reason.

periklis (changes requested):

operator/internal/status/lokistack.go
```diff
@@ -87,6 +93,45 @@ func generateCondition(cs *lokiv1.LokiStackComponentStatus) metav1.Condition {
...
+				return conditionPending
+			}
+
+			for _, pod := range podList.Items {
+				if pod.Status.Phase == corev1.PodPending {
+					nodeList := corev1.NodeList{}
+					if err := k.List(ctx, &nodeList, &client.ListOptions{
+						LabelSelector: labels.SelectorFromSet(labels.Set{
+							lokiv1.AnnotationAvailabilityZoneLabels: pod.Annotations[lokiv1.AnnotationAvailabilityZoneLabels],
+						}),
```
xperimental, line 115: The Pod's annotation can not be used as a selector for Node labels. You don't actually need the Pod at all, as all information is already contained in the LokiStack in the form of the list of `topologyKey`s. These are actually node labels, so the Node list can be filtered using them.

operator/internal/status/lokistack.go
```diff
@@ -87,6 +93,45 @@ func generateCondition(cs *lokiv1.LokiStackComponentStatus) metav1.Condition {
...
+						return conditionPending
+					}
+
+					if len(nodeList.Items) == 0 {
+						degradedError := DegradedError{
+							Message: "Availability zone pod annotation does not match it's node's labels",
+							Reason:  lokiv1.ReasonAvailabilityZoneLabelsMismatch,
+							Requeue: false,
+						}
+						if err := SetDegradedCondition(ctx, k, req, degradedError.Message, degradedError.Reason); err != nil {
```
xperimental, line 127: No need to call to `SetDegradedCondition` as we're already at the point of "setting conditions" here. Simply return the degraded `Condition` (not `DegradedError`) instead of the pending condition.

operator/internal/handlers/lokistack_create_or_update_test.go
```diff
@@ -1609,3 +1609,115 @@ func TestCreateOrUpdateLokiStack_RemovesRulerResourcesWhenDisabled(t *testing.T)
 	// make sure delete was called twice (delete rules configmap and ruler statefulset)
 	require.Equal(t, 2, k.DeleteCallCount())
 }
+
+func TestCreateOrUpdateLokiStack_ZoneAwareNodeLabelMismatch(t *testing.T) {
```
xperimental, line 1613: Needs to be removed.

operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -279,6 +280,13 @@ func CreateOrUpdateLokiStack(
 		}
 	}
 
+	if stack.Spec.Replication != nil && len(stack.Spec.Replication.Zones) > 0 {
+		err = setDegradedConditionOnNodeLabelMismatch(ll, k, ctx, stack)
+		if err != nil {
+			return err
+		}
+	}
```
xperimental, line 288: Needs to be removed.

xperimental (commented):
It's almost there. As you're leaving for vacation in a short time, I have taken the liberty of preparing [a branch](https://github.com/btaani/loki/compare/log4075...xperimental:loki:log4075) with most of the changes already.

xperimental (approved):
LGTM :slightly_smiling_face: 

@periklis As I have done the last few changes, maybe have another look as well.

operator/internal/status/status_test.go
```diff
@@ -9,9 +9,11 @@ import (
 	"github.com/grafana/loki/operator/internal/manifests"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
periklis, line 12: Duplicate import

periklis (commented):
One minor nit

operator/internal/status/status_test.go
xperimental: Removed :heavy_check_mark:

periklis (approved):
One minor bit

PR #9949, opened 2023-07-14
reset pipeline caches when tailing rather than making a new pipeline

MasslessParticle:
Recent query optimizations made pipeline creation more expensive with the assumption that it occurred relatively infrequently. To avoid growing pipline caches without bound, integers recreate pipelines on every push. The combination of these things means that tailing high-volume streams can cause ingester failure.

This fix adds the ability to clear pipeline caches so we don't need to create a new one each time.

### DIFF
--- a/pkg/logql/log/labels.go
+++ b/pkg/logql/log/labels.go
@@ -136,7 +136,7 @@ func (b *BaseLabelsBuilder) ForLabels(lbs labels.Labels, hash uint64) *LabelsBui
 }
 
 // Reset clears all current state for the builder.
-func (b *LabelsBuilder) Reset() {
+func (b *BaseLabelsBuilder) Reset() {
 	b.del = b.del[:0]
 	b.add = b.add[:0]
 	b.err = ""
--- a/pkg/logql/log/pipeline_test.go
+++ b/pkg/logql/log/pipeline_test.go
@@ -12,15 +12,20 @@ import (
 
 func TestNoopPipeline(t *testing.T) {
 	lbs := labels.FromStrings("foo", "bar")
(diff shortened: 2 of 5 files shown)


pkg/logql/log/labels.go
```diff
@@ -136,7 +136,7 @@ func (b *BaseLabelsBuilder) ForLabels(lbs labels.Labels, hash uint64) *LabelsBui
 }
 
 // Reset clears all current state for the builder.
-func (b *LabelsBuilder) Reset() {
+func (b *BaseLabelsBuilder) Reset() {
```
MasslessParticle, line 139: LabelBuilders are BaseLabelBuilders so this change makes `Reset` available to `Pipeline` without affecting it's visibility in `StreamPipeline`

pkg/logql/log/pipeline_test.go
```diff
@@ -12,23 +12,29 @@ import (
...
 	require.Equal(t, true, matches)
 
-	ls, lbr, matches := NewNoopPipeline().ForStream(lbs).ProcessString(0, "")
+	ls, lbr, matches := pipeline.ForStream(lbs).ProcessString(0, "")
 	require.Equal(t, "", ls)
 	require.Equal(t, NewLabelsResult(lbs, lbs.Hash()), lbr)
 	require.Equal(t, true, matches)
+
+	pipeline.Reset()
+	require.Len(t, pipeline.cache, 0)
```
trevorwhitney, line 28: I wish there was a way to better test the behavior rather than reaching in to private data, but I can't think of one, so I'll just whine instead.

trevorwhitney (approved):
looks good!

owen-d (approved):

PR #9957, opened 2023-07-17
WIP: Evaluate topk aggregations with CMS.

jeschkies:
The querier should evaluate topk vector aggregations with the new sketch data structure. This will enable sharding eventually.

This is an exploration of the interfaces that need to be changed. I would like to reuse the old engine for everything but the topk aggregation. However, the step evaluation return type would change from `promql.Vector` to something that can handle `sketch.TopKVector`.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/logql/evaluator.go
```diff
@@ -124,20 +125,20 @@ func Sortable(q Params) (bool, error) {
 }
 
 // Evaluator is an interface for iterating over data at different nodes in the AST
-type Evaluator interface {
-	SampleEvaluator
+type Evaluator[V any] interface {
+	SampleEvaluator[V]
```
jeschkies, line 129: I've made this change originally to easily switch between the return types. However, it seems that it should always be `promql_parser.Value` or rather our own interface.

pkg/logql/evaluator.go
```diff
@@ -626,26 +635,29 @@ func binOpStepEvaluator(
...
-	return newStepEvaluator(func() (bool, int64, promql.Vector) {
+	return NewStepEvaluator(func() (bool, int64, StepResult) {
 		var (
 			ts       int64
 			next     bool
 			lhs, rhs promql.Vector
+			r        StepResult
 		)
-		next, ts, rhs = rse.Next()
+		next, ts, r = rse.Next()
```
cstyan, line 645: everywhere we call `Next()` can we give the result a better name than `r`, took me a minute to track down what `r` actually was

pkg/logql/probabilistic_evaluator.go
```diff
@@ -0,0 +1,325 @@
...
+	switch e := expr.(type) {
+	case *syntax.VectorAggregationExpr:
+		if e.Operation != syntax.OpTypeTopK {
+			return p.DefaultEvaluator.StepEvaluator(ctx, nextEv, expr, q)
+		}
+		return p.newProbabilisticVectorAggEvaluator(ctx, nextEv, e, q)
+	default:
+		return p.DefaultEvaluator.StepEvaluator(ctx, nextEv, expr, q)
+	}
+}
```
cstyan, line 54: Would we eventually want to have a single engine that just executes a query in the way we implement it to, instead of two separate engines? 

It feels a bit odd to me to be doing the check here in the creation of the step evaluator rather than higher up somewhere we would decide to pass a query to one of the two engines we have in this PR. I think we'd also want a type that contains a list of `probabilisticQueries` similar to how we have `shardableQueries` at the moment. Then the list of things that are (or can be) executed in a probabilistic manner is centralized in one place, and we just check `syntax.<expression type>` against that list.

jeschkies: `StepEvaluator` is called recursively. It's a little hidden but in the original engine 

`DefaultEvaluator.StepEvaluator` calls `vectorAggEvaluator(ctx, nextEv, e, q)` which then calls `ev.StepEvaluator(ctx, ev, expr.Left, q)`. Where `ev` is `nextEv` which is `DefaultEvaluator`.

So we have to slip in the probabilistic evaluator and then switch back to the original. However, I agree. This should happen outside of the evaluators. Ideally the evaluators would not construct the step evaluators but something out side that builds up the chain of evaluators.

pkg/logql/probabilistic_evaluator.go
```diff
@@ -0,0 +1,444 @@
...
+// NewDefaultEvaluator constructs a DefaultEvaluator
+func NewProbabilisticEvaluator(querier Querier, maxLookBackPeriod time.Duration) Evaluator {
+	d := NewDefaultEvaluator(querier, maxLookBackPeriod)
+	p := &ProbabilisticEvaluator{Evaluator: d}
+	return p
+}
+
+type StepEvaluatorAdapter struct {
+	d StepEvaluator
+}
```
cstyan, line 138: did I miss something, seems like we're only using this type in this package?

pkg/logql/downstream.go
```diff
@@ -359,6 +361,42 @@ func (ev *DownstreamEvaluator) StepEvaluator(
...
+				Params: params,
+			}
+			if shard := cur.DownstreamTopkSampleExpr.shard; shard != nil {
+				qry.Shards = Shards{*shard}
+			}
+			queries = append(queries, qry)
+			cur = cur.next
+		}
+
+		results, err := ev.Downstream(ctx, queries)
```
cstyan, line 378: we should include "make downstream responses parsable before all reponses are returned" an item in our engine improvements list, since topk sketches should be mergable in any order we don't need to wait for all results here, we could have `DownstreamAsync` that takes a channel that results can be sent to?

pkg/logql/downstream.go
```diff
@@ -466,6 +504,61 @@ func ConcatEvaluator(evaluators []StepEvaluator) (StepEvaluator, error) {
...
+					acc = cur.TopkVector().Topk
+				} else {
+					acc.Merge(cur.TopkVector().Topk)
+				}
+			}
+
+			for _, r := range acc.Topk() {
+				for _, e := range r.Result {
+					// TODO(karsten): handle error.
+					ls, _ := syntax.ParseLabels(e.Event)
```
cstyan, line 524: at this point the labels would be labels that have already been stored in loki or parsed from log lines, so they should be okay? we could treat an error here as non-fatal imo, be more strict about what's acceptable to pass to `topk.Observe` in the first place?

pkg/logql/sketch/topk.go
```diff
@@ -418,17 +418,17 @@ func (t *Topk) TopkForGroupingKey(groupingKey string) TopKResult {
 	}
 	res := TopKResult{
 		groupingKey: "",
-		result:      make([]element, 0, len(*t.heaps[groupingKey])),
+		Result:      make([]element, 0, len(*t.heaps[groupingKey])),
```
cstyan: should both of these be exported?

pkg/logql/downstream.go
```diff
@@ -584,6 +583,37 @@ func ResultStepEvaluator(res logqlmodel.Result, params Params) (StepEvaluator, e
...
+	switch data := res.Data.(type) {
+	case promql.Vector:
+		var exhausted bool
+		return NewProbabilisticStepEvaluator(func() (bool, int64, StepResult) {
+			if !exhausted {
+				exhausted = true
+				return true, start.UnixNano() / int64(time.Millisecond), SampleVector(data)
+			}
+			return false, 0, nil
+		}, nil, nil)
```
cstyan, line 603: how come we're supporting promql vector?

pkg/logql/probabilistic_evaluator.go
jeschkies: I've used it original for the merging. Will revert it now.

pkg/logql/downstream.go
jeschkies: This is future proof instead we are sending other queries to the probabilistic endpoint. I'm happy to remove this again.

pkg/logql/downstream.go
jeschkies: It's tricky. `topk.Observe` could use the label hash to begin with and we could store the labels differently so there's no need for parsing. However, that's quite some work. We have to handle the error somehow so that we catch bugs.

cstyan:
Tested again locally with the removal of all the excess parameter parsing, still seems to work as intended :+1:

pkg/logql/probabilistic_evaluator_test.go
```diff
@@ -0,0 +1,123 @@
...
+			promql.Matrix{
+				promql.Series{
+					Metric: labels.FromStrings("app", "foo"),
+					Floats: []promql.FPoint{{T: 60 * 1000, F: 6}, {T: 90 * 1000, F: 6}, {T: 120 * 1000, F: 6}},
+				},
+			},
+		},
+		// probabilistic queries
+		{
+			// TODO(karsten): I don't understand the relation between step and interval here.
```
jeschkies, line 72: Compare with https://github.com/grafana/loki/blob/main/pkg/logql/engine_test.go#L1421

pkg/logql/probabilistic_evaluator.go
```diff
@@ -0,0 +1,459 @@
...
+		
+		var groupingKey uint64
+		if e.expr.Grouping.Without {
+			groupingKey, buf = metric.HashWithoutLabels(buf, e.expr.Grouping.Groups...)
+		} else {
+			groupingKey, buf = metric.HashForLabels(buf, e.expr.Grouping.Groups...)
+		}
+
+		// TODO(karsten): support floats.
+		topkAggregation.ObserveForGroupingKey(s.Metric.String(), strconv.FormatUint(groupingKey, 10), uint32(s.F))
```
jeschkies, line 130: So we should either use a number for the grouping key or the labels. I think the later is easier to debug.

pkg/logql/shardmapper_test.go
```diff
@@ -145,11 +147,8 @@ func TestMappingStrings(t *testing.T) {
 			)`,
 		},
 		{
-			in: `topk(3, rate({foo="bar"}[5m]))`,
-			out: `topk(3,
-				downstream<rate({foo="bar"}[5m]), shard=0_of_2>
-				++ downstream<rate({foo="bar"}[5m]), shard=1_of_2>
-			)`,
+			in:  `topk(3, rate({foo="bar"}[5m]))`,
+			out: `merge_topk(downstream_topk<topk(3,rate({foo="bar"}[5m])), shard=0_of_2>, downstream_topk<topk(3,rate({foo="bar"}[5m])), shard=1_of_2>)`,
```
jeschkies, line 151: Hm, since probabilistic is `false` this should not have changed...

pkg/logql/shardmapper_test.go
```diff
@@ -585,44 +585,52 @@ func TestMapping(t *testing.T) {
...
 		{
 			in: `topk(3, rate({foo="bar"}[5m]))`,
-			expr: &syntax.VectorAggregationExpr{
-				Grouping:  &syntax.Grouping{},
-				Params:    3,
-				Operation: syntax.OpTypeTopK,
-				Left: &ConcatSampleExpr{
-					DownstreamSampleExpr: DownstreamSampleExpr{
+			expr: &TopkMergeSampleExpr{
+					DownstreamTopkSampleExpr: DownstreamTopkSampleExpr{
```
jeschkies, line 589: Same here. The mapper should have not touched these.

pkg/logql/probabilistic_evaluator_test.go
```diff
@@ -0,0 +1,274 @@
...
+			},
+			[]SelectSampleParams{
+				{&logproto.SampleQueryRequest{Start: time.Unix(0, 0), End: time.Unix(180, 0), Selector: `rate({app=~"foo|bar"}|~".+bar"[1m])`}},
+			},
+			sketch.TopKMatrix{
+				sketch.TopKVector{
+					Topk: &sketch.Topk{
+						Heaps: map[string]*sketch.MinHeap{"17241709254077376921": {
+							{Event: `{app="bar"}`},
+							{Event: `{app="foo"}`},
```
jeschkies, line 122: @cstyan these don't seem to be deterministic.

pkg/logql/downstream.go
jeschkies: Won't do in this PR.

pkg/logql/downstream_test.go
```diff
@@ -405,25 +407,127 @@ func TestRangeMappingEquivalence(t *testing.T) {
...
+			require.NoError(t, err)
+			p40, err := stats.Percentile(allMisses, 30)
+			require.NoError(t, err)
+
+			//p10, err := stats.Percentile(allMisses, 5)
+			//require.NoError(t, err)
+
+			require.LessOrEqual(t, p99, 2.0)
+			require.LessOrEqual(t, p40, 1.0)
+			//require.LessOrEqual(t, p10, 0.0)
```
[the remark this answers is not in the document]
jeschkies: So I'd expect p90 to be 0 and p95 or so to be 1 or 2.

pkg/logql/sketch/topk.go
```diff
@@ -413,25 +413,48 @@ func (t *Topk) Topk() []TopKResult {
...
+	if !ok {
+		return nil
+	}
+
+	events := make([]string, 0, len(*group))
+	for _, node := range *group {
+		events = append(events, node.Event)
+	}
+	return events
+}
```
cstyan: The function names here are a bit confusing. Maybe I'm just missing what we would use these functions for other than in tests?

Couldn't you just use the function that returns the topk for the grouping key, and then iterate over the topk just selecting the event names?

pkg/logql/downstream_test.go
```diff
@@ -480,7 +480,28 @@ func TestSketchEquivalence(t *testing.T) {
...
+					expectedLabels[i] = sample.Metric.String()
+				}
+				actualLabels := make([]string, len(actualVec))
+				for i, sample := range actualVec {
+					actualLabels[i] = sample.Metric.String()
+				}
+				require.Len(t, actualLabels, len(expectedLabels), "labels lengths differ at TS:%d", ts)
+				// TODO(karsten): use a percentage here.
+				require.ElementsMatchf(t, expectedLabels, actualLabels, "labels differ at TS:%d", ts)
+			}
```
cstyan: in general some more comments in this test would be useful, IIUC we're iterating over the non-probabilistic queries iterator and comparing the probabilistic iterators result? even though the ordering and actual counts for the probabilistic result can be different?

pkg/logql/downstream_test.go
```diff
@@ -484,24 +485,41 @@ func TestSketchEquivalence(t *testing.T) {
...
+				require.LessOrEqual(t, misses, 2)
+
+				totalMisses += misses
+
+				ok, ts, expectedVec = expected.Next()
 			}
+
+			// Overall we want to capture 98% of topk.
+			missedPercentage := float64(totalMisses) / float64(totalLabels)
+			require.LessOrEqual(t, missedPercentage, 2.0)
```
cstyan: we should also allow for something like 9/10 correct for topk(10, ...)

pkg/logql/downstream_test.go
```diff
@@ -508,18 +510,13 @@ func TestSketchEquivalence(t *testing.T) {
...
+				f.Update(float64(misses))
 
 				ok, ts, expectedVec = expected.Next()
 			}
 
-			// Overall we want to capture 98% of topk.
-			missedPercentage := float64(totalMisses) / float64(totalLabels)
-			require.LessOrEqual(t, missedPercentage, 2.0)
+			require.LessOrEqual(t, f.Quantile(0.99), 2.0)
+			require.LessOrEqual(t, f.Quantile(0.50), 1.0)
```
cstyan: I like this way of checking the accuracy :+1:

pkg/logql/downstream_test.go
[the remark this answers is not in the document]
cstyan: Not sure I understand what you're saying here. Was our earlier conversation about the accuracy guarantees for CMS part of this, where it's guaranteeing some % chance of being within a range of `real count`?

Again the overall cardinality and distribution of the data set are a factor in whether the CMS size will produce an "accurate" topk, so maybe we want to pick a few distribution models and test sketch sizes for equivalence against datasets of those distribution models?

dannykopping (commented):
Really like how you structured this; it reads quite well to me.
Left a couple nits.
Are you planning on adding benchmarks in this PR, or a subsequent one?

pkg/logql/probabilistic_evaluator.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+}
+
+func (q *probabilisticQuery) aggregateSampleVectors(
+	ctx context.Context,
+	stepEvaluator ProbabilisticStepEvaluator,
+	tenantIDs []string,
+) (promql_parser.Value, error) {
+
+	// Assert stepEvaluator.Type() == VecType
```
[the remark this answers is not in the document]
jeschkies: Thanks for catching this. I've addressed it.

pkg/logql/probabilistic_evaluator.go
```diff
@@ -411,9 +406,6 @@ func (p *ProbabilisticEvaluator) newProbabilisticVectorAggEvaluator(
 		return nil, errors.Errorf("unexpected operation: want 'topk', have '%q'", expr.Operation)
 	}
 
-	// TODO(karsten): Below is just copy-pasta from newVectorAggEvaluator.
-	// We should find better abstractions.
-
```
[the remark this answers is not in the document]
jeschkies: No but our aggregation is so different now that it's ok to copy.

pkg/logql/downstream_test.go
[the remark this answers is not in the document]
jeschkies: > test sketch sizes for equivalence against datasets of those distribution models?

Yes, we need to add more data sets. I'd like to do this in a follow up.

> Not sure I understand what you're saying here. Was our earlier conversation about the accuracy guarantees for CMS part of this, where it's guaranteeing some % chance of being within a range of real count?

Yes and no. I'll talk about this in more detail tomorrow. I've started from the other end. Given we know the accuracy of a topk vector approximation. What's the probability to get ~90% of the results correct. The accuracy of the CMS determines the accuracy of our vector prediction. But it's more complicated. As I understand the accuracy of the CMS is for one count not all k. It's a little bit more complicated to determine the accuracy to get all k picks or k-1 out of k picks correct.

I've setup a meeting tomorrow following our sync to discuss this with Edward from Machine Learning. We don't want to get too complicated. IMHO we need a way to measure or rather estimate our accuracy for a query.

pkg/logql/probabilistic_evaluator.go
jeschkies: When I was running the tests I just saw the hash value as a string. It was not easy to identify each series. With the label string it's straight forward.

pkg/loki/modules.go
```diff
@@ -406,6 +406,11 @@ func (t *Loki) initQuerier() (services.Service, error) {
 			querier.WrapQuerySpanAndTimeout("query.RangeQuery", t.querierAPI),
 		).Wrap(http.HandlerFunc(t.querierAPI.RangeQueryHandler)),
 
+		"/loki/api/v1/probabilistic_query": middleware.Merge(
+			httpMiddleware,
+			querier.WrapQuerySpanAndTimeout("query.ProbabilisticQuery", t.querierAPI),
+		).Wrap(http.HandlerFunc(t.querierAPI.ProbabilisticRangeQueryHandler)),
+
```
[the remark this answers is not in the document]
jeschkies: If the  querier is run in a standalone way, ie without scheduler it's exposing this route but this should not be supported for now.

pkg/logql/sketch/topk.go
jeschkies: `topk` is not including all events. This is just a convenient way to access all observed events without exposing all internals.

pkg/logql/downstream_test.go
jeschkies: The order is ignored with `ElementsMatch`. So yes, we only check that the non-probabilistic and the probabilistic have the same labels/series. I add a comment.

jeschkies: Just saw that I've changed it.

pkg/logql/downstream_test.go
jeschkies: Thanks. I hope that we have a more refined way tomorrow.

jeschkies:
> Are you planning on adding benchmarks in this PR, or a subsequent one?

That's a good question. I'm not adding benchmarks here. I'm more concerned with getting the semantics right here. This should be followed with accuracy tests and then benchmarks. I'll add it to the epic.

pkg/logql/probabilistic_evaluator.go
jeschkies: For this PR we need to:

- [ ] Decide what the grouping key type should be?
- [ ] Add support for floats.

pkg/logql/downstream_test.go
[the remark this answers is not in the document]
jeschkies: Also, we need to figure out why the results are so bad for this particular dataset.

pkg/logql/sketch/topk.go
cstyan: This is essentially just doing what `TopkForGroupingKey` is doing but not returning the counts for each event. The heap already is the entire list of events we're tracking for the grouping key. We don't track _all_ observed events.

pkg/logql/probabilistic_evaluator_test.go
cstyan: not sure what this comment is referring to?

pkg/logql/probabilistic_evaluator.go
cstyan: I think we'd basically need to reimplement the `HashWithoutLabels` minus the last line: https://github.com/prometheus/prometheus/blob/main/model/labels/labels.go#L189-L208

As long as it's not a big perf. hit somehow I don't have a preference either way.

pkg/logql/downstream_test.go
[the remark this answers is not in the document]
cstyan: > The accuracy of the CMS determines the accuracy of our vector prediction. But it's more complicated. As I understand the accuracy of the CMS is for one count not all k. It's a little bit more complicated to determine the accuracy to get all k picks or k-1 out of k picks correct.

Yes, the accuracy guarantee of the CMS is % probability of getting a count within a range of `real count -> real count + x`. However, currently we're not choosing CMS sizes based on this at all. I was simply generating that random set of data and then increasing the CMS size until we got something like >= 97 of topk 100 events correct. Again, just testing the set membership and not the order or count.

Like I mentioned today, the distribution model of a dataset is a big factor in whether we're going to get an accurate result. In the dataset generation, I was forcing there to always be `K` obvious outliers. See how I'm doing that here: https://github.com/grafana/loki/blob/main/pkg/logql/sketch/topk_test.go#L54-L70

While overall the dataset I was generating there might follow a [normal distribution](https://en.wikipedia.org/wiki/Normal_distribution), I think the skew between the mean and the outliers in the topk is large enough in the dataset I was generating to always get an accurate result.

I still think we should have a way to measure the distribution model of our data set.

pkg/logql/sketch/cms.go
```diff
@@ -19,10 +19,10 @@ func NewCountMinSketch(w, d uint32) (*CountMinSketch, error) {
 	}, nil
 }
 
-func make2dslice(col, row uint32) [][]uint32 {
-	ret := make([][]uint32, row)
+func make2dslice(col, row uint32) [][]float64 {
+	ret := make([][]float64, row)
```
cstyan, line 23: we'll have to reevaluate the CMS sizing and overall memory usage with this change, it essentially doubles the memory size of the sketch

pkg/logql/downstream_test.go
```diff
@@ -512,15 +512,15 @@ func TestSketchEquivalence(t *testing.T) {
...
-			//p10, err := stats.Percentile(allMisses, 10)
-			//require.NoError(t, err)
+			p40, err := stats.Percentile(allMisses, 40)
+			require.NoError(t, err)
 
 			require.LessOrEqual(t, p99, 2.0)
-			require.LessOrEqual(t, p40, 1.0)
-			//require.LessOrEqual(t, p10, 0.0)
+			require.LessOrEqual(t, p90, 1.0)
+			require.LessOrEqual(t, p40, 0.0)
```
cstyan: can we leave a comment explaning what these percentiles are?

dannykopping (approved):
LGTM

PR #9964, opened 2023-07-17
lambda-promtail: Support SNS events

sapslaj:

Some use cases (such as S3 bucket notifications) are better suited for SNS rather than SQS when you need to send events to multiple destinations. Adding a little bit of extra parsing logic for SNS events allows directly connecting an SNS topic to the Lambda Promtail function and bypasses the extra complexity of the SQS queue. This lower complexity is very helpful in cross-AWS-account use cases like CloudTrail event forwarding.

Fixes #<issue number>

- I just updated the SQS section of the docs to also reference SNS, since the functionality is basically identical. Happy to take suggestions on improvements there. Not sure if `CHANGELOG.md` needs to be updated as well or not.
- This is semi-difficult to unit test since there isn't an existing test harness for event matching. Testing in situ on our own infrastructure seems to work.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


tools/lambda-promtail/lambda-promtail/sns.go
```diff
@@ -0,0 +1,21 @@
...
+		if err != nil {
+			return err
+		}
+		err = handler(ctx, event)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
cstyan, line 21: I think we could move this and the sqs code to the `s3.go` file. 

Can you also include a small test for the sns event parsing.

sapslaj: Yep, I can get those moved into `s3.go`.

I'm not 100% sure how to test since the logic is all in `handler` and `stringToRawEvent`. `stringToRawEvent` has a test but `handler` doesn't. I can try to come up with something but if you have any suggestions I'd appreciate it!

sapslaj: @cstyan I had to do a little bit of rework but I got tests for both the SNS and SQS processing functions done. I also fixed a panic in the S3 parsing that I ran into while testing.

tools/lambda-promtail/lambda-promtail/s3.go
```diff
@@ -197,10 +197,11 @@ func getLabels(record events.S3EventRecord) (map[string]string, error) {
 	labels["bucket"] = record.S3.Bucket.Name
 	labels["bucket_owner"] = record.S3.Bucket.OwnerIdentity.PrincipalID
 	labels["bucket_region"] = record.AWSRegion
-	var matchingType *string
 	for key, p := range parsers {
 		if p.filenameRegex.MatchString(labels["key"]) {
-			matchingType = aws.String(key)
+			if labels["type"] == "" {
+				labels["type"] = key
+			}
```
cstyan, line 204: am I understanding the reason for moving the code here, to get rid of the chance of a panic if `matchingType` was still a nil pointer?

IMO if having a value for the `type` label is a requirement for being able to correctly parse an AWS or s3 log in `parseS3log` then we should probably also check for the existence of a `type` label explicitly rather than just erroring `there's no parser found for ""`

we could check in `processEvent` or `parseS3log`, but imo we should explicitly check for empty type


@sapslaj what do you think about that?

sapslaj: @cstyan Good point. It shouldn't be too difficult to add a check and return an error in `processS3Event` if `type` isn't set, that way it won't do a GetObject for a log that will inevitably error out.

sapslaj: @cstyan Alright, that's implemented! I wanted to put the check in `processS3Event` but ended up putting it in `getLabels` so it could be tested.

cstyan:
@sapslaj looks good to me. The current drone CI failure looks unrelated to your changes, can you try either merging main into your branch or rebasing to just confirm we're all green? Then I'll merge :+1:

cstyan (approved):

cstyan:
@sapslaj awesome, thanks! :+1:

PR #9966, opened 2023-07-17
add stats, volume, and volume_range commands to logcli

trevorwhitney:
Add `stats`, `volume`, and `volume_range` commands to `logcli`.

Does not implement the file client for now. I think it would be cool if in the future the file client could read from a downloaded TSDB index file for these commands.


MasslessParticle (approved):

MasslessParticle:
Looks good. Just a small note on wording

PR #9976, opened 2023-07-19
Fix buffered iterator

salvacorts:

In https://github.com/grafana/loki/pull/9700, we added support for writing non-indexed labels from entries into chunks. This PR introduced two bugs:

- When the buffered iterator is closed, the buffer to read metadata labels is put back into the pool but not set to nil. Subsequent uses of the iterator may write on top of the buffer that was put back on the pool. This may lead to inconsistent/incorrect results.
- Inside the buffered iterator, we were not correctly handling EOFs while reading the number of labels and each label length. This was due to the `lastAttemp` variable not being reset.

This PR adds a new test for these two bugs and also fixes them.

Fixes https://github.com/grafana/loki/pull/9700

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


vlad-diachenko (approved):
LGTM

PR #9986, opened 2023-07-19
[promtail] non_indexed_labels stage

vlad-diachenko:

added non_indexed_labels stage to copy values from extracted data tonon-indexed-labels field of the Entry.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


clients/pkg/logentry/stages/stage.go
```diff
@@ -107,134 +109,119 @@ func toStage(p Processor) Stage {
...
-		if err != nil {
-			return nil, err
-		}
-	case StageTypeGeoIP:
-		s, err = newGeoIPStage(logger, cfg)
-		if err != nil {
-			return nil, err
-		}
-	default:
+	creator, ok := getStageCreators()[stageType]
```
salvacorts, line 216: nit: IIUC getStageCreatorsn is a factory. I think it's more idiomatic if it takes the stageType as an argument and returns the stage right away instead fo a map.

salvacorts: Moreover I think this function should take the stageType and the params and return the created stage.

clients/pkg/logentry/stages/nonindexedlabels_test.go
```diff
@@ -0,0 +1,88 @@
...
+		"expected non-indexed labels to be extracted with regexp parser and to be added to entry": {
+			pipelineStagesYaml: pipelineStagesNonIndexedLabelsWithRegexParser,
+			logLine:            `2019-01-01T01:00:00.000000001Z stderr P i'm a log message!`,
+			expectedLabels:     push.LabelsAdapter{push.LabelAdapter{Name: "stream", Value: "stderr"}},
+		},
+		"expected non-indexed labels to be extracted with json parser and to be added to entry after rendering the template": {
+			pipelineStagesYaml: pipelineStagesNonIndexedLabelsFromJsonWithTemplate,
+			logLine:            `{"app":"loki" ,"component":"ingester"}`,
+			expectedLabels:     push.LabelsAdapter{push.LabelAdapter{Name: "app", Value: "LOKI"}},
+		},
```
salvacorts, line 75: I think we should have a test using the labels stage:
```go
var pipelineStagesNonIndexedLabelsFromJson = `
pipeline_stages:
- json:
    expressions:
      app:
      job:
- non_indexed_labels:
    app:
- labels:
    job: 
`
```

clients/pkg/logentry/stages/nonindexedlabels.go
```diff
@@ -0,0 +1,42 @@
...
+type nonIndexedLabelsStage struct {
+	cfgs   LabelsConfig
+	logger log.Logger
+}
+
+func (s *nonIndexedLabelsStage) Name() string {
+	return StageTypeNonIndexedLabels
+}
+
+func (s *nonIndexedLabelsStage) Run(in chan Entry) chan Entry {
```
salvacorts, line 35: The labels stage implements the Processor interface and `newLabelStage` returns a stage via the `toStage` function that takes a `Processor` and wraps it into a `Stage`. This is also done in pretty much the rest of stages.
https://github.com/grafana/loki/blob/a7094e9f19c387ab1eb64381741ce5f8d457c656/clients/pkg/logentry/stages/labels.go#L52-L55

I think we should leave this stage as close to the labels one as possible. So I think we should use the same pattern.

Moreover, the `toStage` function adds an inspector. I'm not familiar with what that is, but I wonder if that's something we want to have provided the rest of stages have it.

clients/pkg/logentry/stages/nonindexedlabels_test.go
vlad-diachenko: done

clients/pkg/logentry/stages/nonindexedlabels.go
vlad-diachenko: Processor does not have access to the entry :
```go
type Processor interface {
	Process(labels model.LabelSet, extracted map[string]interface{}, time *time.Time, entry *string)
	Name() string
}
```
that is why I used Stage:
```
type Stage interface {
	Name() string
	Run(chan Entry) chan Entry
}
```
because non-indexed labels are stored in Entry.
JsonParser is also a stage, not a processor.

clients/pkg/logentry/stages/stage.go
vlad-diachenko: refactored in https://github.com/grafana/loki/pull/9986/commits/3dd5be56156c2a788a4b2153c29783aa650e5f74

salvacorts (approved):
Approved. But before merging, I think we should try to run promtail locally and try to send non-indexed labels to Loki to make sure it works as expected.

salvacorts:
Ups, didn't notice auto-merge was enabled. Nvm, we can test it and if it doesn't work we can revert the PR.

PR #9995, opened 2023-07-20
Add initial delay before starting flush loop

chaudum:

Delay the first flush operation by up to 0.8x the flush time period.

This will ensure that multiple ingesters started at the same time do not flush at the same time. Flushing at the same time can cause concurrently writing the same chunk to object storage, which in AWS S3 leads to being rate limited.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ##### Enhancements
 
+* [9995](https://github.com/grafana/loki/pull/9995) **chaudum**: Add jitter to the flush interval to prevent multiple ingesters to flush at the same time.
 * [9797](https://github.com/grafana/loki/pull/9797) **chaudum**: Add new `loki_index_gateway_requests_total` counter metric to observe per-tenant RPS
 * [9710](https://github.com/grafana/loki/pull/9710) **chaudum**: Add shuffle sharding to index gateway
 * [9573](https://github.com/grafana/loki/pull/9573) **CCOLLOT**: Lambda-Promtail: Add support for AWS CloudFront log ingestion.


pkg/ingester/ingester.go
```diff
@@ -577,6 +578,25 @@ func (i *Ingester) removeShutdownMarkerFile() {
 func (i *Ingester) loop() {
 	defer i.loopDone.Done()
 
+	// Delay the first flush operation by up to 0.8x the flush time period.
+	// This will ensure that multiple ingesters started at the same time do not
+	// flush at the same time. Flushing at the same time can cause concurrently
+	// writing the same chunk to object storage, which in AWS S3 leads to being
+	// rate limited.
+	jitter := rand.Int63n(int64(float64(i.cfg.FlushCheckPeriod.Nanoseconds()) * 0.8))
```
DylanGuedes, line 586: isn't that too little? I feel like you'll still hit the race condition if you only sleep for 0.8 ns

DylanGuedes (changes requested):

DylanGuedes (approved):

DylanGuedes (approved):

pkg/ingester/ingester.go
```diff
@@ -577,6 +578,25 @@ func (i *Ingester) removeShutdownMarkerFile() {
...
 	defer i.loopDone.Done()
 
+	// Delay the first flush operation by up to 0.8x the flush time period.
+	// This will ensure that multiple ingesters started at the same time do not
+	// flush at the same time. Flushing at the same time can cause concurrently
+	// writing the same chunk to object storage, which in AWS S3 leads to being
+	// rate limited.
+	jitter := time.Duration(rand.Int63n(int64(float64(i.cfg.FlushCheckPeriod.Nanoseconds()) * 0.8)))
+	initialDelay := time.NewTimer(jitter)
+	defer initialDelay.Stop()
```
slim-bean, line 588: I think we need to do this slightly differently. I think the jitter needs to be on each evaluation such that the flush loop runs at an interval of say approx ever 30s +/- maybe 5s?

The reason being: the example I looked at the other day, the ingesters actually started maybe 10 minutes apart however their flush loops just unluckily ended up synchronized. So adding just an initial jitter would have the same probability of unluckiness which is apparently pretty high as we see this somewhat often.

slim-bean: I think the jitter needs to be applied constantly, if it's only ever applied once I think we end up with the same odds of flush loops being synchronized and stuck that way indefinitely until an ingester restarts.

chaudum: Additionally to the initial delay, I replaced the Ticker with a ticker with jitter, see line 600ff

pkg/util/ticker.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (t *TickerWithJitter) start(d, dev time.Duration) {
+	j := NewJitter(d, dev)
+	timer := time.NewTimer(j.Duration())
+	defer timer.Stop()
+	for {
+		select {
+		case <-timer.C:
+			timer.Reset(j.Duration())
```
DylanGuedes, line 43: hmmm does that mean you'll be calling "Duration()" to every iter, doesn't it? if so, i feel like it might get expensive. WDYT of calculating the randomness a single time when an ingester is being instantiated, but then to use the random factor on every iteration without reevaluating it?

chaudum: I explicitly do not want a fixed ticker interval.

DylanGuedes: sorry I thought he meant to not rely only on an initial sleep but to rather apply the sleep/jitter on every evaluation

pkg/ingester/ingester.go
```diff
@@ -577,7 +578,28 @@ func (i *Ingester) removeShutdownMarkerFile() {
 func (i *Ingester) loop() {
 	defer i.loopDone.Done()
 
-	flushTicker := time.NewTicker(i.cfg.FlushCheckPeriod)
+	// Delay the first flush operation by up to 0.8x the flush time period.
```
kavirajk, line 581: Appreciate your comment here that explains the behavior. Wondering should we make it clear the same in `cfg.FlushCheckPeriod` config as well.

Rationale being, this avoid surprises for loki operators. Without mentioning this delays(during both initial and every iteration) may set different expectation of the `cfg.FlushCheckPeriod` config IMO.

kavirajk (approved):
Nice work :+1: 

Add one minor comment.

pkg/ingester/ingester.go
chaudum: @kavirajk Updated the docs for the `FlushCheckPeriod` setting.

pkg/util/ticker.go
```diff
@@ -0,0 +1,65 @@
...
+			t.C <- time.Now()
+		case <-t.ctx.Done():
+			if !timer.Stop() {
+				<-timer.C
```
ashwanthgoli, line 47: could this block if the channel is empty?

docs/sources/configure/_index.md
```diff
@@ -1442,7 +1442,9 @@ lifecycler:
 # CLI flag: -ingester.concurrent-flushes
 [concurrent_flushes: <int> | default = 32]
 
-# How often should the ingester see if there are any blocks to flush.
+# How often should the ingester see if there are any blocks to flush. The first
+# flush check is delayed by a random time up to 0.8x the flush check period.
```
ashwanthgoli, line 1446: nit: jitter now applies to every flush

ashwanthgoli: nvm

chaudum: See next line :)

pkg/util/ticker.go
chaudum: From the Go docs:

```
// To ensure the channel is empty after a call to Stop, check the
// return value and drain the channel.
// For example, assuming the program has not received from t.C already:
//
//	if !t.Stop() {
//		<-t.C
//	}
//
// This cannot be done concurrent to other receives from the Timer's
// channel or other calls to the Timer's Stop method.
```

Since I don't receive from the timer's channel concurrently - but sequentially in the `for-select` loop - the timer can fire, then it receives from the channel, or the context is done, then the timer hasn't fired yet and the timer channel needs to be drained.

ashwanthgoli: thanks @chaudum 
comment on `Reset` also recommends doing this
https://cs.opensource.google/go/go/+/refs/tags/go1.20.6:src/time/sleep.go;l=109

ashwanthgoli (approved):

PR #10011, opened 2023-07-21
Support non-indexed labels in WAL

salvacorts:
In https://github.com/grafana/loki/pull/9700, we support encoding and decoding non-indexed labels for each entry into the chunks. This PR adds support for writing/reading the non-indexed labels to/from the WAL.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/ingester/wal/encoding.go
+++ b/pkg/ingester/wal/encoding.go
@@ -25,10 +25,13 @@ const (
 	// WALRecordEntriesV2 is the type for the WAL record for samples with an
 	// additional counter value for use in replaying without the ordering constraint.
 	WALRecordEntriesV2
+	// WALRecordEntriesV3 is the type for the WAL record for samples with non-indexed labels.
+	WALRecordEntriesV3
 )
 
 // The current type of Entries that this distribution writes.
 // Loki can read in a backwards compatible manner, but will write the newest variant.
+// TODO: Change to WALRecordEntriesV3?
 const CurrentEntriesRec = WALRecordEntriesV2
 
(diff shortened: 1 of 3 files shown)


pkg/ingester/wal/encoding.go
```diff
@@ -25,10 +25,13 @@ const (
 	// WALRecordEntriesV2 is the type for the WAL record for samples with an
 	// additional counter value for use in replaying without the ordering constraint.
 	WALRecordEntriesV2
+	// WALRecordEntriesV3 is the type for the WAL record for samples with non-indexed labels.
+	WALRecordEntriesV3
 )
 
 // The current type of Entries that this distribution writes.
 // Loki can read in a backwards compatible manner, but will write the newest variant.
+// TODO: Change to WALRecordEntriesV3?
```
salvacorts, line 34: I think it's ok to set `CurrentEntriesRec` to `WALRecordEntriesV3` in this PR. If we are still not writing/reading non-indexed labels into/from chunks, we can still write/read non-indexed labels from the WAL. This way, if an ingester restarts, we won't lose the non-indexed labels. **Wdyt?**

vlad-diachenko: I am not sure here.... let's follow the same approach with headblock and chunkformat. let's merge the changes that allow us to read the next format but does not write in the new format, and next week let's change default formats for: WAL, chunk and head block to the new formats with non indexed labels.

vlad-diachenko (approved):
LGTM

PR #10044, opened 2023-07-24
implementing string interning to optimize resource usage for storing and processing non-indexed labels

sandeepsukhani:
In PR #9700, we added support for storing non-indexed labels in chunks. This PR optimizes resource usage for storing and processing non-indexed labels by doing string interning. We will store deduped label names and values as a list in chunks 
in the newly added non-indexed labels section. The labels would then be referenced in blocks by their index(called symbols).

Additionally, I have started the convention of writing lengths of sections with their offsets within chunks, making it easier to introduce new sections. The section offsets and lengths would be stored at the end of the chunk, similar to [TOC](https://ganeshvernekar.com/blog/prometheus-tsdb-persistent-block-and-its-index/#a-toc) in TSDB.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/chunkenc/memchunk.go
+++ b/pkg/chunkenc/memchunk.go
@@ -1418,8 +1418,8 @@ func (si *bufferedIterator) moveNext() (int64, []byte, labels.Labels, bool) {
 	}
 
 	lastAttempt = 0
-	var labelsWidth, nSymbols int
-	for labelsWidth == 0 { // Read until we have enough bytes for the labels.
+	var symbolsSectionLengthWidth, nSymbolsWidth, nSymbols int
+	for nSymbolsWidth == 0 { // Read until we have enough bytes for the labels.
 		n, err := si.reader.Read(si.readBuf[si.readBufValid:])
 		si.readBufValid += n
 		if err != nil {
@@ -1436,7 +1436,8 @@ func (si *bufferedIterator) moveNext() (int64, []byte, labels.Labels, bool) {
 			}
 		}
 		var l uint64
-		l, labelsWidth = binary.Uvarint(si.readBuf[:si.readBufValid])
+		_, symbolsSectionLengthWidth = binary.Uvarint(si.readBuf[:si.readBufValid])
+		l, nSymbolsWidth = binary.Uvarint(si.readBuf[symbolsSectionLengthWidth:si.readBufValid])
 		nSymbols = int(l)
 		lastAttempt = si.readBufValid
 	}
@@ -1447,7 +1448,31 @@ func (si *bufferedIterator) moveNext() (int64, []byte, labels.Labels, bool) {
 	decompressedNonIndexedLabelsBytes += int64(nSymbols * 2 * binary.MaxVarintLen64)
 
 	// Shift down what is still left in the fixed-size read buffer, if any.
-	si.readBufValid = copy(si.readBuf[:], si.readBuf[labelsWidth:si.readBufValid])
+	si.readBufValid = copy(si.readBuf[:], si.readBuf[symbolsSectionLengthWidth+nSymbolsWidth:si.readBufValid])
+
+	/*
+		Commented out tested code, which lets us skip reading the symbols section altogether.
+		Leaving it here if in case we need it in future.
+
+		symbolsSectionLength -= nSymbolsWidth
+		if symbolsSectionLength > 0 {
+			readBufValid := si.readBufValid
+			if symbolsSectionLength >= si.readBufValid {
+				si.readBufValid = 0
+			} else {
+				si.readBufValid = copy(si.readBuf[:], si.readBuf[symbolsSectionLength:si.readBufValid])
+			}
+			symbolsSectionLength -= readBufValid - si.readBufValid
+			if symbolsSectionLength > 0 {
+				_, err := si.reader.Read(make([]byte, symbolsSectionLength))
+				if err != nil {
+					si.err = err
+					return 0, nil, nil, false
+				}
+			}
+			nSymbols = 0
+		}
+	*/
 
 	// If not enough space for the symbols, create a new buffer slice and put the old one back in the pool.
 	if nSymbols > cap(si.symbolsBuf) {
--- a/pkg/chunkenc/memchunk.go
+++ b/pkg/chunkenc/memchunk.go
@@ -42,6 +42,9 @@ const (
 	// This could wary from configured block size using `ingester.chunks-block-size` flag or equivalent yaml config resulting in
 	// different block size in the new chunk which should be fine.
 	defaultBlockSize = 256 * 1024
+
+	chunkMetasSectionIdx            = 1
+	chunkNonIndexedLabelsSectionIdx = 2
 )
 
 var HeadBlockFmts = []HeadBlockFmt{OrderedHeadBlockFmt, UnorderedHeadBlockFmt, UnorderedWithNonIndexedLabelsHeadBlockFmt}
@@ -157,7 +160,7 @@ func (hb *headBlock) Entries() int { return len(hb.entries) }
 
 func (hb *headBlock) UncompressedSize() int { return hb.size }
 
-func (hb *headBlock) Reset(*symbolizer) {
+func (hb *headBlock) Reset() {
 	if hb.entries != nil {
 		hb.entries = hb.entries[:0]
 	}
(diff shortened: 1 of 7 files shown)


pkg/chunkenc/symbols.go
```diff
@@ -0,0 +1,374 @@
...
+}
+
+type symbols []symbol
+
+// symbolizer holds a collection of label names and values and assign symbols to them.
+// symbols are actually index numbers assigned based on when the entry is seen for the first time.
+type symbolizer struct {
+	mtx            sync.RWMutex
+	symbolsMap     map[string]uint32
+	labels         []string
```
salvacorts, line 29: I find this name a bit confusing. IIUC, this is an array with all the symbols. I think it should be named `symbols`

pkg/chunkenc/symbols.go
```diff
@@ -0,0 +1,374 @@
...
+	if ok {
+		return idx
+	}
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	idx, ok = s.symbolsMap[lbl]
+	if !ok {
+		idx = uint32(len(s.symbolsMap))
```
salvacorts, line 70: Since `idx` is the index within the `labels` slice, I think it would be easier to understand this if we look at the len of `labels`

pkg/chunkenc/symbols.go
```diff
@@ -0,0 +1,374 @@
...
+	var (
+		readBuf      [10]byte // Enough bytes to store one varint.
+		readBufValid int      // How many bytes are left in readBuf from previous read.
+		s            symbolizer
+		buf          []byte
+	)
+
+	s.compressedSize = len(b)
+
+	for {
```
salvacorts, line 319: IIUC this loop will read until there is an EOF. Since in the future, we might want to put something after the symbols. I think we should rather store the number of symbols first and then read that many.

pkg/chunkenc/unordered.go
```diff
@@ -88,14 +90,14 @@ func (hb *unorderedHeadBlock) UncompressedSize() int {
 	return hb.size
 }
 
-func (hb *unorderedHeadBlock) Reset() {
-	x := newUnorderedHeadBlock(hb.format)
+func (hb *unorderedHeadBlock) Reset(symbolizer *symbolizer) {
```
salvacorts, line 93: Looks like, the same symbolizer is reused after calling `Reset`.
https://github.com/grafana/loki/blob/238b174416be3d8d515722527a2802b6abfd9fbc/pkg/chunkenc/memchunk.go#L905

To minimize passing the symbolizer around, I think it makes more sense to remove the symbolizer argument and just pass `hb.symbolizer` to `newUnorderedHeadBlock`

pkg/chunkenc/unordered.go
```diff
@@ -88,14 +90,14 @@ func (hb *unorderedHeadBlock) UncompressedSize() int {
...
+	x := newUnorderedHeadBlock(hb.format, symbolizer)
 	*hb = *x
 }
 
 type nsEntry struct {
-	line             string
-	nonIndexedLabels labels.Labels
+	line    string
+	symbols symbols
```
salvacorts, line 100: I think this should be renamed to `nonIndexedLabelsSymbols` to account for what the symbols are used for

pkg/chunkenc/unordered.go
```diff
@@ -181,7 +184,7 @@ func (hb *unorderedHeadBlock) forEntries(
 	direction logproto.Direction,
 	mint,
 	maxt int64,
-	entryFn func(*stats.Context, int64, string, labels.Labels) error, // returning an error exits early
+	entryFn func(*stats.Context, int64, string, symbols) error, // returning an error exits early
```
salvacorts, line 187: Same wrt naming: `nonIndexedLabelsSymbols`

pkg/chunkenc/unordered.go
```diff
@@ -570,23 +547,21 @@ func (hb *unorderedHeadBlock) LoadBytes(b []byte) error {
 		lineLn := db.uvarint()
 		line := string(db.bytes(lineLn))
 
-		var metaLabels labels.Labels
+		var symbols symbols
```
salvacorts, line 570: Same with naming. I think it should be named `nonIndexedLabelsSymbols`

pkg/chunkenc/unordered.go
```diff
@@ -262,8 +261,8 @@ func (hb *unorderedHeadBlock) Iterator(
 		direction,
 		mint,
 		maxt,
-		func(statsCtx *stats.Context, ts int64, line string, nonIndexedLabels labels.Labels) error {
-			newLine, parsedLbs, matches := pipeline.ProcessString(ts, line, nonIndexedLabels...)
+		func(statsCtx *stats.Context, ts int64, line string, symbols symbols) error {
```
salvacorts, line 264: Same with naming. I think it should be named nonIndexedLabelsSymbols

pkg/chunkenc/memchunk.go
```diff
@@ -316,11 +320,11 @@ func (hb *headBlock) LoadBytes(b []byte) error {
 	return nil
 }
 
-func (hb *headBlock) Convert(version HeadBlockFmt) (HeadBlock, error) {
+func (hb *headBlock) Convert(version HeadBlockFmt, symbolizer *symbolizer) (HeadBlock, error) {
```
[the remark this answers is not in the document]
sandeepsukhani: It was a mistake passing the symbolizer as nil everywhere. I am going to fix it. Nice catch!

pkg/chunkenc/memchunk.go
```diff
@@ -1280,10 +1417,8 @@ func (si *bufferedIterator) moveNext() (int64, []byte, [][]byte, bool) {
 		return ts, si.buf[:lineSize], nil, true
 	}
 
-	// TODO: This is pretty similar to how we read the line size, and the metadata name and value sizes
-	//       Maybe we can extract it to a separate function and reuse it?
 	lastAttempt = 0
-	var labelsWidth, nLabels int
+	var labelsWidth, nSymbols int
```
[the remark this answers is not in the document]
sandeepsukhani: I have already taken care of it in the last commit I pushed

pkg/chunkenc/symbols.go
sandeepsukhani: No, it is a list of labels in string form. `symbols` are actually references by index to the deduped list of labels.

pkg/chunkenc/symbols.go
sandeepsukhani: It should have the same length. `symbolsMap` is actually for deduping while `labels` is used at query time to find labels referenced by index. I will change it the way you suggested since I am fine with either ways.

pkg/chunkenc/symbols.go
```diff
@@ -0,0 +1,394 @@
...
+			if err != nil {
+				// We might get EOF after reading enough bytes to fill the buffer, which is OK.
+				// EOF and zero bytes read when the buffer isn't full is an error.
+				if err == io.EOF && r != 0 {
+					continue
+				}
+				return nil, err
+			}
+		}
+		s.labels = append(s.labels, string(buf))
```
vlad-diachenko, line 389: should we also populate `symbolsMap` to be consistent?

pkg/chunkenc/symbols.go
```diff
@@ -0,0 +1,394 @@
...
+	reader, err := pool.GetReader(bytes.NewBuffer(b))
+	if err != nil {
+		return nil, err
+	}
+	defer pool.PutReader(reader)
+
+	var (
+		readBuf      [10]byte // Enough bytes to store one varint.
+		readBufValid int      // How many bytes are left in readBuf from previous read.
+		s            symbolizer
```
vlad-diachenko, line 331: can we init symbolized with slice and map size enough to store `numSymbols` ?

pkg/chunkenc/unordered.go
```diff
@@ -89,13 +91,13 @@ func (hb *unorderedHeadBlock) UncompressedSize() int {
 }
 
 func (hb *unorderedHeadBlock) Reset() {
-	x := newUnorderedHeadBlock(hb.format)
+	x := newUnorderedHeadBlock(hb.format, hb.symbolizer)
```
vlad-diachenko, line 94: should we also reset symbolizer?

vlad-diachenko: i might be wrong, I believe we need to reuse symbolizer until we cut the chunk. am I right?

vlad-diachenko (approved):
lgtm. small comments

pkg/chunkenc/symbols.go
sandeepsukhani: We load data from encoded bytes from a flushed chunk, so we only need the `labels` list for queries, so I am not populating `symbolsMap` here to save some space and processing.

pkg/chunkenc/unordered.go
sandeepsukhani: yes, that is correct.

pkg/chunkenc/symbols.go
```diff
@@ -0,0 +1,394 @@
...
+func symbolizerFromCheckpoint(b []byte) *symbolizer {
+	if len(b) == 0 {
+		return newSymbolizer()
+	}
+
+	db := decbuf{b: b}
+	s := symbolizer{
+		symbolsMap: map[string]uint32{},
+	}
+	numLabels := db.uvarint()
```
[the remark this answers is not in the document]
sandeepsukhani: good catch, not sure how I missed it. Thanks!

PR #10047, opened 2023-07-24
add volume and volume_range docs

trevorwhitney:
Add docs for the `volume` and `volume_range` endpoints.


docs/sources/reference/api.md
```diff
@@ -868,6 +870,28 @@ It is an approximation with the following caveats:
...
+
+The `query` should be a valid LogQL stream selector, for example `{job="foo", env=~".+"}`. By defult, these endpoints will aggregate into series consiting of all matches for labels included in the query. For example, assuming you have the streams `{job="foo", env="prod", team="alpha"}`, `{job="bar", env="prod", team="beta"}`, `{job="foo", env="dev", team="alpha"}`, and `{job="bar", env="dev", team="beta"}` in your system. The query `{job="foo", env=~".+"}` would return the two metric series `{job="foo", env="dev"}` and `{job="foo", env="prod"}`, each with datapoints representing the accumulate values of chunks for the streams matching that selector, which in this case would be the streams `{job="foo", env="dev", team="alpha"}` and `{job="foo", env="prod", team="alpha"}`, respectively.
```
JStickler, line 877:
```suggestion
The `query` should be a valid LogQL stream selector, for example `{job="foo", env=~".+"}`. By default, these endpoints will aggregate into series consisting of all matches for labels included in the query. For example, assuming you have the streams `{job="foo", env="prod", team="alpha"}`, `{job="bar", env="prod", team="beta"}`, `{job="foo", env="dev", team="alpha"}`, and `{job="bar", env="dev", team="beta"}` in your system. The query `{job="foo", env=~".+"}` would return the two metric series `{job="foo", env="dev"}` and `{job="foo", env="prod"}`, each with datapoints representing the accumulate values of chunks for the streams matching that selector, which in this case would be the streams `{job="foo", env="dev", team="alpha"}` and `{job="foo", env="prod", team="alpha"}`, respectively.
```

docs/sources/reference/api.md
```diff
@@ -868,6 +870,28 @@ It is an approximation with the following caveats:
...
+
+The other way to change aggregations is with the `aggregateBy` parameter. The default value for this is `series`, which aggregates into combinations of matching key-value pairs. This can instead be spefified as `labels`, which will aggregate into labels only. In this case, the response will have a metric series with a label name matching each label, and a label value of `""`. This is useful for exploring logs at a high level. For example, if you wanted to know what percentage of your logs had a `team` label, you could query your logs with `aggregateBy=labels` and a query with either an exact or regex match on `team`, or by including `team` in the list of `targetLabels`.
```
JStickler, line 881:
```suggestion
The other way to change aggregations is with the `aggregateBy` parameter. The default value for this is `series`, which aggregates into combinations of matching key-value pairs. Alternately this can be specified as `labels`, which will aggregate into labels only. In this case, the response will have a metric series with a label name matching each label, and a label value of `""`. This is useful for exploring logs at a high level. For example, if you wanted to know what percentage of your logs had a `team` label, you could query your logs with `aggregateBy=labels` and a query with either an exact or regex match on `team`, or by including `team` in the list of `targetLabels`.
```

JStickler (commented):
[doc team]  Couple of typos.

docs/sources/reference/api.md
```diff
@@ -868,6 +870,28 @@ It is an approximation with the following caveats:
 These make it generally more helpful for larger queries.
 It can be used for better understanding the throughput requirements and data topology for a list of matchers over a period of time.
 
+## Volume
+
+The `/loki/api/v1/index/volume` and `/loki/api/v1/index/volume_range` endpoints can be used to query the index for volume information about label and label-value combinations. This is helpful in exploring the logs Loki has ingested to find high or low volume streams. The `volume` endpoint returns aggregations over the entire time period, returning a Prometheus style vector response with one timestamp and datapoint per matching set of labels or series. The `volume_range` endoint returns a Prometheus style matrix response, with multiple timestamps and corresponding datapoints per matching set of labels or series. The number of timestamps returned when querying `volume_range` will be determined by the provided `step` parameter.
```
MasslessParticle, line 875: I think it would be useful to adopt the same language as the `query` and `query_range` endpoint docs where they talk about `a point in time` vs `over a range of time`

MasslessParticle (approved):
The content looks good. I think we could use some similar language to other parts of the docs to avoid (hopefully?) some confusion.

PR #10055, opened 2023-07-25
Remove redundant spans

DylanGuedes:
- Stop using `instrument.CollectedRequest` on all Index calls since it emits spans that are redundant with existing ones and aren't improving Loki's monitoring while making it more expensive to run
- Add events to `QuerySample` to improve its monitoring so we can better track what is holding down its performance
- Remove redundant Ingester.GetChunkRef span

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -920,6 +920,7 @@ func (i *Ingester) Query(req *logproto.QueryRequest, queryServer logproto.Querie
 func (i *Ingester) QuerySample(req *logproto.SampleQueryRequest, queryServer logproto.Querier_QuerySampleServer) error {
 	// initialize stats collection for ingester queries.
 	_, ctx := stats.NewContext(queryServer.Context())
+	sp := opentracing.SpanFromContext(ctx)
 
 	instanceID, err := tenant.TenantID(ctx)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/storage/stores/index/index.go
DylanGuedes, line 1: All these different `instrument.CollectedRequest` generate another span that doesn't have any valuable information and is redundant with the span already generated by our gRPC middleware. Example in the screenshot: the `chunk_refs` span is redundant is invoked less than 100microseconds after and before the others and doesn't add any information/event. It is also quite expensive, since it will be emitted for every GetChunkRef span.

image

pkg/ingester/instance.go
DylanGuedes, line 1: adding a new event to our `QuerySampleBatch` span reporting the size of the batch. This way we can track if a QuerySampleBatch span is taking too long because its batch size is huge (which is expected) or if it is for a different reason (unexpected).

pkg/ingester/ingester.go
```diff
@@ -1107,8 +1112,7 @@ func (i *Ingester) Series(ctx context.Context, req *logproto.SeriesRequest) (*lo
 }
 
 func (i *Ingester) GetStats(ctx context.Context, req *logproto.IndexStatsRequest) (*logproto.IndexStatsResponse, error) {
-	sp, ctx := opentracing.StartSpanFromContext(ctx, "Ingester.GetStats")
```
DylanGuedes, line 1110: This span is redundant with the `/logproto.Querier/GetStats` span. We won't be losing any information by removing it but we'll end up with a way more concise/smaller trace.
image

ashwanthgoli (approved):
lgtm

PR #10058, opened 2023-07-25
Loki: Tweak how we handle X-Query-Tags header

slim-bean:

Adjust the regex for X-Query-Tags to allow @ and . so that email addresses can be displayed properly. 
Also changed the ToLower to only apply to the key not the value

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/metrics.go
+++ b/pkg/logql/metrics.go
@@ -335,13 +335,13 @@ func tagsToKeyValues(queryTags string) []interface{} {
 		if len(val) != 2 {
 			continue
 		}
-		vals = append(vals, val...)
+		vals = append(vals, strings.ToLower(val[0]), val[1])
 	}
 
 	res := make([]interface{}, 0, len(vals))
 
 	for _, val := range vals {
-		res = append(res, strings.ToLower(val))
+		res = append(res, val)
 	}
 
 	return res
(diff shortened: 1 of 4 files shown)


MasslessParticle (approved):
lgtm

DylanGuedes (approved):

pkg/logql/metrics.go
```diff
@@ -335,13 +335,13 @@ func tagsToKeyValues(queryTags string) []interface{} {
 		if len(val) != 2 {
 			continue
 		}
-		vals = append(vals, val...)
+		vals = append(vals, strings.ToLower(val[0]), val[1])
```
DylanGuedes, line 338: double check: is it fine to treat "level=warn" different than "level=WARN"? i

slim-bean: the only way this is applied is for something we pass in via the X-Query-Tags header, so it's a little hard to reason about what we'll send. previously this was forcing everything to lowercase but I didn't really want that behavior for the value.  in all of our other logfmt the key is lowercase so it made sense to keep lowercasing the key. but we leave the value along now.

PR #10059, opened 2023-07-25
logcli: fix volume routes, add targetLabels and aggregateByLabels

trevorwhitney:
This PR fixes `logcli` to use the new endpoints for volume, which are now `volume` and `volume_range` instead of `series_volume` and `series_volume_range`. This PR also adds support for the additional parameters `targetLabels` and `aggregateBy`.

### DIFF
--- a/pkg/logcli/index/volume.go
+++ b/pkg/logcli/index/volume.go
@@ -11,12 +11,14 @@ import (
 )
 
 type VolumeQuery struct {
-	QueryString string
-	Start       time.Time
-	End         time.Time
-	Step        time.Duration
-	Quiet       bool
-	Limit       int
+	QueryString       string
+	Start             time.Time
+	End               time.Time
+	Step              time.Duration
+	Quiet             bool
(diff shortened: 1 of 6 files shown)


pkg/logcli/index/volume.go
```diff
@@ -45,9 +47,9 @@ func (q *VolumeQuery) volume(rangeQuery bool, c client.Client) *loghttp.QueryRes
 	var err error
 
 	if rangeQuery {
-		resp, err = c.GetVolumeRange(q.QueryString, q.Start, q.End, q.Step, q.Limit, q.Quiet)
+		resp, err = c.GetVolumeRange(q.QueryString, q.Start, q.End, q.Step, q.Limit, q.TargetLabels, q.AggregateByLabels, q.Quiet)
```
MasslessParticle, line 50: Would it be worth just passing the query to these so we don't have to change all the interfaces when we want to add stuff

MasslessParticle (approved):

PR #10065, opened 2023-07-26
fix: parse expression error message based on record or alert

fgouteroux:
Hello,

I want to add this quick fix on the syntax error message  when parsing expression rule.

When there is a syntax error in alert rule expression, it return
```
could not parse expression for record '' in group 'should_fire': parse error at line 3, col 1: syntax error: unexpected IDENTIFIER
```
instead of
```
could not parse expression for alert 'HighPercentageError' in group 'should_fire': parse error at line 3, col 1: syntax error: unexpected IDENTIFIER
```


dannykopping (approved):
Thank you!
Can you please add a CHANGELOG entry and a test?

fgouteroux:
> Thank you! Can you please add a CHANGELOG entry and a test?

Yes, done. One test is in error but seems not related to my change.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ##### Enhancements
 
+* [10065](https://github.com/grafana/loki/pull/10065) **fgouteroux**: fix the syntax error message when parsing expression rule.
```
dannykopping, line 9: Please move this under the Fixes section

pkg/ruler/compat_test.go
```diff
@@ -6,15 +6,52 @@ import (
...
+
 	"github.com/grafana/loki/pkg/iter"
 	"github.com/grafana/loki/pkg/logql"
 	rulerbase "github.com/grafana/loki/pkg/ruler/base"
 	"github.com/grafana/loki/pkg/util/log"
 	"github.com/grafana/loki/pkg/validation"
 )
 
+// TestInvalidRuleGroup tests that a validation error is raised when rule group is invalid
+func TestInvalidRuleGroup(t *testing.T) {
```
dannykopping, line 22: I appreciate that you added this test, but I don't see how this validates your change?

dannykopping (changes requested):

pkg/ruler/compat_test.go
fgouteroux: you right it validate the whole chain but not the error message

dannykopping (approved):
LGTM, thanks a lot @fgouteroux!

PR #10073, opened 2023-07-26
Add docs for non-indexed labels

salvacorts:

Adds documentation for the new experimental non-indexed labels feature.


docs/sources/get-started/labels/_index.md
```diff
@@ -19,6 +19,9 @@ Labels in Loki perform a very important task: They define a stream. More specifi
 
 If you are familiar with Prometheus, the term used there is series; however, Prometheus has an additional dimension: metric name. Loki simplifies this in that there are no metric names, just labels, and we decided to use streams instead of series.
 
+> **Note** This does not apply to non-indexed labels, these do not define a stream but are metadata attached to a log line.
+> See [Non-indexed labels]({{< relref "./non-indexed-labels" >}}) for more information.
+
```
JStickler, line 24:
```suggestion
{{% admonition type="note" %}}
Non-indexed labels do not define a stream, but are metadata attached to a log line.
See [Non-indexed labels]({{< relref "./non-indexed-labels" >}}) for more information.
{{% /admonition %}}
```

docs/sources/get-started/labels/_index.md
```diff
@@ -148,7 +151,7 @@ The two previous examples use statically defined labels with a single value; how
       __path__: /var/log/apache.log
 ```
 
-This regex matches every component of the log line and extracts the value of each component into a capture group. Inside the pipeline code, this data is placed in a temporary data structure that allows using it for several purposes during the processing of that log line (at which point that temp data is discarded). Much more detail about this can be found in the [Promtail pipelines]({{< relref "../clients/promtail/pipelines" >}}) documentation.
+This regex matches every component of the log line and extracts the value of each component into a capture group. Inside the pipeline code, this data is placed in a temporary data structure that allows using it for several purposes during the processing of that log line (at which point that temp data is discarded). Much more detail about this can be found in the [Promtail pipelines]({{< relref "../../clients/promtail/pipelines" >}}) documentation.
```
JStickler, line 154:
```suggestion
This regex matches every component of the log line and extracts the value of each component into a capture group. Inside the pipeline code, this data is placed in a temporary data structure that allows using it for several purposes during the processing of that log line (after processing, the temp data is discarded). More detail about this can be found in the [Promtail pipelines]({{< relref "../../clients/promtail/pipelines" >}}) documentation.
```

docs/sources/get-started/labels/bp-labels.md
```diff
@@ -20,6 +20,9 @@ Too many label value combinations leads to too many streams. The penalties for t
 
 To avoid those issues, don't add a label for something until you know you need it! Use filter expressions (`|= "text"`, `|~ "regex"`, …) and brute force those logs. It works -- and it's fast.
 
+If you often extract a label from a log line at query time, the label has a high cardinality, and extracting that label is expensive in terms of performance; consider extracting the label on the client side and
+attach it as metadata to log lines using a [non-indexed label]({{< relref "./non-indexed-labels" >}}).
```
JStickler, line 24: Is "extract" the verb we want here? It means to remove.  I did some revising, I think this captures what you're trying to say here, but feel free to edit if I'm misunderstanding.
```suggestion
Cardinality is the overall count of values for one label. If you often search for a label that applies to many log lines, that label has a high cardinality. Cardinality can have a major impact on querying where, the higher the cardinality, the more items are required to be iterated over during a query. This makes high cardinality labels expensive in terms of performance; consider extracting the label on the client side and attaching it as metadata to log lines using a [non-indexed label]({{< relref "./non-indexed-labels" >}}).
```

docs/sources/get-started/labels/non-indexed-labels.md
```diff
@@ -0,0 +1,50 @@
...
+description: Attaching metadata to logs.
+---
+# What are non-indexed labels
+
+One of the powerful features of Loki is parsing logs at query time to extract metadata and build labels out of it.
+However, the parsing of logs at query time comes with a cost which can be significantly high for, as an example,
+large json blobs or a poorly written query using complex regex patterns.
+
+In addition, the data extracted from logs at query time is usually high cardinality, which can’t be stored
+in the index without blowing the cardinality.
```
JStickler, line 13: "blowing the cardinality" is confusing here. It's not clear what you mean.

docs/sources/get-started/labels/non-indexed-labels.md
```diff
@@ -0,0 +1,50 @@
...
+
+Non-indexed labels are a way to attach metadata to logs without indexing them. Examples of useful metadata are
+trace IDs, user IDs, and any other label that is often used in queries but has high cardinality and is expensive
+to extract at query time.
+
+## Attaching non-indexed labels to log lines
+
+Non-indexed labels are optionally attached to log lines in the push payload along with each log line and the timestamp.
```
JStickler, line 21: Let's make this active voice, since it has to be configured.
```suggestion
You have the option to attach non-indexed labels to log lines in the push payload along with each log line and the timestamp.
```

docs/sources/get-started/labels/non-indexed-labels.md
```diff
@@ -0,0 +1,50 @@
...
+For more information on how to push logs to Loki via the HTTP endpoint, refer to the [HTTP API documentation]({{< relref "../../reference/api#push-log-entries-to-loki" >}}).
+
+Alternatively, you can use Promtail or the Grafana Agent to extract and attach non-indexed labels to your log lines.
```
JStickler, line 24:
```suggestion
Alternatively, you can use the Grafana Agent or Promtail to extract and attach non-indexed labels to your log lines.
```

docs/sources/reference/api.md
```diff
@@ -583,13 +583,23 @@ JSON post body can be sent in the following format:
       },
       "values": [
           [ "<unix epoch in nanoseconds>", "<log line>" ],
-          [ "<unix epoch in nanoseconds>", "<log line>" ]
+          [ "<unix epoch in nanoseconds>", "<log line>", {"optiona_non_indexed_label": "label_value", "trace_ID": "1234"}]
```
JStickler, line 586:
```suggestion
          [ "<unix epoch in nanoseconds>", "<log line>", {"optional_non_indexed_label": "label_value", "trace_ID": "1234"}]
```

docs/sources/get-started/labels/bp-labels.md
salvacorts: > Cardinality is the overall count of values for one label. If you often search for a label that applies to many log lines, that label has a high cardinality. Cardinality can have a major impact on querying where, the higher the cardinality, the more items are required to be iterated over during a query

I think this isn't quite right in this context. We use stream labels and the time span to reduce the number of logs we need to process. Now, this is talking about labels parsed at query time. So if a log matches a parser, the value for the label parsed is added to the result labels. So we'll still need to process all the logs. 

I think the original paragraph is more correct than the one you suggested. I agree _extract_ may not be the best term here. I'd instead use _parse_ here. Wdyt?

```suggestion
If you often parse a label from a log line at query time, the label has a high cardinality, and extracting that label is expensive in terms of performance; consider extracting the label on the client side
attaching it as metadata to log lines using a [non-indexed label]({{< relref "./non-indexed-labels" >}}).
```

docs/sources/get-started/labels/non-indexed-labels.md
salvacorts: By _blowing the cardinality_, I mean increasing the cardinality too much.

What about:
```suggestion
in the index as it would increase the cardinality too much, and therefore reduce the performance of the index.
```

docs/sources/get-started/architecture.md
```diff
@@ -20,36 +20,52 @@ index and in stored chunks.
...
+  |     #blocks len (uvarint)       |     #blocks offset (uvarint)      |
+  -----------------------------------------------------------------------
 ```
 
 `mint` and `maxt` describe the minimum and maximum Unix nanosecond timestamp,
 respectively.
 
+The `nonIndexedLabels` section stores non-repeated strings. It is used to store label names and label values from
+[non-indexed labels]({{< relref "./labels/non-indexed-labels" >}}).
+Note that the labels strings and lengths within the `nonIndexedLabels` section are stored compressed using Gzip.
```
sandeepsukhani, line 67: The compression format is actually configurable using `ingester.chunk-encoding` and defaults to gzip.

docs/sources/reference/api.md
```diff
@@ -592,6 +592,16 @@ JSON post body can be sent in the following format:
 }
 ```
 
+You can optionally attach [non-indexed labels]({{< relref "../get-started/labels/non-indexed-labels" >}}) to each log line by adding a JSON object to the end of the log line array.
+The JSON object must be a valid JSON object with string keys and string values.
+The JSON object must be the last element in the log line array. Here is an example of a log line with some non-indexed labels attached:
```
sandeepsukhani, line 597: I think we should not say `The JSON object must be the last element in the log line array.` since we will have to update the documentation again if we add a new field in log entry. I think we can say something like this 
```suggestion
The JSON object must be set immediately after the log line. Here is an example of a log entry with some non-indexed labels attached:
```

sandeepsukhani (commented):
Left just 2 minor nits. It looks good to me.

docs/sources/get-started/labels/bp-labels.md
JStickler: I'll go with your rewording (not sure what the etiquette is here for committing suggestions, but I'm assuming that's your role as the PR owner).

docs/sources/get-started/labels/non-indexed-labels.md
JStickler: That's what I thought you probably meant, but I'm not sure everyone would have understood.  The rewording is much clearer.

docs/sources/clients/promtail/stages/non_indexed_labels.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+Given the following log line:
+
+```
+{"log":"log message\n","stream":"stderr","traceID":"0242ac120002",time":"2019-04-30T02:12:41.8443515Z"}
+```
+
+The first stage would extract `stream` with a value of `stderr` and `traceID` with a value of `0242ac120002` into the
+extracted map of labels. The labels stage would turn that `stream` and `stderr` key-value pair into a label.
```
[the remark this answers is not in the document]
salvacorts: I think `data set` may not be the best term to use. I pretty much copy pasted this one form other stages and modified it. I think `map of labels` make sense.

I think we should rather state that the `labels` stage adds the labels to the stream labels, and the `non_indexed_labels` adds the labels to the non-indexed labels set of the log line. Wdyt?

 ```suggestion
The first stage would extract `stream` with a value of `stderr` and `traceID` with a value of `0242ac120002` into the
extracted map of labels. The `labels` stage would turn that `stream` and `stderr` key-value pair into a stream label.
```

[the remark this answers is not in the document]
Comment (vlad-diachenko) [reply docs/sources/clients/promtail/stages/non_indexed_labels.md:43]: in docs, we mostly use the phrase `extracted data` to separate it from labels and to highlight that it's just extracted data and not labels yet. 

https://grafana.com/docs/loki/latest/clients/promtail/stages/logfmt/#:~:text=The%20following%20key%2Dvalue%20pairs%20would%20be%20created%20in%20the%20set%20of%20extracted%20data%3A

https://grafana.com/docs/loki/latest/clients/promtail/stages/json/#:~:text=key%20in%20the-,extracted%20data,-while%20the%20expression

https://grafana.com/docs/loki/latest/clients/promtail/stages/json/#:~:text=those%20types.%20The-,extracted%20data,-can%20hold%20non

https://grafana.com/docs/loki/latest/clients/promtail/stages/json/#:~:text=those%20types.%20The-,extracted%20data,-can%20hold%20non

wdyt?

File: docs/sources/get-started/labels/non-indexed-labels.md [hunk 11]
```diff
@@ -0,0 +1,50 @@
...
+trace IDs, user IDs, and any other label that is often used in queries but has high cardinality and is expensive
+to extract at query time.
+
+## Attaching non-indexed labels to log lines
+
+You have the option to attach non-indexed labels to log lines in the push payload along with each log line and the timestamp.
+For more information on how to push logs to Loki via the HTTP endpoint, refer to the [HTTP API documentation]({{< relref "../../reference/api#push-log-entries-to-loki" >}}).
+
+Alternatively, you can use the Grafana Agent or Promtail to extract and attach non-indexed labels to your log lines.
+See the [Promtail configuration documentation]({{< relref "../../clients/promtail/configuration" >}}) for more information.
```
vlad-diachenko, line 25:
```suggestion
See the [Promtail: Non-indexed labels stage]({{< relref "../../clients/promtail/stages/non-indexed-labels" >}}) for more information.
```

vlad-diachenko: i believe we need to point the user to the concreate page with the new stage. it will be more useful for him

docs/sources/get-started/labels/non-indexed-labels.md
```diff
@@ -0,0 +1,50 @@
...
+{job="example"} | trace_id="0242ac120002" | user_id="superUser123"
+```
+
+Note that since non-indexed labels are extracted automatically to the results labels, some metric queries might return 
+an error like `maximum of series (50000) reached for a single query`. You can use the [Keep]({{< relref "../../query/log_queries#keep-labels-expression" >}}) and [Drop]({{< relref "../../query/log_queries#drop-labels-expression" >}}) stages to filter out labels that you don't need.
+For example:
+
+```logql
+{job="example"} | drop trace_id user_id
+```
```
vlad-diachenko, line 50:
```suggestion
```logql
count_over_time({job="example"} | keep trace_id | trace_id="0242ac120002" [5m])
```
```

**Review by vlad-diachenko — APPROVED** [review]
LGTM

File: docs/sources/get-started/labels/non-indexed-labels.md (hunk 12 above)
Comment (salvacorts) [reply docs/sources/get-started/labels/non-indexed-labels.md:50]: Good catch!

File: docs/sources/clients/promtail/stages/non_indexed_labels.md (hunk 10 above)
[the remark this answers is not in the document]
Comment (salvacorts) [reply docs/sources/clients/promtail/stages/non_indexed_labels.md:43]: Ok, if we are using that already, sounds good to me!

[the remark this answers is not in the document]
Comment (salvacorts) [reply docs/sources/clients/promtail/stages/non_indexed_labels.md:43]: Ok, if we are using that already, sounds good to me!
```suggestion
The first stage would extract `stream` with a value of `stderr` and `traceID` with a value of `0242ac120002` into
the extracted data set. The `labels` stage would turn that `stream` and `stderr` key-value pair into a stream label.
```

docs/sources/get-started/labels/non-indexed-labels.md
```diff
@@ -0,0 +1,50 @@
...
+{job="example"} | trace_id="0242ac120002"`
+```
+
+Of course, you can filter by multiple non-indexed labels at the same time:
+
+```logql
+{job="example"} | trace_id="0242ac120002" | user_id="superUser123"
+```
+
+Note that since non-indexed labels are extracted automatically to the results labels, some metric queries might return
```
gouthamve, line 44: If `traceid` is driving the number of series. For example, if there are 60000 traceids, how does `keep` stage help?

docs/sources/reference/api.md
```diff
@@ -592,6 +592,16 @@ JSON post body can be sent in the following format:
 }
 ```
 
+You can optionally attach [non-indexed labels]({{< relref "../get-started/labels/non-indexed-labels" >}}) to each log line by adding a JSON object to the end of the log line array.
+The JSON object must be a valid JSON object with string keys and string values.
+The JSON object must be set immediately after the log line. Here is an example of a log entry with some non-indexed labels attached:
```
gouthamve, line 597: Can we clarify that the JSON shouldn't have nested keys?

gouthamve (commented):
Looks great! Left some minor improvements.

docs/sources/reference/api.md
salvacorts: Modified with:
_The JSON object must be a valid JSON object with string keys and string values. **The JSON object should not contain any nested object**._

docs/sources/get-started/labels/non-indexed-labels.md
salvacorts: _Keep_ would help with filtering out high-cardinality labels similarly to _drop_. The example I added here is not illustrating this correctly. I've modified it by:
```
count_over_time({job="example"} | trace_id="0242ac120002" | keep job  [5m])
```

JStickler (approved):
[Docs team] LGTM

PR #10096, opened 2023-07-28
Storage: Allow setting a constant prefix for all created keys

### LINKED ISSUE #5889: allow to configure s3 subpath (bucket prefix configuration)
Is your feature request related to a problem? Please describe.

Currently a full S3 bucket is required for Loki - however, in some situations a bucket may need to be shared with other apps (one example would be aws lightsail where there's a fixed cost per bucket).

Describe the solution you'd like

Ability to specify a folder where all Loki data will be stored in a bucket

I've raised a similar issue with Mimir: https://github.com/grafana/mimir/issues/1682

Thank you.

aschleck:

Adds a new option under the aws stanza named key_prefix. This is useful if you have many Loki installations and do not want to create a million buckets. This is different than `compactor.shared_store_key_prefix` because it also affects eg the `loki_cluster_seed.json` file.

Fixes #5889

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


salvacorts:
@aschleck Thank you for your PR! We really appreciate it.

I think being able to set a prefix may make sense for other storages than just S3. IMO, if we want to implement this, we should support it for all (or most) storage backends. Would you be open to giving that a try? Happy to change my opinion tho :)

Here's a PR for Mimir doing something similar. From what I can see, looks like Mimir only support alpha-numeric prefixes, not path-like prefixes (e.g. `some/prefix`)
https://github.com/grafana/mimir/pull/1686

docs/sources/configure/_index.md
```diff
@@ -4141,6 +4141,10 @@ dynamodb:
 # CLI flag: -s3.buckets
 [bucketnames: <string> | default = ""]
 
+# Sets a constant prefix for all keys inserted in S3. Example: loki/
+# CLI flag: -s3.key-prefix
+[key_prefix: <string> | default = ""]
```
salvacorts, line 4146: I would rename this to `object_prefix`

salvacorts:
To add to my comment above, looks like (similarly to mimir), we already have a _prefixed_ client:
https://github.com/grafana/loki/blob/main/pkg/storage/stores/indexshipper/storage/prefixed_object_client.go#L16-L18
(It's currently in the _storage_ pkg, we may need/want to move it to the _client_ pkg along with `object_client.go`)

I think we can take the same approach as Mimir and use it when creating the object client as they do here:
https://github.com/grafana/mimir/pull/1686/files#diff-a17652466a6d455977112abb7636d8ded2f5a2e66fd8727d9e1da6759ef3a886R143-R145

aschleck:
Thank you Salva for looking at this! To make sure I understand, the request is:
* move the prefixed_object_client.go into pkg/storage
* wrap the clients created in https://github.com/grafana/loki/blob/174ef4472c358038e290f4f55a91447876aa869e/pkg/storage/factory.go#L578 with the prefixed object client
* add top level key `object_prefix` (seems like Mimir uses storage_prefix? which do you prefer?) to https://grafana.com/docs/loki/latest/configuration/#storage_config
Is that correct?

aschleck:
I can't figure out why this integration test is failing. I know it's something in this PR, it passes on the parent commit, but I'm not seeing the issue. Current theory is maybe the compactor is stuck but no idea why.

jcdauchy-moodys:
Thanks, I am currently testing on a 2.8.3 branch + your changes, it builds and run.

aschleck:
@salvacorts do you know why the integration test is failing?

cstyan:
Not sure how this is intended to work, but if it circumvents the schema code and stores all data under a single prefix then this is a giant potential footgun for s3 users that should come with documentation and a warning.

aschleck:
@cstyan thank you for thinking about this! My expectation is that changing this would be equivalent to changing the bucket name in the S3 config, so it's just as much as a footgun to the schema stuff as switching the bucket option would be. Does that make sense? I'm certainly happy to make the documentation clearer on that front.

In my original PR this was actually an S3-specific option and lived at the same level as the bucket option, which in light of your point may actually be a better solution. It also passed the CI tests so I kind of like reverting back to it. What do you think of the approach in https://github.com/aschleck/loki/commit/8a71b0e79a6387edd836e239d0178171ba29f178?

cstyan:
@aschleck Would you mind elaborating as to your use case for wanting to have a constant prefix? 

Am I misunderstanding that this would mean all objects are stored under just a single prefix, such as `bucket-name/<chunk id>` instead of using the schema config to store things like `bucket-name/<tenant-id>/<stream-fingerprint>/<chunk-id>`.

aschleck:
@cstyan I am not an expert on Loki's codebase, so please correct me if my change is not doing what I intend. I am *not* trying to force everything into `bucket-name/<chunk id>`. I really appreciate you looking at this PR.

My particular problem is that I have a ton of Kubernetes clusters, and I want to run a separate Loki instance in each of them. However given there are only a finite number of AWS S3 bucket limits, I want to make all of these different Loki instances share the same bucket. So my goal is to make objects write into `bucket-name/<configurable prefix, eg cluster name>/<tenant-id>/<stream-fingerprint>/<chunk-id>`. Does this make sense?

aschleck:
I am trying to do for Loki what https://github.com/grafana/mimir/pull/1686 did for Mimir and https://github.com/thanos-io/thanos/pull/5337 did for Thanos.

slim-bean:
>Not sure how this is intended to work, but if it circumvents the schema code and stores all data under a single prefix then this is a giant potential footgun for s3 users that should come with documentation and a warning.

I believe the prefix is everything up to the final slash

Prefix: `bucket-name/<configurable prefix, eg cluster name>/<tenant-id>/<stream-fingerprint>/`

I'm not thinking of any reason at the moment that adding another static path to the prefix would cause any problems.

docs/sources/configure/_index.md
```diff
@@ -1894,6 +1894,10 @@ hedging:
 # CLI flag: -store.index-cache-validity
 [index_cache_validity: <duration> | default = 5m]
 
+# Sets a constant prefix for all keys inserted in S3. Example: loki/
```
slim-bean, line 1897:
```suggestion
# Sets a constant prefix for all keys inserted into object storage. Example: loki/
```

change this not to be S3 specifc

pkg/storage/factory.go
```diff
@@ -575,8 +577,18 @@ func (c *ClientMetrics) Unregister() {
...
 
-// NewObjectClient makes a new StorageClient of the desired types.
+// NewObjectClient makes a new StorageClient with the prefix in the front.
 func NewObjectClient(name string, cfg Config, clientMetrics ClientMetrics) (client.ObjectClient, error) {
+	actual, err := internalNewObjectClient(name, cfg, clientMetrics)
+	if err != nil {
+		return nil, err
+	}
+
+	return client.NewPrefixedObjectClient(actual, cfg.ObjectPrefix), nil
```
slim-bean, line 587: I don't think we should always wrap the object client with the Prefixed object client.

The prefixed object client does string concatenation on almost every API request, this will incur more allocations and a performance penalty that I don't think we should apply in all cases. 

(Note: I'm not sure that allocation an empty string to a string isn't optimized away? so I guess if we could prove that in the empty string case there isn't any penalty we could leave this as is)

docs/sources/configure/_index.md
aschleck: Done

pkg/storage/factory.go
aschleck: Done, thank you for the suggestion!

aschleck:
Hi @slim-bean, thank you for the review! I think I've addressed your comments. Please let me know if this looks good.

ningyougang:
I like this feature

sjoukedv:
This would be great so that I can also put the <tenant-id> logs under a `prefix` just like for Tempo.

slim-bean:
My only last question is:  I think we either need

* validation that fails if there is no trailing slash on the prefix
* automatically add the trailing slash if it's not present.

If a user configured a prefix without the slash that would lead to munging the prefix with their tenant ID which could potentially cause problems.

I'm struggling to pick between these two approaches however, though I guess I slightly lean towards the automatic option as it gives a bit nicer user experience. (If the validation just fails and says "add a trailing slash" then we might as well just do it for them)

aschleck:
Makes sense @slim-bean and @sjoukedv! Added the automatic `/` concatenation.

ningyougang:
Btw, does the `table_manager's retention` feature works well for the `$bucketName/$prefix`?

ningyougang:
@aschleck can you rebase?

ningyougang:
@aschleck

Btw, i downloaded your branch to my local, and build the image with below command
```
docker build -f loki-build-image/Dockerfile -t loki:2.7.1-key-prefix loki-build-image/
```
the image is built as below
```
$ docker images|grep loki
loki                                     2.7.1-key-prefix       0a8c562db300        28 minutes ago      2.27GB
```
But when restart with the new image `loki:2.7.1-key-prefix`, the loki pod started failed
```
$ kubectl logs loki-0 -n my-namespace
stat: cannot statx '/src/loki': No such file or directory
```
Has any idea?

ningyougang:
@aschleck i downloaded this PR to my local, and build the custom loki image with below command
```
revision=${this pr's latest commit id}
docker build -f cmd/loki/Dockerfile -t loki:key-prefix . --pull=true \
--label org.opencontainers.image.revision=${revision} \
--label org.opencontainers.image.source=${my local loki repo} \
--label org.opencontainers.image.url=${my local loki repo} 
```
And changed loki pod using above custom loki image, changed our loki configuration as below
```
    schema_config:
      configs:
      - from: "2022-11-20"
        store: boltdb-shipper
        object_store: s3
        schema: v11
        index:
          period: 24h
          prefix: index_
        chunks:
          period: 24h
          prefix: chunks_
    server:
      http_listen_port: 3100
      log_level: warn
    storage_config:
      object_prefix: /test-nyg
      boltdb_shipper:
        active_index_directory: /data/loki/boltdb-shipper-active
        cache_location: /data/loki/boltdb-shipper-cache
        cache_ttl: 24h
        shared_store: s3
      aws:
        s3: s3://test-username:test-password@${s3Url}/test-bucket
        s3forcepathstyle: true
```
then, access some page, we can see, the logs were stored to `s3 bucket` finally through loki.
After restarted loki pod and queried loki logs with `http://$lokiSvc:3100/loki/api/v1/query_range`, then, can fetch the log successfully too.
(At this point, it seems works well due to the s3 bucket's data is synced to loki and can be queried by /api/v1/query_range API)

But i did more test.

After wait some time(e.g. 10.minutes),  and query with  `http://$lokiSvc:3100/loki/api/v1/query_range` again, seems can't fetch the log any more.

And if i restarted the loki pod again, same issue happens, can fetch the log successfully with `/api/v1/query_range` initally, but after wait some time(e.g. 10.minutes),  can't fetch the log any more.

@aschleck did you meet this issue?

aschleck:
Thank you for testing this @ningyougang! I just rebased on top of main for you.

Can you confirm that your test behaves differently if you build from a container from the main branch? I want to make sure the config and your method for testing this works on main, so that we know if there's a behavior change that the only thing different is my PR.

ningyougang:
@aschleck 

I build another new image `loki: 2.7.1-main` with https://github.com/grafana/loki/ repo's main branch's latest commit id.
image
And changed the loki pod with new image  `loki: 2.7.1-main` and removed the `object_prefix: /test-nyg`,
after restarted loki pod and wait about 10 minutes, can fetch the log through `/api/v1/query_range` API successfully.
(It is good for grafana/loki's main branch)

And i tested again for your new rebased branch: `key-prefix` in my local.
the same issue happened, e.g.
```
/data/loki/boltdb-shipper-cache/index_19599 $ ls
fake                                   loki-0-1693356663718529500-1693356792

# after wait about 10.minutes

/data/loki/boltdb-shipper-cache/index_19599 $ ls  
fake
```
Restarted a new loki pod and after wait about 10.minutes, can not fetch the log through `/api/v1/query_range` API

ningyougang:
@aschleck i read this PR's source code, seems should use `object_prefix: test-nyg/` rather than `object_prefix: /test-nyg`
And i tested `object_prefix: test-nyg/`, now `data lost issue` does not happen any more after restart a new loki pod.

aschleck:
Oh! That's super interesting, thank you for following up. I would never have thought about that case so appreciate you finding it.

What do you think is the best mitigation? Should we error out if there's a leading / or silently strip it? Personally I think stripping it makes sense. @ningyougang @slim-bean

aschleck:
Changed the logic to strip leading `/`s. @slim-bean what do you think?

pkg/storage/factory.go
```diff
@@ -603,8 +605,23 @@ func (c *ClientMetrics) Unregister() {
...
 func NewObjectClient(name string, cfg Config, clientMetrics ClientMetrics) (client.ObjectClient, error) {
+	actual, err := internalNewObjectClient(name, cfg, clientMetrics)
+	if err != nil {
+		return nil, err
+	}
+
+	if cfg.ObjectPrefix == "" {
+		return actual, nil
+	} else {
+		prefix := strings.Trim(cfg.ObjectPrefix, "/") + "/"
```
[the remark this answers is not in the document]
slim-bean: ah, ignore this! I was mistaken on the behavior of strings.Trim!

aschleck:
Thank you @slim-bean! Do you think you can approve this PR and merge it? What is left to do here?

aschleck:
@slim-bean @salvacorts @JStickler can we get this PR approved and merged?

sandeepsukhani (commented):
nice work @aschleck!
The changes look good to me. I think we need to also make the changes in [the code initializing storage for ruler](https://github.com/grafana/loki/blob/2b6bdc2a29de8d94d78d1c735104f124436a92e5/integration/loki_micro_services_delete_test.go#L28)?

chaudum:
I think we need to add some documentation around what the caveats are when you use a single storage bucket with multiple Loki installations under different prefixes.
The main problem is that rate limits on the bucket caused by one Loki installation affects all other Loki installations that use the same cluster.

chaudum:
An area where the global prefix could cause issues is the caching object client of the index shipper, because it splits object keys by delimiter and checks the length of the resulting parts. 

https://github.com/grafana/loki/blob/main/pkg/storage/stores/indexshipper/storage/cached_client.go#L82-L120

Maybe you could add a test case with a configured global prefix to see how it behaves.

aschleck:
I rebase this PR. CI says "Build is blocked, please, contact repo admin in order to proceed", I'm not sure why but I don't think it's my code.

> An area where the global prefix could cause issues is the caching object client of the index shipper, because it splits object keys by delimiter and checks the length of the resulting parts.
> 
> https://github.com/grafana/loki/blob/main/pkg/storage/stores/indexshipper/storage/cached_client.go#L82-L120
> 
> Maybe you could add a test case with a configured global prefix to see how it behaves.

I wouldn't expect this to cause any problems. The prefix is internal to the prefixed client so this caching client shouldn't see it. I added this test case just in case and it passes.

> I think we need to add some documentation around what the caveats are when you use a single storage bucket with multiple Loki installations under different prefixes. The main problem is that rate limits on the bucket caused by one Loki installation affects all other Loki installations that use the same cluster.

Most users aren't going to hit the S3 concurrency limits where this matters, so I'm not quite sure where you envision putting this. I would imagine it goes in a "Loki tuning guide" but I don't think one exists. This feels like something that might belong in a follow-up PR?

JStickler:
>CI says "Build is blocked, please, contact repo admin in order to proceed", I'm not sure why but I don't think it's my code.

@aschleck Our security team has added a layer of protection to the repo so that any PRs coming from a fork need to have their CI builds approved by an admin every time CI is run.  It has slowed things down considerably, and we're trying to find a different solution to the security problem.  In the meantime, the team has to manually trigger CI builds.  I triggered one just now, before I saw that your branch has conflicts that need to be resolved.

aschleck:
Ahhh you folks are prolific, I just rebased it again. Thank you for triggering the CI build!

slim-bean (approved):
Thank you everyone for all the involvement here and huge thanks to @aschleck for your patience and persistence with this change ❤️ 🎉

PR #10099, opened 2023-07-28
Logging improvements for volume requests

MasslessParticle:
I noticed we were logging volume requests as metric requests in `metrics.go`. This PR logs volume requests on their own with relevant data.

### DIFF
--- a/pkg/logql/metrics.go
+++ b/pkg/logql/metrics.go
@@ -28,6 +28,7 @@ const (
 	QueryTypeLimited = "limited"
 	QueryTypeLabels  = "labels"
 	QueryTypeSeries  = "series"
+	QueryTypeVolume  = "volume"
 
 	latencyTypeSlow = "slow"
 	latencyTypeFast = "fast"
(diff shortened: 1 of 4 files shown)


pkg/logql/metrics.go
```diff
@@ -288,6 +289,52 @@ func RecordSeriesQueryMetrics(
...
+		"status", status,
+		"splits", stats.Summary.Splits,
+		"total_entries", stats.Summary.TotalEntriesReturned,
+	)
+
+	sharded := strconv.FormatBool(false)
+	if stats.Summary.Shards > 1 {
+		sharded = strconv.FormatBool(true)
+	}
+	bytesPerSecond.WithLabelValues(status, queryType, "", latencyType, sharded).
```
trevorwhitney, line 326: can we add `		"throughput", strings.Replace(humanize.Bytes(uint64(stats.Summary.BytesProcessedPerSecond)), " ", "", 1),` to the log line as well?

trevorwhitney: and how about `total_bytes` too? I haven't dug in to if we're tacking these correctly, but if not we should remove the metric, and if so we should add the log line.

trevorwhitney (approved):
can we add `throughput` too?

MasslessParticle:
I removed irrelevant metrics, instead. Those are pretty specific to queries of loglines

PR #10100, opened 2023-07-28
Add logcli default client test

trevorwhitney:
This adds a test for the default client in `logcli`. Currently it only covers happy path for `volume` and `volume_range`. This was per request of @liguozhong resulting from a bug he found. @liguozhong could you add the unhappy path to recreate the bug? I'm still not sure I completely understand what it is. Thanks!

### DIFF
--- a/pkg/logcli/client/client_test.go
+++ b/pkg/logcli/client/client_test.go
@@ -13,7 +13,6 @@ import (
 	"github.com/grafana/loki/pkg/logcli/volume"
 	"github.com/grafana/loki/pkg/loghttp"
 	"github.com/grafana/loki/pkg/logproto"
-	"github.com/grafana/loki/pkg/logqlmodel/stats"
 	"github.com/grafana/loki/pkg/querier/queryrange"
 	"github.com/grafana/loki/pkg/querier/queryrange/queryrangebase"
 	"github.com/prometheus/common/model"
@@ -181,7 +180,7 @@ type mockLokiHTTPServer struct {
 	server   *http.Server
 	tenantID string
 	now      time.Time
-	then      time.Time
+	then     time.Time
 }
 
 func NewMockLokiHTTPServer(t *testing.T, now, then time.Time) *mockLokiHTTPServer {
@@ -205,14 +204,15 @@ func (s *mockLokiHTTPServer) getTenantIDUnsafe() string {
 
 func (s *mockLokiHTTPServer) Run(t *testing.T) {
 	var mux http.ServeMux
-	mux.HandleFunc("/loki/api/v1/index/volume", func(w http.ResponseWriter, request *http.Request) {
-		labels := labels.Labels{
-			{
-				Name:  "foo",
-				Value: "bar",
-			},
-		}
+	labels := labels.Labels{
+		{
+			Name:  "foo",
+			Value: "bar",
+		},
+	}
+	codec := queryrange.Codec{}
 
+	mux.HandleFunc("/loki/api/v1/index/volume", func(w http.ResponseWriter, request *http.Request) {
 		volume := queryrange.LokiPromResponse{
 			Response: &queryrangebase.PrometheusResponse{
 				Status: loghttp.QueryStatusSuccess,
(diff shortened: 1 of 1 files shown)


pkg/logcli/client/client_test.go
```diff
@@ -78,3 +91,198 @@ func Test_getHTTPRequestHeader(t *testing.T) {
...
+	labels := labels.Labels{
+		{
+			Name:  "foo",
+			Value: "bar",
+		},
+	}
+	codec := queryrange.Codec{}
+
+	mux.HandleFunc("/loki/api/v1/index/volume", func(w http.ResponseWriter, request *http.Request) {
+		volume := queryrange.LokiPromResponse{
```
liguozhong, line 216: Very detailed test case, thanks your help. My only doubt here is the return value type of this test.

The return value type here should be logproto.VolumeResponse instead of  queryrange.LokiPromResponse, because the code of http.go return logproto.VolumeResponse, and the test construction in http_test.go is also logproto.VolumeResponse

https://github.com/grafana/loki/blob/main/pkg/querier/http.go#L489
```go
        if resp == nil { // Some stores don't implement this
		resp = &logproto.VolumeResponse{Volumes: []logproto.Volume{}}
	}
	if err := queryrange.WriteResponse(r, nil, resp, w); err != nil {
		serverutil.WriteError(err, w)
	}
```

https://github.com/grafana/loki/blob/main/pkg/querier/http_test.go#L386
```go
	ret := &logproto.VolumeResponse{
		Volumes: []logproto.Volume{
			{Name: `{foo="bar"}`, Volume: 38},
		},
	}
```

https://github.com/grafana/loki/blob/main/pkg/util/marshal/marshal.go#L132
```go
func WriteVolumeResponseJSON(r *logproto.VolumeResponse, w io.Writer) error {
	s := jsoniter.ConfigFastest.BorrowStream(w)
	defer jsoniter.ConfigFastest.ReturnStream(s)
	s.WriteVal(r)
	s.WriteRaw("\n")
	return s.Flush()
}
```

liguozhong (commented):
great

liguozhong:
> This adds a test for the default client in `logcli`. Currently it only covers happy path for `volume` and `volume_range`. This was per request of @liguozhong resulting from a bug he found. @liguozhong could you add the unhappy path to recreate the bug? I'm still not sure I completely understand what it is. Thanks!

My test refers to the result of this test, and the test here also gets this json .`{"volumes":...}`

https://github.com/grafana/loki/blob/main/pkg/querier/http_test.go#L419C5-L422C41
```go
                                        require.Equal(t,
					`{"volumes":[{"name":"{foo=\"bar\"}","volume":38}]}`,
					strings.TrimSpace(w.Body.String()),
```

liguozhong:
> recreate the bug

`logproto.VolumeResponse` type is returned in my test, so this bug should appear. because would interpret an json `{"volumes":...} `, and DefaultClient.go 
   ```go
     var resp loghttp.QueryResponse
	if err := c.doRequest(path, params.Encode(), quiet, &resp); err != nil {
		return nil, err
	}
```
can only decode `{status:...,data:...}` response(loghttp.QueryResponse).



```go
mux.HandleFunc("/loki/api/v1/index/series_volume", func(w http.ResponseWriter, request *http.Request) {
		seriesVolume := logproto.VolumeResponse{
			Volumes: []logproto.Volume{
				{Name: `{foo="bar"}`, Volume: 1024},
				{Name: `{bar="baz"}`, Volume: 3350},
			},
			Limit: 5,
		}
		if err := marshal.WriteVolumeResponseJSON(&seriesVolume, w); err != nil {
			serverutil.WriteError(err, w)
			return
		}
	})
```

trevorwhitney:
@liguozhong sorry for the long delay here. 

I'm still not completely following. The `/loki/api/v1/index/volume` and `/loki/api/v1/index/volume_range` endpoints (their current names, changed since this PR was made) return a `queryrange.LokiPromResponse` (done via this [middleware](https://github.com/grafana/loki/blob/ce3ec024c93212749782b9ee8037af2539a0b865/pkg/querier/queryrange/series_volume.go#L116)). This is by design, as the UIs we're working on that use these endpoints can more easily visualize/plot the return data if it looks like a metrics query as they are just reusing core grafana components. These endpoints are live in our production clusters and are working as expected, so, unless I'm missing something, a change to the test to a different response type would be an incorrect test. Since the logcli test is mimicing hitting Loki from the outside, I don't think it should ever be handling a `logproto.VolumeResponse` since that is a datatype for internal component communication only (ie. queriers -> query frontend).

That being said, I still think this test is valuable to add as is, as we are missing this type of top-level test for logcli.

trevorwhitney:
it may still be worthwhile to add these tests but this PR is pretty out of date without any action on it for a while

PR #10101, opened 2023-07-28
Sharding optimizations

owen-d:
A few bugfixes and more sharding optimizations
* fix bug on `<aggr> by|without ()` groupings which removed the grouping while downstreaming
* shardable implementations for `max+min`, operation specific merge strategies which enable many more types of sharded requests, even when label-reduction is performed at edge.


MichelHollands (approved):

MasslessParticle (approved):
Nice! Looking forward to seeing the effect of this!

PR #10132, opened 2023-08-01
[s3] Add a metric that can be used to get s3 request rate by prefix

cstyan:
Without using AWS observability tools it can be difficult to get more usage information about Loki's requests to s3. In particular, it can be almost impossible to understand _where_ rate limiting is being applied by s3. 

The documentation for s3 states a rate limit of `3,500 PUT/COPY/POST/DELETE or 5,500 GET/HEAD requests per second per Amazon S3 prefix`, where a prefix is the unique path prior to a specific key. For example, `a_thing` and `some/path/another_thing` have two prefixes, empty string and `some/path/`. In our V12 schema, we prefix chunk objects with `user/stream fingerprint/<from:to:checksum>` where `<from:to:checksum>` is the actual chunk identifier. So this should mean that we should be able to get the previously stated rate limits per `user/stream FP` combination.

However, at the moment the only metric we have is the request latency (and because the metric is a histogram, we get the count as well) by method type. In this PR I'm adding a metric that tracks the # of requests by method and prefix. We may want to stick this behind a feature flag (maybe runtime reloadable?) and default it off, but I think a metric is a better option here given the sheer # of requests we do. If we wanted to track the same metric values via log lines we'd be adding significant load to our logging systems.

### DIFF
--- a/pkg/storage/chunk/client/aws/s3_storage_client.go
+++ b/pkg/storage/chunk/client/aws/s3_storage_client.go
@@ -55,6 +55,19 @@ var s3RequestDuration = instrument.NewHistogramCollector(prometheus.NewHistogram
 	Buckets:   []float64{.025, .05, .1, .25, .5, 1, 2},
 }, []string{"operation", "status_code"}))
 
+// s3 rate limits are applied per prefix, we should have a way to see the QPS per prefix to know if we're hitting the max
+// rate limit for that prefix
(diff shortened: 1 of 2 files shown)


MasslessParticle:
Worst case, this will create `(#queriers * #tenants) + (#ingesters * #tenants)` labels, right? I think this might have to be behind a feature flag.

owen-d (changes requested):
Unfortunately, this would create way too much cardinality in our metrics

PR #10140, opened 2023-08-02
Dynamic client-side throttling to avoid object storage rate-limits (GCS only)

dannykopping:
Across the various cloud providers' object storage services, there are different rate-limits implemented. Rate-limits can be imposed under multiple conditions, such as server-side scale up (ramping up from low volume to high, "local" limit), reaching some defined upper limit ("absolute" limit), etc.

We cannot know apriori when these rate-limits will be imposed, so we can't set up a client-side limiter to only allow a certain number of requests through per second. Additionally, that would require global coordination between queriers - which is difficult.

With the above constraints, I have instead taken inspiration from TCP's [congestion control algorithms](https://en.wikipedia.org/wiki/TCP_congestion_control). This PR implements [AIMD](https://en.wikipedia.org/wiki/Additive_increase/multiplicative_decrease) (Additive Increase, Multiplicative Decrease), which is used in the congestion _avoidance_ phase of congestion control. The default window size (object store requests per second) is 2000; in other words, we skip the "slow start" phase.

The controller uses the Go [`rate.Limiter`](https://pkg.go.dev/golang.org/x/time/rate), which implements the token-bucket algorithm.

To put it simply:
- every successful request widens the window (per second client rate-limit)
- every rate-limited response reduces the window size by a backoff factor (0.5 by default, so it will halve)
- when the limit has been reached, the querier will be delayed from making further requests until tokens are available

N/A

image

The above represents the maximum per-second request rate from our queriers across two very large production cells which occasionally encounter rate-limiting. I've set the default baseline (2000) around our 90th percentile, which seems sensible (i.e. there will be no back-pressure by default since the per-second rate-limit starts from a high enough level) and I set the default upper bound to 10000 which is higher than our current max rate.

This congestion control mechanism will introduce some back-pressure, which will cause queues to back up until the congestion clears. I think this will work far better than our current technique, which uses the client libraries' internal retry mechanisms. For example, by default (and we use these defaults) GCS' client library [uses exponential backoff with a start of 1s and a max of 30s, _plus_ some random jitter ](https://github.com/grafana/loki/blob/main/vendor/github.com/googleapis/gax-go/v2/call_option.go#L182-L205)🤯.

I've disabled the GCS' clients retry mechanism if congestion control is enabled.

The congestion control mechanism is pluggable; we might want to implement different control or retry mechanisms in the future. Hedging is included in this PR but _only as a noop_. I will implement this in a subsequent PR.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/storage/chunk/client/gcp/gcs_object_client.go
```diff
@@ -215,6 +227,86 @@ func (s *GCSObjectClient) IsObjectNotFoundErr(err error) bool {
...
+
+	return false
+}
+
+// IsStorageThrottledErr returns true if error means that object cannot be retrieved right now due to throttling.
+func (s *GCSObjectClient) IsStorageThrottledErr(err error) bool {
+	if gerr, ok := err.(*googleapi.Error); ok {
+		// https://cloud.google.com/storage/docs/retry-strategy
+		return gerr.Code == http.StatusTooManyRequests ||
+			(gerr.Code/100 == 5) // all 5xx errors are retryable
```
jeschkies, line 299: Shouldn't this return false according to the doc string above?

pkg/storage/chunk/client/gcp/gcs_object_client.go
```diff
@@ -215,6 +227,86 @@ func (s *GCSObjectClient) IsObjectNotFoundErr(err error) bool {
...
+	}
+
+	return false
+}
+
+// IsRetryableErr returns true if the request failed due to some retryable server-side scenario
+func (s *GCSObjectClient) IsRetryableErr(err error) bool {
+	return s.IsStorageThrottledErr(err) || s.IsStorageThrottledErr(err)
```
jeschkies, line 307:
```suggestion
	return s.IsStorageThrottledErr(err)
```

dannykopping:
This PR is still WIP @jeschkies, it doesn't need review yet - thanks

docs/sources/configure/_index.md
```diff
@@ -1914,6 +1914,55 @@ hedging:
...
+      [upper_bound: <int> | default = 10000]
+
+      # AIMD backoff factor when upstream service is throttled to decrease
+      # number of requests sent per second (default: 0.5).
+      # CLI flag: -store.congestion-control.strategy.aimd.backoff-factor
+      [backoff_factor: <float> | default = 0.5]
+
+  retry:
+    # Congestion control retry strategy to use (default: none, options:
+    # 'limited').
```
jeschkies, line 1945: What does limited mean?

docs/sources/configure/_index.md
```diff
@@ -1914,6 +1914,55 @@ hedging:
...
+  # CLI flag: -store.enabled
+  [enabled: <boolean> | default = false]
+
+  controller:
+    # Congestion control strategy to use (default: none, options: 'aimd').
+    # CLI flag: -store.congestion-control.strategy
+    [strategy: <string> | default = ""]
+
+    aimd:
+      # AIMD starting throughput window size: how many requests can be sent per
```
jeschkies, line 1928: I'd reference https://en.wikipedia.org/wiki/Additive_increase/multiplicative_decrease or something similar here. Or ad least spell out the name :slightly_smiling_face:

pkg/storage/chunk/client/congestion/congestion.go
```diff
@@ -0,0 +1,50 @@
...
+	"strings"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+
+	util_log "github.com/grafana/loki/pkg/util/log"
+)
+
+func NewController(cfg Config, metrics *Metrics) Controller {
+	logger := log.With(util_log.Logger, "component", "congestion_control")
```
jeschkies, line 13: The logger should be injected. We should avoid `util/log`.

pkg/storage/chunk/client/congestion/controller.go
```diff
@@ -0,0 +1,219 @@
...
+				a.metrics.retries.Add(1)
+			}
+
+			// apply back-pressure while rate-limit has been exceeded
+			for {
+				// using Reserve() is slower because it assumes a constant wait time as tokens are replenished, but in experimentation
+				// it's faster to sit in a hot loop and probe every so often if there are tokens available
+				if !a.limiter.Allow() {
+					delay := time.Millisecond * 10
+					time.Sleep(delay)
```
jeschkies, line 102: Usually I'm not a fan of hard coded waiting times or waiting at all. I saw there's `limiter.Wait()` but it seems to be using reserver so :man_shrugging:

pkg/storage/chunk/client/congestion/controller.go
```diff
@@ -0,0 +1,219 @@
...
+		func(attempt int) (io.ReadCloser, int64, error) {
+			a.metrics.requests.Add(1)
+
+			// in retry
+			if attempt > 0 {
+				a.metrics.retries.Add(1)
+			}
+
+			// apply back-pressure while rate-limit has been exceeded
+			for {
```
jeschkies, line 97: What do you think about
```suggestion
			for !a.limiter.Allow() {
```
instead? It saves the `continue` and `break`.

jeschkies (commented):
I need to go through the tests but the controller looks good. Please see my comments.

pkg/storage/chunk/client/congestion/congestion.go
dannykopping: Good call, I was lazy here

docs/sources/configure/_index.md
jeschkies: That was my other thought :slightly_smiling_face:

jeschkies (approved):
Nice! Will you address the todos in this PR or a follow up?

pkg/storage/chunk/client/congestion/retry.go
```diff
@@ -0,0 +1,54 @@
...
+
+		if err != nil {
+			if !isRetryable(err) {
+				return rc, sz, err
+			}
+
+			// TODO(dannyk): consider this more carefully
+			// only decrease rate-limit if error is retryable, otherwise all errors (context cancelled, dial errors, timeouts, etc)
+			// which may be mostly client-side would inappropriately reduce throughput
+			onError()
```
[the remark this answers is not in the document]
dannykopping: Sure, but I don't think it'd improve readability

pkg/storage/chunk/client/congestion/congestion_test.go
```diff
@@ -0,0 +1,69 @@
...
+		Retry: RetrierConfig{
+			Strategy: "limited",
+		},
+	}
+	ctrl := NewController(cfg, log.NewNopLogger(), NewMetrics(t.Name(), cfg))
+
+	require.IsType(t, &AIMDController{}, ctrl)
+	require.IsType(t, &LimitedRetrier{}, ctrl.getRetrier())
+	require.IsType(t, &NoopHedger{}, ctrl.getHedger())
+}
```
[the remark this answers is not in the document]
dannykopping: I find table tests to be quite hard to read; maybe I'll refactor this later but I'd prefer to leave it as is for now

dannykopping:
> Nice! Will you address the todos in this PR or a follow up?

Thanks for the review! I'll address them in a follow-up; this PR is already quite large.

PR #10142, opened 2023-08-02
Fix bugs in non-indexed-labels

salvacorts:

This PR fixes some bugs discovered while setting the default chunk format yo _V4_ and the default head format to _unordered with non-indexed labels_.

Note that this PR doesn't change the defaults, but fixes some bugs that would prevent us from changing them in the future.

The bugs it fixes are:

- Index out-of-range panic when symbols' labels are empty and we do a Lookup. **Test added**.
- When applying retention, new chunks wouldn't contain non-indexed labels (caused by https://github.com/grafana/loki/pull/10090). **Test added**.
- New chunks would fail to serialize non-indexed labels when applying retention because the head format would be the [dummy one (ordered)][1] [(used here via `c.headFmt`)][2]. **Cannot test unless we change the default chunk/head format** but tested manually.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

[1]: https://github.com/grafana/loki/blob/457f2e6ec0f626dfec17faa8ba424a04818168b8/pkg/chunkenc/memchunk.go#L375
[2]: https://github.com/grafana/loki/blob/24fd56683328514b58257972d6ca5baa9aeec51b/pkg/chunkenc/memchunk.go#L1088-L1095

### DIFF
--- a/pkg/chunkenc/dumb_chunk.go
+++ b/pkg/chunkenc/dumb_chunk.go
@@ -72,7 +72,7 @@ func (c *dumbChunk) Encoding() Encoding { return EncNone }
 
 // Returns an iterator that goes from _most_ recent to _least_ recent (ie,
 // backwards).
-func (c *dumbChunk) Iterator(_ context.Context, from, through time.Time, direction logproto.Direction, _ log.StreamPipeline) (iter.EntryIterator, error) {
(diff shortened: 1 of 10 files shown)


vlad-diachenko (approved):
Looks brilliant

PR #10146, opened 2023-08-02
Update chunk and head fmt to v4 (non-indexed labels)

salvacorts:

This PR bumps the default format of the chunks and the heads to support non-indexed labels by default.

Note that the chunk format V4 only supports the UnorderedWithMetadata head format, so we had to update many tests to use the correct head format.


salvacorts:
**NOTE:** we won't merge this PR

PR #10213, opened 2023-08-10
Pin `chunk` and `index` format to `schema` version.

kavirajk:
We pin all three `Chunk`, `HeadBlock` and `TSDB` Version to `schema` version in period config.

This is the following mapping (after being discussed with @owen-d and @sandeepsukhani )

`v12` (current existing schema) - ChunkFormatV3 (UnorderedHeadBlock) + TSDBv2
`v13` (introducing new schema) - ChunkFormatV4 (UnorderedWithNonIndexedLabelsHeadBlockFmt) + TSDBv3

Note the new schema `v13` supports the latest chunk and index format.

**NOTES for Reviewer**

1. General approach is we removed the idea of `index.LiveFormat`, `chunkenc.DefaultChunkFormat` and `chunkenc.DefaultHeadBlockFmt` and made following two changes. These variables were used before to tie chunk and tsdb formats specific to Loki versions. This PR remove that coupling and pin these formats to `schema` version instead.

    1. These variables were replaced with explicit chunk and index formats within those packages (and it's tests)
    2. If these variables were used outside it's own packages say by ingester, compactor, etc. Then we extract correct chunk and index versions from the `schema` config.

2. Add two methods to `periodConfig`. (1) `ChunkFormat()` returning chunk and head format tied to schema (2) `TSDBFormat()` returning tsdb format tied to schema.

2. Other ideas I thought of doing but didn't end up doing is make `ChunkFormat` and `IndexFormat` as separate type (rather than `byte` and `int` currently. Similar to `HeadBlockFmt` type). But didnt' do it eventually to keep the PR small and don't want to complicate with lots of changes.

4. Moved couple of test cases from `chunkenc` to `config` package, because the test case was actually testing methods on `schemaconfig` and it was creating cycling dependencies.


JStickler:
Are there any updates we need to make to the documentation when this change goes through?
We document the chunk format in https://grafana.com/docs/loki/next/get-started/architecture/.
1 - Is the current chunk format correct (IIRC there was a recent update to this section of the docs)
2 - Do we need to add the schema version number to the docs?

kavirajk:
>  Is the current chunk format correct (IIRC there was a recent update to this section of the docs)

I think you mean the non-index labels changes went into ChunkV4. I think yes. @sandeepsukhani @vlad-diachenko @salvacorts worked on this.

> Do we need to add the schema version number to the docs?

Yes. Schema v13 needs to be added. I will do that either in this PR or follow up PR.

pkg/chunkenc/memchunk.go
```diff
@@ -85,7 +83,7 @@ const (
 	UnorderedHeadBlockFmt
 	UnorderedWithNonIndexedLabelsHeadBlockFmt
 
-	DefaultHeadBlockFmt = UnorderedWithNonIndexedLabelsHeadBlockFmt
+	// DefaultHeadBlockFmt = UnorderedWithNonIndexedLabelsHeadBlockFmt
```
chaudum, line 86: Can be removed

pkg/ingester/checkpoint.go
```diff
@@ -103,7 +103,7 @@ func fromWireChunks(conf *Config, wireChunks []Chunk) ([]chunkDesc, error) {
 		// Always use Unordered headblocks during replay
 		// to ensure Loki can effectively replay an unordered-friendly
 		// WAL into a new configuration that disables unordered writes.
-		hbType := chunkenc.DefaultHeadBlockFmt
+		hbType := chunkenc.UnorderedWithNonIndexedLabelsHeadBlockFmt
```
chaudum, line 106: Has this the potential to break WAL recovery when the WAL spans across multiple periods with different schemas?

pkg/storage/config/schema_config.go
```diff
@@ -371,6 +373,42 @@ func (cfg *PeriodConfig) applyDefaults() {
...
+	}
+
+	switch {
+	case sver <= 12:
+		return chunkenc.ChunkFormatV3, chunkenc.UnorderedHeadBlockFmt, nil
+	case sver == 13:
+		fallthrough
+	default:
+		return chunkenc.ChunkFormatV4, chunkenc.UnorderedWithNonIndexedLabelsHeadBlockFmt, nil
+	}
```
chaudum, line 391: This logic partially duplicates `pkg/chunkenc/memchunk.go:416-425`

I think it would be great to have a single source of truth for the binding of chunk format to head block format.

pkg/storage/stores/indexshipper/compactor/retention/retention_test.go
```diff
@@ -523,6 +523,13 @@ func TestChunkRewriter(t *testing.T) {
 
 				for _, interval := range expectedChunks[i] {
 					for curr := interval.Start; curr <= interval.End; curr = curr.Add(time.Minute) {
+						// // Test ready to pass/fail when we change the default chunk and head format.
```
chaudum, line 526: Should this commented out test already pass? I thought there are no "default" chunk and head formats any more, since they are "encoded" into the schema version.

pkg/storage/stores/tsdb/index/index.go
```diff
@@ -61,9 +61,6 @@ const (
 	fingerprintInterval = 1 << 10
 
 	millisecondsInHour = int64(time.Hour / time.Millisecond)
-
-	// The format that will be written by this process
-	LiveFormat = FormatV3
```
chaudum, line 66: Could we keep the "live" format for tests? That way, we don't need to update index tests every time a new version is introduced.

salvacorts (commented):
This is looking great! Thank you Kavi, this will ease our lives a lot when implementing and rolling out new chunk formats 🚀 

I left out some minor comments.

pkg/chunkenc/memchunk_test.go
```diff
@@ -57,27 +57,27 @@ var (
...
 		},
 	}
 )
 
-const DefaultTestHeadBlockFmt = DefaultHeadBlockFmt
+const DefaultTestHeadBlockFmt = UnorderedWithNonIndexedLabelsHeadBlockFmt
 
 func TestBlocksInclusive(t *testing.T) {
-	chk := NewMemChunk(EncNone, DefaultTestHeadBlockFmt, testBlockSize, testTargetSize)
+	chk := NewMemChunk(ChunkFormatV3, EncNone, DefaultTestHeadBlockFmt, testBlockSize, testTargetSize)
```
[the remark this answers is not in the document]
chaudum: I like the idea of running tests in all formats.

pkg/storage/stores/tsdb/index/index.go
kavirajk: May be calling it as `latestFormat` instead (private to the package and not exported) may be better I think.

pkg/ingester/checkpoint.go
kavirajk: Good point. ideally I want to figure out this format from schema config. Will fix it.

pkg/storage/stores/tsdb/index/index.go
kavirajk: Hmm. while implementing it, realized even private `latestFormat` is not needed as all the tests were written to cover both `FormatV2` and `FormatV3` already.

Didn't see the need for it. So removed again. I guess we can add when the needs shows up in future.

pkg/storage/stores/indexshipper/compactor/retention/retention_test.go
kavirajk: Done.

pkg/storage/config/schema_config.go
kavirajk: Good idea. Done.

pkg/ingester/checkpoint.go
kavirajk: Fixed it.

pkg/chunkenc/memchunk.go
kavirajk: Done.

pkg/storage/stores/indexshipper/compactor/retention/retention_test.go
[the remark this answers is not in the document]
kavirajk: Done.

pkg/ingester/stream.go
```diff
@@ -594,7 +611,7 @@ func (s *stream) resetCounter() {
 
 func headBlockType(unorderedWrites bool) chunkenc.HeadBlockFmt {
 	if unorderedWrites {
-		return chunkenc.DefaultHeadBlockFmt
+		return chunkenc.UnorderedWithNonIndexedLabelsHeadBlockFmt
```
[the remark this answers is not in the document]
kavirajk: I like the idea. Only problem is once we export such variables outside the `chunkenc`, people use it instead of taking it from `schema` config.

Besides, particularly here this unordered head fmt should also depends on either `chunkv3 or chunkv4` (cannot use normal UnorderedBlockFmt with chunk4 for example). 

I handled it differently by exporting `ChunkHeadFormatFor(chunkfmt)` api from `chunkenc` package.

sandeepsukhani (commented):
Nice work Kavi! The changes look good to me, modulo Salva's suggestions which are yet to be resolved. While reviewing the code, I had similar concerns about having to update the tests outside of the `chunkenc` package to use the latest chunk format. I think we can put a `LatestPeriodConfig` or `RecommendedPeriodConfig` somewhere for the tests and use it for tests building chunks. The `chunkenc` package can then have most tests running against all supported chunk formats.

pkg/chunkenc/memchunk_test.go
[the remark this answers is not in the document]
kavirajk: Done.

pkg/storage/batch_test.go
```diff
@@ -65,6 +70,15 @@ func Test_batchIterSafeStart(t *testing.T) {
 }
 
 func Test_newLogBatchChunkIterator(t *testing.T) {
+	periodConfig := config.PeriodConfig{
```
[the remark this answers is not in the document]
kavirajk: Agree. Added recent set of schemas to test cases.

pkg/storage/config/schema_config.go
```diff
@@ -371,6 +373,42 @@ func (cfg *PeriodConfig) applyDefaults() {
...
+	sver, err := cfg.VersionAsInt()
+	if err != nil {
+		return 0, 0, fmt.Errorf("failed to get chunk format: %w", err)
+	}
+
+	switch {
+	case sver <= 12:
+		return chunkenc.ChunkFormatV3, chunkenc.UnorderedHeadBlockFmt, nil
+	case sver == 13:
+		fallthrough
```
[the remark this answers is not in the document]
kavirajk: Done. Make sense.

pkg/storage/config/schema_config.go
```diff
@@ -371,6 +373,42 @@ func (cfg *PeriodConfig) applyDefaults() {
...
+	sver, err := cfg.VersionAsInt()
+	if err != nil {
+		return 0, fmt.Errorf("failed to get index format: %w", err)
+	}
+
+	switch {
+	case sver <= 12:
+		return index.FormatV2, nil
+	case sver == 13:
+		fallthrough
```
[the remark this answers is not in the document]
kavirajk: Done.

pkg/storage/hack/main.go
```diff
@@ -91,7 +91,7 @@ func fillStore(cm storage.ClientMetrics) error {
 			labelsBuilder.Set(labels.MetricName, "logs")
 			metric := labelsBuilder.Labels()
 			fp := client.Fingerprint(lbs)
-			chunkEnc := chunkenc.NewMemChunk(chunkenc.EncLZ4_4M, chunkenc.DefaultHeadBlockFmt, 262144, 1572864)
+			chunkEnc := chunkenc.NewMemChunk(chunkenc.ChunkFormatV4, chunkenc.EncLZ4_4M, chunkenc.UnorderedWithNonIndexedLabelsHeadBlockFmt, 262144, 1572864)
```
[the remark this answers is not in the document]
kavirajk: I explained the problem of exposing such function before. Same applies here. I handled it differently (taking it from schema config) 👍

pkg/storage/lazy_chunk_test.go
```diff
@@ -14,17 +14,27 @@ import (
 	"github.com/grafana/loki/pkg/logproto"
 	"github.com/grafana/loki/pkg/logql/log"
 	"github.com/grafana/loki/pkg/storage/chunk"
+	"github.com/grafana/loki/pkg/storage/config"
 	"github.com/grafana/loki/pkg/util"
 )
 
 func TestLazyChunkIterator(t *testing.T) {
+	periodConfig := config.PeriodConfig{
```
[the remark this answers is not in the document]
kavirajk: Done.

pkg/storage/stores/tsdb/head_manager_test.go
```diff
@@ -385,6 +385,7 @@ func TestBuildLegacyWALs(t *testing.T) {
 	schemaCfg := config.SchemaConfig{
 		Configs: []config.PeriodConfig{
 			{
+				Schema:     "v11",
```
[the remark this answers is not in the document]
kavirajk: Done.

kavirajk:
> I think we can put a LatestPeriodConfig or RecommendedPeriodConfig somewhere for the tests and use it for tests building chunks. The chunkenc package can then have most tests running against all supported chunk formats.

Thanks for the review @sandeepsukhani. Agree.

Basically I tweaked tests inside `chunkenc` to run tests against all supported chunk formats and outside the `chunkenc` I tweaked tests to run against all recent schemas (v11, v12 and v13)

pkg/ingester/recovery.go
```diff
@@ -210,7 +210,7 @@ func (r *ingesterRecoverer) Close() {
 			s.unorderedWrites = isAllowed
 
 			if !isAllowed && old {
-				err := s.chunks[len(s.chunks)-1].chunk.ConvertHead(headBlockType(isAllowed))
+				err := s.chunks[len(s.chunks)-1].chunk.ConvertHead(headBlockType(s.chunkFormat, isAllowed))
```
sandeepsukhani, line 213: this code actually doesn't make sense to me because `isAllowed` here is always going to be false. Anyways, not going to block the PR on this.

sandeepsukhani, line 213: this code actually doesn't make sense to me because `isAllowed` here is always going to be false. Anyways, not going to block the PR on this.

sandeepsukhani (approved):
LGTM

kavirajk:
Will merge this once getting some reviews from @owen-d as well.

pkg/ingester/instance.go
```diff
@@ -108,6 +109,12 @@ type instance struct {
 	streamRateCalculator *StreamRateCalculator
 
 	writeFailures *writefailures.Manager
+
+	schemaconfig *config.SchemaConfig
+
+	// precomputed `Chunk` and `Head` formats parsed from schemaconfig
+	chunkfmt byte
```
owen-d, line 116: This should be detected when writing a chunk within the `stream`, not on the instance. The `instance` may be exist across schema boundaries. This leaves us with two choices:
1) set the format based on the min timestamp in a chunk when it's cut
2) set the format based on the max timestamp in a chunk when it's cut

I think (1) is the more conservative approach and thus more appropriate. We may write an older format to a newer schema bucket, but we won't write a newer format to an older bucket. To be exact, the chunk is written once then added to the index buckets for different days. I'd rather us have some v13-schema indices reference v12 compatible chunks than vice versa.

owen-d (commented):
One fix needed then lgtm

pkg/ingester/instance.go
kavirajk: Nice catch. thanks @owen-d for the explanation. 

In fact [my initial commit used "per-stream" chunk format selection](https://github.com/grafana/loki/pull/10213/commits/8b6787fcb8c7d75984be131ea31faeec184af5db#diff-5e8ea6dc696f5e7c429726e1d8c59705869c8a64e1793f399e6e6587b137e4bfL296-L302), but was using `model.Now()` instead of `min/max` timestamp. Makes more sense to use correct timestamp here.

Also agree with `min` timestamp.

PR #10242, opened 2023-08-14
Update chunk and head fmt to v4 (non-indexed labels)

salvacorts:

This PR updates the default chunk and heads format to v4 to support writing and reading non-indexed labels.


kavirajk (approved):
LGTM (don't have much context with v4 formats). But code looks good to me.

Also it's worth noting, we are trying to pin specific chunk/index formats to schema version.
https://github.com/grafana/loki/pull/10213

I guess I will update that PR once this is merged.

PR #10248, opened 2023-08-14
Return intersecting labels when aggregating volume by label

trevorwhitney:
This PR changes the behavior the `index/volume` endpoint to return intersecting labels when requesting volume aggregating by label. Previously it only returned labels that matched selectors in the query. The old behavior is still possible using the `targetLabels` argument.


MasslessParticle (approved):
LGTM.

I wonder if there's a way to generalize across the ingester and the index. The logic here is getting complex and it'd be nice if we didn't have to maintain it in two places

PR #10256, opened 2023-08-16
Promtail: Include geoip fields in the extracted map

### LINKED ISSUE #9076: Allow promtail GeoIP stage to extract data
The new GeoIP stage of Promtail has been just implemented with #3493 and it is great.

However, the GeoIP data flowing directly to labels seems problematic, because most of this data are high cardinality.
Moreover, it seems we can't push the label values to the logs or use them in other stages (tried with the `pack` filter and the `replace` filter as described in the [labeldrop](https://grafana.com/docs/loki/latest/clients/promtail/stages/labeldrop/) documentation, but it seems that those labels values are not reachable at this point of the pipeline).

Wouldn't it be better to pick it as "extracted data" instead of labels, so that users can choose whether they want to add them as label or as log output in later stages ?

Do not send the data to labels at all in this stage.
Only send the data to the "extracted data" map.

This would be more coherent with the stage being a "Parsing stage" according to the [documentation](https://grafana.com/docs/loki/latest/clients/promtail/stages/).

Moreover, there already exists stages that allow to pick extracted data and put it in labels.
This is easier to compose with.

Keep the current implementation (which sends data to labels), but also send it to the extracted data map.

This seems a bit redundant and would ask users to frequently drop most labels with an extra `labeldrop` stage as they are high cardinality.

**Additional context**
I already tried to patch promtail locally (the way described in this feature request) and it seems to work flawlessly.
I'm planning to make multiple clean PRs (because my current patch also add new things and fixes others) ASAP.

In case someone wants to fix this before I do, here is an extract of the full patch that shows the idea of how to implement this:

```diff
-func (g *geoIPStage) populateLabelsWithASNData(labels model.LabelSet, record *geoip2.ASN) {
+func (g *geoIPStage) populateExtractedWithASNData(extracted map[string]interface{}, record *geoip2.ASN) {
 	autonomousSystemNumber := record.AutonomousSystemNumber
 	autonomousSystemOrganization := record.AutonomousSystemOrganization
 	if autonomousSystemNumber != 0 {
-		labels[model.LabelName("geoip_autonomous_system_number")] = model.LabelValue(fmt.Sprint(autonomousSystemNumber))
+		extracted["geoip_autonomous_system_number"] = autonomousSystemNumber
 	}
 	if autonomousSystemOrganization != "" {
-		labels[model.LabelName("geoip_autonomous_system_organization")] = model.LabelValue(autonomousSystemOrganization)
+		extracted["geoip_autonomous_system_organization"] = autonomousSystemOrganization
 	}
 }
```

esev:

The geoip parsing stage currently only populates the labelset with fields. This makes it difficult to use as fields like `geoip_location_latitude`/`geoip_location_longitude` and `geoip_autonomous_system_number` due to them having high cardinality. In addition, stages that operate on the extracted map cannot access these fields.

This PR adds the fields to the extracted map as well. That allows the fields to be usable in other stages, like `pack` & `template`, which can be used to store the fields in the message output.

Fixes #9076

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


JStickler:
>That allows the fields to be usable in other stages, like `pack` and `template` which can be used to store the fields in the message output.

I think it would be helpful to add an example of this ^.

esev:
> I think it would be helpful to add an example of this ^.

Thanks for the feedback. I've added two examples to the documentation.

JStickler (commented):
[docs team] The documentation LGTM, I'll leave it to engineering to approve the Go code.

cstyan:
Hello, thanks for your PR.

We're currently reevaluating promtails position as a project within Grafana Labs. Internally we're actually using the [Agent](https://github.com/grafana/agent) for both metrics and logs collection at this point. Potentially we can consider this a bugfix and move forward, I'll consult with the rest of the team.

longtomjr:
Good day. I just want to check if there have been any decisions made that impacts if this PR will be merged or not? I am running into cardinality issues with some of the log streams I am using with geoip, even after dropping most of the labels.

cstyan:
No update as of right now.

itsmejoeeey:
Hi there, any update on the status of this PR?

Fixing this seems like a vital bugfix to keep label low-cardinality when using the `geoip` parsing stage in promtail. Unless I am missing something and there is another way to use `geoip` safely @cstyan ?

Additionally (it was initially unclear to me), without this fix the issue is exhibited even when using the newer [Grafana Agent](https://github.com/grafana/agent) in static mode.

cstyan:
With the agent now becoming alloy 1.0 ([link](https://grafana.com/blog/2024/04/09/grafana-alloy-opentelemetry-collector-with-prometheus-pipelines/)) promtail is in a "feature complete" status, as I mentioned above. I can still check in on whether we'd include this as a bug fix, if the current state of the geoip stage results in the labels being unusable anyways due to cardinality then I think it's reasonable.

@grafana/agent-squad can you track a separate issue for confirming/fixing the same behaviour in the agent/alloy? thanks

kyrianae:
This PR fullfill a need that we have on log analysis
What can we do to help the process ?

cstyan:
We're okay with moving forward with this PR as a fix rather than a feature. IMO we probably should have some basic tests here, @esev I know this has been sitting here for a while please LMK if you have some time to add tests, otherwise I can try and find to add some myself.

imtipi:
> We're okay with moving forward with this PR as a fix rather than a feature. IMO we probably should have some basic tests here, @esev I know this has been sitting here for a while please LMK if you have some time to add tests, otherwise I can try and find to add some myself.

i kept using this fix since pr appeared,and no crash so far(although I know this isn't the professional test)

JStickler:
Promtail is end of life (EOL) as of  March 2, 2026.  No future support or updates will be provided. All future feature development will occur in [Grafana Alloy](https://github.com/grafana/alloy).

PR #10269, opened 2023-08-16
allows filter-only matchers in series endpoint

owen-d:
We already allow empty matchers `{}` as a special case in series queries to return all series. This PR also allows specifying only filter-style matchers. This is specifically helpful when a tenant is trying to understand all streams which do _not_ have a corresponding label.
```
$ logcli series '{does_not_exist=""}' --analyze-labels --org-id="fake"
2023/08/16 10:13:32 http://localhost:3100/loki/api/v1/series?end=1692206012889997000&match=%7Bdoes_not_exist%3D%22%22%7D&start=1692202412889997000
Total Streams:  6
Unique Labels:  2
```


MichelHollands (approved):
LGTM

salvacorts (approved):

Issue #10153, opened 2023-08-02
`expr.String()` escapes characters, which can result in much larger strings, which can result in panics


trevorwhitney:
**Describe the bug**
Given the query 
```
{cluster="foo", namespace="bar"} |=`msg=~".*?(GET|POST|PUT|DELETE|PATCH|HEAD|OPTIONS) /loki/api/(?i)(\d+[a-z]|[a-z]+\d)\w*/query_range"`
```

This becomes:
```
{cluster=\"foo\", namespace=\"bar\"} |= `msg=~".*?(GET|POST|PUT|DELETE|PATCH|HEAD|OPTIONS) /loki/api/(?i)(\\d+[a-z]|[a-z]+\\d)\\w*/query_range"`
```

If the original query is close to the query size limit of 5120 bytes, then the escaped query can exceed this limit. In the [`clone() function`](https://github.com/grafana/loki/blob/feba83a3ac34c3a63a6bef6e73ddf4ce00572462/pkg/logql/rangemapper.go#L499) we parse the result of `expr.String()`. If this string is bigger than 5120, we throw an error, resulting in a panic.
 
**To Reproduce**
1. Craft a metric query with special characters that is 5120 bytes
2. Execute that query
3. Loki will panic

**Expected behavior**
Loki should not panic

Segflow:
The panic path described here is gone as of v3.0.0. The range mapper no longer clones expressions by re-parsing expr.String() (#11595 switched it to a structural clone via syntax.MustClone), and the parser input limit was raised from 5,120 to 131,072 bytes in #11701.

PR #10277, opened 2023-08-16
Stopped String() from doubly escaping regex characters

akhilanarayanan:
When queries have regex, characters that needed escaping were escaped twice. Now they're only singly escaped. This is good because it reduces the character count so users can submit queries closer to the 5120 byte limit.

Sort of addresses #10153

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/syntax/parser_test.go
+++ b/pkg/logql/syntax/parser_test.go
@@ -3465,3 +3465,13 @@ func TestNoOpLabelToString(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, stages, 0)
 }
+
+// Tests that the regex part of expr doesn't get doubly escaped when we call expr.String()
+func TestParseSampleExpr_String(t *testing.T) {
+	query := `sum(rate({cluster="beep", namespace="boop"} | msg=~` + "`" + `.*?(GET|POST|PUT|DELETE|PATCH|HEAD|OPTIONS) /loki/api/(?i)(\d+[a-z]|[a-z]+\d)\w*/query_range` + "`" + `[1d]))`
+	expr, err := ParseSampleExpr(query)
+	require.NoError(t, err)
+	// need to change some backticks to " in order for query to be exactly equal to expr.String(), otherwise it's the same query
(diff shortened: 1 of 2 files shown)


pkg/logql/syntax/parser_test.go
```diff
@@ -3465,3 +3465,14 @@ func TestNoOpLabelToString(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, stages, 0)
 }
+
+// Tests that the regex part of expr doesn't get doubly escaped when we call expr.String()
+func TestParseSampleExpr_String(t *testing.T) {
+	query := `sum(rate({cluster="beep", namespace="boop"} | msg=~` + "`" + `.*?(GET|POST|PUT|DELETE|PATCH|HEAD|OPTIONS) /loki/api/(?i)(\d+[a-z]|[a-z]+\d)\w*/query_range` + "`" + `[1d]))`
+	require.True(t, len(query) < 5120)
```
MasslessParticle, line 3472: Why are you checking this length here?

pkg/logql/syntax/parser_test.go
```diff
@@ -3465,3 +3465,14 @@ func TestNoOpLabelToString(t *testing.T) {
...
+	expr, err := ParseSampleExpr(query)
+	require.NoError(t, err)
+	// need to change some backticks to "
```
MasslessParticle, line 3475: Is this comment relevant?

pkg/logql/syntax/parser_test.go
```diff
@@ -3465,3 +3465,14 @@ func TestNoOpLabelToString(t *testing.T) {
...
+	query = `sum(rate({cluster="beep", namespace="boop"} | msg=~` + `"` + `.*?(GET|POST|PUT|DELETE|PATCH|HEAD|OPTIONS) /loki/api/(?i)(\d+[a-z]|[a-z]+\d)\w*/query_range` + `"` + `[1d]))`
```
MasslessParticle, line 3476: I don’t think we need to redefine `query` here

MasslessParticle (changes requested):

MasslessParticle:
Nice work! Just some test cleanup and this can be merged

pkg/logql/syntax/parser_test.go
akhilanarayanan: good point. length check can go

MasslessParticle (approved):
LGTM, nice work

PR #10288, opened 2023-08-17
[helm] Make table_manager configuration toggle.

jkroepke:

table_manager is deprecated. It would be nice the add the possibility to remove the key from the config.

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/production/helm/loki/CHANGELOG.md
+++ b/production/helm/loki/CHANGELOG.md
@@ -13,6 +13,10 @@ Entries should include a reference to the pull request that introduced the chang
 
 [//]: # (<AUTOMATED_UPDATES_LOCATOR> : do not remove this line. This locator is used by the CI pipeline to automatically create a changelog entry for each new Loki release. Add other chart versions and respective changelog entries bellow this line.)
 
+## 5.14.0
+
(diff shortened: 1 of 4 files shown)


JStickler:
[docs team]  I feel like this new setting interacts with `tableManager.enabled` in some way?  Like both settings seem like they would need to both be enabled or disabled to avoid errors?
Also, this setting needs to be added to the Helm Charts reference documentation.

jkroepke:
> Like both settings seem like they would need to both be enabled or disabled to avoid errors?

If `retention_period` is set to `0`, then `retention_deletes_enabled`  has no effect.
If `retention_period` is set greater then `0`, then `retention_deletes_enabled` can be still true or false..

The `retention_period` could be act as `enabled` property here.

jkroepke:
The behavior is documented here https://grafana.com/docs/loki/latest/configuration/ Thats, was my source of the comment above.

JStickler:
Ah, I was thinking the Helm configuration reference.

MichelHollands (approved):
LGTM

PR #10318, opened 2023-08-23
Revert "Stopped String() from doubly escaping regex characters"

kavirajk:
Reverts grafana/loki#10277

This currently breaks and cause panics for query with regex label matcher. Reverting it till we find the fix.


salvacorts (approved):

dannykopping (approved):
LGTM

PR #10322, opened 2023-08-23
Deprecate ruler setting with misleading name

chaudum:

The setting defaults to 0s and is not supposed to be changed, therefore it can be removed in a future version (e.g. 3.0) of Loki.


pkg/validation/limits.go
```diff
@@ -564,9 +564,10 @@ func (o *Overrides) MaxQueryLookback(_ context.Context, userID string) time.Dura
 	return time.Duration(o.getOverridesForUser(userID).MaxQueryLookback)
 }
 
+// Deprecated, can be removed in next major version
 // EvaluationDelay returns the rules evaluation delay for a given user.
 func (o *Overrides) EvaluationDelay(userID string) time.Duration {
-	return time.Duration(o.getOverridesForUser(userID).RulerEvaluationDelay)
+	return time.Duration(0)
```
ashwanthgoli, line 570: we should return this value even if it's deprecated right? should we add a warn message instead?

ashwanthgoli (approved):

MichelHollands (approved):
LGTM

PR #10331, opened 2023-08-23
Put unescaped regexes in a raw string literal

MasslessParticle:
[This PR](https://github.com/grafana/loki/pull/10277) fixed an issue where regexes were being escaped multiple times and causing otherwise valid queries to fail our length check. It put the resulting regexes in double quotes which meant that any regexes that _needed_ escaping were now invalid. 

This PR make is so the unescaped regex is put in a raw string literal so any extra escaping in unnecessary.

### DIFF
--- a/pkg/logql/log/label_filter.go
+++ b/pkg/logql/log/label_filter.go
@@ -368,6 +368,16 @@ type lineFilterLabelFilter struct {
 	filter Filterer
 }
 
+// overrides the matcher.String() function in case there is a regexpFilter
+func (s *lineFilterLabelFilter) String() string {
+	if unwrappedFilter, ok := s.filter.(regexpFilter); ok {
+		rStr := unwrappedFilter.String()
+		str := fmt.Sprintf("%s%s`%s`", s.Matcher.Name, s.Matcher.Type, rStr)
+		return str
+	}
+	return s.Matcher.String()
+}
+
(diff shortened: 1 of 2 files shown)


DylanGuedes (approved):

dannykopping (commented):
I'm not understanding why we need this change, to be honest. The query length problem is orthogonal; it's not caused exclusively by escaped chars (but is exacerbated by it) and the problem would better be solved with a configurable max query length rather than a fixed 5K value.

Are the escaped strings illegible, or in some other way problematic besides length?

MasslessParticle:
@dannykopping 
There's a few things:
- Queries that initially pass validation shouldn't fail in later processing. Changing the query length via escaping can cause an otherwise good query to cause errors/panics in the `rangemapper`
- Our maxLength is a hard limit, we can increase it, but modifying a query by escaping internal representations can always cause a query that passed validation to overrun the limit
- maxLength is a hard limit because of the way we've implemented the `lexer`. See [this pr](https://github.com/grafana/loki/pull/9134) for more details. We're in a situation where we need to try and keep the query representation as close to what was passed to us as possible _or_ rewrite the lexer. The latter seems dramatically more costly

dannykopping:
Aha! Got it, thanks for the clarification 👍

dannykopping (approved):
LGTM

PR #10337, opened 2023-08-24
non-indexed labels: support for deletion of logs with non-indexed labels

sandeepsukhani:
We recently added support for storing and querying non-indexed labels with each log line. This PR wires up the code for supporting deletion using non-indexed labels.

**Checklist**
- [x] Tests updated


salvacorts (approved):
LGTM. I'm unfamiliar with the deletion logic but overall looks reasonable and the tests look good.

Issue #9594, opened 2023-06-01
Distinct operator returning too many log lines


svennergr:
**Describe the bug**
I was using the new `distinct` operator which returned more log lines than expected.

**To Reproduce**
1. Started Loki
```json
{"version":"2.8.2","revision":"9f809eda7","branch":"HEAD","buildUser":"root@309137d15b61","buildDate":"2023-05-03T11:09:58Z","goVersion":""}
```
2. Query: `{place="moon"} | json | distinct level | line_format "{{.level}}"`
3. Results where multiple loglines with the same level.

**Expected behavior**
I expected to see distinct levels.

**Environment:**
 - Loki:
```json
{"version":"2.8.2","revision":"9f809eda7","branch":"HEAD","buildUser":"root@309137d15b61","buildDate":"2023-05-03T11:09:58Z","goVersion":""}
```

**Screenshots, Promtail config, or terminal output**

<img width="602" alt="image" src="https://github.com/grafana/loki/assets/8092184/3aea1149-029b-42c3-854d-583e43825943">

svennergr:
fyi @liguozhong

liguozhong:
Thank you for your feedback.
I will take the time to take a look at this bug in the next two weeks.

liguozhong:
https://github.com/grafana/loki/pull/9624
fixed in this pr.

PR #9624, opened 2023-06-05
[bugfix] logq: fix Distinct operator returning too many log lines

liguozhong:
**What this PR does / why we need it**:

Added a global filter function for distinct. Applied in 2 modules.

1: After the querier queries multiple instances and s3. Do a global filter on the final result.
2: Do a global filter on each split result after split by time.

alternative:
We can also introduce a global state map to fix this bug, but this introducing a state is a huge work.

**Which issue(s) this PR fixes**:
Fixes #9594 #8662

**Special notes for your reviewer**:
fix distinct result. only return 3 log .
image

**Checklist**
- [x] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`
- [ ] For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)

PR #10356, opened 2023-08-25
Revert "LogQL: Introduce `distinct` (#8662)"

kavirajk:
This reverts commit 38b298c3fe68004336f078338a548aa6cabb290d.

We are removing this PR #8662 because [known issue tracked here](https://github.com/grafana/loki/issues/9594) and some issues we facing internally with query splitting and sharding work(hard to parallelize, merging two `distinct` sets can have duplicates, hard to make it work with metric queries, etc).

We decided best to remove it before make it to public release.

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


MasslessParticle (approved):

slim-bean:
cc @liguozhong apologies for removing your hard work here, but unfortunately the current implementation has too many issues which are hard/impossible to fix. 

Specifically we can't make it work with metric queries at all, and trying to make it work in log queries as well requires some very special handling that I'm not sure we want to introduce at this time.

Additionally we've been doing some query splitting work in Grafana itself which further lends to this operation having to be excluded.

Sorry for all the folks that have been using this feature ☹️ I'm not sure if/when we'll be able to re-introduce it.

I think the only workaround I'm aware of is using `logcli` with other linux/unix tools like `sort | uniq`

liguozhong:
@slim-bean thanks ,😊 Although it was revert, it can be regarded as an attempt by the loki community.It's a pleasure to be a part of history. 

In addition, `distinct` is used to implement a feature similar to `logreduce`. It is very powerful and can only take out one log for each error code.
image
https://help.sumologic.com/docs/search/logreduce/logreduce-operator/
It is very important for problem discovery when the system is released. I will continue to use the `distinct` feature in my branch to apply to ourselves internal system

and https://github.com/grafana/loki/pull/9624 this pr fixes some obvious problems, maybe there are some difficult problems that I didn’t notice, I just use it to solve the problem of `logreduce`, so I don’t use it very much Very comprehensive. But this is enough for us to release the system more safely

When I implemented the first distinct version before, thanks to the review of @jeschkies , I noticed that we need a distributed state to implement the `distinct` more standardly. I also checked how facebook `presto` `distinct` implements, which requires a larger work. Therefore, distributed state was not introduced at that time.

[jeschkies](https://github.com/jeschkies) [on Apr 17](https://github.com/grafana/loki/pull/8662/files#r1168470204) • 
How do you keep the state across multiple queriers? I think this either has to run in the frontend or we'd need to synchronize the state somehow.

PR #10360, opened 2023-08-25
Storage: Introduce new "ObjectExists" API

DylanGuedes:
Introduce to our `Store` interface a new function named `ObjectExists`, which can be used by external tools to validate if a given file is present on an object store without downloading the file.

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/storage/chunk/client/alibaba/oss_object_client.go
+++ b/pkg/storage/chunk/client/alibaba/oss_object_client.go
@@ -81,10 +81,6 @@ func (s *OssObjectClient) ObjectExists(ctx context.Context, objectKey string) (b
 		return true, nil
 	}
 
-	if s.IsObjectNotFoundErr(err) {
-		return false, nil
-	}
-
 	return false, err
 }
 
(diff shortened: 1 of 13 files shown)


pkg/storage/chunk/client/alibaba/oss_object_client.go
```diff
@@ -71,6 +71,23 @@ func NewOssObjectClient(_ context.Context, cfg OssConfig) (client.ObjectClient,
...
+	var options []oss.Option
+	err := instrument.CollectedRequest(ctx, "OSS.ObjectExists", ossRequestDuration, instrument.ErrorCode, func(ctx context.Context) error {
+		_, requestErr := s.defaultBucket.GetObjectMeta(objectKey, options...)
+		return requestErr
+	})
+	if err == nil {
+		return true, nil
+	}
+
+	if s.IsObjectNotFoundErr(err) {
```
MasslessParticle, line 84: We never make a distinction between if the object is not found or some other error. Would it make sense to just 
```go
if err != nil {
  return false, err
}

return true, nil
```

DylanGuedes: hah agreed. I pushed a commit removing that extra check, lmk what you think.

MasslessParticle (approved):

PR #10362, opened 2023-08-25
Querier: add chunk refs fetch time to metrics

dannykopping:
Adds an entry to the `metrics.go` log line to track how long a query waits for its chunk references to be returned by the index. This is useful when trying to understand where processing time is being spent.


slim-bean (approved):
LGTM

PR #10374, opened 2023-08-28
Bound parallelism of `errgroup`

DylanGuedes:
While investigating ingesters behavior with @dannykopping, he helped me to identify that in ~10s of an ingester execution we had an insane number of goroutines for errgroup: `[golang.org/x/sync/errgroup.(*Group).Go.func1](http://127.0.0.1:37581/goroutine?id=12563616) N=37646`. In the end we were able to track down that the goroutines explosion were due to `IndexSlice.For`. 
He then also showed me that we weren't bounding parallelism at two different places: `table.ForEachConcurrent` and `indexSet.ForEachConcurrent`. I'm including benchmarks for `IndexSlice.For` but not for the other places, because I think it isn't really necessary.

I'm adding a new `max_parallelism_table_ops` config that allows bounding the parallelism of different parts of the storage. The default is 50 but I'm open to picking a different number.

Benchmark results:
```
benchstat old.txt new.txt 
goos: linux
goarch: amd64
                       │   old.txt    │               new.txt               │
                       │    sec/op    │   sec/op     vs base                │
SliceForParallelism-24   4205.8m ± 6%   195.6m ± 2%  -95.35% (p=0.000 n=10)

                       │    old.txt    │               new.txt                │
                       │     B/op      │     B/op      vs base                │
SliceForParallelism-24   20.576Gi ± 0%   1.018Gi ± 0%  -95.05% (p=0.000 n=10)

                       │   old.txt    │               new.txt               │
                       │  allocs/op   │  allocs/op   vs base                │
SliceForParallelism-24   278.81k ± 0%   13.83k ± 0%  -95.04% (p=0.000 n=10)
```

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/storage/stores/tsdb/multi_file_index_test.go
```diff
@@ -154,3 +154,60 @@ func TestMultiIndex(t *testing.T) {
...
+				{
+					MinTime:  8,
+					MaxTime:  9,
+					Checksum: 4,
+				},
+			},
+		},
+	}
+
+	dir := b.TempDir()
```
dannykopping, line 203: Remember to clean this up at the end

chaudum: Temporary directories created with `*testing.T` or `*testing.B` are automatically cleaned up.

```
// TempDir returns a temporary directory for the test to use.
// The directory is automatically removed by Cleanup when the test and
// all its subtests complete.
```

pkg/loki/modules.go
```diff
@@ -1140,6 +1140,7 @@ func (t *Loki) initCompactor() (services.Service, error) {
 		level.Info(util_log.Logger).Log("msg", "-boltdb.shipper.compactor.shared-store not specified, initializing compactor to operator on the following object stores", "stores", strings.Join(stores, ", "))
 	}
 
+	t.Cfg.CompactorConfig.TableOperationsParallelism = t.Cfg.StorageConfig.MaxParallelismTableOps
```
chaudum, line 1143: Mutating the config :/

pkg/storage/stores/indexshipper/compactor/table.go
```diff
@@ -109,6 +112,7 @@ func newTable(ctx context.Context, workingDirectory string, indexStorageClient s
 		baseUserIndexSet:   storage.NewIndexSet(indexStorageClient, true),
 		baseCommonIndexSet: storage.NewIndexSet(indexStorageClient, false),
 		uploadConcurrency:  uploadConcurrency,
+		parallelism:        parallelism,
```
chaudum, line 115: Is there a conceptual difference between `uploadConcurrency` and `parallelism`, except that the former is for uploading tables, the latter is for downloading tables?

Maybe the settings could be consolidated?

pkg/loki/modules.go
DylanGuedes: `CompactorConfig.TableOperationsParallelism` is just reusing `MaxParallelismTableOps` and this is the only place where an assignment for it happens. This way I don't have to change the whole compactor code to hold the Store config everywhere.

pkg/storage/stores/indexshipper/compactor/table.go
DylanGuedes: I considered it, but if I remove `uploadConcurrency` it would be a breaking change, if I reuse `uploadConcurrency` everywhere it would be misleading and can lead to unexpected behavior, as everyone using low `uploadConcurrency` would experience a degradation in parallelism on other parts of the code.

DylanGuedes:
I'm making this a draft because I ran some experiments and the benefits were not super evident. I'll see what happened.

DylanGuedes:
Moving to `ready for review` again since we started to experience bad gateway performance due to scheduler latency recently.

pkg/storage/factory.go
```diff
@@ -317,6 +318,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.DisableBroadIndexQueries, "store.disable-broad-index-queries", false, "Disable broad index queries which results in reduced cache usage and faster query performance at the expense of somewhat higher QPS on the index store.")
 	f.IntVar(&cfg.MaxParallelGetChunk, "store.max-parallel-get-chunk", 150, "Maximum number of parallel chunk reads.")
 	cfg.BoltDBShipperConfig.RegisterFlags(f)
+	f.IntVar(&cfg.MaxParallelismTableOps, "store.max-parallelism-table-ops", 50, "Maximum number of concurrence used for each table operation. We recommend 2*CPUs available")
```
chaudum, line 321: `2 * NUM_CPUs` is unlikely a value of 50 (default)

docs/sources/configure/_index.md
```diff
@@ -1975,6 +1975,10 @@ congestion_control:
 # CLI flag: -store.max-parallel-get-chunk
 [max_parallel_get_chunk: <int> | default = 150]
 
+# Max parallelism used by a table operation.
+# CLI flag: -store.max-parallelism-table-ops
+[max_parallelism_table_ops: <int> | default = 50]
```
chaudum, line 1980: To be consistent with `max_parallel_get_chunk` we should consider naming it like:

```suggestion
[max_parallel_table_ops: <int> | default = 50]
```

pkg/storage/factory.go
```diff
@@ -285,6 +285,7 @@ type Config struct {
 	IndexQueriesCacheConfig  cache.Config `yaml:"index_queries_cache_config"`
 	DisableBroadIndexQueries bool         `yaml:"disable_broad_index_queries"`
 	MaxParallelGetChunk      int          `yaml:"max_parallel_get_chunk"`
+	MaxParallelismTableOps   int          `yaml:"max_parallelism_table_ops" doc:"description=Max parallelism used by a table operation." category:"experimental"`
```
chaudum, line 288: See comment above:

```suggestion
	MaxParallelTableOps   int          `yaml:"max_parallel_table_ops" doc:"description=Max parallelism used by a table operation." category:"experimental"`
```

ashwanthgoli:
we don't need to set the parallelism in `shipper.go` and `table.go` I think given they always spawn atmost 2 goroutines.

DylanGuedes:
> we don't need to set the parallelism in `shipper.go` and `table.go` I think given they always spawn atmost 2 goroutines.

that's a nice observation but while I was doing the change I noticed we'll have to pass the parallelism config down to the shipper and table anyway so that they can reach the indexSet so not sure if there are any benefits on not passing them. WDYT?

chaudum:
Closing because the PR is out-of-date

PR #10375, opened 2023-08-28
fix: respect query matcher in ingester when getting label values

trevorwhitney:
There is a bug when passing a query parameter to the `/label/:label/values`. Currently the ingester is not respecting this query string, and so is always returning all label values. This PR fixes that bug.


trevorwhitney:
Test locally with single binary and SSD mode (docker), and on a microservices dev env. Worked in all 3 places

PR #10341, opened 2023-08-25
deprecate old index types and non-object stores

ashwanthgoli:
**What this PR does / why we need it**:
Deprecates old index types: `aws, aws-dynamo, gcp, gcp-columnkey, bigtable, bigtable-hashed, cassandra, grpc`. Only `tsdb` and `boltdb-shipper` are supported.

Deprecates non-object stores. The next major release might not allow the usage of these as `object_store` in `schema_config`: `aws-dynamo, gcp, gcp-columnkey, bigtable, bigtable-hashed, cassandra, grpc.`

Deprecates table manager which is not needed once when using `boltdb-shipper` or `tsdb` index type

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

**Checklist**
- [x] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated
  - [ ] If the change is worth mentioning in the release notes, add `add-to-release-notes` label
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/setup/upgrade/_index.md`
- [ ] For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)

PR #10383, opened 2023-08-29
Add deprecation notes to Storage documentation

chaudum:

Legacy storage backends (that are not single store aka "shipper") have been deprecated with https://github.com/grafana/loki/pull/10341
The docs should reflect this change.

The structure of the page changed _slightly_:
* Made object storage a separate section under "Chunk storage"
* Refer to this section in the "Single Store" section for supported backends
* Moved Azure example underneath the rest of the single store examples

While it is not yet perfect, I tried to change as little as possible and keep the big change for when the legacy storage backends are removed.

### DIFF
--- a/docs/sources/storage/_index.md
+++ b/docs/sources/storage/_index.md
@@ -14,7 +14,7 @@ simplifies the operation and significantly lowers the cost of Loki.
 Prior to Loki 2.0, chunks and index data were stored in separate backends:
 object storage (or filesystem) for chunk data and NoSQL/Key-Value databases for index data.
 
-Loki 2.0 brings an index mechanism named 'boltdb-shipper' and is what we now call [Single Store](#loki-single-store).
+Loki 2.0 brings an index mechanism named 'boltdb-shipper' and is what we now call [Single Store](#single-store).
 This type only requires one store, the object store, for both the index and chunks.
 More detailed information can be found on the [operations page]({{< relref "../operations/storage/boltdb-shipper.md" >}}).
 
--- a/docs/sources/storage/_index.md
+++ b/docs/sources/storage/_index.md
@@ -11,85 +11,110 @@ is then compressed and stored in chunks in object stores such as S3 or GCS, or
 even locally on the filesystem. A small index and highly compressed chunks
 simplifies the operation and significantly lowers the cost of Loki.
 
-Until Loki 2.0, index data was stored in a separate index.
+Prior to Loki 2.0, chunks and index data were stored in separate backends:
+object storage (or filesystem) for chunk data and NoSQL/Key-Value databases for index data.
 
-Loki 2.0 brings an index mechanism named 'boltdb-shipper' and is what we now call Single Store Loki.
-This index type only requires one store, the object store, for both the index and chunks.
+Loki 2.0 brings an index mechanism named 'boltdb-shipper' and is what we now call [Single Store](#loki-single-store).
+This type only requires one store, the object store, for both the index and chunks.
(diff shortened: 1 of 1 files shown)


docs/sources/storage/_index.md
```diff
@@ -11,85 +11,110 @@ is then compressed and stored in chunks in object stores such as S3 or GCS, or
 even locally on the filesystem. A small index and highly compressed chunks
 simplifies the operation and significantly lowers the cost of Loki.
 
-Until Loki 2.0, index data was stored in a separate index.
+Prior to Loki 2.0, chunks and index data were stored in separate backends:
+object storage (or filesystem) for chunk data and NoSQL/Key-Value databases for index data.
 
-Loki 2.0 brings an index mechanism named 'boltdb-shipper' and is what we now call Single Store Loki.
-This index type only requires one store, the object store, for both the index and chunks.
+Loki 2.0 brings an index mechanism named 'boltdb-shipper' and is what we now call [Single Store](#loki-single-store).
```
JStickler, line 17: How does TSDB fit in here?  Should we update line 17 to point new users to TSDB instead of hoping they read all the way down to line 102?

docs/sources/storage/_index.md
```diff
@@ -11,85 +11,110 @@ is then compressed and stored in chunks in object stores such as S3 or GCS, or
...
+### Cassandra (deprecated)
 
-#### BoltDB (deprecated)
+Cassandra is a popular database and one of Loki's possible chunk stores and is production safe.
 
-Also known as "boltdb-shipper" during development (and is still the schema `store` name). The single store configurations for Loki utilize the chunk store for both chunks and the index, requiring just one store to run Loki.
+> **Note:** This storage type for chunks is deprecated and may be removed in future major versions of Loki.
 
-Performance is comparable to a dedicated index type while providing a much less expensive and less complicated deployment.
+## Index storage
```
JStickler, line 68: I think it would make sense to add one or two sentences here to explain that if you're using boltdb or TSDB you do not need a separate index storage. Otherwise it looks odd that everything under this heading is marked deprecated.

docs/sources/storage/_index.md
```diff
@@ -11,85 +11,110 @@ is then compressed and stored in chunks in object stores such as S3 or GCS, or
...
 
-### BoltDB
+### BoltDB (deprecated)
 
 BoltDB is an embedded database on disk. It is not replicated and thus cannot be used for high availability or clustered Loki deployments, but is commonly paired with a `filesystem` chunk store for proof of concept deployments, trying out Loki, and development. The [boltdb-shipper]({{< relref "../operations/storage/boltdb-shipper" >}}) aims to support clustered deployments using `boltdb` as an index.
 
-### Azure Storage Account
+> **Note:** This storage type for indexes is deprecated and may be removed in future major versions of Loki.
+
+## Loki Single Store
```
JStickler, line 98: I kinda want to move this whole Loki Single Store section (lines 98 - 110) higher up on the page.  Above both the Chunk storage and Index Storage headings, since it is what we want new users to be using.

JStickler (commented):
[docs team]

ashwanthgoli:
thanks for this @chaudum

docs/sources/storage/_index.md
chaudum: Was thinking exactly the same. I did not want to change the page for the 2.9 docs too much, but only add the deprecation notes. For the 3.0 docs, we should definitely re-organise the whole page.

chaudum:
@JStickler thanks for the review. I pushed an update which I think addresses your suggestions.

PR #10387, opened 2023-08-29
Fix k3d and docker-compose local development envs

trevorwhitney:
This PR fixes bugs in two of the local dev envs we have.

* production/docker was missing a required CLI flag for the backend target
* tools/dev/k3d was failing to bring up a k8s cluster due to an issue with the prometheus operator crd bundled with the kube-prometheus-stack chart.

### DIFF
--- a/production/docker/docker-compose.yaml
+++ b/production/docker/docker-compose.yaml
@@ -42,7 +42,7 @@ services:
   prometheus:
     image: prom/prometheus:v2.27.0
     ports:
-      - 9090:9090
+      - 9090
     volumes:
       - ./config/prometheus.yaml:/etc/prometheus/prometheus.yml
       - prometheus:/prometheus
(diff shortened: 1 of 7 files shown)


production/docker/docker-compose.yaml
```diff
@@ -42,7 +42,7 @@ services:
   prometheus:
     image: prom/prometheus:v2.27.0
     ports:
-      - 9090:9090
+      - 9090
```
trevorwhitney, line 45: I have monitoring on my local system, so mapping to the default ports on localhost conflicts with that. This syntax will map container port `9090` to a random port on the host, so it's still accessible without conflicting

production/docker/docker-compose.yaml
```diff
@@ -171,7 +171,7 @@ services:
       - "7946"
       # uncomment to use interactive debugging
       # - "60000-60002:40000" #  makes the replicas available on ports 60000, 60001, 60002
-    command: "-config.file=/etc/loki/loki.yaml -target=backend"
+    command: "-config.file=/etc/loki/loki.yaml -target=backend -legacy-read-mode=false"
```
trevorwhitney, line 174: this was the fix needed to spin up the backend pod

MasslessParticle (approved):

PR #10400, opened 2023-08-30
Remove anonymous step evaluator and add `explain`.

jeschkies:
This removes the `newStepEvaluator(...)` and replaces most `StepEvaluator`s with a struct. This makes tracing the evaluators simpler. It also enables us to support something like the `Explain` method.

Please note that `StepEvaluator.Explain()` is not usable right now because creating a step evaluator already executes the query. See the comment in `TestExplain`. Ideally we remove that behaviour. However, that would be a change in the semantics which this PR should not do.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/logql/explain.go
```diff
@@ -0,0 +1,59 @@
+package logql
+
+import (
+	"github.com/xlab/treeprint"
```
jeschkies, line 4: So I'd prefer [treeprinter](https://github.com/cockroachdb/cockroach/blob/master/pkg/util/treeprinter/tree_printer.go.) from CockroachDB but it's under a business license. It was under Apache before. I need to check with legal if we could copy it before it changed.

pkg/logql/downstream.go
```diff
@@ -382,43 +381,49 @@ func (ev *DownstreamEvaluator) NewIterator(
...
-			case 0:
-				return nil
-			case 1:
-				return errs[0]
-			default:
-				return util.MultiError(errs)
-			}
-		},
-	)
+func NewConcatStepEvaluator(evaluators []StepEvaluator) StepEvaluator {
```
chaudum, line 390: Can we remove the [preemtive interface](https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a) by returning the concrete implementation?

```suggestion
func NewConcatStepEvaluator(evaluators []StepEvaluator) *concatStepEvaluator {
```

`concatStepEvaluator` implicitly implements `StepEvaluator` already.

Ref: https://github.com/golang/go/wiki/CodeReviewComments#interfaces

Same goes with other implementations.

(`concatStepEvaluator` might need to become `ConcactStepEvaluator`, because an exported function should only return exported objects)

pkg/logql/explain.go
```diff
@@ -0,0 +1,59 @@
...
+	b := parent.AddBranch("Concat")
+	if len(e.evaluators) < 3 {
+		for _, child := range e.evaluators {
+			child.Explain(b)
+		}
+	} else {
+		e.evaluators[0].Explain(b)
+		b.AddBranch("...")
+		e.evaluators[len(e.evaluators)-1].Explain(b)
+	}
```
chaudum: Did you do the ellipsis because concatStepEvaluator can possibly have _a lot_ of child evaluators?
If so, I would still print the first 3 and then do the ellipsis.

jeschkies: Yes. My original idea was to print the index as well. That's why I chose the last one.

pkg/logql/downstream.go
jeschkies: Done.

PR #10401, opened 2023-08-30
Rename and move index and chunk store interfaces

chaudum:

This is a first PR of a series to simplify and streamline the "store interfaces" to make them easier to understand and speed up development of new features.

This PR does not change any functionality or interface functions. Changes are:

- `RequestChunkFilterer` moved from `pkg/storage/chunk` to `pkg/storage/stores/index`
- `Filterer` moved from `pkg/storage/chunk` to `pkg/storage/stores/index`
- `RequestChunkFilterer` and `Filterer` are now in the same place as `Filterable`
- `ChunkReader`, `ChunkWriter`, `ChunkReaderWriter` and `ChunkFetcher` are renamed to `ReadStore`, `WriteStore`, `ReadWriteStore`, and `Fetcher` respectively and move from `pkg/storage/stores` to `pkg/storage/stores/chunk`
- `pkg/storage/stores/index` is usually aliased as `indexstore` when imported (to not confuse it with other `index` packages. the package contains the index store interfaces
- `pkg/storage/stores/chunk` is usually aliased as `chunkstore` when imported (to not confuse it with other `chunk` packages. the package contains the chunk store interfaces

### DIFF
--- a/pkg/storage/stores/chunk/chunkstore.go
+++ b/pkg/storage/stores/chunk/chunkstore.go
@@ -13,23 +13,23 @@ import (
 	"github.com/grafana/loki/pkg/storage/chunk/fetcher"
 )
 
-// Reader is an interface for a chunk store that reads data
-type Reader interface {
+// ReadStore is an interface for a chunk store that reads data
+type ReadStore interface {
 	SelectSamples(ctx context.Context, req logql.SelectSampleParams) (iter.SampleIterator, error)
 	SelectLogs(ctx context.Context, req logql.SelectLogParams) (iter.EntryIterator, error)
 	Series(ctx context.Context, req logql.SelectLogParams) ([]logproto.SeriesIdentifier, error)
 }
 
-// Writer is an interface for a chunk store that writes data
-type Writer interface {
+// WriteStore is an interface for a chunk store that writes data
+type WriteStore interface {
 	Put(ctx context.Context, chunks []chunk.Chunk) error
 	PutOne(ctx context.Context, from, through model.Time, chunk chunk.Chunk) error
 }
 
-// ReaderWriter is an interface for a chunk store that reads and writes data
-type ReaderWriter interface {
-	Reader
-	Writer
+// ReadWriteStore is an interface for a chunk store that reads and writes data
+type ReadWriteStore interface {
+	ReadStore
+	WriteStore
 }
 
 // Fetcher is an interface for a chunk store that needs to fetch chunks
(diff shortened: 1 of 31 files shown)


pkg/storage/stores/chunk/chunkstore.go
```diff
@@ -0,0 +1,40 @@
...
+	Series(ctx context.Context, req logql.SelectLogParams) ([]logproto.SeriesIdentifier, error)
+}
+
+// Writer is an interface for a chunk store that writes data
+type Writer interface {
+	Put(ctx context.Context, chunks []chunk.Chunk) error
+	PutOne(ctx context.Context, from, through model.Time, chunk chunk.Chunk) error
+}
+
+// ChunkWriter is an interface for a chunk store that reads and writes data
```
ashwanthgoli, line 29:
```suggestion
// ReaderWriter is an interface for a chunk store that reads and writes data
```

pkg/storage/stores/chunk/chunkstore.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/grafana/loki/pkg/iter"
+	"github.com/grafana/loki/pkg/logproto"
+	"github.com/grafana/loki/pkg/logql"
+	"github.com/grafana/loki/pkg/storage/chunk"
+	"github.com/grafana/loki/pkg/storage/chunk/fetcher"
+)
+
+// Reader is an interface for a chunk store that reads data
+type Reader interface {
```
ashwanthgoli, line 17: i feel that `Reader` is not exactly a reader for chunks
looking at the code refs, should we move this to querier instead. wdyt?

jeschkies (approved):

pkg/storage/stores/chunk/chunkstore.go
chaudum: I agree, `Reader` is probably not the best name for what the interface defines.

chaudum: `Reader -> ReadStore`

chaudum:
After a discussion with @ashwanthgoli about where the store interfaces should be located (`pkg/storages`, `pkg/storages/stores`, `pkg/storages/stores/{chunk,index}`) and how they should be named (`Writer` vs. `WriteStore`, etc), we came to the agreement that the `ChunkReader` interface, even though associated with the chunks package, is a different abstraction than the `ChunkWriter` at the same package/level.

So I opened a new, smaller **[PR](https://github.com/grafana/loki/pull/10435)**, which does mainly move the `ChunkReader` interface to `pkg/storages.SelectStore`.

PR #10403, opened 2023-08-30
Enable results cache for volume queries

trevorwhitney:
This PR enables the results cache for volume queries to the `index/volume` and `index/volume_range` endpoints


pkg/querier/queryrange/volume_cache.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/grafana/loki/pkg/util/validation"
+)
+
+type VolumeSplitter struct {
+	cacheKeyLimits
+}
+
+// GenerateCacheKey generates a cache key based on the userID, Request and interval.
+func (i VolumeSplitter) GenerateCacheKey(ctx context.Context, userID string, r queryrangebase.Request) string {
+	cacheKey := i.cacheKeyLimits.GenerateCacheKey(ctx, userID, r)
```
MasslessParticle, line 27: This cache key doesn't take into consideration `targetLabels` or `aggregateBy`. I think that means it will return incorrect results if, for example, a user requests the same matchers but a different target label.

MasslessParticle (changes requested):
Generally this looks great but I think we need to take request params into account.

pkg/querier/queryrange/volume_cache.go
trevorwhitney: very good point, thank you!

MasslessParticle (approved):

PR #10413, opened 2023-08-31
pkg/loki: provide more help to users sending bad queries

kevinburkesegment:
**What this PR does / why we need it**: Currently, if you have no idea you're supposed to use LogQL, and just send a random query to the API, here is the entirety of the HTTP response body you get back:

    parse error at line 1, col 1: syntax error: unexpected IDENTIFIER

This doesn't help users figure out what they did wrong, or explain what sort of syntax Loki is looking for, so attempt to provide more help by pointing people to the LogQL documentation.

**Which issue(s) this PR fixes**: No issue

**Checklist**
- [ ] Documentation added (No documentation)
- [X] Tests updated
- [ ] `CHANGELOG.md` updated
  - [ ] If the change is worth mentioning in the release notes, add `add-to-release-notes` label (Not needed)
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/setup/upgrade/_index.md` (Not needed)
- [ ] For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213) (Not needed)

### DIFF
--- a/pkg/loki/format_query_handler.go
+++ b/pkg/loki/format_query_handler.go
@@ -3,6 +3,7 @@ package loki
 import (
 	"encoding/json"
 	"net/http"
+	"strings"
 
 	"github.com/grafana/loki/pkg/logql/syntax"
 	"github.com/grafana/loki/pkg/util/server"
@@ -22,6 +23,12 @@ func formatQueryHandler() http.HandlerFunc {
 			statusCode = http.StatusBadRequest
 			status = "invalid-query"
 			errStr = err.Error()
+			if strings.Contains(errStr, "parse error at line 1") && strings.Contains(errStr, "unexpected IDENTIFIER") {
+				// User probably sent a query like "foo", ie. they are trying
+				// to use a grep like syntax. Help them find the right place to
+				// submit a proper query
+				errStr += " (queries should use LogQL: https://grafana.com/docs/loki/latest/logql/log_queries/)"
+			}
 		}
 
 		if err == nil {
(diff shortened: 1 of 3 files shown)


pkg/loki/format_query_handler.go
```diff
@@ -22,6 +23,12 @@ func formatQueryHandler() http.HandlerFunc {
 			statusCode = http.StatusBadRequest
 			status = "invalid-query"
 			errStr = err.Error()
+			if strings.Contains(errStr, "parse error at line 1") && strings.Contains(errStr, "unexpected IDENTIFIER") {
```
dannykopping, line 26: I appreciate what you're trying to do here, but I think this is too generic; an unexpected identifier could still be encountered in a LogQL query.

I do think, though, that it would be very nice to return more context to users when errors are encountered. I would be in favour of adding an additional field to the response JSON where we could insert links and/or help text under certain conditions.

kevinburkesegment: OK - do you think we could expose that field in the LogCLI output as well? 

I'm actually a bit confused about this because what I see when I get the error back from the server is `text/plain` output but I can't figure out where that is being generated - I looked at all instances in the source tree where we send back text/plain responses and they don't seem to make sense. 

```
$ curl -v -u "$USER:$(cat disk-tokens/grafana-cloudaccesspolicy-token)" https://logs-prod-006.grafana.net/loki/api/v1/query_range\?direction\=BACKWARD\&end\=1693289918339003000\&limit\=30\&query\=segment\&start\=1693286318339003000
> GET /loki/api/v1/query_range?direction=BACKWARD&end=1693289918339003000&limit=30&query=segment&start=1693286318339003000 HTTP/1.1
> Host: logs-prod-006.grafana.net
> Authorization: Basic <redacted>
> User-Agent: curl/8.2.1
> Accept: */*
>
< HTTP/1.1 400 Bad Request
< Content-Length: 65
< Content-Type: text/plain; charset=utf-8
< Date: Thu, 31 Aug 2023 19:59:09 GMT
< Vary: Accept-Encoding
<
parse error at line 1, col 1: syntax error: unexpected IDENTIFIER
```

kevinburkesegment: I opened #10414 to discuss

chaudum:
While I agree that the parser error is very generic, this is not a technical problem to, it's an educational one.
Improving the parser is quite a big challenge.

PR #10416, opened 2023-09-01
lambda-promtail: Add support for WAF logs in S3

lpugoy:
Adds support for WAF logs in S3 in lambda-promtail

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


lpugoy:
Used https://github.com/grafana/loki/pull/8254 as a guide which was unfortunately closed without merging.

tools/lambda-promtail/lambda-promtail/s3.go
```diff
@@ -182,9 +203,19 @@ func parseS3Log(ctx context.Context, b *batch, labels map[string]string, obj io.
...
+				if err != nil {
+					return err
+				}
+			} else if parser.timestampType == "unix" {
+				// convert to microseconds so that we only use one function
+				usec, err := toMicroseconds(match[1])
+				if err != nil {
+					return err
+				}
+				timestamp = time.UnixMicro(usec).UTC()
```
cstyan, line 218: we could simplify this to a switch statement, and include a default cause with some warning level logging

tools/lambda-promtail/lambda-promtail/s3.go
```diff
@@ -301,3 +332,16 @@ func stringToRawEvent(body string) (map[string]interface{}, error) {
...
+	i, err := strconv.ParseInt(s, 10, 64)
+	if err != nil {
+		return usec, err
+	}
+	iPow10 := int(math.Log10(float64(i)))
+	multiplier := math.Pow10(15 - iPow10)
+	usec = int64(float64(i) * multiplier)
+
+	return usec, err
+}
```
cstyan, line 347: could you add another comment in here about the transformation that's being done? an example would help

lpugoy:
@cstyan Thank you for your comments, added commits to address them.

tools/lambda-promtail/lambda-promtail/s3.go
```diff
@@ -333,15 +335,42 @@ func stringToRawEvent(body string) (map[string]interface{}, error) {
...
-	usec = int64(float64(i) * multiplier)
 
-	return usec, err
+	iLog10 := int(math.Log10(float64(i)))
+	multiplier := math.Pow10(UNIX_SEC_LOG10 - iLog10)
+	sec = int64(float64(i) * multiplier)
+
+	fractionalSec := float64(i % sec)
+	if fractionalSec == 0 {
+		return sec, 0, err
```
cstyan, line 368: I think you want to set an actual value for `err` here? otherwise it has the nil value from `strconv.ParseInt`

tools/lambda-promtail/lambda-promtail/s3_test.go
```diff
@@ -652,64 +682,71 @@ func TestProcessSQSEvent(t *testing.T) {
 	require.True(t, handlerCalled)
 }
 
-func TestToMicroseconds(t *testing.T) {
+func TestGetUnixSecNsec(t *testing.T) {
```
cstyan, line 685: this test needs a case for the `fractionalSec == 0` case as well

tools/lambda-promtail/lambda-promtail/s3.go
lpugoy: Yes, `err` would be nil if `fractionalSec` is 0. There would only be an error during parsing, `fractionalSec` being 0 just means that the Unix time only has seconds resolution. It is not an error.

tools/lambda-promtail/lambda-promtail/s3_test.go
lpugoy: [This](https://github.com/grafana/loki/pull/10416/files#diff-f89d8b6a0bfefcada6292d108d96a954c426fa6850340440149542bb3ab613b4R696-R704) tests `fractionalSec == 0`.

cstyan (approved):
just need to update based on main one last time I think

lpugoy:
Sorry, just saw a typo in the test; it was referring to the old function name. Pushed a fix.

PR #10418, opened 2023-09-01
operator: Use a condition to warn when labels for zone-awareness are empty

btaani:
For zone-aware replication, if the `spec.replication.topologyKey` node label has no value, then the lokistack will report a degraded condition

Fixes [LOG-4454](https://issues.redhat.com/browse/LOG-4454)

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated


operator/apis/loki/v1/lokistack_types.go
```diff
@@ -992,6 +992,8 @@ const (
 	ReasonQueryTimeoutInvalid LokiStackConditionReason = "ReasonQueryTimeoutInvalid"
 	// ReasonNoZoneAwareNodes when the cluster does not contain any nodes with the labels needed for zone-awareness.
 	ReasonNoZoneAwareNodes LokiStackConditionReason = "ReasonNoZoneAwareNodes"
+	// ReasonTopologyKeyNodeLabelsEmpty when the topologyKey node label has an empty value
+	ReasonTopologyKeyNodeLabelsEmpty LokiStackConditionReason = "ReasonTopologyKeyNodeLabelsEmpty"
```
xperimental, line 996: Now that we have two degraded conditions regarding zone-awareness, maybe it does make sense to "group them". Maybe something like this?

```suggestion
	// ReasonZoneAwareNodesMissing when the cluster does not contain any nodes with the labels needed for zone-awareness.
	ReasonZoneAwareNodesMissing LokiStackConditionReason = "ReasonZoneAwareNodesMissing"
	// ReasonZoneAwareEmptyLabel when the node-label used for zone-awareness has an empty value.
	ReasonZoneAwareEmptyLabel LokiStackConditionReason = "ReasonZoneAwareEmptyLabel"
```

operator/internal/status/lokistack.go
```diff
@@ -17,10 +17,11 @@ import (
...
 const (
-	messageReady              = "All components ready"
-	messageFailed             = "Some LokiStack components failed"
-	messagePending            = "Some LokiStack components pending on dependencies"
-	messageDegradedNodeLabels = "Cluster contains no nodes matching the labels used for zone-awareness"
+	messageReady                      = "All components ready"
+	messageFailed                     = "Some LokiStack components failed"
+	messagePending                    = "Some LokiStack components pending on dependencies"
+	messageDegradedNodeLabels         = "Cluster contains no nodes matching the labels used for zone-awareness"
+	messageTopologyKeyNodeLabelsEmpty = "No value for the labels used for zone-awareness"
```
xperimental, line 24: For better readability I would use the `messageDegraded` prefix for the second condition as well:

```suggestion
	messageDegradedMissingNodes   = "Cluster contains no nodes matching the labels used for zone-awareness"
	messageDegradedEmptyNodeLabel = "Node-label used for zone-awareness has an empty value"
```

operator/internal/status/lokistack.go
```diff
@@ -44,6 +45,11 @@ var (
 		Message: messageDegradedNodeLabels,
 		Reason:  string(lokiv1.ReasonNoZoneAwareNodes),
 	}
+	conditionTopologyKeyNodeLabelsEmpty = metav1.Condition{
```
xperimental, line 48: I'd use the `conditionDegraded` prefix here as well:

```suggestion
	conditionDegradedEmptyNodeLabel = metav1.Condition{
```

operator/internal/status/lokistack.go
```diff
@@ -113,18 +123,26 @@ func generateCondition(ctx context.Context, cs *lokiv1.LokiStackComponentStatus,
...
+	for _, node := range nodeList.Items {
+		for _, nodeLabel := range nodeLabels {
+			if node.Labels[nodeLabel] == "" {
+				return true, false, nil
+			}
+		}
 	}
 
-	return len(nodeList.Items) > 0, nil
+	return len(nodeList.Items) > 0, true, nil
```
xperimental, line 145: I think I'd move the "are there nodes in the list?" check above the check for labels:

```suggestion
	if len(nodeList.Items) == 0 {
		return false, false, nil
	}

	for _, node := range nodeList.Items {
		for _, nodeLabel := range nodeLabels {
			if node.Labels[nodeLabel] == "" {
				return true, false, nil
			}
		}
	}

	return true, true, nil
```

operator/internal/status/lokistack.go
```diff
@@ -113,18 +123,26 @@ func generateCondition(ctx context.Context, cs *lokiv1.LokiStackComponentStatus,
 	return conditionReady, nil
 }
 
-func checkForZoneawareNodes(ctx context.Context, k client.Client, zones []lokiv1.ZoneSpec) (bool, error) {
+func checkForZoneawareNodes(ctx context.Context, k client.Client, zones []lokiv1.ZoneSpec) (bool, bool, error) {
```
xperimental, line 126: You can give the return values names:

```suggestion
func checkForZoneawareNodes(ctx context.Context, k client.Client, zones []lokiv1.ZoneSpec) (nodesOk bool, labelsOk bool, err error) {
```

xperimental (commented):
Just a few minor comments, this works fine. :+1: 

Unfortunately I had an idea during testing, that this issue might be more complicated than originally thought: The "label is empty" case is not caught by the scheduler like "there are no nodes matching that labelset" which means that the pods actually get scheduled and are stuck in our init-container then.

But here it gets more complicated: The init-container only waits if the value is "non empty", so if the customer provides multiple `topologyKey`s and there is at least one key with a value then the containers will start up completely. If this only happens for some nodes (there are nodes where a selected label is empty and nodes where this is not the case) then this effectively reduces the "zone-awareness".

I currently do not have a good idea how to guard for this, but at least we need to make sure that "all node-labels selected for zone-awareness can not have empty values" is somewhere in the docs :slightly_smiling_face: 

The pods belonging to a StatefulSet which are stuck in Init also need to be cleaned up manually. I'm currently checking if this is also the case for Pending pods.

xperimental:
Just noticed there's also an error in one of the checks: You can update the `api.md` by running

```
make docs/operator/api.md
```

in the `operator` directory.

xperimental (approved):

PR #10419, opened 2023-09-01
Flag categorize labels on streams response

salvacorts:
We recently introduced support for ingesting and querying structured metadata in Loki. This adds a new dimension to Loki's labels since now we arguably have three categories of labels: _stream_, _structured metadata_, and _parsed_ labels.

Depending on the origin of the labels, they should be used in LogQL expressions differently to achieve optimal performance. _stream_ labels should be added to stream matchers, _structured metadata_ labels should be used in a filter expression before any parsing expression, and _parsed_ labels should be placed after the parser expression extracting them.

The Grafana UI has a hard time dealing with this same problem. Before https://github.com/grafana/grafana/pull/73955, the filtering functionality in Grafana was broken since it was not able to distinguish between _stream_ and _structured metadata_ labels. Also, as soon as a parser expression was added to the query, filters added by Grafana would be appended to the end of the query regardless of the label category. The PR above implements a workaround for this problem but needs a better API on Loki's end to mitigate all corner cases.

Loki currently returns the following JSON for log queries:
```json
...
{
	"stream": {
	  "cluster": "us-central",
	  "container": "query-frontend",
	  "namespace": "loki",
	  "level": "info",
	  "traceID": "68810cf0c94bfcca"
	},
	 "values": [
                    [
                        "1693996529000222496",
                        "1693996529000222496 aaaaaaaaa.....\n"
                    ],
          ...
},
{
	"stream": {
	  "cluster": "us-central",
	  "container": "query-frontend",
	  "namespace": "loki",
	  "level": "debug",
	  "traceID": "a7116cj54c4bjz8s"
	},
	 "values": [
                    [
                        "1693996529000222497",
                        "1693996529000222497 bbbbbbbbb.....\n"
                    ],
          ...
},
...
```

As can be seen, there is no way we can distinguish the category of each label.

This PR introduces a new flag `X-Loki-Response-Encoding-Flags: categorize-labels` that makes Loki return categorized labels as follows:

```json
...
{
        "stream": {
	  "cluster": "us-central",
	  "container": "query-frontend",
	  "namespace": "loki",
	},
	"values": [
                    [
                        "1693996529000222496",
                        "1693996529000222496 aaaaaaaaa.....\n",
                       {
                           "structuredMetadata": {
                                "traceID": "68810cf0c94bfcca"
                           },
                          "parsed": {
                                 "level": "info"
                          }
                       }
                    ],
                    [
                        "1693996529000222497",
                        "1693996529000222497 bbbbbbbbb.....\n",
                       {
                           "structuredMetadata": {
                                "traceID": "a7116cj54c4bjz8s"
                           },
                          "parsed": {
                                 "level": "debug"
                          }
                       }
                    ],
    ...
},
...
```

Note that this PR only supports log queries, not metric queries. From a UX perspective, being able to categorize labels in metric queries doesn't have any benefit yet. Having said that, supporting this for metric queries would require some minor refactoring on top of what has been implemented here. If we decide to do that, I think we should do it on a separate PR to avoid making this PR even larger.

I also decided to leave out support for Tail queries to avoid making this PR even larger. Once this one gets merged, we can work to support tailing.

---

**Note to reviewers**

This PR is huge since we need to forward categorized all over the codebase (from parsing logs all the way to marshaling), fortunately, many of the changes come from updating tests and refactoring iterators.

Tested out in a dev cell with query `'{stream="stdout"} | label_format new="text"`.
- Without the new flag:
```
$ http http://127.0.0.1:3100/loki/api/v1/query_range\?direction\=BACKWARD\&end\=1693996529322486000\&limit\=30\&query\=%7Bstream%3D%22stdout%22%7D+%7C+label_format+new%3D%22text%22\&start\=1693992929322486000 X-Scope-Orgid:REDACTED
{
    "data": {
        "result": [
            {
                "stream": {
                    "new": "text",
                    "pod": "loki-canary-986bd6f4b-xqmb7",
                    "stream": "stdout"
                },
                "values": [
                    [
                        "1693996529000222496",
                        "1693996529000222496 pppppppppppp...\n"
                    ],
                    [
                        "1693996528499160852",
                        "1693996528499160852 pppppppppppp...\n"
                    ],
...
```

- With the new flag
```
$ http http://127.0.0.1:3100/loki/api/v1/query_range\?direction\=BACKWARD\&end\=1693996529322486000\&limit\=30\&query\=%7Bstream%3D%22stdout%22%7D+%7C+label_format+new%3D%22text%22\&start\=1693992929322486000 X-Scope-Orgid:REDACTED X-Loki-Response-Encoding-Flags:categorize-labels
{
    "data": {
        "encodingFlags": [
            "categorize-labels"
        ],
        "result": [
            {
                "stream": {
                    "pod": "loki-canary-986bd6f4b-xqmb7",
                    "stream": "stdout"
                },
                "values": [
                    [
                        "1693996529000222496",
                        "1693996529000222496 pppppppppppp...\n",
                        {
                            "parsed": {
                                "new": "text"
                            }
                        }
                    ],
                    [
                        "1693996528499160852",
                        "1693996528499160852 pppppppppppp...\n",
                        {
                            "parsed": {
                                "new": "text"
                            }
                        }
                    ],
...
```


integration/loki_micro_services_test.go
```diff
@@ -710,6 +711,272 @@ func TestQueryTSDB_WithCachedPostings(t *testing.T) {
 
 }
 
+func TestCategorizedLabels(t *testing.T) {
+	clu := cluster.New(nil, cluster.SchemaWithTDSBAndNonIndexedLabels)
```
vlad-diachenko, line 715: let's rename it to `SchemaWithTDSBAndStructuredMetadata`

vlad-diachenko (approved):
Looks brilliant 💎 . great job @salvacorts

owen-d (commented):
I like the idea, but I'd personally prefer something that's not a header to determine version (`version` parameter in query string|body, different api version prefix in path), but that's a nit. 

I also think it makes more sense and will be more efficient for us to not treat each stream distinct based on `(labels, structuredMetadata, parsed)`, but rather `labels` alone. It also feels more semantically correct as different log lines with separate structuredMetadata or parsed labels are still part of the same stream.

For instance:

```json
{
  "stream": {
    "labels": {
      "cluster": "us-central",
      "namespace": "loki",
      "container": "query-frontend",
    },
    "structuredMetadata": {
        "traceID": "68810cf0c94bfcca"
    },
    "parsed": {
         "level": "info"
     }
  },
  "values": [...]
}
```
should become
```json
{
  "stream": {
    "cluster": "us-central",
    "namespace": "loki",
    "container": "query-frontend",
  },
  "values": [
    {
      "ts": 0,
      "entry": "entry0",
      "structuredMetadata": {
        "traceID": "68810cf0c94bfcca"
      },
      "parsed": {
        "level": "info"
      }
    },
    ...
  ]
}
```
There are a few places in codecs, response merging (iirc) that will benefit from the reduced number of streams this will produces. Since it also feels more semantically correct, I think it's the way to go.

integration/loki_micro_services_test.go
salvacorts: Done

sandeepsukhani:
> I also think it makes more sense and will be more efficient for us to not treat each stream distinct based on (labels, structuredMetadata, parsed), but rather labels alone. It also feels more semantically correct as different log lines with separate structuredMetadata or parsed labels are still part of the same stream.

Would it not mean a difference in behaviour between metric and log queries? I can't think of a way to do a similar categorization of labels in metric queries.

pkg/push/push.proto
```diff
@@ -46,4 +46,8 @@ message EntryAdapter {
     (gogoproto.nullable) = false,
     (gogoproto.jsontag) = "structuredMetadata,omitempty"
   ];
+  repeated LabelPairAdapter parsed = 4 [
+    (gogoproto.nullable) = false,
+    (gogoproto.jsontag) = "parsed,omitempty"
+  ];
```
sandeepsukhani, line 52: I would love to avoid doing this change since it adds an unexpected field to the push payload. I know it is hard but maybe leave a todo and let us try fixing it in a separate PR to avoid overcomplicating this one.

pkg/util/marshal/query.go
```diff
@@ -310,16 +346,30 @@ func encodeStream(stream logproto.Stream, s *jsoniter.Stream) error {
 		s.WriteRaw(`"`)
 		s.WriteMore()
 		s.WriteStringWithHTMLEscaped(e.Line)
-		if len(e.StructuredMetadata) > 0 {
+
+		if categorizeLabels && (len(e.StructuredMetadata) > 0 || len(e.Parsed) > 0) {
```
sandeepsukhani, line 350: I think for consistency we should write empty object(`{}`) instead of skipping it when we have not structured metadata and parsed labels.

sandeepsukhani (commented):
nice work Salva! I have left some minor suggestions. I would like to do another review once those are taken care of.

integration/cluster/schema.go
```diff
@@ -45,6 +45,17 @@ schema_config:
...
+schema_config:
+  configs:
+    - from: {{.curPeriodStart}}
+      store: tsdb
+      object_store: filesystem
+      schema: v13
+      index:
+        prefix: index_
+        period: 24h
 `
```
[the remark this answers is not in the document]
salvacorts: Done thank you!

pkg/logql/log/labels.go
```diff
@@ -333,34 +458,64 @@ func (b *LabelsBuilder) Map() map[string]string {
...
+	stream := b.labels(StreamLabel).Copy()
+	structuredMetadata := b.labels(StructuredMetadataLabel).Copy()
+	parsed := b.labels(ParsedLabel).Copy()
+	b.buf = flattenLabels(b.buf, stream, structuredMetadata, parsed)
+	hash := b.hasher.Hash(b.buf)
+	if cached, ok := b.resultCache[hash]; ok {
+		return cached
+	}
+
+	return NewLabelsResult(b.buf.String(), hash, stream, structuredMetadata, parsed)
```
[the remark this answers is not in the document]
salvacorts: Good catch. Thank you!

pkg/logql/log/labels_test.go
```diff
@@ -61,19 +72,38 @@ func TestLabelsBuilder_LabelsResult(t *testing.T) {
...
+	expected = append(expected, expectedStreamLbls...)
+	expected = append(expected, expectedStucturedMetadataLbls...)
+	expected = append(expected, expectedParsedLbls...)
+	expected = labels.New(expected...)
+
+	actual := b.LabelsResult()
+	assertLabelResult(t, expected, actual)
 	// cached.
-	assertLabelResult(t, expected, b.LabelsResult())
+	assertLabelResult(t, expected, actual)
```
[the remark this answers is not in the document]
salvacorts: You are right, thank you.

pkg/util/marshal/query.go
salvacorts: Do you mean always serializing:
```json
{
    "structuredMetadata": {},
    "parsed": {}
}
```

If so, I think it's better not to do it. It will add a lot of overhead to the response.

pkg/push/push.proto
salvacorts: > I would love to avoid doing this chang

Likewise! The problem is this is the object we move around everywhere an throughout our components so we need to send it around somewhere.

> it adds an unexpected field to the push payload

I don't think that's a problem. Protobuf is extensible and backwards compatible when adding fields to the end of an object definition. So in this case, [it's still backwards compatible](https://protobuf.dev/programming-guides/proto3/#updating).

pkg/loghttp/entry.go
```diff
@@ -81,6 +103,27 @@ func (e *Entry) UnmarshalJSON(data []byte) error {
 	return err
 }
 
+func parseLabels(data []byte) (labels.Labels, error) {
+	var lbls labels.Labels
+	err := jsonparser.ObjectEach(data, func(key []byte, value []byte, t jsonparser.ValueType, _ int) error {
+		if t != jsonparser.String && t != jsonparser.Number {
```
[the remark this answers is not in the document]
salvacorts: I wanted to make this consistent with https://github.com/grafana/loki/pull/10550

pkg/loghttp/entry.go
```diff
@@ -52,26 +54,46 @@ func (e *Entry) UnmarshalJSON(data []byte) error {
...
+						structuredMetadata = lbls
+					}
+					if string(key) == "parsed" {
+						lbls, err := parseLabels(value)
+						if err != nil {
+							return err
+						}
+						parsed = lbls
+					}
+					return nil
```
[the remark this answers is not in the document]
salvacorts: The problem is, right now, the HTTP Push API expects structured metadata to be sent inside an object like:
```
[ "<unix epoch in nanoseconds>", "<log line>", {"trace_id": "0242ac120002", "user_id": "superUser123"}]
```

Unfortunately, to be backwards compatible, we need to be able to deserialize both the format above and:
```
[ "<unix epoch in nanoseconds>", "<log line>", { "structuredMetadata": {"trace_id": "0242ac120002", "user_id": "superUser123"}, "parsed": {"msg": "text"}}]
```

pkg/logql/log/labels.go
```diff
@@ -333,34 +458,67 @@ func (b *LabelsBuilder) Map() map[string]string {
...
+}
+
+func (b *BaseLabelsBuilder) toUncategorizedResult(buf labels.Labels) LabelsResult {
 	hash := b.hasher.Hash(buf)
 	if cached, ok := b.resultCache[hash]; ok {
 		return cached
 	}
-	res := NewLabelsResult(buf.Copy(), hash)
+
+	res := NewLabelsResult(buf.String(), hash, buf, nil, nil)
```
sandeepsukhani, line 505: I think we should be doing a copy of the buf since it is reused?

pkg/loghttp/entry.go
```diff
@@ -52,26 +54,46 @@ func (e *Entry) UnmarshalJSON(data []byte) error {
...
 				parseError = jsonparser.MalformedObjectError
 				return
 			}
+
 			var structuredMetadata labels.Labels
+			var parsed labels.Labels
 			if err := jsonparser.ObjectEach(value, func(key []byte, value []byte, dataType jsonparser.ValueType, _ int) error {
-				if dataType != jsonparser.String {
-					return jsonparser.MalformedStringError
+				if dataType == jsonparser.Object {
```
sandeepsukhani, line 66: can you please leave a comment saying we parse both push request and query response json here to make it clear for others?

pkg/logql/engine.go
```diff
@@ -498,28 +501,54 @@ func PopulateMatrixFromScalar(data promql.Scalar, params Params) promql.Matrix {
...
+					return nil, fmt.Errorf("failed to parse series labels to categorize labels: %w", err)
+				}
+
+				builder := labels.NewBuilder(lbls)
+				for _, label := range entry.StructuredMetadata {
+					builder.Del(label.Name)
+				}
+				for _, label := range entry.Parsed {
+					builder.Del(label.Name)
+				}
```
sandeepsukhani, line 541: Since this is on a hot path, it would be good to avoid doing this and try building the final response in the iterators. Maybe leave a todo and try doing it in a separate PR.

pkg/push/push.proto
sandeepsukhani: I am not worried about backwards compatibility, my ony concern is having an unwanted field on the proto definitions on the write path. I understand it is hard to do it so maybe leave a todo for now.

pkg/logql/log/labels.go
sandeepsukhani, line 505: I think we should be doing a copy of the buf since it is reused?

sandeepsukhani (changes requested):
the changes look great. Left just a minor comments and then we are good to merge it.

pkg/loghttp/entry.go
salvacorts: Done

pkg/logql/log/labels.go
salvacorts: Good catch, done

pkg/logql/engine.go
salvacorts: Added a todo

sandeepsukhani (approved):
LGTM

PR #10442, opened 2023-09-04
Jsonnet: Update shipper config to allow use of both boltdb and tsdb shippers for transitioning to tsdb from boltdb

liam-howe-maersk:
Allows configuring both boltdb and tsdb by setting both `using_boltdb_shipper` and `using_tsdb_shipper` to true and setting the necessary config for both. If you transition from boltdb to tsdb for the shipper there will still be some data in the boltdb index and so we must be able to support both until no more data is retained in boltdb.

N/A

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/production/ksonnet/loki/ruler.libsonnet
+++ b/production/ksonnet/loki/ruler.libsonnet
@@ -9,10 +9,7 @@ local k = import 'ksonnet-util/kausal.libsonnet';
 
   ruler_args:: $._config.commonArgs {
     target: 'ruler',
-  } + if $._config.using_boltdb_shipper && !$._config.using_tsdb_shipper then {
-    // Use PVC for caching
-    'boltdb.shipper.cache-location': '/data/boltdb-cache',
-  } else {},
+  },
 
   _config+:: {
     // run rulers as statefulsets when using boltdb-shipper to avoid using node disk for storing the index.
--- a/production/ksonnet/loki/ruler.libsonnet
+++ b/production/ksonnet/loki/ruler.libsonnet
@@ -9,7 +9,7 @@ local k = import 'ksonnet-util/kausal.libsonnet';
 
   ruler_args:: $._config.commonArgs {
     target: 'ruler',
-  } + if $._config.using_boltdb_shipper then {
+  } + if $._config.using_boltdb_shipper && !$._config.using_tsdb_shipper then {
     // Use PVC for caching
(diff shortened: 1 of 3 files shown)


production/ksonnet/loki/ruler.libsonnet
```diff
@@ -9,7 +9,7 @@ local k = import 'ksonnet-util/kausal.libsonnet';
 
   ruler_args:: $._config.commonArgs {
     target: 'ruler',
-  } + if $._config.using_boltdb_shipper then {
+  } + if $._config.using_boltdb_shipper && !$._config.using_tsdb_shipper then {
     // Use PVC for caching
     'boltdb.shipper.cache-location': '/data/boltdb-cache',
```
chaudum, line 14: IMO these lines can be removed altogether, since the cache location is already defined in the config, see your additions (lines 10-19) in `production/ksonnet/loki/shipper.libsonnet`. Wdyt?

chaudum (commented):
Thanks for the contribution, the change makes sense overall.

production/ksonnet/loki/ruler.libsonnet
liam-howe-maersk: It looks like here the config is being passed to the ruler component, I'm not super familiar with this component so to be honest I'm not sure if this is actually being used by the ruler, if you think that it isn't then happy to delete it?

chaudum: The global config is picked up by all components. So it's safe to delete it here.

liam-howe-maersk: I see makes sense, I have removed this config altogether here then.

Sheikh-Abubaker:
Is this issue resolved by the above mentioned PR ?

liam-howe-maersk:
There are a couple of tests failing in the integration tests but they don't look like they are related to my changes.
```
1.20.6-bullseye: Pulling from library/golang
received unexpected HTTP status: 503 Service Unavailable
```
I don't have permissions to retry them I don't think is there someone that can retry them or advise on what to do?

liam-howe-maersk:
It looks like this functionality has already been covered in this commit https://github.com/grafana/loki/commit/09927db78da8a3904c39452877a9b68a9d6a75db, will close this PR as no longer needed.

PR #10447, opened 2023-09-04
Add terraform example in ruler docs

fgouteroux:

Adding terraform example to manager alerts and recording rules.


dannykopping (approved):
Thanks very much for this @fgouteroux 🎉

Issue #10209, opened 2023-08-09
lambda-promtail: Add EventBridge event support


spedersen-emailage:
**Is your feature request related to a problem? Please describe.**

The only way to create an S3 bucket notification in CloudFormation is [via the bucket itself](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-notificationconfig.html). Even the `Serverless::Function` resource `Events` property proxies the `AWS::S3::Bucket` resource. Other notifications like CloudWatch Logs [can be created independently](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-subscriptionfilter.html) of the LogGroup resource.

Because of this, they have a [circular dependency problem](https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/79).

One workaround is to [use EventBridge (EB) as a middle-man](https://aws.amazon.com/blogs/aws/new-use-amazon-s3-event-notifications-with-amazon-eventbridge/) for event notifications. You send S3 events to EB, then create an EventsRule to invoke the Lambda when EB receives an S3 event. 

**Describe the solution you'd like**

lambda-promtail [supports these events](https://github.com/grafana/loki/blob/b57d260ddc3e3bbd8caf8286e82da5421069f717/tools/lambda-promtail/lambda-promtail/main.go#L135
): 

* events.S3Event
* events.S3TestEvent
* events.CloudwatchLogsEvent
* events.KinesisEvent
* events.SQSEvent
* events.SNSEvent

Unfortunately, EB is not one of them. I did try to send an EB event; the Lambda produces an "invalid event" error, followed by the data outlined in the EB link below. 

* [S3 Notification Content Structure](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html)
* [EventBridge Event Message Structure](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html)

Would it be possible to add EB event support to lambda-promtail? Perhaps the current event types could be used to "translate" the EB event. A lot of the event properties are similar, or can be transposed. Basically, take the EB event, translate it, then feed it into the correct event sub-type: S3, CW, etc.

**Describe alternatives you've considered**

More or less outlined in the first section.

* Use `AWS::Serverless::Function`; the S3 bucket must be in the same template, which is not always possible or ideal.
[...]

PR #10449, opened 2023-09-04
Add support for EventBridge s3 events in lambda-promtail

thepalbi:
Lambda promtail supports s3 events, which are used for scraping several log sources such as ALB access logs. This works by configuring at the S3 bucket level "s3 event notification", that are configured to target the lambda deployment of lambda-promtail.

However, if one is configuring this through CloudFormation, there's a known issue with AWS that doesn't allow to configure both the lambda, the bucket, and the notifications in the same stack. See [this issue](https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/79) for details.

For that, AWS introduced EventBridge notifications, which can be used to ship s3 events to a lambda deployment as well. This flow looks like:
s3 -> eventbridge bus -> eventbridge rule -> lambda

EventBridge has it's own message structure for s3 notifications. This PR adds a translation layer, just for `Object created` events (since they are the only ones we should take into account), so that EventBridge events can be received, and trigger the lambda as if they were from s3.

Fixes https://github.com/grafana/loki/issues/10209

- [x] Pending testing this with an actual deployment of the s3 -> event bridge -> lambda flow
- [x] ~~Add CF template for the `s3 -> event bridge -> lambda` deployment~~ Follow up PR

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/tools/lambda-promtail/lambda-promtail/eventbridge.go
+++ b/tools/lambda-promtail/lambda-promtail/eventbridge.go
@@ -26,24 +26,36 @@ type S3ObjectDetail struct {
 	Sequencer string `json:"sequencer"`
 }
 
-func processEventBridgeEvent(ctx context.Context, ev *events.CloudWatchEvent, pc Client, log *log.Logger) error {
+type s3EventProcessor func(ctx context.Context, ev *events.S3Event, pc Client, log *log.Logger) error
+
+func processEventBridgeEvent(ctx context.Context, ev *events.CloudWatchEvent, pc Client, log *log.Logger, process s3EventProcessor) error {
 	// lambda-promtail should only be used with S3 object creation events, since those indicate that a new file has been
 	// added to bucket, and need to be fetched and parsed accordingly.
 	if !(ev.Source == "aws.s3" && ev.DetailType == "Object Created") {
 		return fmt.Errorf("event bridge event type not supported")
 	}
 
-	var s3Detail S3Detail
-	if err := json.Unmarshal(ev.Detail, &s3Detail); err != nil {
+	var eventDetail S3Detail
+	if err := json.Unmarshal(ev.Detail, &eventDetail); err != nil {
 		return err
 	}
 
 	// TODO(thepalbi): how to fill bucket owner?
-	infos := []S3EventInfo{{
-		ObjectKey:    s3Detail.Object.Key,
-		BucketName:   s3Detail.Bucket.Name,
-		BucketRegion: ev.Region,
-	}}
+	var s3Event = events.S3Event{
+		Records: []events.S3EventRecord{
+			{
+				AWSRegion: ev.Region,
+				S3: events.S3Entity{
+					Bucket: events.S3Bucket{
+						Name: eventDetail.Bucket.Name,
+					},
+					Object: events.S3Object{
+						Key: eventDetail.Object.Key,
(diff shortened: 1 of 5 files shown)


tools/lambda-promtail/lambda-promtail/eventbridge.go
```diff
@@ -0,0 +1,61 @@
...
+	if !(ev.Source == "aws.s3" && ev.DetailType == "Object Created") {
+		return fmt.Errorf("event bridge event type not supported")
+	}
+
+	var eventDetail S3Detail
+	if err := json.Unmarshal(ev.Detail, &eventDetail); err != nil {
+		return err
+	}
+
+	// TODO(thepalbi): how to fill bucket owner?
```
thepalbi, line 43: the current logic extracts [this other field](https://github.com/grafana/loki/blob/4cd76919100fe59d41bd832bb07518f267d4ea1d/tools/lambda-promtail/lambda-promtail/s3.go#L208) from the original s3 events (the owner's principal ID), but this ends up not being used. See [here](https://github.com/grafana/loki/blob/4cd76919100fe59d41bd832bb07518f267d4ea1d/tools/lambda-promtail/lambda-promtail/s3.go#L248).

EventBridge events don't include the principal ID, but include the bucket AWS acct id, which is the value that should be used in [here](https://github.com/grafana/loki/blob/4cd76919100fe59d41bd832bb07518f267d4ea1d/tools/lambda-promtail/lambda-promtail/s3.go#L248):

See doc below in the field where it was supposed to be used in:
```
	// The account ID of the expected bucket owner. If the bucket is owned by a
	// different account, the request will fail with an HTTP 403 (Access Denied) error.
	ExpectedBucketOwner *string
```

cstyan:
@thepalbi should we provide updated terraform and cloudformation examples with this?

thepalbi:
> @thepalbi should we provide updated terraform and cloudformation examples with this?

So the CloudFormation/terraform templates we provide are just for configuring the basic use case of deploying the lambda, IAM roles and a CloudWatch subscription filter (which makes AWS forward logs from a CW Log group into the Lambda).

This PR is just for tackling a really specific use case, which is allowing the lambda to process logs from s3 (which for example, is the only method of logging for application load balancer access logs), using EventBridge as the middle man to notify the lambda when a file get's added to S3. EventBridge is actually and extra-hop, to tackle a limitation CloudFormation has when defining where a bucket should send notifications. I'm almost certain no such limitation exists in Terraform, but let me double check.

Either way, I think it would be nice to have a separate CloudFormation template for this use case, since it's a known problem. I'll do the testing I have to do with CloudFormation, and add the template in the PR.

cstyan (approved):
We can also add examples for tf and cf in a follow up PR, I'll let you decide.

thepalbi:
@cstyan tested on a real scenario with: `application LB -> access logging -> s3 -> notification to EventBridge -> EB -> lambda promtail -> Loki`, and working as expected:

image

[Here's](https://github.com/grafana/loki/compare/pablo/lambda-promtail-event-bridge-support...pablo/lambda-promtail-event-bridge-setup?expand=1) the edited CF template, which I'll add as a separate one, alongside with some docs, in a follow up PR.

PR #10462, opened 2023-09-05
[Style updates] Capitalize Loki when referring to project.

JStickler:
Went through and looked for places where Loki should be capitalized because we are talking about the product or project (vs. in configuration examples, etc. where lower case is acceptable or required).


MichelHollands (approved):
LGTM

PR #10464, opened 2023-09-06
lambda-promtail: Enable assume role to get from S3

lpugoy:
In our organization, direct cross account access to S3 buckets is not allowed. We need to assume a role in the account that owns the S3 buckets to download from them.

In the Terraform, the policy to download from the S3 bucket is not created when an assume role is given. Rather, a policy to enable role assumption is added.

Management of the assume role is not added to the Terraform. It is assumed that the permissions to download from the S3 bucket and the trust policies to allow the Lambda to assume it will be managed separately.

This also assumes that a single role is assumed to download from all the S3 buckets. It might be more correct to set a role for each bucket, including none for direct access, but that could complicate the solution.

The IAM policy to download from the S3 bucket is not created in the CloudFormation template, so I also didn't modify it.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 #### Promtail
 
 ##### Enhancements
+* [10464](https://github.com/grafana/loki/pull/10464) **lpugoy**: Lambda-Promtail: Enable assume role to get from S3
 * [10416](https://github.com/grafana/loki/pull/10416) **lpugoy**: Lambda-Promtail: Add support for WAF logs in S3
 * [10301](https://github.com/grafana/loki/pull/10301) **wildum**: users can now define `additional_fields` in cloudflare configuration.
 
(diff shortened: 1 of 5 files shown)


cstyan:
I'm sorry, I don't know enough about s3 access roles right now to determine if this is something lambda-promtail should be doing. Hopefully someone else from the team can help here.

lpugoy:
Updated the PR to use the [documented](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/credentials/stscreds) way of getting credentials when assuming a role.

cstyan:
@lpugoy sorry, I still don't think I know enough about s3 and roles to know whether including this in code makes sense. Is there precedent for doing so within other software that you're aware of?

lpugoy:
Hi @cstyan . I am not aware of other scenarios with the same workflow, I am not sure if this is even best practice with regards to accessing S3 buckets cross account. This may just be limited to our organization, I just opened this in case others may find it useful.

JStickler:
@lpugoy as of https://github.com/grafana/loki/pull/18531, the lambda-promtail code has moved to its own repo as another team is taking ownership of it. Could you re-open your PR against the [new lambda promtail repo](https://github.com/grafana/lambda-promtail) please?

PR #10435, opened 2023-09-04
Chore(storage): Simplify store interfaces and abstractions (pt 1)

chaudum:
**What this PR does / why we need it**:

In an attempt to simplify the store abstractions/interfaces, this PR does the following cleanups:

* Remove duplicate store field in `pkg/storage.*store`
* Remove duplicate index.Filterable interface inheritance
* Remove nesting from CompositeStore
* Rename/move ChunkReader to ./pkg/storage.SelectStore

**Checklist**
- [x] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated
  - [ ] If the change is worth mentioning in the release notes, add `add-to-release-notes` label
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/setup/upgrade/_index.md`
- [ ] For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)

PR #10465, opened 2023-09-06
Make the Store field of the Loki struct an interface type

chaudum:

Applications that extends the `Loki` struct must be able to also override the `Store`. In order to allow this, the field must be an interface type, so any `storage.Store` implementation can be assigned.

This reverts the change from interface type to struct type made in PR #10435.


ashwanthgoli (approved):
lgtm

PR #10519, opened 2023-09-08
helm: Add complete object schema to single-binary volumeClaimTemplate to avoid synchronization issues

ceriath:
This PR completes the object schema for the volumeClaim template inside the single-binary helm chart. Currently it omits the apiVersion and the kind. 
This is working, because kubernetes auto-fills those values, however it is causing an issue when you are trying to synchronize your helm chart with your deployment (in my case with argoCD). It will be shown as out of sync, because kubernetes has the fields, while the chart does not. 
Syncing works, because kubernetes accepts it without apiVersion and kind, but adds it itself again. That results in an endless sync-loop, that could be avoided by adding these to the helm template.

Since both values are already filled in by kubernetes, this will not break anything nor result in any diff for existing deployments.   

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


ceriath:
Sorry for the mess with the force-pushes, i created this PR literally during the release of 5.20.0, which messed up the whole versioning...

ceriath:
@MichelHollands Thanks! I updated it to v5.25.0

MichelHollands:
@ceriath Can you fix the merge conflicts? Once that is done I'll merge it.

MichelHollands (approved):

PR #10550, opened 2023-09-12
Fix regression when parsing numbers in Push request

salvacorts:
Even though the[ Loki HTTP API docs for the push endpoint][1] state that the stream label values should be strings, we previously didn't enforce this requirement. With https://github.com/grafana/loki/pull/9694, we started enforcing this requirement, and that broke some users.

In this PR we are reverting this type of assertion and adding a bunch of tests to avoid the regression in the future.

[1]: https://grafana.com/docs/loki/latest/reference/api/#push-log-entries-to-loki


vlad-diachenko (approved):
Looks brilliant 💎 thanks @salvacorts

PR #10563, opened 2023-09-13
Update loki artifacts to be 2.9.0.  Make TSDB the default.

### LINKED ISSUE #10266: Change config files to use TDSB as a default, like we recommend in the docs.
TODO: changing the config files to use tsdb, too. This way anyone starting from scratch is already using it

- https://github.com/grafana/loki/blob/42b8a6cbcaf8211c0344c0f8e7d00a1617c911bf/cmd/loki/loki-local-config.yaml#L26-L34
- https://github.com/grafana/loki/blob/42b8a6cbcaf8211c0344c0f8e7d00a1617c911bf/cmd/loki/loki-docker-config.yaml#L18-L25

_Originally posted by @MasslessParticle in https://github.com/grafana/loki/issues/9246#issuecomment-1680748876_

paul1r:
This PR addresses a few minor currency issues.
1) The issue brought up in Issue 10266 - that is, making the config files utilize TSDB over BoltDB.  New installs will default to the same store as recommended by the Loki documentation.
2) Update the artifacts referenced from 2.8.3 to the current release of 2.9.0

Fixes #10266

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/cmd/loki/loki-docker-config.yaml
+++ b/cmd/loki/loki-docker-config.yaml
@@ -17,7 +17,7 @@ common:
 schema_config:
   configs:
     - from: 2020-10-24
-      store: boltdb-shipper
+      store: tsdb
       object_store: filesystem
       schema: v11
       index:
--- a/cmd/loki/loki-local-config.yaml
+++ b/cmd/loki/loki-local-config.yaml
@@ -26,7 +26,7 @@ query_range:
 schema_config:
   configs:
     - from: 2020-10-24
-      store: boltdb-shipper
+      store: tsdb
       object_store: filesystem
       schema: v11
       index:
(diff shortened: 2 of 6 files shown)


cmd/loki/loki-docker-config.yaml
```diff
@@ -17,7 +17,7 @@ common:
 schema_config:
   configs:
     - from: 2020-10-24
-      store: boltdb-shipper
+      store: tsdb
       object_store: filesystem
       schema: v11
```
poyzannur, line 22: Does the date and schema need to be updated too? 
I'm not sure about `from` but i know tsdb is using `v12`

slim-bean (changes requested):
We need to add some notes to the upgrade guide around changing the schema in these default configuration files, if not we risk causing some problems for folks in ways they may not have expected.

An example of how we did this before: [2.0 upgrade nodes](https://grafana.com/docs/loki/latest/setup/upgrade/#200)

I would suggest breaking this up into 2 prs, maybe keep this one for the fix to the links and a separate for changing the config files w/ upgrade notes

PR #10567, opened 2023-09-13
Return `StepResult` from `StepEvaluator.Next()`.

jeschkies:
This change makes the `StepEvaluator` more flexible for changes like #10417 that introduce a different step result. With the interface `StepResult` we can support other vector types in the future.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/step_evaluator.go
+++ b/pkg/logql/step_evaluator.go
@@ -5,14 +5,14 @@ import (
 )
 
 type StepResult interface {
-	PromVec() promql.Vector
+	SampleVector() promql.Vector
 }
 
-type PromVec promql.Vector
+type SampleVector promql.Vector
 
-var _ StepResult = PromVec{}
+var _ StepResult = SampleVector{}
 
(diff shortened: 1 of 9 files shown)


pkg/logql/step_evaluator.go
```diff
@@ -4,10 +4,22 @@ import (
 	"github.com/prometheus/prometheus/promql"
 )
 
+type StepResult interface {
+	PromVec() promql.Vector
+}
+
+type PromVec promql.Vector
```
dannykopping, line 11: I'm wondering if we should move away from the Prometheus terminology... can we perhaps just call this `Vector` or `Samples`?

dannykopping (approved):
LGTM, just a naming suggestion nit.

pkg/logql/step_evaluator.go
jeschkies: Hm, vector is too generic as we'll have other vector implementations but we could call it `SampleVector`.

PR #6821, opened 2022-08-01
embedded-cache: Bring `fifocache` and `groupcache` into single tent.

kavirajk:
Signed-off-by: Kaviraj <kavirajkanagaraj@gmail.com>

**What this PR does / why we need it**:
This PR introduces have following changes

1. new cache type called "embedded-cache"
2. Now we bring both fifocache and groupcache into single tent called `embedded-cache`.
3. old fifocache is depreicated. New config is `embedded-cache.enabled=true && embedded-cache.distributed=false`
4. replaces `groupcache` references to embedded cache accordingly.
5. distributed embedded cache is enabled only for `results` cache.

Usage: Embedded Cache.

```yaml
embedded_cache:
    enabled: true
    distributed: true|false
```

Usage: Distributed Embedded cache (supported only for results cache for now)
```yaml
query_range:
  results_cache:
    cache:
      embedded_cache:
        enabled: true
        distributed: true`
```

or via flag
```
-frontend.embedded-cache.distributed -frontend.embedded-cache.enabled
```

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:
This is part of config cleanup before shipping groupcache feature for results cache.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] Is this an important fix or new feature? Add an entry in the `CHANGELOG.md`.
- [x] Changes that require user attention or interaction to upgrade are documented in `docs/sources/upgrading/_index.md`

PR #10620, opened 2023-09-18
cache: Enable embedded cache if no other cache is configured.

ashwanthgoli:

FIFO cache is replaced by embedded-cache in https://github.com/grafana/loki/pull/6821 
This PR enables embedded cache for query range results, chunks cache, index_stats_results and volume_results if no other cache is explicitly configured (redis, memcache).

Also removes a deprecated flag `cache-split-interval`

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/loki/config_wrapper.go
```diff
@@ -557,23 +557,18 @@ func betterTSDBShipperDefaults(cfg, defaults *ConfigWrapper, period config.Perio
...
 }
 
-// applyFIFOCacheConfig turns on FIFO cache for the chunk store, for the query range results,
-// and for the index stats results, but only if no other cache storage is configured (redis or memcache).
-// This behavior is only applied for the chunk store cache, for the query range results cache, and for
-// the index stats results (i.e: not applicable for the index queries cache or for the write dedupe cache).
-func applyFIFOCacheConfig(r *ConfigWrapper) {
+// applyEmbeddedCacheConfig turns on Embedded cache for the chunk store, query range results,
+// index stats and volume results only if no other cache storage is configured (redis or memcache).
+// Not applicable for the index queries cache or for the write dedupe cache.
```
dannykopping, line 562: > write dedupe cache

Possibly not for this PR, but I think we can remove this cache; it's been deprecated for ages.

pkg/loki/config_wrapper.go
```diff
@@ -557,23 +557,18 @@ func betterTSDBShipperDefaults(cfg, defaults *ConfigWrapper, period config.Perio
...
+		r.ChunkStoreConfig.ChunkCacheConfig.EmbeddedCache.Enabled = true
 	}
 
 	resultsCacheConfig := r.QueryRange.ResultsCacheConfig.CacheConfig
 	if !cache.IsCacheConfigured(resultsCacheConfig) {
-		r.QueryRange.ResultsCacheConfig.CacheConfig.EnableFifoCache = true
-		// The query results fifocache is still in Cortex so we couldn't change the flag defaults
-		// so instead we will override them here.
-		r.QueryRange.ResultsCacheConfig.CacheConfig.Fifocache.MaxSizeBytes = "1GB"
-		r.QueryRange.ResultsCacheConfig.CacheConfig.Fifocache.TTL = 1 * time.Hour
```
dannykopping, line 576: We might be changing behaviour here?

pkg/storage/chunk/cache/cache.go
```diff
@@ -116,24 +101,11 @@ func New(cfg Config, reg prometheus.Registerer, logger log.Logger, cacheType sta
...
-		}
-
-		if cfg.EmbeddedCache.IsEnabled() {
-			fifocfg = FifoCacheConfig{
-				MaxSizeBytes:  fmt.Sprint(cfg.EmbeddedCache.MaxSizeMB * 1e6),
-				TTL:           cfg.EmbeddedCache.TTL,
-				PurgeInterval: cfg.EmbeddedCache.PurgeInterval,
-			}
+	if cfg.EmbeddedCache.IsEnabled() {
+		fifocfg := FifoCacheConfig{
```
dannykopping, line 105: Shouldn't this be `EmbeddedCacheConfig`?

dannykopping: Not always, it seems:
https://github.com/grafana/loki/blob/e5c425eef6980604d96ed5f341598b32e706b9e6/pkg/loki/modules.go#L587-L598

Things are getting pretty weird in this space. I think we need to standardise on a single, clear approach.
I don't think we want to keep groupcache at all

pkg/loki/config_wrapper.go
ashwanthgoli: I feel `1GB` is on the higher side for single binary installations. `200MB` is also inline with what we advertise as default cache size (in docs).
https://github.com/grafana/loki/blob/835d0452d48dcadd85f9672477f511e435e3ddc4/pkg/storage/chunk/cache/embeddedcache.go#L25

pkg/storage/chunk/cache/cache.go
ashwanthgoli: ``` but the former is now removed``` 
sorry correction. I meant to say groupcache is now removed.

pkg/loki/config_wrapper.go
dannykopping: I agree with you, but my point here is that users may be unknowingly relying on this setting. We should at least document the behaviour change. Looks like we define 100MB by default not 200MB though?

ashwanthgoli: added a note in upgrade guide https://github.com/grafana/loki/pull/10620/commits/bdb20e67469dc8348484e1a5efe86edf20c1207a

JStickler (approved):
[docs team] Docs LGTM

pkg/storage/chunk/cache/cache.go
dannykopping: This is still a bit weird for me. `EmbeddedCacheConfig` and `FifoCacheConfig` are almost identical, so we should prefer `EmbeddedCacheConfig` to reduce complexity. We're nominally removing fifocache in this PR but by still using its configs and internal structs.

Maybe for another PR, but I think we should rename `FifoCache` to `EmbeddedCache` and simplify this part of the code.

dannykopping (approved):
Approving to unblock, left a comment about still using `FifoCache` internally which feels weird.

pkg/querier/queryrange/roundtrip_test.go
```diff
@@ -856,19 +855,13 @@ func TestNewTripperware_Caches(t *testing.T) {
...
+					require.NotNil(t, c)
 					caches = append(caches, c)
 				}
 			}
 
 			require.Equal(t, tc.numCaches, len(caches))
-
-			if tc.numCaches == 2 {
-				if tc.equalCaches {
-					require.Equal(t, caches[0], caches[1])
```
ashwanthgoli, line 867: cannot compare them as they are bound to be different instances of embedded cache.
what we want to compare are configs passed down to the cache

> results cache enabled, stats cache enabled but different
I'll raise a separate pr to improve this test

pkg/storage/chunk/cache/embeddedcache.go
```diff
@@ -1,32 +1,338 @@
...
 	f.DurationVar(&cfg.TTL, prefix+"embedded-cache.ttl", time.Hour, description+"The time to live for items in the cache before they get purged.")
-
 }
 
 func (cfg *EmbeddedCacheConfig) IsEnabled() bool {
 	return cfg.Enabled
 }
+
+// NewEmbeddedCache returns a new initialised EmbeddedCache.
+func NewEmbeddedCache(name string, cfg EmbeddedCacheConfig, reg prometheus.Registerer, logger log.Logger, cacheType stats.CacheType) *EmbeddedCache {
```
ashwanthgoli, line 90: a few changes to wire thing after renaming `fifo_cache.go/embeddedcache.go`

pkg/storage/chunk/cache/cache.go
ashwanthgoli: agree with this. pushed commits to remove/rename fifocache

docs/sources/configure/query-frontend.md
```diff
@@ -53,11 +53,11 @@ data:
 
       results_cache:
         cache:
-          # We're going to use the in-process "FIFO" cache
-          enable_fifocache: true
-          fifocache:
-            size: 1024
-            validity: 24h
+          # We're going to use the in-process embedded cache
```
chaudum, line 56: "in-process" has basically the same meaning as "embedded", right?

```suggestion
          # We're going to use the embedded cache
```

pkg/storage/chunk/cache/embeddedcache.go
```diff
@@ -1,32 +1,338 @@
...
+		entriesAdded: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Namespace:   "querier",
+			Subsystem:   "cache",
+			Name:        "added_total",
+			Help:        "The total number of Put calls on the cache",
+			ConstLabels: prometheus.Labels{"cache": name},
+		}),
+
+		entriesAddedNew: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Namespace:   "querier",
```
dannykopping, line 122: I wonder if we should take this opportunity to correct these metrics, too.
All our metrics start with `loki_` (well... except the ones that start with `cortex_`), but these don't. It also feels weird to limit the scope of this to the querier since many components can use this cache. Again, not necessarily for this PR 👍

dannykopping (approved):
LGTM, thanks a lot @ashwanthgoli!

pkg/storage/chunk/cache/embeddedcache.go
ashwanthgoli: agree, was talking to @chaudum about renaming these. i'll raise a separate pr

chaudum (approved):

PR #10634, opened 2023-09-18
Don't allow unbounded parallelism when downloading indices

MasslessParticle:
Index Gateways can stall when certain queries try to download tons of indices because. `ErrGroup` needs an explicit limit set on it. This PR sets the limit provisionally until a more configurable solution can be introduced.


trevorwhitney (commented):
LGTM, thanks for adding to multi_file_index as well

PR #10641, opened 2023-09-18
clients: update fluentd base container

isimluk:
1.14.0 is over 2 years old and the latest 1.16 continues to work just fine with loki.


isimluk:
It is all green. And it works like a charm.

isimluk:
Replaced with 10966

PR #10649, opened 2023-09-19
Unbreak foreign builds that depend on a newer golang.org/x/exp/slices

Jille:
They broke backwards compatibility (it's /exp/ after all), but now people using a newer version of slices get build errors (e.g. issue

Update to the latest version so the build will work for people using a newer version of slices again.


paul1r:
Closing this, I believe it was addressed in https://github.com/grafana/loki/pull/10843, but is no longer relevant

PR #10656, opened 2023-09-20
Log process startup and running time

chaudum:

This helps to get insights about slow startup of Loki, e.g. for the index gateway, which may pre-fetch a lot of data before it becomes ready.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -399,6 +399,8 @@ func (t *Loki) ListTargets() {
 
 // Run starts Loki running, and blocks until a Loki stops.
 func (t *Loki) Run(opts RunOpts) error {
+	startTime := time.Now()
+
 	serviceMap, err := t.ModuleManager.InitModuleServices(t.Cfg.Target...)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/loki/loki.go
```diff
@@ -399,6 +399,8 @@ func (t *Loki) ListTargets() {
 
 // Run starts Loki running, and blocks until a Loki stops.
 func (t *Loki) Run(opts RunOpts) error {
```
chaudum, line 401: If we want a more precise startup time, including some config parsing that happens prior to calling `Run()` we could start the timer in the `main()` function and pass it as a `RunOpt`.
Not sure if it really makes much difference, though.

chaudum: Actually, let me do that!

ashwanthgoli (approved):
lgtm

pkg/loki/loki.go
```diff
@@ -369,6 +368,9 @@ type RunOpts struct {
 	// CustomConfigEndpointHandlerFn is the handlerFunc to be used by the /config endpoint.
 	// If empty, default handlerFunc will be used.
 	CustomConfigEndpointHandlerFn func(http.ResponseWriter, *http.Request)
+	// StartTime is the time at which the main() function started executing.
+	// It is used to determin the startup time as well as the running time of the Loki process.
```
ashwanthgoli, line 372:
```suggestion
	// It is used to determine the startup time as well as the running time of the Loki process.
```

PR #10664, opened 2023-09-21
drone(docker-driver): Fetch tags before pushing image during release

kavirajk:
We realized, loki-docker-driver image was not being pushed after 2.8.2. And looking into it's drone pipeline and comparing it with other steps, docker-driver step is not fetching the tags,[ thus the final hash calculated to push is some HEAD branch.](https://github.com/grafana/loki/blob/main/tools/image-tag#L13-L14)

e.g: even if we have tag 2.9.1, it pushes only with latest HEAD commit hash.

This PR fixes it.
Screenshot 2023-09-21 at 08 14 09
Screenshot 2023-09-21 at 08 14 24
Screenshot 2023-09-21 at 08 14 42

Fixes #<issue number>

I would prefer this instead of https://github.com/grafana/loki/pull/10663. Because that would make fetching tags more repetitive in all the drone tasks (can also be expensive in terms of running time if we put all together)

**Checklist**

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -911,7 +911,7 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
       {
         name: 'build and push',
         image: 'grafana/loki-build-image:%s' % build_image_version,
-        depends_on: ['clone'],
+        depends_on: ['clone', 'image-tag'],
         environment: {
           DOCKER_USERNAME: { from_secret: docker_username_secret.name },
           DOCKER_PASSWORD: { from_secret: docker_password_secret.name },


.drone/drone.yml
```diff
@@ -2014,8 +2015,3 @@ get:
   path: infra/data/ci/packages-publish/gpg
 kind: secret
 name: gpg_private_key
----
```
kavirajk, line 2017: auto removed during `make drone`.

.drone/drone.jsonnet
```diff
@@ -917,6 +916,7 @@ local manifest_ecr(apps, archs) = pipeline('manifest-ecr') {
           DOCKER_PASSWORD: { from_secret: docker_password_secret.name },
         },
         commands: [
+	  'git fetch origin --tags',
```
kavirajk, line 919: tried one more approach before this.

Tried to add it as separate step say `fetch-git-tags` on this pipeline. In that case, git tags only fetched in that particular step and cannot be used during `make docker-driver-push` step. Rationale being other pipelines that do use separate step to fetch git tags, stores it in `./tools/image-tag > arch.tags` and re-use it other step.

Ended up doing this way without changing much of the pipelines.

.drone/drone.yml
kavirajk: Looks like drone is failing without it. Adding it back

dannykopping (approved):
LGTM but you'll need to update the drone signature

PR #10678, opened 2023-09-22
Jsonnet: Update the 'cluster' label in the logs and operational dashboards as currently the custom one is not picked up

liam-howe-maersk:

We use a custom label for the kubernetes cluster in our Loki set up, which we can do by making use of the config `per_cluster_label`. However, we have found that for 2 dashboards, the logs and operational ones, that the `cluster` label is not being overridden, I have added logic in their libsonnet files to override this value so that these dashboards can work for us when we store the cluster name in a different label.

N/A

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/production/loki-mixin/dashboards/loki-logs.libsonnet
+++ b/production/loki-mixin/dashboards/loki-logs.libsonnet
@@ -61,6 +61,19 @@ local template = import 'grafonnet/template.libsonnet';
                           p {
                             targets: [
                               e {
+                                expr: std.strReplace(super.expr, 'cluster="$cluster", ', $._config.per_cluster_label + '="$cluster", '),
+                              }
+                              for e in p.targets
+                            ],
+                          }
+                          for p in super.panels
+                        ],
+                      } +
+                      {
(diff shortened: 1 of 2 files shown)


liam-howe-maersk:
Another place that needed updating was within the logic to replace the cortex-gw labels, so I have updated the `replaceCortexGateway` method to also update the cluster label. When I generate the dashboard json I still find 7 cases where the `cluster` label has not been overwritten by the custom label and I can't figure out where this update needs to happen for these last 7 panels. These panels are:

- "Discarded Lines"
- "Discarded Lines Per Interval"
- "Active Streams"
- "Streams Created/Sec"
- "Chunks Flushed/Sec" - "Chunks" query
- "Chunk Size Bytes"
- "Chunk Flush Reason %"

liam-howe-maersk:
I figured out the issue with why not all were being replaced and have updated with a solution, I believe this should now be ready for review.

PR #10681, opened 2023-09-22
Update 0003-QueryFairnessInScheduler.md image references

dimitarvdimitrov:
The doc and images were moved, but the references weren't updated


JStickler:
The image references are relative, so since the files and the images are still in the same folder,  the image references do not need to be updated.  The images appear in the current docs.

LIDS_Diagrams

dimitarvdimitrov:
Screenshot 2023-09-25 at 10 10 42

unfortunately they don't render properly when viewing the raw markdown

JStickler:
>unfortunately they don't render properly when viewing the raw markdown

That is a limitation of GitHub.

PR #10687, opened 2023-09-22
[sharding] [bugfix] Correctly account for child groupings in count/min/max sharding

owen-d:
Previously, we didn't always consider that some label reductions from downstream vector|range-vector aggregations come from groupings. This is particularly useful in the min/max/count sharding mapper.

edit: also fixes an issue with the range extractor mishandling `without ()` as `by ()`


vlad-diachenko (approved):
looks awesome!

Issue #10683, opened 2023-09-22
Can't search for structured metadata


mathias-ioki:
**Describe the bug**
We want to use the new `structured metadata` feature. For that, we have configured Loki to accept this new kind of metadata and we configured Promtail to ship the structured metadata. Everything looks fine so far. But if we now add these non-indexed fields to our search query, we don't get any response. We tried logcli and Grafana Explore to search for the logs.

We are pretty sure, that these structured metadata are correctly shipped to Loki, because:
* It's possible to ship the field as label, by replacing structured_metadata with labels in promtail
* structured_metadata is the right keyword in promtail, as promtail complains when we use an unknown one
* structured metadata are send to loki by promtail, as promtail is logging an error, when we disable it again in loki by setting allow_structured_metadata to false.

But we don't know how to confirm that they are correctly stored in S3.

**To Reproduce**
1. Enable structured metadata in Loki and restart:
```
limits_config:
  allow_structured_metadata: true
```
2. Configure Promtail to add structured metadata to your logs:
```
scrape_configs:
  - job_name: docker
    pipeline_stages:
      - static_labels:
          job: docker
      - match:
          selector: '{application="appname"} |= "log message"'
          stages:
            - json:
                expressions:
                  payload:
            - json:
                expressions:
                  request_id:
                source: payload
            - structured_metadata:
                request_id:
```
3. Query: `'{application="appname"} | request_id = "123abc"'`

**Expected behavior**
The expected behavior is to get log lines, which contain the mentioned request_id. But instead of that, you get an empty response.

**Environment:**
- Loki version: 2.9.1
- Promtail version: 2.9.1
- LogCli version: 2.9.1
- Infrastructure: Cloud VMs with Read-Write mode
- Deployment tool: apt packages

luxaritas:
Also ran into this. Can be reproduced with a single post via cURL:
```
DATA_PRE='{"streams": [{ "stream": { "label": "labelvalue" }, "values": [ [ "'
TIME=$(date +%s%N)
DATA_POST='", "log line content", {"structured": "structuredvalue"} ] ] }]}'

curl -v -H "Content-Type: application/json" -XPOST -s "http://localhost:3100/loki/api/v1/push" --data-raw "$DATA_PRE$TIME$DATA_POST"
```
and then doing a logcli or grafana lookup (I also tried a grafana nightly snapshot from ~ a week ago)

vlad-diachenko:
I will check today/tomorrow

[...]

PR #10698, opened 2023-09-25
Update structured-metadata docs

vlad-diachenko:
OSS users tried to use the structured metadata feature and faced an issue because they used previous schema versions (V11, V12) and they were not aware that it was necessary to use V13 because the docs did not cover it.  

re: https://github.com/grafana/loki/issues/10683


sandeepsukhani (approved):
LGTM

salvacorts (approved):

PR #10712, opened 2023-09-26
[release-2.9.x][sharding] [bugfix] Correctly account for child groupings in count/mi…

poyzannur:
Manually backported 8e79c3a1a9280ab91b832df1d7506008bddffdc6 from https://github.com/grafana/loki/pull/10687 due to merge conflicts

---------------------

Previously, we didn't always consider that some label reductions from downstream vector|range-vector aggregations come from groupings. This is particularly useful in the min/max/count sharding mapper.

edit: also fixes an issue with the range extractor mishandling without () as by ()


poyzannur:
decided not to patch this release as the original changes were not included in 2.9.x

PR #10727, opened 2023-09-27
ingestion: native otlp ingestion support

sandeepsukhani:
Add support for natively supporting logs ingestion in OTLP format.
`/otlp/v1/logs` is the new endpoint where users can push logs in OTLP format. It accepts logs serialized in JSON or proto format. Since OTEL format is very different than what Loki storage model, here is how data in OTEL format will be mapped to Loki data model:

* Index labels: The Resource Attributes map quite well to Index labels in Loki since both usually identify the source of the logs. The problem however is that Resource attributes in OTLP can have an unbounded number of values while Loki has a default limit of having up to 30 labels. Since Index labels in Loki can largely drive the kind of querying experience the users are going to have, we have chosen select attributes which would be picked as Index Labels. The ones that are not picked up as Index labels would be stored as Structured Metadata with each log entry.
* Timestamp: LogRecord.TimeUnixNano
* LogLine: LogRecord.Body holds the body of the log. However, since Loki only supports Log body in string format, we will stringify non-string values using [AsString method from OTEL collector lib](https://github.com/open-telemetry/opentelemetry-collector/blob/ab3d6c5b64701e690aaa340b0a63f443ff22c1f0/pdata/pcommon/value.go#L353).
* Structured Metadata: Anything which can’t be stored in Index labels and LogLine. Here is a non-exhaustive list of what will be stored in Structured Metadata to give a sense of what it will hold:
  * Resource Attributes not stored as Index labels is replicated and stored with each log entry.
  * Everything under InstrumentationScope is replicated and stored with each log entry.
  * Everything under LogRecord except LogRecord.Body, LogRecord.TimeUnixNano and sometimes LogRecord.ObservedTimestamp.

*NOTES*:
* Since Loki does not support `.` or any other special characters other than `_` in label names, we replace all non-supported characters with `_`.
* Since Loki only supports string in values of Index Labels and Structured Metadata, all the complex types are converted as follows:
  * Map would be flattened into label keys using `_` as separator, same as how we do it in [json parser in LogQL](https://grafana.com/docs/loki/latest/query/log_queries/#json).
  * Everything else is stringified using [AsString method from OTEL collector lib](https://github.com/open-telemetry/opentelemetry-collector/blob/ab3d6c5b64701e690aaa340b0a63f443ff22c1f0/pdata/pcommon/value.go#L353)

I will open follow-up PRs for:
* Documentation
* Make blessed attributes list configurable per tenant.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/loghttp/push/otlp.go
```diff
@@ -0,0 +1,322 @@
...
+		streamLabels := model.LabelSet{}
+		flattenedResourceAttributes := labels.NewBuilder(logproto.FromLabelAdaptersToLabels(flattenAttributes(rls.At(i).Resource().Attributes(), "")))
+		// service.name is a required Resource Attribute. If it is not present, we will set it to "unknown_service".
+		if flattenedResourceAttributes.Get("service_name") == "" {
+			flattenedResourceAttributes = flattenedResourceAttributes.Set("service_name", "unknown_service")
+		}
+
+		// copy blessed attributes to stream labels
+		for _, ba := range blessedAttributes {
+			normalizedBlessedAttribute := prometheustranslator.NormalizeLabel(ba)
```
vlad-diachenko, line 128: we do not need to normalize it at runtime, we can generate normalizedBlessedAttributes on startup

pkg/loghttp/push/otlp.go
```diff
@@ -0,0 +1,322 @@
...
+
+			// remove the blessed attributes copied to stream labels
+			flattenedResourceAttributes.Del(normalizedBlessedAttribute)
+		}
+
+		if err := streamLabels.Validate(); err != nil {
+			stats.errs = append(stats.errs, fmt.Errorf("invalid labels: %w", err))
+			continue
+		}
+		labelsStr := streamLabels.String()
```
vlad-diachenko, line 143: I believe we could skip converting `streamLabels` to string and parsing it back `syntax.ParseLabels(labelsStr)` . we could just iterate over the key<>values of streamLabels and create `labels.Labels` . wdyt?

integration/client/client.go
```diff
@@ -177,6 +169,54 @@ func (c *Client) pushLogLine(line string, timestamp time.Time, structuredMetadat
...
+	}
+
+	if res.StatusCode/100 == 2 {
+		defer res.Body.Close()
+		return nil
+	}
+
+	buf, err := io.ReadAll(res.Body)
+	if err != nil {
+		return fmt.Errorf("reading request failed with status code %v: %w", res.StatusCode, err)
```
vlad-diachenko, line 214:
```suggestion
		return fmt.Errorf("reading response failed for status code %v: %w", res.StatusCode, err)
```

integration/client/client.go
```diff
@@ -177,6 +169,54 @@ func (c *Client) pushLogLine(line string, timestamp time.Time, structuredMetadat
...
+	}
+
+	return fmt.Errorf("request failed with status code %v: %w", res.StatusCode, errors.New(string(buf)))
```
vlad-diachenko, line 217: what is the difference between these two formats? I believe we can just skip creating error from the string and use the string directly in the error message
```suggestion
	return fmt.Errorf("request failed with status code %v: %s", res.StatusCode, string(buf))
```

integration/loki_micro_services_test.go
```diff
@@ -400,12 +398,12 @@ func TestMicroServicesIngestQueryOverMultipleBucketSingleProvider(t *testing.T)
...
-				require.NoError(t, cliDistributor.PushLogLineWithTimestampAndStructuredMetadata("lineA", time.Now().Add(-48*time.Hour), map[string]string{"traceID": "123"}, map[string]string{"job": "fake"}))
-				require.NoError(t, cliDistributor.PushLogLineWithTimestampAndStructuredMetadata("lineB", time.Now().Add(-36*time.Hour), map[string]string{"traceID": "456"}, map[string]string{"job": "fake"}))
+				require.NoError(t, cliDistributor.PushLogLine("lineA", time.Now().Add(-48*time.Hour), map[string]string{"traceID": "123"}, map[string]string{"job": "fake"}))
+				require.NoError(t, cliDistributor.PushLogLine("lineB", time.Now().Add(-36*time.Hour), map[string]string{"traceID": "456"}, map[string]string{"job": "fake"}))
 
 				// ingest logs to the current period
-				require.NoError(t, cliDistributor.PushLogLineWithStructuredMetadata("lineC", map[string]string{"traceID": "789"}, map[string]string{"job": "fake"}))
-				require.NoError(t, cliDistributor.PushLogLineWithStructuredMetadata("lineD", map[string]string{"traceID": "123"}, map[string]string{"job": "fake"}))
+				require.NoError(t, cliDistributor.PushLogLine("lineC", now, nil, map[string]string{"job": "fake"}))
```
vlad-diachenko, line 405: why did we remove structured metadata from these log lines? 
`map[string]string{"traceID": "123"}`

integration/loki_micro_services_test.go
```diff
@@ -708,6 +706,128 @@ func TestQueryTSDB_WithCachedPostings(t *testing.T) {
...
+				require.Len(t, stream.Values, 1)
+				require.Equal(t, "lineC", stream.Values[0][1])
+				require.Equal(t, map[string]string{
+					"service_name": "varlog",
+					"order_ids":    "[5,6]",
+				}, stream.Stream)
+			default:
+				t.Errorf("unexpected case %d", i)
+			}
+		}
```
vlad-diachenko, line 827: these assertions do not make sure that there are exactly 3 log lines in the response. If the response has only 1 line it will assert the first line and will mark the test as passed

vlad-diachenko (approved):
Looks great! minor comments

pkg/loghttp/push/otlp_test.go
```diff
@@ -0,0 +1,444 @@
...
+						Name:  "dropped_attributes_count",
+						Value: "1",
+					},
+					{
+						Name:  "flags",
+						Value: fmt.Sprintf("%d", plog.DefaultLogRecordFlags.WithIsSampled(true)),
+					},
+					{
+						Name:  "trace_id",
+						Value: "12345678123456781234567812345678",
```
salvacorts, line 316: It's a bit difficult to see how the traceID set in the payload maps to this output. Maybe we can have the traceID as a variable and call traceID.String() here?

pkg/loghttp/push/otlp.go
```diff
@@ -0,0 +1,373 @@
...
+		structuredMetadata = append(structuredMetadata, push.LabelAdapter{
+			Name:  "flags",
+			Value: fmt.Sprintf("%d", logRecordFlags),
+		})
+	}
+
+	if traceID := log.TraceID(); !traceID.IsEmpty() {
+		structuredMetadata = append(structuredMetadata, push.LabelAdapter{
+			Name:  "trace_id",
+			Value: hex.EncodeToString(traceID[:]),
```
salvacorts, line 298: We can call `traceID.String()` here

pkg/loghttp/push/otlp.go
```diff
@@ -0,0 +1,373 @@
...
+		})
+	}
+	if spanID := log.SpanID(); !spanID.IsEmpty() {
+		structuredMetadata = append(structuredMetadata, push.LabelAdapter{
+			Name:  "span_id",
+			Value: hex.EncodeToString(spanID[:]),
```
salvacorts, line 304: We can call `spanID.String()` here

pkg/loki/modules.go
```diff
@@ -321,19 +321,25 @@ func (t *Loki) initDistributor() (services.Service, error) {
 		tenant.WithDefaultResolver(tenant.NewMultiResolver())
 	}
 
-	pushHandler := middleware.Merge(
+	httpPushHandler := middleware.Merge(
```
salvacorts, line 324: Same comment wrt naming. I'd change it for `lokiPushHandler`

salvacorts: In fact provided they will likely always have the same middlewares I'd do:
```go
httpPushHandlerMiddleware := middleware.Merge(
    serverutil.RecoveryHTTPMiddleware,
    t.HTTPAuthMiddleware,
)
lokiPushHandler := httpPushHandlerMiddleware.Wrap(http.HandlerFunc(t.distributor.PushHandler))
otlpPushHandler := httpPushHandlerMiddleware.Wrap(http.HandlerFunc(t.distributor.OTLPPushHandler))
```

salvacorts (commented):
Awesome work Sandeep! I left some comments.

pkg/loghttp/push/otlp.go
```diff
@@ -0,0 +1,322 @@
...
+	"github.com/grafana/loki/pkg/logql/syntax"
+	"github.com/grafana/loki/pkg/push"
+)
+
+const (
+	pbContentType       = "application/x-protobuf"
+	gzipContentEncoding = "gzip"
+)
+
+var blessedAttributes = []string{
```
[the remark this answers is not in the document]
salvacorts: This'll be done on a followup PR

pkg/loghttp/push/otlp.go
```diff
@@ -0,0 +1,373 @@
...
+			flattenedResourceAttributes = flattenedResourceAttributes.Set("service_name", "unknown_service")
+		}
+
+		if dac := res.DroppedAttributesCount(); dac != 0 {
+			flattenedResourceAttributes = flattenedResourceAttributes.Set("resource_dropped_attributes_count", fmt.Sprintf("%d", dac))
+		}
+
+		// copy blessed attributes to stream labels
+		streamLabels := make(model.LabelSet, len(blessedAttributesNormalized))
+		for _, ba := range blessedAttributesNormalized {
```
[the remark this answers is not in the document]
salvacorts: The blessed attributes should be an ordered list by preference.

integration/client/client.go
sandeepsukhani: `%w` is used for wrapping the errors, which the consumer can then unwrap. I just copied the code from the other push function. Wrapping the error here doesn't make sense, so I will change it.

integration/loki_micro_services_test.go
sandeepsukhani: good catch, it was a mistake. Fixed it.

pkg/loghttp/push/otlp.go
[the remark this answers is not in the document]
sandeepsukhani: We are not going to enforce `max_label_names_per_series` limit here. The limits would be enforced later at the common code path which processed the parsed push requests. It would be the user's responsibility to set appropriate blessed attributes(we are going to make it configurable) to avoid rejecting data when the label count goes beyond the limits.

pkg/loghttp/push/otlp.go
```diff
@@ -0,0 +1,373 @@
...
+
+			// it would be rare to have multiple scopes so if the entries slice is empty, pre-allocate it for the number of log entries
+			if cap(pushRequestsByStream[labelsStr].Entries) == 0 {
+				stream := pushRequestsByStream[labelsStr]
+				stream.Entries = make([]push.Entry, 0, logs.Len())
+				pushRequestsByStream[labelsStr] = stream
+			}
+
+			// use fields and attributes from scope as structured metadata
+			scopeAttributesAsStructuredMetadata := attributesToLabels(scope.Attributes(), "")
```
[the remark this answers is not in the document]
sandeepsukhani: there is no way for the caller to know how big the slice has to be since the flattening of attributes could result in more elements when there are nested attributes. Also, passing that extra argument complicates the code. I would love to keep it simple and optimize it later if it happens to cause any with real usage.

pkg/loghttp/push/otlp_test.go
sandeepsukhani: I am not sure if that would be an ideal test. I actually copied the test case from [otel repo](https://github.com/open-telemetry/opentelemetry-collector/blob/287b98f6973fd6baa278150b9fca8c83abea0af4/pdata/pcommon/traceid_test.go#L29).

pkg/loghttp/push/otlp.go
sandeepsukhani: It is not ideal to call `String()` as recommended by [otel lib](https://github.com/open-telemetry/opentelemetry-collector/blob/287b98f6973fd6baa278150b9fca8c83abea0af4/pdata/pcommon/traceid.go#L22-L26).

pkg/loghttp/push/otlp.go
sandeepsukhani: same here

salvacorts (approved):

PR #10733, opened 2023-09-28
Support case insensetive logql keywords

### LINKED ISSUE #10638: Support case-insensitive LogQL keywords
Loki's keywords & functions are currently case-sensitive.

Invalid:

```logql
count_over_time({app="distributor"}[$__auto]) > 1
AND 
count_over_time({app="distributor"}[$__auto]) < 100
```

Produces: `parse error at line 2, col 1: syntax error: unexpected IDENTIFIER`

Valid:

```logql
count_over_time({app="distributor"}[$__auto]) > 1
and 
count_over_time({app="distributor"}[$__auto]) < 100
```

Folks coming from a predominantly SQL background might find writing the above query as the following to read better:

```logql
COUNT_OVER_TIME({app="distributor"}[$__auto]) > 1
AND 
COUNT_OVER_TIME({app="distributor"}[$__auto]) < 100
```

At the very least, improve the error message.

shantanualsi:
Adds support for case-insensetive logql queries that fixes #10638.

https://github.com/grafana/loki/issues/10638

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated


pkg/logql/syntax/lex_test.go
```diff
@@ -68,13 +68,16 @@ func TestLex(t *testing.T) {
 		{`count_over_time({foo="bar"}[5m])`, []int{COUNT_OVER_TIME, OPEN_PARENTHESIS, OPEN_BRACE, IDENTIFIER, EQ, STRING, CLOSE_BRACE, RANGE, CLOSE_PARENTHESIS}},
 		{`count_over_time({foo="bar"} |~ "\\w+" | unwrap foo[5m])`, []int{COUNT_OVER_TIME, OPEN_PARENTHESIS, OPEN_BRACE, IDENTIFIER, EQ, STRING, CLOSE_BRACE, PIPE_MATCH, STRING, PIPE, UNWRAP, IDENTIFIER, RANGE, CLOSE_PARENTHESIS}},
 		{`sum(count_over_time({foo="bar"}[5m])) by (foo,bar)`, []int{SUM, OPEN_PARENTHESIS, COUNT_OVER_TIME, OPEN_PARENTHESIS, OPEN_BRACE, IDENTIFIER, EQ, STRING, CLOSE_BRACE, RANGE, CLOSE_PARENTHESIS, CLOSE_PARENTHESIS, BY, OPEN_PARENTHESIS, IDENTIFIER, COMMA, IDENTIFIER, CLOSE_PARENTHESIS}},
+		{`SUM(Count_Over_Time({foo="bar"}[5m])) BY (foo,bar)`, []int{SUM, OPEN_PARENTHESIS, COUNT_OVER_TIME, OPEN_PARENTHESIS, OPEN_BRACE, IDENTIFIER, EQ, STRING, CLOSE_BRACE, RANGE, CLOSE_PARENTHESIS, CLOSE_PARENTHESIS, BY, OPEN_PARENTHESIS, IDENTIFIER, COMMA, IDENTIFIER, CLOSE_PARENTHESIS}},
```
ashwanthgoli, line 71: should we only support either all caps or all nocaps?

pkg/logql/syntax/lex.go
```diff
@@ -226,7 +228,7 @@ func (l *lexer) Lex(lval *exprSymType) int {
 		}
 	}
 
-	if tok, ok := functionTokens[tokenText]; ok {
+	if tok, ok := functionTokens[tokenTextLower]; ok {
```
ashwanthgoli, line 231: `isFunction` also needs a check for `by/without` tokens

pkg/logql/syntax/lex.go
```diff
@@ -215,7 +215,9 @@ func (l *lexer) Lex(lval *exprSymType) int {
 	}
 
 	tokenText := l.TokenText()
-	tokenNext := tokenText + string(l.Peek())
+	tokenTextLower := strings.ToLower(l.TokenText())
+	tokenNext := strings.ToLower(tokenText + string(l.Peek()))
```
ashwanthgoli, line 219: i don't know why we peek here for functions, next token might also need to be covered I think

shantanualsi: not sure either, I was also wondering about it.. I just added the toLower func to the existing implementation

pkg/logql/syntax/lex.go
shantanualsi: can you explain what you mean here exactly? they should be covered (see test case line 71)

pkg/logql/syntax/lex_test.go
shantanualsi: hmm.. not sure about this nuance! need to check with @dannykopping

pkg/logql/syntax/lex.go
ashwanthgoli: sorry I wasn't clear. I think `isFunction` would return false for `SUM BY (foo)` since we only check against lower case `by` and `without`
https://github.com/grafana/loki/blob/4c3e8f5958d8251bfba2ae2e807f1a75f7bdc01f/pkg/logql/syntax/lex.go#L396

shantanualsi: good catch! but `isFunction` is called only if `functionTokens[tokenTextLower]` is `ok`. But in case of by and without, it would never be called.. so now I'm wondering why is by and without a separate case in `isFunction`

pkg/logql/syntax/lex_test.go
dannykopping: I can't think of a reason why we'd want to enforce a single case over the whole function.
Stylistically it's ¯\_(ツ)_/¯ if folks `waNt_To_qUeRy_liKE_tHIS()` but at least it's not case-sensitive.

pkg/logql/syntax/lex.go
shantanualsi: good catch!

shantanualsi: fixed this and added a test case to cover this. please verify.

PR #10748, opened 2023-09-29
Add service structure for bloom-compactor

poyzannur:
Initial service structure for bloom-compactor component

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -0,0 +1,39 @@
...
+
+type Compactor struct {
+	services.Service
+
+	cfg    Config
+	logger log.Logger
+}
+
+//TODO do we want to configure retention in the bloom compactors?
+//var (
```
owen-d, line 18: Probably fine to start without retention 👍

pkg/bloomcompactor/config.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// RegisterFlags registers flags for the Bloom-Compactor configuration.
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	cfg.Ring.RegisterFlags("bloom-compactor.", "collectors/", f)
+	f.BoolVar(&cfg.Enabled, "bloom-compactor.enabled", false, "Flag to enable or disable the usage of the bloom-compactor component.")
+}
+
+// RingCfg is a wrapper for our internally used ring configuration plus the replication factor.
+type RingCfg struct {
```
owen-d, line 26: I think we can just use `util.RingConfig` rather than redefining another type here. I also don't think we'll need replication factors in the bloom compactors.

pkg/bloomcompactor/ringmanager.go
```diff
@@ -0,0 +1,206 @@
...
+	ringNameForServer = "bloom-compactor"
+	// start with a single instance
+	ringNumTokens   = 1
+	ringCheckPeriod = 3 * time.Second
+
+	// ringKey is the key under which we register different instances of bloom-compactor in the KVStore.
+	ringKey = "bloom-compactor"
+)
+
+type RingManager struct {
```
owen-d, line 32: can we reuse one of the other implementations of ringmanager instead of recreating it here? If the logic is the same, it feels better+more extensible to reuse.

owen-d (approved):
Left some suggestions but approving. Wiring looks fine but would be nice if we can reuse rather than copy/paste in a few places.

pkg/loki/loki.go
```diff
@@ -621,9 +627,11 @@ func (t *Loki) setupModuleManager() error {
 		TableManager:             {Server, Analytics},
 		Compactor:                {Server, Overrides, MemberlistKV, Analytics},
 		IndexGateway:             {Server, Store, Overrides, Analytics, MemberlistKV, IndexGatewayRing, IndexGatewayInterceptors},
+		BloomCompactor:           {Server, BloomCompactorRing, Analytics},
 		IngesterQuerier:          {Ring},
 		QuerySchedulerRing:       {Overrides, Server, MemberlistKV},
 		IndexGatewayRing:         {Overrides, Server, MemberlistKV},
+		BloomCompactorRing:       {Overrides, MemberlistKV},
```
salvacorts, line 634: `IndexGatewayRing` adds the `Server` dependency despite the `IndexGateway` already having the  `Server` dependency. Should `BloomCompactorRing` also add `Server` as well?

poyzannur: I see your point. But I can't justify why they were added in the first place? They already have `MemberlistKV` which has `Server` dependency. i prefer we remove it from the others as well, unless there's another reason i may be missing. wdyt?

pkg/storage/stores/shipper/bloomshipper/client.go
```diff
@@ -68,7 +68,7 @@ type MetaClient interface {
 	// Returns all metas that are within MinFingerprint-MaxFingerprint fingerprint range
 	// and intersect time period from StartTimestamp to EndTimestamp.
 	GetMetas(ctx context.Context, params MetaSearchParams) ([]Meta, error)
-	PutMeta(ctx context.Context, meta Meta) (Meta, error)
+	PutMeta(ctx context.Context, meta Meta) error
```
poyzannur, line 71: @vlad-diachenko the `PutObject` method that is called in the implementation of `PutMeta` does not return the object back. Are you planning to fetch the object after putting? 
I removed the Meta from return value to make life easier for this PR. Hope that's ok.

pkg/bloomcompactor/config.go
poyzannur: we all decided to extract ring code in a separate PR for all these new services.

pkg/loki/loki.go
cyriltovena: I think you should add everything you need just in case a transient dependency changes ? but anyway this is not thing that changes much often so 🤷 not important.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -0,0 +1,241 @@
...
+	//
+
+	client, err := bloomshipper.NewBloomClient(periodConfigs, storageCfg, clientMetrics)
+	if err != nil {
+		return nil, err
+	}
+
+	shipper, err := bloomshipper.NewShipper(
+		client,
+		bloomshipperconfig.Config{WorkingDirectory: cfg.WorkingDirectory},
```
chaudum, line 84:
```suggestion
		storageCfg.BloomShipperConfig,
```

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -0,0 +1,241 @@
...
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	store, err := bloomshipper.NewBloomStore(*shipper)
```
chaudum, line 90: NewBloomStore should take a pointer

```suggestion
	store, err := bloomshipper.NewBloomStore(shipper)
```

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -0,0 +1,241 @@
...
+	services.Service
+
+	cfg                Config
+	logger             log.Logger
+	bloomCompactorRing ring.ReadRing
+	periodConfigs      []config.PeriodConfig
+
+	// temporary workaround until store has implemented read/write shipper interface
+	bloomShipperClient bloomshipper.Client
+	bloomStore         bloomshipper.Store
```
chaudum, line 56: Once you rebased, you can get rid of these.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -0,0 +1,241 @@
...
+	c := &Compactor{
+		cfg:                cfg,
+		logger:             logger,
+		bloomCompactorRing: readRing,
+		periodConfigs:      periodConfigs,
+	}
+
+	//objectClient, err := storage.NewObjectClient(storageCfg.TSDBShipperConfig.SharedStoreType, storageCfg, clientMetrics)
+	//if err != nil { return nil, err}
+	//
```
chaudum, line 75: Can be removed.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -0,0 +1,241 @@
...
+
+	store, err := bloomshipper.NewBloomStore(*shipper)
+	if err != nil {
+		return nil, err
+	}
+
+	// temporary workaround until store has implemented read/write shipper interface
+	c.bloomShipperClient = client
+	c.bloomStore = store
+	c.Service = services.NewIdleService(c.starting, c.stopping)
```
chaudum, line 98: We probably don't want an IdleService, but a service that runs a job in an interval. It's ok to leave it for the scaffolding, though.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -0,0 +1,241 @@
...
+
+func (c *Compactor) runCompact(ctx context.Context) error {
+	//TODO set MaxLookBackPeriod to Max ingester accepts
+	maxLookBackPeriod := c.cfg.MaxLookBackPeriod
+
+	st_fp, end_fp := NoopGetFingerprintRange()
+	tenantId := NoopGetUserId()
+
+	end := time.Now().UTC().UnixMilli()
+	start := end - maxLookBackPeriod.Milliseconds()
```
chaudum, line 197: Our timestamps are nanosecond precision:

```suggestion
	end := time.Now().UTC().Nanos()
	start := end - maxLookBackPeriod.Nanos()
```

owen-d (approved):
Approach looks good. Looks like the next pieces we need to unblock you are building blooms and bloom-blocks from various starting points:
* adding new series (create bloom)
* appending chunks to existing series (extend bloom)

pkg/bloomcompactor/bloomcompactor.go
poyzannur: added a TODO for myself anyways

pkg/bloomcompactor/bloomcompactor.go
poyzannur: PR is quite large and all over the place atm. If i remove now, i need to add interfaces with get/putMetas functions. 
can i do that in a follow up PR?

PR #10780, opened 2023-10-03
Format + Library for building bloom filter "blocks" from chunks

owen-d:
This PR introduces both a binary format for bloom filter "blocks" and a library for building/querying them.

Each "block" consists of two files:
1) `Index`. This is a list of individually compressed pages of stream+chunk information sorted in ascending fingerprint order. Each series lists both the chunks indexed in the bloom filters as well as the relevant offsets for said series in the bloom file.
2) `Blooms`. This is a list of individually compressed pages of bloom filters for each series, again sorted in ascending fingerprint order.

There is still more work to do, but I'm opening this PR as an initial pass.

PR #10782, opened 2023-10-03
Add basic structure of bloom gateways

chaudum:
### Summary

This pull requests adds the basic structure for the new bloom gateway component.

- Adds new `bloom-gateway` target that runs with multiple instances joined by a ring
- Adds a querier and client component on the index gateway to filter chunk refs
- Adds the gRPC protobuf definitions for commication between index gateways and bloom gateways
- Adds a store component used on the bloom gateways to query binary bloom files

```

			     Querier   Query Frontend
			        |           |
			................................... service boundary
			        |           |
			        +----+------+
			             |
			     indexgateway.Gateway**
			             |
			   bloomgateway.BloomQuerier
			             |
			   bloomgateway.GatewayClient
			             |
			  logproto.BloomGatewayClient
			             |
			................................... service boundary
			             |
			      bloomgateway.Gateway
			             |
			       bloomshipper.Store
			             |
			      bloomshipper.Shipper
			             |
                        bloomshipper.BloomFileClient**
			             |
			        ObjectClient**
			             |
			................................... service boundary
			             |
		         object storage

** not part of this PR
```

This PR still contains a lot of TODOs and possibilities for optimisations, which will be addressed in subsequent pull requests.


pkg/bloomgateway/bloomgateway.go
```diff
@@ -0,0 +1,169 @@
...
+type Gateway struct {
+	services.Service
+
+	cfg     Config
+	logger  log.Logger
+	metrics *metrics
+
+	bloomStore   bloomshipper.Store
+	bloomShipper bloomshipper.Shipper
+	bloomClient  bloom_shipper.Shipper
```
chaudum, line 75: These two fields can be removed. They won't be used in the future.

pkg/storage/bloom/bloom-shipper/bloom_shipper.go
```diff
@@ -0,0 +1,304 @@
+package bloom_shipper
```
chaudum, line 1: Files from the `bloom_shipper` package have been copied from https://github.com/grafana/loki/pull/10697 and need to be replaced.
Location will change to `pkg/storage/stores/shipper/bloomshipper/`

pkg/storage/stores/shipper/bloomshipper/store.go
```diff
@@ -0,0 +1,152 @@
...
+// TODO(chaudum): This is just a placeholder and needs to be replaced by actual
+// bloom filter interface.
+type BloomFilter interface {
+}
+
+type ForEachBloomFilterCallback func(f BloomFilter) error
+
+type ReadShipper interface {
+	ForEachBloom(ctx context.Context, tenant string, from, through time.Time, fingerprints []uint64, callback ForEachBloomFilterCallback) error
+}
```
chaudum, line 21: Interface for BloomShipper is still tbd.

pkg/bloomgateway/client.go
```diff
@@ -0,0 +1,287 @@
...
+				Filters: filters,
+			}
+			resp, err := client.FilterChunkRefs(ctx, req)
+			if err != nil {
+				return err
+			}
+			for _, res := range resp.Chunks {
+				chunkRefs := make([]*logproto.ChunkRef, 0, len(res.ChunkIDs))
+				for _, chunkID := range res.ChunkIDs {
+					chk, err := chunk.ParseExternalKey(tenant, chunkID)
```
owen-d, line 166: It probably makes more sense to send the chunkref structs directly in protos rather than parse them on either end. This may be more efficient and removes these error conditions.

pkg/bloomgateway/client.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	return err
+}
+
+// serverAddrsForFingerprints returns a slices of server address slices for
+// each fingerprint of given fingerprints.
+// The indexes of the returned slices correspond to each other.
+// Returns an error in case the bloom gateway ring could not get the
+// corresponding replica set for a given fingerprint.
+func (c *GatewayClient) serverAddrsForFingerprints(tenantID string, fingerprints []uint64) ([]uint64, [][]string, error) {
```
owen-d, line 270: As we talked about earlier, it's less efficient to iterate fingerprints than tokens on large queries.

pkg/bloomgateway/querier.go
```diff
@@ -0,0 +1,62 @@
...
+	return &BloomQuerier{c: c, logger: logger}
+}
+
+func (bq *BloomQuerier) FilterChunkRefs(ctx context.Context, tenant string, from, through model.Time, chunkRefs []*logproto.ChunkRef, filters ...*logproto.LineFilterExpression) ([]*logproto.ChunkRef, error) {
+	// Shortcut that does not require any filtering
+	if len(chunkRefs) == 0 || len(filters) == 0 {
+		return chunkRefs, nil
+	}
+	// The indexes of the chunks slice correspond to the indexes of the fingerprint slice.
+	fingerprints := make([]uint64, 0, len(chunkRefs))
```
owen-d, line 29: todo(pool)

pkg/bloomgateway/querier.go
```diff
@@ -0,0 +1,62 @@
...
+	chunks := make([][]*logproto.ChunkRef, 0, len(chunkRefs))
+
+	// Sort the chunkRefs by their stream fingerprint
+	// so we can easily append them to the target slice by iterating over them.
+	sort.Slice(chunkRefs, func(i, j int) bool {
+		return chunkRefs[i].Fingerprint > chunkRefs[j].Fingerprint
```
owen-d, line 35: should this be `<`?

pkg/bloomgateway/ringmanager.go
```diff
@@ -0,0 +1,264 @@
...
+	subservicesWatcher *services.FailureWatcher
+
+	RingLifecycler *ring.BasicLifecycler
+	Ring           *ring.Ring
+	Mode           ManagerMode
+}
+
+// NewRingManager instantiates a new RingManager instance.
+// The other functions will assume the RingManager was instantiated through this function.
+func NewRingManager(mode ManagerMode, cfg Config, logger log.Logger, registerer prometheus.Registerer) (*RingManager, error) {
```
owen-d, line 67: nit: probably a good thing to refactor into independent code. This looks to be repeated in both the scheduler & index-gw

pkg/bloomgateway/sharding.go
```diff
@@ -0,0 +1,162 @@
...
+		// bloom gateway keeps their previously owned blocks until new owners are ACTIVE).
+		return s != ring.ACTIVE
+	})
+)
+
+type Limits interface {
+	BloomGatewayShardSize(tenantID string) int
+}
+
+type ShardingStrategy interface {
```
owen-d, line 40: nit: this also looks copy+pasted from the index-gw, can't we reuse?

pkg/logproto/bloomgateway.proto
```diff
@@ -0,0 +1,34 @@
...
+  ];
+  int64 through = 2 [
+    (gogoproto.customtype) = "github.com/prometheus/common/model.Time",
+    (gogoproto.nullable) = false
+  ];
+  repeated logproto.ChunkRef refs = 3;
+  repeated logproto.LineFilterExpression filters = 4;
+}
+
+message ChunkIDsForStream {
```
owen-d, line 23: This can probably be represented more effectively by mirroring https://github.com/grafana/loki/pull/10780 which only encodes the fingerprint once per group. The chunk ID can be constructed from the `(tenant, fingerprint, start, end, checksum)`

```go
type Series struct {
	Fingerprint model.Fingerprint
	Chunks      []ChunkRef
}

type ChunkRef struct {
	Start, End model.Time
	Checksum   uint32
}
```

owen-d (approved):
Left some suggestions but approving 👍

pkg/bloomgateway/sharding.go
chaudum: Yes, will consolidate both the ring manager scaffolding and the sharding strategy in a separate PR after merging this one.

pkg/bloomgateway/client.go
chaudum: Added a comment and log a warning in case the alogo iterating over fingerprints becomes less efficient that an algo that iterates over ring tokens and binary searches fingerprint ranges.

pkg/bloomgateway/sharding.go
```diff
@@ -0,0 +1,162 @@
...
+// order to guarantee the same logic is used.
+func GetShuffleShardingSubring(ring ring.ReadRing, tenantID string, limits Limits) ring.ReadRing {
+	shardSize := limits.BloomGatewayShardSize(tenantID)
+
+	// A shard size of 0 means shuffle sharding is disabled for this specific user,
+	// so we just return the full ring so that indexes will be sharded across all index gateways.
+	// Since we set the shard size to replication factor if shard size is 0, this
+	// can only happen if both the shard size and the replication factor are set
+	// to 0.
+	if shardSize <= 0 {
```
poyzannur, line 139: can shardSize be negative? why not `==`

pkg/storage/stores/shipper/bloomshipper/store.go
```diff
@@ -0,0 +1,105 @@
...
+type ReadShipper interface {
+	ForEachBlock(ctx context.Context, tenant string, from, through time.Time, fingerprints []uint64, callback ForEachBlockCallback) error
+}
+
+type WriteShipper interface {
+}
+
+type Shipper interface {
+	ReadShipper
+	WriteShipper
```
poyzannur, line 31: Does this block also have bloom-compactor in mind? otherwise why is WriteShipper needed?

poyzannur (approved):
very minor comments/questions. LGTM.

pkg/bloomgateway/sharding.go
```diff
@@ -0,0 +1,162 @@
...
+
+		rs, err = subRing.Get(uint32(blockRef.ThroughFp), BlocksOwnerSync, bufDescs, bufHosts, bufZones)
+		if err != nil {
+			return nil, err
+		}
+		// Include the block only if it belongs to this bloom gateway shard.
+		if rs.Includes(s.instanceID) {
+			filteredBlockRefs = append(filteredBlockRefs, blockRef)
+			continue
+		}
```
vlad-diachenko, line 123: If we include the block by two factors: 1) if `FromFp` belongs to this instance 2) if ThroughFp belongs to this instance, it means that two separate bloomGw might own the same block, one by `FromFp` another one by   `ThroughFp` .
We should check only `FromFp` , it's enough to identify the owner

vlad-diachenko (changes requested):
Looks awesome. just one important comment that I believe we need to address before merging it

pkg/bloomgateway/sharding.go
chaudum: It is actually OK that multiple instances own the same block, if the same block span into the key ranges from both instances.

screenshot_20231006_082906

pkg/bloomgateway/sharding.go
chaudum: `shardSize` is an `int`. And the config would allow setting a negative value.

pkg/storage/stores/shipper/bloomshipper/store.go
chaudum: I could remove it, because it's not needed for the purposes of this PR.

pkg/bloomgateway/client.go
```diff
@@ -0,0 +1,309 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/health/grpc_health_v1"
+
+	"github.com/grafana/loki/pkg/distributor/clientpool"
+	"github.com/grafana/loki/pkg/logproto"
+	"github.com/grafana/loki/pkg/util"
+)
+
+// GRPCPool represents a pool of gRPC connections to different index gateway instances.
+type GRPCPool struct {
```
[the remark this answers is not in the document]
chaudum: Yes, it is to satisfy interface functions without the need to implement all of them on the GRPCPool itself.

Usually I also prefer composition over (this fake) inheritance, but in this case I prefer the simplicity.

vlad-diachenko (approved):
Looks cool

pkg/storage/stores/shipper/bloomshipper/shipper.go
```diff
@@ -125,19 +141,40 @@ func (s *Shipper) findBlocks(
...
+	if b.MaxFingerprint < minFp || b.MinFingerprint > maxFp {
+		return true
+	}
+
+	// Check if the block range is inside a "gap" in the fingerprint slice
+	// e.g.
+	// fingerprints = [1, 2,          6, 7, 8]
+	// block =              [3, 4, 5]
+	idx := getPosition[[]uint64](fingerprints, b.MinFingerprint)
+	return b.MaxFingerprint < fingerprints[idx]
```
vlad-diachenko, line 173: this check should give a `true` if no one item from `fingerprints []uint64` is greater or equal to `b.MinFingerprint` and less or equal to `b.MaxFingerprint`. 
however, this method returns `false` for the block `b.MinFingerprint: 3 , b.MaxFingerprint: 5` if fingerprints []uint64 is :
1.  `[2]` - should be true because the block does not cover this fingerprint
2. `[6]` - should be true because the block does not cover this fingerprint

could you please add more test cases to cover this method?

chaudum: These cases are actually covered in line 163-166

vlad-diachenko (approved):
LGTM

PR #10794, opened 2023-10-05
Move standalone examples to a new top-level section

salvacorts:

Moves the standalone examples inside the _Configuration_ section to a new top-level _Examples_ section.
We can use this section to add more examples in the future.


docs/sources/examples/_index.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Grafana Loki Examples
+menuTitle: Examples
+description: Configuration reference for the parameters used to configure Grafana Loki.
```
ashwanthgoli, line 4: update description. "Examples to help configure Grafana Loki"?

salvacorts: Done :)

JStickler:
@salvacorts  The docs team has spent a lot of time this year discussing what should be top level categories as part of our effort to create a more consistent information architecture across the various products and projects.   I hadn't realized you were proposing making the new Examples section a top level section.  I assumed the Examples would still be staying under the Configuration heading, but that was my mistake for making that assumption.  Since these are configuration examples, I would prefer to keep them under the Configuration heading as that aligns with what the docs team agrees is a best practice, keeping the examples close to the content that references them.

salvacorts:
@JStickler I have no strong preference about where these should go. I pushed one commit to move the _Examples_ section inside the _Configure_ section. Now it looks like:
image

Whereas the current docs are arranged as:
image

I think it still makes sense to add a new _Examples_ section to group the query frontend example and the example snippets.
Wdyt? Happy to close the PR otherwise.

JStickler:
I'm OK with having an examples folder under configure, as your screenshots show.

JStickler (approved):
[docs team]

PR #10709, opened 2023-09-26
Do not transfer off chunks on ingester shutdown

chaudum:
**What this PR does / why we need it**:

The **Write Ahead Log** (WAL) supersedes the chunk transfer feature.

**Checklist**
- [ ] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated
  - [ ] If the change is worth mentioning in the release notes, add `add-to-release-notes` label
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/setup/upgrade/_index.md`
- [ ] For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)

PR #10850, opened 2023-10-11
Remove -ingester.max-transfer-retries

salvacorts:

In https://github.com/grafana/loki/pull/10709, we removed `-ingester.max-transfer-retries` in favor of the WAL. Then, in https://github.com/grafana/loki/pull/10844 we removed it from the config file but we did not remove it from the ingester args. This PR solves this by removing it.


vlad-diachenko (approved):
LGTM

ashwanthgoli (approved):

PR #10886, opened 2023-10-13
Blooms/chunk check

owen-d:
[WIP]

Adds ability to return the list of chunks needed to be downloaded by a query based on bloom membership results

### DIFF
--- a/pkg/storage/bloom/v1/block.go
+++ b/pkg/storage/bloom/v1/block.go
@@ -1,6 +1,8 @@
 package v1
 
 import (
+	"fmt"
+
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
 )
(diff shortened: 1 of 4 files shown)


pkg/storage/bloom/v1/block.go
```diff
@@ -111,3 +113,61 @@ func (bq *BlockQuerier) Err() error {
...
+	bq.blooms.Seek(series.Offset)
+	if !bq.blooms.Next() {
+		return chks, fmt.Errorf("seeking to bloom for fp: %v", fp)
+	}
+
+	bloom := bq.blooms.At()
+
+	// First, see if the search passes the series level bloom before checking for chunks individually
+	for _, search := range searches {
+		if !bloom.sbf.Test(search) {
```
cyriltovena, line 144: Not sure if this hits the network, but if it does you could consider running this in parallel.

owen-d: Thanks. I've been thinking about how to control parallelism+access to the underlying data, but will do that in another PR

owen-d:
There is some more work to do here, but I'll continue work in a followup PR

salvacorts (approved):

PR #10896, opened 2023-10-13
Use lazy bloom filter

salvacorts:

When querying bloom filters, use lazy decoders from https://github.com/owen-d/BoomFilters/pull/1 so we don't make a copy of bloom filters.


pkg/storage/bloom/v1/bloom.go
```diff
@@ -32,16 +32,14 @@ func (b *Bloom) Encode(enc *encoding.Encbuf) error {
 	return nil
 }
 
-func (b *Bloom) Decode(dec *encoding.Decbuf) error {
+type BloomQuerier struct {
```
owen-d, line 35: What's the reason for adding the `BloomQuerier` struct here?

salvacorts: Removed with my last commit.

salvacorts (commented):
@owen-d I applied the changes we discussed in Slack

Issue #10859, opened 2023-10-11
Make `Test(Local|Remote)RuleEval` integration test more reliable


kavirajk:
**Describe the bug**
Read this PR for more history and context
https://github.com/grafana/loki/pull/10857

Make two test cases more reliable
1. TestLocalRuleEval
2. TestRemoteRuleEval

These two test cases were flaky for very long time. But it increased in frequency after upgrading to prometheus 
https://github.com/grafana/loki/pull/10777

We suspect this is some form of regression either in ruler evaluation path or how this test is setup. This issue make sure to fix those.

**To Reproduce**
1. Change timeout on those test cases from 20 seconds to 10 seconds (or less)
2. Run those test cases

**Expected behavior**
Shouldn't fail for lesser timeouts.

**Environment:**
 - Infrastructure: [e.g., Kubernetes, bare-metal, laptop]
 - Deployment tool: [e.g., helm, jsonnet]

**Screenshots, Promtail config, or terminal output**
If applicable, add any output to help explain your problem.

PR #10906, opened 2023-10-15
ruler: Support `writeNotify` on Loki ruler

kavirajk:
prometheus ruler added a feature of notifying the reader when a sample is appended, instead of waiting in a loop burning the CPU cycles. https://github.com/prometheus/prometheus/pull/11949

This changes a default behaviour a bit. Now if `notify` is not enabled, next read is done only when next readTicker is triggered.

Also should fix https://github.com/grafana/loki/issues/10859

Adding few more details for the sake of completeness.

We found this via more frequent failures of rule-evaluation integration tests linked on the issue above. After some investigation, we tracked down to prometheus changes.

Prometheus introduced new type `wlog.WriteNotified` interface with `Notify()` method with a goal to notify any waiting readers, that some write is done. 

Two types implements this type `wlog.Watcher` and `remote.Storage`. `remote.Storage` implements `Notify()` by just calling it's queues `wlog.Watcher`'s `Notify()` under the hood.

How are these types impacts Loki ruler?

Loki ruler also uses `remote.Storage`. So when any samples got committed via `appender`, we have to notify the remote storage.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/ruler/storage/wal/wal.go
+++ b/pkg/ruler/storage/wal/wal.go
@@ -128,10 +128,6 @@ func (w *Storage) SetWriteNotified(writeNotified wlog.WriteNotified) {
 	w.writeNotified = writeNotified
 }
 
-func (w *Storage) getWriteNotified() wlog.WriteNotified {
-	return w.writeNotified
-}
-
 func (w *Storage) replayWAL() error {
 	w.walMtx.RLock()
 	defer w.walMtx.RUnlock()
--- a/pkg/ruler/storage/wal/wal.go
+++ b/pkg/ruler/storage/wal/wal.go
@@ -89,12 +89,17 @@ func NewStorage(logger log.Logger, metrics *Metrics, registerer prometheus.Regis
 	}
 
 	storage.appenderPool.New = func() interface{} {
+		var notify func() = nil
+
+		if storage.writeNotified != nil {
+			notify = storage.writeNotified.Notify
+		}
 		return &appender{
-			w:             storage,
-			writeNotified: storage.getWriteNotified,
-			series:        make([]record.RefSeries, 0, 100),
-			samples:       make([]record.RefSample, 0, 100),
-			exemplars:     make([]record.RefExemplar, 0, 10),
+			w:         storage,
+			notify:    notify,
+			series:    make([]record.RefSeries, 0, 100),
+			samples:   make([]record.RefSample, 0, 100),
+			exemplars: make([]record.RefExemplar, 0, 10),
 		}
 	}
 
@@ -543,11 +548,12 @@ func dirSize(path string) (int64, error) {
 }
 
 type appender struct {
-	w             *Storage
-	writeNotified func() wlog.WriteNotified
-	series        []record.RefSeries
-	samples       []record.RefSample
-	exemplars     []record.RefExemplar
+	w *Storage
+	// Notify the underlying storage that some sample is written
+	notify    func()
+	series    []record.RefSeries
+	samples   []record.RefSample
+	exemplars []record.RefExemplar
 }
 
 var _ storage.Appender = (*appender)(nil)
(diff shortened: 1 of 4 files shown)


pkg/ruler/storage/wal/wal.go
```diff
@@ -62,6 +62,8 @@ type Storage struct {
 	deleted    map[chunks.HeadSeriesRef]int // Deleted series, and what WAL segment they must be kept until.
 
 	metrics *Metrics
+
+	writeNotified wlog.WriteNotified
```
dannykopping, line 66: Howcome this is not a pointer?

pkg/ruler/storage/wal/wal.go
```diff
@@ -677,6 +689,11 @@ func (a *appender) Commit() error {
 		buf = buf[:0]
 	}
 
+	// Notify so that reader waiting for it can read without needing to wait for next read ticker.
+	if a.writeNotified() != nil {
+		a.writeNotified().Notify()
+	}
```
dannykopping, line 695: This is a little clunky; how about just passing a `Notify()` function to the appender? The appender doesn't need any knowledge of the `wlog.WriteNotified`.

jeschkies: I wouldn't have thought about this fix. This is cool. That said I'm a little confused. Shouldn't we implement `Notify()` and pass it on?

kavirajk: Agree on the clunkiness :) 

It comes from 

```
type appender struct {
	w             *Storage
	writeNotified func() wlog.WriteNotified
	series        []record.RefSeries
	samples       []record.RefSample
	exemplars     []record.RefExemplar
}
```

`writeNotified` being `func()` type.

It got bit tricky because we initialize few things during `NewStorage()` and few other things during `initialize()` call. 

wal is created during `wal.NewStorage` but the remoteStorage (that has `Notify()`) only created during `initialize`. We have to it as func type because we can only know if `wal.Storage` has write notify set that way. Also note, `wal.Storage` uses `appender` as a pool.

We can make it more clean by refactoring few things on `NewStorage` and `initialize`. But AFAIK not worth it now?

kavirajk: Also about the nit: `SetWriteNotified` -> `SetWriteNotifier`. 

I think they calling it `WriteNotified` and not `WriteNotifier` because, `WriteNotifier` is confusing in a sense, is it notifying the "write" that is done? or notifying the "writer".

We want to notify the "write" that is actually done (potentially to readers waiting for it).

pkg/ruler/storage/wal/wal.go
```diff
@@ -690,8 +696,8 @@ func (a *appender) Commit() error {
 	}
 
 	// Notify so that reader waiting for it can read without needing to wait for next read ticker.
-	if a.writeNotified() != nil {
-		a.writeNotified().Notify()
+	if a.notify != nil {
+		a.notify()
```
dannykopping, line 700: I think we should log when `notify` is nil instead of silently not notifying.

dannykopping (commented):
Almost there, thanks!

pkg/ruler/storage/wal/wal.go
```diff
@@ -677,6 +691,12 @@ func (a *appender) Commit() error {
 		buf = buf[:0]
 	}
 
+	// Notify so that reader waiting for it can read without needing to wait for next read ticker.
+	if a.notify != nil {
+		level.Warn(a.w.logger).Log("msg", "missing to notify WAL writes because notifier is not set")
```
dannykopping, line 696: Minor grammatical fix

```suggestion
		level.Warn(a.w.logger).Log("msg", "not notifying about WAL writes because notifier is not set")
```

dannykopping (approved):
LGTM, one small grammar nit

PR #10926, opened 2023-10-17
Adjust bloom gateway request structure

chaudum:

Since we already group chunkRefs by fingerprint on the client side to determine the correct bloom gateway to send the request to, we can keep the grouping for the gRPC request.
This will not only optimize the request size, but also helps to reduce complexity of querying blooms on the bloom gateway itself.

There is still conversion between `[]*logproto.ShortRef` and `v1.ChunkRefs` required, which could be removed by sending `v1.ChunkRefs` over the wire.


pkg/bloomgateway/client.go
```diff
@@ -295,14 +269,16 @@ func (c *GatewayClient) serverAddrsForFingerprints(tenantID string, fingerprints
 		level.Warn(c.logger).Log("msg", "using an inefficient algorithm to determin server addresses for fingerprints", "fingerprints", numFingerprints, "tokens", numTokens)
 	}
 
+	fingerprints := make([]uint64, numFingerprints)
 	addresses := make([][]string, numFingerprints)
 	bufDescs, bufHosts, bufZones := ring.MakeBuffersForGet()
 
-	for idx, key := range fingerprints {
-		rs, err = subRing.Get(uint32(key), BlocksRead, bufDescs, bufHosts, bufZones)
+	for idx, key := range groups {
```
owen-d, line 276: It'll likely be more efficient to iterate through the token ranges directly rather than the fingerprints since there can easily be thousands (or millions) of fingerprints touched in a query. This is probably fine for now, but a TODO will be helpful. We could also detect which set is smaller: tokens in the ring or number of fingerprints in the query.

owen-d (approved):

pkg/bloomgateway/client.go
chaudum: There is a TODO on lines 264 to 268 regarding the algorithm to more efficiently find servers.

vlad-diachenko (commented):
looks great!

PR #10957, opened 2023-10-18
Initial loki bloom creation library

paul1r:

Optimization and benchmarking of various LRUs that sit in front of the bloom filters.
Initial creation of a library that can be used to create blooms from chunks

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/bloomtokenizer/bloomtokenizer.go
```diff
@@ -0,0 +1,98 @@
...
+func (bt *BloomTokenizer) PopulateSBF(sbf *filter.ScalableBloomFilter, chunks []chunk.Chunk) {
+	clearCache(bt.cache)
+	for idx := range chunks {
+		lc := chunks[idx].Data.(*chunkenc.Facade).LokiChunk()
+		bt.chunkIDTokenizer.reinit(chunks[idx].ChunkRef)
+
+		itr, err := lc.Iterator(
+			context.Background(),
+			time.Unix(0, 0),
+			time.Unix(0, math.MaxInt64),
```
poyzannur, line 67: We will need to parameterise these from and to right? Can we add a todo?

pkg/bloomtokenizer/tokenizer_test.go
```diff
@@ -0,0 +1,558 @@
...
+			chunkTokenizer.reinit(logproto.ChunkRef{From: 0, Through: 999999, Checksum: 1})
+			require.Equal(t, tc.exp, chunkTokenizer.Tokens(tc.input))
+		})
+	}
+}
+
+func BenchmarkTokens(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		file, _ := os.Open("big.txt")
```
owen-d, line 548: It doesn't look like this file exists. We should include a dataset in the repo for bench consistency.

pkg/bloomtokenizer/tokenizer.go
```diff
@@ -0,0 +1,175 @@
+package bloomtokenizer
+
+import (
+	"encoding/binary"
+	"unicode/utf8"
+
+	"github.com/grafana/loki/pkg/logproto"
+)
+
+type Token struct {
```
owen-d, line 10: I don't think there's an advantage to a specific token struct, especially if the fields are effectively duplicates (one is just a string cast of the other). It could be better served as `[]byte`

poyzannur (approved):
LGTM

pkg/bloomtokenizer/bloomtokenizer.go
```diff
@@ -0,0 +1,98 @@
...
+	metrics *metrics
+
+	lineTokenizer    Tokenizer
+	chunkIDTokenizer *WrappedTokenizer
+	cache            map[string]interface{}
+}
+
+const CacheSize = 150000
+
+// New returns a new instance of the Bloom Tokenizer.
```
poyzannur, line 32: Actually if we can add here a two liner of what Tokenizer should do, that'd be very nice. i don't think this is a self-explanatory component.

pkg/bloomtokenizer/tokenizer.go
```diff
@@ -0,0 +1,175 @@
...
+
+type ngramTokenizer struct {
+	// [min,max) exclusivity
+	min, max, skip      int
+	buffers             [][]rune // circular buffers used for ngram generation
+	runeBuffer          []byte   // buffer used for token generation
+	tokenBuffer         []Token  // buffer used for holding tokens that is returned
+	internalTokenBuffer []Token  // circular buffer for tokens
+}
+
```
poyzannur, line 32: maybe here too, a link to what an n-gram is and how we use it.

pkg/bloomtokenizer/bloomtokenizer.go
```diff
@@ -0,0 +1,98 @@
...
+
+	return t, nil
+}
+
+// TODO: Something real here with metrics
+func newMetrics(r prometheus.Registerer) *metrics {
+	return &metrics{}
+}
+
+func clearCache(cache map[string]interface{}) {
```
owen-d, line 52: For future improvement: we can version the cache so clearing it doesn't require iterating all the keys. It could be something like `map[key]int` where `int` is the version number and if a key exists but the version isn't the current expected version, treat it as absent

pkg/bloomtokenizer/tokenizer.go
```diff
@@ -0,0 +1,175 @@
...
+				}
+				t.internalTokenBuffer[numToks].Key = t.internalTokenBuffer[numToks].Key[:0]
+				t.internalTokenBuffer[numToks].Key = append(t.internalTokenBuffer[numToks].Key, t.runeBuffer...)
+				t.internalTokenBuffer[numToks].Value = string(t.internalTokenBuffer[numToks].Key)
+				numToks++
+			}
+		}
+		i++
+	}
+	t.tokenBuffer = append(t.tokenBuffer, t.internalTokenBuffer[:numToks]...)
```
owen-d, line 97: What's the advantage of having `tokenBuffer` vs `internalTokenBuffer`? Because we reset the tokenbuffer on each `Tokens()` call, there's no protection against mutability issues. Using one slice would be strictly more efficient here afaict.

pkg/bloomtokenizer/tokenizer.go
```diff
@@ -0,0 +1,175 @@
...
+	w.prefix = append(w.prefix, w.i64buf...)
+	w.prefix = append(w.prefix, 58)
+	binary.PutVarint(w.i64buf, int64(chk.Through))
+	w.prefix = append(w.prefix, w.i64buf...)
+	w.prefix = append(w.prefix, 58)
+	binary.LittleEndian.PutUint32(w.i32buf, chk.Checksum)
+	w.prefix = append(w.prefix, w.i32buf...)
+	w.prefix = append(w.prefix, 58)
+
+	w.f = func(tok Token) Token {
```
owen-d, line 170: `w.f` can just be a static method rather than a function instantiated on every reinit call

pkg/bloomtokenizer/tokenizer.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+}
+
+func (w *WrappedTokenizer) reinit(chk logproto.ChunkRef) {
+	//prefix := fmt.Sprintf("%d:%d:%d:", chk.From, chk.Through, chk.Checksum)
+	w.prefix = w.prefix[:0]
+
+	binary.PutVarint(w.i64buf, int64(chk.From))
+	w.prefix = append(w.prefix, w.i64buf...)
+	w.prefix = append(w.prefix, 58)
```
owen-d, line 162: nit: adding the colons to our prefix don't serve a purpose here because they don't help disambiguate keys from each other

pkg/bloomtokenizer/tokenizer.go
```diff
@@ -0,0 +1,175 @@
...
+	binary.PutVarint(w.i64buf, int64(chk.Through))
+	w.prefix = append(w.prefix, w.i64buf...)
+	w.prefix = append(w.prefix, 58)
+	binary.LittleEndian.PutUint32(w.i32buf, chk.Checksum)
+	w.prefix = append(w.prefix, w.i32buf...)
+	w.prefix = append(w.prefix, 58)
+
+	w.f = func(tok Token) Token {
+		tok.Key = append(append(tok.Key, w.prefix...), tok.Key...)[len(tok.Key):]
```
owen-d, line 171: We don't need to do this append-chaining just to reorder the prefix vs key structure. What we do need is generate unique keys for each `(chunk, ngram)` combination.

owen-d (commented):
Some comments + let's move this into the `bloom/v1` pkg

pkg/bloomtokenizer/tokenizer.go
slim-bean: do we ever do any unsafe casting? I could maybe see a value here in guaranteeing memory consistency by doing it like this if we are into the unsafe casting game.

owen-d: Shouldn't have to -- these are just generated from logs for immediate bloom filter testing

paul1r: I've removed the string value of the struct.  I've left the Token struct there, it does provide a level of readability, but I'm not married to the idea.

pkg/bloomtokenizer/bloomtokenizer.go
paul1r: I understand where you're going with that, but I believe it will lead to the map growing unbounded.  For an extreme example, if the map was filled to capacity with a set of tokens with version=1, when we bump to version 2, if the next set of tokens were all unique, we'd have capacity*2 number of tokens.
I don't like iterating through the keys, but it saves reallocation of the map and has proven faster in my benchmarks

owen-d (commented):
A couple comments. I think we can get this merged soon to unblock some other work streams and improve it after the fact.

pkg/storage/bloom/v1/tokenizer.go
```diff
@@ -0,0 +1,162 @@
...
+	prefix      []byte
+	i64buf      []byte
+	i32buf      []byte
+}
+
+func (w *WrappedTokenizer) Tokens(line string) []Token {
+	w.tokenBuffer = w.tokenBuffer[:0] // Reset the result slice
+	toks := w.t.Tokens(line)
+	for _, tok := range toks {
+		w.tokenBuffer = append(w.tokenBuffer, chunkIDTransformer(tok, w.prefix), tok)
```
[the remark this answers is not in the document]
paul1r: My intention is that w.tokenBuffer doesn't need to be reallocated, which is why I set it to :0, as opposed to setting it to nil.

As far as the appending to the tokenBuffer, I've spent a lot of time running performance tests, and this way has been the fastest out of what I've tried.  Open to other ideas.

owen-d (approved):

PR #10959, opened 2023-10-18
Loki: introduce an exponential backoff on failed query retries.

slim-bean:

Currently any subquery which fails with a 5xx status code is retried by the query frontend immediately, and by default will make a total of 5 attempts to execute the subquery, ultimately failing the query if you exhaust those attempts.

We notice there are times when subqueries can fail extremely quickly, mostly around queriers handling various GRPC related issues between components like ingesters or index gateways.

In these cases a query will exhaust all the retries in less than a second and fail back to the user.

Also in cases where a query is causing a lot of memory pressure on a pool of queriers leading to OOM crashes of queriers, the lack of a backoff on retries can make this behavior quite aggressive and potentially cause disruption if you don't have enough queriers available.

This PR introduces a simple exponential backoff on retries of failed queries to allow for the downstream GRPC issues to have time to sort out as well as causing queries causing a lot of memory pressure and OOMing to be slowed down and minimize their impact slightly.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/pkg/querier/queryrange/queryrangebase/retry.go
+++ b/pkg/querier/queryrange/queryrangebase/retry.go
@@ -2,9 +2,11 @@ package queryrangebase
 
 import (
 	"context"
+	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/backoff"
 	"github.com/grafana/dskit/httpgrpc"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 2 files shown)


owen-d (approved):

pkg/querier/queryrange/queryrangebase/retry.go
```diff
@@ -57,6 +59,20 @@ func (r retry) Do(ctx context.Context, req Request) (Response, error) {
...
+	// try 2: 500ms wait
+	// try 3: 1s wait
+	// try 4: 2s wait
+
+	cfg := backoff.Config{
+		MinBackoff: 250 * time.Millisecond,
+		MaxBackoff: 5 * time.Second,
+		MaxRetries: 0,
+	}
+	bk := backoff.New(ctx, cfg)
```
dannykopping, line 75: Can you add a test please?

pkg/querier/queryrange/queryrangebase/retry.go
```diff
@@ -70,7 +86,8 @@ func (r retry) Do(ctx context.Context, req Request) (Response, error) {
 		httpResp, ok := httpgrpc.HTTPResponseFromError(err)
 		if !ok || httpResp.Code/100 == 5 {
 			lastErr = err
-			level.Error(util_log.WithContext(ctx, r.log)).Log("msg", "error processing request", "try", tries, "query", req.GetQuery(), "err", err)
+			level.Error(util_log.WithContext(ctx, r.log)).Log("msg", "error processing request", "try", tries, "query", req.GetQuery(), "retry_in", bk.NextDelay(), "err", err)
```
dannykopping, line 89: `NextDelay()` 👍 nice touch

dannykopping (approved):
LGTM, a test would be helpful

pkg/querier/queryrange/queryrangebase/retry.go
slim-bean: I don't think I understand what we would test here?

dannykopping: Was initially thinking it'd validate the delays you'd defined there, but that's tested by the underlying lib. Disregard

PR #10972, opened 2023-10-19
LRU Cache wrapper

vlad-diachenko:
Add a simple LRU cache wrapper that locks the key until the value supplier provides the value
It will be used as a bloom-blocks cache that will provide guarantees that the block is downloaded only once.

### DIFF
--- a/pkg/util/cache/lru/lru.go
+++ b/pkg/util/cache/lru/lru.go
@@ -1,0 +1,70 @@
+package lru
+
+import (
+	"fmt"
+
+	lru_hashicorp "github.com/hashicorp/golang-lru"
+	"k8s.io/utils/keymutex"
+)
+
+func defaultKeyStringifier[K comparable](key K) string {
+	return fmt.Sprintf("%+v", key)
+}
+
+// Cache is a simple LRU cache wrapper that locks the key until the value supplier provides the value.
+// It can be used if it's necessary to make sure the value for the key was created/fetched only once even during concurrent access.
+type Cache[K comparable, V any] struct {
+	delegate       *lru_hashicorp.Cache
+	onKeyMissing   func(key K) (V, error)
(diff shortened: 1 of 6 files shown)


pkg/util/cache/lru/lru.go
```diff
@@ -0,0 +1,70 @@
+package lru
```
dannykopping, line 1: Can we not make use of our existing LRU cache implementation?

vlad-diachenko: Do you mean the package name?

dannykopping: We already have `pkg/storage/chunk/cache/embeddedcache.go` which is an LRU, plus we have `github.com/hashicorp/golang-lru` imported as well. I just want to make sure we've exhausted all other options before including yet another LRU implementation.

vlad-diachenko: Oh, I see. yep, I have not seen the embedded cache, I checked all the places where we use `github.com/hashicorp/golang-lru` and that is why I decided to create a wrapper over hashicorp-lru. 

I believe it makes sense to change my wrapper to use embedded cache under the hood, instead of using hasicorp-lru because it provides reach instrumentation... 
I will check if it's possible to refactor the embedded cache, because we need to have an eviction callback and also we need to store the struct as a value, not  `[]byte` , and if it's possible, then I will switch to it. 

Also, I will try to find a better name for my implementation to not collide with other LRU implementations. 
Does it make sense?

dannykopping: OK super 👍

vlad-diachenko:
closed in favor of https://github.com/grafana/loki/pull/10972 . 
The new implementation of embedded cache already provides the ability to control the cache size by item count or total size, and it also provides metrics out of the box.

PR #10974, opened 2023-10-19
adds a merge-builder for building blocks from other blocks+storage

owen-d:
Some minor refactoring + a (simple) `MergeBuilder` struct to build bloom blocks from other blocks and storage, depending on what's already been indexed into the blocks.


poyzannur (approved):
LGTM. This is super helpful Owen. If we can merge it I can play around and test it.

PR #10976, opened 2023-10-20
Introduce worker queue in bloom gateway

chaudum:
Instead of calling the bloom store directly on each and every request to filter chunk refs based on the given filters, we want to queue requests in per-tenant queues and process batches of requests that can be multiplexed to avoid excessive seeking in the bloom block queriers when checking chunk matches.

This PR re-uses the request queue implementation used in the query scheduler. To do so, it moves the queue related code from the scheduler into a separate package `pkg/queue` and renames occurrences of "querier" to "consumer" to be more generic.

The bloom gateway instantiates the request queue when starting the service. The gRPC method `FilterChunkRefs` then enqueues incoming requests to that queue.

For testing purposes, this PR also contains a dummy implementation of the workers. The worker implementation - which includes multiplexing of multiple tasks - is subject to a separate PR.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/bloomgateway/bloomgateway.go
```diff
@@ -99,18 +194,116 @@ func New(cfg Config, schemaCfg config.SchemaConfig, storageCfg storage.Config, s
...
+	return services.StopManagerAndAwaitStopped(context.Background(), g.serviceMngr)
+}
+
+// This is just a dummy implementation of the worker!
+// TODO(chaudum): Implement worker that dequeues multiple pending tasks and
+// multiplexes them prior to execution.
+func (g *Gateway) startWorker(_ context.Context, id string) error {
+	level.Info(g.logger).Log("msg", "starting worker", "worker", id)
+
+	g.queue.RegisterQuerierConnection(id)
```
vlad-diachenko, line 267: should it be RegisterWorkerConnection? or maybe RegisterConsumerConnection

pkg/bloomgateway/bloomgateway.go
```diff
@@ -69,20 +151,33 @@ type Gateway struct {
...
 	g := &Gateway{
-		cfg:      cfg,
-		logger:   logger,
-		metrics:  newMetrics(reg),
-		sharding: shardingStrategy,
+		cfg:          cfg,
+		logger:       logger,
+		metrics:      newMetrics("bloom_gateway", reg),
+		sharding:     shardingStrategy,
+		pendingTasks: makePendingTasks(1024),
```
kavirajk, line 174: nit: Move `1024` to some const?

pkg/bloomgateway/bloomgateway.go
```diff
@@ -99,18 +194,116 @@ func New(cfg Config, schemaCfg config.SchemaConfig, storageCfg storage.Config, s
...
+			level.Error(g.logger).Log("msg", "failed to gracefully stop bloom gateway dependencies", "err", err)
+		}
+	}()
+
+	if err := services.StartManagerAndAwaitHealthy(ctx, g.serviceMngr); err != nil {
+		return errors.Wrap(err, "unable to start bloom gateway subservices")
+	}
+
+	for i := 0; i < numWorkers; i++ {
+		go g.startWorker(ctx, fmt.Sprintf("worker-%d", i))
```
kavirajk, line 227: Q: How do know these worker go routines are stopped correctly? (asking because it uses resource like querierConnection)

Using some shard `stopChanel` (and worker multiplex on it to listen when to shutdown) across workers or tracking all the spanned workers and calling explicity `Stop()` on them during shutdown can help?

pkg/bloomgateway/bloomgateway.go
```diff
@@ -99,18 +194,116 @@ func New(cfg Config, schemaCfg config.SchemaConfig, storageCfg storage.Config, s
...
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case err := <-g.serviceWatcher.Chan():
+			return errors.Wrap(err, "bloom gateway subservice failed")
+		case <-inflightTasksTicker.C:
+			g.pendingTasks.RLock()
+			inflight := len(g.pendingTasks.Map)
+			g.pendingTasks.RUnlock()
```
vlad-diachenko, line 250: could we just create a few methods of SyncMap that will do the locking/unlocking inside these methods?

vlad-diachenko (approved):
LGTM

pkg/bloomgateway/bloomgateway.go
```diff
@@ -99,18 +194,116 @@ func New(cfg Config, schemaCfg config.SchemaConfig, storageCfg storage.Config, s
...
+			g.metrics.inflightRequests.Observe(float64(inflight))
```
kavirajk, line 251: Q: I'm bit confused with how we calculate `inflightRequests` metric here. From the description, I see it's combination of requests both waiting and getting processed. Why not just increment when creating and successful enqueue in `FilterChunksRefs` method and decrement after dequeue and processed on the worker loop?

why need in-flights ticker here?

> We also do it with
// a ticker so that we keep tracking it even if we have no new requests but stuck inflight
// tasks (eg. worker are all exhausted).

I understand we are calculating `inflight` requests based on `pendingTasks` shared map. And in that way, once task is dequeue and it removed from pendingTasks but if it get stuck on fetching it from BloomStore then inflights metrics can go invalid. But do we have to that way?

Instead why not increament `inflight` metrics during creation + successful enqueue (in FilterChunkRefs) and decrement it only after dequeue + success processing (after getting response from BloomStore)

Am i missing anything here?

pkg/bloomgateway/bloomgateway.go
chaudum: Should be `RegisterConsumerConnection`. I thought I changed that at some point, but seems to have slipped through.

pkg/bloomgateway/bloomgateway.go
```diff
@@ -99,18 +194,116 @@ func New(cfg Config, schemaCfg config.SchemaConfig, storageCfg storage.Config, s
...
+	level.Info(g.logger).Log("msg", "starting worker", "worker", id)
+
+	g.queue.RegisterQuerierConnection(id)
+	defer g.queue.UnregisterQuerierConnection(id)
+
+	idx := queue.StartIndexWithLocalQueue
+
+	for {
+		ctx := context.Background()
+		item, newIdx, err := g.queue.Dequeue(ctx, idx, id)
```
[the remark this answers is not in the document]
chaudum: Yes, I will handle that in a follow up PR.

pkg/bloomgateway/bloomgateway.go
chaudum: The goroutine has a `for` loop which dequeues from the queue. The queue will yield an error when it's stopped, then the goroutine returns and it is stopped.

Note, this implementation is going to be replaced in a follow up PR.

pkg/bloomgateway/bloomgateway.go
chaudum: Afaik, the idea of using a Summary for the inflight requests is that we can better observer spikes. If we'd use a Gauge and do Inc()/Dec() for every task, we can easily miss spikes in between the 15s scrape interval.

The reasoning is also explained in the [PR](https://github.com/grafana/loki/pull/5658) of the original implementation in the query scheduler.

PR #10977, opened 2023-10-20
Deprecated/Deleted Config Checker tool

salvacorts:

This PR adds a new _deprecated/deleted config checker_ tool to allow users upgrading from previous Loki versions to check their Loki config and runtime config for deprecated and deleted options, as well as providing them with alternatives and reasons why those configs were deprecated/deleted.

Mimir provides a [tool][1] that allows users to migrate their old configs to new ones. Having that is our future goal but the new tool in this PR is a good middle ground for the short-mid term.

We also update the checklist in the PR template to ask contributors to update the tool source of truth if the contributor is deprecating or deleting a config.

Here's an example for how to use the tool:
```bash
go run tools/deprecated-config-checker/main.go \
  -config.file tools/deprecated-config-checker/test-fixtures/config.yaml \
  -runtime-config.file tools/deprecated-config-checker/test-fixtures/runtime-config.yaml
```

And this is an example output:
image

### Notes

I think I've gone through all the deprecated and deleted configs. I cannot find the config pointed by the deprecated `ruler.storage`. It states 
```
ruler:
    ...
    # Deprecated: Use -ruler-storage. CLI flags and their respective YAML config
    # options instead. 
    storage:
          ...
```

I couldn't find that `-ruler-storage` cli flag nor the YAML definition.

[1]: https://grafana.com/docs/mimir/latest/manage/tools/mimirtool/#convert


tools/deprecated-config-checker/checker/checker.go
```diff
@@ -0,0 +1,289 @@
...
+
+		// To this point, the deprecate item is not a leaf, so we need to recurse into it.
+		if deprecateYaml, is := deprecate.(RawYaml); is {
+			switch v := inputValue.(type) {
+			case RawYaml:
+				deprecations = enumerateDeprecatesFields(deprecateYaml, v, path, deprecations)
+			case []interface{}:
+				// If the config is a list, recurse into each item.
+				for i, item := range v {
+					itemYaml := item.(RawYaml)
```
ashwanthgoli, line 281: could this fail?

tools/deprecated-config-checker/deleted-config.yaml
```diff
@@ -0,0 +1,33 @@
+# This file should contain a list fo deleted config options.
+#
+# The syntax for this file is the same as the syntax for deprecated-config.yaml.
+
+ingester:
+  max_transfer_retries: "Enable the ingester WAL and rely on new ingesters to replay the WAL."
+
+querier:
+  engine:
+    timeout: "Configure global or per-tenant query_timeout limit."
```
ashwanthgoli, line 10:
```suggestion
    timeout: "Configure global or per-tenant query_timeout limit using limits_config."
```

ashwanthgoli: same comment for all other runtime configs.

tools/deprecated-config-checker/deleted-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+  split_queries_by_interval: "Configure global or per-tenant split_queries_by_interval limit"
+  forward_headers_list: "This setting is removed without an alternative"
+
+frontend_worker:
+  parallelism: "Configure querier.max_concurrent to limit the max concurrent requests processed by the queriers."
+  match_max_concurrent: "Configure querier.max_concurrent to limit the max concurrent requests processed by the queriers."
+
+common:
+  storage:
+    s3: &s3_deletes
```
ashwanthgoli, line 22: `s3_deletes`? is this part of the syntax.

tools/deprecated-config-checker/deprecated-config.yaml
```diff
@@ -0,0 +1,70 @@
...
+#       _msg: "Use tsdb (preferred) or boltdb-shipper instead."
+#
+# Note that even though the configs in schema_config takes a list, here we specify the deprecated fields for each item in the list.
+
+ruler:
+  remote_write:
+    client: "Use clients instead."
+
+compactor:
+  deletion_mode: "Use deletion_mode per tenant configuration instead."
```
ashwanthgoli, line 22:
```suggestion
  deletion_mode: "Use deletion_mode in limits_config instead."
```

tools/deprecated-config-checker/deprecated-config.yaml
```diff
@@ -0,0 +1,70 @@
...
+  ruler_remote_write_queue_retry_on_ratelimit: "Use ruler_remote_write_config instead."
+  ruler_remote_write_sigv4_config: "Use ruler_remote_write_config instead."
+  per_tenant_override_config: "Feature renamed to 'runtime configuration', flag deprecated in favor of runtime_config.file"
+  per_tenant_override_period: "Feature renamed to 'runtime configuration', flag deprecated in favor of runtime_config.period"
+  allow_deletes: "Use deletion_mode per tenant configuration instead."
+
+# TODO:
+# - ruler.storage: Use -ruler-storage.
+#      I cannot find where ruler-storage is defined.
+# - index_gateway.replication_factor: Use -index-gateway.shard-size instead. The shard size is also a per-tenant setting.
```
ashwanthgoli, line 69: this one I think https://github.com/grafana/loki/blob/53416d6930e4d83515585c98b08866176fa8796c/pkg/validation/limits.go#L184

tools/deprecated-config-checker/deprecated-config.yaml
```diff
@@ -0,0 +1,70 @@
...
+  ruler_remote_write_queue_min_backoff: "Use ruler_remote_write_config instead."
+  ruler_remote_write_queue_max_backoff: "Use ruler_remote_write_config instead."
+  ruler_remote_write_queue_retry_on_ratelimit: "Use ruler_remote_write_config instead."
+  ruler_remote_write_sigv4_config: "Use ruler_remote_write_config instead."
+  per_tenant_override_config: "Feature renamed to 'runtime configuration', flag deprecated in favor of runtime_config.file"
+  per_tenant_override_period: "Feature renamed to 'runtime configuration', flag deprecated in favor of runtime_config.period"
+  allow_deletes: "Use deletion_mode per tenant configuration instead."
+
+# TODO:
+# - ruler.storage: Use -ruler-storage.
```
ashwanthgoli, line 67: not sure either, don't see anything in the upgrade doc

tools/deprecated-config-checker/deleted-config.yaml
salvacorts: I'm using YAML [aliases and anchors][1] so I don't repeat the deprecated fields for s3. 

[1]: https://www.educative.io/blog/advanced-yaml-syntax-cheatsheet#anchors

tools/deprecated-config-checker/deleted-config.yaml
salvacorts: Thank you! I think `query_timeout limit inside limits_config` (note **inside** instead of **using**) is more correct.

tools/deprecated-config-checker/deprecated-config.yaml
salvacorts: Thank you!

tools/deprecated-config-checker/deprecated-config.yaml
salvacorts: Done

tools/deprecated-config-checker/deleted-config.yaml
salvacorts: Changed to `Use global or per-tenant query_timeout configuration from limits_config.`

tools/deprecated-config-checker/deprecated-config.yaml
salvacorts: I've added a note to the PR description.

tools/deprecated-config-checker/checker/checker.go
salvacorts: This may panic if the item is a list of strings or numbers. Looking at the config, looks like we do have lists of strings (e.g. `distributor.ring.instance_interface_names`). 

But this will not happen because we iterate through the deprecated/deleted fields. So if the field is deprecated, we won't recurse into it.

We need to update how we handle the deprecation early on to be able to check deprecated values on a list of strings. Updated in my last commit.

tools/deprecated-config-checker/deleted-config.yaml
ashwanthgoli: aha, nice

tools/deprecated-config-checker/checker/checker.go
ashwanthgoli: nice, this wasn't that important of a case to fix I think. thanks for doing this

ashwanthgoli (approved):

PR #10983, opened 2023-10-20
Recover querier handler from panic.

jeschkies:
Commit 56437fa0c44efae3727f05972ab1d7e54ef59f64 introduced a querier handler that would work without an HTTP layer. This removed the HTTP recovery middleware that would catch panics and return them as HTTP 500 errors. This change introduces a new middlware for the querier handler that has the same behviour.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -507,7 +507,7 @@ func (t *Loki) initQuerier() (services.Service, error) {
 	svc, err := querier.InitWorkerService(
 		querierWorkerServiceConfig,
 		prometheus.DefaultRegisterer,
-		handler,
+		serverutil.RecoveryMiddleware.Wrap(handler),
 		t.Codec,
 	)
 	if err != nil {
--- a/pkg/util/server/recovery.go
+++ b/pkg/util/server/recovery.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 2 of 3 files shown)


pkg/loki/modules.go
```diff
@@ -507,7 +507,7 @@ func (t *Loki) initQuerier() (services.Service, error) {
 	svc, err := querier.InitWorkerService(
 		querierWorkerServiceConfig,
 		prometheus.DefaultRegisterer,
-		handler,
+		serverutil.RecoveryMiddleware.Wrap(handler),
```
DylanGuedes, line 510: double checking: shouldn't you wrap `httpHandler` here instead of `handler`?

DylanGuedes (changes requested):

pkg/loki/modules.go
cstyan: probably worth clarifying this via a comment somewhere that's easy to find, my understanding is this;
- there's now the base level querier API, which supports protobuf
- on top of that we use a wrapper function to turn that APIs responses into the currently used http responses (httpgrpc)

so I think the same question is likely to come up again, and we should make it obvious when a middleware needs to be added to both the `handler` and `httpHandler` vs just one of the two

jeschkies: Should I rename the handler? I can add some documentation.

cstyan: Lets rename or add docs/comments when you're back

cstyan (approved):

PR #10989, opened 2023-10-21
embedded cache refactoring

vlad-diachenko:
refactored embedded cache to allow using it for any types, not only for `string -> []byte`. 
Also, the new implementation allows passing the callback that will be called for the removed entry.


pkg/storage/chunk/cache/embeddedcache.go
```diff
@@ -205,41 +225,54 @@ func (c *EmbeddedCache) Store(_ context.Context, keys []string, values [][]byte)
...
+func (c *EmbeddedCache[K, V]) Stop() {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
 	close(c.done)
-
-	c.entries = make(map[string]*list.Element)
+	if c.onEntryRemoved != nil {
+		for _, entry := range c.entries {
+			castedEntry := entry.Value.(*cacheEntry[K, V])
```
dannykopping, line 235: Please check that this cast succeeds, otherwise the next line will panic

dannykopping: Hhmm, do we really want to call this function for _every_ item once `Stop()` is called? That could be quick slow. Is it necessary?

pkg/storage/chunk/cache/embeddedcache.go
```diff
@@ -82,8 +86,26 @@ func (cfg *EmbeddedCacheConfig) IsEnabled() bool {
...
+// NewEmbeddedCache returns a new initialised EmbeddedCache where the key is a string and the value is a slice of bytes.
+func NewEmbeddedCache(name string, cfg EmbeddedCacheConfig, reg prometheus.Registerer, logger log.Logger, cacheType stats.CacheType) *EmbeddedCache[string, []byte] {
+	return NewTypedEmbeddedCache[string, []byte](name, cfg, reg, logger, cacheType, sizeOf, nil)
+}
+
+// NewTypedEmbeddedCache returns a new initialised EmbeddedCache with the key and value of requested types.
+// To limit the memory allocated by items in the cache, it's necessary to pass cacheEntrySizeCalculator
+// that calculates the size of an entry in bytes.
+// Also, this constructor allows passing the callback that will be called for the entry whenever it is removed from the cache.
+func NewTypedEmbeddedCache[K comparable, V any](
```
dannykopping, line 100: Nice!!

dannykopping (commented):
Very nice work @vlad-diachenko! Left a couple thoughts

pkg/storage/chunk/cache/embeddedcache.go
```diff
@@ -161,40 +185,36 @@ func (c *EmbeddedCache) runPruneJob(interval, ttl time.Duration) {
...
 
 // pruneExpiredItems prunes items in the cache that exceeded their ttl
-func (c *EmbeddedCache) pruneExpiredItems(ttl time.Duration) {
+func (c *EmbeddedCache[K, V]) pruneExpiredItems(ttl time.Duration) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
 	for k, v := range c.entries {
-		entry := v.Value.(*cacheEntry)
+		entry := v.Value.(*cacheEntry[K, V])
```
chaudum, line 193: Even though the cache value should not be of a different type, we should still check the cast result:

```suggestion
		entry, ok := v.Value.(*cacheEntry[K, V])
		if ok && time.Since(entry.updated) > ttl {
		    c.remove(k, v, expiredReason)
		}
```

pkg/storage/chunk/cache/embeddedcache.go
vlad-diachenko: Yes, if the value must be properly closed, we need to call on `onEntryRemoved` callback. In our case, we will use this function to remove the data from the disk.
It's not always necessary if you run your pod against ephemeral storage, but if you run Loki on our local machine, then we need to clean up the data from the disk.
If it becomes unnecessary for some cases in the future, we can add an additional toggle to disable calling this function during the stop. 
wdyt?

dannykopping: What are you storing on disk?

One way you could solve this efficiently is to generate a unique ID, and store all your objects on disk under a directory with that ID. When the cache is `Stop()`-ed, you can delete the directory.

Doing an O(N) scan on a potentially full cache will be very slow.

vlad-diachenko:
hey @chaudum , @dannykopping
. I have a question about the need to assert the type of entry from the cache.
The underlying cache is a private field, so, nobody can put the entries from outside and currently, it’s only possible to do using Store function that forces the entries to have types [K, V] .
 I am not sure that we need to assert if the type is casted properly because it means that somebody produced the bug in the cache itself, and this bug also is not caught by the test…. and I do not know if the check of the type will help us here…. maybe it’s better to panic because in this case, I do not think that application can operate
wdyt?

pkg/storage/chunk/cache/embeddedcache.go
dannykopping: If the type is not expected, a panic seems appropriate

pkg/storage/chunk/cache/embeddedcache.go
vlad-diachenko: >What are you storing on disk?

 we will store extracted bloom-blocks on the disk. and we need to be able to delete the directory with this block if entry is evicted from the cache. 

>One way you could solve this efficiently is to generate a unique ID, and store all your objects on disk under a directory with that ID. When the cache is Stop()-ed, you can delete the directory.

yes, we could do it but it will require having different callbacks, one for `onEntryRemoved` and one more for `onStop`... anyway, for now, such cache with `onEntryRemoved` callback will be used only for bloom-blocks so it does not affect rest of the implementation because the callback will be `nil` there...  
Also, even if it will be a thousand of the directories to delete, it should not take a lot of time. 
wdyt?

dannykopping: When will `Stop()` be called?

I also don't understand the relationship between the data on disk and the values going into this cache. Can you elaborate?

vlad-diachenko: > When will Stop() be called?

I believe that for each cache instance, it might be different. We will call `Stop()` during application shutdown.

>I also don't understand the relationship between the data on disk and the values going into this cache. Can you elaborate?

Yep, sure...
In the cache, we will store some readers of the data that will be stored on the disk. In the beginning, we will limit the cache by the maximum items count, so when we reach the max capacity, old items will be removed, and on removal we need to remove the data from the disk also.

dannykopping: OK thanks.
Why can't you store all the items on disk under 1 directory (with some unique ID generated at startup), and at shutdown just drop the whole directory?

And why do you need to drop the data from disk between runs of the process?

vlad-diachenko: > Why can't you store all the items on disk under 1 directory (with some unique ID generated at startup), and at shutdown just drop the whole directory?

we can, and we already have a config property `working_directory` and all blocks will be downloaded to this folder. So, we can just drop the whole directory.
However, I prefer to move this responsibility to the cache. The cache will remove individual blocks when they are evicted, so I think we need to delete the rest of the blocks during the shutdown. 

> And why do you need to drop the data from disk between runs of the process?

It is not so important for the way we are going to use it in our cloud because we will use ephemeral storage that will be removed once the pod is shut down. 
However, I still want to remove all the blocks during the cache shutdown to handle the case when the users run Loki binary on VMs or bare metal because otherwise this data might be abandoned on the disk...

dannykopping: This sounds like you're just building memcached extstore.

I'd highly recommend that you try use that instead of introducing all of this complexity.

vlad-diachenko: yes, completely agree with you that we might end up using Memcached with extstore. The current implementation is just the initial version and maybe after this, we will think about using Memcached. 
For now, we want to simplify it using LRU cache just to check how things go...

dannykopping: Why not just use memcached from the beginning? memcached is an LRU...

vlad-diachenko:
@chaudum @dannykopping what do you think about the PR? can we merge it?

pkg/storage/chunk/cache/embeddedcache.go
vlad-diachenko: I am not sure what will be better to store the blocks on the disk or download them from the Memcache every time. we expect these blocks to be big enough, so we want to store them on a local disk because it might be just faster...
Let's say if the query touches 100GB of data, then it might be necessary to download about 1 GB of blocks, it might take some time ...

dannykopping: If you want, we can jump on a call tomorrow or later today to discuss.

vlad-diachenko: yep, let's discuss it tomorrow

vlad-diachenko: removed execution onEntryRemoved callback during the cache stop https://github.com/grafana/loki/pull/10989/commits/a6f7a435fc05b686a6cf5257cbbc75b7739ca2bd

owen-d (approved):

PR #10991, opened 2023-10-22
compactor: pick index storage client from index storage config

### LINKED ISSUE #10985: Wrong compactor index client for schema
**Describe the bug**
After these changes #6537 & #7447, the compactor will pick the `indexStorageClient` from the `object_store` of the given schema: https://github.com/grafana/loki/blob/v2.9.2/pkg/storage/stores/indexshipper/compactor/compactor.go#L581
This is causing a problem in the compactor trying to access the index from a store which for example `boltdb_shipper` is not pointing to (the scenario having multiple schemas with multiple stores but one s3 store for boltdbshipper index).

**To Reproduce**
Having multiple schemas with multiple stores (`named_stores` feature) and one boltdbshipper index storage config.

**Expected behavior**
`indexStorageClient` should be picked from `shared_store` of the index storage config.

**Environment:**
- Loki v2.9.2

clwluvw:
indexStorageClient should be picked from shared_store of the index storage config.

Fixes #10985

### DIFF
--- a/pkg/bloomgateway/bloomgateway_test.go
+++ b/pkg/bloomgateway/bloomgateway_test.go
@@ -245,6 +245,6 @@ func TestBloomGateway_FilterChunkRefs(t *testing.T) {
 			_, err = gw.FilterChunkRefs(ctx, req)
 			require.NoError(t, err)
 		}
-		require.Equal(t, tenants, gw.activeUsers.ActiveUsers())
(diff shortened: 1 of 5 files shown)


chaudum:
This PR is out-of-date. The linked issue was with Loki 2.9.x which is not supported any more.

Issue #10741, opened 2023-09-28
[Bug/LogQL]: Results mismatch with binary comparison operators


kavirajk:
**Describe the bug**
Following LogQL queries gives different results. But supposed to return same results according to doc (`2000` is just used for example, any scaler value would have same behaviour)

```
2000 < sum(count_over_time({cluster="play-db-cluster"}[1h]))
```
```
sum(count_over_time({cluster="play-db-cluster"}[1h])) > 2000
```

![Screenshot 2023-09-28 at 16 50 42](https://github.com/grafana/loki/assets/3735252/460a047d-74d6-41b5-9cd0-a124063afaa6)

we’re returning the value of the thing on the left side of the binary operation when the comparison is truthy, as opposed to the vector value which is what the docs say it should return: https://grafana.com/docs/loki/latest/query/#comparison-operators
```
Between a vector and a scalar, these operators are applied to the value of every data sample in the vector...
```

[Grafana play link for above queries](https://play.grafana.org/explore?panes=%7B%22fXy%22:%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%222000%20%3C%20sum%28count_over_time%28%7Bcluster%3D%5C%22play-db-cluster%5C%22%7D%5B1h%5D%29%29%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22SCALAR_FRONT%22%7D,%7B%22refId%22:%22B%22,%22expr%22:%22sum%28count_over_time%28%7Bcluster%3D%5C%22play-db-cluster%5C%22%7D%5B1h%5D%29%29%20%3E%2020000%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22SCALAR_BACK%22%7D%5D,%22range%22:%7B%22from%22:%22now-30m%22,%22to%22:%22now%22%7D%7D%7D&schemaVersion=1&orgId=1)

Thanks @gtk-grafana for reporting the bug

**To Reproduce**
Bug seems to be present for ever, tested it on 2.9.1, 2.8.5 and 2.7.7.

**Expected behavior**
Both queries should return same results.

[...]

PR #10997, opened 2023-10-23
Bug(LogQL): Fix mismatch results on `scalar` and `vector` binary ops

kavirajk:

[According to LogQL doc,](https://grafana.com/docs/loki/latest/query/#comparison-operators) binary operators between `scalar` and `vector` should behave as follows (either it's `scalar` op `vector` or `vector` op `scalar`)

```
Between a vector and a scalar, these operators are applied to the value of every data sample in the vector
```
PromQL (where LogQL inspired from) [also works that way](https://prometheus.io/docs/prometheus/latest/querying/operators/).

Currently LogQL violates this behaviour. It returns whatever on the `left` (even if it's scalar). Look into the attached issue for more details.

Changes:
1. Fixed `MergeBinOps` by making it aware when to return left or right (depending on which is from vector originally)
2. Add tests in both `engine_test.go` and `evaluator_test.go` to lock this behaviour
3. Added appropriate comments to the types and functions.

Fixes https://github.com/grafana/loki/issues/10741

**Checklist**
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


MichelHollands (approved):
LGTM

pkg/logql/syntax/ast.go
```diff
@@ -1825,12 +1832,17 @@ func MergeBinOp(op string, left, right *promql.Sample, filter, isVectorCompariso
 		return res, nil
 	}
 
-	if filter {
-		// if a filter-enabled vector-wise comparison has returned non-nil,
-		// ensure we return the left hand side's value (2) instead of the
-		// comparison operator's result (1: the truthy answer)
+	if notReturnBool {
```
owen-d, line 1835: The `bool` flag doesn't have anything to do with whether we should perform `swap`. For an example, try `1 < bool $METRIC` in prometheus -- it'll return a vector of values with `0` or `1`.

owen-d (commented):
Great start, thanks!

pkg/logql/syntax/ast.go
kavirajk: I think it's been there because, swapping has no impact when return `bool` (vector of 0 or 1)

Means, both 
`1 < bool $METRIC` and `$METRIC > bool 1` will return exact same vector of values (0 or 1) without needed to swap. In other words, whether to return `left` or `right` sample  is irrelevant in case of returning bool.

owen-d (approved):

PR #11018, opened 2023-10-25
Refactor the proto Request interface to return time.Time

cstyan:
Part of our query path improvements. In addition to Karstens new protobuf formats for all the query request and responses, plus our use of just proto grpc rather than httpgrpc, we'd like refactor the proto Request interface to return time.Time for start and end instead of int64. This will allow his protobuf formats to use more `logproto` proto types.

There's still one or two tests that I need to fix.


jeschkies (commented):
The Protobuf should generate these getter methods.

pkg/querier/queryrange/limits.go
```diff
@@ -150,26 +149,26 @@ func (l limitsMiddleware) Do(ctx context.Context, r queryrangebase.Request) (que
...
-		if r.GetEnd() < minStartTime {
+		if r.GetEnd().Before(minStartTime) {
 			// The request is fully outside the allowed range, so we can return an
 			// empty response.
 			level.Debug(log).Log(
 				"msg", "skipping the execution of the query because its time range is before the 'max query lookback' setting",
-				"reqStart", util.FormatTimeMillis(r.GetStart()),
-				"redEnd", util.FormatTimeMillis(r.GetEnd()),
+				"reqStart", r.GetStart().String(),
+				"redEnd", r.GetEnd().String(),
```
jeschkies, line 160: I started breaking things but the nice bit about using `time.Time` is that we can do this now and void roundtrip conversions.

pkg/querier/queryrange/queryrangebase/queryrange.proto
```diff
@@ -14,8 +15,14 @@ option (gogoproto.unmarshaler_all) = true;
 
 message PrometheusRequest {
   string path = 1;
-  int64 start = 2;
-  int64 end = 3;
```
jeschkies, line 18: I don't think we can change the type. That would no be backwards compatible.

jeschkies: ..I think `PrometheusRequest` can actually go.

jeschkies: Ok, `PrometheusRequest` is only used for testing. I think this is fine.

jeschkies (approved):

PR #11028, opened 2023-10-25
logql: Add decryptAES template function

### LINKED ISSUE #11016: Add decryptAES to the available template functions
Our team currently logs some AES-encrypted fields. We frequently need to find logs with particular decrypted values, which means we need to encrypt and base64-encode the values first and then search for the resulting strings.

Include the decryptAES function from Sprig in the function map. Then we could do the decryption and search for the values directly in the query.

We could move the encrypted fields to a different data source entirely and join based on a key, but our current setup makes it very difficult to control who would have access to the decrypted information in that case.

precisionconage:
Exposes the `decryptAES` template function from sprig. This is useful for searching logs containing data encrypted with AES-256 CBC.

Fixes #11016.

None.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


docs/sources/query/template_functions.md
```diff
@@ -788,6 +788,24 @@ Examples:
...
 ```
 
+## decryptAES
+
+Decrypts a base64-encoded AES-256 CBC encrypted payload. The encoded payload bytes **must** have the following format
+for successful decryption:
+* bytes \[0:16\]: 16-byte IV (initialization vector)
+* bytes \[16:\]: ciphertext
+
+Additionally, only the first 32 bytes of the key will be used. Keys shorter than 32 bytes will be padded with NUL bytes.
```
JStickler, line 798: I think `null` is more widely used than `NUL`.
```suggestion
Additionally, only the first 32 bytes of the key will be used. Keys shorter than 32 bytes will be padded with null bytes. 
```

JStickler (commented):
[docs team] Thanks for updating the documentation as part of your PR.

Someone from the engineering team will need to review and approve your Go code before this can be merged.

PR #11029, opened 2023-10-25
WIP: return empty responses for logproto request types from

cstyan:
We need to return empty responses for logproto request types from NewEmptyResponse based on types that can be returned from DecodeRequest.

### DIFF
--- a/pkg/logproto/compat.go
+++ b/pkg/logproto/compat.go
@@ -276,6 +276,10 @@ func (m *IndexStatsRequest) LogToSpan(sp opentracing.Span) {
 	)
 }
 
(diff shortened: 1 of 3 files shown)


jeschkies (approved):
Could you add some tests? Otherwise it looks good to me 🙂

cstyan:
> Could you add some tests? Otherwise it looks good to me 🙂

I added a test for the actual function call that lead to the unsupported type. IMO we need a nicer refactoring of some of the middleware to be able to test this better but that can be done in a follow up.

jeschkies (commented):
This is odd. Why is this required? There shots be another response type

cstyan:
> This is odd. Why is this required? There shots be another response type

I'm not sure what you mean?

PR #11048, opened 2023-10-26
Handle proto type for volume endpoint response encoding

dannykopping:
Fixes the `invalid response format, got (*logproto.VolumeResponse)` error received when issuing requests against the `/loki/api/v1/index/volume` API.

### DIFF
--- a/pkg/querier/queryrange/codec.go
+++ b/pkg/querier/queryrange/codec.go
@@ -1005,6 +1005,10 @@ func encodeResponseJSONTo(version loghttp.Version, res queryrangebase.Response, 
 		if err := marshal.WriteVolumeResponseJSON(response.Response, w); err != nil {
 			return err
 		}
+	case *logproto.VolumeResponse:
+		if err := marshal.WriteVolumeResponseJSON(response, w); err != nil {
+			return err
+		}
 	default:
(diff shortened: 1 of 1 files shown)


pkg/querier/queryrange/codec.go
```diff
@@ -1005,6 +1005,10 @@ func encodeResponseJSONTo(version loghttp.Version, res queryrangebase.Response,
 		if err := marshal.WriteVolumeResponseJSON(response.Response, w); err != nil {
 			return err
 		}
+	case *logproto.VolumeResponse:
+		if err := marshal.WriteVolumeResponseJSON(response, w); err != nil {
+			return err
+		}
```
jeschkies, line 1011: Really? We use `logproto.VolumeResponse` and `VolumeResponse`?

jeschkies (approved):

jeschkies:
@dannykopping could you add a simple test case?

pkg/querier/queryrange/codec.go
jeschkies: I've just checked. We seem to handle both :see_no_evil: We probably should only use `logproto.VolumeResponse` and throw the other out.

jeschkies:
Replaced by https://github.com/grafana/loki/pull/11049

PR #11064, opened 2023-10-27
Improve observability of index queries

dannykopping:
Requests to the following APIs previously had limited or incorrect logging, which made understanding their behaviour at runtime difficult.

[GET /loki/api/v1/labels](https://grafana.com/docs/loki/latest/reference/api/#list-labels-within-a-range-of-time)
[GET /loki/api/v1/label/<name>/values](https://grafana.com/docs/loki/latest/reference/api/#list-label-values-within-a-range-of-time)
[GET /loki/api/v1/series](https://grafana.com/docs/loki/latest/reference/api/#list-series)
[GET /loki/api/v1/index/stats](https://grafana.com/docs/loki/latest/reference/api/#index-stats)
[GET /loki/api/v1/index/volume](https://grafana.com/docs/loki/latest/reference/api/#volume)
[GET /loki/api/v1/index/volume_range](https://grafana.com/docs/loki/latest/reference/api/#volume)

All of these APIs now have querier and query-frontend logs; sharding and time-based splitting are applied to these requests, and it's valuable to see all requests.

N/A

Logs are produced by both queriers and query-frontends, but only the QFs produced logs which showed `caller=metrics.go`. We regularly search on this substring to find our request log, but querier logs were shown as coming from `spanlogger.go`. I've included a small hack in `fixLogger` which adds `caller=metrics.go` to the log line if a `SpanLogger` is used. It's ugly, but I wanted to get this change in to help with our operational issues rather than refactoring the way we log on some occasions.

We're also considering adding limits (and later pagination) to these endpoints. I propose that we let this change soak in production for a couple weeks so we can make some data-driven decisions about what limits are reasonable.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/metrics.go
+++ b/pkg/logql/metrics.go
@@ -20,6 +20,7 @@ import (
 	logql_stats "github.com/grafana/loki/pkg/logqlmodel/stats"
 	"github.com/grafana/loki/pkg/util/httpreq"
 	util_log "github.com/grafana/loki/pkg/util/log"
+	"github.com/grafana/loki/pkg/util/spanlogger"
 )
 
 const (
@@ -89,7 +90,7 @@ func RecordRangeAndInstantQueryMetrics(
 	result promql_parser.Value,
 ) {
 	var (
-		logger        = util_log.WithContext(ctx, log)
+		logger        = fixLogger(ctx, log)
 		rt            = string(GetRangeType(p))
 		latencyType   = latencyTypeFast
 		returnedLines = 0
@@ -197,7 +198,7 @@ func RecordLabelQueryMetrics(
 	stats logql_stats.Result,
 ) {
 	var (
-		logger      = util_log.WithContext(ctx, log)
+		logger      = fixLogger(ctx, log)
 		latencyType = latencyTypeFast
 		queryType   = QueryTypeLabels
 	)
@@ -239,6 +240,17 @@ func RecordLabelQueryMetrics(
 	ingesterLineTotal.Add(float64(stats.Ingester.TotalLinesSent))
 }
 
+// fixLogger forces the given logger to include a caller=metrics.go kv pair.
+// The given logger might be a spanlogger instance, in which case it only logs caller=spanlogger.go:<line>.
+// We use `caller=metrics.go` when querying our logs for performance issues, and some logs were missing.
+func fixLogger(ctx context.Context, logger log.Logger) log.Logger {
+	nl := util_log.WithContext(ctx, logger)
+	if _, ok := logger.(*spanlogger.SpanLogger); ok {
+		return log.With(nl, "caller", "metrics.go")
+	}
+	return nl
+}
+
 func PrintMatches(matches []string) string {
 	// not using comma (,) as separator as matcher may already have comma (e.g: `{a="b", c="d"}`)
 	return strings.Join(matches, ":")
@@ -253,7 +265,7 @@ func RecordSeriesQueryMetrics(
 	stats logql_stats.Result,
 ) {
 	var (
-		logger      = util_log.WithContext(ctx, log)
+		logger      = fixLogger(ctx, log)
 		latencyType = latencyTypeFast
 		queryType   = QueryTypeSeries
 	)
@@ -303,7 +315,7 @@ func RecordVolumeQueryMetrics(
 	stats logql_stats.Result,
 ) {
 	var (
-		logger      = util_log.WithContext(ctx, log)
+		logger      = fixLogger(ctx, log)
 		latencyType = latencyTypeFast
 		queryType   = QueryTypeVolume
 	)
--- a/pkg/logql/metrics.go
+++ b/pkg/logql/metrics.go
@@ -223,6 +223,7 @@ func RecordLabelQueryMetrics(
 		"status", status,
 		"label", label,
 		"query", query,
+		"query_hash", HashedQuery(query),
 		"total_entries", stats.Summary.TotalEntriesReturned,
 	)
 
@@ -272,6 +273,7 @@ func RecordSeriesQueryMetrics(ctx context.Context, log log.Logger, start, end ti
 		"duration", time.Duration(int64(stats.Summary.ExecTime*float64(time.Second))),
 		"status", status,
 		"match", PrintMatches(match),
+		"query_hash", HashedQuery(PrintMatches(match)),
 		"total_entries", stats.Summary.TotalEntriesReturned)
 
 	if shard != nil {
@@ -311,6 +313,7 @@ func RecordStatsQueryMetrics(ctx context.Context, log log.Logger, start, end tim
 		"duration", time.Duration(int64(stats.Summary.ExecTime*float64(time.Second))),
 		"status", status,
 		"query", query,
+		"query_hash", HashedQuery(query),
 		"total_entries", stats.Summary.TotalEntriesReturned)
 
 	level.Info(logger).Log(logValues...)
@@ -318,14 +321,7 @@ func RecordStatsQueryMetrics(ctx context.Context, log log.Logger, start, end tim
 	execLatency.WithLabelValues(status, queryType, "").Observe(stats.Summary.ExecTime)
 }
 
-func RecordVolumeQueryMetrics(
-	ctx context.Context,
-	log log.Logger,
-	start, end time.Time,
-	query string,
-	status string,
-	stats logql_stats.Result,
-) {
+func RecordVolumeQueryMetrics(ctx context.Context, log log.Logger, start, end time.Time, query string, limit uint32, step time.Duration, status string, stats logql_stats.Result) {
 	var (
 		logger      = fixLogger(ctx, log)
 		latencyType = latencyTypeFast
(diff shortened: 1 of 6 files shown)


pkg/logql/metrics.go
```diff
@@ -211,49 +214,41 @@ func RecordLabelQueryMetrics(
...
+		"start_delta", time.Since(start),
+		"end_delta", time.Since(end),
 		"length", end.Sub(start),
 		"duration", time.Duration(int64(stats.Summary.ExecTime*float64(time.Second))),
 		"status", status,
 		"label", label,
 		"query", query,
-		"splits", stats.Summary.Splits,
-		"throughput", strings.Replace(humanize.Bytes(uint64(stats.Summary.BytesProcessedPerSecond)), " ", "", 1),
-		"total_bytes", strings.Replace(humanize.Bytes(uint64(stats.Summary.TotalBytesProcessed)), " ", "", 1),
```
slim-bean, line 221: why remove these? these seem useful?

pkg/logql/metrics.go
```diff
@@ -264,46 +259,71 @@ func RecordSeriesQueryMetrics(
...
+		"start_delta", time.Since(start),
+		"end_delta", time.Since(end),
 		"length", end.Sub(start),
 		"duration", time.Duration(int64(stats.Summary.ExecTime*float64(time.Second))),
 		"status", status,
 		"match", PrintMatches(match),
-		"splits", stats.Summary.Splits,
-		"throughput", strings.Replace(humanize.Bytes(uint64(stats.Summary.BytesProcessedPerSecond)), " ", "", 1),
-		"total_bytes", strings.Replace(humanize.Bytes(uint64(stats.Summary.TotalBytesProcessed)), " ", "", 1),
-		"total_entries", stats.Summary.TotalEntriesReturned,
```
slim-bean, line 278: splits is still useful for series queries

pkg/logql/metrics.go
```diff
@@ -264,46 +259,71 @@ func RecordSeriesQueryMetrics(
...
-	bytesPerSecond.WithLabelValues(status, queryType, "", latencyType, sharded).
-		Observe(float64(stats.Summary.BytesProcessedPerSecond))
-	execLatency.WithLabelValues(status, queryType, "").
-		Observe(stats.Summary.ExecTime)
-	chunkDownloadLatency.WithLabelValues(status, queryType, "").
-		Observe(stats.ChunksDownloadTime().Seconds())
-	duplicatesTotal.Add(float64(stats.TotalDuplicates()))
-	chunkDownloadedTotal.WithLabelValues(status, queryType, "").
-		Add(float64(stats.TotalChunksDownloaded()))
-	ingesterLineTotal.Add(float64(stats.Ingester.TotalLinesSent))
```
slim-bean, line 294: are these metrics being removed? I'm not seeing where they are going?

pkg/logql/metrics.go
```diff
@@ -314,23 +334,36 @@ func RecordVolumeQueryMetrics(
...
+		"step", step,
+		"limit", limit,
 		"length", end.Sub(start),
 		"duration", time.Duration(int64(stats.Summary.ExecTime*float64(time.Second))),
 		"status", status,
 		"splits", stats.Summary.Splits,
 		"total_entries", stats.Summary.TotalEntriesReturned,
-		"cache_volume_results_req", stats.Caches.VolumeResult.EntriesRequested,
-		"cache_volume_results_hit", stats.Caches.VolumeResult.EntriesFound,
-		"cache_volume_results_download_time", stats.Caches.VolumeResult.CacheDownloadTime(),
```
slim-bean, line 329: if this is the volume metrics endpoint, woudln't we want the cache stats included with it?

pkg/logql/metrics.go
```diff
@@ -264,46 +259,71 @@ func RecordSeriesQueryMetrics(
...
-		"total_entries", stats.Summary.TotalEntriesReturned,
+		"query_hash", HashedQuery(PrintMatches(match)),
+		"total_entries", stats.Summary.TotalEntriesReturned)
+
+	if shard != nil {
+		logValues = append(logValues,
+			"shard_num", shard.Shard,
+			"shard_count", shard.Of,
+		)
+	}
```
slim-bean, line 284: nice, this seems like it could be helpful

pkg/logql/metrics.go
dannykopping: Removed, yeah. They never worked; we don't capture these stats in the index

pkg/logql/metrics.go
dannykopping: I don't understand the question; those stats are still included

pkg/logql/metrics.go
dannykopping: I _think_ this stat was not captured. Will check again, same for the other instance

pkg/querier/http.go
```diff
@@ -271,21 +269,38 @@ func (q *QuerierAPI) SeriesHandler(ctx context.Context, req *logproto.SeriesRequ
...
+	statResult := statsCtx.Result(time.Since(start), queueTime, 1)
+	log := spanlogger.FromContext(ctx)
+	statResult.Log(level.Debug(log))
+
+	status := 200
+	if err != nil {
+		status, _ = serverutil.ClientHTTPStatusAndError(err)
+	}
+
+	logql.RecordStatsQueryMetrics(ctx, log, req.Start, req.End, req.Query, strconv.Itoa(status), statResult)
```
slim-bean, line 302: this kind of feels like we are breaking an existing pattern that we have here: https://github.com/grafana/loki/blob/a05744a385167e994f8d8812b0c2649c2321e5fd/pkg/querier/queryrange/stats.go#L49

not sure if it's possible to add additional types to that statement to follow that pattern?

pkg/querier/http.go
```diff
@@ -302,6 +323,18 @@ func (q *QuerierAPI) VolumeHandler(ctx context.Context, req *logproto.VolumeRequ
...
+	statResult := statsCtx.Result(time.Since(start), queueTime, 1)
+	log := spanlogger.FromContext(ctx)
+	statResult.Log(level.Debug(log))
+
+	status := 200
+	if err != nil {
+		status, _ = serverutil.ClientHTTPStatusAndError(err)
+	}
+
+	logql.RecordVolumeQueryMetrics(ctx, log, req.From.Time(), req.Through.Time(), req.GetQuery(), uint32(req.GetLimit()), time.Duration(req.GetStep()), strconv.Itoa(status), statResult)
```
slim-bean, line 336: same question as above

slim-bean:
I love this PR!!! sorry it feels like I am nitting it to death 😬

pkg/logql/metrics.go
dannykopping: Confirmed, in neither case is `stats.Summary.Splits` recorded

pkg/querier/http.go
dannykopping: We capture at both levels; this one captures individual requests from queriers, and the code you referenced captures the stats at the QF level

pkg/querier/http.go
dannykopping: Ditto

dannykopping:
> I love this PR!!! sorry it feels like I am nitting it to death 😬

Not at all! I appreciate the high level of scrutiny. Much prefer it to a rubber stamp

pkg/logql/metrics.go
dannykopping: Apparently I was wrong; I added them back 👍

pkg/logql/metrics.go
```diff
@@ -314,23 +334,36 @@ func RecordVolumeQueryMetrics(
...
+
 	level.Info(logger).Log(
 		"latency", latencyType,
 		"query_type", queryType,
 		"query", query,
 		"query_hash", HashedQuery(query),
+		"start", start.Format(time.RFC3339Nano),
+		"end", end.Format(time.RFC3339Nano),
+		"start_delta", time.Since(start),
```
kavirajk, line 351: Q: for my understanding:

why `start_delta` and `end_delta`? And both calculated from `time.Since` (meaning the time where it is logging in `metrics.go`?. Failing to understand what's the relation here.

why can't be just `delta: end-start`?

pkg/logql/metrics.go
slim-bean: ooohhh, I was just looking at the names. not the values setting them.

I think I would prefer to leave "volume" in the name?

hrmmm still thinking about this

pkg/logql/metrics.go
dannykopping: We use the same thing in our query logs. These figures useful for knowing the difference in time between the start of the query range and when the query began (or more strictly, when it completed - which is normally not long after).

This is used to see how many queries are being executed for some time-range other than from now to now-6h, for example.

pkg/logql/metrics.go
```diff
@@ -211,49 +214,41 @@ func RecordLabelQueryMetrics(
...
-	chunkDownloadLatency.WithLabelValues(status, queryType, "").
-		Observe(stats.ChunksDownloadTime().Seconds())
-	duplicatesTotal.Add(float64(stats.TotalDuplicates()))
-	chunkDownloadedTotal.WithLabelValues(status, queryType, "").
-		Add(float64(stats.TotalChunksDownloaded()))
-	ingesterLineTotal.Add(float64(stats.Ingester.TotalLinesSent))
+// fixLogger forces the given logger to include a caller=metrics.go kv pair.
+// The given logger might be a spanlogger instance, in which case it only logs caller=spanlogger.go:<line>.
+// We use `caller=metrics.go` when querying our logs for performance issues, and some logs were missing.
+func fixLogger(ctx context.Context, logger log.Logger) log.Logger {
```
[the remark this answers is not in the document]
dannykopping: It might not be the best name, but it's the most descriptive I could think of. It's much clearer about its intent than `metricsLogger` IMHO; I wanted to explicitly show that something weird was being done.

pkg/logql/metrics.go
dannykopping: I removed them because they felt unnecessary (since it's a volume query after all) and it made the keys very long

pkg/querier/http.go
slim-bean: hrm, interesting, maybe my question now is: why don't we do this for the instant and range query handlers?

Maybe we don't need to unpack that right now though.

pkg/logql/metrics.go
slim-bean: my argument towards keeping volume in the name would be to avoid ambiguity and removing a case where I could see a common mistake made:

```
sum(sum_over_time({} |= "metrics.go | logfmt | unwrap cache_results_hit[1m]))
/
sum(sum_over_time({} |= "metrics.go | logfmt | unwrap cache_results_req[1m]))
```

This is a query I've done a lot in the past to see how effective our results cache is, but with this change I would both need to know and remember it would be including stats for volume queries which I think is too error prone. WDYT?

dannykopping: Fair point 👍 I'll add them back in

slim-bean (approved):
LGTM!

kavirajk (approved):

PR #11079, opened 2023-10-30
Support categorized labels in Tailing

salvacorts:

This is a follow-up PR for https://github.com/grafana/loki/pull/10419 adding support for tailing.

I tested it on a dev cell and works fine.
image

**Note**: With these changes, the JSON marshal unmarshal functions for the tail are no longer used ([example][1]) so I think we can remove them. Also, the new Tail response is no longer used, so we can also make it an alias to the _legacy_ one. Let's do it on a follow-up PR to avoid making this one bigger.

[1]: https://github.com/grafana/loki/blob/52a3f16039dd5ff655fc3681257d99794f620ec4/pkg/loghttp/entry.go#L210-L238


vlad-diachenko (approved):
looks great! thanks @salvacorts

PR #11088, opened 2023-10-31
Bloom/fuse queries

owen-d:
Introduces a "Fuse" operation on a single block querier which multiplexes queries against a single block. This is advantageous for a few reasons:
* A "query" contains one of more series, with each series needing to test one or more chunks for some search string(s).
* Blocks are sorted by fingerprint, which are the result of a hash function. This means series are ~uniformly distributed across the hash keyspace (`uint64` in our case). Because of this, a query for some number of series are likely to be spread out across the block itself.
* Blocks are organized into individually compressed "pages". In essence, querying them requires Seeking to the correct offset, decompressing the page, then seeking within the page to the actual offset for a particular series.
* Since the series for any query are likely spread out across the block, performing `n` queries decompresses each page `n` times.
* By multiplexing many queries into a single pass iteration of the block, we amortize the cost of loading & decompressing

Misc:
Also introduces a bunch of generic-powered iterators. These can probably be optimized (I bet they use dynamic dispatch under the hood), but they're very convenient to prototype with and we can improve them in the future if it proves warranted.

### DIFF
--- a/pkg/storage/bloom/v1/TODO.md
+++ b/pkg/storage/bloom/v1/TODO.md
@@ -1,5 +1,3 @@
-* Should be able to read bloom as a []byte without copying it during decoding
-  * It's immutable + partition offsets are calculable, etc
 * Less copying! I've taken some shortcuts we'll need to refactor to avoid copying []byte around in a few places
 * more sophisticated querying methods
 * queue access to blooms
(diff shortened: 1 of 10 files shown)


slim-bean (approved):
LTM

PR #10840, opened 2023-10-10
Remove shared_store and shared_store_key_prefix from shipper and compactor

ashwanthgoli:
**What this PR does / why we need it**:
#### Removes `shared_store` and `shared_store_key_prefix` from index shipper and compactor configs and their corresponding CLI flags.

- `-tsdb.shipper.shared-store`
- `-boltdb.shipper.shared-store`
- `-tsdb.shipper.shared-store.key-prefix`
- `-boltdb.shipper.shared-store.key-prefix`
- `-boltdb.shipper.compactor.shared-store`
- `-boltdb.shipper.compactor.shared-store.key-prefix`

`shared_store` has been a confusing option allowing users to easily misconfigure Loki.
Going forward `object_store` setting in the [period_config](https://grafana.com/docs/loki/latest/configure/#period_config) (which already configured the store for chunks) will be used to configure store for the index.

And the newly added `path_prefix` option under the `index` key in `period_config` will configure the path under which index tables are stored. 

This change enforces chunks and index files for a given period reside together in the same storage bucket. More details in the upgrade guide.

---

`-compactor.delete-request-store` has to be **explicitly configured** going forward. Without setting this, loki wouldn't know which object store to use for storing delete requests. Path prefix for storing deletes is decided by `-compactor.delete-request-store.key-prefix` which defaults to `index/`.

**Checklist**
- [X] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [X] Documentation added
- [X] Tests updated
- [x] `CHANGELOG.md` updated
  - [ ] If the change is worth mentioning in the release notes, add `add-to-release-notes` label
- [X] Changes that require user attention or interaction to upgrade are documented in `docs/sources/setup/upgrade/_index.md`
- [ ] For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)

PR #11091, opened 2023-10-31
operator: Add automatic stream sharding support

periklis:
The following PR adds [automatic stream sharding](https://grafana.com/docs/loki/v2.9.x/operations/automatic-stream-sharding/) support for all `LokiStack` deployment sizes in the Loki Operator. In particular it replaces present defaults like `per_stream_rate_limit: 3MB` with a `desiredRate: 3MB`

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/internal/manifests/internal/config/loki-config.yaml
```diff
@@ -200,9 +200,18 @@ limits_config:
 {{- end }}
 {{- end }}{{- end }}
   max_cache_freshness_per_query: 10m
+{{- if .Stack.Limits.Global.IngestionLimits.PerStreamRateLimit }}
   per_stream_rate_limit: {{ .Stack.Limits.Global.IngestionLimits.PerStreamRateLimit }}MB
+{{- end }}
+{{- if .Stack.Limits.Global.IngestionLimits.PerStreamRateLimitBurst }}
   per_stream_rate_limit_burst: {{ .Stack.Limits.Global.IngestionLimits.PerStreamRateLimitBurst }}MB
+{{- end }}
```
xperimental, line 208: Use `with`?

```suggestion
{{- with .Stack.Limits.Global.IngestionLimits.PerStreamRateLimit }}
  per_stream_rate_limit: {{ . }}MB
{{- end }}
{{- with .Stack.Limits.Global.IngestionLimits.PerStreamRateLimitBurst }}
  per_stream_rate_limit_burst: {{ . }}MB
{{- end }}
```

operator/internal/manifests/internal/config/loki-config.yaml
```diff
@@ -200,9 +200,18 @@ limits_config:
...
   split_queries_by_interval: 30m
+{{- if .Stack.Limits.Global.IngestionLimits.DesiredRate }}
+  shard_streams:
+    enabled: true
+    desired_rate: {{ .Stack.Limits.Global.IngestionLimits.DesiredRate }}
```
xperimental, line 213: Use `with`?

```suggestion
{{- with .Stack.Limits.Global.IngestionLimits.DesiredRate }}
  shard_streams:
    enabled: true
    desired_rate: {{ . }}
```

operator/internal/manifests/internal/config/loki-runtime-config.yaml
```diff
@@ -32,6 +32,11 @@ overrides:
     {{- if $l.PerStreamRateLimitBurst }}
     per_stream_rate_limit_burst: {{ $l.PerStreamRateLimitBurst }}MB
     {{- end }}
+    {{- if $l.DesiredRate }}
+    shard_streams:
+      enabled: true
+      desired_rate: {{ $l.DesiredRate }}
```
xperimental, line 38: Use `with`?

```suggestion
    {{- with $l.DesiredRate }}
    shard_streams:
      enabled: true
      desired_rate: {{ . }}
```

operator/internal/validation/lokistack.go
```diff
@@ -115,6 +120,89 @@ func (v *LokiStackValidator) validateHashRingSpec(s lokiv1.LokiStackSpec) field.
...
+		))
+	}
+
+	return errs
+}
+
+// validateTenantIngestionLimits checks if a given tenant IngestionLimitSpec is valid for:
+// - Tenant fields `desiredRate` and `perStreamRateLimit` not both set at a time.
+// - Global field `desiredRate` and tenant field `perStreamRateLimit` not both set at a time.
+// - Tenant field `desiredRate` and global field `perStreamRateLimit` not both set at a time.
```
xperimental, line 173: Am I understanding this correctly that the configuration must be "either all use desiredRate or all use perStreamRateLimit" and there's never a valid combination which uses desiredRate on one side and perStreamRateLimit on the other?

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -632,6 +632,14 @@ type QueryLimitSpec struct {
 
 // IngestionLimitSpec defines the limits applied at the ingestion path.
 type IngestionLimitSpec struct {
+	// DesiredRate defines the desired ingestion rate per second that LokiStack should
+	// target applyfing automatic stream sharding. Units Bytes.
```
JoaoBraveCoding, line 636:
```suggestion
	// target applying automatic stream sharding. Units Bytes.
```

operator/internal/manifests/internal/config/build_test.go
JoaoBraveCoding, line 1: Didn't check but if we replaced all occurrences of the old limits I'm afraid bug might slip through in the future. I would be on the side of at least keeping 1 instance of these

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -632,6 +632,14 @@ type QueryLimitSpec struct {
...
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="Desired Ingestion Rate (in Bytes)"
+	DesiredRate int32 `json:"desiredRate,omitempty"`
```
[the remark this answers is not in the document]
periklis: I will stick with MB and an `int32` to keep all knobs similar. On the rendering side we have many discrepancies because of the Loki Config format, some accept bytes (i.e. `desired_rate`) while other accept megabytes w/o a suffix (i.e. `ingestion_rate_mb`) and others with a suffix (`per_stream_rate_limit`).

We can save the overhead of `resource.Quantity` for another day, e.g. LokiStack v2. ;)

operator/internal/validation/lokistack.go
periklis: It's not a choice of valid vs. invalid although the webhook implies this. My call here is to keep things uniform and simple (as we try to keep the stacks similar), either you opt-in for Loki doing the sharding or you keep managing the rates yourself.

operator/internal/validation/lokistack.go
```diff
@@ -115,6 +120,89 @@ func (v *LokiStackValidator) validateHashRingSpec(s lokiv1.LokiStackSpec) field.
...
+		fp := field.NewPath("spec", "limits", "global", "ingestion", "desiredRate")
+		errors = append(errors, validateIngestionLimits(s.Global.IngestionLimits, fp)...)
+	}
+
+	for id, tenant := range s.Tenants {
+		if tenant.IngestionLimits == nil {
+			continue
+		}
+
+		fp := field.NewPath("spec", "limits", "tenants", id, "ingestion", "desiredRate")
```
JoaoBraveCoding, line 143: Question, why do we have in this function the field path lead to `desiredRate` but then in `validateTenantIngestionLimits` we use on the second set of if's `validateTenantIngestionLimits`? Shouldn't we be consistent and always use the same? Especially because it's the same error. I would always go with `perStreamRateLimit` since that is the oldest field.

periklis: I have simplifed the validation funcs a bit, PTAL

operator/apis/loki/v1/v1.go
```diff
@@ -83,6 +83,8 @@ var (
 	ErrReplicationSpecConflict = errors.New("replicationSpec and replicationFactor (deprecated) cannot be used at the same time")
 	// ErrIPv6InstanceAddrTypeNotAllowed when the default InstanceAddrType is used with enableIPv6.
 	ErrIPv6InstanceAddrTypeNotAllowed = errors.New(`instanceAddrType "default" cannot be used with enableIPv6 at the same time`)
+	// ErrDesiredRateAndPerStreamRateNotAllowed when there's a conflict in the rate limiting configuration because both desiredRate and perStreamRateLimit are set.
+	ErrDesiredRateAndPerStreamRateNotAllowed = errors.New("fields desiredRate and perStreamRateLimit cannot be used together")
```
DylanGuedes, line 87: fyi: in some scenarios you'd benefit by tuning both. Ex: if you ingest a spike going from 0MB/s ingestion to 5MB/s, you'll hit the per-stream rate limit even with stream sharding enabled, because Loki wouldn't have enough time to adjust to the current load. In this scenario, setting the perStreamRateLimit to 5MB/s would help avoiding discards.

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -636,11 +636,11 @@ type QueryLimitSpec struct {
 // IngestionLimitSpec defines the limits applied at the ingestion path.
 type IngestionLimitSpec struct {
 	// DesiredRate defines the desired ingestion rate per second that LokiStack should
-	// target applyfing automatic stream sharding. Units Bytes.
+	// target applying automatic stream sharding. Unit MB.
```
xperimental, line 639: What we're actually using are `MiB`, should we mention this in the comment?

```suggestion
	// target applying automatic stream sharding. Unit MiB.
```

operator/apis/loki/v1/v1.go
xperimental: I'm not sure I understand this correctly. I was under the impression, that using `desiredRate` without setting a `perStreamRateLimit` would effectively turn the per-stream rate-limiting into a "soft limit" where peaks would not return a 429 and having a stream with high log rate would be turned into multiple streams internally automatically to avoid having a single large stream.

operator/internal/manifests/internal/config/build.go
```diff
@@ -54,3 +56,7 @@ func Build(opts Options) ([]byte, []byte, error) {
 	}
 	return cfg, rcfg, nil
 }
+
+func toBytes(s int32) int32 {
+	return int32(float32(s) * (1 << 20))
```
xperimental, line 61: I think I would simplify this with a constant and less conversions (also, because I'm not converting it to 32bit in the end, it produces the correct value in case we ever exceed a Rate of 2GiB):

```suggestion
// move the constant to the const block at the top
const mebiBytesFactor = 1 << 20

func toBytes(s int32) int {
	return int(s) * mebiBytesFactor
```

operator/internal/manifests/internal/config/loki-config.yaml
```diff
@@ -201,18 +201,19 @@ limits_config:
...
+{{- with .Stack.Limits.Global.IngestionLimits.PerStreamRateLimitBurst }}
+  per_stream_rate_limit_burst: {{ . }}MB
 {{- end }}
   split_queries_by_interval: 30m
-{{- if .Stack.Limits.Global.IngestionLimits.DesiredRate }}
+{{- with .Stack.Limits.Global.IngestionLimits.DesiredRate }}
   shard_streams:
     enabled: true
-    desired_rate: {{ .Stack.Limits.Global.IngestionLimits.DesiredRate }}
+    desired_rate: {{ toBytes . }}
```
xperimental, line 214: (this invalidates the other comment about `toBytes`)

Looking at the Loki sources, it looks like `desired_rate` is using the same type as `per_stream_rate_limit` (`flagext.ByteSize`) and should be able to take a `MB` suffix, even though this is not mentioned in the docs. Both are listed as `<int>` in the docs though as well.

```suggestion
    desired_rate: {{ . }}MB
```

operator/internal/manifests/internal/config/loki-config.yaml
```diff
@@ -201,18 +201,19 @@ limits_config:
...
 {{- end }}
+  allow_structured_metadata: true
```
xperimental, line 216: Does this belong to the current PR?

operator/internal/manifests/internal/config/loki-runtime-config.yaml
```diff
@@ -32,10 +32,10 @@ overrides:
     {{- if $l.PerStreamRateLimitBurst }}
     per_stream_rate_limit_burst: {{ $l.PerStreamRateLimitBurst }}MB
     {{- end }}
-    {{- if $l.DesiredRate }}
+    {{- with $l.DesiredRate }}
     shard_streams:
       enabled: true
-      desired_rate: {{ $l.DesiredRate }}
+      desired_rate: {{ . }}
```
xperimental, line 38: I think this is either missing the `toBytes` or `MB` suffix:

```suggestion
      desired_rate: {{ . }}MB
```

xperimental (commented):
Review in code only as I'm on the move and not testing on a cluster.

operator/internal/manifests/internal/config/build.go
periklis: Sticking with the suggestion below to use the `flagext.bytesize` version some sanity in our config.

operator/internal/manifests/internal/config/loki-config.yaml
periklis: Nice find, will use that and get rid of `toBytes`

operator/internal/manifests/internal/config/loki-runtime-config.yaml
periklis: Good catch!

operator/apis/loki/v1/v1.go
DylanGuedes: it is a little confusing. Because it takes time to propagate the right number of shards, if you get a spike too quickly for a stream the system won't have enough time to adjust. In this scenario, the per-stream-rate-limit dictates if your logs will be rejected or not.

periklis: FWIW and IIF I am reading the rate store code correctly, the rates are based on observed data points on the ingester side. So for things not seen yet we cannot create adapt the shard count. Thus spiky streams will be rate limited by the per-stream rate limit. Correct?

periklis: I am referring to this: https://github.com/grafana/loki/blob/v2.9.2/pkg/distributor/ratestore.go#L301

DylanGuedes: > FWIW and IIF I am reading the rate store code correctly, the rates are based on observed data points on the ingester side. So for things not seen yet we cannot create adapt the shard count. Thus spiky streams will be rate limited by the per-stream rate limit. Correct?

exactly it.

DylanGuedes: > FWIW and IIF 

just learned what FWIW means, but couldn't really find what IIF stands for :rofl:

periklis: The emoji and acronym domination among engineering folks is sometimes excessive. sorry for that :see_no_evil:

DylanGuedes: nrta, idimsf

(no reason to apologize, I do it myself)

operator/internal/manifests/internal/config/loki-config.yaml
```diff
@@ -201,9 +201,18 @@ limits_config:
...
 {{- end }}{{- end }}
   max_cache_freshness_per_query: 10m
-  per_stream_rate_limit: {{ .Stack.Limits.Global.IngestionLimits.PerStreamRateLimit }}MB
-  per_stream_rate_limit_burst: {{ .Stack.Limits.Global.IngestionLimits.PerStreamRateLimitBurst }}MB
+{{- with .Stack.Limits.Global.IngestionLimits.PerStreamRateLimit }}
+  per_stream_rate_limit: {{ . }}MB
+{{- end }}
+{{- with .Stack.Limits.Global.IngestionLimits.PerStreamRateLimitBurst }}
+  per_stream_rate_limit_burst: {{ . }}MB
+{{- end }}
```
xperimental, line 209: Doesn't the other conversation mean that these limits are not optional at all and the operator should expect a limit and burst limit to be configured (even if it is just the default)?

operator/apis/loki/v1/v1.go
DylanGuedes: cool, I'll resolve the thread. if you need any clarifications regarding stream sharding please let me know (you can ping me on Slack)

operator/internal/manifests/internal/config/loki-config.yaml
periklis: Actually the `sizes.go` declares a new default per-stream-rate limit of 5MB and leaves the default for burst to the Loki config. WDYT? (maybe just removing the template conditional for `per_stream_rate_limit` is enough?!?)

operator/internal/manifests/internal/sizes.go
```diff
@@ -235,14 +235,14 @@ var StackSizeTable = map[lokiv1.LokiStackSizeType]lokiv1.LokiStackSpec{
...
 				IngestionLimits: &lokiv1.IngestionLimitSpec{
 					// Defaults from Loki docs
-					IngestionRate:           4,
-					IngestionBurstSize:      6,
-					MaxLabelNameLength:      1024,
-					MaxLabelValueLength:     2048,
-					MaxLabelNamesPerSeries:  30,
-					MaxLineSize:             256000,
-					PerStreamRateLimit:      3,
-					PerStreamRateLimitBurst: 15,
```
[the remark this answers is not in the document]
periklis: The old defaults disappeared because of the initial validation case that not both `desired_rate` and `per_stream_rate_limit` can co-exist. As per the above discussion with @DylanGuedes I added the `per_stream_rate_limit` back but with a new value set to `5 MB`. This means our setups aim for a desired rate of `3 MB/s` but have plenty of room for bursty streams. WDYT?

xperimental (commented):
Seems to work nicely :+1: 

Left two more comments.

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -698,6 +690,14 @@ type IngestionLimitSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="Max Line Size"
 	MaxLineSize int32 `json:"maxLineSize,omitempty"`
 
+	// PerStreamDesiredRate defines the desired ingestion rate per second that LokiStack should
+	// target applying automatic stream sharding. Units MB.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="Per Stream Desired Rate (in MB)"
+	PerStreamDesiredRate int32 `json:"desiredRate,omitempty"`
```
xperimental, line 699: I think you missed to update the JSON name? It's still `desiredRate`.

xperimental (approved):

PR #11094, opened 2023-10-31
operator: Add support for blocking queries per tenant

periklis:
The following PR adds support for blocking queries per tenant for the Loki Operator.

[LOG-4552](https://issues.redhat.com/browse/LOG-4552)

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/apis/loki/v1/lokistack_types.go
```diff
@@ -630,6 +632,68 @@ type QueryLimitSpec struct {
...
+//
+// +kubebuilder:validation:Enum=filter;limited;metric
+type BlockedQueryType string
+
+const (
+	// BlockedQueryFilter defines the blocking type for queries with at least one log filter.
+	BlockedQueryFilter = "filter"
+	// BlockedQueryLimited defines the blocking type for queries without a filter or a metric aggregation.
+	BlockedQueryLimited = "limited"
+	// BlockedQueryMetric findes the blocking type for queries with an aggregation.
```
JoaoBraveCoding, line 645:
```suggestion
	// BlockedQueryMetric defines the blocking type for queries with an aggregation.
```

JoaoBraveCoding (commented):
One comment but lgtm

operator/config/manager/kustomization.yaml
```diff
@@ -5,5 +5,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openshift-logging/loki-operator
-  newTag: 0.1.0
+  newName: quay.io/periklis/loki-operator
```
JoaoBraveCoding, line 8: Probably a miss `git add`

operator/config/overlays/openshift/manager_related_image_patch.yaml
```diff
@@ -7,6 +7,7 @@ spec:
     spec:
       containers:
         - name: manager
+          imagePullPolicy: Always
```
JoaoBraveCoding, line 10: Probably a miss `git add`

operator/hack/lokistack_gateway_ocp.yaml
JoaoBraveCoding, line 1: Probably a miss `git add`

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -630,6 +630,60 @@ type QueryLimitSpec struct {
...
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch",displayName="Regex"
+	Regex bool `json:"regex,omitempty"`
+	// Types defines the list of query types that should be considered for blocking.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Query Types"
+	Types []BlockedQueryType `json:"types,omitempty"`
```
JoaoBraveCoding, line 672: Can users provide duplicate types? Does it have any impact on the final config?

periklis: :thinking: That's a good question, will check this out on the Loki code base. If it is an error state, we should add some validation code.

periklis: @JoaoBraveCoding Loki does not care if we add the same value multiple times.

operator/internal/manifests/internal/config/loki-runtime-config.yaml
```diff
@@ -54,6 +54,23 @@ overrides:
     {{- if $spec.QueryLimits.CardinalityLimit }}
     cardinality_limit: {{ $spec.QueryLimits.CardinalityLimit }}
     {{- end }}
+    {{- with $l.Blocked }}
+    blocked_queries:
+    {{- range $blockedQuery := . }}
+    - {{ if $blockedQuery.Pattern -}}
+      pattern: '{{ $blockedQuery.Pattern }}'
```
xperimental, line 61: Do we need to pass `Pattern` through a function that escapes `'` or `"` (depending on what we use in the output)?

periklis: You might be right. After looking on the Loki source code it expects the pattern to be a valid compiling regex. if we leave single quotes in the output `loki-runtime.yaml` the only case IMHO that needs attention is using single quotes in the regex statement (e.g. `.*msg="this 'service' is bad.*"`).

Creating a proper custom resource YAML with properly quoted single quotes results in an error as of now, because when we use the `pattern` fields value we get the non-escaped form, e.g.

For:
```yaml
  limits:
    tenants:
      application:
        queries:
          blocked:
          - pattern: '.*msg="this ''service'' is bad.*'
            regex: true
            types:
            - metric
```
results in an invalid `loki-runtime.yaml`:
```yaml
overrides:
  application:
    query_timeout: 3m
    blocked_queries:
    - pattern: '.*msg="this 'service' is bad.*'
      regex: true
      types: metric
```
and the pods get stuck with the following error:
```
level=error ts=2023-11-27T09:45:38.019488231Z caller=loki.go:517 msg="module failed" module=runtime-config error="invalid service state: Failed, expected: Running, failure: failed to load runtime config: unmarshal file \"/etc/loki/config/runtime-config.yaml\": yaml: line 5: did not find expected key"
```

WDYT?

periklis: @xperimental Do you think we should render the regex pattern field's value as YAML literal scalar to escape the above? Something like this:
```
    {{- range $blockedQuery := . }}
    - {{ with $blockedQuery.Pattern -}}
      pattern: |
          {{ . }}`
      {{- end }}
```

xperimental: The block literal is probably a good idea :+1: 

We still need to process `Pattern` a bit, because it _might_ contain new-lines which we then need to treat specially. I have put comments in the different places to address this (`yamlBlock`).

operator/internal/manifests/internal/config/loki-runtime-config.yaml
```diff
@@ -54,6 +54,23 @@ overrides:
     {{- if $spec.QueryLimits.CardinalityLimit }}
     cardinality_limit: {{ $spec.QueryLimits.CardinalityLimit }}
     {{- end }}
+    {{- with $l.Blocked }}
+    blocked_queries:
+    {{- range $blockedQuery := . }}
+    - {{ with $blockedQuery.Pattern -}}
+      pattern: '{{ . }}'
+      {{- end -}}
```
xperimental, line 62:
```suggestion
    - {{ with $blockedQuery.Pattern -}}
      pattern: |
        {{ . | yamlBlock "        " }}
      {{- end -}}
```

---

This also needs changes in `build.go` which I can not attach comments to, because it is not changed anymore:

```golang
func yamlBlock(in reflect.Value, indent string) string {
	inStr := in.String()
	lines := strings.Split(strings.TrimRight(inStr, "\n"), "\n")
	return strings.Join(lines, "\n"+indent)
}
```

and

```golang
	lokiRuntimeConfigYAMLTmpl = template.Must(template.New("runtime-config").Funcs(template.FuncMap{
		"yamlBlock": yamlBlock,
	}).ParseFS(lokiRuntimeConfigYAMLTmplFile, "loki-runtime-config.yaml"))
```

operator/internal/manifests/internal/config/loki-runtime-config.yaml
```diff
@@ -54,6 +54,23 @@ overrides:
     {{- if $spec.QueryLimits.CardinalityLimit }}
     cardinality_limit: {{ $spec.QueryLimits.CardinalityLimit }}
     {{- end }}
+    {{- with $l.Blocked }}
+    blocked_queries:
+    {{- range $blockedQuery := . }}
+    - {{ with $blockedQuery.Pattern -}}
```
xperimental, line 60: I think we need to add another validation to the LokiStack: all attributes of `BlockedQuerySpec` are optional, because depending on the use-case you only need a few of them. But our config-generation depends on _at least one_ of the attributes being set, otherwise we produce invalid YAML.

Would it make sense to add a validation to the LokiStack that emits an error if there is a `BlockedQuerySpec` which contains no values? Example:

```yaml
spec:
  limits:
    tenants:
      application:
        queries:
          blocked:
          - {}
```

xperimental: Ok, small correction: The generated YAML does not seem to be invalid, like I thought. So it's not technically needed to validate this. Is it still a good idea, though?

xperimental (approved):

PR #11096, opened 2023-10-31
add vulnerability scanning github action

trevorwhitney:

Adds a GitHub action that will run `snyk` and `trivy` scans against PRs and publish the results as a comment on the PR.

Examples of what the comments will look like can be seen on this PR. It will only add a single comment for each scanner in the future, this one has multiples because of some of the iterations it went though.


.github/workflows/vulnerability-scan.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Checkout code
+        uses: actions/checkout@master
+      - name: Run Snyk to check for vulnerabilities
+        uses: snyk/actions/golang@master
+        continue-on-error: true  # To make sure that PR comment is made
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          command: test
+          args: --severity-threshold=high --json-file-output=snyk.json
```
kavirajk, line 22: Q: rationale behind `high` value here?

kavirajk (approved):
LGTM. thanks @trevorwhitney. Just one non-blocking clarification.

.github/workflows/vulnerability-scan.yml
trevorwhitney: thought was to reduce noise by only showing high and above. I need to add that to the trivy action as well. would you prefer we make it medium? remove completely?

PR #11099, opened 2023-11-01
Fix scalar timestamp

shantanualsi:
This PR fixes an issue in literal expressions where the timestamp returned in response was incorrect. The expected time should be the current time.

image

The date returned here is from 2021.
image

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/querier/queryrange/codec.go
```diff
@@ -1314,7 +1314,7 @@ func toProtoScalar(v loghttp.Scalar) []queryrangebase.SampleStream {
 	res = append(res, queryrangebase.SampleStream{
 		Samples: []logproto.LegacySample{{
 			Value:       float64(v.Value),
-			TimestampMs: v.Timestamp.UnixNano() / 1e6,
+			TimestampMs: v.Timestamp.Unix(),
```
jeschkies, line 1317: Is this returning milliseconds? I'll take a look in a bit

shantanualsi: i changed the PR to return seconds from `evalLiteral`. ptal.

pkg/logql/engine.go
```diff
@@ -448,7 +448,7 @@ func (q *query) evalLiteral(_ context.Context, expr *syntax.LiteralExpr) (promql
 		return nil, err
 	}
 	s := promql.Scalar{
-		T: q.params.Start().UnixNano() / int64(time.Millisecond),
```
jeschkies, line 451: Checkout line 468. I think that one is wron. And then there's 474 that just uses `UnixMilli()` :thinking:

pkg/querier/queryrange/codec.go
jeschkies: So, `v.Timestamp` is using the Prometheus `model.Time` which is in milliseconds. So this should be `TimestampMs: int64(v.Timestamp)`.

PR #11115, opened 2023-11-01
Add first step for creating a single bloom from series

poyzannur:
This is the first step to create a bloom from a series.

Fixes #<issue number>

It does not handle appending multiple blooms into blocks. 
It does not handle creating and uploading of meta files. 
It assumes a single periodConfig. 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +145,273 @@ func (c *Compactor) stopping(_ error) error {
...
+		},
+		Data: blockFile,
+	}
+
+	return blocks, nil
+}
+
+func listSeries(ctx context.Context, objectClient client.ObjectClient) ([]Series, error) {
+	// Returns all the TSDB files, including subdirectories
+	prefix := "index/"
```
owen-d, line 247: Heads up, this prefix is [configurable](https://github.com/grafana/loki/blob/main/pkg/storage/config/schema_config.go#L482) and should be factored into inputs or handled elsewhere.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +145,273 @@ func (c *Compactor) stopping(_ error) error {
...
-	bloomPath := filepath.Join(dst, "bloom")
+	var result []Series
+
+	for _, index := range indices {
+		s := strings.Split(index.Key, "/")
+
+		if len(s) > 3 {
+			tableName := s[1]
+
+			if !strings.HasPrefix(tableName, "loki_") || strings.Contains(tableName, "backup") {
```
owen-d, line 262: These should also also be not defined as constants here (they're both configurable). The file name prefix is set [here](https://github.com/grafana/loki/blob/main/pkg/storage/config/schema_config.go#L512). The `backup` is also not something we can reliably ignore here (it's feasible to name the table prefix "backup"). Instead, try using a `PeriodConfig` as an argument to this function.

It might also be worthwhile to build a file discovery method into the index shipper's API since it already does a lot of this logic internally.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +145,273 @@ func (c *Compactor) stopping(_ error) error {
...
+			userID := s[2]
+			_, err := strconv.Atoi(userID)
+			if err != nil {
+				continue
+			}
 
-	blockRef := bloomshipper.BlockRef{
-		IndexPath: indexPath,
-		BlockPath: bloomPath,
+			result = append(result, Series{tableName: tableName, tenant: userID, indexPath: index.Key})
```
owen-d, line 272: nit: A series is a single set of label names+values, but this is an entire index file consisting of many series.

pkg/storage/bloom/v1/block.go
```diff
@@ -176,6 +176,6 @@ outer:
 		// chunk passed all searches, add to the list of chunks to download
 		mustCheck = append(mustCheck, chk)
 
-	}
+	} // again is this desired behaviour. only returns must check, not in bloom?
```
owen-d, line 179: Yes. A positive test from a bloom filter doesn't tell for certain if a key was indexed, but a negative test is certain a key wasn't indexed. 

We send a list of chunks to the blooms. The ones that don't pass the queries are ignored (they're guaranteed to not have the data we want), and we return the list of chunks we'll need to scan.

pkg/bloomcompactor/bloomcompactor.go
poyzannur: I wasn't super clear what this `indexPath` actually corresponds to [here](https://github.com/grafana/loki/blob/main/pkg/storage/stores/shipper/bloomshipper/client.go#L40-L44)
i'll clarify with Vlad.

pkg/bloomcompactor/bloomcompactor.go
poyzannur: This `listSeries` method was my temporary hack to get tableNames and tenants,  i'm aware that it's partial and fragile.
 
[Salva's ring work here](https://github.com/grafana/loki/compare/salvacorts/bloomcompactor-sharding) provides them nicely, it uses `indexShipper/storage`'s `ListTables` [method](https://github.com/grafana/loki/blob/main/pkg/storage/stores/shipper/indexshipper/storage/client.go#L65-L77)

 `Job` struct in Salva's branch will replace the `Series` struct in this PR.

pkg/bloomcompactor/bloomcompactor.go
poyzannur: same reasoning as above

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +145,273 @@ func (c *Compactor) stopping(_ error) error {
...
+	blockFile, err := os.Open(filepath.Join(localDst, "bloom"))
 	if err != nil {
-		return err
+		level.Info(util_log.Logger).Log("reading bloomBlock", err)
 	}
-	// BuildFrom closes itself
-	err = builder.BuildFrom(v1.NewSliceIter[v1.SeriesWithBloom](seriesList))
+
+	// read the checksum
+	blockFile.Seek(-4, 2)
```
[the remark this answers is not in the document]
poyzannur: Sure, I'll address in a follow up PR.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
-// TODO get series from objectClient (TSDB) instead of params
-func NoopGetSeries() *v1.Series { return nil }
+	return chunkRefs
+}
 
-// TODO Then get chunk data from series
-func NoopGetChunks() []byte { return nil }
+// TODO Revisit this step once v1/bloom lib updated to combine blooms in the same series
+func buildBloomBlock(bloomForChks v1.SeriesWithBloom, series Series) (bloomshipper.Block, error) {
+	localDst := createLocalFileName(series)
```
vlad-diachenko, line 204: I believe we need to create a file in workingDir that can be set in config

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
-			// TODO point to the data to be read
-			Data: nil,
+func CompactNewChunks(ctx context.Context, series Series, bloomShipperClient bloomshipper.Client) (err error) {
+	// Create a bloom for this series
+	bloomForChks := v1.SeriesWithBloom{
+		Series: &v1.Series{
+			Fingerprint: series.fingerPrint,
+		},
+		Bloom: &v1.Bloom{
+			ScalableBloomFilter: *filter.NewDefaultScalableBloomFilter(0.01),
```
vlad-diachenko, line 311: nit: let's move fpRate(`0.01`) to the config with visibility `advanced` ... just in case we decide to tune it in the future to avoid the necessity to rebuild it

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
+		return bloomshipper.Block{}, err
+	}
 
-	// block and seriesList
-	seriesList := []v1.SeriesWithBloom{
-		{
-			Series: series,
-			Bloom:  &bloom,
-		},
+	blockFile, err := os.Open(filepath.Join(localDst, "bloom"))
```
vlad-diachenko, line 219: nit: let's make constants public and use them from package `v1` instead of using string values "bloom"
```
const (
	bloomFileName  = "bloom"
	seriesFileName = "series"
)
```

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
+	if err != nil {
+		level.Info(util_log.Logger).Log("reading bloomBlock", err)
 	}
 
-	writer := v1.NewDirectoryBlockWriter(dst)
+	// read the checksum
+	if _, err := blockFile.Seek(-4, 2); err != nil {
```
vlad-diachenko, line 225: could you please move out magic numbers to the separate constants

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
+		return bloomshipper.Block{}, errors.Wrap(err, "seeking to bloom checksum")
+	}
+	checksum := make([]byte, 4)
```
vlad-diachenko, line 228: magic numbers ;)

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
-			BloomPageSize:  10 << 10,
-		}, writer)
-	if err != nil {
-		return err
+	// Reset back to beginning
+	if _, err := blockFile.Seek(0, 0); err != nil {
+		return bloomshipper.Block{}, errors.Wrap(err, "seeking to back to beginning of the file")
+	}
+
+	objectStoreDst := createObjStorageFileName(series, fmt.Sprint(binary.BigEndian.Uint32(checksum)))
```
vlad-diachenko, line 238: BloomClient PutBlocks will populate the path when the file is uploaded to the storage.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
+			Ref: bloomshipper.Ref{
+				TenantID:       series.tenant,
+				TableName:      series.tableName,
+				MinFingerprint: uint64(series.fingerPrint), //TODO will change once we compact multiple blooms into a block
+				MaxFingerprint: uint64(series.fingerPrint),
+				StartTimestamp: series.from.Unix(),
+				EndTimestamp:   series.through.Unix(),
+				Checksum:       binary.BigEndian.Uint32(checksum),
+			},
+			IndexPath: series.indexPath, // For reviewer, this is the same filepath with the original TSDBindex for this series?
```
vlad-diachenko, line 251: yes, it's the path to the index file from which the block is built, to understand that we need to rebuild the block file if the index file is changed.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
+	return result, nil
+}
 
-	blocks := []bloomshipper.Block{
-		{
-			BlockRef: blockRef,
+func createLocalFileName(series Series) string {
+	return fmt.Sprintf("bloomBlock-%s-%s-%s-%s-%s-%s", series.tableName, series.tenant, series.fingerPrint, series.fingerPrint, series.from, series.through)
+}
+func createObjStorageFileName(series Series, checksum string) string {
```
vlad-diachenko, line 296: we do not need this logic. BloomClient will populate it.

vlad-diachenko (commented):
Looks good, I left some comments that we could apply I believe.

owen-d (approved):

pkg/bloomcompactor/bloomcompactor.go
poyzannur: I'll move to a constant for now to make it more clear.
But i'm not sure what you mean by "to avoid the necessity to rebuild it" because changing falsePositive rate will require bloomFilter to be re-created. We can't change it in runtime. I've checked a while ago with @paul1r that it's a decision to configure it, i don't think we need it in config.

pkg/storage/bloom/v1/bloom_tokenizer.go
```diff
@@ -89,6 +91,11 @@ func (bt *BloomTokenizer) PopulateSeriesWithBloom(seriesWithBloom *SeriesWithBlo
 
 		defer itr.Close()
 
+		if err != nil {
+			return
+		}
+		defer itr.Close()
```
salvacorts, line 97: I think this already done few lines above

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
-		// 1. for each period at hand, get TSDB table indexes for given fp range
-		// 2. Check blocks for given uniqueIndexPaths and TSDBindexes
-		//	if bloomBlock refs are a superset (covers TSDBIndexes plus more outside of range)
-		//	create a new meta.json file, tombstone unused index/block paths.
-
-		// else if: there are TSDBindexes that are not covered in bloomBlocks (a subset)
-		// then call compactNewChunks on them and create a new meta.json
-
-		// else: all good, no compaction
+			_ = idx.(*tsdb.TSDBFile).Index.(*tsdb.TSDBIndex).ForSeries(
```
salvacorts, line 362: We should make these castings safe.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
+				ctx,
+				nil,              // TODO no shards for now, revisit with ring work
```
salvacorts, line 364: I think shard arg refers to the query-level sharding. IIUC we want to process all shards here so I think this TODO can be removed.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
+				0, math.MaxInt64, // Replace with MaxLookBackPeriod
```
salvacorts, line 365: It looks like we are removing that setting in this PR, shouldn't we use it in here or will we use something different?

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
+
+				// Get chunks for a series label and a fp
+				func(ls labels.Labels, fp model.Fingerprint, chksMetas []tsdbindex.ChunkMeta) {
+
+					// TODO call bloomShipperClient.GetMetas to get existing meta.json
+					var metas []bloomshipper.Meta
+
+					if len(metas) == 0 {
+						// Get chunks data from list of chunkRefs
```
salvacorts, line 374: nit: I'd extract the login inside here into a new function provided the else statement will probably be big as well (I'd put it on a separate function as well)

pkg/bloomcompactor/bloomcompactor.go
poyzannur: I have a follow up PR to return checksum from `BuildFrom` method. I'll remove this code of reading checksum.

pkg/bloomcompactor/bloomcompactor.go
poyzannur: same as above re reading checksum will be removed.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -106,141 +173,263 @@ func (c *Compactor) stopping(_ error) error {
...
 
-	builder, err := v1.NewBlockBuilder(
-		v1.BlockOptions{
-			SeriesPageSize: 100,
-			BloomPageSize:  10 << 10,
-		}, writer)
-	if err != nil {
-		return err
+	// Reset back to beginning
+	if _, err := blockFile.Seek(0, 0); err != nil {
```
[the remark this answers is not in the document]
poyzannur: same as above re reading checksum will be removed.

pkg/bloomcompactor/bloomcompactor.go
poyzannur: added a todo 👍

pkg/bloomcompactor/bloomcompactor.go
poyzannur: instead of getting it from config, I'll use the `RejectOldSamplesMaxAge` limit that the distributors are using.

salvacorts (approved):

pkg/bloomcompactor/bloomcompactor.go
vlad-diachenko: it's not critical, so np if we do not have it in the config.

>to avoid the necessity to rebuild it

`it` here is loki image ) sorry for the confusion 

I meant to add this field to the loki config (not runtime) so we could tune it and restart the compactor and it would use the new fp rate...

vlad-diachenko (approved):
lgtm

PR #11124, opened 2023-11-02
[Style] Data source as two words.

JStickler:
Updating "datasource" to be "data source" per recent decision by Style Council.


cstyan (approved):
Makes sense to me :+1: 

I'd always assumed Grafana referred to them as `datasource` one word, but I just checked and they use `data source`.

JStickler:
It's currently not very standardized.  The docs team is trying to get make things more consistent.

PR #11125, opened 2023-11-02
[Style] Update timeseries to time series.

JStickler:
Small update to conform to recent Style decision re: spelling of `time series`.
https://grafana.com/docs/writers-toolkit/write/style-guide/word-list/#t


cstyan (approved):
same as before, I'd always assumed `timeseries` was correct in this context but prometheus uses `time series` as well

PR #11132, opened 2023-11-03
feat(storage): GCS backend using thanos.io/objstore

JoaoBraveCoding:

This PR adds support to use the thanos.io/objstore backend for the GCS provider for all components ~~including the Ruler~~

Fixes #<issue number>

- We tried to the best of our efforts to preserve feature parity so the new backend still uses the ObjectClient interface
- Thanos.io/objstore doesn't support all the CLI flags Loki supports nowadays (see snippet below)
- Hedging support was added in https://github.com/thanos-io/objstore/pull/86 and https://github.com/grafana/loki/pull/11132/commits/32a9e4c6945e2bc1f28c9deaaa529114bd77fb7f 
- This PR was inspired by the way Mimir uses thanos.io/objstore intending to make this part of the codebase as similar as possible
- Support for rulers will be added in a follow-up pr https://github.com/grafana/loki/pull/11713
- Congestion control support in a follow-up. pending on https://github.com/thanos-io/objstore/pull/147
- Named store support will be added separately

CLI Table
```
gcs.bucketname
gcs.service-account
gcs.chunk-buffer-size

NOT PRESENT AT THE MOMENT

gcs.enable-opencensus (Would need objstore patch)
gcs.enable-retries (we have an upstream patch)
gcs.request-timeout (will be added as a common option for all providers)
gcs.enable-http2 (not required for newer version of cristalhq/hedgedhttp)

```

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


chaudum:
> This PR was inspired by the way Mimir uses thanos.io/objstore with the goal of making this part of the codebase as similar as possible

Thank you! :heart:

kavirajk:
@JoaoBraveCoding thanks for the PR :)

Update: 
I'm testing GCS changes on two things internally.
1. On our internal cluster to make sure nothing breaks
2. Our Cloud logs archiver project uses Loki's storage layer to copy chunks between buckets. I'm making sure this PR doesn't break anything there (particularly authentication side of things).

I will keep it posted here. Once those testing is done, we can approve and merge this PR

kavirajk:
@JoaoBraveCoding I was testing this PR and found it is breaking couple of our internal services.
```
not enough arguments in call to storage.NewObjectClient
	have (string, "github.com/grafana/loki/v3/pkg/storage".Config, "github.com/grafana/loki/v3/pkg/storage".ClientMetrics)
	want (string, string, "github.com/grafana/loki/v3/pkg/storage".Config, "github.com/grafana/loki/v3/pkg/storage".ClientMetrics, "github.com/prometheus/client_golang/prometheus".Registerer)
```
Basically, looks like we cannot afford to change the signature of `storage.NewObjectClient` atomically in single PR. These two internal services consume those heavily and I think we need to take multi step approach.
1. Keep `storage.NewObjectClient  as is
2. expose something like storage.NewObjectClientV2 (or similar) with thanos object store support
3. I will migrate our internal services to V2 separately.
wdyt?

kavirajk:
I'm going to merge this PR. 

There is one edge case when we enable `thanos_objstore: true` in some components (compactor, ruler, idx-gw and ingester) thanos metrics are getting registered more than once, causing duplicate panic. But this never happens with `thanos_objstore: false` which is default. **Basically it's no-op if not enabled, tested in on some internal dev cells**

The rationale for that is mainly if loki config have multiple period_configs, we create one object client per period_config and [each will try to register the metrics collector in thanos currently](https://github.com/thanos-io/objstore/blob/main/objstore.go#L409-L452). But in Loki currently [we create those metrics once](https://github.com/grafana/loki/blob/5f2a02f14222dab891b7851e8f48052d6c9b594a/pkg/storage/chunk/client/gcp/instrumentation.go#L30) and [make it part of transport](https://github.com/grafana/loki/blob/5f2a02f14222dab891b7851e8f48052d6c9b594a/pkg/storage/chunk/client/gcp/instrumentation.go#L54-L55) so that all the object storage calls are tracked in single set of metrics.

I have [raised an upstream PR](https://github.com/thanos-io/objstore/pull/130) to decouple metrics creation and instrumenting the bucket. 

With this, we can create thanos metrics once and use it multiple instances of object clients. I have a changes ready to fix it in compactor, I will raise as separate PR (we can do similar things on other components as well)

All these can be done iteratively, meaning we can merge this PR without keeping it open.

Another important rationale is, this PR often get's outdated and needing some manual work. With RF=1 feature, people also touch object client API's so keeping this PR open going to make only worse with more frequent rebase & conflicts.

Looping some people for visibility @JoaoBraveCoding @slim-bean

pkg/storage/bucket/client.go
```diff
@@ -117,26 +179,35 @@ func NewClient(ctx context.Context, cfg Config, name string, logger log.Logger,
...
-	return objstore.WrapWithMetrics(
-		bucketClient,
-		prometheus.WrapRegistererWith(prometheus.Labels{"component": name}, reg),
-		"")
+	if metrics.BucketMetrics == nil {
+		reg := metrics.Registerer
+		reg = prometheus.WrapRegistererWithPrefix("loki_", reg)
+		reg = prometheus.WrapRegistererWith(prometheus.Labels{"component": name}, reg)
+		// Save metrics to be assigned to other buckets created with the same component name
+		metrics.BucketMetrics = objstore.BucketMetrics(reg, "")
```
ashwanthgoli, line 209: can we define the bucket metrics as a global in this file and register it with the default registerer during pkg init?
we can revise this in follow-up prs if needed. wdyt?

pkg/loki/modules.go
```diff
@@ -1738,7 +1753,13 @@ func (t *Loki) initAnalytics() (services.Service, error) {
 		return nil, err
 	}
 
-	objectClient, err := storage.NewObjectClient(period.ObjectType, t.Cfg.StorageConfig, t.ClientMetrics)
+	var objectClient client.ObjectClient
+	if t.Cfg.StorageConfig.ThanosObjStore {
+		metrics := &client.Metrics{Registerer: prometheus.DefaultRegisterer}
+		objectClient, err = storage.NewObjectClientV2("analytics", period.ObjectType, t.Cfg.StorageConfig, metrics)
```
ashwanthgoli, line 1759: loki already has a thanos specific ruler storage section [`t.Cfg.RulerStorage`](https://github.com/grafana/loki/pull/11713/files#diff-dfe68c6d3018ac004616d53414bdfc82d0bbb04eea91fdff7f63e0369b22b022R1055) as the code was taken from cortex project. Should we remove ruler specific changes from this? I can revive https://github.com/grafana/loki/pull/11713 and rebase changes from this branch to it

DylanGuedes (commented):
I know that the change was introduced in a previous PR but anyway, now we're doing this everywhere:
```
		var objectClient client.ObjectClient
		var err error
		if t.Cfg.StorageConfig.UseThanosObjstore {
			objectClient, err = storage.NewObjectClientV2("compactor", periodConfig.ObjectType, t.Cfg.StorageConfig)
		} else {
			objectClient, err = storage.NewObjectClient(periodConfig.ObjectType, t.Cfg.StorageConfig, t.ClientMetrics)
		}
		if err != nil ...
```
WDYT of instead we moving the if conditional to the `storage.NewObjectClientV2` instead, and adding a new parameter to it like `legacy=true` that dictates if internally it will instantiate `V2` or `V1`?
the way we're doing here we're having to duplicate the condition everywhere (which has the risk of we missing something) and external libs will also have to add the same conditional.

JoaoBraveCoding (commented):
Just some small comments but overall lgmt 🙌  thank you for the work to get this merged!

pkg/storage/bucket/client.go
```diff
@@ -35,70 +33,112 @@ const (
...
 )
 
 var (
 	SupportedBackends = []string{S3, GCS, Azure, Swift, Filesystem}
 
-	ErrUnsupportedStorageBackend = errors.New("unsupported storage backend")
+	ErrUnsupportedStorageBackend        = errors.New("unsupported storage backend")
+	ErrInvalidCharactersInStoragePrefix = errors.New("storage prefix contains invalid characters, it may only contain digits and English alphabet letters")
+
+	// TODO: Remove this global if we can tag all clients with component name
```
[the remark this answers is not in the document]
ashwanthgoli: yeah, i think the comment can be more descriptive. i'll fix it :)

pkg/storage/chunk/client/gcp/gcs_thanos_object_client.go
```diff
@@ -0,0 +1,203 @@
...
+	reader, err := s.hedgedClient.Get(ctx, objectKey)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	size, err := objstore.TryToGetSize(reader)
+	if err != nil {
+		size = -1
+		level.Warn(s.logger).Log("msg", "failed to get size of object", "err", err)
+	}
```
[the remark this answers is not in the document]
ashwanthgoli: likely not for gcs, but it might also fail if the provider specific code fails to get the size for some reason. Seems very unlikely but I added this to not fail the Get request entirely if we cannot figure out the size. Size is only being used to reduce allocations, does not affect any functionality

docs/sources/shared/configuration.md
```diff
@@ -3685,6 +3725,20 @@ ruler_remote_write_sigv4_config:
...
+
+# S3 server-side encryption KMS Key ID. Ignored if the SSE type override is not
+# set.
+[s3_sse_kms_key_id: <string> | default = ""]
+
+# S3 server-side encryption KMS encryption context. If unset and the key ID
+# override is set, the encryption context will not be provided to S3. Ignored if
+# the SSE type override is not set.
+[s3_sse_kms_encryption_context: <string> | default = ""]
+
```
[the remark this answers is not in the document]
ashwanthgoli: updated in f47392878

pkg/storage/chunk/client/object_client.go
```diff
@@ -183,6 +183,11 @@ func (o *client) getChunk(ctx context.Context, decodeContext *chunk.DecodeContex
 	}
 	defer readCloser.Close()
 
+	// reset if GetObject fails to return a valid size
+	if size < 0 {
+		size = 0
+	}
```
DylanGuedes, line 189: hey would that be possible a bug on our side? If so, i think we should at least log something. If that's a known problem with the GCP API please ignore this comment.

DylanGuedes (approved):
lgtm just a quick question on https://github.com/grafana/loki/pull/11132#discussion_r1804604896

pkg/storage/chunk/client/object_client.go
ashwanthgoli: we log it in the called method [here](https://github.com/grafana/loki/pull/11132/files#diff-4ccb492979e70aa118915885936aa252cc68d2e3164a0dc681e511e8192b1432R102)

I added some context [here](https://github.com/grafana/loki/pull/11132#discussion_r1800962297) about the check, but this is unlikely to fail. I'll monitor for errors as i test this out

PR #11133, opened 2023-11-03
Tokenizer tests and TokenizeLine updates

paul1r:
The thrust of this PR is to ensure we have tests for each major function of the Bloom Tokenizer.  In addition, there was some cleanup, in that constants are used to set some common parameters.

Lastly, the TokenizeLine() call was updated to correctly tokenize a line when a "skip tokenizer" is utilized.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


poyzannur (approved):
LGTM with a minor question on the use of the two methods you added. 
Am i correct to assume `TokenizeLineWithChunkPrefix` should be used when writing blooms, whereas `TokenizeLine` should be used to tokenize search queries before querying the blooms?

paul1r:
> LGTM with a minor question on the use of the two methods you added. Am i correct to assume `TokenizeLineWithChunkPrefix` should be used when writing blooms, whereas `TokenizeLine` should be used to tokenize search queries before querying the blooms?

PopulateSeriesWithBloom would be used when writing blooms.  TokenizeLine is just to get a quick set of tokens, for the quick sniff test if we need to dig into a chunk or not.  The TokenizeLineWithChunkPrefix is to validate those tokens exist for a specific chunk.  Should be more obvious once we start wiring this together

owen-d (dismissed):

pkg/storage/bloom/v1/bloom_tokenizer.go
```diff
@@ -116,6 +118,54 @@ func (bt *BloomTokenizer) PopulateSeriesWithBloom(seriesWithBloom *SeriesWithBlo
...
+	}
+	return allTokens
+}
+
+// TokenizeLineWithChunkPrefix returns a slice of tokens for the given line, based on the current value of the tokenizer,
+// and prepends the chunk ID to the tokens
+// If the tokenizer has a skip value, then the line will be tokenized multiple times,
+// starting at the beginning of the line, with "skip" number of iterations, offset by one each time
+// Each offset is kept as a separate slice of tokens, and all are returned in a slice of slices
+func (bt *BloomTokenizer) TokenizeLineWithChunkPrefix(line string, chk logproto.ChunkRef) [][]Token {
```
owen-d, line 150: There's no need for two functions here -- you can just use something like

```go
func SearchesForTokenizerAndLine(t Tokenizer, line string) (res [][]Token) {
  for i := 0; i < t.Skip()+1; i++ {
    res = append(res, t.Tokens(line[i:])) // this needs to account for runes vs bytes, but you get the idea 
  }
  return
}

```

pkg/storage/bloom/v1/bloom_tokenizer.go
```diff
@@ -116,6 +118,54 @@ func (bt *BloomTokenizer) PopulateSeriesWithBloom(seriesWithBloom *SeriesWithBlo
...
+	allTokens := make([][]Token, 0, 10)
+
+	if len(line) >= bt.chunkIDTokenizer.GetMin() && len(line) >= bt.chunkIDTokenizer.GetSkip() {
```
owen-d, line 153: Two things:
* This actually needs to ensure the length is `>= min + skip`. 
* len(str) doesn't return the number of `runes`, but the number of bytes. We need to account for `runes` since that's how we index. See [this](https://go.dev/blog/strings) for more detail.

pkg/storage/bloom/v1/bloom_tokenizer.go
```diff
@@ -116,6 +119,30 @@ func (bt *BloomTokenizer) PopulateSeriesWithBloom(seriesWithBloom *SeriesWithBlo
 	} // for each chunk
 }
 
-func (bt *BloomTokenizer) TokenizeLine(line string) []Token {
-	return bt.lineTokenizer.Tokens(line)
+// SearchesForTokenizerAndLine is for taking a given search string (ex: on the read/query path) and returning
+// all the possible tokens, given a tokenizer.
+// This is a multi-dimensional slice where the first slice is the offset into the line, and the
+// second slice is the tokens for that offset.
```
owen-d, line 125: This is only true if all of the skip offsets return at least one token. Otherwise, the length of the result will be less than the number of skips+1.

pkg/storage/bloom/v1/bloom_tokenizer.go
```diff
@@ -116,6 +119,30 @@ func (bt *BloomTokenizer) PopulateSeriesWithBloom(seriesWithBloom *SeriesWithBlo
...
+// The offset is used if the Tokenizer has a skip value being utilized.
+func SearchesForTokenizerAndLine(t Tokenizer, line string) (res [][]Token) {
+	res = make([][]Token, 0, 10)
+	for i := range line { // iterate by runes
```
owen-d, line 129: This unnecessarily iterates all runes in the line, including offsets beyond `skip+1`.

paul1r: ack, added a break clause

pkg/storage/bloom/v1/bloom_tokenizer.go
paul1r: yep, I've updated the doc accordingly, good catch

owen-d (approved):

PR #11143, opened 2023-11-06
otel: add support for per tenant configuration for mapping otlp data to loki format

sandeepsukhani:
In OTEL, we pick select Resource Attributes to identify the streams and store all the other attributes as Structured Metadata, as explained here. The problem however is that the list of Resource Attributes that are picked as Stream labels are hardcoded, and there is no way to drop unwanted data. This PR adds support for configuring how data is mapped from OTEL to Loki format per tenant. It also adds support for dropping unwanted data. We decided to make the config look similar to Prometheus's relabling config to make it familiar.

Opening a draft PR to get some initial feedback. I will add documentation in a separate PR.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/loghttp/push/push.go
```diff
@@ -58,7 +58,11 @@ type TenantsRetention interface {
 	RetentionPeriodFor(userID string, lbs labels.Labels) time.Duration
 }
 
-type RequestParser func(userID string, r *http.Request, tenantsRetention TenantsRetention) (*logproto.PushRequest, *Stats, error)
+type Limits interface {
+	OTLPConfig(userID string) OTLPConfig
+}
```
slim-bean, line 63: Not sure why a new interface here is helpful? Can't we use the existing overrides?

sandeepsukhani: It makes it easier to write tests, and we can't reference overrides since that would cause circular dependency.

slim-bean:
This looks good to me @sandeepsukhani!

salvacorts (approved):
LGTM

PR #11148, opened 2023-11-06
Default build-image target arch to amd64

trevorwhitney:

The changes in https://github.com/grafana/loki/pull/11130 broke `make build-image` and `make build-image-push`. This defaults the target arch added in #11130 so those make targets still work.

### DIFF
--- a/loki-build-image/Dockerfile
+++ b/loki-build-image/Dockerfile
@@ -7,7 +7,7 @@
 
 # Install helm (https://helm.sh/) and helm-docs (https://github.com/norwoodj/helm-docs) for generating Helm Chart reference.
 FROM golang:1.21.3-bullseye as helm
-ARG TARGETARCH="amd64"
+ARG TARGETARCH
 ARG HELM_VER="v3.2.3"
 RUN curl -L "https://get.helm.sh/helm-${HELM_VER}-linux-$TARGETARCH.tar.gz" | tar zx && \
     install -t /usr/local/bin "linux-$TARGETARCH/helm"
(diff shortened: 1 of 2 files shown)


cstyan (approved):

loki-build-image/Dockerfile
```diff
@@ -7,7 +7,7 @@
 
 # Install helm (https://helm.sh/) and helm-docs (https://github.com/norwoodj/helm-docs) for generating Helm Chart reference.
 FROM golang:1.21.3-bullseye as helm
-ARG TARGETARCH
+ARG TARGETARCH="amd64"
```
jeschkies, line 10: This would be set by `DOCKER_BUILDKIT: 1` which I would favour.

jeschkies (changes requested):
Could you try with `DOCKER_BUILDKIT: 1` in the `Makefile`. Overall I'm not sure we should keep this target around. Alternatively we could migrate to all build targets like suggest [here](https://github.com/grafana/deployment_tools/blob/master/docs/infrastructure/arm.md#if-your-project-lives-elsewhere).

loki-build-image/Dockerfile
trevorwhitney: I'm confused on how this is supposed to work. The [Makefile alread sets DOCKER_BUILDKIT=1](https://github.com/grafana/loki/blob/6385b195739bd7d4e9706faddd0de663d8e5331a/Makefile#L104), but running `make build-image` fails. However, if I run `DOCKER_BUILDKIT=1 make build-image` it works as you have mentioned.

trevorwhitney:
@jeschkies I changed how this was done, is 3b8bfc028559fce0a0a533f56d844f6f03945d63 a better solution?

jeschkies (approved):

PR #11154, opened 2023-11-07
Salvacorts/bloomcompactor sharding

salvacorts:

Fixes #<issue number>


pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -90,22 +125,335 @@ func New(cfg Config,
...
+	}
+
+	c.metrics.compactionRunDiscoveredTenants.Add(float64(len(tenants)))
+	level.Info(c.logger).Log("msg", "discovered tenants from bucket", "users", len(tenants))
+	return c.compactUsers(ctx, sc, tableName, tenants)
+}
+
+// See: https://github.com/grafana/mimir/blob/34852137c332d4050e53128481f4f6417daee91e/pkg/compactor/compactor.go#L566-L689
+func (c *Compactor) compactUsers(ctx context.Context, sc storeClient, tableName string, tenants []string) error {
+	// When starting multiple compactor replicas nearly at the same time, running in a cluster with
```
owen-d, line 310: I don't think this really applies here. The tsdb compactor is currently run as a singleton, meaning it compacts everything from one process. This code was a safeguard to minimize duplicate work when multiple compactors were erroneously configured.

In the bloom compactor, we expect to run multiple compactors in a ring. Each replica should operate on different data, making this safeguard unnecessary.

There is one exception: when the ring topology changes and the compactor replicas have different views of ring membership when determining what data to compact.

In summary, this shouldn't help us much, but also doesn't really hurt. Just wanted to make sure you understood why this optimization isn't necessary for the bloom compactors compared to the tsdb compactor.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -90,22 +125,335 @@ func New(cfg Config,
...
+
+func (c *Compactor) compactTenant(ctx context.Context, sc storeClient, tableName string, tenant string) error {
+	level.Info(c.logger).Log("msg", "starting compaction of tenant", "tenant", tenant)
+
+	// Ensure the context has not been canceled (ie. compactor shutdown has been triggered).
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
+	// TODO: Use ForEachConcurrent?
```
owen-d, line 372: Probably unnecessary to multithread here. There are better ways to introduce a configurable worker parallelism elsewhere (likely more centralized for reuse across tenants as well as periods and tsdbs). Happy to leave that for a followup PR if/when it proves warranted

pkg/bloomcompactor/config.go
```diff
@@ -14,13 +15,29 @@ type Config struct {
...
+	MaxLookBackPeriod  time.Duration `yaml:"max_look_back_period"`
+	CompactionInterval time.Duration `yaml:"compaction_interval"`
+
+	// No need to add options to customize the retry backoff,
+	// given the defaults should be fine, but allow to override
+	// it in tests.
+	retryMinBackoff          time.Duration `yaml:"-"`
+	retryMaxBackoff          time.Duration `yaml:"-"`
+	CompactionRetries        int           `yaml:"compaction_retries" category:"advanced"`
+	TablesToCompact          int           `yaml:"tables_to_compact"`
```
owen-d, line 29: I don't think this should be a config. Say `tables_to_compact=4`, but we accept `7d` old data. That creates an easy misconfiguration that would never compact the older or newer tables, depending on which are preferred since each cycle would only take `4` from either end. Instead, I think we should have tenant configurations that specify the min and maximum durations we'll attempt compaction for (e.g. don't compact tables older than 7d or newer than 1d). This is similar to what I mentioned in a previous comment.

pkg/bloomcompactor/job.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"github.com/prometheus/common/model"
+
+	"github.com/grafana/loki/pkg/storage/stores/shipper/indexshipper/tsdb/index"
+)
+
+// Job holds a compaction job, which consists of a group of blocks that should be compacted together.
+// Not goroutine safe.
+// TODO: A job should probably contain series or chunks
+type Job struct {
```
owen-d, line 12: Nit: I expect we'll want a job to contain multiple series rather than one per fingerprint (fingerprint is roughly analagous to stream). Creating one per fingerprint will likely result in logging millions of streams when the happy path isn't satisfied. That's an inconvenience, but there may be places where it's a performance hit as well.

Instead, we could either
* Create one job for all fingerprints in a TSDB owned by the compactor replica
* Choose some batch-size and create a number of jobs that have that many series|chunks in each (more work but more predictable)

pkg/bloomcompactor/sharding.go
```diff
@@ -0,0 +1,73 @@
...
+	ring           *ring.Ring
+	ringLifeCycler *ring.BasicLifecycler
+	limits         Limits
+
+	// Buffers to avoid allocations in ring.Get
+	bufDescs           []ring.InstanceDesc
+	bufHosts, bufZones []string
+}
+
+func NewShuffleShardingStrategy(r *ring.Ring, ringLifecycler *ring.BasicLifecycler, limits Limits) *ShuffleShardingStrategy {
```
owen-d, line 28: I don't think we need to reimplement (copy/paste) this code. If we only need to extend functionality for `OwnsJob`, we can try struct-embedding. Is there something I'm missing?

owen-d (commented):
Very good start, left some thoughts.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -158,26 +170,288 @@ func New(cfg Config,
...
+	// multiple replicas at the same time. Shuffling users helps reduce the likelihood this will happen.
+	rand.Shuffle(len(tenants), func(i, j int) {
+		tenants[i], tenants[j] = tenants[j], tenants[i]
+	})
+
+	// Keep track of tenants owned by this shard, so that we can delete the local files for all other users.
+	errs := multierror.New()
+	ownedTenants := make(map[string]struct{}, len(tenants))
+	for _, tenant := range tenants {
+		// Ensure the context has not been canceled (ie. compactor shutdown has been triggered).
```
chaudum, line 317: You have a mix of `tenant` and `tenantID` for the tenant field in the logs.
To avoid that, you can create a logger with the field upfront:

```go
logger := log.With(c.logger, "tenant", tenant)
```

pkg/bloomcompactor/bloomcompactor_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"flag"
+	"fmt"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/grafana/dskit/flagext"
+	"github.com/grafana/dskit/kv/consul"
+	"github.com/grafana/dskit/ring"
+	ww "github.com/grafana/dskit/server"
```
chaudum, line 14: There is no other imported module named `server`

```suggestion
	"github.com/grafana/dskit/server"
```

pkg/bloomcompactor/sharding.go
chaudum: Right, we can use the sharding implementation from the gateway, move it to `github.com/grafana/loki/pkg/util/ring` and extend it.

pkg/bloomcompactor/sharding.go
```diff
@@ -0,0 +1,73 @@
...
+	s := ShuffleShardingStrategy{
+		ring:           r,
+		ringLifeCycler: ringLifecycler,
+		limits:         limits,
+	}
+	s.bufDescs, s.bufHosts, s.bufZones = ring.MakeBuffersForGet()
```
chaudum, line 34: Usage of these buffers is not threadsafe

pkg/bloomcompactor/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	compactionRunsStarted          prometheus.Counter
+	compactionRunsCompleted        prometheus.Counter
+	compactionRunsErred            prometheus.Counter
+	compactionRunDiscoveredTenants prometheus.Gauge
+	compactionRunSkippedTenants    prometheus.Gauge
+	compactionRunSucceededTenants  prometheus.Gauge
+	compactionRunFailedTenants     prometheus.Gauge
+	compactionRunSkippedJobs       prometheus.Gauge
+	compactionRunSucceededJobs     prometheus.Gauge
+	compactionRunFailedJobs        prometheus.Gauge
```
vlad-diachenko, line 22: all these metrics should be the counters.
we can use `rate` , `increase` . otherwise, we might miss relevant data if the value changes up and down between the scraping.

pkg/bloomcompactor/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	compactionRunInterval          prometheus.Gauge
+	compactorRunning               prometheus.Gauge
+}
+
+func newMetrics(r prometheus.Registerer) *metrics {
+	m := metrics{
+		compactionRunsStarted: promauto.With(r).NewCounter(prometheus.CounterOpts{
+			Namespace: metricsNamespace,
```
chaudum, line 30: namespace is `loki`, subsystem is `bloomcompactor`

```suggestion
			Namespace: metricsNamespace,
			Subsystem: metricsSubsystem,
```

pkg/validation/limits.go
```diff
@@ -297,6 +298,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 
 	f.IntVar(&l.IndexGatewayShardSize, "index-gateway.shard-size", 0, "The shard size defines how many index gateways should be used by a tenant for querying. If the global shard factor is 0, the global shard factor is set to the deprecated -replication-factor for backwards compatibility reasons.")
 	f.IntVar(&l.BloomGatewayShardSize, "bloom-gateway.shard-size", 1, "The shard size defines how many bloom gateways should be used by a tenant for querying.")
+	f.IntVar(&l.BloomCompactorShardSize, "bloom-compactor.shard-size", 0, "The shard size defines how many bloom compactors should be used by a tenant when computing blooms.")
```
vlad-diachenko, line 301: it's a bit confusing that shard-size for Bloom Gateway is set to 1 by default where is for the compactor it's set to 0.
Logically it should work the same way, but the users might be confused a bit...  
```suggestion
	f.IntVar(&l.BloomCompactorShardSize, "bloom-compactor.shard-size", 0, "The shard size defines how many bloom compactors should be used by a tenant when computing blooms. If it's set to 0, shuffle sharding is disabled.")
```

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -158,26 +168,288 @@ func New(cfg Config,
...
+					c.metrics.compactionRunSkippedJobs.Inc()
+					level.Debug(c.logger).Log("msg", "skipping job because it is not owned by this shard", "job", job)
+					return
+				}
+
+				if err := c.runCompact(ctx, c.bloomShipperClient, sc, job); err != nil {
+					c.metrics.compactionRunFailedJobs.Inc()
+					errs.Add(errors.Wrap(err, "runBloomCompact"))
+					return
+				}
```
chaudum, line 399: Probably a personal preference, but I would separate job collection and job execution into separate functions.

pkg/bloomcompactor/metrics.go
salvacorts: Done thank you!

pkg/bloomcompactor/sharding.go
salvacorts: Indeed. Thank you! If we end up making many allocations here, we can use a buffer pool.

pkg/bloomcompactor/bloomcompactor_test.go
salvacorts: Thank you!

pkg/bloomcompactor/bloomcompactor.go
salvacorts: Awesome, thanks!

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -90,22 +125,335 @@ func New(cfg Config,
...
+		err := c.compactTable(ctx, tableName)
+		if err != nil {
+			errs.Add(err)
+			return nil
+		}
+		level.Info(util_log.Logger).Log("msg", "finished compacting table", "table-name", tableName)
+		return nil
+	}); err != nil {
+		errs.Add(err)
+	}
```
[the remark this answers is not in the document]
salvacorts: Done

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -90,22 +125,335 @@ func New(cfg Config,
...
+			}
+			c.metrics.compactionRunsCompleted.Inc()
+		case <-ctx.Done():
+			return nil
+		}
+	}
 }
 
 func (c *Compactor) stopping(_ error) error {
+	c.metrics.compactorRunning.Set(0)
```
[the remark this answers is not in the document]
salvacorts: We no longer need the `bloomStore` after the last merge with main.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -90,22 +125,335 @@ func New(cfg Config,
...
+		tables = tables[:c.cfg.TablesToCompact]
+	}
+
+	// Reset discovered tenants metric since we will increase it in compactTable.
+	c.metrics.compactionRunDiscoveredTenants.Set(0)
+
+	parallelism := c.cfg.MaxCompactionParallelism
+	if parallelism == 0 {
+		parallelism = len(tables)
+	}
```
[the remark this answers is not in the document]
salvacorts: The parallelism is configurable and defaults to 1 (with the latest commit). Therefore, this will only happen if we set `MaxCompactionParallelism` to 0 (unlimited).

pkg/bloomcompactor/config.go
```diff
@@ -14,13 +15,29 @@ type Config struct {
...
+	WorkingDirectory   string        `yaml:"working_directory"`
+	MaxLookBackPeriod  time.Duration `yaml:"max_look_back_period"`
+	CompactionInterval time.Duration `yaml:"compaction_interval"`
+
+	// No need to add options to customize the retry backoff,
+	// given the defaults should be fine, but allow to override
+	// it in tests.
+	retryMinBackoff          time.Duration `yaml:"-"`
+	retryMaxBackoff          time.Duration `yaml:"-"`
+	CompactionRetries        int           `yaml:"compaction_retries" category:"advanced"`
```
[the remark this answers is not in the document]
salvacorts: Done, thank you!

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -90,22 +125,335 @@ func New(cfg Config,
...
+	ticker := time.NewTicker(util.DurationWithJitter(c.cfg.CompactionInterval, 0.05))
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			c.metrics.compactionRunsStarted.Inc()
+			if err := c.runCompaction(ctx); err != nil {
+				c.metrics.compactionRunsErred.Inc()
+				level.Error(util_log.Logger).Log("msg", "failed to run compaction", "err", err)
```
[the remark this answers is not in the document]
salvacorts: Done, thank you! Note that there are other functions using `util_log.Logger` such as `CompactNewChunks`. I'm not changing those yet as we may want to move that logic elsewhere.

pkg/validation/limits.go
salvacorts: Agree. My original idea was to, by default, allow all tenants to use all compactors. Changed the default to 1 as well and added your suggestion as well.

pkg/bloomcompactor/metrics.go
salvacorts: Done, thanks for the tip!

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -90,22 +125,335 @@ func New(cfg Config,
...
-func (c *Compactor) starting(_ context.Context) error {
-	return nil
+func (c *Compactor) starting(_ context.Context) (err error) {
+	c.metrics.compactorRunning.Set(1)
+	return err
+}
+
+func (c *Compactor) running(ctx context.Context) error {
+	// Run an initial compaction before starting the interval.
+	if err := c.runCompaction(ctx); err != nil {
```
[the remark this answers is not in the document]
salvacorts: Done

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -90,22 +125,335 @@ func New(cfg Config,
...
+		level.Error(util_log.Logger).Log("msg", "failed to run compaction", "err", err)
```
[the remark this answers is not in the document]
salvacorts: Done

pkg/bloomcompactor/metrics.go
poyzannur: I'd use counters for definitives ones like err, success, skipped etc once the job is completed
but I'd argue gauge is appropriate only to record ongoing jobs (CompactionRunning) 
I can see how many compactions are running at the same time, but the gauge will drop once it's completed and record a success, fail etc counter.

pkg/bloomcompactor/sharding.go
salvacorts: Done, I created `pkg/util/ring/sharding.go` with strategies to shard by tenant and by fingerprint. Refactored the sharding strategies from bloom-gw and bloom-compactor to reuse the new ones from the ring utils lib.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -343,98 +579,89 @@ func CompactNewChunks(ctx context.Context, series Series, bt *v1.BloomTokenizer,
...
-			return nil
-		})
+		series := Series{
+			tableName:   job.tableName,
+			tenant:      job.Tenant(),
+			labels:      job.Labels(),
+			fingerPrint: job.Fingerprint(),
+			chunks:      chks,
+			from:        minFrom,
+			through:     maxThrough,
```
poyzannur, line 618: suggestion: we can get `from` and `through` from chunkMetas. 
and you don't necessarily have to have chunks here in this struct. This way you can remove this and use Job instead. Pass the chunks to `CompactNewChunks` separately. 
`CompactNewChunks` will not always compact all chunks returned. Once completed, we will have another logic before calling `CompactNewChunks` to filter only the chunks that are not been compacted into blooms before.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -90,22 +125,335 @@ func New(cfg Config,
...
 
 func (c *Compactor) stopping(_ error) error {
+	c.metrics.compactorRunning.Set(0)
 	return nil
 }
 
+func (c *Compactor) runCompaction(ctx context.Context) error {
+	var (
+		tables []string
+		// it possible for two periods to use the same storage bucket and path prefix (different indexType or schema version)
```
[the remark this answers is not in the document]
salvacorts: Agree. Removed.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -90,22 +125,335 @@ func New(cfg Config,
...
+			tables = append(tables, table)
+			seen[table] = struct{}{}
+		}
+	}
+
+	// process most recent tables first
+	sortTablesByRange(tables)
+
+	// apply passed in compaction limits
+	if c.cfg.SkipLatestNTables <= len(tables) {
```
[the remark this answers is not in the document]
salvacorts: Done, replaced `SkipLatestNTables` by `MaxTableAge` to filter out tables older than `now - MaxTableAge`

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -90,22 +125,335 @@ func New(cfg Config,
...
+		tables = tables[c.cfg.SkipLatestNTables:]
+	}
+	if c.cfg.TablesToCompact > 0 && c.cfg.TablesToCompact < len(tables) {
```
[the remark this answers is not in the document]
salvacorts: Agree. Done.

pkg/bloomcompactor/bloomcompactor.go
salvacorts: That makes sense, thank you! I removed that shuffling.

pkg/bloomcompactor/bloomcompactor.go
salvacorts: Nice, I removed the TODO. Note that we are already parallelizing at the table and tenant level.

pkg/bloomcompactor/job.go
salvacorts: That was my original understanding/idea as well. But looking at https://github.com/grafana/loki/pull/11115 looked like we would be listing and processing one series at a time. So I though it was actually easier to filter by the series fingerprint right away.

> - Create one job for all fingerprints in a TSDB owned by the compactor replica
> - Choose some batch-size and create a number of jobs that have that many series|chunks in each (more work but more predictable)

The "problem" I see with this is that we'd need to iterate throughout the whole index to build those batches anyway. And then, iterate again skipping those unowned. So I don't quite see the benefit of creating those batches other than not checking the ring as often (which shouldn't be a problem since`ShuffleShard` ops are cached at the ring level)

> will likely result in logging millions of streams

For each series, the first thing we do is checking if we use the series' FP and skip it if we don't. By logging here I guess you mean the log line _`"skipping job because it is not owned by this shard"`_. I agree this will be too verbose. Since we have a metric to track that (`loki_bloomcompactor_jobs_skipped`), I think we can just remove this log line. Maybe we can even do the same for the tenant?

pkg/bloomcompactor/bloomcompactor.go
salvacorts: Agree! Let's do it on a followup PR once we add more logic to the actual compaction of chunks. Maybe moving that logic to `chunkcompactor.go`

pkg/bloomcompactor/bloomcompactor.go
salvacorts: Done, thank you!

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -343,98 +539,72 @@ func CompactNewChunks(ctx context.Context, series Series, bt *v1.BloomTokenizer,
...
-				})
-			return nil
-		})
-		if err != nil {
-			return errors.Wrap(err, "getting each series")
+// filterAndSortTablesByRange returns most-recent first sorted list of tables that are within the max age.
+func filterAndSortTablesByRange(tables []string, maxAge time.Duration) []string {
+	tableRanges := make(map[string]model.Interval, len(tables))
+	tablesToSort := make([]string, 0, len(tables))
+	maxAgeTime := model.Now().Add(-maxAge)
```
poyzannur, line 582: How can we ensure we exclude un-compacted index tables from potentially last 15 mins? we agreed that it will be tricky to compact most recent uncompacted tables.

pkg/bloomcompactor/bloomcompactor.go
owen-d: Parallelizing at the table level is probably ok for now. In the future, I'd rather have a worker pool for jobs (work is effectively a `(user, table, tsdb_shard_range)` tuple).

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -158,33 +165,256 @@ func New(cfg Config,
...
+
+				ownsJob, err := c.sharding.OwnsJob(job)
+				if err != nil {
+					c.metrics.compactionRunSkippedJobs.Inc()
+					level.Error(c.logger).Log("msg", "failed to check if compactor owns job", "job", job, "err", err)
+					errs.Add(err)
+					return
+				}
+				if !ownsJob {
+					c.metrics.compactionRunSkippedJobs.Inc()
```
poyzannur, line 357: I can see this metric name `compactionRunSkippedJobs` can get confusing in the future because there will be whole compactions skipped because there is no change in indexes and chunks that are already compacted. 
why do we want to count jobs not owned by the shard? we must assume it will be processed by another shard.

 I can see it as a back up to see if we are missing a subgroup of jobs all together, maybe we can use a metric like jobNotOwned, still it doesn't help us find which jobs.

pkg/bloomcompactor/bloomcompactor.go
salvacorts: Good call. I think we'll need to add a new limit `MinTableAge`? and look `End` time of the table's interval.

pkg/bloomcompactor/bloomcompactor.go
salvacorts: > because there will be whole compactions skipped because there is no change in indexes

What about renaming the metric to `unowned_jobs` instead?

> why do we want to count jobs not owned by the shard? we must assume it will be processed by another shard.

I think this will be useful for debugging proposes. We can look at the total number started compaction jobs and the number of finished jobs. ideally the number of started should be the same as finished. If they differ we can use the unowned metric to see if there are jobs not being owned by any compactor (if the sum of unowned jobs. more than jobs started).

Having said that, we'd need to add the `started` metric...

pkg/bloomcompactor/bloomcompactor.go
salvacorts: Added a TODO for a followup PR.

pkg/bloomcompactor/bloomcompactor.go
poyzannur: yes i agree that it's good for debugging purposes. `unowned_jobs` sounds the most appropriate.

pkg/bloomcompactor/bloomcompactor.go
poyzannur: Is there any other way of determining an un-compacted indexes I wonder? 
I've seen some tar files in the index bucket in gcp, something like `timestamp-ingester-12.tsdb.gz`, i assume these are un-compacted indexes. I'll check the code tomorrow to confirm. 
I know the underlying object client doesn't filter them out, returns them all. But maybe something else in the middle does.

pkg/bloomcompactor/job.go
salvacorts: Agreed with Owen to continue with 1 series - 1 job for now. We can easily reevaluate this in future iterations.

pkg/bloomcompactor/bloomcompactor.go
salvacorts: > Is there any other way of determining an un-compacted indexes I wonder?

I'm not sure. Let me know the output of your investigation. We can modify this in a followup PR.

Added MinTableAge defaulting to 1h for now.

pkg/bloomcompactor/bloomcompactor.go
salvacorts: Done :)

vlad-diachenko (approved):
LGTM ;)

pkg/bloomcompactor/job.go
```diff
@@ -0,0 +1,97 @@
...
+			minFrom = from
+		}
+		if maxThrough < through {
+			maxThrough = through
+		}
+	}
+
+	j.from = &minFrom
+	j.through = &maxThrough
+}
```
chaudum, line 97: This will yield incorrect result when `len(j.chunks)` is `0`, because `minFrom` and `maxThrough` are never updated.

poyzannur (approved):
lgtm, thanks for working this out.

pkg/bloomcompactor/metrics.go
```diff
@@ -0,0 +1,105 @@
...
+
+const (
+	metricsNamespace = "loki"
+	metricsSubsystem = "bloomcompactor"
+)
+
+type metrics struct {
+	compactionRunsStarted          prometheus.Counter
+	compactionRunsCompleted        prometheus.Counter
+	compactionRunsErred            prometheus.Counter
```
chaudum, line 16:
```suggestion
	compactionRunsFailed            prometheus.Counter
```

pkg/bloomcompactor/job.go
poyzannur: great catch, I can also address that in my next PR.

PR #11158, opened 2023-11-07
operator: Add warning for old schema configuration

btaani:
This PR adds a new warning condition to the LokiStack in case the object storage schema version is older than V13

Fixes [LOG-4548](https://issues.redhat.com/browse/LOG-4548)

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -354,8 +355,14 @@ func CreateOrUpdateLokiStack(
...
-		ll.Error(err, "failed to set storage schema status")
-		return err
+		if errors.Is(err, status.WarningError) {
+			if er := status.SetWarningCondition(ctx, k, req, status.StorageSchemaNeedsUpgrade, lokiv1.ReasonObjectStorageSchemaShouldBeUpgraded); er != nil {
+				return kverrors.Wrap(er, "error setting warning condition")
+			}
+		} else {
+			ll.Error(err, "failed to set storage schema status")
+			return err
+		}
```
periklis, line 365: I would simplify this moving SetWarningCondition call direct into `SetStorageSchemaStatus`.

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -1078,6 +1086,12 @@ type LokiStackStorageStatus struct {
 	// +optional
 	// +kubebuilder:validation:Optional
 	Schemas []ObjectStorageSchema `json:"schemas,omitempty"`
+	// SchemaStatus conveys any warnings/messages that apply
+	// to the currently applied schemas
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	SchemaStatus string `json:"schemaStatus,omitempty"`
```
periklis, line 1094: Having a custom warning condition adds more visibility for the user that in general some action needs to be taken (e.g. conditions are separately visible in the OpenShift Console). However I would keep the `SchemaStatus` in a slightly different form namely integrating this field in the `ObjectStorageSchema` struct, so that we can indicate which schemas can be removed.

periklis (changes requested):
In general this PR goes into the right direction. A couple of suggestions for better UX:
- Keep the Warning condition that the user should upgrade to V13 if not done.
- Add another Warning condition that indicates that some schemas are off the retention limit.
- Move `SchemaStatus` into the `ObjectStorageSchema` struct to indicate which schemas are off the the retention limit.

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -563,6 +563,13 @@ type ObjectStorageSchema struct {
 	// +required
 	// +kubebuilder:validation:Required
 	EffectiveDate StorageSchemaEffectiveDate `json:"effectiveDate"`
+
+	// SchemaStatus conveys any warnings/messages that apply
+	// to the currently applied schema
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	SchemaStatus string `json:"schemaStatus,omitempty"`
```
xperimental, line 572: The status should not be part of the type used in the `spec`. We currently re-use the same type for both `spec` and `status` but just because there's no difference in the data. This will change that, which means that we need to define a separate type for the `status` side. That type _can_ embed the current `ObjectStorageSchema` type though, so that it has the same attributes which can be initialized from an existing type.

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -1007,6 +1017,8 @@ const (
 	ReasonZoneAwareNodesMissing LokiStackConditionReason = "ReasonZoneAwareNodesMissing"
 	// ReasonZoneAwareEmptyLabel when the node-label used for zone-awareness has an empty value.
 	ReasonZoneAwareEmptyLabel LokiStackConditionReason = "ReasonZoneAwareEmptyLabel"
+	// ReasonObjectStorageSchemaShouldBeUpgraded when the object storage schema should be upgraded to the newest version V13
+	ReasonObjectStorageSchemaShouldBeUpgraded LokiStackConditionReason = "ReasonObjectStorageSchemaShouldBeUpgraded"
```
xperimental, line 1021: How about something more succinct like `ReasonSchemaUpgradeRecommended`?

operator/internal/status/storage.go
```diff
@@ -25,5 +42,47 @@ func SetStorageSchemaStatus(ctx context.Context, k k8s.Client, req ctrl.Request,
...
+			if err := updateSchemaStatus(s.Status.Storage.Schemas, StorageSchemaOutOfRetention); err != nil {
+				return kverrors.Wrap(err, "error updating schema status")
+			}
+		} else {
+			//s.Status.Storage.SchemaStatus = StorageSchemaNeedsUpgrade
+			if err := updateSchemaStatus(s.Status.Storage.Schemas, StorageSchemaNeedsUpgrade); err != nil {
+				return kverrors.Wrap(err, "error updating schema status")
+			}
+
+			if err := SetWarningCondition(ctx, k, req, StorageSchemaNeedsUpgrade, lokiv1.ReasonObjectStorageSchemaShouldBeUpgraded); err != nil {
```
xperimental, line 71: I think updating the condition is better done in `status.Refresh()` where (most) of the other conditions are set. Updating the status in the schema list can still happen here, it makes sense that this happens when the rest of `.status.storage.schemas` is updated.

operator/internal/status/storage.go
```diff
@@ -25,5 +42,47 @@ func SetStorageSchemaStatus(ctx context.Context, k k8s.Client, req ctrl.Request,
...
+		} else {
+			schemaVersionMap[sc.Version] = false
+			oldSchemas = append(oldSchemas, sc)
+		}
+	}
+
+	if len(oldSchemas) > 0 {
+		if schemaVersionMap[lokiv1.ObjectStorageSchemaV13] {
+			//s.Status.Storage.Schemas[] = StorageSchemaOutOfRetention
+			if err := updateSchemaStatus(s.Status.Storage.Schemas, StorageSchemaOutOfRetention); err != nil {
```
xperimental, line 62: I think this can be done easier by going through the slice of schemas once and setting the matching status on it.

operator/internal/status/lokistack.go
```diff
@@ -74,6 +75,16 @@ func SetDegradedCondition(ctx context.Context, k k8s.Client, req ctrl.Request, m
 	return updateCondition(ctx, k, req, degraded)
 }
 
+func SetWarningCondition(ctx context.Context, k k8s.Client, req ctrl.Request, msg string, reason lokiv1.LokiStackConditionReason) error {
```
xperimental, line 78: This does not really work as I expected. The warning conditions need to be able to be active in addition to the other conditions. I think the overall "health" of the LokiStack should still be "Ready" even when there are warning conditions active. This would also leave us the option of having more than one warning condition active at the same time, which is already possible with the two warnings we currently have.

operator/internal/status/storage.go
```diff
@@ -25,5 +42,47 @@ func SetStorageSchemaStatus(ctx context.Context, k k8s.Client, req ctrl.Request,
...
+				return kverrors.Wrap(err, "error updating schema status")
+			}
+		} else {
+			//s.Status.Storage.SchemaStatus = StorageSchemaNeedsUpgrade
+			if err := updateSchemaStatus(s.Status.Storage.Schemas, StorageSchemaNeedsUpgrade); err != nil {
```
xperimental, line 67: I don't know if I would "attach" the "needs upgrade" warning to a schema as this might be interpreted as "oh, I need to change this version to `v13` instead of "you need to add a new schema in the future which uses v13".

xperimental (commented):
Overall a good start :+1: 
I have not tested this on a cluster yet, only looked at the source so far, but I added some comments already.

I second Peri's hint for an alert. I'm currently assuming that this is how the "insights" feature works in OpenShift.

operator/apis/loki/v1/lokistack_types.go
btaani: You're right, thanks for pointing this out

operator/internal/status/storage.go
btaani: The reason I did this is because the status `StorageSchemaOutOfRetention` is only activated if we have a newer v13 schema added whose effective date is in the past.

operator/internal/status/storage.go
btaani: it's called `StorageSchemaNeedsUpgrade` but the text is actually `"Consider upgrading to schema V13 to use TSDB shipper"`
so it's more a recommendation than a requirement

operator/apis/loki/v1beta1/lokistack_types.go
```diff
@@ -766,6 +766,32 @@ type LokiStackComponentStatus struct {
 	Ruler PodStatusMap `json:"ruler,omitempty"`
 }
 
+// ObjectStorageStatusSchema defines the requirements needed to configure a new
+// storage schema status.
+type ObjectStorageStatusSchema struct {
```
xperimental, line 771: This is in the `v1beta1` API. Even when we decide to keep the new status struct in `v1`, I don't think we should change the old version of the API.

operator/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Main
 
+- [11158](https://github.com/grafana/loki/pull/11158) **btaani**: operator: Add support to add additional storage schema V13
```
xperimental, line 3: I think the feature we're adding is the warning nudging the user to upgrade the schema:

```suggestion
- [11158](https://github.com/grafana/loki/pull/11158) **btaani**: operator: Add warning for old schema configuration
```

operator/internal/status/status.go
```diff
@@ -59,6 +59,31 @@ func Refresh(ctx context.Context, k k8s.Client, req ctrl.Request, now time.Time)
...
+				Reason:             string(reason),
+				Status:             metav1.ConditionTrue,
+				LastTransitionTime: metav1.Now(),
+			}
+
+			if warningExists(warning, stack.Status.Conditions) {
+				continue
+			}
+			stack.Status.Conditions = append(stack.Status.Conditions, warning)
+		}
```
xperimental, line 86: This does not work as expected. The warning is added with `status: true` but on subsequent updated the status is reset to false, rendering the condition inactive. I think re-using the "support multiple conditions" [commit](https://github.com/grafana/loki/commit/f390f3734f5a123affe5f6438888b7f9a5ab6387) from my spike might be a good candidate.

operator/internal/status/storage.go
xperimental: That's true, but I still think attaching this text to an existing schema is not what we want, because the user can not upgrade an existing schema, they have to add a new one that gets activated in the future.

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -567,6 +567,32 @@ type ObjectStorageSchema struct {
 	EffectiveDate StorageSchemaEffectiveDate `json:"effectiveDate"`
 }
 
+// ObjectStorageStatusSchema defines the requirements needed to configure a new
+// storage schema status.
+type ObjectStorageStatusSchema struct {
```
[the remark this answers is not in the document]
btaani: But the warning condition will not specify which schema needs upgrade/is out of retention. This status field is associated with each specific schema

operator/internal/status/status.go
btaani: Yeah thanks, I forgot to change this one

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -1097,6 +1101,8 @@ const (
 	ReasonZoneAwareNodesMissing LokiStackConditionReason = "ReasonZoneAwareNodesMissing"
 	// ReasonZoneAwareEmptyLabel when the node-label used for zone-awareness has an empty value.
 	ReasonZoneAwareEmptyLabel LokiStackConditionReason = "ReasonZoneAwareEmptyLabel"
+	// ReasonStorageSchemaVersionIsOld when the object storage schema version is older than V13
+	ReasonStorageSchemaVersionIsOld LokiStackConditionReason = "ReasonStorageSchemaVersionIsOld"
```
xperimental, line 1105: At some point we started using the `Reason` prefix in the value as well, I don't think we need to do that, though:

```suggestion
	ReasonStorageNeedsSchemaUpdate LokiStackConditionReason = "StorageNeedsSchemaUpdate"
```

operator/internal/status/lokistack.go
```diff
@@ -22,6 +19,7 @@ const (
 	messagePending                = "Some LokiStack components pending on dependencies"
 	messageDegradedMissingNodes   = "Cluster contains no nodes matching the labels used for zone-awareness"
 	messageDegradedEmptyNodeLabel = "No value for the labels used for zone-awareness"
+	messageOldSchemaVersion       = "The schema configuration contains one or more schemas that do not use the most recent version."
```
xperimental, line 22: I just realized, that this warning is still not actionable by the user. The "has future old versions" warning in the spike did go a step into the right direction but it also failed to make the point:

Currently the user can not remove old schema versions, so having a warning trigger on "there is an old schema version in the configuration" does not make sense, it only matters if the configuration ends with the latest version, so that the warning goes away once the v13 schema version is added.

So the following warning should make more sense:

```suggestion
	messageWarningNeedsSchemaVersionUpdate = "The schema configuration does not contain the most recent schema version and needs an update."
```

operator/internal/status/lokistack.go
```diff
@@ -149,53 +159,18 @@ func checkForZoneawareNodes(ctx context.Context, k client.Client, zones []lokiv1
...
-func updateCondition(ctx context.Context, k k8s.Client, req ctrl.Request, condition metav1.Condition) error {
-	var stack lokiv1.LokiStack
-	if err := k.Get(ctx, req.NamespacedName, &stack); err != nil {
-		if apierrors.IsNotFound(err) {
-			return nil
+func generateWarnings(ctx context.Context, cs *lokiv1.LokiStackComponentStatus, k k8s.Client, req ctrl.Request, stack *lokiv1.LokiStack) ([]metav1.Condition, error) {
+	warnings := make([]metav1.Condition, 0)
+	for _, sc := range stack.Status.Storage.Schemas {
+		if sc.Version != lokiv1.ObjectStorageSchemaV13 {
+			warnings = append(warnings, metav1.Condition{
```
xperimental, line 166: See the comment on the warning message: I think we should change this to only check if the schema configuration ends in a schema that uses the latest version.

operator/internal/status/lokistack.go
```diff
@@ -149,53 +159,18 @@ func checkForZoneawareNodes(ctx context.Context, k client.Client, zones []lokiv1
 	return true, true, nil
 }
 
-func updateCondition(ctx context.Context, k k8s.Client, req ctrl.Request, condition metav1.Condition) error {
-	var stack lokiv1.LokiStack
-	if err := k.Get(ctx, req.NamespacedName, &stack); err != nil {
-		if apierrors.IsNotFound(err) {
-			return nil
+func generateWarnings(ctx context.Context, cs *lokiv1.LokiStackComponentStatus, k k8s.Client, req ctrl.Request, stack *lokiv1.LokiStack) ([]metav1.Condition, error) {
```
xperimental, line 162: Function has unused parameters:

```suggestion
func generateWarnings(stack *lokiv1.LokiStack) ([]metav1.Condition, error) {
```

operator/internal/status/lokistack.go
```diff
@@ -63,18 +61,30 @@ func (e *DegradedError) Error() string {
...
+		return nil, err
 	}
 
-	return updateCondition(ctx, k, req, degraded)
+	conditions = append(conditions, mainCondition)
+	return conditions, nil
 }
 
-func generateCondition(ctx context.Context, cs *lokiv1.LokiStackComponentStatus, k k8s.Client, req ctrl.Request, stack *lokiv1.LokiStack) (metav1.Condition, error) {
+func generateCondition(ctx context.Context, cs *lokiv1.LokiStackComponentStatus, k k8s.Client, req ctrl.Request, stack *lokiv1.LokiStack, degradedErr *DegradedError) (metav1.Condition, error) {
```
xperimental, line 79: Function has unused parameters:

```suggestion
func generateCondition(ctx context.Context, cs *lokiv1.LokiStackComponentStatus, k k8s.Client, stack *lokiv1.LokiStack, degradedErr *DegradedError) (metav1.Condition, error) {
```

operator/internal/status/lokistack.go
btaani: This makes sense, thanks

operator/internal/manifests/internal/config/build_test.go
```diff
@@ -5418,3 +5418,296 @@ analytics:
 		})
 	}
 }
+
+func TestBuild_ConfigAndRuntimeConfig_AddSchema(t *testing.T) {
```
xperimental, line 5422: I'm not sure what we're testing here. Do we still need this test or is it a left-over from previous versions of this PR?

operator/internal/status/lokistack_test.go
```diff
@@ -306,10 +200,164 @@ func TestGenerateCondition_ZoneAwareLokiStack(t *testing.T) {
...
+			require.NotEmpty(t, s.Status.Storage.Schemas)
+
+			require.NotZero(t, k.StatusCallCount())
+			require.NotZero(t, sw.UpdateCallCount())
+
+			condition := generateWarnings(s.Status.Storage.Schemas)
+			require.Equal(t, condition, tc.wantCondition)
+
+		})
+	}
```
xperimental, line 362: A lot of the boilerplate for this test can be removed now:

```suggestion
	tt := []struct {
		desc          string
		schemas       []lokiv1.ObjectStorageSchema
		wantCondition []metav1.Condition
	}{
		{
			desc: "no V13 in schema config",
			schemas: []lokiv1.ObjectStorageSchema{
				{
					Version:       lokiv1.ObjectStorageSchemaV11,
					EffectiveDate: "2020-10-11",
				},
				{
					Version:       lokiv1.ObjectStorageSchemaV12,
					EffectiveDate: "2023-10-11",
				},
			},
			wantCondition: []metav1.Condition{{
				Type:    string(lokiv1.ConditionWarning),
				Reason:  string(lokiv1.ReasonStorageNeedsSchemaUpdate),
				Message: messageWarningNeedsSchemaVersionUpdate,
			}},
		},
		{
			desc: "with V13 not as the last element in schema config",
			schemas: []lokiv1.ObjectStorageSchema{
				{
					Version:       lokiv1.ObjectStorageSchemaV11,
					EffectiveDate: "2020-10-11",
				},
				{
					Version:       lokiv1.ObjectStorageSchemaV13,
					EffectiveDate: "2023-10-11",
				},
				{
					Version:       lokiv1.ObjectStorageSchemaV12,
					EffectiveDate: "2024-10-11",
				},
			},
			wantCondition: []metav1.Condition{{
				Type:    string(lokiv1.ConditionWarning),
				Reason:  string(lokiv1.ReasonStorageNeedsSchemaUpdate),
				Message: messageWarningNeedsSchemaVersionUpdate,
			}},
		}, {
			desc: "with V13 as the last element in schema config",
			schemas: []lokiv1.ObjectStorageSchema{
				{
					Version:       lokiv1.ObjectStorageSchemaV11,
					EffectiveDate: "2020-10-11",
				},
				{
					Version:       lokiv1.ObjectStorageSchemaV12,
					EffectiveDate: "2023-10-11",
				},
				{
					Version:       lokiv1.ObjectStorageSchemaV13,
					EffectiveDate: "2024-10-11",
				},
			},
			wantCondition: []metav1.Condition{},
		},
	}

	for _, tc := range tt {
		tc := tc
		t.Run(tc.desc, func(t *testing.T) {
			t.Parallel()

			condition := generateWarnings(tc.schemas)
			require.Equal(t, condition, tc.wantCondition)
		})
	}
```

operator/internal/status/lokistack.go
```diff
@@ -149,53 +156,19 @@ func checkForZoneawareNodes(ctx context.Context, k client.Client, zones []lokiv1
...
-	var stack lokiv1.LokiStack
-	if err := k.Get(ctx, req.NamespacedName, &stack); err != nil {
-		if apierrors.IsNotFound(err) {
-			return nil
-		}
-		return kverrors.Wrap(err, "failed to lookup LokiStack", "name", req.NamespacedName)
+func generateWarnings(schemas []lokiv1.ObjectStorageSchema) []metav1.Condition {
+	warnings := make([]metav1.Condition, 0)
+	if len(schemas) == 0 {
+		return nil
```
xperimental, line 162: The return value of `generateWarnings` is used to `append` the other conditions to, so I let it always return an initialized slice, even though it might be empty (GH does not let me make a suggestion for the whole function, so I can only list the new function):

```go
func generateWarnings(schemas []lokiv1.ObjectStorageSchema) []metav1.Condition {
	warnings := make([]metav1.Condition, 0, 2)

	if len(schemas) > 0 && schemas[len(schemas)-1].Version != lokiv1.ObjectStorageSchemaV13 {
		warnings = append(warnings, metav1.Condition{
			Type:    string(lokiv1.ConditionWarning),
			Reason:  string(lokiv1.ReasonStorageNeedsSchemaUpdate),
			Message: messageWarningNeedsSchemaVersionUpdate,
		})
	}

	return warnings
}
```

xperimental (commented):
The new warning works nicely :+1: 

I left a few comments in the PR. I'll defer to @periklis for another pass.

operator/internal/manifests/internal/config/build_test.go
btaani: It's a leftover from the previous version of the PR

operator/internal/status/lokistack.go
```diff
@@ -63,18 +61,27 @@ func (e *DegradedError) Error() string {
 	return fmt.Sprintf("cluster degraded: %s", e.Message)
 }
 
-// SetDegradedCondition appends the condition Degraded to the lokistack status conditions.
-func SetDegradedCondition(ctx context.Context, k k8s.Client, req ctrl.Request, msg string, reason lokiv1.LokiStackConditionReason) error {
-	degraded := metav1.Condition{
-		Type:    string(lokiv1.ConditionDegraded),
-		Message: msg,
-		Reason:  string(reason),
+func generateConditions(ctx context.Context, cs *lokiv1.LokiStackComponentStatus, k k8s.Client, req ctrl.Request, stack *lokiv1.LokiStack, degradedErr *DegradedError) ([]metav1.Condition, error) {
```
periklis, line 64:
```suggestion
func generateConditions(ctx context.Context, cs *lokiv1.LokiStackComponentStatus, k k8s.Client, stack *lokiv1.LokiStack, degradedErr *DegradedError) ([]metav1.Condition, error) {
```

Param `req ctrl.Request` is unused here.

operator/internal/status/lokistack.go
```diff
@@ -149,53 +156,16 @@ func checkForZoneawareNodes(ctx context.Context, k client.Client, zones []lokiv1
...
-func updateCondition(ctx context.Context, k k8s.Client, req ctrl.Request, condition metav1.Condition) error {
-	var stack lokiv1.LokiStack
-	if err := k.Get(ctx, req.NamespacedName, &stack); err != nil {
-		if apierrors.IsNotFound(err) {
-			return nil
-		}
-		return kverrors.Wrap(err, "failed to lookup LokiStack", "name", req.NamespacedName)
-	}
+func generateWarnings(schemas []lokiv1.ObjectStorageSchema) []metav1.Condition {
+	warnings := make([]metav1.Condition, 0, 2)
```
periklis, line 160:
```suggestion
	warnings := make([]metav1.Condition, 0, 1)
```

nit: We allocate only one warning, the backing array of this slice can stay with a capacity of 1. Or do I miss something, like a second warning condition somewhere else?

operator/internal/status/lokistack_test.go
periklis, line 1: Do we have replacements of the three tests removed here? Was there an overlap with some other test file?

periklis (commented):
LGTM and smoke testing working fine too. A couple of nits and one remaining question about removed tests.

operator/internal/status/lokistack.go
btaani: This one is just a typo I guess

operator/internal/status/lokistack_test.go
btaani: Those three were testing the function `SetDegradedCondition()` which no longer exists because of the refactoring that I borrowed from Robert's spike (check the first two commits that are authored by him)

operator/internal/status/lokistack.go
xperimental: The two was there on purpose, to have room for the condition that is always there (ready, pending, failed, ...) and one warning.

periklis (approved):

PR #11162, opened 2023-11-07
config: remove unordered_writes setting

ashwanthgoli:
accepting out of order writes has been the default behavior since v2.4.x and the option to configure `unordered_writes` has been deprecated in v2.9.x. this pr removes the config option.

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


salvacorts (approved):

docs/sources/operations/request-validation-rate-limits.md
```diff
@@ -125,13 +125,13 @@ This validation error is returned when a stream is submitted without any labels.
 | Sample discarded        | **Yes**       |
 | Configurable per tenant | No            |
 
-## `too_far_behind` and `out_of_order`
+## `too_far_behind`
 
-The `too_far_behind` and `out_of_order` reasons are identical. Loki clusters with `unordered_writes=true` (the default value as of Loki v2.4) use `reason=too_far_behind`. Loki clusters with `unordered_writes=false` use `reason=out_of_order`.
+Loki enforces a validity window on the earliest timestamp for entries in a given log stream.
+Log entries older than `highest_timestamp_of_entries_in_stream - (max_chunk_age/2)` will be rejected with `too_far_behind` error.
```
JStickler, line 131:
```suggestion
Log entries older than `highest_timestamp_of_entries_in_stream - (max_chunk_age/2)` will be rejected with a `too_far_behind` error.
```

docs/sources/operations/request-validation-rate-limits.md
```diff
@@ -125,13 +125,13 @@ This validation error is returned when a stream is submitted without any labels.
...
 
-This validation error is returned when a stream is submitted out of order. More details can be found [here](/docs/loki/latest/configuration/#accept-out-of-order-writes) about Loki's ordering constraints.
-
-The `unordered_writes` config value can be modified globally in the [`limits_config`](/docs/loki/latest/configuration/#limits_config) block, or on a per-tenant basis in the [runtime overrides](/docs/loki/latest/configuration/#runtime-configuration-file) file, whereas `max_chunk_age` is a global configuration.
+For example, if `max_chunk_age` is 2 hours and the stream `{foo="bar"}` has one entry at `8:00`, Loki will accept data for that stream as far back in time as `7:00`.
+If another log line is written at `10:00`, Loki will accept data for that stream as far back in time as `9:00`.
```
JStickler, line 134: I know it's not part of this PR, but line 136 is really unclear (what problem are we referring to?).  And then line 136 and 137 give completely conflicting advice about whether or not users should change the `max_chunk_age` value or not.  Should we clean those lines up now as part of this PR?

cstyan:
If the config option has only just become deprecated shouldn't we for a time include a warning notification when used that it is deprecated and the value is being ignored, plus that the flag will be removed in a future release? 

I guess this is safe since you're merging into main instead of the 2.9.x branch, but at the same time I don't see any mention of the this config being deprecated in the 2.9.x release notes/upgrade guide.

ashwanthgoli:
@cstyan agree with you. deprecation note is present in changelog and the configuration page, but it's not highlighted where users would notice. I think we could've been a bit more vocal about it by including it in the 2.9.x release notes and adding a warning log.


making this change now keeping the major release in mind, but I am okay with pausing it if you feel we should hold off for now. Loki would still be able to read ordered chunks, this only removes the option to make out of order writes configurable.

cstyan:
@ashwanthgoli the only concern is that removing the flag when people don't expect it to be removed means loki will fail to start if they haven't removed that line from their config or cli invocation. Since this is for the next public release and not going into 2.9 I think it's fine, but if there's a way we can go back to update the notes for 2.9 and also make sure we make this obvious that it's been removed for 2.10/3.0 lets make sure we do that. I'm not here to block this PR :)

JStickler:
We have the ability to update the 2.9.x release notes if we want to.

cstyan:
@ashwanthgoli how would you like to proceed?

ashwanthgoli:
@cstyan i'll merge this one and update the release notes for 2.9.x to make it explicit that this setting is deprecated

JStickler:
Do we want to add this PR to the [Deprecation Epic](https://github.com/grafana/loki/issues/21372)?

chaudum:
@ashwanthgoli do you want to update the PR and resolve the merge conflicts, or should I open a new one (which seems easier IMO)?

ashwanthgoli:
@chaudum starting fresh seems a better option as the code must have changed quite a bit since this pr.

PR #11166, opened 2023-11-07
Test query tags propagation.

jeschkies:
This is a follow up to https://github.com/grafana/loki/pull/11147 and tests the query tags propagation withing the scheduler processor.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/querier/worker/scheduler_processor_test.go
+++ b/pkg/querier/worker/scheduler_processor_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/grafana/loki/pkg/querier/queryrange"
 	"github.com/grafana/loki/pkg/querier/queryrange/queryrangebase"
 	"github.com/grafana/loki/pkg/scheduler/schedulerpb"
+	"github.com/grafana/loki/pkg/util/httpreq"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/querier/worker/scheduler_processor_test.go
```diff
@@ -101,6 +102,45 @@ func TestSchedulerProcessor_processQueriesOnSingleStream(t *testing.T) {
 		loopClient.AssertCalled(t, "Send", &schedulerpb.QuerierToScheduler{QuerierID: "test-querier-id"})
 	})
 
+	t.Run("should pass the query tags from the request in the context", func(t *testing.T) {
```
dannykopping, line 105: I think this is testing the internals, not the desired behaviour.

What we want is for the query tags to be received in the querier when sent by the frontend; whether it's passed in the context or not is an implementation detail.

IMO this test should be instantiating an HTTP request, passing it to the frontend, and the querier worker should receive it with all the required parameters present.

jeschkies: But how would you test that the querier worker receives it? The only place it's applied is on the logs as far as I can tell.

cstyan:
do we still want this?

PR #11178, opened 2023-11-08
Return checksum from BuildFrom function

poyzannur:
Follow up from https://github.com/grafana/loki/pull/11115 
We want to know the checksum of the bloom blocks before uploading them to storage. Instead of opening the file and reading it, this PR returns the checksum from where it's created.


cyriltovena (approved):
LGTM

owen-d (approved):

PR #11181, opened 2023-11-09
Bloom Gateway: Implement chunk filtering using workers that multiplex requests

chaudum:

This PR adds the worker implementation in the bloom gateways. The workers pull multiple items from the queue, multiplex them, and execute the chunk matching on the resolved bloom blocks.

The multiplexing is used to minimise the overhead of seeking and skipping through bloom blocks when matching chunks.

**Todo**

- [x] Test cases for multiplexing
- [x] Functional tests of the FilterChunkRefs() method

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/bloomgateway/util.go
```diff
@@ -0,0 +1,129 @@
...
+
+func getDay(ts model.Time) int64 {
+	return ts.Unix() / int64(24*time.Hour/time.Second)
+}
+
+func getDayTime(ts model.Time) time.Time {
+	return time.Date(ts.Time().Year(), ts.Time().Month(), ts.Time().Day(), 0, 0, 0, 0, time.UTC)
+}
+
+func filterRequestForDay(r *logproto.FilterChunkRefRequest, day time.Time) *logproto.FilterChunkRefRequest {
```
chaudum, line 59: Could be the schema's DayTime type:

```suggestion
func filterRequestForDay(r *logproto.FilterChunkRefRequest, day config.DayTime) *logproto.FilterChunkRefRequest {
```

pkg/bloomgateway/bloomgateway.go
```diff
@@ -169,34 +141,45 @@ func makePendingTasks(n int) *pendingTasks {
...
 
 	serviceMngr    *services.Manager
 	serviceWatcher *services.FailureWatcher
+
+	workerConfig workerConfig
 }
 
 // New returns a new instance of the Bloom Gateway.
 func New(cfg Config, schemaCfg config.SchemaConfig, storageCfg storage.Config, shardingStrategy ShardingStrategy, cm storage.ClientMetrics, logger log.Logger, reg prometheus.Registerer) (*Gateway, error) {
+	metricsSubsystem := "bloom_gateway"
```
salvacorts, line 167: nit: I'd move this to a const

pkg/bloomgateway/bloomgateway.go
```diff
@@ -355,19 +312,250 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
 		}
 	}
 }
+
+type workerConfig struct {
+	maxWaitTime time.Duration
+	maxItems    int
+}
+
+type workerMetrics struct {
```
salvacorts, line 346: Since this file is getting quite big, I think it may make sense to create a new `metrics.go` file and move these there.

Same for the `metrics` struct here. In fact, I think in metrics.go we should define `metrics` to contain the worker and queue metrics:
```go
type metrics struct {
 workerMetrics
 queue.Metrics
 otherMetric prometheus.Summary
}
```

pkg/bloomgateway/bloomgateway.go
```diff
@@ -355,19 +312,250 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
+		}, append(labels, "operation")),
+	}
+}
+
+// worker is a datastructure that consumes tasks from the request queue,
+// processes them and returns the result/error back to the response channels of
+// the tasks.
+// It is responsible for multiplexing tasks so they can be processes in a more
+// efficient way.
+type worker struct {
```
salvacorts, line 388: I think we could move the worker logic to a new sub pkg `pkg/bloomgateway/worker`. And have:
- Logic inside `worker.go`
- Config inside `config.go`
- Metrics inside `metrics.go`

pkg/bloomgateway/bloomgateway.go
```diff
@@ -215,19 +198,32 @@ func New(cfg Config, schemaCfg config.SchemaConfig, storageCfg storage.Config, s
...
+	}
+	g.Service = services.NewBasicService(g.starting, g.running, g.stopping).WithName("bloom-gateway")
+
+	return g, nil
+}
+
+func (g *Gateway) initServices() error {
+	var err error
 	svcs := []services.Service{g.queue, g.activeUsers}
+	for i := 0; i < numWorkers; i++ {
```
salvacorts, line 215: From the TODO earlier in this file I understand this will be made configurable on a followup PR.

pkg/bloomgateway/bloomgateway.go
```diff
@@ -355,19 +312,250 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
+					}
+				}
+				// continue with tasks of next day
+				continue
+			}
+
+			hasNext := it.Next()
+			for _, bq := range bqs {
+				requests = requests[:0]
+				for hasNext && it.At().Fp <= bq.MaxFp {
```
salvacorts, line 525: nit: I think this would better read as:
```go
for it.Next() && it.At().Fp <= bq.MaxFp {
...
}
```

pkg/bloomgateway/bloomgateway.go
```diff
@@ -355,19 +312,250 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
+				err := fq.Run()
+				if err != nil {
+					for _, t := range tasks {
+						t.ErrCh <- errors.Wrap(err, "failed to run chunk check")
+					}
+					continue
+				}
+			}
+
+			for hasNext {
```
salvacorts, line 544: nit: same as above: `for it.Next() {`

pkg/bloomgateway/bloomgateway.go
```diff
@@ -355,19 +312,250 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
+			}
+
+			it.Reset()
+
+			storeFetchStart := time.Now()
+			// GetBlockQueriers() waits until all blocks are downloaded and available for querying.
+			// TODO(chaudum): Add API that allows to process blocks as soon as they become available.
+			// This will require to change the taskMergeIterator to a slice of requests so we can seek
+			// to the appropriate fingerprint range within the slice that matches the block's fingerprint range.
+			bqs, err := w.store.GetBlockQueriers(taskCtx, tasks[0].Tenant, day, day.Add(24*time.Hour).Add(-1*time.Nanosecond), fingerprints)
```
salvacorts, line 495: nit: rename `bqs` for `blockQueriers` for better readability throughout the rest of the code.

pkg/bloomgateway/bloomgateway.go
```diff
@@ -355,19 +312,250 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
+							Chks: convertToChunkRefs(ref.Refs),
+						}
+					}
+				}
+				// continue with tasks of next day
+				continue
+			}
+
+			hasNext := it.Next()
+			for _, bq := range bqs {
```
salvacorts, line 523: nit: related to above, rename `bq` to `blockQuerier`.

pkg/bloomgateway/multiplexing.go
```diff
@@ -0,0 +1,135 @@
...
+	// Request is the original request
+	Request *logproto.FilterChunkRefRequest
+	// ErrCh is a send-only channel to write an error to
+	ErrCh chan<- error
+	// ResCh is a send-only channel to write partial responses to
+	ResCh chan<- v1.Output
+}
+
+// NewTask returns a new Task that can be enqueued to the task queue.
+// As additional arguments, it returns a result and an error channel, as well
```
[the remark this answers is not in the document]
chaudum: The Task should hold a send-only channel, but the function that creates the task needs to read from the channel.

pkg/bloomgateway/bloomgateway.go
```diff
@@ -355,19 +312,250 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
+
+			fromDay, throughDay := task.Bounds()
+
+			if fromDay.Equal(throughDay) {
+				tasksPerDay[fromDay] = append(tasksPerDay[fromDay], task)
+			} else {
+				// split task into separate tasks per day
+				for i := fromDay; i.Before(throughDay); i = i.Add(24 * time.Hour) {
+					r := filterRequestForDay(task.Request, i)
+					t := task.WithRequest(r)
```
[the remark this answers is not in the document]
chaudum: Maybe name it `CopyWithRequest()`?

pkg/bloomgateway/bloomgateway.go
chaudum: Right.

pkg/queue/queue.go
```diff
@@ -125,6 +125,31 @@ func (q *RequestQueue) Enqueue(tenant string, path []string, req Request, maxQue
 	}
 }
 
+func (q *RequestQueue) DequeueMany(ctx context.Context, last QueueIndex, consumerID string, maxItems int, maxWait time.Duration) ([]Request, QueueIndex, error) {
```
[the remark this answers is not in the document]
chaudum: Good call!

pkg/bloomgateway/bloomgateway.go
chaudum: It would be better to read, but it would not work as you think:

When you call `it.Next()` the iterator will proceed to the next item. However, if the condition `it.At().Fp <= bq.MaxFp` does not match, the loop is exited.
Then the loop is started for the next `bq` again, and first, `it.Next()` will be called, proceeding to the next item, and therefore skipping one item.

pkg/bloomgateway/bloomgateway.go
chaudum: IMO that's not necessary. Golang is not like Java where every single class has its own file.
500-ish LOC are not that many for a go file.
The only thing I could imagine is splitting the worker related code into a separate file.

pkg/storage/stores/shipper/bloomshipper/store.go
```diff
@@ -40,84 +45,18 @@ func (bs *BloomStore) Stop() {
...
-	}
-	return chunkRefs, nil
-}
-
-func (bs *BloomStore) queriers(ctx context.Context, tenant string, from, through time.Time, fingerprints []uint64) (*bloomQueriers, error) {
-	bf := newBloomFilters(1024)
-	err := bs.shipper.ForEachBlock(ctx, tenant, from, through, fingerprints, func(bq *v1.BlockQuerier) error {
-		bf.queriers = append(bf.queriers, bq)
+func (bs *BloomStore) GetBlockQueriers(ctx context.Context, tenant string, from, through time.Time, fingerprints []uint64) ([]BlockQuerierWithFingerprintRange, error) {
+	bqs := make([]BlockQuerierWithFingerprintRange, 0, 32)
```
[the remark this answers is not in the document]
chaudum: I'll address the reduction of allocations in a separate PR it that's ok.

pkg/bloomgateway/multiplexing.go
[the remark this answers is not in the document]
salvacorts: Ah gotya, thanks for the clarification.

pkg/bloomgateway/bloomgateway.go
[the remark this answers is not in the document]
salvacorts: That sounds better!

pkg/storage/stores/shipper/bloomshipper/store.go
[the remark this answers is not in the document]
salvacorts: Sure ;)

pkg/bloomgateway/bloomgateway.go
salvacorts: I was more concerned about logically splitting files and following a pattern we already follow elsewhere with config and metrics on separate files. Also, I do feel like workers can be logically separated as right now we have workers as goroutines, but maybe in the future we want workers running like we run queriers.

It's just a nit tho, so feel free to disregard this.

pkg/bloomgateway/bloomgateway.go
salvacorts: Oh I see. Thanks for the clarification. I'd probably add a comment explaining that.

salvacorts (approved):

pkg/bloomgateway/bloomgateway.go
chaudum: I did split the worker code in a separate `worker.go` file :)

pkg/bloomgateway/worker.go
```diff
@@ -0,0 +1,240 @@
...
+			}
+		}
+
+		for day, tasks := range tasksPerDay {
+			logger := log.With(w.logger, "day", day)
+			level.Debug(logger).Log("msg", "process tasks", "tasks", len(tasks))
+
+			it := newTaskMergeIterator(tasks...)
+
+			fingerprints = fingerprints[:0]
```
owen-d, line 156: nit: I don't think we need another list of all the fingerprints here. We could instead build a function which found the intersecting blocks for a list of fingerprints from the original requests.

Simplified below,

```go
func OverlappingBlocksForRequests(reqs [][]model.Fingerprint, blocks []Block) []Block
```

We could binary search over the fp lists, comparing them to blocks rather than iterate over every fp directly (can easily be `n=millions`).

pkg/bloomgateway/worker.go
```diff
@@ -0,0 +1,240 @@
...
+			for it.Next() {
+				// fingerprints are already sorted. we can skip duplicates by checking
+				// if the next is greater than the previous
+				fp := uint64(it.At().Fp)
+				if len(fingerprints) > 0 && fp <= fingerprints[len(fingerprints)-1] {
```
owen-d, line 161: Have you seen `NewDedupingIter`? Might be a bit heavyweight here, but I've used it to wrap a heap-iter before to handle items with the same keys.

pkg/bloomgateway/worker.go
```diff
@@ -0,0 +1,240 @@
...
+							Fp:   model.Fingerprint(ref.Fingerprint),
+							Chks: convertToChunkRefs(ref.Refs),
+						}
+					}
+				}
+				// continue with tasks of next day
+				continue
+			}
+
+			hasNext := it.Next()
```
owen-d, line 201: Can use the `view` iter idea described earlier for this

pkg/queue/queue.go
```diff
@@ -125,6 +125,31 @@ func (q *RequestQueue) Enqueue(tenant string, path []string, req Request, maxQue
...
+	// create a context for dequeuing with a max time we want to wait to fullfill the desired maxItems
+
+	dequeueCtx, cancel := context.WithTimeout(ctx, maxWait)
+	defer cancel()
+
+	var idx QueueIndex
+	items := make([]Request, 0, maxItems)
```
owen-d, line 135: todo: pool

owen-d (commented):
This is looking really good.

I need to refactor the bloom querying to return the list of chunks which can be removed rather than the ones which need to be queried in storage. The reasoning for this is we can merge a removal list of chunks across bloom blocks by unioning them, but we can't do the opposite.

pkg/bloomgateway/bloomgateway.go
```diff
@@ -278,71 +270,36 @@ func (g *Gateway) stopping(_ error) error {
...
+		return &logproto.FilterChunkRefResponse{
+			ChunkRefs: req.Refs,
+		}, nil
+	}
+
 	for _, ref := range req.Refs {
 		if ref.Tenant != tenantID {
 			return nil, errors.Wrapf(errInvalidTenant, "expected chunk refs from tenant %s, got tenant %s", tenantID, ref.Tenant)
 		}
+		// Sort ShortRefs by From time in ascending order
```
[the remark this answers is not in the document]
chaudum: > We should also assume these chunkrefs are already sorted -- no need to sort them again here. The index gateway should take care of that (I think it does already since the index is laid out in this order as well)

You're right, they are sorted already.

pkg/bloomgateway/worker.go
owen-d: I've gone ahead and put together something like this here: https://github.com/grafana/loki/pull/11237

chaudum: You'd still need the list of fingerprints to calculating the overlapping blocks. Or do I miss something?

pkg/bloomgateway/bloomgateway.go
```diff
@@ -278,71 +270,36 @@ func (g *Gateway) stopping(_ error) error {
...
 		return nil, err
 	}
 
+	// Shortcut if request does not contain filters
+	if len(req.Filters) == 0 {
+		return &logproto.FilterChunkRefResponse{
+			ChunkRefs: req.Refs,
+		}, nil
+	}
+
```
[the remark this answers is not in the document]
chaudum: :+1: Removed the assertion

pkg/bloomgateway/bloomgateway.go
```diff
@@ -354,19 +301,58 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
+			// log line is helpful for debugging tests
+			// level.Debug(g.logger).Log("msg", "got partial result", "task", task.ID, "tenant", tenantID, "fp", uint64(res.Fp), "chunks", res.Removals.Len(), "progress", fmt.Sprintf("%d/%d", len(responses), requestCount))
 			// wait for all parts of the full response
-			response = append(response, res)
-			if len(response) == len(req.Refs) {
-				return &logproto.FilterChunkRefResponse{ChunkRefs: response}, nil
+			if len(responses) == requestCount {
+				for _, o := range responses {
+					// we must not remove items from req.Refs as long as the worker may iterater over them
+					g.removeNotMatchingChunks(req, o)
```
vlad-diachenko, line 322: we could skip the call to this method if `Removals.Len()` is `0`.

vlad-diachenko (approved):
lgtm

pkg/bloomgateway/worker.go
```diff
@@ -0,0 +1,236 @@
...
+								Removals: nil,
+							}
+						}
+					}
+					// continue with tasks of next day
+					continue
+				}
+
+				boundedRefs := partitionFingerprintRange(tasks, blockRefs)
+				blockRefs = blockRefs[0:]
```
owen-d, line 184:
```suggestion
				blockRefs = blockRefs[:0]
```

pkg/bloomgateway/worker.go
```diff
@@ -0,0 +1,236 @@
...
+					}
+					// continue with tasks of next day
+					continue
+				}
+
+				for i, blockQuerier := range blockQueriers {
+					it := newTaskMergeIterator(day, boundedRefs[i].tasks...)
+					requests = requests[:0]
+					for it.Next() {
+						requests = append(requests, it.At().Request)
```
owen-d, line 208: What's the advantage of collecting these into a slice rather than building an iterator over the underlying iterator?

pkg/queue/queue.go
```diff
@@ -125,6 +127,35 @@ func (q *RequestQueue) Enqueue(tenant string, path []string, req Request, maxQue
...
+	// create a context for dequeuing with a max time we want to wait to fullfill the desired maxItems
+
+	dequeueCtx, cancel := context.WithTimeout(ctx, maxWait)
+	defer cancel()
+
+	var idx QueueIndex
+
+	items := q.pool.Get(maxItems)
+	defer func() {
+		q.pool.Put(items)
```
owen-d, line 140: This looks like it puts the items back into the pool when it returns them to the caller as well, creating a mutability bug waiting to happen.

pkg/queue/util.go
```diff
@@ -0,0 +1,25 @@
+package queue
+
+import "github.com/prometheus/prometheus/util/pool"
+
+// BufferPool uses a bucket pool and wraps the Get() and Put() functions for
+// simpler access.
+type BufferPool[T any] struct {
```
owen-d, line 7: super-nit: Maybe just `SlicePool[T]` would be a better name?

pkg/storage/bloom/v1/merge.go
```diff
@@ -76,15 +76,15 @@ func (mbq *HeapIterator[T]) pop() (T, bool) {
 			return mbq.zero, false
 		}
 
-		cur := mbq.itrs[0]
-		if ok := cur.Next(); !ok {
+		curr := mbq.itrs[0]
+		if ok := curr.Next(); !ok {
```
owen-d, line 80: don't like my naming? 😭

pkg/storage/stores/shipper/bloomshipper/client.go
```diff
@@ -29,6 +28,14 @@ const (
 	fileNamePartDelimiter = "-"
 )
 
+type BoundsCheck uint8
```
owen-d, line 31: I'd love to keep this in the `v1` lib so I can use it there as well (can't import this package there)

owen-d (commented):
Looking good

pkg/bloomgateway/multiplexing.go
```diff
@@ -0,0 +1,225 @@
...
+		return false
+	}
+	for next && !it.matches(it.iter.At()) {
+		next = it.iter.Next()
+		if !next {
+			it.cache = it.zero
+			return false
+		}
+	}
+	it.cache = it.transform(it.iter.At())
```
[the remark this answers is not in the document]
chaudum: I'd like to avoid the `it.transform(it.iter.At())` function call every time `it.At()` is called.
Depending on the transform function, it could be expensive to do so.

pkg/bloomgateway/util.go
```diff
@@ -0,0 +1,163 @@
...
+	if len(refs) == 0 {
+		return model.Earliest, model.Latest
+	}
+	return refs[0].From, refs[len(refs)-1].Through
+}
+
+func convertToSearches(filters []*logproto.LineFilterExpression) [][]byte {
+	searches := make([][]byte, 0, len(filters))
+	for _, f := range filters {
+		searches = append(searches, []byte(f.Match))
```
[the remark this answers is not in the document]
chaudum: Right, I haven't thought about other operators than `=`.
Gonna add a TODO comment.

pkg/bloomgateway/worker.go
chaudum: Good catch! :see_no_evil:

pkg/storage/bloom/v1/merge.go
chaudum: 😂

chaudum: This happened accidentally when I reverted a temporary change.

pkg/bloomgateway/worker.go
chaudum: There is none.

PR #11196, opened 2023-11-10
Handle unexpected response types in volume API

dannykopping:
We're currently seeing panics in `k174` due to `nil` responses being passed due to an unsuccessful type assertion which was not checked.

This PR moves the type assertion to where it's needed and handles it gracefully.

We don't yet know what response is being passed into the channel; this log line will help us identify it.

N/A

There was also some variable shadowing in this code which could produce some interesting bugs, so I renamed the variables to avoid that.


dannykopping:
Superseded by #11209

PR #11201, opened 2023-11-10
bloom blocks downloading queue

vlad-diachenko:
implemented bloom blocks downloading queue to control the concurrency of downloading the blocks from the storage


pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,224 @@
...
+
+	for i := 0; i < config.BlocksDownloadingQueue.WorkersCount; i++ {
+		go b.serveDownloadingTasks(fmt.Sprintf("worker-%d", i))
+	}
+	return b
+}
+
+type BlockDownloadingTask struct {
+	ctx context.Context
+	// Request is the original request
```
chaudum, line 63: I guess the comment can be removed.

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,224 @@
...
+type BlockDownloadingResult struct {
+	extractedBlockPath string
+	block              BlockRef
+}
+
+func (d *blockDownloader) downloadBlocks(ctx context.Context, tenantID string, references []BlockRef) (chan blockWithQuerier, chan error) {
+	d.activeUsersService.UpdateUserTimestamp(tenantID, time.Now())
+	errCh := make(chan error)
+	blocksCh := make(chan blockWithQuerier, len(references))
+	go func() {
```
chaudum, line 133: I'm questioning whether we want to spawn a goroutine for each `downloadBlocks` call. Enqueueing is already non-blocking.

I think you can avoid that by returning actual results by the `downloadBlocks()` function (instead of the channels). That would remove the `for { select {} }` in the `ForEachBlock()` function of the shipper.

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,224 @@
...
+	return workingDirectoryPath, nil
+}
+
+func (s *blockDownloader) createBlockQuerier(directory string) *v1.BlockQuerier {
+	reader := v1.NewDirectoryBlockReader(directory)
+	block := v1.NewBlock(reader)
+	return v1.NewBlockQuerier(block)
+}
+
+func writeDataToTempFile(workingDirectoryPath string, block *Block) (string, error) {
```
chaudum, line 202: Maybe add a comment that this function is supposed to close the block.

pkg/storage/stores/shipper/bloomshipper/shipper.go
```diff
@@ -4,32 +4,35 @@ import (
...
-	logger log.Logger
+	client          Client
+	config          config.Config
+	logger          log.Logger
+	blockDownloader *blockDownloader
 }
 
-func NewShipper(client Client, config config.Config, logger log.Logger) (*Shipper, error) {
+type Limits interface {
+	BloomGatewayBlocksDownloadingParallelism(tenantID string) int
```
chaudum, line 25: Could be more concise IMO

```suggestion
	BloomShipperDownloadParallelism(tenantID string) int
```

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,224 @@
...
+		workingDirectory:   config.WorkingDirectory,
+		queueMetrics:       queueMetrics,
+		queue:              downloadingQueue,
+		blockClient:        blockClient,
+		activeUsersService: activeUsersService,
+		limits:             limits,
+	}
+
+	for i := 0; i < config.BlocksDownloadingQueue.WorkersCount; i++ {
+		go b.serveDownloadingTasks(fmt.Sprintf("worker-%d", i))
```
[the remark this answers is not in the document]
owen-d: agree, but be careful with the `for ctx.Err` bit as explained in detail [here](https://github.com/grafana/loki/pull/11181#discussion_r1393498266). It should use `select` to avoid saturating the cpu.

owen-d (commented):
took a brief look, letting @chaudum do the lions share of review on this one

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,224 @@
+package bloomshipper
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
+	"github.com/grafana/loki/pkg/util"
```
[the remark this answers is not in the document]
vlad-diachenko: done

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,224 @@
+package bloomshipper
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
```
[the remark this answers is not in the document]
vlad-diachenko: done

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,224 @@
...
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/grafana/loki/pkg/queue"
+	"github.com/grafana/loki/pkg/storage/stores/shipper/bloomshipper/config"
+	"github.com/grafana/loki/pkg/util/constants"
+	"github.com/prometheus/client_golang/prometheus"
```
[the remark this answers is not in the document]
vlad-diachenko: done

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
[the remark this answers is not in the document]
vlad-diachenko: I expected queue to be stopped in this case and the worker would receive `ErrStopped` and would close the goroutine 
https://github.com/grafana/loki/blob/27411cff087e6fa2c660c6bfa577c31945ab5232/pkg/queue/queue.go#L146 
wdyt?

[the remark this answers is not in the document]
vlad-diachenko: ok, but it looks like we do not call StopAsync on the queue ;)

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,224 @@
...
+		ctx:       ctx,
+		block:     block,
+		ErrCh:     errCh,
+		ResultsCh: resCh,
+	}
+}
+
+func (d *blockDownloader) serveDownloadingTasks(workerID string) {
+	logger := log.With(d.logger, "worker", workerID)
+	level.Debug(logger).Log("msg", "starting worker", "worker", workerID)
```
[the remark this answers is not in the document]
vlad-diachenko: done

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
vlad-diachenko: done

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
vlad-diachenko: I want to return the channels because some blocks might be already in the cache, so we can start processing those immediately but for the rest, we will have to download and extract them, which might take some time ... 
wdyt?

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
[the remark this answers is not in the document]
vlad-diachenko: I managed to achieve the workers shutdown without `select` statement, because `dequeue`  function also checks if the context is not canceled.  
https://github.com/grafana/loki/pull/11201/commits/6b05ec5e236b8747f9c0222bd318f7963419fe7b

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
vlad-diachenko: done

pkg/storage/stores/shipper/bloomshipper/shipper.go
vlad-diachenko: yes, it is also correct, but `BloomGatewayBlocksDownloadingParallelism` is more descriptive and explains immediately that it's an option that affects exactly blocks downloading parallelism.  so, I would leave it as is if it's not critical ;)

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,233 @@
...
+	return &BlockDownloadingTask{
+		ctx:       ctx,
+		block:     block,
+		ErrCh:     errCh,
+		ResultsCh: resCh,
+	}
+}
+
+// noop implementation
+var onWorkerStopCallback = func() {}
```
chaudum, line 93: Could we move this package global callback function into the `blockDownloader` struct?
It would be even easier to use in the tests:

```go
downloader, err := newBlockDownloader(...)
downloader.onWorkerStopCallback = func() { stoppedWorkersCount.Inc() }
```

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,233 @@
...
+	for {
+		item, newIdx, err := d.queue.Dequeue(ctx, idx, workerID)
+		if err != nil {
+			if !errors.Is(err, queue.ErrStopped) && !errors.Is(err, context.Canceled) {
+				level.Error(logger).Log("msg", "failed to dequeue task", "err", err)
+				continue
+			}
+			level.Info(logger).Log("msg", "stopping worker")
+			//this callback is used only in the tests to assert that worker is stopped
+			onWorkerStopCallback()
```
chaudum, line 113: I would invoke the callback as a deferred function

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,233 @@
...
+			BlockRef:     task.block,
+			BlockQuerier: blockQuerier,
+		}
+	}
+}
+
+type BlockDownloadingResult struct {
+	extractedBlockPath string
+	block              BlockRef
+}
```
chaudum, line 150: An exported struct with only non-exported fields is pretty useless for using in another package, unless it implements an interface.
If the type is not used outside of the package, you can make it non-exported.

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,233 @@
...
+
+func (d *blockDownloader) downloadBlocks(ctx context.Context, tenantID string, references []BlockRef) (chan blockWithQuerier, chan error) {
+	d.activeUsersService.UpdateUserTimestamp(tenantID, time.Now())
+	// we need to have errCh with size that can keep max count of errors to prevent the case when
+	// the queue worker reported the error to this channel before the current goroutine
+	// and this goroutine will go to the deadlock because it won't be able to report an error
+	// because nothing reads this channel at this moment.
+	errCh := make(chan error, len(references))
```
chaudum, line 158: Not sure if I understand that correctly. On  line 168 you return after sending the first and only error to the `errCh`. Could you just make buffered channel of size `1`?

pkg/storage/stores/shipper/bloomshipper/config/config.go
```diff
@@ -8,11 +8,23 @@ import (
 )
 
 type Config struct {
-	WorkingDirectory string `yaml:"working_directory"`
+	WorkingDirectory       string                 `yaml:"working_directory"`
+	BlocksDownloadingQueue DownloadingQueueConfig `yaml:"blocks_downloading_queue"`
```
chaudum, line 12: Do we really need the nesting here? Leave that up to you.

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
vlad-diachenko: this channel is also passed to the queue and queue workers will report errors to this channel also...

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
vlad-diachenko: it is just a leftover, removed it completely

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
vlad-diachenko: make sense. done

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
vlad-diachenko: make sense ;) thanks

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,233 @@
...
+		block:     block,
+		ErrCh:     errCh,
+		ResultsCh: resCh,
+	}
+}
+
+// noop implementation
+var onWorkerStopCallback = func() {}
+
+func (d *blockDownloader) serveDownloadingTasks(ctx context.Context, workerID string) {
```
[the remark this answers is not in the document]
vlad-diachenko: done

pkg/storage/stores/shipper/bloomshipper/config/config.go
vlad-diachenko: I believe that it will be better to have it separated ;) also do not have a strong opinion on it... just separated it because in the next PR I will add cache config and do not want to mix the properties in the single block....

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
vlad-diachenko: fixed

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,233 @@
...
+	logger := log.With(d.logger, "worker", workerID)
+	level.Debug(logger).Log("msg", "starting worker")
+
+	d.queue.RegisterConsumerConnection(workerID)
+	defer d.queue.UnregisterConsumerConnection(workerID)
+
+	idx := queue.StartIndexWithLocalQueue
+
+	for {
```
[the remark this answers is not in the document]
vlad-diachenko: it's not necessary here because the context will be checked during `Dequeue` function call...
also, we have discussed it here https://raintank-corp.slack.com/archives/C056KUQFBFT/p1700665187391789

pkg/storage/stores/shipper/bloomshipper/client.go
```diff
@@ -190,42 +190,23 @@ func (b *BloomClient) DeleteMeta(ctx context.Context, meta Meta) error {
...
-		if err != nil {
-			errChannel <- fmt.Errorf("error downloading block file: %w", err)
-			return
-		}
-		//close blocks channel only if there is no error
-		close(blocksChannel)
-	}()
-	return blocksChannel, errChannel
+// GetBlock downloads the blocks from objectStorage and returns the downloaded block
+func (b *BloomClient) GetBlock(ctx context.Context, reference BlockRef) (Block, error) {
```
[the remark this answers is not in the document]
vlad-diachenko: we will download many blocks in parallel using the downloading queue. I see that in the compactor it might be necessary to download many blocks in parallel also, but I believe we can spin up the goroutines in compactor's code to download many blocks in parallel. wdyt?

pkg/storage/stores/shipper/bloomshipper/client.go
```diff
@@ -190,42 +190,23 @@ func (b *BloomClient) DeleteMeta(ctx context.Context, meta Meta) error {
...
+	period, err := findPeriod(b.periodicConfigs, reference.StartTimestamp)
+	if err != nil {
+		return Block{}, fmt.Errorf("error while period lookup: %w", err)
+	}
+	objectClient := b.periodicObjectClients[period]
+	readCloser, _, err := objectClient.GetObject(ctx, createBlockObjectKey(reference.Ref))
```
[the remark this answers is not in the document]
vlad-diachenko: it's already handled in https://github.com/grafana/loki/blob/d96c9396933ebf015568c7d4a7ff0f4568915273/pkg/storage/stores/shipper/bloomshipper/block_downloader.go#L134

pkg/bloomgateway/bloomgateway_test.go
```diff
@@ -248,3 +248,21 @@ func TestBloomGateway_FilterChunkRefs(t *testing.T) {
 		require.ElementsMatch(t, tenants, gw.activeUsers.ActiveUsers())
 	})
 }
+
+type fakeLimits struct {
+}
+
+func (f fakeLimits) BloomGatewayShardSize(_ string) int {
+	//TODO implement me
+	panic("implement me")
```
chaudum, line 257:
```suggestion
	return 0
```

pkg/bloomgateway/bloomgateway_test.go
```diff
@@ -248,3 +248,21 @@ func TestBloomGateway_FilterChunkRefs(t *testing.T) {
...
+}
+
+func (f fakeLimits) BloomGatewayEnabled(_ string) bool {
+	//TODO implement me
+	panic("implement me")
```
chaudum, line 262:
```suggestion
	return true
```

pkg/bloomgateway/bloomgateway_test.go
```diff
@@ -248,3 +248,21 @@ func TestBloomGateway_FilterChunkRefs(t *testing.T) {
...
+}
+
+func (f fakeLimits) BloomGatewayBlocksDownloadingParallelism(_ string) int {
+	//TODO implement me
+	panic("implement me")
```
chaudum, line 267:
```suggestion
	return 1
```

pkg/storage/stores/shipper/bloomshipper/block_downloader.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func (d *blockDownloader) createBlockQuerier(directory string) *v1.BlockQuerier {
+	reader := v1.NewDirectoryBlockReader(directory)
+	block := v1.NewBlock(reader)
+	return v1.NewBlockQuerier(block)
+}
+
+func (d *blockDownloader) stop() {
+	_ = services.StopManagerAndAwaitStopped(d.ctx, d.manager)
```
chaudum, line 205: Maybe we should log the error?

pkg/storage/stores/shipper/bloomshipper/config/config.go
```diff
@@ -8,11 +8,23 @@ import (
...
+}
+
+type DownloadingQueueConfig struct {
+	WorkersCount              int `yaml:"workers_count"`
+	MaxTasksEnqueuedPerTenant int `yaml:"max_tasks_enqueued_per_tenant"`
+}
+
+func (cfg *DownloadingQueueConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	f.IntVar(&cfg.WorkersCount, prefix+"workers-count", 100, "The count of parallel workers that download Bloom Blocks.")
```
chaudum, line 21: I think we should be more conservative regarding the default value. `10` should  be more than sufficient?

pkg/storage/stores/shipper/bloomshipper/config/config.go
```diff
@@ -8,11 +8,23 @@ import (
...
+	f.IntVar(&cfg.MaxTasksEnqueuedPerTenant, prefix+"max_tasks_enqueued_per_tenant", 10_000, "Maximum number of task in queue per tenant per bloom-gateway. Enqueuing the tasks above this limit will fail an error.")
```
chaudum, line 22: Same here.

chaudum (approved):
Left a few comments, but nothing that would block merging.

PR #11209, opened 2023-11-10
populate empty resonse for IndexStatsResponse and VolumeResponse

owen-d:
Fixes a panic when the underlying response struct is accessed but not populated after an out-of-bounds request is short-circuited via `NewEmptyResponse` [here](https://github.com/grafana/loki/blob/main/pkg/querier/queryrange/limits.go#L155-L164). This PR ensures the embedded types are not nil in accordance with the other variants in this function.


jeschkies (approved):
Nice one. Could you extend the tests?

PR #11222, opened 2023-11-13
Owen d/fix/nil ptr due to empty resp

trevorwhitney:
Replaces https://github.com/grafana/loki/pull/11209 to fix CI

### DIFF
--- a/.github/workflows/vulnerability-scan.yml
+++ b/.github/workflows/vulnerability-scan.yml
@@ -4,14 +4,12 @@ on: pull_request
 permissions:
   pull-requests: write
   contents: write
+  issues: write
 
 jobs:
   snyk:
     name: Snyk Scan
     runs-on: ubuntu-latest
-    permissions:
-      issues: write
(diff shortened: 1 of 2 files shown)


.github/workflows/vulnerability-scan.yml
```diff
@@ -4,6 +4,7 @@ on: pull_request
 permissions:
   pull-requests: write
   contents: write
+  issues: write
```
jeschkies, line 7: I don't think this is related is it?

trevorwhitney: I wasn't sure if the problem was caused by https://github.com/mshick/add-pr-comment/issues/43 or https://github.com/mshick/add-pr-comment/issues/67, so in the interest of speed I fixed both. I can remove this if we'd prefer to narrow this down now, or we can punt as this fixed seemed urgent?

trevorwhitney:
11209 was merged

PR #11233, opened 2023-11-15
Remove querier wait time metric.

jeschkies:
We would like to know how long a querier worker is idle to understand if workstealing would have an impact. The original metric was too noisy and its cardinality was too high. Instead, we are going to log the wait time.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/querier/worker/processor_manager.go
```diff
@@ -64,7 +65,9 @@ func (pm *processorManager) concurrency(n int) {
 		n = 0
 	}
 
+	workerId := 0
```
ashwanthgoli, line 68: i am guessing the workerId does not matter much here. but there is a chance here of re-using workerIds if the concurrency value changes.

jeschkies: Good question. I wonder what happens then. Are we restarting the querier?

ashwanthgoli: when more schedulers get added? which means each worker gets a smaller concurrency value.
not something that is often done in practice, so it should be alright I think

ashwanthgoli (approved):
lgtm

pkg/querier/worker/frontend_processor.go
```diff
@@ -58,7 +58,7 @@ func (fp *frontendProcessor) notifyShutdown(ctx context.Context, conn *grpc.Clie
 }
 
 // runOne loops, trying to establish a stream to the frontend to begin request processing.
-func (fp *frontendProcessor) processQueriesOnSingleStream(ctx context.Context, conn *grpc.ClientConn, address string) {
+func (fp *frontendProcessor) processQueriesOnSingleStream(ctx context.Context, conn *grpc.ClientConn, address, _ string) {
```
cstyan, line 61: is the `_` param intentional here?

pkg/querier/worker/frontend_processor_test.go
```diff
@@ -39,7 +39,7 @@ func TestRecvFailDoesntCancelProcess(t *testing.T) {
 		running.Store(true)
 		defer running.Store(false)
 
-		mgr.processQueriesOnSingleStream(ctx, cc, "test:12345")
+		mgr.processQueriesOnSingleStream(ctx, cc, "test:12345", "1")
```
cstyan, line 42: is `"1"` necessary?

cstyan (commented):
would the new native histograms be helpful here as well? these are much less costly and have less relative error between buckets

there's essentially a single series per native histogram metric and buckets are populated sparsely in storage

pkg/querier/worker/frontend_processor.go
jeschkies: Yes, `processQueriesOnSingleStream` implements the `processor` interface and we don't use the worker ID in the case of the frontend processor.

pkg/querier/worker/frontend_processor_test.go
jeschkies: You are right. Since it's ignored we don't need it here.

PR #11234, opened 2023-11-15
[bloom-compactor] Move meta.json creation at the end of compaction cycle

poyzannur:
This is a follow up from https://github.com/grafana/loki/pull/11115
Instead of a creating a meta file per bloom creation, create a meta file per compaction cycle. 

Fixes #<issue number>


salvacorts (approved):

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -559,23 +545,42 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
...
 		}
 
-		err = CompactNewChunks(ctx, logger, job, chks, bt, bloomShipperClient, c.cfg.WorkingDirectory)
+		storedBlocks, err := CompactNewChunks(ctx, logger, job, chks, bt, bloomShipperClient, c.cfg.WorkingDirectory)
 		if err != nil {
-			return err
+			return level.Error(logger).Log("compacting new chunks", err)
 		}
+
+		storedBlockRefs := make([]bloomshipper.BlockRef, len(storedBlocks))
```
salvacorts, line 576: TODO: use pool for block refs and for tombstone refs. We should pass the storedBlocks buffer to CompactNewChunks to avoid allocations.

PR #11235, opened 2023-11-15
[bloom-compactor] Add configs to enable compactor per tenant

poyzannur:
We want to control bloom compaction per tenant basis. Adding configs to enable/disable bloom compactor. 

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/bloomcompactor/bloomcompactor.go
+++ b/pkg/bloomcompactor/bloomcompactor.go
@@ -293,6 +293,11 @@ func (c *Compactor) compactUsers(ctx context.Context, logger log.Logger, sc stor
 			return fmt.Errorf("interrupting compaction of tenants: %w", err)
 		}
 
+		// Skip tenant if compaction is not enabled
+		if !c.limits.BloomCompactorEnabled(tenant) {
+			return level.Info(tenantLogger).Log("msg", "compaction disabled for tenant")
+		}
+
 		// Skip this table if it is too new/old for the tenant limits.
 		now := model.Now()
 		tableMinAge := c.limits.BloomCompactorMinTableAge(tenant)
@@ -557,10 +562,6 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job, 
 	// Ensure the context has not been canceled (ie. compactor shutdown has been triggered).
 	if err := ctx.Err(); err != nil {
 		return err
-	}
-
-	if !c.limits.BloomCompactorEnabled(job.tenantID) {
-		return level.Error(c.logger).Log("msg", "compaction disabled for tenant: ", job.tenantID)
 	}
 
 	// TODO call bloomShipperClient.GetMetas to get existing meta.json
--- a/pkg/bloomcompactor/bloomcompactor.go
+++ b/pkg/bloomcompactor/bloomcompactor.go
@@ -295,7 +295,8 @@ func (c *Compactor) compactUsers(ctx context.Context, logger log.Logger, sc stor
 
 		// Skip tenant if compaction is not enabled
 		if !c.limits.BloomCompactorEnabled(tenant) {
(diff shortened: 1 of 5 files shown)


pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -559,6 +559,10 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
 		return err
 	}
 
+	if !c.limits.BloomCompactorEnabled(job.tenantID) {
```
salvacorts, line 562: I think this limit should be applied inside the loop over tenants in `compactUsers`. The limit should probably just print a warning (even info?) message and continue to the next tenant. This way you also skip doing a bunch of work for the tenant.

pkg/validation/limits.go
```diff
@@ -301,6 +302,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&l.BloomCompactorShardSize, "bloom-compactor.shard-size", 1, "The shard size defines how many bloom compactors should be used by a tenant when computing blooms. If it's set to 0, shuffle sharding is disabled.")
 	f.DurationVar(&l.BloomCompactorMaxTableAge, "bloom-compactor.max-table-age", 7*24*time.Hour, "The maximum age of a table before it is compacted. Do not compact tables older than the the configured time. Default to 7 days. 0s means no limit.")
 	f.DurationVar(&l.BloomCompactorMinTableAge, "bloom-compactor.min-table-age", 1*time.Hour, "The minimum age of a table before it is compacted. Do not compact tables newer than the the configured time. Default to 1 hour. 0s means no limit. This is useful to avoid compacting tables that will be updated with out-of-order writes.")
+	f.BoolVar(&l.BloomCompactorEnabled, "bloom-compactor.enable-compaction", false, "Whether to compact chunks into bloom filters.")
```
salvacorts, line 305: IIUC we want this to be false by default at the beginning so we selectively enable it for testing, right? 

But eventually, this will be true by default. Wondering because the log line about skipping the tenant can get a bit noisy.

pkg/bloomcompactor/bloomcompactor.go
poyzannur: Yeah i was torn in between, happy to move there.

pkg/validation/limits.go
poyzannur: Yes correct, we want to selectively enable at first. Probably true for large cells, but i don't think for dev. 
We can cross that bridge then?

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -293,6 +293,11 @@ func (c *Compactor) compactUsers(ctx context.Context, logger log.Logger, sc stor
 			return fmt.Errorf("interrupting compaction of tenants: %w", err)
 		}
 
+		// Skip tenant if compaction is not enabled
+		if !c.limits.BloomCompactorEnabled(tenant) {
+			return level.Info(tenantLogger).Log("msg", "compaction disabled for tenant")
```
chaudum: Not sure about returning the return value of `Log()`.
While it works perfectly fine, I think writing the `return` in a separate line looks cleaner.

chaudum (approved):

pkg/bloomcompactor/bloomcompactor.go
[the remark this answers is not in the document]
poyzannur: oof end of day silliness. sorry

poyzannur: it wasn't suppose to end the loop for this. it has a `continue` on a separate line now.

owen-d (approved):

PR #11237, opened 2023-11-15
boundscheck & partitioning fingerprints between blocks

owen-d:
Adds a few utilities for comparing fingerprints to blocks that cover a specific fingerprint range. Will likely need to be refactored with more comprehensive types, but the logic still applies.

### DIFF
--- a/pkg/storage/bloom/v1/fuse.go
+++ b/pkg/storage/bloom/v1/fuse.go
@@ -1,6 +1,8 @@
 package v1
 
 import (
+	"sort"
+
 	"github.com/efficientgo/core/errors"
 	"github.com/prometheus/common/model"
 )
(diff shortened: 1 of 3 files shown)


pkg/storage/bloom/v1/fuse.go
```diff
@@ -142,3 +144,43 @@ func (fq *FusedQuerier) Run() error {
...
+	reqs   [][]model.Fingerprint
+}
+
+// reqs models a set of requests covering many fingerprints.
+// consumers models a set of blocks covering different fingerprint ranges
+func partitionFingerprintRange(reqs [][]model.Fingerprint, blocks []FingerprintBounds) (res []boundedRequests) {
+	for _, block := range blocks {
+		bounded := boundedRequests{
+			bounds: block,
+		}
```
chaudum, line 160: Does the slice `reqs` not need to be allocated?

pkg/storage/bloom/v1/fuse.go
```diff
@@ -142,3 +144,43 @@ func (fq *FusedQuerier) Run() error {
...
+
+// boundedRequests is a set of requests that are clamped to a specific range
+type boundedRequests struct {
+	bounds FingerprintBounds
+	reqs   [][]model.Fingerprint
+}
+
+// reqs models a set of requests covering many fingerprints.
+// consumers models a set of blocks covering different fingerprint ranges
+func partitionFingerprintRange(reqs [][]model.Fingerprint, blocks []FingerprintBounds) (res []boundedRequests) {
```
chaudum, line 156: You have a named return value, but you return the value also explicitly.
Personally I'm not a big fan of named return values, because they can hide allocation problems.

chaudum (approved):

PR #11263, opened 2023-11-20
compaction: Separate metrics for tracking retention and compaction

sandeepsukhani:
In PR #9884, we separated the retention loop from compaction to avoid blocking compaction for too long due to some intensive delete requests. Currently, we track retention and compaction using the same metrics. This PR adds separate metrics for monitoring retention operation. I have also updated the Retention dashboard to use the new metrics.

### DIFF
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -126,6 +126,8 @@ func (cfg *Config) Validate() error {
 
 		if cfg.ApplyRetentionInterval == 0 {
 			cfg.ApplyRetentionInterval = cfg.CompactionInterval
+			// add some jitter to avoid running retention and compaction at same time
+			cfg.ApplyRetentionInterval += cfg.CompactionInterval / 2
 		}
 
 		if err := config.ValidatePathPrefix(cfg.DeleteRequestStoreKeyPrefix); err != nil {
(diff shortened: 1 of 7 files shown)


pkg/compactor/compactor.go
```diff
@@ -614,6 +616,8 @@ func (c *Compactor) CompactTable(ctx context.Context, tableName string, applyRet
 		// wait for lock to be released since we can't mark delete requests as processed without checking all the tables
 		select {
 		case <-lockWaiterChan:
+			// refresh index list cache since the compaction would have changed the index files in the object store
+			sc.indexStorageClient.RefreshIndexTableNamesCache(ctx)
```
ashwanthgoli, line 620: we refresh the table [here](https://github.com/grafana/loki/blob/05c4b77ed02df5682d35db6582058c9e2764558c/pkg/compactor/table.go#L119), is this still required when compacting a table?

pkg/compactor/compactor.go
```diff
@@ -126,6 +126,8 @@ func (cfg *Config) Validate() error {
 
 		if cfg.ApplyRetentionInterval == 0 {
 			cfg.ApplyRetentionInterval = cfg.CompactionInterval
+			// add some jitter to avoid running retention and compaction at same time
+			cfg.ApplyRetentionInterval += cfg.CompactionInterval / 2
```
ashwanthgoli, line 130: nit: should we instead add a jitter to when the first retention kicks in?

ashwanthgoli (approved):

pkg/compactor/compactor.go
sandeepsukhani: ahh, right, this one is not required. Will remove it.

PR #11276, opened 2023-11-21
utilities for reducing ngram allocations in token iterator construction

owen-d:
Previously, the `calculatePrefix` function was used to generate the token buffer used, but this wouldn't necessarily have enough space to append the ngram afterwards without reallocating+copying the token buffer. This PR ensures we allocate enough space for both prefix and ngram in the token buffer.

### DIFF
--- a/pkg/storage/bloom/v1/bloom_tokenizer.go
+++ b/pkg/storage/bloom/v1/bloom_tokenizer.go
@@ -2,7 +2,6 @@ package v1
 
 import (
 	"context"
-	"encoding/binary"
 	"math"
 	"time"
 
@@ -14,6 +13,7 @@ import (
 	"github.com/grafana/loki/pkg/logql/log"
 
 	"github.com/grafana/loki/pkg/storage/chunk"
+	"github.com/grafana/loki/pkg/util/encoding"
 	util_log "github.com/grafana/loki/pkg/util/log"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/storage/bloom/v1/bloom_tokenizer.go
```diff
@@ -68,27 +68,32 @@ func clearCache(cache map[string]interface{}) {
...
-	prefix := make([]byte, 32)
-
-	binary.PutVarint(i64buf, int64(chk.From))
-	prefix = append(prefix, i64buf...)
-	binary.PutVarint(i64buf, int64(chk.Through))
-	prefix = append(prefix, i64buf...)
-	binary.LittleEndian.PutUint32(i32buf, chk.Checksum)
-	prefix = append(prefix, i32buf...)
-
-	return prefix
```
owen-d, line 83: This does a few funny things like using intermediate buffers for int encoding, but mostly it returns the prefix after only allocating 32 bytes which is only enough for the chunk information (ignoring varint-encoding, which makes this more complex). Instead, we want to provision enough space for both the prefix and the maximum expected token length. Otherwise, each iteration of the token iterator can and likely will need to reallocate a new slice of larger length, especially when the initial one has a power-of-2 capacity, which we do in this case (32).

paul1r (approved):

paul1r (approved):
lgtm, just need to do the make format

PR #11284, opened 2023-11-22
config: adds `frontend.max-query-capacity` to tune per-tenant query capacity

ashwanthgoli:
Adds a new config `frontend.max-query-capacity` that allows users to configure what portion of the the available querier replicas can be used by a tenant. `max_query_capacity` is the corresponding YAML option that can be configured in limits or runtime overrides. 

For example, setting this to 0.5 would allow a tenant to use half of the available queriers.

This complements the existing `frontend.max-queriers-per-tenant`. When both are configured, the smaller value of the resulting querier replica count is considered:
```
min(frontend.max-queriers-per-tenant, ceil(querier_replicas * frontend.max-query-capacity))
```
*All* queriers will handle requests for a tenant if neither limits are applied.

Fixes #<issue number>

noticed that we don't pass down the shuffle sharding limits for frontend (only using it with schedulers)
https://github.com/grafana/loki/blob/26f097162a856db48ecbd16bef2f0b750029855b/pkg/loki/modules.go#L895
but the [docs](https://github.com/grafana/loki/blob/26f097162a856db48ecbd16bef2f0b750029855b/pkg/validation/limits.go#L276) mention that`frontend.max-queriers-per-tenant` applies to frontend as well.
```
This option only works with queriers connecting to the query-frontend / query-scheduler, not when using downstream URL.
```

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


JStickler (commented):
[docs team]  Docs LGTM

pkg/queue/tenant_queues.go
```diff
@@ -151,6 +154,20 @@ func (q *tenantQueues) getOrAddQueue(tenant string, path []string, maxQueriers i
...
+	}
+
+	maxQueriers := int(math.Ceil(float64(consumers) * uq.maxQueryCapacity))
+
+	if uq.maxQueriers != 0 && uq.maxQueriers < maxQueriers {
+		return uq.maxQueriers
+	}
+
+	return maxQueriers
+}
```
dannykopping, line 169: `tenantQueue` should not have an additional responsibility of calculating the max capacity, and know how to juggle between the two settings.

This code can also expect some subtle bugs in the future because it relies on something outside to set `uq.maxQueriers` = `max_queriers_per_tenant` correctly. Right now it works just fine, but after a refactor in future...?

To improve the safety of this code, I'd suggest creating a type which accepts all the required inputs (`max_queriers_per_tenant`, `max_query_capacity`) and returns the capacity with a function which accepts `consumers` as input. You can then externalise all this business logic, it'll be self-documenting, and still equally testable.

pkg/queue/tenant_queues_test.go
```diff
@@ -173,6 +173,60 @@ func TestQueuesWithQueriers(t *testing.T) {
 	assert.InDelta(t, stdDev, 0, mean*0.2)
 }
 
+func TestTenantQueueMaxQueiers(t *testing.T) {
```
dannykopping, line 176: 👌

pkg/queue/queue.go
```diff
@@ -79,15 +79,15 @@ func NewRequestQueue(maxOutstandingPerTenant int, forgetDelay time.Duration, met
 // between calls.
 //
 // If request is successfully enqueued, successFn is called with the lock held, before any querier can receive the request.
-func (q *RequestQueue) Enqueue(tenant string, path []string, req Request, maxQueriers int, successFn func()) error {
+func (q *RequestQueue) Enqueue(tenant string, path []string, req Request, maxQueriers int, maxQueryCapacity float64, successFn func()) error {
```
dannykopping, line 82: I find this to be a bit of a code smell. We shouldn't need to push down settings like this for each tenant through the call graph because we know the tenant ID and can retrieve it using `Overrides`, no?

Besides that, it just _feels_ strange to specify the capacity (or multiple capacity settings) when enqueueing a request; the original design may have been a bit flawed, but this seems like a good opportunity to correct it?

dannykopping (commented):
Looks pretty good, but I think there's some refactoring that needs to be done for safety and clarity.

pkg/queue/queue.go
ashwanthgoli: totally agree with this. I update the pr with the suggested changes.

pkg/scheduler/scheduler.go
```diff
@@ -130,13 +128,13 @@ func NewScheduler(cfg Config, limits Limits, log log.Logger, ringManager *lokiri
...
 		log:    log,
-		limits: limits,
+		limits: schedulerLimits,
 
 		pendingRequests:    map[requestKey]*schedulerRequest{},
 		connectedFrontends: map[string]*connectedFrontend{},
 		queueMetrics:       queueMetrics,
 		ringManager:        ringManager,
-		requestQueue:       queue.NewRequestQueue(cfg.MaxOutstandingPerTenant, cfg.QuerierForgetDelay, queueMetrics),
+		requestQueue:       queue.NewRequestQueue(cfg.MaxOutstandingPerTenant, cfg.QuerierForgetDelay, limits.QueueLimits(schedulerLimits), queueMetrics),
```
ashwanthgoli, line 137: this would look neater if we move it to the queue pkg `queue.Limits(frontendLimits)`
but I left it in scheduler pkg since it is an adapter between scheduler and queue limits.

pkg/queue/tenant_queues.go
```diff
@@ -120,19 +122,15 @@ func (q *tenantQueues) deleteQueue(tenant string) {
 }
 
 // Returns existing or new queue for a tenant.
-// MaxQueriers is used to compute which consumers should handle requests for this tenant.
-// If maxQueriers is <= 0, all consumers can handle this tenant's requests.
-// If maxQueriers has changed since the last call, consumers for this are recomputed.
-func (q *tenantQueues) getOrAddQueue(tenant string, path []string, maxQueriers int, maxQueryCapacity float64) Queue {
+// MaxConsumers is used to compute which consumers should handle requests for this tenant.
```
dannykopping, line 125: This feels like an odd place to put this comment

pkg/queue/tenant_queues.go
```diff
@@ -142,10 +140,13 @@ func (q *tenantQueues) getOrAddQueue(tenant string, path []string, maxQueriers i
 		q.mapping.Put(tenant, uq)
 	}
 
-	if uq.maxQueriers != maxQueriers || uq.maxQueryCapacity != maxQueryCapacity {
-		uq.maxQueriers = maxQueriers
-		uq.maxQueryCapacity = maxQueryCapacity
-		uq.consumers = shuffleConsumersForTenants(uq.seed, uq.computeMaxQueriers(len(q.sortedConsumers)), q.sortedConsumers, nil)
+	consumersToSelect, err := q.limits.MaxConsumers(tenant, len(q.sortedConsumers))
+	if err != nil {
+		level.Error(util_log.Logger).Log("msg", "computing max consumers", "tenant", tenant, "error", err)
```
dannykopping, line 145: Are we allowing a fallthrough here because `consumersToSelect` is expected to be set to `0` in the `MaxConsumers` function?

For the same reasons mentioned before, try not to rely on implementation outside this function. Set your expectations explicitly, or be very explicit about what the intended behaviour is. Assuming `consumersToSelect` is `0`, failing to compute the max consumers would result in... all queriers being allowable? That feels like a dangerous fallback.

pkg/queue/tenant_queues.go
```diff
@@ -311,8 +298,14 @@ func (q *tenantQueues) forgetDisconnectedConsumers(now time.Time) int {
 func (q *tenantQueues) recomputeUserConsumers() {
 	scratchpad := make([]string, 0, len(q.sortedConsumers))
 
-	for _, uq := range q.mapping.Values() {
-		uq.consumers = shuffleConsumersForTenants(uq.seed, uq.computeMaxQueriers(len(q.sortedConsumers)), q.sortedConsumers, scratchpad)
+	for _, tenant := range q.mapping.Keys() {
+		if uq := q.mapping.GetByKey(tenant); uq != nil {
+			maxConsumers, err := q.limits.MaxConsumers(tenant, len(q.sortedConsumers))
+			if err != nil {
+				level.Error(util_log.Logger).Log("msg", "computing max consumers", "tenant", tenant, "error", err)
```
dannykopping, line 305: Ditto

pkg/scheduler/limits/definitions.go
```diff
@@ -8,3 +15,38 @@ type Limits interface {
...
+
+func QueueLimits(limits Limits) *queueLimits {
+	return &queueLimits{limits: limits}
+}
+
+type queueLimits struct {
+	limits Limits
+}
+
+// MaxConsumers is used to compute how many of the available consumers are allowed to handle requests for a given tenant.
```
dannykopping, line 27: "consumers" is a term used by the tenant queue I think (correct me if I'm wrong), and I'm not sure it makes sense to use this language here. If we're talking about query workers, so let's refer to query workers.

pkg/scheduler/limits/definitions.go
```diff
@@ -8,3 +15,38 @@ type Limits interface {
...
+func (c *queueLimits) MaxConsumers(tenantID string, allConsumers int) (int, error) {
+	if c == nil || c.limits == nil {
+		return 0, nil
+	}
+
+	// aggregate the max queriers limit in the case of a multi tenant query
+	tenantIDs, err := tenant.TenantIDsFromOrgID(tenantID)
+	if err != nil {
+		return 0, err
```
dannykopping, line 36: If we can't lookup the tenant, I'd argue this is an invalid query.

pkg/scheduler/limits/definitions_test.go
```diff
@@ -0,0 +1,79 @@
...
+		})
+	}
+}
+
+type mockLimits struct {
+	maxQueriers      int
+	maxQueryCapacity float64
+}
+
+func (l mockLimits) MaxQueriersPerUser(_ string) int {
```
dannykopping, line 73: Still required?

dannykopping (commented):
Thanks for the refactor! It reads much more clearly now.
There were a couple troubling things that caught my eye though with the changes, but they shouldn't take long to resolve

pkg/queue/tenant_queues.go
ashwanthgoli: agree, this would allow all queriers.

the other option is to not create/return the queue which would cause the enqeue to fail. I thought that would be a bit extreme to do just because we are not able to fetch a limit.

But given this could only happen for an empty tenant and looking at the other comment, I think I'll go ahead with this.

pkg/scheduler/limits/definitions_test.go
ashwanthgoli: yeah, these are the options that get passed down from modules.go as overrides to scheduler & frontend.
I am wrapping these limits with an adapter to return `MaxConsumers`, which gets passed down to the queue

docs/sources/configure/_index.md
```diff
@@ -2760,6 +2760,20 @@ The `limits_config` block configures global and per-tenant limits in Loki.
 # CLI flag: -frontend.max-queriers-per-tenant
 [max_queriers_per_tenant: <int> | default = 0]
 
+# How much of the available query capacity (queriers) can be used by a single
```
dannykopping, line 2763:
```suggestion
# How much of the available query capacity ("querier" components in distributed mode, "read" components in SSD mode) can be used by a single
```

docs/sources/configure/_index.md
```diff
@@ -2760,6 +2760,20 @@ The `limits_config` block configures global and per-tenant limits in Loki.
...
+# tenant. Allowed values are 0.0 to 1.0. For example, setting this to 0.5 would
+# allow a tenant to use half of the available queriers for processing the query
+# workload. If set to 0, query capacity is determined by
+# frontend.max-queriers-per-tenant. When both frontend.max-queriers-per-tenant
+# and frontend.max-query-capacity are configured, smaller value of the resulting
+# querier replica count is considered: min(frontend.max-queriers-per-tenant,
+# ceil(querier_replicas * frontend.max-query-capacity)). *All* queriers will
+# handle requests for the tenant if neither limits are applied. This option only
+# works with queriers connecting to the query-frontend / query-scheduler, not
+# when using downstream URL.
```
dannykopping, line 2773:
```suggestion
# when using downstream URL.
# Use this feature in a multi-tenant setup where you need to limit query capacity for certain tenants.
```

pkg/scheduler/limits/definitions.go
```diff
@@ -1,7 +1,43 @@
...
+
+func NewQueueLimits(limits Limits) *QueueLimits {
+	return &QueueLimits{limits: limits}
+}
+
+type QueueLimits struct {
+	limits Limits
+}
+
+// MaxConsumers is used to compute how many of the available queriers are allowed to handle requests for a given tenant.
```
dannykopping, line 24: Please add some detail to the comment here explaining the business logic

pkg/validation/limits.go
```diff
@@ -368,6 +370,21 @@ func (l *Limits) Validate() error {
 		level.Warn(util_log.Logger).Log("msg", "The compactor.allow-deletes configuration option has been deprecated and will be ignored. Instead, use deletion_mode in the limits_configs to adjust deletion functionality")
 	}
 
+	if l.MaxQueriersPerTenant < 0 {
+		level.Warn(util_log.Logger).Log("msg", "setting frontend.max-queriers-per-tenant to 0 as it is configured to a value less than 0")
+		l.MaxQueriersPerTenant = 0
+	}
+
+	if l.MaxQueryCapacity < 0 {
```
dannykopping, line 378: Can't we just make them `UintVar`s?

dannykopping (commented):
Looking _much_ cleaner now, thanks @ashwanthgoli!
Just minor nits now

pkg/validation/limits.go
dannykopping: Naming it `MaxQueryCapacityPercent` with a uint would be even more intuitive, I'd argue, but I don't have any strong opinions here. Feel free to leave as is

PR #11285, opened 2023-11-22
Fix bloom compactor startup when using unsupported index type

chaudum:
This PR fixes the check in the bloom compactor startup to ignore all other index types than TSDB.

When updating an existing Loki installation that uses BoltDB Shipper in the schema config, you get the following error:

```
invalid tsdb path:
/data/boltdb-shipper-cache/XXXXXXXXX_index_19682/write-0-1639160260677624223-1700595900
││ create index shipper
││ github.com/grafana/loki/pkg/bloomcompactor.New
││     /src/loki/pkg/bloomcompactor/bloomcompactor.go:159
││ github.com/grafana/loki/pkg/loki.(*Loki).initBloomCompactor
││     /src/loki/pkg/loki/modules.go:1406
││ github.com/grafana/dskit/modules.(*Manager).initModule
││     /src/loki/vendor/github.com/grafana/dskit/modules/modules.go:136
││ github.com/grafana/dskit/modules.(*Manager).InitModuleServices
││     /src/loki/vendor/github.com/grafana/dskit/modules/modules.go:108
││ github.com/grafana/loki/pkg/loki.(*Loki).Run
```

### DIFF
--- a/pkg/bloomcompactor/bloomcompactor.go
+++ b/pkg/bloomcompactor/bloomcompactor.go
@@ -119,14 +119,8 @@ func New(
 	c.storeClients = make(map[config.DayTime]storeClient)
 
 	for i, periodicConfig := range schemaConfig.Configs {
-		var indexStorageCfg indexshipper.Config
-		switch periodicConfig.IndexType {
-		case config.TSDBType:
-			indexStorageCfg = storageCfg.TSDBShipperConfig
-		case config.BoltDBShipperType:
-			indexStorageCfg = storageCfg.BoltDBShipperConfig.Config
-		default:
-			level.Warn(c.logger).Log("msg", "skipping period because index type is unsupported")
+		if periodicConfig.IndexType != config.TSDBType {
+			level.Warn(c.logger).Log("msg", "skipping schema period because index type is not supported", "index_type", periodicConfig.IndexType, "period", periodicConfig.From)
 			continue
 		}
 
@@ -143,7 +137,7 @@ func New(
 
 		indexShipper, err := indexshipper.NewIndexShipper(
 			periodicConfig.IndexTables.PathPrefix,
-			indexStorageCfg,
+			storageCfg.TSDBShipperConfig,
 			objectClient,
 			limits,
 			nil,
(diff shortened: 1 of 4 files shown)


ashwanthgoli (approved):

chaudum:
@ashwanthgoli Still need to add a test case :)

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -344,16 +340,16 @@ func (c *Compactor) compactUsers(ctx context.Context, logger log.Logger, sc stor
 	// TODO: Delete local files for unowned tenants, if there are any.
 }
 
-func (c *Compactor) compactTenant(ctx context.Context, logger log.Logger, sc storeClient, tableName string, tenant string) error {
```
salvacorts, line 347: Why are we removing the logger here? This logger is a child of c.logger but with the tenant key.
https://github.com/grafana/loki/pull/11285/files#diff-a88e1e7a49788f4b1dbe6ea4b03845217353b07f7cae25e0f74042b693f50680R285

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -344,16 +340,16 @@ func (c *Compactor) compactUsers(ctx context.Context, logger log.Logger, sc stor
...
+func (c *Compactor) compactTenant(ctx context.Context, sc storeClient, tableName string, tenant string) error {
+	level.Info(c.logger).Log("msg", "starting compaction of tenant")
 
 	// Ensure the context has not been canceled (ie. compactor shutdown has been triggered).
 	if err := ctx.Err(); err != nil {
 		return err
 	}
 
 	// Tokenizer is not thread-safe so we need one per goroutine.
-	bt, _ := v1.NewBloomTokenizer(prometheus.DefaultRegisterer)
```
salvacorts, line 356: 👍 good catch.

pkg/bloomcompactor/bloomcompactor.go
chaudum: Ah I see. Gonna revert that!

salvacorts (approved):
LGTM. Thank you 👏

poyzannur (approved):
Thanks a lot for quickly fixing it

vlad-diachenko (approved):
lgtm

PR #11287, opened 2023-11-22
update dskit to latest version

sandeepsukhani:
Update dskit to latest version


MichelHollands (approved):
LGTM

periklis:
Hate to say this, but we should always present some list of changes to highlight what's changed and what we picked up from dskit on every bump. It's easy to see the intent, but sometimes we do not incorporate all changes directly or other times we just benefit from indirect changes w/o api/config impact.

PR #11293, opened 2023-11-22
Bloom Compactor: Use distributor's RejectOldSamplesMaxAge to filter newer chunks

salvacorts:

This PR limits the chunks that will be processed by the bloom compactor by filtering out chunks older than `RejectOldSamplesMaxAge` (aka `reject_old_samples_max_age`).


poyzannur:
relates to https://github.com/grafana/loki-private/issues/856

PR #11295, opened 2023-11-22
Compact multiple series into a bloom block

poyzannur:

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/bloomcompactor/bloomcompactor.go
+++ b/pkg/bloomcompactor/bloomcompactor.go
@@ -358,42 +358,48 @@ func (c *Compactor) compactTenant(ctx context.Context, logger log.Logger, sc sto
 	// TODO: Use ForEachConcurrent?
 	errs := multierror.New()
 	if err := sc.indexShipper.ForEach(ctx, tableName, tenant, func(isMultiTenantIndex bool, idx shipperindex.Index) error {
-		if isMultiTenantIndex {
+		if isMultiTenantIndex { // TODO handle multitenant tables
 			return fmt.Errorf("unexpected multi-tenant")
 		}
 
+		var indices []Index
 		// TODO: Make these casts safely
 		if err := idx.(*tsdb.TSDBFile).Index.(*tsdb.TSDBIndex).ForSeries(
 			ctx, nil,
 			0, math.MaxInt64, // TODO: Replace with MaxLookBackPeriod
 			func(labels labels.Labels, fingerprint model.Fingerprint, chksMetas []tsdbindex.ChunkMeta) {
-				job := NewJob(tenant, tableName, idx.Path(), fingerprint, labels, chksMetas)
-				jobLogger := log.With(logger, "job", job.String())
-
-				ownsJob, err := c.sharding.OwnsJob(job)
-				if err != nil {
-					c.metrics.compactionRunUnownedJobs.Inc()
-					level.Error(jobLogger).Log("msg", "failed to check if compactor owns job", "err", err)
-					errs.Add(err)
-					return
-				}
-				if !ownsJob {
-					c.metrics.compactionRunUnownedJobs.Inc()
-					level.Debug(jobLogger).Log("msg", "skipping job because it is not owned by this shard")
-					return
-				}
-
-				if err := c.runCompact(ctx, jobLogger, job, c.bloomShipperClient, bt, sc); err != nil {
-					c.metrics.compactionRunFailedJobs.Inc()
-					errs.Add(errors.Wrap(err, "runBloomCompact"))
-					return
-				}
-
-				c.metrics.compactionRunSucceededJobs.Inc()
+				var temp []tsdbindex.ChunkMeta
+				copy(temp, chksMetas)
+				indices = append(indices, Index{seriesFP: fingerprint, seriesLbs: labels, chunks: temp})
 			},
 		); err != nil {
 			errs.Add(err)
 		}
+
+		job := NewJob(tenant, tableName, idx.Path(), indices)
+		jobLogger := log.With(logger, "job", job.String())
+
+		ownsJob, err := c.sharding.OwnsJob(job) // TODO: A shard should either own all the fps of a job or not
+
+		if err != nil {
+			c.metrics.compactionRunUnownedJobs.Inc()
+			level.Error(jobLogger).Log("msg", "failed to check if compactor owns job", "err", err)
+			errs.Add(err)
+			return errs.Err()
+		}
+		if !ownsJob {
+			c.metrics.compactionRunUnownedJobs.Inc()
+			level.Debug(jobLogger).Log("msg", "skipping job because it is not owned by this shard")
+			return nil
+		}
+
+		if err := c.runCompact(ctx, jobLogger, job, c.bloomShipperClient, bt, sc); err != nil {
+			c.metrics.compactionRunFailedJobs.Inc()
+			errs.Add(errors.Wrap(err, "runBloomCompact"))
+			return errs.Err()
+		}
+
+		c.metrics.compactionRunSucceededJobs.Inc()
 
 		return nil
 	}); err != nil {
@@ -459,7 +465,7 @@ func makeChunkRefs(chksMetas []tsdbindex.ChunkMeta, tenant string, fp model.Fing
 }
 
 // TODO Revisit this step once v1/bloom lib updated to combine blooms in the same series
-func buildBloomBlock(ctx context.Context, logger log.Logger, bloomForChks v1.SeriesWithBloom, job Job, workingDir string) (bloomshipper.Block, error) {
+func buildBloomBlock(ctx context.Context, logger log.Logger, blooms []v1.SeriesWithBloom, job Job, workingDir string) (bloomshipper.Block, error) {
 	// Ensure the context has not been canceled (ie. compactor shutdown has been triggered).
 	if err := ctx.Err(); err != nil {
 		return bloomshipper.Block{}, err
@@ -474,7 +480,7 @@ func buildBloomBlock(ctx context.Context, logger log.Logger, bloomForChks v1.Ser
 		return bloomshipper.Block{}, err
 	}
 
-	checksum, err := builder.BuildFrom(v1.NewSliceIter([]v1.SeriesWithBloom{bloomForChks}))
+	checksum, err := builder.BuildFrom(v1.NewSliceIter(blooms))
 	if err != nil {
 		level.Error(logger).Log("writing bloom", err)
 		return bloomshipper.Block{}, err
@@ -490,8 +496,8 @@ func buildBloomBlock(ctx context.Context, logger log.Logger, bloomForChks v1.Ser
 			Ref: bloomshipper.Ref{
 				TenantID:       job.Tenant(),
 				TableName:      job.TableName(),
-				MinFingerprint: uint64(job.Fingerprint()), // TODO will change once we compact multiple blooms into a block
-				MaxFingerprint: uint64(job.Fingerprint()),
+				MinFingerprint: uint64(job.minFp),
+				MaxFingerprint: uint64(job.maxFp),
 				StartTimestamp: job.From().Unix(),
 				EndTimestamp:   job.Through().Unix(),
 				Checksum:       checksum,
@@ -505,7 +511,7 @@ func buildBloomBlock(ctx context.Context, logger log.Logger, bloomForChks v1.Ser
 }
 
 func createLocalDirName(workingDir string, job Job) string {
-	dir := fmt.Sprintf("bloomBlock-%s-%s-%s-%s-%s-%s", job.TableName(), job.Tenant(), job.Fingerprint(), job.Fingerprint(), job.From(), job.Through())
+	dir := fmt.Sprintf("bloomBlock-%s-%s-%s-%s-%s-%s", job.TableName(), job.Tenant(), job.minFp, job.maxFp, job.From(), job.Through())
 	return filepath.Join(workingDir, dir)
 }
 
@@ -518,21 +524,27 @@ func CompactNewChunks(ctx context.Context, logger log.Logger, job Job,
 		return nil, err
 	}
 
-	// Create a bloom for this series
-	bloomForChks := v1.SeriesWithBloom{
-		Series: &v1.Series{
-			Fingerprint: job.Fingerprint(),
-		},
-		Bloom: &v1.Bloom{
-			ScalableBloomFilter: *filter.NewDefaultScalableBloomFilter(fpRate),
-		},
+	var blooms []v1.SeriesWithBloom
+
+	for _, index := range job.indices {
+		// Create a bloom for this series
+		bloomForChks := v1.SeriesWithBloom{
+			Series: &v1.Series{
+				Fingerprint: index.Fingerprint(),
+			},
+			Bloom: &v1.Bloom{
+				ScalableBloomFilter: *filter.NewDefaultScalableBloomFilter(fpRate),
+			},
+		}
+
+		// Tokenize data into n-grams
+		bt.PopulateSeriesWithBloom(&bloomForChks, chunks)
+
+		blooms = append(blooms, bloomForChks)
 	}
 
-	// Tokenize data into n-grams
-	bt.PopulateSeriesWithBloom(&bloomForChks, chunks)
-
 	// Build and upload bloomBlock to storage
-	blocks, err := buildBloomBlock(ctx, logger, bloomForChks, job, dst)
+	blocks, err := buildBloomBlock(ctx, logger, blooms, job, dst)
 	if err != nil {
 		level.Error(logger).Log("building bloomBlocks", err)
 		return nil, err
@@ -553,8 +565,8 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job, 
 
 	metaSearchParams := bloomshipper.MetaSearchParams{
 		TenantID:       job.tenantID,
-		MinFingerprint: uint64(job.seriesFP),
-		MaxFingerprint: uint64(job.seriesFP),
+		MinFingerprint: uint64(job.minFp),
+		MaxFingerprint: uint64(job.maxFp),
 		StartTimestamp: int64(job.from),
 		EndTimestamp:   int64(job.through),
 	}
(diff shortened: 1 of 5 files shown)


pkg/bloomcompactor/sharding.go
```diff
@@ -14,7 +14,7 @@ var (
 // ShardingStrategy describes whether compactor "owns" given user or job.
 type ShardingStrategy interface {
 	util_ring.TenantSharding
-	OwnsJob(job Job) (bool, error)
+	OwnsFp(tenantID string, fp uint64) (bool, error)
```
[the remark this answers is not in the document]
salvacorts: this would map easier to the already existing method in the interface definition we are overriding: https://github.com/grafana/loki/pull/11295/files#diff-6ef3b2d19fc3ebaf9ba3ae94986583eb6d468b74c6b872438ec8832791aca6b5R42

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -577,35 +588,42 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
...
-			return level.Error(logger).Log("compacting new chunks", err)
-		}
+		for _, index := range job.indices {
+			chks, err := storeClient.chunk.GetChunks(ctx, makeChunkRefs(index.Chunks(), job.Tenant(), index.Fingerprint()))
+			if err != nil {
+				return err
+			}
 
-		storedBlockRefs := make([]bloomshipper.BlockRef, len(storedBlocks))
+			storedBlocks, err := CompactNewChunks(ctx, logger, job, chks, bt, fpRate, bloomShipperClient, c.cfg.WorkingDirectory)
```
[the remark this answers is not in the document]
salvacorts: In addition to the above, here https://github.com/grafana/loki/pull/11295/files#diff-a88e1e7a49788f4b1dbe6ea4b03845217353b07f7cae25e0f74042b693f50680R554 we are incorrectly adding all chunks for each series.

salvacorts (commented):
Left some comments. Also, I think we should add some test cases around this.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -354,46 +353,53 @@ func (c *Compactor) compactTenant(ctx context.Context, logger log.Logger, sc sto
...
 
+		var indices []Index
 		// TODO: Make these casts safely
 		if err := idx.(*tsdb.TSDBFile).Index.(*tsdb.TSDBIndex).ForSeries(
 			ctx, nil,
 			0, math.MaxInt64, // TODO: Replace with MaxLookBackPeriod
 			func(labels labels.Labels, fingerprint model.Fingerprint, chksMetas []tsdbindex.ChunkMeta) {
-				job := NewJob(tenant, tableName, idx.Path(), fingerprint, labels, chksMetas)
-				jobLogger := log.With(logger, "job", job.String())
+				ownsFingerprint, err := c.sharding.OwnsFp(tenant, uint64(fingerprint))
```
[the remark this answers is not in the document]
salvacorts: Agree, but also my bad: this inefficiency comes from my implementation for `OwnsJob`. I think once we have the tenant, we can create a new`util_ring.NewFingerprintShuffleSharding` object passing the tenant subring to it and we can get rid of all the logic inside sharding.go (I'd move the test inside sharding_test.go to bloomcompactor_test.go tho)

I see one problem with this tho. If we don't refresh the tenant subring, the tenant compaction takes long enough, and there is an ownership change while compacting the tenant, we may end up processing fingerprints for a tenant the instance no longer owns. This may be an over-optimization at this point, but I wonder if we should have some tenant ring timeout that forces us to reload the tenant ring.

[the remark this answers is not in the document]
vlad-diachenko: I would expect that the ring could give us the range of the fingerprints that we own, and after this inside `.ForSeries()` we could just check that FP is within the range, something like: `min_fp <= fingerprint <= max_fp`. without the need to work with the ring inside the callback.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -354,46 +353,53 @@ func (c *Compactor) compactTenant(ctx context.Context, logger log.Logger, sc sto
...
+					level.Error(fpLogger).Log("msg", "failed to check if compactor owns fp", "err", err)
 					errs.Add(err)
 					return
 				}
-				if !ownsJob {
-					c.metrics.compactionRunUnownedJobs.Inc()
-					level.Debug(jobLogger).Log("msg", "skipping job because it is not owned by this shard")
+				if !ownsFingerprint {
+					c.metrics.compactionRunSkippedFp.Inc()
+					level.Debug(fpLogger).Log("msg", "skipping fp because it is not owned by this shard")
```
vlad-diachenko, line 379: I believe this message will be too noisy because if we have 12 compactors,  each compactor will report all fingerprints from the rest 11 ranges ... let's say if we have 100k streams, each compactor will report 91.6k fingerprints

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -354,46 +353,53 @@ func (c *Compactor) compactTenant(ctx context.Context, logger log.Logger, sc sto
...
+		var indices []Index
 		// TODO: Make these casts safely
 		if err := idx.(*tsdb.TSDBFile).Index.(*tsdb.TSDBIndex).ForSeries(
 			ctx, nil,
 			0, math.MaxInt64, // TODO: Replace with MaxLookBackPeriod
 			func(labels labels.Labels, fingerprint model.Fingerprint, chksMetas []tsdbindex.ChunkMeta) {
-				job := NewJob(tenant, tableName, idx.Path(), fingerprint, labels, chksMetas)
-				jobLogger := log.With(logger, "job", job.String())
+				ownsFingerprint, err := c.sharding.OwnsFp(tenant, uint64(fingerprint))
+				fpLogger := log.With(logger, "fp", tableName+"_"+tenant+"_"+fingerprint.String())
```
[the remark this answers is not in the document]
vlad-diachenko: yes, and moreover, for the FP that we own we do not print anything. 
I would just add this `KV` on line 373 and 379

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -564,11 +576,10 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
 	if err := ctx.Err(); err != nil {
 		return err
 	}
-
 	metaSearchParams := bloomshipper.MetaSearchParams{
 		TenantID:       job.tenantID,
-		MinFingerprint: uint64(job.seriesFP),
-		MaxFingerprint: uint64(job.seriesFP),
+		MinFingerprint: uint64(job.minFp),
```
vlad-diachenko, line 581: these values are not initialised at this moment yet. you need to call `computeBounds` at the beginning.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -577,35 +588,42 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
...
-			return err
-		}
 
-		fpRate := c.limits.BloomFalsePositiveRate(job.Tenant())
-		storedBlocks, err := CompactNewChunks(ctx, logger, job, chks, bt, fpRate, bloomShipperClient, c.cfg.WorkingDirectory)
-		if err != nil {
-			return level.Error(logger).Log("compacting new chunks", err)
-		}
+		for _, index := range job.indices {
+			chks, err := storeClient.chunk.GetChunks(ctx, makeChunkRefs(index.Chunks(), job.Tenant(), index.Fingerprint()))
```
vlad-diachenko, line 601: it looks like this line will download all the chunks to memory ... we need to use lazyChunk to download the chunk, immediately process it, and allow GC to release the memory. wdyt?

pkg/bloomcompactor/job.go
```diff
@@ -66,36 +69,53 @@ func (j *Job) IndexPath() string {
...
 	maxThrough := model.Earliest
 
-	for _, chunk := range j.chunks {
-		from, through := chunk.Bounds()
-		if minFrom > from {
-			minFrom = from
+	minFp := model.Fingerprint(math.MaxInt64)
+	maxFp := model.Fingerprint(0)
+
+	for _, index := range j.indices {
```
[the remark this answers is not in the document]
vlad-diachenko: @salvacorts using these helper functions would require iterating a few times over the elements to find `min` and `max` . but here we can do both within a single iteration.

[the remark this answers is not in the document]
paul1r: We could potentially get some parallelism by using some channels and goroutines, like:
```
for _, chunk := range idx.chunks {
				from, through := chunk.Bounds()
				if localMinFrom > from {
					localMinFrom = from
				}
				if localMaxThrough < through {
					localMaxThrough = through
				}
			}

			fromChan <- localMinFrom
			throughChan <- localMaxThrough
```
That being said, I'm not sure the loss of readability for an unknown small performance gain right here is a good move.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -61,10 +61,9 @@ import (
 	"github.com/grafana/loki/pkg/util"
 )
 
-// TODO: Make a constants file somewhere
 const (
-	bloomFileName  = "bloom"
-	seriesFileName = "series"
+	bloomFileName  = v1.BloomFileName
+	seriesFileName = v1.SeriesFileName
 )
```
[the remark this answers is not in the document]
poyzannur: Nope.

pkg/bloomcompactor/bloomcompactor.go
poyzannur: Good point, totally agreed. But AFAICT [lazyChunk](https://github.com/grafana/loki/blob/main/pkg/storage/store.go#L384-L421) is not exposed (please let me know if i'm looking at the wrong place). Making a change will result in a larger PR. Should we address that later?

pkg/bloomcompactor/bloomcompactor.go
poyzannur: good catch, addressed right after job creation.

pkg/bloomcompactor/metrics.go
```diff
@@ -18,7 +18,7 @@ type metrics struct {
 	compactionRunSkippedTenants    prometheus.Counter
 	compactionRunSucceededTenants  prometheus.Counter
 	compactionRunFailedTenants     prometheus.Counter
-	compactionRunUnownedJobs       prometheus.Counter
+	compactionRunSkippedFp         prometheus.Counter
```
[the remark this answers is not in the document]
poyzannur: good call, added. but i removed skippedFP as i don't think it adds much.

pkg/bloomcompactor/bloomcompactor.go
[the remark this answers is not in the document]
poyzannur: I removed the fingerprint related metrics altogether as I don't think they were adding much.

pkg/bloomcompactor/bloomcompactor.go
poyzannur: agreed, Salva had the same concern when adding. I removed.

pkg/bloomcompactor/bloomcompactor.go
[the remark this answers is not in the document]
poyzannur: > we may end up processing fingerprints for a tenant the instance no longer owns. 
this was discussed during the first weeks and it was agreed that much duplicate work should be fine for a starting point. In the next compaction (eg. in 15 mins) that should be cleared up.  

@vlad-diachenko AFAICT the ring can't give a range, at least as it is. But in the same spirit, we can ask if the job's entire fingerprint range is within the ring by checking the bounds by the ring and go from there. wdyt? 

Overall the goal of this PR is to merge series into a single bloomblock, are we ok to address it in a follow up?

pkg/bloomcompactor/job.go
```diff
@@ -66,36 +69,53 @@ func (j *Job) IndexPath() string {
...
+			maxFp = index.seriesFP
 		}
 	}
 
 	j.from = minFrom
 	j.through = maxThrough
+
+	j.minFp = minFp
+	j.maxFp = maxFp
 }
```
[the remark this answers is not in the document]
poyzannur: Agreed it's fairly crude but does the job. Any pointers of an existing example or do you have anything else in mind?

pkg/bloomcompactor/job.go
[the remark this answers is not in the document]
poyzannur: Yes I agree Paul it's a very small gain imo. If we are to keep this rudimentary way of calculating, i kinda like the single iteration over the list. 
but if you have a better place/way of doing it, like a existing method in shipper etc, that might be nice.

pkg/bloomcompactor/chunkcompactor.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	return &builder, nil
+}
+
+func (p *PersistentBlockBuilder) BuildFrom(itr v1.Iterator[v1.SeriesWithBloom]) (uint32, error) {
+	return p.builder.BuildFrom(itr)
+}
+
+func (p *PersistentBlockBuilder) Data() (io.ReadCloser, error) {
+	blockFile, err := os.Open(filepath.Join(p.localDst, v1.BloomFileName))
```
vlad-diachenko, line 59: bloomClient will close the reader once the uploading is completed, but we also need to remove the file from the local disk. Do we have such mechanisms?

vlad-diachenko (approved):
Looks brilliant 💎 
 just a single comment but I believe it might require refactoring or adding additional functionality to address it (if it's not implemented yet, because I could just miss it;) )

paul1r (approved):

pkg/bloomcompactor/bloomcompactor.go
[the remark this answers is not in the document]
chaudum: @poyzannur I faced a somewhat similar problem when I implemented the bloom gateway client: There I also need to find the fingerprint ranges for each server instance in the ring. 

... while looking at my PR #11336 I saw you commented on that part of the code.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -583,29 +457,50 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
...
 
 	if len(metas) == 0 {
-		// Get chunks data from list of chunkRefs
-		chks, err := storeClient.chunk.GetChunks(ctx, makeChunkRefs(job.Chunks(), job.Tenant(), job.Fingerprint()))
+		localDst := createLocalDirName(c.cfg.WorkingDirectory, job)
+		defer func() {
+			//clean up the bloom directory
+			if err := os.RemoveAll(localDst); err != nil {
+				level.Error(logger).Log(fmt.Sprintf("failed to remove block directory %s", localDst), err)
+			}
```
chaudum, line 465: Could a non-empty directory lead to a problem later on? Should we fail?

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -583,29 +457,50 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
 	}
 
 	if len(metas) == 0 {
-		// Get chunks data from list of chunkRefs
-		chks, err := storeClient.chunk.GetChunks(ctx, makeChunkRefs(job.Chunks(), job.Tenant(), job.Fingerprint()))
+		localDst := createLocalDirName(c.cfg.WorkingDirectory, job)
+		defer func() {
+			//clean up the bloom directory
+			if err := os.RemoveAll(localDst); err != nil {
+				level.Error(logger).Log(fmt.Sprintf("failed to remove block directory %s", localDst), err)
```
chaudum, line 464: We structure our error logs so they contain a `msg` field and `err` field:

```suggestion
				level.Error(logger).Log("msg", "failed to remove block directory", "dir", localDst, "err", err)
```

chaudum: Please check occurrences further down the line

pkg/bloomcompactor/chunkcompactor.go
```diff
@@ -0,0 +1,184 @@
...
+	return block, nil
+}
+
+func createLocalDirName(workingDir string, job Job) string {
+	dir := fmt.Sprintf("bloomBlock-%s-%s-%s-%s-%s-%s", job.tenantID, job.tenantID, job.minFp, job.maxFp, job.from, job.through)
+	return filepath.Join(workingDir, dir)
+}
+
+// Compacts given list of chunks, uploads them to storage and returns a list of bloomBlocks
+func CompactNewChunks(
```
chaudum, line 149: IMO this function should not be exported, since it's meant for package internal use only

```suggestion
func compactNewChunks(
```

pkg/bloomcompactor/chunkcompactor.go
```diff
@@ -0,0 +1,184 @@
...
+			IndexPath: job.indexPath,
+		},
+		Data: data,
+	}
+
+	return block, nil
+}
+
+func createLocalDirName(workingDir string, job Job) string {
+	dir := fmt.Sprintf("bloomBlock-%s-%s-%s-%s-%s-%s", job.tenantID, job.tenantID, job.minFp, job.maxFp, job.from, job.through)
```
chaudum, line 144: I think one tenantID is enough ;-)

```suggestion
	dir := fmt.Sprintf("bloomBlock-%s-%s-%s-%s-%s", job.tenantID, job.minFp, job.maxFp, job.from, job.through)
```

pkg/bloomcompactor/chunkcompactor.go
```diff
@@ -0,0 +1,184 @@
...
+	bt compactorTokenizer,
+	storeClient chunkClient,
+	builder blockBuilder,
+) (bloomshipper.Block, error) {
+	// Ensure the context has not been canceled (ie. compactor shutdown has been triggered).
+	if err := ctx.Err(); err != nil {
+		return bloomshipper.Block{}, err
+	}
+
+	var blooms []v1.SeriesWithBloom
```
chaudum, line 163: Always allocate correct capacity if you know size upfront. This will safe allocation cycles.

pkg/bloomcompactor/chunkcompactor.go
```diff
@@ -0,0 +1,184 @@
...
+		},
+	}
+
+	// Tokenize data into n-grams
+	tokenizer.PopulateSeriesWithBloom(&bloomForChks, chunks)
+	return bloomForChks
+}
+
+// TODO Test this when bloom block size check is implemented
+func buildBlockFromBloom(
```
chaudum, line 100: The function accepts a list of blooms, so I would reflect this in the function name as well:

```suggestion
func buildBlockFromBlooms(
```

pkg/bloomcompactor/bloomcompactor.go
poyzannur: No it shouldn't lead problems now, i don't think we should fail compaction for that. And how and where we do clean up will certainly change as we'll compress the blocks before uploading.

PR #11307, opened 2023-11-23
Revert "Compression of bloom blocks (#11267)"

vlad-diachenko:
This reverts commit af177034 (https://github.com/grafana/loki/pull/11267)
because the compression/uncompression must be done outside of the client
context: https://raintank-corp.slack.com/archives/C056KUQFBFT/p1700494151990649


paul1r (approved):
Approved, but with the caveat that I (or someone else) needs to augment that bloom block structure as discussed in that slack thread.

PR #11313, opened 2023-11-24
bug(volume): return 400 error for invalid volume request

dannykopping:
Making a request to the `/volume{,_range}` endpoints results in a 500 is a LogQL expression is given instead of the expected label matchers. The reason for this is that the error is not wrapped in such a way as to be interpreted by `ClientHTTPStatusAndError` correctly.

5xx responses from the querier are automatically retried, so any invalid request currently results in 5 retries which are unnecessary.

### DIFF
--- a/clients/pkg/logentry/logql/parser.go
+++ b/clients/pkg/logentry/logql/parser.go
@@ -1,13 +1,14 @@
 package logql
 
 import (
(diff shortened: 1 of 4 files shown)


MichelHollands (approved):
LGTM

PR #11321, opened 2023-11-27
Make Blooms-Gateway queue settings configurable

salvacorts:

This PR adds three new settings to the Bloom-Gateway config:

- `worker_concurrency`: replaces the `numWorkers` constant. Named after the scheduler's `scheduler_worker_concurrency` config.
- `max_outstanding_per_tenant`: replaces the `maxTasksPerTenant` constant. Named after the frontend's  `max_outstanding_per_tenant` config.
- `pending_tasks_initial_capacity`: replaces the pendingTasksInitialCap constant.

**Note to the reviewers:**
- I think in the future `max_outstanding_per_tenant` should be configurable per-tenant. This would require refactoring the queue pkg and the frontend/scheduler, so I think it doesn't make sense it in this PR.
- I moved the `Limits` struct definition to `config.go` so have all the adjustable settings in the same place.

### DIFF
--- a/pkg/bloomgateway/bloomgateway.go
+++ b/pkg/bloomgateway/bloomgateway.go
@@ -69,13 +69,6 @@ import (
 
 var errGatewayUnhealthy = errors.New("bloom-gateway is unhealthy in the ring")
 
-// TODO(chaudum): Make these configurable
-const (
-	numWorkers             = 4
-	maxTasksPerTenant      = 1024
-	pendingTasksInitialCap = 1024
-)
-
 const (
 	metricsSubsystem = "bloom_gateway"
 )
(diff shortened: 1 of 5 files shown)


pkg/bloomgateway/bloomgateway.go
```diff
@@ -69,13 +69,6 @@ import (
 
 var errGatewayUnhealthy = errors.New("bloom-gateway is unhealthy in the ring")
 
-// TODO(chaudum): Make these configurable
-const (
-	numWorkers             = 4
-	maxTasksPerTenant      = 1024
-	pendingTasksInitialCap = 1024
```
chaudum, line 76: I don't think this needs to be configurable. It defines the initial capacity of the `pendingTasks` map, which is used to track the queue size.
There is not much to tweak with this value.

salvacorts: Done :)

vlad-diachenko (approved):
LGTM, completely agree with Christian's comment.

PR #11322, opened 2023-11-27
Avoid allocations when collecting responses from the bloom block searches

chaudum:
by using a slice pool

PR #11324, opened 2023-11-27
Bloom-Gateway: Use pools to reduce allocations

salvacorts:

This PR addresses some TODOs in the Bloom-Gateway component to reduce allocations by pooling.

**Note for reviewers:**
- I'm not really sure about the max size for the preallocated pools. Used 16K for all of them but that's probably too much for pools such as the `responsesPool`, `addressesPool`... Feel free to leave a comment in the PR with an estimate on the size you think it would make sense to set as the max.

### DIFF
--- a/pkg/bloomgateway/bloomgateway.go
+++ b/pkg/bloomgateway/bloomgateway.go
@@ -56,6 +56,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/prometheus/util/pool"
 
 	"github.com/grafana/loki/pkg/logproto"
 	"github.com/grafana/loki/pkg/queue"
@@ -78,6 +79,11 @@ const (
 
 const (
 	metricsSubsystem = "bloom_gateway"
+)
+
+var (
+	// responsesPool pooling array of v1.Output [512,1024,...,16k]
+	responsesPool = pool.New(1<<9, 1<<14, 2, func(size int) interface{} { return make([]v1.Output, 0, size) })
 )
 
 type metrics struct {
@@ -302,8 +308,8 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
 	})
 
 	requestCount := len(req.Refs)
-	// TODO(chaudum): Use pool
-	responses := make([]v1.Output, 0, requestCount)
+	responses := responsesPool.Get(requestCount).([]v1.Output)
+	defer responsesPool.Put(responses)
 
 	for {
 		select {
--- a/pkg/bloomgateway/bloomgateway.go
+++ b/pkg/bloomgateway/bloomgateway.go
@@ -82,7 +82,7 @@ const (
 
 var (
 	// responsesPool pooling array of v1.Output [64, 128, 256, ..., 65536]
-	responsesPool = queue.NewSlicePool[v1.Output](1<<6, 1<<32, 2)
+	responsesPool = queue.NewSlicePool[v1.Output](1<<6, 1<<16, 2)
 )
 
 type metrics struct {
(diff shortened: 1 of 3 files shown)


pkg/bloomgateway/client.go
```diff
@@ -274,12 +307,17 @@ func (c *GatewayClient) serverAddrsForFingerprints(tenantID string, groups []*lo
...
 	}
 
-	fingerprints := make([]uint64, numFingerprints)
-	addresses := make([][]string, numFingerprints)
-	bufDescs, bufHosts, bufZones := ring.MakeBuffersForGet()
+	fingerprints := fingerprintsPool.Get(numFingerprints).([]uint64)
+	defer fingerprintsPool.Put(fingerprints)
+
+	addresses := addressesPool.Get(numFingerprints).([][]string)
+	defer addressesPool.Put(addresses)
```
chaudum, line 314: You cannot both return the slices to the pool and return them in the function call.
The caller would need to be in charge of returning the received buffer to the pool.
Could you remove this in your PR, because I do some refactoring around the grouping of fingerprints by server instance.

pkg/bloomgateway/bloomgateway.go
```diff
@@ -80,6 +81,11 @@ const (
 	metricsSubsystem = "bloom_gateway"
 )
 
+var (
+	// responsesPool pooling array of v1.Output [512,1024,...,16k]
+	responsesPool = pool.New(1<<9, 1<<14, 2, func(size int) interface{} { return make([]v1.Output, 0, size) })
```
chaudum, line 86: Can we use a SlicePool? See https://github.com/grafana/loki/pull/11322

pkg/bloomgateway/client.go
salvacorts: Done

pkg/bloomgateway/client.go
```diff
@@ -276,10 +304,12 @@ func (c *GatewayClient) serverAddrsForFingerprints(tenantID string, groups []*lo
 
 	fingerprints := make([]uint64, numFingerprints)
 	addresses := make([][]string, numFingerprints)
-	bufDescs, bufHosts, bufZones := ring.MakeBuffersForGet()
+
+	buf := ringGetBuffersPool.Get().(*ringGetBuffers)
+	defer ringGetBuffersPool.Put(buf)
```
chaudum, line 309: Should we clear the `buf.Descs`, `buf.Hosts`, and `buf.Zones` before releasing them back to the pool? It would release the items in the slices earlier for GC, right?

chaudum (approved):

pkg/bloomgateway/bloomgateway.go
```diff
@@ -80,6 +80,11 @@ const (
 	metricsSubsystem = "bloom_gateway"
 )
 
+var (
+	// responsesPool pooling array of v1.Output [64, 128, 256, ..., 65536]
+	responsesPool = queue.NewSlicePool[v1.Output](1<<6, 1<<16, 2)
```
vlad-diachenko, line 85: It's easier to understand the number in decimal form. 
```suggestion
	responsesPool = queue.NewSlicePool[v1.Output](64, 65536, 2)
```

pkg/bloomgateway/client.go
vlad-diachenko: yep, I believe we clean up the slice only when we fetch it from the pool, so, we will keep the objects in memory until the next time we work with this slice. counting that we have the slices with different sizes, some slices might be rarely used and they will keep some memory.

vlad-diachenko (approved):
lgtm

pkg/bloomgateway/bloomgateway.go
salvacorts: Happy to change it, but note that we usually declare pools with that notation:
https://github.com/grafana/loki/blob/c405db43179c58e6aef18a8aed8129fb5b83c5a2/pkg/chunkenc/pool.go#L49-L60

pkg/bloomgateway/client.go
salvacorts: Good idea, done.

chaudum (approved):

Issue #11332, opened 2023-11-28
Move release process to GitHub Actions and `release-please`


trevorwhitney:
We are planning to overhaul our release process and move it off of Drone and over to GitHub actions. The new process with leverage [`release-please`](https://github.com/googleapis/release-please), which will require us to start using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). This commit message style will be enforced at PR merge time via a [merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue).

The goals of this project are to:
1. Enable any maintainer to do a release (currently there are only a few of us who know how all the pieces work)
2. Enable more frequent releases, particularly faster turn around on patch releases

trevorwhitney:
This has been done for a while now, whoops.

PR #11145, opened 2023-11-06
Fix minimum Go version for module

xperimental:
**What this PR does / why we need it**:

The main module of Loki already uses features of Go 1.21 (`cmp` module, `context.WithTimeoutCause`), so it can't be built with Go versions lower than 1.21 anymore. The module file should reflect this.

The `go.sum` file was updated automatically by the Go tool `go mod tidy` after updating the version.

**Which issue(s) this PR fixes**:

I have not found an existing issue mentioning this yet.

**Special notes for your reviewer**:

Go 1.21 module files introduce a new `toolchain` attribute. I have set this to the currently available latest version.

**Checklist**
- [x] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)

PR #11333, opened 2023-11-28
feat: add release please config

trevorwhitney:

The new release process is being developed in another repo for now as it enables much faster feedback loops. We are at the stage where we are pulling the Loki repo and building images, so it's time to move the `release-please` config into Loki, and configure it for the various packages we will be releasing.

This PR adds the two basic configs that `release-please` requires. This PR does not change anything about the release process, it just enables us to keep iterating out of band until we're ready to move over.

Re #11332


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/grafana/loki
 
-go 1.20
+go 1.21
```
kavirajk, line 3: wonder if this upgraded necessary for the integrating `prepare-release`? 

Asking because, we may need to update the loki build image and few other places as well in that case.

kavirajk (commented):
changes look good to me 👍 Left a comment about upgrading Go version in this PR.

kavirajk (approved):
Approving to unblock.

go.mod
periklis: Please consider this PR waiting for some time now that bumps the directive too:
- #11145

trevorwhitney: @kavirajk our build does not work on 1.20. We have 1.21 in our build image. The new release process moves everything ti GitHub actions, and so we're relying on the `go.mod` file to setup go in the actions containers the build and tests are running in. Without this bump, the build fails.

PR #11338, opened 2023-11-29
Bloom tokenization performance tweaks

paul1r:

The main work of this PR is to improve the performance of the reassemble() function.  By changing the assignment of the 'cur' variable to not do addition, and incrementing the 'pos' variable separately, the benchmarks improved as follows:

BenchmarkTokens/threeSkip1Chunk/v2-10  	   88864	     13487 ns/op
BenchmarkTokens/threeSkip1ChunkAt2/v2At2-10         	   98607	     12256 ns/op

(This was validated with multiple runs, on multiple size inputs)

The remaining changes in the PR are just to add some benchmarking for PopulateSeriesWithBloom, in case we need to do other comparisons for future improvements.

### DIFF
--- a/pkg/storage/bloom/v1/bloom_tokenizer.go
+++ b/pkg/storage/bloom/v1/bloom_tokenizer.go
@@ -115,10 +115,7 @@ func newMetrics(r prometheus.Registerer, namespace, subsystem string) *metrics {
 }
 
 func clearCache(cache map[string]interface{}) {
-	// TODO: Once we move Loki to Go 1.21, we can just use clear() here.
-	for k := range cache {
-		delete(cache, k)
-	}
+	clear(cache)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/storage/bloom/v1/bloom_tokenizer.go
```diff
@@ -115,6 +115,7 @@ func newMetrics(r prometheus.Registerer, namespace, subsystem string) *metrics {
 }
 
 func clearCache(cache map[string]interface{}) {
+	// TODO: Once we move Loki to Go 1.21, we can just use clear() here.
```
chaudum, line 118: We already require 1.21 in the go.mod

paul1r: argh, it was 1.20 when I made the PR.. thanks

PR #11343, opened 2023-11-30
Extract results cache into new pkg

salvacorts:

This extracts the results cache from `queryrangebase` into its own pkg so we can reuse it in other components such as the bloom-gateway without having to import `queryrangebase`.

- Most of the logic inside `pkg/querier/queryrange/queryrangebase/results_cache.go` now lives in `pkg/storage/chunk/cache/results_cache/cache.go`.  
- Some of the tests in `pkg/querier/queryrange/queryrangebase/results_cache.go` are moved into pkg/storage/chunk/cache/results_cache/cache_test.go.
   - Note that here we don't have access to the types we use in `queryrangebase` so we created a new set of mock request/response types to test with.


pkg/storage/chunk/cache/resultscache/cache.go
```diff
@@ -0,0 +1,467 @@
...
+// extents must be in order by start time.
+func (s ResultsCache) partition(req Request, extents []Extent) ([]Request, []Response, error) {
+	var requests []Request
+	var cachedResponses []Response
+	start := req.GetStart().UnixMilli()
+	end := req.GetEnd().UnixMilli()
+
+	for _, extent := range extents {
+		// If there is no overlap, ignore this extent.
+		if extent.GetEnd() < start || extent.Start > end {
```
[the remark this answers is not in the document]
salvacorts: Note this was copied from the original code. Why would accessing .End be better/different from calling GetEnd()?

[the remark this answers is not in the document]
paul1r: Probably slightly more performant, but it's really just the fact that one of the OR clauses uses the getter, and the other references the variable directly.  I'd vote for one or the other to be consistent, that's all

pkg/storage/chunk/cache/resultscache/interface.go
```diff
@@ -0,0 +1,56 @@
...
+
+type Response interface {
+	proto.Message
+}
+
+// ResponseMerger is used by middlewares making multiple requests to merge back all responses into a single one.
+type ResponseMerger interface {
+	// MergeResponse merges responses from multiple requests into a single Response
+	MergeResponse(...Response) (Response, error)
+}
```
chaudum, line 34: I find only one implementation (`queryMergerAsCacheResponseMerger`) of the ResponseMerger interface. Is this really needed, since it only casts back and forth?

pkg/querier/queryrange/index_stats_cache_test.go
```diff
@@ -79,7 +82,7 @@ func TestIndexStatsCache(t *testing.T) {
 	// should reuse part of the previous request and issue a new request for the remaining time till end.
 	// The new start time is 15m (i.e. 25%) in the future with regard to the previous request time span.
 	*calls = 0
-	req := statsReq.WithStartEnd(statsReq.GetStart().Add(15*time.Minute), statsReq.GetEnd().Add(15*time.Minute))
+	req := statsReq.WithStartEnd(statsReq.GetStart().Add(15*time.Minute), statsReq.GetEnd().Add(15*time.Minute)).(queryrangebase.Request)
```
kavirajk, line 85: this type conversion necessary?

pkg/querier/queryrange/queryrangebase/definitions/interface.go
```diff
@@ -32,28 +32,16 @@ type Merger interface {
...
-	GetEnd() time.Time
-	// GetStep returns the step of the request in milliseconds.
-	GetStep() int64
-	// GetQuery returns the query of the request.
-	GetQuery() string
-	// GetCachingOptions returns the caching options.
-	GetCachingOptions() CachingOptions
-	// WithStartEnd clone the current request with different start and end timestamp.
-	WithStartEnd(start time.Time, end time.Time) Request
+	resultscache.Request
```
kavirajk, line 35: This refactoring bothers me a little for two reasons. 
1. conceptually, this is confusing. What has `resultscache.Request` to do anything with main `Request` interface type used in all the read path for all the queries?

Not sure about coupling these two types this way. It should be the other way around, "results cache using main request interface".

2. Embedding types are comparatively make it hard for readability IMHO. Given this is one of the main interface types.

the way I think about this is, if circular dependencies are causing problems to reuse these types, then better we move these types completely to different package.

Makes more sense for both `queryrange` and `resultscache` package uses exactly same type say `query.Request` instead of coupling them with two different types + embedding. Seems complex to me.

wdyt @salvacorts ?

pkg/storage/chunk/cache/resultscache/cache.go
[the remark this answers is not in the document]
salvacorts: > Probably slightly more performant

I don't think so, I'd expect these kind of methods to be inlined.

> Probably slightly more performant, but it's really just the fact that one of the OR clauses uses the getter, and the other references the variable directly. I'd vote for one or the other to be consistent, that's all

Agree and done! Used `GetStart()`

pkg/querier/queryrange/index_stats_cache_test.go
salvacorts: No longer necessary since separated `queryrangebase.Request/Response` from `resultscache.Request/Response`. Removed.

pkg/querier/queryrange/queryrangebase/definitions/interface.go
salvacorts: (we discussed this in Slack).

I ended up separating the definition of `queryrangebase.Request/Response` from `resultscache.Request/Response`. The return type of the `WithStartEnd` method was being problematic (ambiguous and duplicated definition) So I ended up following the [same pattern as LokiRequest][1]: have `WithStartEnd` to implement `queryrangebase.Request` and `WithStartEndTime` to implement `resultscache.Request`.

[1]: https://github.com/grafana/loki/blob/5d84adfe372997dc09bc9f2425b1cff1998bfbbb/pkg/querier/queryrange/codec.go#L59-L71

pkg/storage/chunk/cache/resultscache/interface.go
```diff
@@ -0,0 +1,56 @@
...
+	// GetEnd returns the end timestamp of the request in milliseconds.
+	GetEnd() time.Time
+	// GetStep returns the step of the request in milliseconds.
+	GetStep() int64
+	// GetQuery returns the query of the request.
+	GetQuery() string
+	// GetCachingOptions returns the caching options.
+	GetCachingOptions() CachingOptions
+	// WithStartEndTime clone the current request with different start and end timestamp.
+	WithStartEndTime(start time.Time, end time.Time) Request
```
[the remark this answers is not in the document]
salvacorts: Done :)

PR #11353, opened 2023-12-01
Wait for all workers to stop when stopping blockdownloader

chaudum:
The `blockDownloader` does not wait until all of its worker goroutines are finished. This leads to a flaky test, see #11347 

To properly fix the behaviour, the `stop()` function should wait until all workers are done.

Closes #11347


vlad-diachenko (approved):
LGTM

PR #11356, opened 2023-12-01
Remove worker callback function that is only used in tests

chaudum:
The queue already exposes the functionality to track connected consumers (workers), so there is no need to have callback that is only used in tests.


vlad-diachenko (approved):
LGTM

PR #11360, opened 2023-12-01
Bloom Gateway: process blocks immediately when they are available

chaudum:
At current point in time, the worker calls `w.store.GetBlockQueriersForBlockRefs()` to obtain a list of block queriers sorted by fingerprint in ascending order.
However, the shipper used by the store to fetch blocks by its refs, calls a callback function when a requested block is available. Instead of waiting for all callbacks to be called and then returning a single list of block queriers, the store now exposes a new function `ForEach()` that also accepts a callback that is passed to the underlying shipper.

This allows to process blocks in the worker as soon as they become available. Some block may be served from cache, some may need to be downloaded. That way, we reduce waiting time.

### DIFF
--- a/pkg/bloomgateway/bloomgateway_test.go
+++ b/pkg/bloomgateway/bloomgateway_test.go
@@ -3,6 +3,7 @@ package bloomgateway
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"os"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


pkg/bloomgateway/bloomgateway_test.go
```diff
@@ -263,9 +264,13 @@ func TestBloomGateway_FilterChunkRefs(t *testing.T) {
 		// replace store implementation and re-initialize workers and sub-services
 		bqs, data := createBlockQueriers(t, 5, now.Add(-8*time.Hour), now, 0, 1024)
 		gw.bloomStore = newMockBloomStore(bqs)
+		// randomize between sequentical processing and processing using callback
+		gw.workerConfig.processBlocksSequentially = rand.Float32() > 0.5
```
paul1r, line 268: Would it be better to have two tests, one that did sequential, and one that used the callback?  I'm only thinking of flaky test cases, and that if there was something strange happening here, it may be hard to pin down.

paul1r (approved):
lgtm

PR #11363, opened 2023-12-02
bugfix(memcached): Make `memcached` batch fetch truely context aware

kavirajk:
Currently, `Fetch` method of memcached is not truely respecting the `ctx` passed to it. It get lost in `fetchKeysBatched` method
https://github.com/grafana/loki/blob/489ac8d529ec39352/pkg/storage/chunk/cache/memcached.go#L171-L226

So even if client cancels the context, memcached would still be fetching keys.

This PR make it respect context cancelation.

Changes:
1. Add select case to check for `ctx.Done()` before submitting work to `inputCh`.
2. Given, now we have two exit point to stop the mecached actor one via old `Stop()` method and other via context cancel, I Added aditional flag `stopped` to make sure we are not closing channels that are already closed.
3. Updated tests to lock this behaviour.

Fixes NA

This bug played a role in some incident recently on our Loki SaaS. cc @slim-bean 

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/storage/chunk/cache/memcached.go
```diff
@@ -171,11 +178,13 @@ func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, b
 func (c *Memcached) fetchKeysBatched(ctx context.Context, keys []string) (found []string, bufs [][]byte, missed []string, err error) {
 	resultsCh := make(chan *result)
 	batchSize := c.cfg.BatchSize
-
 	go func() {
 		for i, j := 0, 0; i < len(keys); i += batchSize {
 			batchKeys := keys[i:math.Min(i+batchSize, len(keys))]
 			select {
+			case <-ctx.Done():
+				c.closeAndStop()
```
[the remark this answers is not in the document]
kavirajk: We start the go routines that waits on `inputCh` [here,](https://github.com/grafana/loki/blob/kavirajk%2Ffix-chunk-fetcher-cancellation/pkg/storage/chunk/cache/memcached.go#L96-L108) even when `NewMemcached` is called. 

We are `ranging` over the inputCh which is nice and closing the inputCha stops the `range` and exit those go routines safely. Without closing `inputCh` ranging would hang for ever.

pkg/storage/chunk/cache/memcached.go
```diff
@@ -171,11 +178,13 @@ func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, b
...
+				return
```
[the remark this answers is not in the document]
kavirajk: Good point about fetcher 👍 Actually thought about sending error back didn't end up doing because need some rewrite to make all those go routines return error and make other main goroutine to listen on some `errCh` correctly.

But seeing how fetcher uses it, it's definitely worth sending error up. Agree. will fix it.

pkg/storage/chunk/cache/memcached.go
```diff
@@ -101,7 +106,12 @@ func NewMemcached(cfg MemcachedConfig, client MemcachedClient, name string, reg
 					batchID: input.batchID,
 				}
 				res.found, res.bufs, res.missed, res.err = c.fetch(input.ctx, input.keys)
-				input.resultCh <- res
+				// NOTE: This check is needed because goutines submitting work via `inputCh` may exit in-between because of context cancellation or timeout. This helps to close these worker goroutines to exit without hanging around.
```
dannykopping, line 109:
```suggestion
				// NOTE: This check is needed because goroutines submitting work via `inputCh` may exit in-between because of context cancellation or timeout. This helps to close these worker goroutines to exit without hanging around.
```

pkg/storage/chunk/cache/memcached_test.go
```diff
@@ -17,44 +18,124 @@ import (
...
+				bufs := [][]byte{[]byte("1"), []byte("2")}
+				err := m.Store(ctx, keys, bufs)
+				require.NoError(t, err)
+				<-wait                            // wait before fetching
+				_, _, _, err = m.Fetch(ctx, keys) // will try to write to `intputChan` and shouldn't panic
+				require.NoError(t, err)
+			})
+		}()
+
+		ctxCancel() // cancel even before single fetch is done.
```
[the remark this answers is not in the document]
kavirajk: Done 👍

pkg/storage/chunk/cache/memcached.go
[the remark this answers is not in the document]
kavirajk: Done 👍

pkg/storage/chunk/cache/memcached.go
```diff
@@ -266,13 +286,10 @@ func (c *Memcached) Stop() {
 // Assumes c.inputCh, c.closed channels are non-nil
 // Go routine safe and idempotent.
 func (c *Memcached) closeAndStop() {
-	c.stopMu.Lock()
-	defer c.stopMu.Unlock()
-
-	if !c.stopped {
+	if !c.stopped.Load() {
```
dannykopping, line 289: I think you should use https://pkg.go.dev/sync/atomic#Bool.CompareAndSwap here for increased safety and protection against races.

pkg/storage/chunk/cache/memcached_test.go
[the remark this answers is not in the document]
dannykopping: Looks good, thanks 👍

pkg/storage/chunk/cache/memcached.go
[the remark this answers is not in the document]
dannykopping: Looks good, thanks 👍

pkg/storage/chunk/cache/memcached.go
```diff
@@ -11,14 +11,20 @@ import (
...
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"go.uber.org/atomic"
 
 	"github.com/grafana/loki/pkg/logqlmodel/stats"
 	"github.com/grafana/loki/pkg/util/constants"
 	"github.com/grafana/loki/pkg/util/math"
 )
 
+var (
+	ErrMemcachedClosed = errors.New("memcached is closed")
```
dannykopping, line 25: `memcached is closed` is not a very helpful error message. Is the client closed? Is the server down?
Please clarify this.

pkg/storage/chunk/cache/memcached.go
```diff
@@ -222,6 +248,10 @@ func (c *Memcached) fetchKeysBatched(ctx context.Context, keys []string) (found
 		}
 	}
 
+	if workerErr != nil {
```
dannykopping, line 251: This code block is not covered by a test

pkg/storage/chunk/cache/memcached.go
```diff
@@ -94,7 +106,12 @@ func NewMemcached(cfg MemcachedConfig, client MemcachedClient, name string, reg
 					batchID: input.batchID,
 				}
 				res.found, res.bufs, res.missed, res.err = c.fetch(input.ctx, input.keys)
-				input.resultCh <- res
+				// NOTE: This check is needed because goutines submitting work via `inputCh` may exit in-between because of context cancellation or timeout. This helps to close these worker goroutines to exit without hanging around.
+				select {
+				case <-c.closed:
+					return
```
dannykopping, line 112: This is not covered by a test, is it worth adding one?

dannykopping (commented):
Almost there 👍

pkg/storage/chunk/cache/memcached.go
kavirajk: This is the error returned when context cancellation happened in this method.

And should be captured in these test asserts.

```
				require.ErrorIs(t, err, context.Canceled)
```

and

```
				require.ErrorIs(t, err, context.DeadlineExceeded)
```

dannykopping: I ran the tests with coverage and it showed this code block was not covered

pkg/storage/chunk/cache/memcached.go
kavirajk: We are actually in-directly testing this, as without this, some of the worker go-routine may hang on `resultsCh` so that not all `wg.Done()` is run so `wg.Wait()` on the `Stop()` method will not finish.

We in-directly testing this by explicitly calling `m.Stop()` after every sub tests.

Hard to test this specific line deterministically in test if that's what you mean also tbh not necessary I think (testing `Stop()` is good enough IMHO)

pkg/storage/chunk/cache/memcached.go
kavirajk: Also double checked by removing those lines (these asserts would fail in that case)

kavirajk: Ah. I think I know why. I accidentially removed one of the return when it's closed. Caught in another test case. Fixing it.

pkg/storage/chunk/cache/memcached.go
kavirajk: Done.

pkg/storage/chunk/cache/memcached.go
kavirajk: Done

pkg/storage/chunk/cache/memcached.go
kavirajk: I'm bit confused to use `CompareAndSwap` for bool here.

The expected behaviour I want is I want some channels to be closed after Loading and before Swapping. which is not possible with single `CompareAndSwap`?

If I use `CompareAndSwap()` instead of `Load` here, then bool may be already updated even before closing this channels. 

The more I think about it, mutex instead of atomic.Bool would be right type here. wdyt?

pkg/storage/chunk/cache/memcached.go
```diff
@@ -22,7 +22,7 @@ import (
 )
 
 var (
-	ErrMemcachedClosed = errors.New("memcached is closed")
+	ErrMemcachedClosedByClient = errors.New("memcached is closed by client")
```
dannykopping, line 25: This doesn't make sense; do you mean a memcached _connection_ is closed by the client?
```suggestion
	ErrMemcachedConnClosed = errors.New("memcached connection is closed")
```

pkg/storage/chunk/cache/memcached.go
```diff
@@ -63,6 +63,9 @@ type Memcached struct {
 	stopped atomic.Bool
 
 	logger log.Logger
+
+	// NOTE: testFetchDelay should be used only for testing. See `SetTestFetchDelay()` method for more details.
+	testFetchDelay chan struct{}
```
dannykopping, line 68: The idiomatic approach here would be to define a mock client or wrap this client and shadow the function where the delay is required.

If there's absolutely no other way to implement this without refactoring a lot then I'll approve, but this feels like a hack in its current form.

pkg/storage/chunk/cache/memcached.go
dannykopping: Yeah you're right.
The more I look at this, the more I think this should be a `sync.Once`, actually.

pkg/storage/chunk/cache/memcached.go
kavirajk: > If there's absolutely no other way to implement this without refactoring a lot then I'll approve, but this feels like a hack in its current form.

Looks like it. I spent some time to clean it up. Tried several ways with mock as well. But there are some major blocks 

1. we use `cache_test` intead of `cache` (original package name) in all the tests to test these caches as external clients, which prevent us to use something internal for testing.

2. I tried to make it just `cache` touching tons of files, even after doing that lots of cyclic dependencies break this particular package. Then realized that's why they introduced this testing via `cache_test` in the first place.

3. Another core problem is in general with this package is how we do parallelism with Go routines. Ideally, only sender goroutine of the channel should be closing it as well (e.g: inputCh, resultsCh, etc). But here we are doing at random places. This prevents us to do some idiomatic tests deterministically. I tried to clean it up but ended up rewriting a lots of code, seems overkill for this PR. Better to do it separately with more testing.

dannykopping: OK, are you planning to follow up this PR with the refactor?

pkg/storage/chunk/cache/memcached.go
kavirajk: I don't want to confuse this with `connection` here. Because we just closing the `Memcached` struct created by client. So that it can no longer be used for storing or fetching.

May be `memcached` word is confusing. Going to change to `cache is stopped by client` instead. The actual message is client called `Stop()` method deliberately.

dannykopping: Makes sense 👍

pkg/storage/chunk/cache/memcached.go
kavirajk: yes :)

pkg/storage/chunk/cache/memcached.go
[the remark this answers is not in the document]
slim-bean: I think ashwanth is correct here, I don't think we can close this channel, it will never be re-opened?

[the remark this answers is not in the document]
kavirajk: So true. Pity I didn't pay close attention. Looking into it.

PR #11380, opened 2023-12-04
Bloom-Gateway cache

salvacorts:

This PR adds caching to the bloom-gateway client. It uses the result cache from https://github.com/grafana/loki/pull/11343.

Here's how we:
- Merge responses: group all chunks by FP and remove duplicated chunk checksums.
- Extract responses based on time span: For all chunks in each FP, add to the extracted response only the chunks that overlaps with the desired start and end time.

### DIFF
--- a/docs/sources/configure/_index.md
+++ b/docs/sources/configure/_index.md
@@ -1838,6 +1838,18 @@ client:
   # CLI flag: -bloom-gateway-client.log-gateway-requests
   [log_gateway_requests: <boolean> | default = false]
 
+  cache:
+    # The cache block configures the cache backend.
+    # The CLI flags prefix for this block configuration is:
+    # store.index-cache-read
+    [cache: <cache_config>]
+
+    [compression: <string> | default = ""]
+
+    [parallelism: <int>]
+
+  [cache_results: <boolean>]
+
 # Number of workers to use for filtering chunks concurrently.
 # CLI flag: -bloom-gateway.worker-concurrency
 [worker_concurrency: <int> | default = 4]
@@ -3027,6 +3039,10 @@ shard_streams:
 # Maximum number of blocks will be downloaded in parallel by the Bloom Gateway.
 # CLI flag: -bloom-gateway.blocks-downloading-parallelism
 [bloom_gateway_blocks_downloading_parallelism: <int> | default = 50]
+
+# Interval for computing the cache key in the Bloom Gateway.
+# CLI flag: -bloom-gateway.cache-key-interval
+[bloom_gateway_cache_key_interval: <duration> | default = 15m]
 
 # Allow user to send structured metadata in push payload.
 # CLI flag: -validation.allow-structured-metadata
--- a/docs/sources/configure/_index.md
+++ b/docs/sources/configure/_index.md
@@ -1849,7 +1849,9 @@ client:
     # CLI flag: -bloom-gateway-client.cache.compression
     [compression: <string> | default = ""]
 
-  [cache_results: <boolean>]
+  # Flag to control whether to cache bloom gateway client requests/responses.
+  # CLI flag: -bloom-gateway-client.cache_results
+  [cache_results: <boolean> | default = false]
 
 # Number of workers to use for filtering chunks concurrently.
 # CLI flag: -bloom-gateway.worker-concurrency
--- a/docs/sources/configure/_index.md
+++ b/docs/sources/configure/_index.md
@@ -1838,12 +1838,15 @@ client:
   # CLI flag: -bloom-gateway-client.log-gateway-requests
   [log_gateway_requests: <boolean> | default = false]
 
-  cache:
+  results_cache:
     # The cache block configures the cache backend.
     # The CLI flags prefix for this block configuration is:
-    # store.index-cache-read
+    # bloom-gateway-client.cache
     [cache: <cache_config>]
 
+    # Use compression in cache. The default is an empty value '', which disables
+    # compression. Supported values are: 'snappy' and ''.
+    # CLI flag: -bloom-gateway-client.cache.compression
     [compression: <string> | default = ""]
 
     [parallelism: <int>]
@@ -4249,6 +4252,7 @@ The TLS configuration.
 
 The cache block configures the cache backend. The supported CLI flags `<prefix>` used to reference this configuration block are:
 
+- `bloom-gateway-client.cache`
 - `frontend`
 - `frontend.index-stats-results-cache`
 - `frontend.volume-results-cache`
--- a/docs/sources/configure/_index.md
+++ b/docs/sources/configure/_index.md
@@ -1849,8 +1849,6 @@ client:
     # CLI flag: -bloom-gateway-client.cache.compression
     [compression: <string> | default = ""]
 
-    [parallelism: <int>]
-
   [cache_results: <boolean>]
 
 # Number of workers to use for filtering chunks concurrently.
--- a/pkg/bloomgateway/cache.go
+++ b/pkg/bloomgateway/cache.go
@@ -146,8 +146,9 @@ func NewBloomGatewayClientCacheMiddleware(
 	cacheGen resultscache.CacheGenNumberLoader,
 	retentionEnabled bool,
 ) *ClientCache {
-	nextAsHandler := resultscache.HandlerFunc(func(ctx context.Context, req resultscache.Request) (resultscache.Response, error) {
-		return next.FilterChunkRefs(ctx, req.(*logproto.FilterChunkRefRequest))
+	nextAsHandler := resultscache.HandlerFunc(func(ctx context.Context, cacheReq resultscache.Request) (resultscache.Response, error) {
+		req := cacheReq.(requestWithGrpcCallOptions)
+		return next.FilterChunkRefs(ctx, req.FilterChunkRefRequest, req.grpcCallOptions...)
 	})
 
 	resultsCache := resultscache.NewResultsCache(
@@ -174,11 +175,27 @@ func NewBloomGatewayClientCacheMiddleware(
 	}
 }
 
-func (c *ClientCache) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunkRefRequest, _ ...grpc.CallOption) (*logproto.FilterChunkRefResponse, error) {
-	res, err := c.cache.Do(ctx, req)
+func (c *ClientCache) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunkRefRequest, opts ...grpc.CallOption) (*logproto.FilterChunkRefResponse, error) {
+	cacheReq := requestWithGrpcCallOptions{
+		FilterChunkRefRequest: req,
+		grpcCallOptions:       opts,
+	}
+	res, err := c.cache.Do(ctx, cacheReq)
 	if err != nil {
 		return nil, err
 	}
 
 	return res.(*logproto.FilterChunkRefResponse), nil
+}
+
+type requestWithGrpcCallOptions struct {
+	*logproto.FilterChunkRefRequest
+	grpcCallOptions []grpc.CallOption
+}
+
+func (r requestWithGrpcCallOptions) WithStartEndForCache(start time.Time, end time.Time) resultscache.Request {
+	return requestWithGrpcCallOptions{
+		FilterChunkRefRequest: r.FilterChunkRefRequest.WithStartEndForCache(start, end).(*logproto.FilterChunkRefRequest),
+		grpcCallOptions:       r.grpcCallOptions,
+	}
 }
--- a/pkg/bloomgateway/cache.go
+++ b/pkg/bloomgateway/cache.go
@@ -1,0 +1,177 @@
+package bloomgateway
+
+import (
+	"context"
+	"flag"
+	"sort"
+
+	"github.com/go-kit/log"
+	"github.com/prometheus/common/model"
+	"golang.org/x/exp/slices"
+	"google.golang.org/grpc"
+
+	"github.com/grafana/loki/pkg/logproto"
+	"github.com/grafana/loki/pkg/storage/chunk/cache"
+	"github.com/grafana/loki/pkg/storage/chunk/cache/resultscache"
+)
+
+type CacheConfig struct {
+	resultscache.Config `yaml:",inline"`
+	Parallelism         int `yaml:"parallelism"`
+}
+
+// RegisterFlags registers flags.
+func (cfg *CacheConfig) RegisterFlags(f *flag.FlagSet) {
+	cfg.RegisterFlagsWithPrefix(f, "bloom-gateway.cache.")
+	f.IntVar(&cfg.Parallelism, "bloom-gateway.cache.parallelism", 1, "Maximum number fo concurrent requests to gateway server.")
+}
+
+type keyGen struct {
+	Limits
+}
+
+func NewCacheKeyGen(limits Limits) resultscache.KeyGenerator {
+	return keyGen{limits}
+}
+
+func (k keyGen) GenerateCacheKey(ctx context.Context, userID string, r resultscache.Request) string {
+	return resultscache.ConstSplitter(k.BloomGatewayCacheKeyInterval(userID)).GenerateCacheKey(ctx, userID, r)
+}
+
+type extractor struct{}
+
+func NewExtractor() resultscache.Extractor {
+	return extractor{}
+}
+
+// Extract extracts a subset of a response from the `start` and `end` timestamps in milliseconds.
+// We remove chunks that are not within the given time range.
+func (e extractor) Extract(start, end int64, r resultscache.Response, _, _ int64) resultscache.Response {
+	res := r.(*logproto.FilterChunkRefResponse)
+
+	chunkRefs := make([]*logproto.GroupedChunkRefs, 0, len(res.ChunkRefs))
+	for _, chunkRef := range res.ChunkRefs {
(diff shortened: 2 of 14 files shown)


pkg/bloomgateway/cache.go
```diff
@@ -0,0 +1,177 @@
...
+
+	chunkRefs := make([]*logproto.GroupedChunkRefs, 0, len(res.ChunkRefs))
+	for _, chunkRef := range res.ChunkRefs {
+		refs := make([]*logproto.ShortRef, 0, len(chunkRef.Refs))
+		for _, ref := range chunkRef.Refs {
+			if model.Time(end) < ref.From || ref.Through <= model.Time(start) {
+				continue
+			}
+			refs = append(refs, ref)
+		}
```
chaudum, line 60: I wonder if we can reduce the mount of iterations by sorting `chunkRef.Refs` by `From` date and binary search the first item where `From > end` .
Then iterate only over `chunkRef.Refs[idx:]`.

pkg/bloomgateway/cache.go
```diff
@@ -0,0 +1,177 @@
...
+	"google.golang.org/grpc"
+
+	"github.com/grafana/loki/pkg/logproto"
+	"github.com/grafana/loki/pkg/storage/chunk/cache"
+	"github.com/grafana/loki/pkg/storage/chunk/cache/resultscache"
+)
+
+type CacheConfig struct {
+	resultscache.Config `yaml:",inline"`
+	Parallelism         int `yaml:"parallelism"`
```
chaudum, line 20: I would omit this setting, since it's rather confusing. I don't see a need to parallelise requests to the bloom gateway.

pkg/bloomgateway/cache.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// RegisterFlags registers flags.
+func (cfg *CacheConfig) RegisterFlags(f *flag.FlagSet) {
+	cfg.RegisterFlagsWithPrefix(f, "bloom-gateway.cache.")
+	f.IntVar(&cfg.Parallelism, "bloom-gateway.cache.parallelism", 1, "Maximum number fo concurrent requests to gateway server.")
+}
+
+type keyGen struct {
+	Limits
```
chaudum, line 30: I would accept a more limited interface that has only the necessary method for the cache

```suggestion
	CacheLimits
```

pkg/bloomgateway/cache.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func NewCacheKeyGen(limits Limits) resultscache.KeyGenerator {
+	return keyGen{limits}
+}
+
+func (k keyGen) GenerateCacheKey(ctx context.Context, userID string, r resultscache.Request) string {
```
chaudum, line 37: We use `tenant` throughout the bloomgateway and bloomcompactor code:

```suggestion
func (k keyGen) GenerateCacheKey(ctx context.Context, tenant string, r resultscache.Request) string {
```

pkg/bloomgateway/cache.go
```diff
@@ -0,0 +1,177 @@
...
+	return resultscache.ConstSplitter(k.BloomGatewayCacheKeyInterval(userID)).GenerateCacheKey(ctx, userID, r)
+}
+
+type extractor struct{}
+
+func NewExtractor() resultscache.Extractor {
```
chaudum, line 43: Please return concrete types. The consumer accepts an interface and so enforces the implementation of the interface on this type.

```suggestion
func NewExtractor() extractor {
```

pkg/bloomgateway/cache.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Merge duplicated fingerprints by:
+// 1. Sort the chunkRefs by their stream fingerprint
+// 2. Append all chunks with the same fingerprint into the first fingerprint's chunk list.
+func mergeGroupedChunkRefs(chunkRefs []*logproto.GroupedChunkRefs) []*logproto.GroupedChunkRefs {
+	sort.Slice(chunkRefs, func(i, j int) bool {
+		return chunkRefs[i].Fingerprint < chunkRefs[j].Fingerprint
+	})
+	return slices.CompactFunc(chunkRefs, func(next, prev *logproto.GroupedChunkRefs) bool {
```
chaudum, line 108: TIL: There is a `slices.CompactFunc` :)

pkg/bloomgateway/cache.go
```diff
@@ -0,0 +1,177 @@
...
+func mergeShortRefs(refs []*logproto.ShortRef) []*logproto.ShortRef {
+	sort.Slice(refs, func(i, j int) bool {
+		return refs[i].Checksum < refs[j].Checksum
+	})
+	return slices.CompactFunc(refs, func(a, b *logproto.ShortRef) bool {
+		return a.Checksum == b.Checksum
+	})
+}
+
+type bloomQuerierCache struct {
```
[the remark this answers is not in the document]
salvacorts: It is! I started using that name because I was putting the cache in the querier instead of on the client level. Renamed.

pkg/bloomgateway/cache.go
```diff
@@ -0,0 +1,177 @@
...
+
+func NewBloomGatewayClientCacheMiddleware(
+	cfg CacheConfig,
+	logger log.Logger,
+	next logproto.BloomGatewayClient,
+	c cache.Cache,
+	limits Limits,
+	cacheGen resultscache.CacheGenNumberLoader,
+	retentionEnabled bool,
+) logproto.BloomGatewayClient {
```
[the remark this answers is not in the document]
salvacorts: Thank you! Changed.

pkg/bloomgateway/client.go
```diff
@@ -124,7 +131,15 @@ type GatewayClient struct {
...
 
-func NewGatewayClient(cfg ClientConfig, limits Limits, registerer prometheus.Registerer, logger log.Logger, metricsNamespace string) (*GatewayClient, error) {
+func NewGatewayClient(
+	cfg ClientConfig,
+	limits Limits,
+	registerer prometheus.Registerer,
+	logger log.Logger,
+	metricsNamespace string,
+	cacheGen resultscache.CacheGenNumberLoader,
+	retentionEnabled bool,
```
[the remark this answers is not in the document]
salvacorts: Yes, it's used to indicate that the flag may be invalidated. It works with the cache gen number generator.
https://github.com/grafana/loki/blob/957ce119d2090739391323c99437cdab0006c3b7/pkg/storage/chunk/cache/resultscache/cache.go#L112-L114

pkg/bloomgateway/cache.go
salvacorts: I don't think that will save us any iterations: Sorting + bin search is `O(n*logn + logn)`, whereas here we are only iterating once so `O(n)`.

pkg/bloomgateway/cache.go
salvacorts: Removed

pkg/bloomgateway/cache.go
salvacorts: Done

pkg/bloomgateway/cache.go
salvacorts: Done

pkg/bloomgateway/cache.go
salvacorts: Done

pkg/bloomgateway/cache.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	return &logproto.FilterChunkRefResponse{
+		ChunkRefs: chunkRefs,
+	}
+}
+
+type merger struct{}
+
+func NewMerger() resultscache.ResponseMerger {
```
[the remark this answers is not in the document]
salvacorts: Done

pkg/bloomgateway/cache.go
salvacorts: Yes! It's quite useful to remove / aggregate duplicated items in a sorted list

pkg/bloomgateway/cache.go
```diff
@@ -0,0 +1,177 @@
...
+
+type CacheConfig struct {
+	resultscache.Config `yaml:",inline"`
+	Parallelism         int `yaml:"parallelism"`
+}
+
+// RegisterFlags registers flags.
+func (cfg *CacheConfig) RegisterFlags(f *flag.FlagSet) {
+	cfg.RegisterFlagsWithPrefix(f, "bloom-gateway.cache.")
+	f.IntVar(&cfg.Parallelism, "bloom-gateway.cache.parallelism", 1, "Maximum number fo concurrent requests to gateway server.")
```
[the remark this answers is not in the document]
salvacorts: Removed

docs/sources/configure/_index.md
```diff
@@ -1838,6 +1838,18 @@ client:
...
+    # The cache block configures the cache backend.
+    # The CLI flags prefix for this block configuration is:
+    # store.index-cache-read
+    [cache: <cache_config>]
+
+    [compression: <string> | default = ""]
+
+    [parallelism: <int>]
+
+  [cache_results: <boolean>]
```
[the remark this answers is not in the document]
salvacorts: The docs were missing due to the errors reported by CI. Should be fixed now.

pkg/bloomgateway/cache.go
```diff
@@ -0,0 +1,177 @@
...
+	)
+
+	return &bloomQuerierCache{
+		cache:  resultsCache,
+		limits: limits,
+		logger: logger,
+	}
+}
+
+func (c *bloomQuerierCache) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunkRefRequest, _ ...grpc.CallOption) (*logproto.FilterChunkRefResponse, error) {
```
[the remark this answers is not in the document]
salvacorts: Fixed

pkg/logproto/compat.go
```diff
@@ -334,3 +337,71 @@ func (m *VolumeRequest) LogToSpan(sp opentracing.Span) {
...
+}
+
+// GetCachingOptions returns the caching options.
+func (m *FilterChunkRefRequest) GetCachingOptions() (res resultscache.CachingOptions) { return }
+
+// WithStartEndForCache implements resultscache.Request.
+func (m *FilterChunkRefRequest) WithStartEndForCache(start, end time.Time) resultscache.Request {
+	clone := *m
+	clone.From = model.TimeFromUnixNano(start.UnixNano())
+	clone.Through = model.TimeFromUnixNano(end.UnixNano())
```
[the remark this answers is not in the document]
salvacorts: Good call, fixed.

pkg/bloomgateway/client.go
```diff
@@ -138,22 +154,47 @@ func NewGatewayClient(cfg ClientConfig, limits Limits, registerer prometheus.Reg
 		return nil, err
 	}
 
+	var c cache.Cache
+	if cfg.CacheResults {
+		if !cache.IsCacheConfigured(cfg.Cache.CacheConfig) {
+			return nil, errors.New("cache is not configured")
+		}
```
vlad-diachenko, line 161: it should be validated by the config.

pkg/logproto/compat.go
```diff
@@ -334,3 +337,71 @@ func (m *VolumeRequest) LogToSpan(sp opentracing.Span) {
...
+}
+
+// GetStep returns the step of the request in milliseconds. Always 0.
+func (m *FilterChunkRefRequest) GetStep() int64 {
+	return 0
+}
+
+// GetQuery returns the query of the request.
+// The query is the hash for the input chunks refs and the filter expressions.
+func (m *FilterChunkRefRequest) GetQuery() string {
```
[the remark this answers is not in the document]
salvacorts: Yes, it's used to compute the key. I agree it's probably too expensive.

pkg/logproto/compat.go
```diff
@@ -334,3 +337,71 @@ func (m *VolumeRequest) LogToSpan(sp opentracing.Span) {
...
+		s.WriteObjectStart()
+		s.WriteObjectField("op")
+		s.WriteInt64(filter.Operator)
+		s.WriteMore()
+		s.WriteObjectField("match")
+		s.WriteString(filter.Match)
+		s.WriteObjectEnd()
+	}
+	s.WriteArrayEnd()
+	_ = s.Flush()
```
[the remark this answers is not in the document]
salvacorts: I was thinking about debuggability to have cache keys that we could easily identify. But probably not?
Having said that, I don't think this is too expensive compared to https://github.com/grafana/loki/pull/11380#discussion_r1415010740.

maybe we can just create a string like: `op-filter,op-filter`; e.g. `0-1a23h4g,1-UserName,0-textILookFor`

pkg/logproto/compat.go
```diff
@@ -334,3 +337,71 @@ func (m *VolumeRequest) LogToSpan(sp opentracing.Span) {
...
+}
+
+// GetQuery returns the query of the request.
+// The query is the hash for the input chunks refs and the filter expressions.
+func (m *FilterChunkRefRequest) GetQuery() string {
+	var chunksHash uint64
+	if len(m.Refs) > 0 {
+		h := xxhash.New()
+		for _, ref := range m.Refs {
+			_, _ = h.WriteString(fmt.Sprintf("%s-%d", ref.Tenant, ref.Fingerprint))
```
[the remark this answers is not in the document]
salvacorts: I agree. I thought about:
1. Call `WriteString` for all seen tenant, and
2. Summing up all the fingerprints
Then compute the hash of it but: 
1. We need to allocate a map for that (should be fine comparatively?)
2. There are other combinations of FPs that can lead to the same sum.
   - Given that the cache key also has the time span and the filter exr, I guess the probability is low enough for this to make sense?

pkg/logproto/compat.go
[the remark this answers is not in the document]
chaudum: that sounds good to me

pkg/logproto/compat.go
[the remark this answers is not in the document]
chaudum: Requests only contain a single tenant, so there is no need to "encode" the tenant multiple times.

vlad-diachenko (commented):
reviewed partially and left some comments, I believe Chirstian has more context here, so completely rely on his review

pkg/bloomgateway/client.go
salvacorts: Done :)

pkg/bloomgateway/cache.go
chaudum: Right. Ideally, chunkRef.Refs was already sorted, but I cannot guarantee that in the bloom gateway

pkg/logproto/compat.go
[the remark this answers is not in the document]
salvacorts: I'm still worried that the summing all chunks FPs may create collisions. wdyt?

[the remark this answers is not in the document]
chaudum: > I'm still worried that the summing all chunks FPs may create collisions

That can always be a problem when using a hash as a key ...

> summing too many FPs may overflow the integer.

Not sure what you mean. You can append the `[]byte` representation of the fingerprint?

pkg/bloomgateway/client.go
```diff
@@ -109,9 +116,25 @@ func (i *ClientConfig) RegisterFlags(f *flag.FlagSet) {
 // RegisterFlagsWithPrefix registers flags for the Bloom Gateway client configuration with a common prefix.
 func (i *ClientConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	i.GRPCClientConfig.RegisterFlagsWithPrefix(prefix+"grpc", f)
+	i.Cache.RegisterFlagsWithPrefix(f, prefix+"cache.")
```
chaudum, line 119: This is a bit of a trap:

You call `RegisterFlagsWithPrefix` on the inlined object `resultscache.Config`. Therefore not all flags are registered.

pkg/logproto/compat.go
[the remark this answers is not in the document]
salvacorts: I just pushed a commit that reduces allocations by reusing a buffer to be used along with `fmt.Appendf`

pkg/bloomgateway/client.go
salvacorts: Fixed by implementing `RegisterFlagsWithPrefix` in `CacheConfig`

pkg/logproto/compat.go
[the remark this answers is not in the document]
salvacorts: Done

pkg/logproto/compat.go
```diff
@@ -334,3 +335,85 @@ func (m *VolumeRequest) LogToSpan(sp opentracing.Span) {
...
+// GetQuery returns the query of the request.
+// The query is the hash for the input chunks refs and the filter expressions.
+func (m *FilterChunkRefRequest) GetQuery() string {
+	var encodeBuf []byte
+	var chunksHash uint64
+	if len(m.Refs) > 0 {
+		h := xxhash.New()
+		for _, ref := range m.Refs {
+			encodeBuf = encodeBuf[:0]
+			_, _ = h.Write(fmt.Appendf(encodeBuf, "%d", ref.Fingerprint))
```
chaudum, line 365: Alternative that does not require string->[]byte conversions:

```suggestion
			_, _ = h.Write(binary.LittleEndian.AppendUint64(nil, ref.Fingerprint))
		
```

paul1r (approved):

vlad-diachenko (approved):
lgtm

PR #11400, opened 2023-12-06
Add alt text to images

JStickler:
Adding alt text to images that were flagged in this week's 404 and build error report.


MichelHollands (approved):
LGTM

PR #11407, opened 2023-12-07
Update out_grafana_loki.go to handle timestamp slice

dvpost:
If you, like me, have encountered the error "level=warn caller=out_grafana_loki.go:135 id=0 msg='timestamp isn't known format. Use current time.'" while using Fluent Bit ~ version 2.2.0 along with the fluent-bit-go-loki plugin ~ version 2.9.2, this fix might be beneficial for you.

This commit updates the out_grafana_loki.go file to handle cases where the timestamp is received as a slice of interfaces ([]interface{}). The update includes a check and extraction process for timestamps provided as a slice, ensuring proper recognition and processing of the time data.

This fix is particularly relevant for users who are experiencing issues with timestamp formatting in logs sent to Loki from Fluent Bit using the fluent-bit-go-loki plugin. It aims to provide a more reliable logging data flow and resolve the timestamp format issues that have been observed in these versions.

**What this PR does / why we need it**: fix "timestamp isn't known format. Use current time."

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/clients/cmd/fluent-bit/out_grafana_loki.go
+++ b/clients/cmd/fluent-bit/out_grafana_loki.go
@@ -126,6 +126,9 @@ func FLBPluginFlushCtx(ctx, data unsafe.Pointer, length C.int, _ *C.char) int {
 
 		// Get timestamp
 		var timestamp time.Time
+		if tsSlice, ok := ts.([]interface{}); ok && len(tsSlice) > 0 {
+			ts = tsSlice[0]
+		}
 		switch t := ts.(type) {
 		case output.FLBTime:
 			timestamp = ts.(output.FLBTime).Time


cstyan:
Some examples of when/why you would receive multiple timestamps plus some tests would be useful here.

Honestly, with the limited knowledge I have here, it feels more like the problem is the fact that we can receive multiple timestamps not that we don't handle the receiving of multiple timestamps.

leo198706:
> Some examples of when/why you would receive multiple timestamps plus some tests would be useful here.
> 
> Honestly, with the limited knowledge I have here, it feels more like the problem is the fact that we can receive multiple timestamps not that we don't handle the receiving of multiple timestamps.

I also encountered the same problem, using fluent-bit 2.2.3 to collect logs and send them to loki, "timestamp isn't known format. Use current time."

After looking through the documentation of fluent-bit, I found that the fluent-bit event format has changed, and I think this is the reason. Reference documentation: https://docs.fluentbit.io/manual/v/dev-2.2/concepts/key-concepts#event-format

Current event format: [[TIMESTAMP, METADATA], MESSAGE]
Prior to v2.1.0 used: [TIMESTAMP, MESSAGE]

Is adjusting the output of GetRecord a better approach:
https://github.com/grafana/loki/blob/f6529c293e3281ba1c19ff36d6018c0ef338f3e6/clients/cmd/fluent-bit/out_grafana_loki.go#L122

chaudum:
The original bug report for this "fix" was closed ("Looks like nothing's wrong. I packaged the plugin with the following dockerfile and it works fine on aws eks"), so closing this PR as well.

PR #9409, opened 2023-05-05
[chore]docs; update admonition syntax

MattDodsonEnglish:
Following the conventions in https://grafana.com/docs/writers-toolkit/style-guide/style-conventions/#admonitions

Part of ongoing chore to standardize admonitions across docs

PR #11408, opened 2023-12-07
Update notes to use shortcode

JStickler:
Updates Notes to use the Hugo admonition shortcode.

Replaces PR #9409 which has gone stale.


MichelHollands (approved):
LGTM

Issue #11413, opened 2023-12-08
No default redis port.


nstankov-bg:
### Describe the Bug
When initializing Loki in a Docker Swarm on-premise infrastructure, the system fails to set up the Redis client due to an error in deriving endpoints. Specifically, the error occurs when a port is not specified in the address for the Loki Redis query result cache, resulting in a failure in splitting the host and port. 

### To Reproduce
1. Start Loki (version 2.8.3, SHA: 0d81144cf)
2. Start Promtail to tail specific logs.
3. Perform a query.
4. Encounter the error related to the endpoint derivation.

### Expected Behavior
The expected behavior is that when a port is not specified in the endpoint, a default port should be assumed or derived without causing an error.

### Environment
- Infrastructure: Docker-Swarm on-premise
- Deployment tool: docker-swarm

### Screenshots or Terminal Output
![Screenshot of Issue](https://github.com/grafana/loki/assets/27363885/65a8fb6d-fcef-4d76-9235-74edf990c971)

Terminal output error message:
```md
redis client setup failed: failed to derive endpoints: splitting host:port failed :address loki_redis_query_result_cache: missing port in address

error initialising module: query-frontend-tripperware

github.com/grafana/dskit/modules.(*Manager).initModule

	/src/loki/vendor/github.com/grafana/dskit/modules/modules.go:122

github.com/grafana/dskit/modules.(*Manager).InitModuleServices

	/src/loki/vendor/github.com/grafana/dskit/modules/modules.go:92

github.com/grafana/loki/pkg/loki.(*Loki).Run

	/src/loki/pkg/loki/loki.go:458

������

n

	/src/loki/cmd/loki/main.go:110

������
runtime.main

	/usr/local/go/src/runtime/proc.go:250

runtime.goexit

	/usr/local/go/src/runtime/asm_amd64.s:1598

level=warn ts=2023-12-08T15:01:05.176258628Z caller=store.go:51 msg="running with DEPRECATED flag -store.max-look-back-period, use -querier.max-query-lookback instead."

level=warn ts=2023-12-08T15:01:05.178666999Z caller=loki.go:286 msg="per-tenant timeout not configured, using default engine timeout (\"5m0s\"). This behavior will change in the next major to always use the default per-tenant timeout (\"5m\")."

level=info ts=2023-12-08T15:01:05.180732638Z caller=main.go:108 msg="Starting Loki" version="(version=2.8.3, branch=HEAD, revision=0d81144cf)"

[...]

PR #11414, opened 2023-12-08
Handle Missing Redis Ports by Defaulting to Standard Redis Port in Loki

nstankov-bg:
**Summary of Changes**:
- Modified the `deriveEndpoints` function in `pkg/storage/chunk/cache/redis_client.go` to handle cases where the Redis port is not specified in the Loki configuration.
- Implemented a fallback mechanism that sets the default Redis port (6379) when the port is absent in the endpoint address.
- This change addresses a bug encountered when initializing Loki in Docker Swarm environments, specifically in setting up the Redis client for the query result cache.

**Reason for Changes**:
- The bug was identified when Loki failed to initialize properly due to an error in deriving endpoints for the Redis client. This occurred when the port was not specified in the endpoint address, leading to an error in splitting host and port.
- The implemented fix ensures that a default port is used in such scenarios, preventing initialization failures and enhancing the robustness of the endpoint derivation process.

**Impact on the Project**:
- These changes directly impact the initialization process of Loki, particularly in environments where explicit port specification might be omitted.
- By defaulting to the standard Redis port, the changes ensure smoother and more reliable deployment of Loki in various infrastructures, especially in Docker Swarm setups.

**Additional Notes for Reviewers**:
- Reviewers are encouraged to verify the default port setting aligns with standard Redis configurations and does not interfere with custom setups.
- It is important to assess if this change might affect any existing deployments where specific port configurations are expected.

**Related Issue**: 
- This PR addresses the bug described in the GitHub issue titled "Redis Client Setup Fails in Loki When Port is Missing in Address". [Issue](https://github.com/grafana/loki/issues/11413)

---

**Checklist for Review and Merge**:
- [x] Ensure changes align with project standards and requirements.
- [x] Verify that the solution adequately addresses the reported issue.
- [x] Assess potential impacts on other parts of the codebase.
- [x] Review for coding best practices and performance implications.
- [ ] Ensure adequate test coverage for the new logic.
- [x] Update relevant documentation to reflect these changes.
- [x] Check for any necessary updates in configuration files or upgrade guides.

### DIFF
--- a/docs/sources/setup/install/helm/install-monolithic/_index.md
+++ b/docs/sources/setup/install/helm/install-monolithic/_index.md
@@ -47,7 +47,7 @@ If you set the `singleBinary.replicas` value to 2 or more, this chart configures
           type: 'filesystem'
         schemaConfig:
           configs:
-          - from: 2024-01-01
+          - from: "2024-01-01"
             store: tsdb
             index:
               prefix: loki_index_
@@ -72,7 +72,7 @@ If you set the `singleBinary.replicas` value to 2 or more, this chart configures
           replication_factor: 3
         schemaConfig:
           configs:
-          - from: 2024-01-01
+          - from: "2024-01-01"
             store: tsdb
             index:
               prefix: loki_index_


pkg/storage/chunk/cache/redis_client.go
```diff
@@ -84,38 +84,41 @@ func NewRedisClient(cfg *RedisConfig) (*RedisClient, error) {
...
+        return endpoints, nil
+    }
+
+    // Handle single configuration endpoint which resolves multiple nodes.
+    host, port, err := net.SplitHostPort(endpoints[0])
+    if err != nil {
+        // If the port is missing, use the host as is and set a default port, if desired
+        host = endpoints[0]
+        // Example: set a default port, or you can choose to handle this differently
+        port = "6379" // Default Redis port
```
cstyan, line 104: this feels like something we should do via a `validateConfig` function before passing the config to the client creation

@dannykopping I think you worked on the caching code the most recently, any thoughts here?

nstankov-bg: Let me know if you would like me to approach this differently. I would be happy to contribute.

nstankov-bg: Following up on this @dannykopping  and @cstyan

cstyan: I would have a new function that's called before `deriveEndpoints` in `NewRedisClient` that logs a warning for every endpoint that doesn't have a port set clarifying that the default port will be used. Alternatively, this new function could be called before `NewRedisClient` in `cache.go` `New`.

PR #9031, opened 2023-04-05
Updated to reference Loki's usage of object storage

Alex3k:
**What this PR does / why we need it**:
Being 100% backed by object storage is a massive competitive win for us over tools like Elasticsearch and Splunk. I have updated this page to reference explicitly

PR #11443, opened 2023-12-11
Add note about use of object storage

JStickler:
PR #9031 from @Alex3k has gone quite stale as the topic has been revised since it was submitted.

This PR attempts to get at least part of Alex's update into the documentation so that we can close his original PR.


dannykopping (approved):
LGTM

PR #11446, opened 2023-12-12
Shortcut callback execution if there are 1 or fewer elements

dannykopping:
In the case where we have zero elements, we can exit early; with 1 element there's no need for concurrency.
This will provide a small optimisation and also make our profiles/execution traces easier to read.

We have a fairly significant amount of requests which don't require concurrency:
image

### DIFF
--- a/pkg/storage/stores/shipper/indexshipper/downloads/index_set.go
+++ b/pkg/storage/stores/shipper/indexshipper/downloads/index_set.go
@@ -224,8 +224,8 @@ func (t *indexSet) ForEachConcurrent(ctx context.Context, callback index.ForEach
 		}
 	}
 
-	var g *errgroup.Group
-	g, ctx = errgroup.WithContext(ctx)
+	//nolint:ineffassign,staticcheck
+	g, ctx := errgroup.WithContext(ctx)
 	if t.maxConcurrent == 0 {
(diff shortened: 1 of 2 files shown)


pkg/storage/stores/shipper/indexshipper/downloads/index_set.go
```diff
@@ -209,15 +209,27 @@ func (t *indexSet) ForEachConcurrent(ctx context.Context, callback index.ForEach
...
+	if len(t.index) == 0 {
+		return nil
+	}
+
+	// shortcut; if there's only one index, there's no need for bounded concurrency
+	if len(t.index) == 1 {
+		for i := range t.index {
+			idx := t.index[i]
+			return callback(t.userID == "", idx)
+		}
```
chaudum, line 224:
```suggestion
		return callback(t.userID == "", t.index[0])
```

chaudum (approved):

pkg/storage/stores/shipper/indexshipper/downloads/index_set.go
dannykopping: Haha I also tried this, but it's a map

chaudum: :see_no_evil:

PR #11447, opened 2023-12-12
Avoid reading block files into memory

chaudum:

Bloom blocks in the wild likely exceed the memory resources of a compactor. Therefore we need to make sure that we don't copy block data into memory, but rather write to and read from disk.


paul1r (approved):
lgtm

vlad-diachenko (approved):
looks great!

poyzannur (approved):
great! ty!

PR #11449, opened 2023-12-12
Map rpc Status back to user error

jeschkies:
Errors forwarded from the querier to the frontend would be reported with the wrong error code and message. This change fixes this.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/util/server/error.go
```diff
@@ -60,10 +63,22 @@ func ClientHTTPStatusAndError(err error) (int, error) {
 		return http.StatusBadRequest, err
 	case errors.Is(err, user.ErrNoOrgID):
 		return http.StatusBadRequest, err
+	case isRPC && (int(s.Code())/100 == 4 || int(s.Code())/100 == 5):
+		return int(s.Code()), errors.New(s.Message())
```
jeschkies, line 67: This fix is a little weird as we are not testing with `error.Is`. Furthermore this method is used to determine whether a call should be retried. I think error handling needs some refacotring. The retry should not be based on the status code but with a method checking `error.Is`. Which leads us to deserializing errors. I do not know how that's possible.

pkg/util/server/error.go
```diff
@@ -60,10 +63,22 @@ func ClientHTTPStatusAndError(err error) (int, error) {
...
+
+// WrapError wraps an error in a protobuf status.
+func WrapError(err error) *rpc.Status {
+	code, err := ClientHTTPStatusAndError(err)
+	return status.New(codes.Code(code), err.Error()).Proto()
+}
+
+func UnwrapError(s *rpc.Status) error {
+	return status.ErrorProto(s)
+}
```
jeschkies, line 84: I think this breaks the retry logic https://github.com/grafana/loki/blob/main/pkg/querier/queryrange/queryrangebase/retry.go#L91-L92 :/

jeschkies: We should use an embedded HTTP Response here like in here https://github.com/grafana/loki/blob/3555001aee06f01d26a33880326aa1bb9ab1fc93/pkg/querier/worker/util.go#L114-L117

jeschkies: After a talk with @duricanikolic I think we should migrate all errors to the gRPC `Status` interface eventually.

jeschkies: TIL: `httpgrpc.Errof` is returning an `httpgrpc.HTTPResponse` inside a `rpc.Status` as an error. So the current solution is fine but all user errors should be changed to `status.Errorf`.

pkg/util/server/error.go
```diff
@@ -60,10 +63,22 @@ func ClientHTTPStatusAndError(err error) (int, error) {
 		return http.StatusBadRequest, err
 	case errors.Is(err, user.ErrNoOrgID):
 		return http.StatusBadRequest, err
+	case isRPC && (int(s.Code())/100 == 4 || int(s.Code())/100 == 5):
```
duricanikolic, line 65: Wouldn't the logic be more intuitive if you separate cases where you need `isRPC` to be true from those where you don't need it?
```
s, isRPC := status.FromError(err) // I also recommend grpcutil.ErrorToStatus(err)
if isRPC {
  if s.Code() == codes.DeadlineExceeded {
    return http.StatusGatewayTimeout, errors.New(ErrDeadlineExceeded)
  } else if int(s.Code())/100 == 4 || int(s.Code())/100 == 5 {
    return int(s.Code()), errors.New(s.Message())
  }
  return http.StatusInternalServerError, err
}
// at this point, it is not a gRPC error, so both status.FromError(err) or httpgrpc.HTTPResponseFromError(err) would return false
switch {
  case errors.Is(err, context.Canceled) || 
	(errors.As(err, &promErr) && errors.Is(promErr.Err, context.Canceled)):
    return StatusClientClosedRequest, errors.New(ErrClientCanceled)
  case errors.Is(err, context.DeadlineExceeded): 
    return http.StatusGatewayTimeout, errors.New(ErrDeadlineExceeded)
    case errors.As(err, &queryErr):
      return http.StatusBadRequest, err
    case errors.Is(err, logqlmodel.ErrLimit) || errors.Is(err, logqlmodel.ErrParse) || errors.Is(err, logqlmodel.ErrPipeline) || errors.Is(err, logqlmodel.ErrBlocked) || errors.Is(err, logqlmodel.ErrParseMatchers):
      return http.StatusBadRequest, err
    case errors.Is(err, user.ErrNoOrgID):
      return http.StatusBadRequest, err
    default:
      return http.StatusInternalServerError, err
}
```

duricanikolic: Note: all these `errors.Is()` and `errors.As()` could be expensive. Imagine that most of the time you end in the `default` case. You would, at this point, try to execute all the `errors.Is()` and `errors.As()`, and only then you would return `http.StatusInternalServerError, err`. You might think about how to improve that.

pkg/util/server/error.go
duricanikolic: I also thing that the error handling needs some refactoring :)
But I also think that you cannot use `errors.Is` with `Status` errors, at least not to check whether they correspond to some of the types mentioned in this code.

pkg/util/server/error.go
```diff
@@ -60,10 +62,22 @@ func ClientHTTPStatusAndError(err error) (int, error) {
...
 		if grpcErr, ok := httpgrpc.HTTPResponseFromError(err); ok {
 			return int(grpcErr.Code), errors.New(string(grpcErr.Body))
 		}
 		return http.StatusInternalServerError, err
 	}
 }
+
+// WrapError wraps an error in a protobuf status.
+func WrapError(err error) *rpc.Status {
+	code, err := ClientHTTPStatusAndError(err)
```
duricanikolic, line 77: Hm... This is strange. So, in the case of a `Status` error, i.e., an error recognized by `status.FromError()` or `grpcutil.ErrorToStatus()`, you first split those errors into `<code, error message>` pairs (inside `ClientHTTPStatusAndError()`), and then you create another `Status` error containing exactly the same `code` and `error message`

pkg/util/server/error_test.go
```diff
@@ -9,13 +9,12 @@ import (
...
 
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
-
 	"github.com/grafana/dskit/httpgrpc"
 	"github.com/grafana/dskit/user"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
duricanikolic, line 17: Do you really want this dependency? The rest of the Loki repository uses `github.com/gogo/status`.
I also don't see where it is actually used in this source.

pkg/util/server/error.go
jeschkies: Thanks for the feedback. It's actually worse. `error.As` and `error.Is` is used to determine the code on the [querier side](https://github.com/grafana/loki/blob/main/pkg/querier/worker/util.go#L113). The error is then packaged in the `httpgrpc.HTTPResponse`. That triggers the default case of this method on the [frontend](https://github.com/grafana/loki/blob/main/pkg/querier/queryrange/serialize.go#L69). Unless ofc the error happened within the frontend 🙈

pkg/util/server/error.go
jeschkies: Yes. We probably should check if we have a status or not. Good point!

pkg/util/server/error_test.go
jeschkies: This is older code. I think the assumption was that gRPC might return a status for something like `status.New(codes.Canceled, context.Canceled.Error())`. That's how it's used in the test cases. Would a gogo protobuf generated client return a `gogo.status.Status`?

duricanikolic: You can simply use `github.com/gogo/status`.
I've check the rest of the code. The `status` package is used only for creating errors. This is an example of its usage:
```
status.New(codes.Canceled, context.Canceled.Error()).Err()
```
This can be simplified by using:
```
status.Error(codes.Canceled, context.Canceled.Error())
```

PR #11451, opened 2023-12-12
Tag queries with rule name

jeschkies:
Debugging remote rule evaluation is tough not knowing the rule name. This change adss the rule name to the tags.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/ruler/evaluator_remote.go
```diff
@@ -230,7 +233,7 @@ func (r *RemoteEvaluator) query(ctx context.Context, orgID, query string, ts tim
 			{Key: textproto.CanonicalMIMEHeaderKey("User-Agent"), Values: []string{userAgent}},
 			{Key: textproto.CanonicalMIMEHeaderKey("Content-Type"), Values: []string{mimeTypeFormPost}},
 			{Key: textproto.CanonicalMIMEHeaderKey("Content-Length"), Values: []string{strconv.Itoa(len(body))}},
-			{Key: textproto.CanonicalMIMEHeaderKey(string(httpreq.QueryTagsHTTPHeader)), Values: []string{"ruler"}},
+			{Key: textproto.CanonicalMIMEHeaderKey(string(httpreq.QueryTagsHTTPHeader)), Values: []string{"ruler,Source=ruler,Rule=" + detail.Name}},
```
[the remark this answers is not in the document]
dannykopping: This feels quite flimsy. Can we make it more robust and add some tests?

dannykopping (commented):
Thanks for adding this! This will be very valuable.

pkg/ruler/evaluator_remote.go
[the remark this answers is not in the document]
jeschkies: > This feels quite flimsy.

What I would expect is that once could set `X-Query-Tags` multiple times. Unfortunately it's not documented anywhere. I did see a set for the comma separated case. What do you think would break? What should I test?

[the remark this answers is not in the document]
dannykopping: The things I find flimsy about this are:

1. We're concatenating a string rather than using a struct which produces the final value and is independently testable
2. There aren't any tests here which prove this works
3. Having the name alone may not be enough information for us to triangulate which rule is acting up

pkg/ruler/evaluator_remote.go
```diff
@@ -341,6 +344,18 @@ func metricToLabels(m model.Metric) labels.Labels {
 	return b.Labels()
 }
 
+func detailToQueryTags(detail rules.RuleDetail) []string {
```
dannykopping, line 347: This really feels like a responsibility the ruler should not have.
In the `httpreq` we should have a helper which allows additional values to be injected into and extracted from these headers.

pkg/ruler/evaluator_remote_test.go
```diff
@@ -411,6 +414,77 @@ func TestRemoteEvalUnsupportedResultResponse(t *testing.T) {
...
+			for _, h := range in.Headers {
+				if h.Key == string(httpreq.QueryTagsHTTPHeader) {
+					queryTags = strings.Join(h.Values, ",")
+				}
+			}
+			require.Equal(t, "ruler,Source=ruler,Rule=foo,Kind=alerting", queryTags)
+
+			// this is somewhat bleeding the abstraction, but it's more idiomatic/readable than constructing
+			// the expected JSON response by hand
+			resp := loghttp.QueryResponse{
```
dannykopping, line 437: Why do we need a valid response? Aren't we just looking at the headers in the test?

pkg/ruler/evaluator_remote_test.go
```diff
@@ -411,6 +414,77 @@ func TestRemoteEvalUnsupportedResultResponse(t *testing.T) {
...
+
+	cli := mockClient{
+		handleFn: func(ctx context.Context, in *httpgrpc.HTTPRequest, opts ...grpc.CallOption) (*httpgrpc.HTTPResponse, error) {
+			var queryTags string
+			for _, h := range in.Headers {
+				if h.Key == string(httpreq.QueryTagsHTTPHeader) {
+					queryTags = strings.Join(h.Values, ",")
+				}
+			}
+			require.Equal(t, "ruler,Source=ruler,Rule=foo,Kind=alerting", queryTags)
```
dannykopping, line 433: I think the more correct/readable approach here would be to assign this value to a variable and move the `require.Equal` out into the main scope of the test.

For example, if this function is never called the test will pass.

pkg/ruler/evaluator_remote.go
jeschkies: I don't disagree. The query tags need some work and a standard format (seems logfmt would do). That said this blows up the work for this small enhancement.

PR #11350, opened 2023-11-30
Update Getting Started topic and graphics

JStickler:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Final PR to complete fixes #9228 

**Special notes for your reviewer**:
Updates Getting Started topic to be a general overview, links to Grafana topics, and Agent configuration file.
Adds new Quick Start topic for running Loki Docker compose file locally that includes working sample queries.
Adds three new graphics
Removes two unused graphics

PR #11478, opened 2023-12-13
fix image links in getting started

JStickler:
Fixes two image links as images aren't showing up from previous PR (#11350).

https://grafana.com/docs/loki/next/get-started/quick-start/  should display a graphic and a screenshot of Grafana Explore, but only the captions are visible.


MichelHollands (approved):
LGTM

PR #11060, opened 2023-10-26
docs(operator): Add proposal for short-lived token authentication

periklis:
**What this PR does / why we need it**:
The following PR provides an enhancement proposal document for the Loki Operator about the design and implementation details for short-lived token authentications across the three major cloud providers. More details about each provider:
- Azure: [Workload Identity Federation](https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation)
- AWS: [Secure Service Token](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html)
- Google Cloud Platform: [Workload Identity Federation](https://cloud.google.com/iam/docs/workload-identity-federation)

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

**Checklist**
- [ ] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated
  - [ ] If the change is worth mentioning in the release notes, add `add-to-release-notes` label
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/setup/upgrade/_index.md`
- [ ] For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)
- [ ] If the change is deprecating or removing a configuration option, update the `deprecated-config.yaml` and `deleted-config.yaml` files respectively in the `tools/deprecated-config-checker` directory.

PR #11481, opened 2023-12-13
operator: adds AWS sts support

JoaoBraveCoding:

Adds support for users to configure the S3 object storage with AWS STS.

- https://issues.redhat.com/browse/LOG-4544

- This PR depends on https://github.com/grafana/loki/pull/11500

To test you need:
- Run the AWS commands as explained in https://github.com/grafana/loki/pull/11060/files#diff-27b5aa2c87cf76afa33759f0ba55f749b2a54e1d4b1b6c7bc69a2cf1cbfc1c24R169-R213;
- Create a secret with `role_arn`, `region`, `bucketnames` and `audience`;
- Create a LokiStack.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/internal/manifests/storage/configure.go
```diff
@@ -186,29 +186,58 @@ func ensureObjectStoreCredentials(p *corev1.PodSpec, opts Options) corev1.PodSpe
...
 		}
 	case lokiv1.ObjectStorageSecretS3:
-		storeEnvVars = []corev1.EnvVar{
-			{
-				Name: EnvAWSAccessKeyID,
-				ValueFrom: &corev1.EnvVarSource{
-					SecretKeyRef: &corev1.SecretKeySelector{
-						LocalObjectReference: corev1.LocalObjectReference{
-							Name: secretName,
+		sts := opts.S3 != nil && len(opts.S3.RoleArn) != 0
```
periklis, line 189: WDYT pushing this condition into a method of S3, e.g. `opts.IsSTS()` (please feel free to improve naming). We can make use of it in the `loki-config.yaml` template and improve readability.

operator/internal/manifests/storage/configure.go
```diff
@@ -186,29 +186,58 @@ func ensureObjectStoreCredentials(p *corev1.PodSpec, opts Options) corev1.PodSpe
...
+							LocalObjectReference: corev1.LocalObjectReference{
+								Name: secretName,
+							},
+							Key: KeyAWSAccessKeySecret,
 						},
-						Key: KeyAWSAccessKeySecret,
 					},
 				},
-			},
+			}
```
periklis, line 214: Maybe it's time and should have been in done in the first place to separate these monster switch branch contents in small function in the `var.go` file, e.g. `func S3EnvVars(secretName string, isSTS bool){}`. WDYT? If yes we should create a prerequisite PR to backport it.

operator/internal/manifests/storage/configure.go
```diff
@@ -186,29 +186,58 @@ func ensureObjectStoreCredentials(p *corev1.PodSpec, opts Options) corev1.PodSpe
...
+					// TODO (JoaoBraveCoding) fix
+					Name: EnvAWSWebIdentityTokenFile,
+					ValueFrom: &corev1.EnvVarSource{
+						SecretKeyRef: &corev1.SecretKeySelector{
+							LocalObjectReference: corev1.LocalObjectReference{
+								Name: secretName,
+							},
+							Key: KeyAWSAccessKeySecret,
+						},
+					},
```
periklis, line 238: Since we will never expose setting the value of this to the user, we should consider setting up the value with a helper function when mounting the coresponding serviceaccount token projection, i.e.

1. On plain Kubernetes the token mount looks like this:
```yaml
---
serviceAccountName: lokistack-dev
---
volumeMounts:
# Watch the path is is over the openshift subdir
- mountPath: /var/run/secrets/kubernetes/serviceaccount 
  name: bound-sa-token
---
volumes:
- name: bound-sa-token
  projected:
    sources:
      - serviceAccountToken:
          # Watch out the audience is set to api here
          audience: api        
          expirationSeconds: 3600
          path: token
```
2. On OpenShift  the token mount looks like this:
```yaml
---
serviceAccountName: lokistack-dev
---
volumeMounts:
# Watch the path is is over the openshift subdir
- mountPath: /var/run/secrets/openshift/serviceaccount 
  name: bound-sa-token
---
volumes:
- name: bound-sa-token
  projected:
    sources:
      - serviceAccountToken:
          # Watch out the audient is openshift here *NOT* api
          audience: openshift         
          expirationSeconds: 3600
          path: token
```

periklis (commented):
Looks good to me. Couple of points still missing:
- Introduce the custom serviceaccount per LokiStack. Maybe we should do this as a pre-requisite PR and consider backporting this?
- Should we make the `audience` field in the service account projection volume configurable for non-OpenShift clusters?

operator/internal/manifests/storage/configure.go
JoaoBraveCoding: Let me know if it looks more easy to read. I tend to like deep functions with simple interfaces, functions should also be easy to read so if you still feel this too much we can split it

operator/internal/manifests/storage/configure.go
JoaoBraveCoding: I think I went a bit in this direction but maybe not what you had in mind. I'll not mark this as resolved and I want to have this into consideration when the code is finished as to see if I can improve it

operator/internal/manifests/storage/configure.go
JoaoBraveCoding: Added audience to the configuration as an optional parameter in the configMap and I'm thinking of passing down to `ExtractSecret`  the concept of an openshift or k8s cluster so we can correctly set the mountPath.

Thinking about it we can also make the path agnostic no? `/var/run/secrets/wif/serviceaccount` I'm picking WIF as the common denomination for this feature but we can pick something else

operator/internal/manifests/storage/var.go
```diff
@@ -99,6 +101,11 @@ const (
 	// KeySwiftPassword is the secret data key for the OpenStack Swift password.
 	KeySwiftUsername = "username"
 
+	saTokenVolumeName         = "bound-sa-token"
+	saTokenVolumeK8sDirectory = "/var/run/secrets/kubernetes/serviceaccount"
```
periklis, line 105:
```suggestion
	saTokenVolumeK8sDirectory = "/var/run/secrets/kubernetes.io/serviceaccount"
```

operator/internal/manifests/storage/var.go
```diff
@@ -99,6 +101,11 @@ const (
...
+	saTokenVolumeOcpDirectory = "/var/run/secrets/openshift/serviceaccount"
+	saTokenExpiration         = 3600
```
periklis, line 107: We need to investigate if we can rely on a default constant here. IIRC some provides like GCP allow you to set the expiration date on their side. If these two sides are related maybe we need to give the freedom back to keep in sync. Will lookup for you and inform you on this thread ASAP.

operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -124,38 +124,58 @@ func extractGCSConfigSecret(s *corev1.Secret) (*storage.GCSStorageConfig, error)
...
+		}
+	} else {
+		// In the STS case region is not an optional field
+		if len(region) == 0 {
+			return nil, kverrors.New("missing secret field", "field", storage.KeyAWSRegion)
+		}
+		wiTokenFile = storage.SATokenVolumeK8sDirectory
+		if osEnabled {
+			wiTokenFile = storage.SATokenVolumeOcpDirectory
+		}
```
periklis, line 163: Can we skip doing this extra branching here to improve readibility and making this function OCP-unaware? In fact we have our storage configure method that takes care of patching the envVars anyway. We could make the decision making there? WDYT?

operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -155,6 +167,8 @@ func extractS3ConfigSecret(s *corev1.Secret) (*storage.S3StorageConfig, error) {
 		Endpoint: string(endpoint),
 		Buckets:  string(buckets),
 		Region:   string(region),
+		RoleArn:  string(roleArn),
```
periklis, line 170: I suggest not to pass the role_arn through the operator callstack, because the secret will be projected into environment variables for the Loki pods anyway so will this secret field become `ROLE_ARN`.

operator/internal/manifests/internal/config/build_test.go
```diff
@@ -5418,3 +5418,199 @@ analytics:
...
+					Buckets:              "my-bucket",
+				},
+				Schemas: defaultSchema,
+			},
+			expStorageConfig: `
+    s3:
+      s3: s3://my-region/my-bucket
+      s3forcepathstyle: false`,
+		},
+	} {
```
periklis, line 5450: Any reason to keep this folded as a table test? In general in this test file I would keep one test case one test function because the expected values become ugly long and table test add more reading burden.

operator/internal/manifests/storage/configure.go
```diff
@@ -187,68 +166,26 @@ func ensureObjectStoreCredentials(p *corev1.PodSpec, opts Options) corev1.PodSpe
...
-							Name: secretName,
-						},
-						Key: KeyAWSAccessKeySecret,
-					},
+			envVarFromSecret(EnvAWSAccessKeyID, secretName, KeyAWSAccessKeyID),
+			envVarFromSecret(EnvAWSAccessKeySecret, secretName, KeyAWSAccessKeySecret),
+		}
+		// STS Auth Workflow
+		if wifEnabled(opts) {
+			storeEnvVars = []corev1.EnvVar{
```
periklis, line 174: I almost read that we added all four env vars when `wifEnabled` returns `true`. This whole switch case is hard to read. Need to :thinking: about possible improvements.

operator/internal/manifests/internal/config/build_test.go
JoaoBraveCoding: My goal would be to add the other test cases for the other WIF implementations but if we want to keep 1 function per test case I'm happy to oblige. Although I believe that when they are like this it's easier to read because we see all the important parts of the test at the same time.

operator/internal/handlers/internal/storage/secrets.go
JoaoBraveCoding: Ahhh I see!  I hadn't thought about that! I'll improve it then

operator/internal/manifests/compactor.go
```diff
@@ -26,7 +26,7 @@ func BuildCompactor(opts Options) ([]client.Object, error) {
 		}
 	}
 
-	if err := storage.ConfigureStatefulSet(statefulSet, opts.ObjectStorage); err != nil {
+	if err := storage.ConfigureStatefulSet(statefulSet, opts.ObjectStorage, opts.Gates.OpenShift.Enabled); err != nil {
```
periklis, line 29: When we need more than one field from `Options`, I suggest to pass the whole object instead of exploding it in parameters. This encapsulates usage in the callee site and in turn when you look up for references of Options fields you have getters/setters always within the context of the callee site.

operator/internal/manifests/storage/configure.go
```diff
@@ -17,12 +18,12 @@ import (
 // - All: Ensure object storage secret mounted and auth projected as env vars.
 // - GCS: Ensure env var GOOGLE_APPLICATION_CREDENTIALS in container
 // - S3: Ensure mounting custom CA configmap if any TLSConfig given
-func ConfigureDeployment(d *appsv1.Deployment, opts Options) error {
+func ConfigureDeployment(d *appsv1.Deployment, opts Options, osEnabled bool) error {
```
periklis, line 21: As mentioned above, if we pass the full `manifests.Options` struct here we don't need to reinvent naming for references to individual fields. In fact` osEnabled` does not immediately ring the bell for everyone not long acquainted with the code bases to `opts.FeatureGates.OpenShift.Enabled`, WDYT?

operator/internal/manifests/storage/configure.go
```diff
@@ -126,139 +126,69 @@ func ensureObjectStoreCredentials(p *corev1.PodSpec, opts Options) corev1.PodSpe
...
-	var storeEnvVars []corev1.EnvVar
-	switch storeType {
+	if !wifEnabled(opts) {
+		container.Env = append(container.Env, envVarsNoWIF(opts)...)
+	} else {
+		setSATokenPath(&opts, osEnabled)
+		container.Env = append(container.Env, envVarsWIF(opts)...)
+		volumes = append(volumes, saTokenVolume(opts))
+		container.VolumeMounts = append(container.VolumeMounts, saTokenVolumeMount(opts))
+	}
```
periklis, line 136: I really like the style introduced here. It makes this function more readable. Although I feel `wif` as a prefix is a bit hard to grasp (as in general the whole managed identity federation concepts). I suggest the following rename that we should pass into the [enhancement proposal](https://github.com/grafana/loki/pull/11060):
1. `StaticAuthCredentials`: For everything that is using a static serviceaccount and accompanying credentials.
2. `ManagedAuthCredentials`: For everything that is using a federated/managed identity to access an authentication token at the end.

We can leave `Credentials` out if it yields too long function/variable names.

operator/internal/manifests/storage/configure.go
```diff
@@ -126,139 +126,69 @@ func ensureObjectStoreCredentials(p *corev1.PodSpec, opts Options) corev1.PodSpe
...
-				ValueFrom: &corev1.EnvVarSource{
-					SecretKeyRef: &corev1.SecretKeySelector{
-						LocalObjectReference: corev1.LocalObjectReference{
-							Name: secretName,
-						},
-						Key: KeyAWSSseKmsEncryptionContext,
-					},
-				},
-			})
+			storeEnvVars = append(storeEnvVars, envVarFromSecret(EnvAWSSseKmsEncryptionContext, secretName, KeyAWSSseKmsEncryptionContext))
```
periklis, line 169: This is missing in the STS variant. We may need to create two appending functions:
1. One pair as proposed here `envVarsWithWIF/envVarsNoWIF` that appends authentication env vars per store type.
2. One `envVarsServerSideEncryption` that appends env vars related to server side encryption per store type if applicable.

WDYT?

operator/internal/manifests/storage/configure.go
```diff
@@ -294,3 +224,83 @@ func ensureCAForS3(p *corev1.PodSpec, tls *TLSConfig) corev1.PodSpec {
...
+	}
+}
+
+func saTokenVolume(opts Options) corev1.Volume {
+	var audience string
+	storeType := opts.SharedStore
+	switch storeType {
+	case lokiv1.ObjectStorageSecretS3:
+		audience = opts.S3.Audience
+	}
```
periklis, line 268: This is missing the default `audience` here. I believe in plain Kuberentes the audience is always `api`, but please double this suggestion before applying it.

JoaoBraveCoding: I was researching this and as far as I found `api` is only for Azure `audience: api://AzureADTokenExchange` like so. For aws I believe it would be `audience: sts.amazonaws.com`

operator/internal/manifests/compactor.go
JoaoBraveCoding: I'm not able to import the `manifests.Options` as that creates a circular dependency. After some consideration, I believe we have 2 options:
1. We set the SATokeFilePath before reaching the manifests generation (I was thinking after extract secret)
2. We replicate the OpenShift enabled fields inside `storage.Options` (This was what i ended up following, let me know how it looks)

operator/internal/manifests/storage/var.go
JoaoBraveCoding: I was researching this and I was not able to find anything on this being a configurable field from the provider side 🤔

operator/internal/manifests/storage/configure.go
```diff
@@ -236,3 +257,65 @@ func envVarFromValue(name, value string) corev1.EnvVar {
...
+func saTokenVolume(opts Options) corev1.Volume {
+	var audience string
+	storeType := opts.SharedStore
+	switch storeType {
+	case lokiv1.ObjectStorageSecretS3:
+		audience = awsDefaultAudience
+		if opts.S3.Audience != "" {
+			audience = opts.S3.Audience
+		}
+	}
```
periklis, line 304: I believe we miss to default the audience for OpenShift ServiceAccount tokens to `openshift` here, right? 
```suggestion
	case lokiv1.ObjectStorageSecretS3:
		audience = awsDefaultAudience
		if opts.S3.Audience != "" {
			audience = opts.S3.Audience
		}
		if opts.OpenShiftEnabled != "" {
			audience = "openshift"
		}
	}
```

operator/internal/manifests/storage/configure.go
```diff
@@ -236,3 +257,65 @@ func envVarFromValue(name, value string) corev1.EnvVar {
...
+func saTokenVolumeMount(opts Options) corev1.VolumeMount {
+	var wiToken string
+	switch opts.SharedStore {
+	case lokiv1.ObjectStorageSecretS3:
+		wiToken = opts.S3.WebIdentityTokenFile
+	}
+	return corev1.VolumeMount{
+		Name:      saTokenVolumeName,
+		MountPath: wiToken,
+	}
```
periklis, line 292:
```suggestion
	var tokenPath string
	switch opts.SharedStore {
	case lokiv1.ObjectStorageSecretS3:
		tokenPath = opts.S3.WebIdentityTokenFile
	}
	return corev1.VolumeMount{
		Name:      saTokenVolumeName,
		MountPath: tokenPath,
	}
```

operator/internal/manifests/storage/configure.go
```diff
@@ -236,3 +257,65 @@ func envVarFromValue(name, value string) corev1.EnvVar {
...
+	var wiToken string
+	switch opts.SharedStore {
+	case lokiv1.ObjectStorageSecretS3:
+		wiToken = saTokenVolumeK8sDirectory
+		if opts.OpenShiftEnabled {
+			wiToken = saTokenVolumeOcpDirectory
+		}
+		opts.S3.WebIdentityTokenFile = wiToken
+	}
+}
```
periklis, line 281:
```suggestion
func setSATokenPath(opts *Options) {
	switch opts.SharedStore {
	case lokiv1.ObjectStorageSecretS3:
		opts.S3.WebIdentityTokenFile = saTokenVolumeK8sDirectory
		if opts.OpenShiftEnabled {
			opts.S3.WebIdentityTokenFile = saTokenVolumeOcpDirectory
		}
	}
}
```

operator/internal/manifests/storage/configure.go
```diff
@@ -236,3 +257,65 @@ func envVarFromValue(name, value string) corev1.EnvVar {
...
 }
+
+func managedAuthEnabled(opts Options) bool {
+	storeType := opts.SharedStore
+	switch storeType {
+	case lokiv1.ObjectStorageSecretS3:
+		return opts.S3 != nil && opts.S3.STS
+	default:
+		return false
+	}
```
periklis, line 268:
```suggestion
	switch opts.SharedStore {
	case lokiv1.ObjectStorageSecretS3:
		return opts.S3 != nil && opts.S3.STS
	default:
		return false
	}
```

operator/internal/manifests/storage/configure.go
```diff
@@ -168,6 +176,19 @@ func staticAuthCredentials(opts Options) []corev1.EnvVar {
...
+	secretName := opts.SecretName
+	switch opts.SharedStore {
+	case lokiv1.ObjectStorageSecretS3:
+		return []corev1.EnvVar{
+			envVarFromSecret(EnvAWSRoleArn, secretName, KeyAWSRoleArn),
+			envVarFromValue(EnvAWSWebIdentityTokenFile, path.Join(opts.S3.WebIdentityTokenFile, "token")),
+		}
+	default:
+		return []corev1.EnvVar{}
+	}
```
periklis, line 189:
```suggestion
	switch opts.SharedStore {
	case lokiv1.ObjectStorageSecretS3:
		return []corev1.EnvVar{
			envVarFromSecret(EnvAWSRoleArn, opts.SecretName, KeyAWSRoleArn),
			envVarFromValue(EnvAWSWebIdentityTokenFile, path.Join(opts.S3.WebIdentityTokenFile, "token")),
		}
	default:
		return []corev1.EnvVar{}
	}
```

periklis (approved):
Minor improvements, I think we arrived at the finish line.

shriyo101:
Is this feature open for new users who wants to use Loki with s3 in STS?

periklis:
@shriyo101 Yes of course. Please consider the enhancement doc: https://loki-operator.dev/docs/short_lived_tokens_authentication.md/#aws-secure-token-service

PR #11486, opened 2023-12-14
Run compaction over existing bloom blocks

poyzannur:
Upon changes to TSDB indexes, the existing blooms needs to be amended with newly added chunks, or new series should be added to blocks. 
This PR compares the existing meta information with the most up to date TSDB information in the job. In case of no changes to indexes, compaction is skipped; if there's any block needs amending, it does that and merges all blocks in fp range; if there's no previous compaction, it compacts from scratch. 

This logic does not support cutting blocks upon topology changes to bloom-compactors. It can create blocks with series outside of the fp range of the compactor. Cutting blocks will be addressed in a follow-up PR.


pkg/bloomcompactor/utils.go
```diff
@@ -0,0 +1,37 @@
...
+	var oldTombstonedBlockRefs map[bloomshipper.BlockRef]struct{}
+
+	for _, meta := range metas {
+		if meta.TableName != job.tableName {
+			continue
+		}
+		metasMatchingJob = append(metasMatchingJob, meta)
+
+		for _, tombstonedBlockRef := range meta.Tombstones {
+			oldTombstonedBlockRefs[tombstonedBlockRef] = struct{}{}
```
paul1r, line 17: Isn't the map nil here?

pkg/bloomcompactor/mergecompactor.go
```diff
@@ -0,0 +1,125 @@
...
+			Fingerprint: s.seriesFP,
+			Chunks:      crefs,
+		}
+	}
+	seriesIter := v1.NewSliceIter(seriesFromSeriesMeta)
+
+	// TODO: Make blockIters an actual, lazy iterator
+	// Download existing blocks that needs compaction
+	blockIters := make([]v1.PeekingIterator[*v1.SeriesWithBloom], len(blocksToUpdate))
+	for i, b := range blocksToUpdate {
```
chaudum, line 67: Should we make this concurrent?

pkg/bloomcompactor/compress_utils.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/go-kit/log/level"
+	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
+	"github.com/grafana/loki/pkg/storage/stores/shipper/bloomshipper"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func compressBloomBlock(storedBlock bloomshipper.Block, archivePath, localDst string, logger log.Logger) (bloomshipper.Block, error) {
```
chaudum, line 15: I think we should only pass the block's `BlockRef`, since `Data` is not used.

```suggestion
func compressBloomBlock(ref bloomshipper.BlockRef, archivePath, localDst string, logger log.Logger) (bloomshipper.Block, error) {
```

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -463,92 +463,88 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
...
-			}
-		}()
+	}
 
-		// Do not change the signature of PutBlocks yet.
-		// Once block size is limited potentially, compactNewChunks will return multiple blocks, hence a list is appropriate.
-		storedBlocks, err := c.bloomShipperClient.PutBlocks(ctx, []bloomshipper.Block{blockToUpload})
+	archivePath := filepath.Join(c.cfg.WorkingDirectory, uuid.New().String())
+
+	blockToUpload, err := compressBloomBlock(resultingBlock, archivePath, localDst, logger)
```
chaudum, line 518: `resultingBlock.Data` needs to be closed

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -463,92 +463,88 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
...
+	if err != nil {
+		level.Error(logger).Log("msg", "putting blocks to storage", "err", err)
```
chaudum, line 520: The `msg` does not reflect the cause of the error.

pkg/bloomcompactor/bloomcompactor.go
vlad-diachenko: it will be closed by the client https://github.com/grafana/loki/blob/5f4285963c5c4a796094e67afce51b2f67045094/pkg/storage/stores/shipper/bloomshipper/client.go#L231

vlad-diachenko: maybe it's responsibility of the compactor to close the reader. but then we need to remove closing the reader in the client

pkg/bloomcompactor/compress_utils.go
```diff
@@ -0,0 +1,79 @@
+package bloomcompactor
```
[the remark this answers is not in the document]
poyzannur: Did you mean bloomutils?

pkg/bloomcompactor/bloomcompactor.go
poyzannur: It's used in multiple places, happy to do a follow up PR for this.

pkg/bloomcompactor/utils.go
```diff
@@ -0,0 +1,37 @@
...
+				continue
+			}
+
+			if blockRef.IndexPath == job.indexPath {
+				// index has not changed, no compaction needed
+				continue
+			}
+			blocksMatchingJob = append(blocksMatchingJob, blockRef)
+		}
+	}
```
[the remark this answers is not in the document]
poyzannur: The idea of two loops is to accumulate all previously tombstoned refs in one go first. if a block is later seen to in tombstoned list then i will have to remove it from `blocksMatchingJobs`. 

Instead of removing from a list, i think this is cleaner to read. block lists should not grow super big, so i didn't think it's a deal breaker. Please let me know what you prefer here though.

pkg/bloomcompactor/bloomcompactor.go
chaudum: > it will be closed by the client 

`resultingBlock` is not the object that is written using `PutBlocks`

pkg/bloomcompactor/compress_utils.go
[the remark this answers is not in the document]
chaudum: No. I think the shipper would be a more appropriate place. However, bloomutils is also ok.

pkg/bloomcompactor/bloomcompactor.go
poyzannur: ah fair, thanks for catching. added.

pkg/bloomcompactor/compress_utils.go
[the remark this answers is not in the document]
poyzannur: thanks moved now.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -463,92 +463,95 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
...
+	}
 
-		// Do not change the signature of PutBlocks yet.
-		// Once block size is limited potentially, compactNewChunks will return multiple blocks, hence a list is appropriate.
-		storedBlocks, err := c.bloomShipperClient.PutBlocks(ctx, []bloomshipper.Block{blockToUpload})
+	archivePath := filepath.Join(c.cfg.WorkingDirectory, uuid.New().String())
+
+	blockToUpload, err := bloomshipper.CompressBloomBlock(resultingBlock.BlockRef, archivePath, localDst, logger)
+	if err != nil {
+		level.Error(logger).Log("msg", "compressing bloom blocks in to tar files", "err", err)
```
chaudum, line 527:
```suggestion
		level.Error(logger).Log("msg", "failed to compress bloom block directory into tar file", "err", err)
```

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -463,92 +463,95 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
...
+		return err
+	}
+	defer func() {
+		err = os.Remove(archivePath)
 		if err != nil {
-			level.Error(logger).Log("msg", "putting blocks to storage", "err", err)
-			return err
+			level.Error(logger).Log("msg", "removing archive file", "err", err, "file", archivePath)
```
chaudum, line 533: I would write messages of logs that are only logged in case of an error in negative form, such as:
`failed to ...`, or `did not ...`

poyzannur: i think I made a wrong observation previously in another components, bc I was only writing like this modelling how we do elsewhere. I'll review all and fix the errors with a negative phrase.

PR #11488, opened 2023-12-14
Refactor Distributor.Push() to be able to Tee log streams.

na--:

This allows us to duplicate (i.e. "tee") some of the log streams from the `Distributor` by introducing an optional interface in a follow-up PR.

None

See inline comments

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -272,9 +272,14 @@ func (d *Distributor) stopping(_ error) error {
 	return services.StopManagerAndAwaitStopped(context.Background(), d.subservices)
 }
 
+type KeyedStream struct {
+	HashKey uint32
+	Stream  logproto.Stream
+}
+
 // TODO taken from Cortex, see if we can refactor out an usable interface.
 type streamTracker struct {
-	stream      logproto.Stream
+	KeyedStream
 	minSuccess  int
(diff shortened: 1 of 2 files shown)


pkg/distributor/distributor.go
```diff
@@ -413,6 +417,7 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
 	const maxExpectedReplicationSet = 5 // typical replication factor 3 plus one for inactive plus one for luck
 	var descs [maxExpectedReplicationSet]ring.InstanceDesc
 
+	streamTrackers := make([]streamTracker, len(streams))
```
na--, line 420: This slice isn't really necessary, I could just have something like this on line 439:
```go
st = &streamTracker{
    // ...
}
```

And then just pass `st` instead of `&streamTrackers[i]` on line 445. However, the memory usage won't be significantly lower, considering we still have to allocate the `streamTracker{}` objects, but the GC trashing will probably be quite a bit higher, considering how many of these heap objects will be created separately.

MasslessParticle (approved):
I think this is an unambiguous clean up. LGTM

PR #11492, opened 2023-12-14
Bloom querying: Propagate line filters from querier to bloom filter

chaudum:
In order to query the bloom filters with the search strings from the user query, we need to propagate the line filters through the different implementations of the store, namely through the GetChunks function.

In order to avoid converting data structures, the line filter from the syntax tree is used as custom protobuf type, because it is passed from querier to index gateway and from index gateway to bloom gateway.

~~For querying the bloom blocks, line filters from the user queries need to be tokenized into n-grams. This is momentarily done using the per-tenant n-gram length setting.
This is problematic, because the n-gram length needs to match the one with which the bloom filters were created. However, the bloom block still lacks this information.~~

~~Adding the ngram size to the bloom block, and tokenizing the search string at query time based on this value are subject to a follow up PR (see https://github.com/grafana/loki/pull/11512)~~

**Update**: The follow-up PR has been merged into this one

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/bloomgateway/multiplexing.go
```diff
@@ -15,6 +15,10 @@ const (
 	Day = 24 * time.Hour
 )
 
+type tokenSettings struct {
+	nGramLen int
```
paul1r, line 19: also will need a `skipVal int`

pkg/logproto/bloomgateway.proto
```diff
@@ -17,7 +17,10 @@ message FilterChunkRefRequest {
     (gogoproto.nullable) = false
   ];
   repeated GroupedChunkRefs refs = 3;
-  repeated logproto.LineFilterExpression filters = 4;
+  repeated bytes filters = 4 [
```
jeschkies, line 20: This fails. You need to define a message type:
```suggestion
  message Filter {
    bytes raw = 1;
  }
  repeated Filter filters = 4 [
```

pkg/bloomgateway/bloomgateway.go
```diff
@@ -298,6 +300,10 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
 	})
 
 	task, resCh, errCh, err := NewTask(tenantID, req)
+	task.tokenSettings = tokenSettings{
+		nGramLen: g.overrides.BloomNGramLength(tenantID),
+	}
```
chaudum, line 305: This is only a temporary solution

pkg/storage/chunk/predicate.go
```diff
@@ -0,0 +1,16 @@
+package chunk
+
+import (
+	"github.com/prometheus/prometheus/model/labels"
+
+	"github.com/grafana/loki/pkg/logql/syntax"
+)
+
+type Predicate struct {
```
chaudum, line 9: Maybe we can come up with a better name.

pkg/bloomgateway/config.go
```diff
@@ -37,8 +37,13 @@ func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	cfg.Client.RegisterFlags(f)
 }
 
+type NGramSettings interface {
+	BloomNGramLength(tenantID string) int
+}
+
 type Limits interface {
```
jeschkies, line 44: That's probably the answer to my question 🙂

pkg/bloomgateway/querier_test.go
```diff
@@ -55,8 +56,8 @@ func TestBloomQuerier(t *testing.T) {
 		through := model.Now()
 		from := through.Add(-12 * time.Hour)
 		chunkRefs := []*logproto.ChunkRef{}
-		filters := []*logproto.LineFilterExpression{
-			{Operator: 0, Match: "uuid"},
+		filters := []syntax.LineFilter{
+			{Ty: 0, Match: "uuid", Op: ""},
```
jeschkies, line 60: Could you use the enums and constants from `syntax` instead of `0` and `""`?

pkg/bloomgateway/util.go
```diff
@@ -36,12 +38,22 @@ func getFromThrough(refs []*logproto.ShortRef) (model.Time, model.Time) {
...
+	// Can we assume an avg filter string length of 12 chars?
+	// Then the capacity would be (len(filterString) - (nGramLen-1)) * numFilters
+	searches := make([][]byte, 0, (13-cfg.nGramLen)*len(filters))
 	for _, f := range filters {
-		searches = append(searches, []byte(f.Match))
+		if f.Ty == labels.MatchEqual {
+			it := t.Tokens(f.Match)
+			for it.Next() {
+				key := make([]byte, cfg.nGramLen)
+				_ = copy(key, it.At())
```
jeschkies, line 53: Shouldn't this work or will you get a linter error?
```suggestion
				copy(key, it.At())
```

pkg/logproto/bloomgateway.proto
jeschkies: I think you need to change this as well, don't you?

pkg/logql/syntax/linefilter.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Binary encoding of the LineFilter
+// integer is varint encoded
+// strings are variable-length encoded
+//
+// +---------+--------------+-------------+
+// | Ty      | Match        | Op          |
+// +---------+--------------+-------------+
+// | value   | len  | value | len | value |
+// +---------+--------------+-------------+
```
jeschkies, line 17: That's probably a nice compromise but I'd go with the JSON one we have to keep consistency.

pkg/logql/syntax/visit.go
```diff
@@ -168,8 +168,12 @@ func (v *DepthFirstTraversal) VisitLineFilter(e *LineFilterExpr) {
...
 		v.VisitLineFilterFn(v, e)
 	} else {
-		e.Left.Accept(v)
-		e.Or.Accept(v)
+		if e.Left != nil {
+			e.Left.Accept(v)
+		}
+		if e.Or != nil {
+			e.Or.Accept(v)
+		}
```
jeschkies, line 176: Nice catch!

pkg/storage/chunk/predicate.go
```diff
@@ -0,0 +1,16 @@
...
+	Matchers []*labels.Matcher
```
jeschkies, line 10: Matchers are for selecting the series, right?

pkg/storage/chunk/predicate.go
jeschkies: `Filter` or `Selector` are the only alternatives I find fitting.

pkg/storage/store.go
```diff
@@ -461,15 +460,34 @@ func (s *LokiStore) SelectSeries(ctx context.Context, req logql.SelectLogParams)
...
+	lineFilters := make([]syntax.LineFilter, 0)
+	visitor := &syntax.DepthFirstTraversal{
+		VisitLineFilterFn: func(v syntax.RootVisitor, e *syntax.LineFilterExpr) {
+			if e.Left != nil {
+				e.Left.Accept(v)
+			}
+			if e.Or != nil {
+				e.Or.Accept(v)
+			}
+			lineFilters = append(lineFilters, e.LineFilter)
```
jeschkies, line 473: Neat! I would have created a new visitor that embeds the default visitor but this is much cleaner and easier to understand.

pkg/bloomgateway/util.go
chaudum: Linter says "No!" :D

pkg/storage/chunk/predicate.go
chaudum: Right, these are the label selectors.

cmd/migrate/main.go
```diff
@@ -310,7 +310,7 @@ func (m *chunkMover) moveChunks(ctx context.Context, threadID int, syncRangeCh <
 			var totalBytes uint64
 			var totalChunks uint64
 			//log.Printf("%d processing sync range %d - Start: %v, End: %v\n", threadID, sr.number, time.Unix(0, sr.from).UTC(), time.Unix(0, sr.to).UTC())
-			schemaGroups, fetchers, err := m.source.GetChunks(m.ctx, m.sourceUser, model.TimeFromUnixNano(sr.from), model.TimeFromUnixNano(sr.to), m.matchers...)
+			schemaGroups, fetchers, err := m.source.GetChunks(m.ctx, m.sourceUser, model.TimeFromUnixNano(sr.from), model.TimeFromUnixNano(sr.to), chunk.NewPredicate(m.matchers, nil))
```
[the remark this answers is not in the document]
chaudum: Yes

pkg/bloomgateway/querier_test.go
chaudum: Op is usually "", unless there are nested OR filters

chaudum: I think I can omit `Op`

pkg/logql/syntax/linefilter.go
chaudum: Consistent with the Plan serialization?

chaudum: I used the encoding utils that we already have, which makes the code much simpler: https://github.com/grafana/loki/pull/11492/commits/335d79ce4a909bc2738b52cbf2934c76c5aa1a63

poyzannur (approved):

pkg/logql/syntax/linefilter.go
jeschkies: Nice. Something to consider for the query plan.

jeschkies (approved):

PR #11493, opened 2023-12-14
Implement hooks to instrument query pipelines

MasslessParticle:
This PR implements interfaces to inject new pipelines/extractors into the query path of queriers and ingesters.

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	lokilog "github.com/grafana/loki/pkg/logql/log"
 	"math/rand"
 	"net/http"
 	"os"
@@ -12,6 +11,8 @@ import (
 	"path/filepath"
 	"sync"
 	"time"
+
+	lokilog "github.com/grafana/loki/pkg/logql/log"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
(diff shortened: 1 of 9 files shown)


pkg/ingester/ingester.go
```diff
@@ -99,7 +101,10 @@ type Config struct {
 
 	WAL WALConfig `yaml:"wal,omitempty" doc:"description=The ingester WAL (Write Ahead Log) records incoming logs and stores them on the local file systems in order to guarantee persistence of acknowledged data in the event of a process crash."`
 
-	ChunkFilterer chunk.RequestChunkFilterer `yaml:"-"`
+	ChunkFilterer          chunk.RequestChunkFilterer     `yaml:"-"`
```
na--, line 104: It's quite weird to see these objects in a `Config`, but I see they are not the only ones like that :confused: Is that the only way these non-config objects can be injected? :confused:

na-- (commented):
LGTM except for stuffing interface objects in a `Config`, but I don't know any of the code and assumptions here yet, so I don't feel like I can properly give it a :+1: and someone else should also take a look.

pkg/ingester/ingester.go
MasslessParticle: It's not the only way, just the way they happen to be injected in the ingester. I've followed the pattern here rather than bifurcating how it's wired together.

paul1r (approved):
lgtm

PR #11495, opened 2023-12-15
feat: mixin, allow overriding of some labels by parameterizing mixin recording/alert rules

alex5517:
Allows for more dynamic generation of recording/alert rules. 

Fixes N/A

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


production/loki-mixin/config.libsonnet
```diff
@@ -11,6 +14,20 @@
 
     // The label used to differentiate between different clusters.
     per_cluster_label: 'cluster',
+    per_namespace_label: 'namespace',
+    per_job_label: 'job',
+
+    // Grouping labels, to uniquely identify and group by {jobs, clusters}
+    job_labels: [$._config.per_cluster_label, $._config.per_namespace_label, $._config.per_job_label],
+    cluster_labels: [$._config.per_cluster_label, $._config.per_namespace_label],
```
cstyan, line 22: just to avoid confusion lets make the names as explicit as possible, I would assume that `job_labels` and `cluster_labels` select only `job` or `cluster`, not a combination of those plus other labels.

production/loki-mixin/config.libsonnet
```diff
@@ -11,6 +14,20 @@
...
+
+    // Each group prefix is composed of `_`-separated labels
+    group_prefix_jobs: makePrefix($._config.job_labels),
+    group_prefix_clusters: makePrefix($._config.cluster_labels),
+
+    // Each group-by label list is `, `-separated and unique identifies
+    group_by_job: makeGroupBy($._config.job_labels),
+    group_by_cluster: makeGroupBy($._config.cluster_labels),
```
cstyan, line 30: IIUC this allows for any of the labels that are defaulted to be in `x_labels`, for example `cluster, namespace, job`, to actually be removed, in case someone only wants to select on `namespace, job` for example?

production/loki-mixin/mixin-ssd.libsonnet
```diff
@@ -1,6 +1,4 @@
-(import 'dashboards.libsonnet') +
-(import 'alerts.libsonnet') +
-(import 'recording_rules.libsonnet') + {
+(import 'mixin.libsonnet') + {
```
cstyan, line 1: was this change automated?

cstyan (commented):
I think making the label selectors more modular is a good idea

production/loki-mixin/config.libsonnet
alex5517: @cstyan,
Sorry for the late reply, need to have a look at my notification settings... 

I do not necessarily disagree with that, my thoughts behind it was that i wanted it to be familiar for user to use it, and since Mimir mixins does it like that... https://github.com/grafana/mimir/blob/main/operations/mimir-mixin/config.libsonnet#L179-L182

production/loki-mixin/config.libsonnet
alex5517: That is correct.

production/loki-mixin/mixin-ssd.libsonnet
alex5517: No i just figured there was no need to define it twice.

cstyan: we should just double check that everything still works here, I'll have to double check if we have an ssd deployment ourselves to deploy this against but if you have one yourself just lmk

production/loki-mixin/config.libsonnet
cstyan: I don't have a very strong preference either way, but I think the simplified naming is for historical reasons and we've just tacked on additional labels as we went without changing the name. Your reasoning is valid as well, I'll leave the end decision here to you.

production/loki-mixin/mixin-ssd.libsonnet
alex5517: I have several SSD deployments running myself, and this update does not render anything that previously worked with SSD unusable now :)

I also noticed that I forgot to remove the import of config.libsonnet from dashboards.libsonnet, this is done now.

Previously, the only variables found in config.libsonnet were used in the dashboards. Therefore, it was properly decided to simply import it within the dashboards.libsonnet file. This change ensures its availability for alerts and recording rules as well.

Considering the only difference between mixin.libsonnet and mixin-ssd.libsonnet is some alterations to the _config+:: object and grafanaDashboardFolder, I believe it's acceptable to import everything from mixin.libsonnet and perform the necessary overwrites.

Alternatively, we could completely remove mixin-ssd.libsonnet and allow users to specify their preferences in config.libsonnet.

cstyan:
@alex5517 looking good! I think we're more or less good to go. Unfortunately we need to merge/rebase on master to pull in changes for CI.

alex5517:
@cstyan 

Found the culprit :)

cstyan:
We also need to update the compiled mixins that are used for the helm chart `Please build mixin by running 'make loki-mixin'`

alex5517:
@cstyan,

Done :)

cstyan:
Sorry @alex5517 because I missed your earlier updates it looks like we need to pull in `main` and update the compiled dashboards again. I can get this merged this week assuming we can get all the CI checks green :+1:

alex5517:
@cstyan, i have merged changes from main, and re-run make loki-mixin

cstyan:
@alex5517 sorry, the mixin check is still failing because it thinks there should be additional changes to the compiled mixin. It might be due to jsonnet versions? The check uses jsonnet v0.20 and doesn't use our build image (`Run make BUILD_IN_CONTAINER=false loki-mixin-check`). I still need to update our build image to use that version of jsonnet.

alex5517:
@cstyan,

Sorry for the delay, i synced the branch and ran the commands you asked.

production/loki-mixin/config.libsonnet
cstyan, line 22: can we make these names more explicit? even if it's long, I'd prefer to know all labels that are present from the name OR have to use each `per_x_label` for readability sake

cstyan (commented):
thanks for your patience @alex5517 , just one last comment I think

production/loki-mixin/config.libsonnet
alex5517: @cstyan,

I can change it, but would it not be better if it matches what mimir-mixin does, så that it easier for one who has used mimir-mixin to also use loki-mixin?

cstyan (approved):
cool, thanks @alex5517

PR #11505, opened 2023-12-18
structured metadata: track usage of structured metadata in queries

sandeepsukhani:
Adds support for tracking queries referencing structured metadata. It tracks the following kinds of references:
* Fields in structured metadata used as label filters in the query.
* Fields in structured metadata used in `by` or `without` clauses in aggregations.

**Checklist**
- [x] Tests updated


pkg/logql/log/labels.go
```diff
@@ -596,9 +607,12 @@ Outer:
 				continue Outer
 			}
 		}
-		for _, la := range b.add {
+		for category, la := range b.add {
 			for _, l := range la {
 				if g == l.Name {
+					if LabelCategory(category) == StructuredMetadataLabel {
+						b.referencedStructuredMetadata = true
+					}
```
vlad-diachenko, line 615: I believe these lines will set `b.referencedStructuredMetadata` to true whenever we have structured metadata extracted from the storage... even if the query does not use it. is it expected?

pkg/logql/log/labels_test.go
```diff
@@ -16,8 +16,19 @@ func TestLabelsBuilder_Get(t *testing.T) {
 	b.Reset()
 	b.Set(StructuredMetadataLabel, "foo", "bar")
 	b.Set(ParsedLabel, "bar", "buzz")
+
+	_, category, ok := b.GetWithCategory("bar")
```
vlad-diachenko, line 20: Can we ensure that `GetWithCategory` function is used only when we try to read the value of the label specified in the query?

vlad-diachenko (commented):
is it possible to add the testcases when the logs have the structured metadata but we do not use these labels in the query and assert that `referencedStructuredMetadata` is false? 
other than that, everything looks awesome!

pkg/logql/log/labels.go
sandeepsukhani: This is for queries referencing structured metadata in `by` clause for example if you have `foo` in structured metadata then the flag would be turned on for query `sum by (foo) (...)`

pkg/logql/log/labels_test.go
sandeepsukhani: Not sure if I understand the concern. `Get` and `GetWithCategory` are for getting value for a given label name.

pkg/storage/store_test.go
```diff
@@ -1831,3 +1842,210 @@ func TestStore_BoltdbTsdbSameIndexPrefix(t *testing.T) {
...
+				expectedReferencedStructuredMetadata: true,
+			},
+			{
+				name:  "referencing structured metadata in by aggregation clause",
+				query: `sum by (fizz) (count_over_time({sm="true"}[1m]))`,
+
+				expectedReferencedStructuredMetadata: true,
+			},
+			{
+				name:  "referencing structured metadata in without aggregation clause",
```
ashwanthgoli, line 2028: should we also include unwrap?

ashwanthgoli (approved):
lgtm

PR #11512, opened 2023-12-18
Tokenize line filters on a per-block basis

chaudum:

So far, the line filters from the user query have been tokenized to n-grams based on the per-tenant setting.
However, since the setting can change over time, block may be written with different n-gram sizes.

This PR adds the ability to read the n-gram length from the block schema at query time and tokenizes the line filters from the input just ahead of searching the block.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/storage/bloom/v1/test_util.go
+++ b/pkg/storage/bloom/v1/test_util.go
@@ -26,8 +26,10 @@ func MakeBlockQuerier(t testing.TB, fromFp, throughFp model.Fingerprint, fromTs,
 	builder, err := NewBlockBuilder(
 		BlockOptions{
 			schema: Schema{
-				version:  DefaultSchemaVersion,
-				encoding: chunkenc.EncSnappy,
+				version:     DefaultSchemaVersion,
+				encoding:    chunkenc.EncSnappy,
+				nGramLength: 4, // see DefaultNGramLength in bloom_tokenizer_test.go
+				nGramSkip:   0, // see DefaultNGramSkip in bloom_tokenizer_test.go
 			},
 			SeriesPageSize: 100,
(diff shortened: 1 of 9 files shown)


pkg/storage/bloom/v1/test_util.go
```diff
@@ -26,8 +26,10 @@ func MakeBlockQuerier(t testing.TB, fromFp, throughFp model.Fingerprint, fromTs,
 	builder, err := NewBlockBuilder(
 		BlockOptions{
 			schema: Schema{
-				version:  DefaultSchemaVersion,
-				encoding: chunkenc.EncSnappy,
+				version:     DefaultSchemaVersion,
+				encoding:    chunkenc.EncSnappy,
+				nGramLength: 4, // see DefaultNGramLength in bloom_tokenizer_test.go
+				nGramSkip:   0, // see DefaultNGramSkip in bloom_tokenizer_test.go
```
poyzannur, line 32: why not directly use those defaults? or move them here?

poyzannur (approved):
LGTM!
I think we need a reminder to check in [mergeBuilder](https://github.com/grafana/loki/blob/main/pkg/storage/bloom/v1/builder.go#L500), that we should not try to merge blocks tokenized with different `ngram` values. 
This can be done in a follow up PR.

pkg/storage/bloom/v1/test_util.go
chaudum: The test package cannot be imported into a regular package.

poyzannur: ah sorry, got it.

PR #11528, opened 2023-12-19
Add context and tenant to pipeline/extractor wrappers

MasslessParticle:
Adding the context here means the wrapping pipeline can know when the query has stopped or been cancelled.


pkg/ingester/instance.go
```diff
@@ -10,6 +10,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/grafana/dskit/tenant"
```
cstyan, line 13: nit: this can go below with the other "loki external" imports

cstyan (approved):
A description of why we need the context for adaptive logs would be helpful

PR #11535, opened 2023-12-20
Query-frontend: customisable query splitting for queries overlapping `query_ingester_within` window

dannykopping:
The config option `query_ingesters_within` defines the window during which logs _could_ be present on ingesters, and as such queriers will send queries to ingesters instead.

`split_queries_by_interval` is defined to split queries into subqueries for increased parallelism.

Aggressive query splitting within the `query_ingesters_within` window can result in overloading ingesters with unnecessarily large numbers of subqueries, which perversely can impact writes.

`query_ingesters_within` is set to 3h by default. In Grafana Cloud Logs we set `split_queries_by_interval` as low as 15m (defaults to 1h), which would result in result in 3*60/15=12 requests. Every querier queries every ingester during this window, so that's 12 requests _per ingester per query_ which has the `query_ingesters_within` window in its time range _(i.e. a query from now to now-7d would include the `query_ingesters_within` window as well, now-3h to now-7d would not)_.

However, we _do_ want to split queries so an ingester won't have to handle a query for a full `query_ingesters_within` window - this could involve a large amount of data. To account for this, this PR introduces a new option `split_ingester_queries_by_interval` on the query-frontend; this setting is disabled by default.

image

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/querier/queryrange/splitters.go
```diff
@@ -0,0 +1,269 @@
...
+
+		return reqs, nil
+	}
+
+	// Treat range of given query within the `query_ingesters_within` window differently by splitting using the `split_ingester_queries_by_interval`
+	// instead of the default `split_queries_by_interval`; rebound the start/end time after doing so to build intervals
+	// for queries outside the `query_ingesters_within` window.
+	//
+	// The given factory is responsible for building the splits and appending to reqs.
+	newStart, newEnd := buildIngesterQuerySplitsAndRebound(execTime, s.limits, s.iqo, tenantIDs, lokiReq, factory, false)
```
ashwanthgoli, line 200: for metric queries, this does not round the split boundaries to the step. is that intentional?

pkg/querier/queryrange/splitters.go
```diff
@@ -0,0 +1,269 @@
...
+	if newStart.Equal(newEnd) {
+		// Nothing to do, there are no more intervals to process.
+		return reqs, nil
+	}
+
+	// copy the splits, reset the results
+	ingesterSplits := reqs
+	reqs = nil
+
+	for start := newStart; start.Before(newEnd); start = s.nextIntervalBoundary(start, r.GetStep(), interval).Add(time.Duration(r.GetStep()) * util.SplitGap) {
```
ashwanthgoli, line 210: why not leave this as `time.Millisecond` since we are essentially converting it to ns?
I feel `SplitGap` is not the right thing to use here

ashwanthgoli:
something to be aware of: caching middlewares downstream expect the sub-queries to be split and aligned by `split_queries_by_interval` or other related configs based on the request type. This change breaks that assumption.

for ex: a cache key spanning an interval of 15m might store a larger 1h extent.
this may not affect the correctness for metric queries since we extract what we need, but it stores results that do not belong to the cache key.

pkg/querier/queryrange/splitters.go
dannykopping: You're right, I can't use this function here; thanks! I'll fix this.

dannykopping: Great catch on this @ashwanthgoli! I've addressed it in https://github.com/grafana/loki/pull/11535/commits/b7dfa09cbf74112d4d9a0e572ade02e36207ec36, PTAL.

pkg/querier/queryrange/splitters.go
```diff
@@ -0,0 +1,292 @@
...
+
+	start, end, needsIngesterSplits = ingesterQueryBounds(execTime, s.iqo, lokiReq)
+	start, end = s.alignStartEnd(r.GetStep(), start, end)
+
+	if ingesterQueryInterval := validation.MaxDurationPerTenant(tenantIDs, s.limits.IngesterQuerySplitDuration); ingesterQueryInterval != 0 && needsIngesterSplits {
+		// perform splitting using special interval (`split_ingester_queries_by_interval`)
+		s.buildMetricSplits(lokiReq.GetStep(), ingesterQueryInterval, start, end, factory)
+
+		// rebound after ingester queries have been split out
+		end = start
```
ashwanthgoli, line 220: we should also substract step from end to leave a step gap on the ingester split boundary similar to what we do with each split.


https://github.com/grafana/loki/blob/b7dfa09cbf74112d4d9a0e572ade02e36207ec36/pkg/querier/queryrange/splitters.go#L155

ashwanthgoli:
@dannykopping should we also make the cache layer aware of the ingester splits and pick a different interval for the cache key based on the query range being served?

>  I guess we shouldn't be caching metric queries within query_ingesters_within since the values can change anyway while logs are in the head block?

problem with this is that the current defaults for max_cache_freshness are much smaller, sub 30m. this might have some performance impact

pkg/querier/queryrange/splitters.go
dannykopping: Implemented in https://github.com/grafana/loki/pull/11535/commits/75ee0159621c0eb5d96890fcf7076192a75d2389, good catch!

dannykopping:
> @dannykopping should we also make the cache layer aware of the ingester splits and pick a different interval for the cache key based on the query range being served?
> 
> > I guess we shouldn't be caching metric queries within query_ingesters_within since the values can change anyway while logs are in the head block?
> 
> problem with this is that the current defaults for max_cache_freshness are much smaller, sub 30m. this might have some performance impact

Great idea! I'll do this in a follow-up PR.

ashwanthgoli (approved):
lgtm

pkg/querier/queryrange/splitters.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// move the ingester splits to the end to maintain correct order
+	reqs = append(reqs, ingesterSplits...)
+
+	return reqs, nil
+}
+
+func (s *metricQuerySplitter) alignStartEnd(step int64, start, end time.Time) (time.Time, time.Time) {
+	// step align start and end time of the query. Start time is rounded down and end time is rounded up.
+	stepNs := step * 1e6
```
vlad-diachenko, line 252: fingers crossed that step is `ms` ;)

pkg/validation/limits.go
```diff
@@ -299,6 +300,9 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	_ = l.MetadataQuerySplitDuration.Set("24h")
 	f.Var(&l.MetadataQuerySplitDuration, "querier.split-metadata-queries-by-interval", "Split metadata queries by a time interval and execute in parallel. The value 0 disables splitting metadata queries by time. This also determines how cache keys are chosen when label/series result caching is enabled.")
 
+	_ = l.IngesterQuerySplitDuration.Set("0s")
+	f.Var(&l.IngesterQuerySplitDuration, "querier.split-ingester-queries-by-interval", "Interval to use for time-based splitting when a request is within the `query_ingesters_within` window; defaults to `split-queries-by-interval` by setting to 0.")
```
vlad-diachenko, line 304: how do we fallback to `split-queries-by-interval` in case it is set to 0? just curious because I can not find it.

vlad-diachenko (commented):
looks awesome 🚀

pkg/validation/limits.go
dannykopping: https://github.com/grafana/loki/pull/11535/files#diff-5ae98add35d728910d8469070693a406831e482d1597d2a249303fa39d26ec43R120-R123

Effectively we don't handle the splits specially if the limit is 0, which uses `split-queries-by-interval`.

PR #11536, opened 2023-12-20
Bloom Compactor: Fix panic at startup if multiple TSDB period configs are defined

chaudum:
### Bloom Compactor: Fix duplicate metrics registration

If there are multiple period configs with TSDB, then each shipper
instance needs to register their metrics with a unique static label set,
otherwise, the bloom compactor would panic at startup due to duplicate
metric registration attempt.

### Bloom compactor: Start indexshipper in RO mode

We don't want/need to upload any index files in the bloom compactor, but
we only download files.

### DIFF
--- a/pkg/bloomcompactor/bloomcompactor.go
+++ b/pkg/bloomcompactor/bloomcompactor.go
@@ -114,6 +114,8 @@ func New(
 
 	c.storeClients = make(map[config.DayTime]storeClient)
 
+	indexShipperReg := prometheus.WrapRegistererWithPrefix("loki_bloom_compactor_tsdb_shipper_", r)
+
 	for i, periodicConfig := range schemaConfig.Configs {
 		if periodicConfig.IndexType != config.TSDBType {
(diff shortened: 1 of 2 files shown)


pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -131,6 +133,16 @@ func New(
...
 
+		pReg := prometheus.WrapRegistererWith(
+			prometheus.Labels{
+				"component": fmt.Sprintf(
+					"index-store-%s-%s",
+					periodicConfig.IndexType,
+					periodicConfig.From.String(),
+				),
+			}, indexShipperReg)
+		pLogger := log.With(logger, "index-store", fmt.Sprintf("%s-%s", periodicConfig.IndexType, periodicConfig.From.String()))
```
chaudum, line 144: Borrowed from https://github.com/grafana/loki/blob/main/pkg/storage/store.go#L269-L277

poyzannur (approved):

PR #11537, opened 2023-12-20
Fix panic in bloom gateway client

chaudum:
The bloom gateway client panicked, because the ring, from which it gets the subring for the tenant was never passed as argument in the constructor, even though it was part of the client struct.


poyzannur (approved):
lgtm

PR #11542, opened 2023-12-21
Add labels to bloom tokenizer metrics

poyzannur:
Bloom-compactor panics for duplicate metrics registration in bloom-tokenizers.
This PR adds labels to bloom-tokenizer metrics, adopting same solution in https://github.com/grafana/loki/pull/11536


pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -486,6 +481,22 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
...
+	NGramSkip := c.limits.BloomNGramSkip(job.tenantID)
+
+	btReg := prometheus.WrapRegistererWithPrefix("loki_bloom_tokenizer", c.reg)
+	pReg := prometheus.WrapRegistererWith(
+		prometheus.Labels{
+			"component": fmt.Sprintf(
+				"bloom-tokenizer-%s-%s",
+				job.tenantID,
+				job.tableName,
+			),
```
chaudum, line 495: This would create quite a high cardincality label.
Could we instead instantiate the metrics for the tokenizer once, store it on the bloom-compactor and then pass it like you do now.

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -486,6 +481,22 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
...
+		}, btReg)
+
+	bt, _ := v1.NewBloomTokenizer(pReg, NGramLength, NGramSkip)
```
chaudum, line 498: I would pass the more important arguments first.

```suggestion
	bt, _ := v1.NewBloomTokenizer(NGramLength, NGramSkip, reg)
```

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -357,7 +361,7 @@ func (c *Compactor) compactTenant(ctx context.Context, logger log.Logger, sc sto
 	// Tokenizer is not thread-safe so we need one per goroutine.
 	NGramLength := c.limits.BloomNGramLength(tenant)
 	NGramSkip := c.limits.BloomNGramSkip(tenant)
-	bt, _ := v1.NewBloomTokenizer(c.reg, NGramLength, NGramSkip)
+	bt, _ := v1.NewBloomTokenizer(NGramLength, NGramSkip, c.btMetrics)
```
chaudum, line 364: We should check the returned error.

chaudum (commented):
Otherwise LGTM

pkg/bloomcompactor/bloomcompactor.go
poyzannur: I removed the error return value instead. There's wasn't anything to return an error in `NewBloomTokenizer `, creates a map, creates . that's why we never checked it anywhere, seems like an afterthought.

chaudum (approved):

tools/tsdb/bloom-tester/tokenizer.go
```diff
@@ -0,0 +1,220 @@
+package main
```
poyzannur, line 1: With my changes, it became tricky to use v1.BloomTokenizer because it uses v1.Metrics vs the bloom tester uses it's own metrics. 

I choose an easy solution to copy the file. As the project gets in shape, we don't rely on this tool to benchmark blooms anymore.

PR #11546, opened 2023-12-21
[bloom-compactor] remove BloomCompactorMinTableAge check

poyzannur:
`BloomCompactorMinTableAge` was added with the idea that we will skip most recent index files from ingesters that are not compacted into TSDB indexes yet. 

The default value is an hour. This blocks bloom-compacter processing TSBD of the day, because end timestamp is always in the last 15 mins. We can either reduce it something lower than the frequency of TSDB indexes being built (< 15mins). 
Here I choose to remove it altogether, assuming uncompacted indexes from ingesters will not be processed as a table as there won't be a schema for that table with the [check here](https://github.com/grafana/loki/blob/main/pkg/bloomcompactor/bloomcompactor.go#L268-L272).


pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -307,12 +305,7 @@ func (c *Compactor) compactUsers(ctx context.Context, logger log.Logger, sc stor
 
 		// Skip this table if it is too new/old for the tenant limits.
 		now := model.Now()
-		tableMinAge := c.limits.BloomCompactorMinTableAge(tenant)
```
chaudum, line 310: Please also remove per-tenant setting as it is not used any more.

paul1r (approved):

Issue #7103, opened 2022-09-08
Promtail - docker_sd re-sending logs on container shutdown


heapdavid:
**Describe the bug**
Hi,

We're using `docker_sd_config` to scrape docker containers using json-file logging. The symptoms we're seeing when a container shuts down are:
- Recent logs are duplicated once the container has been shut down
- Any older logs cause the following error example:
  - `final error sending batch server returned HTTP status 400 Bad Request (400): entry with timestamp 2022-09-07 08:32:22.028843795 +0000 UTC ignored, reason: 'entry too far behind' for stream: `
  - related to `max_chunk_age`/2 as far as I can tell, as per https://github.com/grafana/loki/issues/5936#issuecomment-1102921754

**To Reproduce**
1. Configure Promtail (v2.6.1) with docker_sd_config as per config below
2. Run a docker container (for longer than `max_chunk_age`/2 since the most recent log entry if you want to reproduce the 2nd symptom)
3. Check the logs don't contain duplicates, then stop or restart the container.
4. Check the logs for the container, which now show duplicates.
5. If long running container as above, check the logs for promtail itself, which shows the 400 error.

**Expected behavior**
Promtail should only send the logs once.

**Environment:**
- Loki & Promtail v2.6.1
- Docker v20.10.17 (reproduced on both Swarm clusters and stand-alone instances)

**Screenshots, Promtail config, or terminal output**
```
  - job_name: docker
    docker_sd_configs:
      - host: unix:///var/run/docker.sock
        refresh_interval: 5s
    relabel_configs:
      - regex: '__meta_docker_container_label_com_(docker_compose_(project|service))'
        replacement: '$1'
        action: labelmap
      - source_labels: [__meta_docker_container_label_com_docker_swarm_service_name]
        target_label: swarm_service
      - source_labels: [__meta_docker_container_label_com_docker_stack_namespace]
        target_label: swarm_stack
      - source_labels: [__meta_docker_container_name]
        regex: '/(.*)'
        target_label: container
    pipeline_stages:
      - docker: {}
      - static_labels:
          job: docker
```

heapdavid:
Just spotted that we get 2 `"finished transferring logs"` messages logged when this happens:

```
[...]

PR #11563, opened 2023-12-29
Fix duplicate logs from docker containers

### LINKED ISSUE #7382: Duplicate log entries for short-lived container
## **Describe the bug**
I have a container for which I would like to get the logs but it is short-lived. A host cron job wakes it up every minute with a `docker start <container>`. The container checks in a mounted folder if it has work to do.  
If it has no work to do, nothing is logged and the container goes back to sleep almost instantly.  
If it has work to do, it will do its job and log lots of thing to the standard output.  

The log entries for my short-lived container are duplicated several times. It seems like I get a duplicate for everytime the container wakes up without work to do.  All log entries have **exactly** the same `tsNs`.

The Promtail documentation mentions the following

> Please note that the discovery will not pick up finished containers. That means Promtail will not scrape the remaining logs from finished containers after a restart.

Maybe it can be related to my bug? I'm not sure.  

This issue could also be related to the following ones which all mention duplication problems:
- https://github.com/grafana/loki/issues/7103. This one could very well be my own problem.
- https://github.com/grafana/loki/issues/7207
- https://github.com/grafana/loki/issues/7141
- https://github.com/grafana/loki/issues/6929
- https://github.com/grafana/loki/issues/6790
- https://github.com/grafana/loki/issues/6225
- https://github.com/grafana/loki/issues/5252

## **To Reproduce**
Here are my Loki and Promtail services in docker-compose:
```yml
  loki:
    image: grafana/loki:2.6.0
    container_name: loki
    restart: unless-stopped
    user: 1001:1001
    command: -config.file=/etc/loki/config.yml
    volumes:
      - ./config/loki:/etc/loki
      - ./storage/loki:/tmp/loki
  promtail:
    image: grafana/promtail:2.6.0
    container_name: promtail
    restart: unless-stopped
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
      - ./config/promtail/config.yml:/etc/promtail/config.yml:ro
      - ./storage/promtail:/storage/promtail
    command:
      - "-config.file=/etc/promtail/config.yml"
```

## **Expected behavior**
I should not have duplicate log entries.

## **Environment:**
 - Infrastructure: Bare metal.
 - Deployment tool: Docker/docker-compose.

## **Screenshots, Promtail config, or terminal output**
The following config files:

**Promtail config:**
```
server:
  disable: true

positions:
  filename: /storage/promtail/positions.yml

clients:
  - url: https://loki.example.com/loki/api/v1/push

scrape_configs:
  - job_name: docker
    docker_sd_configs:
      - host: unix:///var/run/docker.sock
        refresh_interval: 5s
        filters:
          - name: label
            values:
              - company.logging.enable=True
    relabel_configs:
      - target_label: env
        replacement: dev

      - source_labels:
        - __meta_docker_container_label_com_docker_compose_project_working_dir
        target_label: env
        regex: ".*(stage1).*"

      - source_labels:
        - __meta_docker_container_label_com_docker_compose_project_working_dir
        target_label: env
        regex: ".*(stage2).*"
```

**Loki Config:**
```
auth_enabled: false

server:
  http_listen_port: 3100

ingester:
  lifecycler:
    address: 127.0.0.1
    ring:
      kvstore:
        store: inmemory
      replication_factor: 1
    final_sleep: 0s
  chunk_idle_period: 5m
  chunk_retain_period: 30s
  wal:
    dir: /tmp/loki/wal

schema_config:
  configs:
  - from: 2020-05-15
    store: boltdb
    object_store: filesystem
    schema: v11
    index:
      prefix: index_
      period: 24h

storage_config:
  boltdb:
    directory: /tmp/loki/index

  filesystem:
    directory: /tmp/loki/chunks

compactor:
  working_directory: /tmp/loki/compactor
  shared_store: filesystem
  compaction_interval: 10m
  retention_enabled: true
  retention_delete_delay: 2h

limits_config:
  enforce_metric_name: false
  reject_old_samples: true
  retention_period: 30d

chunk_store_config:
  max_look_back_period: 14d

table_manager:
  retention_deletes_enabled: true
  retention_period: 30d
```

ptqa:
This fixes various issues with docker_sd on promtail. Mostly related to duplicate logs being send to loki from promtail.

Root case of issue is that positions file is updated after process function, but in memory field `since` of `Target` struct is not updated.

Fixes #7382 #7103 and some others

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


choooze:
bump

cstyan:
A test should be added here to show the bug and that this is fixing it.

ptqa:
@cstyan I would like to point out few things:

1. This bug was confirmed a year ago by one of the contributors (see https://github.com/grafana/loki/issues/7382#issuecomment-1280484688)
2. Since we only can test docker API interactions via mock API, to reproduce this bug with mock API I would need to:
* fully reproduce `log` endpoint behavior with timestamp support
* make mock API dynamic, so it would simulate container start/stop/start

This is non-trivial and requires lots of work for a simple bug in logic. Feel like too much for this.

cstyan:
Given that it's been more than a year since someone on our team looked into it and we don't use the promtail with THE docker target ourselves (and actually aren't running promtail anymore, we run https://github.com/grafana/agent so this fix is likely relevant to them as well), some kind of a test to validate the behaviour feels like a reasonable ask.

On paper this does seem like the right fix, your explanation of the root cause makes sense. There's an existing test in `target_test.go` that creates an instance of the docker client, a positions file, and a set of fake data to send to the client via a mock entry handler. It seems like this could be extended to then stop the targets process loop, start it again, resend the same dataset, and ensure that we threw away all of those log lines because `t.since` was set properly. If you take a look at everything here and it still doesn't seem feasible to add a test then please let me know.

ptqa:
@cstyan gotcha, your suggested approach indeed is simpler and doesn't require this much work. I've added tests to this bug, can you check?

cstyan:
Very nice, confirmed the test fails without the `t.since` change and passes with it. Just waiting to confirm that the drone build is green :+1:

cstyan (approved):
green, lgtm

cstyan:
@ptqa not sure what the cause of the conflict is but since there isn't a file specified my first assumption is that it's within your fork somehow? I'd suggest merging `grafana/loki:main` into your `lidofinance:main` branch.

ptqa:
@cstyan yeah it's a CHANGELOG file, I've merged main branch

cstyan:
@ptqa awesome, thanks for your contribution and patience

ptqa:
@cstyan thank you for the guidance and merging this PR

PR #11566, opened 2024-01-02
Bloomshipper: Use `model.Time` in `MetaRef` and `BlockRef`

chaudum:

The `Ref` of `MetaRef` and `BlockRef` do have start and end timestamps, which are of type `int64`.
This however, can lead to the problem that different components might use different precisions (`s`, `ms`, `ns`) for these timestamps.

This PR changes the current implementation of the bloom shipper to use `ms` precision timestamps (using `model.Time`) instead of `s`, to match the behaviour that we have to encode ChunkRefs to object store keys.


pkg/storage/stores/shipper/bloomshipper/client.go
```diff
@@ -33,20 +33,28 @@ type Ref struct {
 	TenantID                       string
 	TableName                      string
 	MinFingerprint, MaxFingerprint uint64
-	StartTimestamp, EndTimestamp   int64
+	StartTimestamp, EndTimestamp   int64 // must be millisecond precision timestamp aka model.Time
```
chaudum, line 36: It would make even more sense to use `model.Time` type.

poyzannur: I agree 👍

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -472,10 +472,10 @@ func (c *Compactor) runCompact(ctx context.Context, logger log.Logger, job Job,
 	}
 	metaSearchParams := bloomshipper.MetaSearchParams{
 		TenantID:       job.tenantID,
-		MinFingerprint: uint64(job.minFp),
-		MaxFingerprint: uint64(job.maxFp),
-		StartTimestamp: int64(job.from),
-		EndTimestamp:   int64(job.through),
+		MinFingerprint: job.minFp,
+		MaxFingerprint: job.maxFp,
```
poyzannur, line 476: doesn't this complain? `min/maxFp` are `model.Fingerprint`

pkg/storage/stores/shipper/bloomshipper/client.go
```diff
@@ -67,11 +67,9 @@ type Meta struct {
 }
 
 type MetaSearchParams struct {
-	TenantID       string
-	MinFingerprint uint64
-	MaxFingerprint uint64
-	StartTimestamp int64
-	EndTimestamp   int64
+	TenantID                       string
+	MinFingerprint, MaxFingerprint model.Fingerprint
```
poyzannur, line 71: I think it's good to use `model.Fingerprint` here instead of `uint64`

If you are changing `fp` here, maybe change above in `Ref` struct too?

chaudum: Yes, we can change it in the Ref as well. I would create a separate PR for that, though.

poyzannur (approved):
lgtm

PR #11571, opened 2024-01-03
Add metrics.go log line when reading from an ingester

MichelHollands:

Add a metrics.go log line when a querier reads from an ingester. The output is like this:

```
level=info ts=2024-01-03T13:59:06.919965546Z caller=metrics.go:275 component=ingester org_id=fake traceID=0b216e7d014c5f87 latency=fast query_type=ingester_series start=2024-01-03T13:00:00Z end=2024-01-03T13:59:06.834Z start_delta=59m6.91996413s end_delta=85.964255ms length=59m6.834s duration=110.792µs status=200 query=<omitted> query_hash=2166136261 total_entries=9
```

**Checklist**
- [ ] Documentation added
- [X] Tests updated
- [X] `CHANGELOG.md` updated


pkg/ingester/ingester.go
```diff
@@ -871,10 +871,13 @@ func (i *Ingester) GetOrCreateInstance(instanceID string) (*instance, error) { /
 }
 
 // Query the ingests for log streams matching a set of matchers.
-func (i *Ingester) Query(req *logproto.QueryRequest, queryServer logproto.Querier_QueryServer) error {
+func (i *Ingester) Query(req *logproto.QueryRequest, queryServer logproto.Querier_QueryServer) (err error) {
 	// initialize stats collection for ingester queries.
 	_, ctx := stats.NewContext(queryServer.Context())
 
+	start := time.Now()
```
JordanRushing, line 878: Should this be UTC?

pkg/ingester/ingester.go
```diff
@@ -885,6 +888,17 @@ func (i *Ingester) Query(req *logproto.QueryRequest, queryServer logproto.Querie
...
 	}
 
+	defer func() {
+		status := "200"
+		if err != nil {
+			status = "400"
+		}
+		statsCtx := stats.FromContext(ctx)
+		execTime := time.Since(start)
+		logql.RecordIngesterSeriesQueryMetrics(ctx, i.logger, req.Start, req.End, req.Selector, status,
```
dannykopping, line 898: I don't think this is a series request; this is a streams request.
`QuerySample`, which should also have logging I believe (?), returns series.

pkg/logql/metrics.go
```diff
@@ -243,6 +244,37 @@ func PrintMatches(matches []string) string {
...
+
+	// Tag throughput metric by latency type based on a threshold.
+	// Latency below the threshold is fast, above is slow.
+	if stats.Summary.ExecTime > slowQueryThresholdSecond {
+		latencyType = latencyTypeSlow
+	}
+
+	logValues := make([]interface{}, 0, 11)
+	logValues = append(logValues,
+		"latency", latencyType,
```
dannykopping, line 262: Would be useful to log the number of lines here, too
If we have the bytes returned as well that'd be very valuable

pkg/ingester/ingester.go
MichelHollands: Good catch

pkg/ingester/ingester.go
```diff
@@ -885,6 +888,17 @@ func (i *Ingester) Query(req *logproto.QueryRequest, queryServer logproto.Querie
 		}
 	}
 
+	defer func() {
+		status := "200"
+		if err != nil {
+			status = "400"
```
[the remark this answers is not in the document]
MichelHollands: Yeah, I'm not sure what to use here otherwise. The other logging functions with a status field are called from a HTTP handler. Here that's not the case. Perhaps I should use an error field.

pkg/ingester/ingester.go
MichelHollands: Good idea, I've renamed it and also added logging to QuerySample.

pkg/logql/metrics.go
MichelHollands: Good idea as well. I've added as many as possible.
Note that the nr of lines was stored in the stats type before. Now it's a separate field to make it consistent with the other logging functions.

pkg/ingester/ingester.go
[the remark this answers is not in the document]
dannykopping: Yeah it's an open string so feel free to use something more specific like "successful"/"failed".

dannykopping (approved):
LGTM 👍

PR #11581, opened 2024-01-04
Update API docs

### LINKED ISSUE #11565: Docs feedback: /docs/sources/reference/api.md
when using /loki/api/v1/push, the timestamp needs to be sent as a string and not a number, otherwise the endpoint will return a 400 error

JStickler:
Fixes #11565


MichelHollands (approved):
LGTM

PR #11588, opened 2024-01-05
builder baseMap concurrent write fix lock

cstyan:
It looks like with either sufficiently large line formats, or line formats that have some kind of operation that results in two line formats being executed concurrently (somehow, I do not understand how), it's possible for us to end up with a concurrent write as part of `LineFormatter.Process` calls. 

This AFAICT existed before https://github.com/grafana/loki/pull/11484 but a query was executed in one of our environments today that made us think the panic was from those changes. I suspect the baseMap usage is the underlying cause, and can confirm by deploying the changes from here into one of our environments tomorrow and running the same query I saw today that caused the panic.

### DIFF
--- a/pkg/logql/log/fmt.go
+++ b/pkg/logql/log/fmt.go
@@ -221,8 +221,8 @@ func (lf *LineFormatter) Process(ts int64, line []byte, lbs *LabelsBuilder) ([]b
 	lf.currentLine = line
 	lf.currentTs = ts
 
-	// map now is taking from a pool
-	m := lbs.Map()
+	m := smp.Get()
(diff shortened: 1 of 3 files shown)


dannykopping (commented):
Left a couple notes, but what's unclear to me is why this would address the issue fully... The panic occurs at `pkg/logql/log/labels.go:468` which is when the newly-retrieved map from the pool is `clear`-ed, suggesting that we're putting a map back into the pool without removing all references to it. We can add the locks and that'll make the problem go away, but is the underlying issue not the references?

cstyan:
superseded by https://github.com/grafana/loki/pull/11607

PR #11589, opened 2024-01-05
results cache: add query_length_served

ashwanthgoli:
cache hit rate that is currently being measured using metrics and stats does not account for the fact that the hit could only return partial results. When we query that cache for a key, we get back are a list of extents and these need no cover the entire (split) range of the cache key.

This pr adds a new cache stat called `query_length_served` to better measure the cache effectiveness.
```
query_length - sum(length of downstream queries)
```

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


dannykopping (commented):
Trying to understand: how would this will be used to measure cache effectiveness? Would we use this field and perform some arithmetic against the query length to determine the overall "hit rate"?

On its own it'll be difficult to use as a measure of cache effectiveness, I feel.

ashwanthgoli:
> Would we use this field and perform some arithmetic against the query length to determine the overall "hit rate"?

yes @dannykopping, this is what I was thinking. Just logging the actual number same as what we do with other stats like timings.

dannykopping:
Gotcha, ok. Can we even do arithmetic on durations in this way?

ashwanthgoli:
not pretty, but that should do it I think
```
divf (.cache_label_results_query_length_served | duration) (.length | duration))
```

dannykopping:
> not pretty, but that should do it I think
> 
> ```
> divf (.cache_label_results_query_length_served | duration) (.length | duration))
> ```

Seems plausible, can you confirm before we merge?

dannykopping (approved):
LGTM

PR #11592, opened 2024-01-05
operator: Refactor CreateOrUpdateLokiStack handler

periklis:
Refactor the `CreateOrUpdateLokiStack` handler to separate functions to re-instantiate readability.

Fixes #<issue number>

The gist is moving and isolating all internal packages code into the internal package itself for gateway, storage and rules. This enhances readability in the handler itself. In addition the internal package's public interface is down to 1 or max 2 functions (i.e. `{gateway,storage,rules}.BuildOptions`) and the use cases can be tested locally w/o the heavy handler setup code.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

cc @JoaoBraveCoding


operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -414,6 +221,252 @@ func CreateOrUpdateLokiStack(
...
+	if err != nil {
+		log.Error(err, "failed to check OCP AlertManager")
+		return nil, nil, ruler, ocpOpts, err
+	}
+
+	ocpUWAmEnabled, err := openshift.UserWorkloadAlertManagerSVCExists(ctx, stack.Spec, k)
+	if err != nil {
+		log.Error(err, "failed to check OCP User Workload AlertManager")
+		return nil, nil, ruler, ocpOpts, err
+	}
```
JoaoBraveCoding, line 453: Minor comment, it looks weird that the RulerOptions generate the OCP Options, although they are related, however, I also have a hard time justifying a function with only these two steps. Maybe in the future, it will become more evident

JoaoBraveCoding (commented):
Minor comment but apart from that it doesn't make `CreateOrUpdateLokiStack` simpler on the eyes

operator/internal/handlers/internal/storage/storage.go
```diff
@@ -0,0 +1,91 @@
...
+
+	configv1 "github.com/grafana/loki/operator/apis/config/v1"
+	lokiv1 "github.com/grafana/loki/operator/apis/loki/v1"
+	"github.com/grafana/loki/operator/internal/external/k8s"
+	"github.com/grafana/loki/operator/internal/manifests/storage"
+	"github.com/grafana/loki/operator/internal/status"
+)
+
+// BuildOptions returns the object storage options to generate Kubernetes resource  manifests
+// which require access to object storage buckets. Or else it returns  a degraded error if one
```
JoaoBraveCoding, line 16:
```suggestion
// which require access to object storage buckets. Or else it returns a degraded error if one
```

operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -1,24 +1,46 @@
...
+	"github.com/grafana/loki/operator/internal/external/k8s"
 	"github.com/grafana/loki/operator/internal/manifests/storage"
+	"github.com/grafana/loki/operator/internal/status"
 )
 
 var hashSeparator = []byte(",")
 
-// ExtractSecret reads a k8s secret into a manifest object storage struct if valid.
-func ExtractSecret(s *corev1.Secret, secretType lokiv1.ObjectStorageSecretType) (*storage.Options, error) {
+func getSecret(ctx context.Context, k k8s.Client, stack *lokiv1.LokiStack) (*corev1.Secret, error) {
```
JoaoBraveCoding, line 22: Thinking about error handling, if we plan for this function to be used to fetch other secrets other than the object storage one then, we should also provide a way to configure the DegradedError struct. 

I am however more of the opinion of changing the name of this function to be more specific and not allude to the idea of reusing it when we need to get secrets from k8s. Instead, we should create a new function for getting new secrets out of k8s.

JoaoBraveCoding (commented):
Yes! I like this version much more! Just a small comment but this is going in a great direction

operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -1,24 +1,46 @@
...
+			}
+		}
+		return nil, kverrors.Wrap(err, "failed to lookup lokistack storage secret", "name", key)
+	}
+
+	return &storageSecret, nil
+}
+
+// extractSecret reads a k8s secret into a manifest object storage struct if valid.
+func extractSecret(s *corev1.Secret, secretType lokiv1.ObjectStorageSecretType) (storage.Options, error) {
```
xperimental, line 40: Why don't we stick to the pointered return type?

periklis: Because when we pass back to the `manifests.Options` we use a non-pointered version.

operator/internal/handlers/internal/rules/cleanup.go
```diff
@@ -4,25 +4,50 @@ import (
...
-// RemoveRulesConfigMap removes the rules configmaps if any exists.
-func RemoveRulesConfigMap(ctx context.Context, req ctrl.Request, c client.Client) error {
+// Cleanup removes the ruler component's statefulset and configmaps if available, or
+// else it returns an error to retry the reconciliation loop.
+func Cleanup(ctx context.Context, log logr.Logger, k k8s.Client, stack *v1.LokiStack) error {
+	if stack.Spec.Rules != nil && stack.Spec.Rules.Enabled {
+		return nil
+	}
+
+	stackKey := client.ObjectKeyFromObject(stack)
```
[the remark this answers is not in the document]
periklis: After checking I figured only one usage where we could skip passing the stack as pointer and pass the namespace instead namely:
```go
func getCAConfigMap(ctx context.Context, k k8s.Client, name, namespace string) (*corev1.ConfigMap, error) {
```

All other usages usually use the stack to traverse the spec and create custom object keys, e.g.:

```go
key := client.ObjectKey{Name: tenant.OIDC.Secret.Name, Namespace: stack.Namespace}

key := client.ObjectKey{Name: rulerConfig.RemoteWriteSpec.ClientSpec.AuthorizationSecretName, Namespace: stack.Namespace}
```

xperimental (approved):

PR #11595, opened 2024-01-05
fix: use syntax.MustClone to avoid panics when cloning expressions

MasslessParticle:
The clone function in rangemapper has been a persistent source of panics due to the way stringifying exprs changed the way the original query was represented. This PR replaces the `clone` function with `syntax.MustClone`. It clones the expression by actually walking the AST rather than doing `expr -> string -> parse`. It  might still panic but for a more fundamental reason than because of issues with the string representation.


dannykopping (commented):
This makes a whole lotta sense!
Do we have tests which cover this? I didn't see any in `pkg/logql/rangemapper_test.go` in a very quick scan.

MasslessParticle:
@dannykopping  :+1: I've added a test that demonstrates the issue. It panics with the old `clone` but not now

dannykopping (approved):
LGTM, thanks

pkg/logql/rangemapper.go
```diff
@@ -120,7 +120,7 @@ func (m RangeMapper) Parse(expr syntax.Expr) (bool, syntax.Expr, error) {
 // is pushed down to the downstream expression.
 func (m RangeMapper) Map(expr syntax.SampleExpr, vectorAggrPushdown *syntax.VectorAggregationExpr, recorder *downstreamRecorder) (syntax.SampleExpr, error) {
 	// immediately clone the passed expr to avoid mutating the original
-	expr = clone(expr)
+	expr = syntax.MustClone(expr)
```
jeschkies, line 123: Nice check. I thought I've replaced all of them.

jeschkies (approved):

PR #11600, opened 2024-01-08
deletion: delete chunk overlapping multiple tables when processing oldest table that has it indexed

sandeepsukhani:
When a delete request has a line filter and/or is deleting a chunk partially, we create a new chunk from the source chunk and mark the original chunk for deletion. However, when a chunk overlaps multiple tables, we apply heuristics to retain the chunk until all the tables that have it indexed get processed since we need the original chunk to build a new one.

Earlier, we used to process tables from oldest to newest, so the logic to detect if all the tables indexing a chunk are processed was done by seeing if `chunk-end-time` <= `table-end-time`. To put it in simple terms, if the chunk ends in the current table we are processing, it means this is the last table that we are processing, which has the chunk indexed.

However, this does not work anymore since we changed the order of processing tables to process the newest tables first. To get it working back, I have updated the logic to `chunk-start-time` >= `table-start-time`. In simple terms, if the chunk starts in the current table we are processing, it means this is the last table we are processing, which has the chunk indexed.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/compactor/retention/retention_test.go
+++ b/pkg/compactor/retention/retention_test.go
@@ -175,11 +175,16 @@ func Test_Retention(t *testing.T) {
 	}
 }
 
-type noopWriter struct{}
+type noopWriter struct {
+	count int64
+}
 
-func (noopWriter) Put(_ []byte) error { return nil }
-func (noopWriter) Count() int64       { return 0 }
-func (noopWriter) Close() error       { return nil }
+func (n *noopWriter) Put(_ []byte) error {
+	n.count++
+	return nil
+}
+func (n *noopWriter) Count() int64 { return n.count }
+func (n *noopWriter) Close() error { return nil }
 
 func Test_EmptyTable(t *testing.T) {
 	schema := allSchemas[0]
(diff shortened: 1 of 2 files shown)


pkg/compactor/retention/retention_test.go
```diff
@@ -675,6 +684,9 @@ func TestMarkForDelete_SeriesCleanup(t *testing.T) {
 			expectedModified: []bool{
 				false,
 			},
+			numChunksDeleted: []int64{
+				0,
```
JoaoBraveCoding, line 688: by the `name` of the test case here shouldn't it be 1?

JoaoBraveCoding (commented):
Trying to help out with PR reviews & learn the project, if my comment doesn't make sense please disregard it

pkg/compactor/retention/retention_test.go
sandeepsukhani: Thanks for taking a look! This tc is testing a case where there is a delete request with a line filter but the filter is not matching any lines. Only the lines which are matched by filter would get deleted so essentially it is a no-op.

pkg/compactor/retention/retention_test.go
```diff
@@ -825,6 +852,9 @@ func TestMarkForDelete_SeriesCleanup(t *testing.T) {
 			expectedModified: []bool{
 				true, true,
 			},
+			numChunksDeleted: []int64{
+				1, 0,
+			},
```
ashwanthgoli, line 857: should we add another test here where the chunk is partially deleted from two tables, but only gets marked in the oldest one?

ashwanthgoli: had a chat with sandeep offline, this test already covers the above scenario. Without the chunk start time check, we would have marked the chunk from both tables after a rewrite

ashwanthgoli (approved):
lgtm

PR #11604, opened 2024-01-08
compactor" track number of times table compactions were skipped consecutively

sandeepsukhani:
Earlier we had a counter for the number of times tables were skipped by compaction due to them being locked by retention. Counter makes it hard to reliably write an alert without making it noisy. I am changing the counter to gauge which would count the number of times tables were consecutively skipped by compaction. This would make it easier to write alerts.

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/compactor/metrics.go
+++ b/pkg/compactor/metrics.go
@@ -60,7 +60,7 @@ func newMetrics(r prometheus.Registerer) *metrics {
 		}),
 		skippedCompactingLockedTables: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
 			Namespace: "loki_compactor",
-			Name:      "skipped_compacting_locked_table_consecutive",
+			Name:      "locked_table_successive_compaction_skips",
 			Help:      "Number of times uncompacted tables were consecutively skipped due to them being locked by retention",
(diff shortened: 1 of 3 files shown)


pkg/compactor/metrics.go
```diff
@@ -58,10 +58,10 @@ func newMetrics(r prometheus.Registerer) *metrics {
 			Name:      "compactor_running",
 			Help:      "Value will be 1 if compactor is currently running on this instance",
 		}),
-		skippedCompactingLockedTables: promauto.With(r).NewCounterVec(prometheus.CounterOpts{
+		skippedCompactingLockedTables: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
 			Namespace: "loki_compactor",
-			Name:      "skipped_compacting_locked_table_total",
-			Help:      "Count of uncompacted tables being skipped due to them being locked by retention",
+			Name:      "skipped_compacting_locked_table_consecutive",
```
ashwanthgoli, line 63:
```suggestion
			Name:      "locked_table_successive_compaction_skips",
```

ashwanthgoli (approved):
lgtm. minor naming suggestion

PR #11605, opened 2024-01-08
feat: improve performance of `first_over_time` and `last_over_time` queries by sharding them

jeschkies:
With the introduction of sending query plans with custom nodes to the querier we can now shard `first_over_time` queries.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/logql/first_over_time.go
```diff
@@ -0,0 +1,191 @@
...
+
+			if len(series.Floats) == 0 || !e.inRange(series.Floats[0].T, ts) {
+				continue
+			}
+
+			// Merge
+			if vec[j].T > series.Floats[0].T {
+				vec[j].F = series.Floats[0].F
+				vec[j].T = series.Floats[0].T
+			}
```
cstyan, line 149: This is the bit that selects the earliest timestamp out of all of the downstreamed queries. For `last_over_time` we reverse this to `if vec[j].T < series.Floats[0].T {` as we merge them together.

cstyan (commented):
I'm extended this to include implementing `last_over_time` [here](https://github.com/grafana/loki/compare/main...callum-shard-last).

My only comments would be:
- that renaming `first_over_time.go` to something more generic, or `first_last_over_time.go` like I did in my branch, probably makes sense
- some comments would be useful (such as where I commented on the PR) for first time readers of the code

pkg/logql/engine.go
```diff
@@ -419,15 +419,16 @@ func (q *query) JoinSampleVector(next bool, ts int64, r StepResult, stepEvaluato
 				seriesIndex[hash] = series
 			}
 			series.Floats = append(series.Floats, promql.FPoint{
-				T: ts,
+				//T: ts,
```
jeschkies, line 422: I'm not sure we can do it like this. What do you think, @cstyan?

pkg/logql/first_over_time.go
```diff
@@ -0,0 +1,189 @@
...
+		r.at = make([]promql.Sample, 0, len(r.window))
+	}
+	r.at = r.at[:0]
+	// convert ts from nano to milli seconds as the iterator work with nanoseconds
+	ts := r.current/1e+6 + r.offset/1e+6
+	for _, series := range r.window {
+		s := r.agg(series.Floats)
+		r.at = append(r.at, promql.Sample{
+			F:      s.F,
+			T:      s.T / int64(time.Millisecond),
```
jeschkies, line 48: Vlad mentioned we should be using nanos here. What do you think?

cstyan: I don't think it matters? When we return the result to users we interploate forwards in time to the next interval of the `Step` anyways. If we were going to change it, we should probably not use internal prometheus types, since prometheus is only ever millisecond precise. It might not be a problem right now, but we shouldn't rely on something _not_ changing in a way that would make that millisecond precision more strict in some way.

pkg/logql/first_over_time.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+	}
+
+	// Align vector timestamps with step
+	for i := range vec {
+		vec[i].T = ts
+	}
+
+	return true, ts, SampleVector(vec)
+}
```
cstyan, line 160: I think we could simplify this into one loop. At least, the tests still pass for me. I think however that we're missing a test case for when there's only a single matrix, in that case we're not calling the code later on in this function that aligns the timestamps with the step.

```
func (e *firstOverTimeStepEvaluator) Next() (bool, int64, StepResult) {

	var (
		vec promql.Vector
		ok  bool
	)

	// TODO: build index metric to vec pos

	e.ts = e.ts.Add(e.step)
	if e.ts.After(e.end) {
		return false, 0, nil
	}
	ts := e.ts.UnixNano() / int64(time.Millisecond)

	// Merge other results
	for i, m := range e.matrices {
		// TODO: verify length and same labels/metric
		for j, series := range m {

			if len(series.Floats) == 0 || !e.inRange(series.Floats[0].T, ts) {
				continue
			}

			// Merge
			if len(vec) < len(e.matrices) {
				vec = append(vec, promql.Sample{
					Metric: series.Metric,
					T:      series.Floats[0].T,
					F:      series.Floats[0].F,
				})
			} else if vec[j].T > series.Floats[0].T {
				vec[j].F = series.Floats[0].F
				vec[j].T = series.Floats[0].T
			}

			// We've omitted the first matrix. That's why +1.
			e.pop(i, j)
		}
	}

	if len(e.matrices) == 1 {
		return ok, ts, SampleVector(vec)
	}

	if len(vec) == 0 {
		return e.hasNext(), ts, SampleVector(vec)
	}

	// Align vector timestamps with step
	for i := range vec {
		vec[i].T = ts
	}

	return true, ts, SampleVector(vec)
}
```

cstyan: > I think however that we're missing a test case for when there's only a single matrix

I guess this would only happen if there's only a single series returned AND a single shard. Which I'm not sure is possible? But it's probably safer to guard against it just in case.

pkg/logql/first_over_time.go
```diff
@@ -0,0 +1,189 @@
...
+
+	var (
+		start = params.Start()
+		end   = params.End()
+		step  = params.Step()
+	)
+
+	index := make(map[int64]int, 0)
+	for i, series := range m[1] {
+		index[int64(series.Metric.Hash())] = i
```
cstyan, line 82: I don't think this and the `streamVec` field are needed, we set it but never use it.

cstyan (commented):
if you want you can pull in my branch and we can merge first and last sharding at the same time, I cleaned up some stuff based on my comments here in the latest commit https://github.com/grafana/loki/commit/f8c9c2bdbf2013cc151e986330263bf08dbeeefc

pkg/logql/downstream_test.go
```diff
@@ -132,7 +134,7 @@ func TestMappingEquivalenceSketches(t *testing.T) {
 		query         string
 		realtiveError float64
 	}{
-		{`quantile_over_time(0.70, {a=~".+"} | logfmt | unwrap value [1s]) by (a)`, 0.03},
+		{`quantile_over_time(0.70, {a=~".+"} | logfmt | unwrap value [1s]) by (a)`, 0.05},
```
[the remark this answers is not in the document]
jeschkies: When I change the test series to have the values spread out this would fail.

pkg/logql/first_over_time.go
```diff
@@ -0,0 +1,189 @@
...
+	selRange, step, start, end, offset int64) RangeVectorIterator {
+	inner := &batchRangeVectorIterator{
+		iter:     it,
+		step:     step,
+		end:      end,
+		selRange: selRange,
+		metrics:  map[string]labels.Labels{},
+		window:   map[string]*promql.Series{},
+		agg:      nil,
+		current:  start - step, // first loop iteration will set it to start
```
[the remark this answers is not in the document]
jeschkies: This is how it's in the other cases. I think this is super brittle but couldn't think of a nicer way.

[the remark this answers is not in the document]
cstyan: > This is how it's in the other cases. I think this is super brittle but couldn't think of a nicer way.

Okay, I'll create an issue for that as a follow up, if this is following the existing pattern then lets just move forward with this as we have it now.

pkg/logql/downstream_test.go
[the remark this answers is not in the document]
cstyan: which change is that? the addition of nanos in the timestamp within `test_utils.go`?

pkg/logql/downstream.go
```diff
@@ -394,6 +422,42 @@ func (ev *DownstreamEvaluator) NewStepEvaluator(
...
+		}
+
+		xs := make([]promql.Matrix, 0, len(queries))
+		for _, res := range results {
+
+			switch data := res.Data.(type) {
+			case promql.Matrix:
+				xs = append(xs, data)
+			default:
+				return nil, fmt.Errorf("unexpected type (%s) uncoercible to StepEvaluator", data.Type())
```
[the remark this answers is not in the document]
jeschkies: It would be a matrix with one vector I believe. Only for instant queries we return a vector https://github.com/grafana/loki/blob/main/pkg/logql/engine.go#L388

pkg/logql/downstream_test.go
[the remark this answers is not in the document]
jeschkies: Yes.

pkg/logql/engine.go
jeschkies: The current code on `main` overrides he timestamps in each vector.

pkg/logql/first_over_time.go
jeschkies: Done

pkg/logql/engine.go
kavirajk: isn't the timestamp from the actual vector samples more accurate than the one got from `stepEvaluator.Next()`?

Can you help me understand why it won't make sense?

pkg/logql/syntax/ast.go
```diff
@@ -1170,7 +1170,9 @@ const (
 	// internal expressions not represented in LogQL. These are used to
 	// evaluate expressions differently resulting in intermediate formats
 	// that are not consumable by LogQL clients but are used for sharding.
-	OpRangeTypeQuantileSketch = "__quantile_sketch_over_time__"
+	OpRangeTypeQuantileSketch     = "__quantile_sketch_over_time__"
+	OpRangeTypeFirstWithTimestamp = "__first_over_time_ts__"
```
kavirajk, line 1174: AFAIU, this is used to represent that both `first_over_time` and `last_over_time` is shardable with downstream engine. why not call it `__first_over_time_sharded__` instead?. Failing to understand the purpose of `_ts__` suffix here.

pkg/logql/first_last_over_time.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	return ts, SampleVector(r.at)
+}
+
+// The inner batchRangeVectorIterator has preloaded valid samples into each
+// series (see the window filed and load). They're in increasing timestamp
+// order (see next, moves the current timestamp forward by step). So this
+// means that for each downstreamed shard of a first_over_time, selecting the
+// first sample here in each iterator is getting us the earliest timestamped
+// value. Later on when we merge we select the earliest from all the shards.
```
kavirajk, line 62: I believe, this comment should be broken down into pieces and moved to three different iterators.
`firstWithTimestampBatch`, `lastWithTimestampBatch` and `mergeOverTimeEvaluator`?

kavirajk (commented):
Overall looks great. Nice work folks! learned few things about sharding in downstream engine.

Added few comments. One question for me is, how do we make sure the correctness of `first_over_time` and `last_over_time` running with old engine vs downstream engine.?

I see you added tests in `TestMappingEquivalence`. But do you think should be also added on other tests like `TestRangeMappingEquivalence` on downstream engine tests?

pkg/logql/syntax/ast.go
jeschkies: It's for the queriers to tell them to include the original timestamp in the result and not override it with the timestamp of the step.

pkg/logql/engine.go
jeschkies: It does make sense for us. I'm just not sure some other component is relying on this overridden timestamp.

pkg/logql/first_last_over_time.go
jeschkies: Done.

kavirajk (approved):
LGTM. thanks for addressing the feedbacks.

Still this one applies I think.

I see you added tests in `TestMappingEquivalence`. But do you think should be also added on other tests like `TestRangeMappingEquivalence` on downstream engine tests? Rationale being the later tests for more cases of different range aggregation.

approving to unblock.

cstyan:
hey @kavirajk, actually something fun I learned this past week when finally finding the cause of the issues with our `quantile_over_time`, `TestMappingEquivalence` is actually testing the results equivalency for each query as a **range** query. What we don't have (AFAICT) is tests for the results equivalency for **instant** queries. The test names here are actually a bit misleading; `TestMappingEquivalence` is testing for equivalency with sharded queries while `TestRangeMappingEquivalence` is testing for equivalency with time range split queries.

cstyan:
@jeschkies any opinions on merging this as is vs as behind a feature flag? given that it's not being executed via a probabilistic data structure I don't think we need a feature flag

PR #6110, opened 2022-05-06
Improve labels computation during LogQL pipeline

cyriltovena:
This does two things:

- Introduce a buffer for computing labels and the buffer is only copied if we never seen that label set before.
- Add a Map function to the label builders used for `label_format` and `line_format` that avoids slice allocations and sorting.

```
❯ benchstat before.txt after.txt
name                                 old time/op    new time/op    delta
_Pipeline/pipeline_bytes-16            10.6µs ± 0%     6.9µs ± 1%  -34.85%  (p=0.008 n=5+5)
_Pipeline/pipeline_string-16           10.6µs ± 0%     6.9µs ± 0%  -35.16%  (p=0.008 n=5+5)
_Pipeline/line_extractor_bytes-16      10.6µs ± 0%     6.9µs ± 1%  -34.98%  (p=0.008 n=5+5)
_Pipeline/line_extractor_string-16     10.7µs ± 1%     6.9µs ± 0%  -35.27%  (p=0.016 n=5+4)
_Pipeline/label_extractor_bytes-16     10.7µs ± 1%     7.1µs ± 4%  -33.97%  (p=0.008 n=5+5)
_Pipeline/label_extractor_string-16    10.8µs ± 1%     7.0µs ± 0%  -34.90%  (p=0.016 n=5+4)

name                                 old alloc/op   new alloc/op   delta
_Pipeline/pipeline_bytes-16            9.46kB ± 0%    3.26kB ± 0%  -65.58%  (p=0.008 n=5+5)
_Pipeline/pipeline_string-16           9.46kB ± 0%    3.26kB ± 0%  -65.57%  (p=0.008 n=5+5)
_Pipeline/line_extractor_bytes-16      9.46kB ± 0%    3.26kB ± 0%  -65.58%  (p=0.008 n=5+5)
_Pipeline/line_extractor_string-16     9.46kB ± 0%    3.26kB ± 0%  -65.58%  (p=0.008 n=5+5)
_Pipeline/label_extractor_bytes-16     9.46kB ± 0%    3.26kB ± 0%  -65.58%  (p=0.008 n=5+5)
_Pipeline/label_extractor_string-16    9.46kB ± 0%    3.26kB ± 0%  -65.58%  (p=0.008 n=5+5)

name                                 old allocs/op  new allocs/op  delta
_Pipeline/pipeline_bytes-16              44.0 ± 0%      33.0 ± 0%  -25.00%  (p=0.008 n=5+5)
[...]

PR #11607, opened 2024-01-08
labels/line formatter pooling panic fix

cstyan:
Big thanks to @slim-bean, even after ~1 day of trying to find a cause here I didn't track down this issue, that I was returning something to the pool that hadn't actually been retrieved from the pool.

The test here in the 2nd commit will fail if you pass the `-race` flag without the fix from the 3rd commit.

### DIFF
--- a/pkg/logql/log/fmt_test.go
+++ b/pkg/logql/log/fmt_test.go
@@ -925,7 +925,7 @@ func TestInvalidUnixTimes(t *testing.T) {
 	require.Error(t, err)
 }
 
-func TestMapPool(t *testing.T) {
+func TestMapPoolPanic(_ *testing.T) {
 	wg := sync.WaitGroup{}
 	wg.Add(1)
 	wgFinished := sync.WaitGroup{}
--- a/pkg/logql/log/fmt_test.go
+++ b/pkg/logql/log/fmt_test.go
@@ -932,7 +932,9 @@ func TestMapPoolPanic(_ *testing.T) {
 
 	ls := labels.FromStrings("cluster", "us-central-0")
 	builder := NewBaseLabelsBuilder().ForLabels(ls, ls.Hash())
-	a := newMustLineFormatter(`[1m{{if .level }}{{alignRight 5 .level}}{{else if .severity}}{{alignRight 5 .severity}}{{end}}[0m [90m[{{alignRight 10 .resources_service_instance_id}}{{if .attributes_thread_name}}/{{alignRight 20 .attributes_thread_name}}{{else if eq "java" .resources_telemetry_sdk_language }}                    {{end}}][0m [36m{{if .instrumentation_scope_name }}{{alignRight 40 .instrumentation_scope_name}}{{end}}[0m {{.body}} {{if .traceid}} [37m[3m[traceid={{.traceid}}]{{end}}`)
+	// this specific line format was part of the query that first alerted us to the panic caused by map pooling in the label/line formatter Process functions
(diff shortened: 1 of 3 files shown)


pkg/logql/log/labels.go
```diff
@@ -495,22 +495,21 @@ func (b *LabelsBuilder) IntoMap(m map[string]string) {
...
 			b.baseMap = b.base.Map()
 		}
-		return b.baseMap
+		return b.baseMap, false
 	}
 	b.buf = b.UnsortedLabels(b.buf)
 	// todo should we also cache maps since limited by the result ?
 	// Maps also don't create a copy of the labels.
 	res := smp.Get()
-	clear(res)
```
dannykopping, line 509: Why are we losing this?

pkg/logql/log/labels.go
```diff
@@ -495,22 +495,21 @@ func (b *LabelsBuilder) IntoMap(m map[string]string) {
...
 	if !b.hasDel() && !b.hasAdd() && !b.HasErr() {
 		if b.baseMap == nil {
 			b.baseMap = b.base.Map()
 		}
-		return b.baseMap
+		return b.baseMap, false
 	}
 	b.buf = b.UnsortedLabels(b.buf)
 	// todo should we also cache maps since limited by the result ?
 	// Maps also don't create a copy of the labels.
```
dannykopping, line 507: Is this still relevant?

pkg/logql/log/fmt_test.go
```diff
@@ -923,3 +924,46 @@ func TestInvalidUnixTimes(t *testing.T) {
...
 }
+
+func TestMapPoolPanic(_ *testing.T) {
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	wgFinished := sync.WaitGroup{}
+
+	ls := labels.FromStrings("cluster", "us-central-0")
+	builder := NewBaseLabelsBuilder().ForLabels(ls, ls.Hash())
+	a := newMustLineFormatter(`[1m{{if .level }}{{alignRight 5 .level}}{{else if .severity}}{{alignRight 5 .severity}}{{end}}[0m [90m[{{alignRight 10 .resources_service_instance_id}}{{if .attributes_thread_name}}/{{alignRight 20 .attributes_thread_name}}{{else if eq "java" .resources_telemetry_sdk_language }}                    {{end}}][0m [36m{{if .instrumentation_scope_name }}{{alignRight 40 .instrumentation_scope_name}}{{end}}[0m {{.body}} {{if .traceid}} [37m[3m[traceid={{.traceid}}]{{end}}`)
```
dannykopping, line 935: Might be helpful to explain here why such an exotic test case is required.

dannykopping (commented):
Left a couple minor questions

Please give the PR a more meaningful title

pkg/logql/log/labels.go
cstyan: this wasn't needed in the first place, the pool wrappers `Get` call clears the map before returning it to us, this was leftover from before I had the wrapper

pkg/logql/log/fmt_test.go
cstyan: This was the only query that I found caused the panic in the first place, though not reliably within smaller environments/smaller time ranges. I tracked it down to the actual panics occurring when we first noticed this, and then reran every query I could find in that time range. This large line format was the only one that reliably caused the panic to occur in larger environments, and in dev/testing with a siginficantly large time range (7-30d) it would happen reliably as well.

pkg/logql/log/labels.go
cstyan: I think so, yes. It was added in Cyrils PR that added the `baseMap` in the first place, and I haven't changed that behaviour at all: https://github.com/grafana/loki/pull/6110/commits/e6002637698e89b1fe888005dd0f46fcd00ba676

pkg/logql/log/fmt_test.go
dannykopping: I meant as a comment, but at least we have it here in the PR for archaeology's sake 👍 thanks

dannykopping (approved):
Please add a changelog entry and backport these
Thanks for the fix!

cstyan:
supersedes https://github.com/grafana/loki/pull/11588 and https://github.com/grafana/loki/pull/11583

PR #11608, opened 2024-01-09
Bump build image

trevorwhitney:

Bumps alpine in the loki build image and ass the build image to the `scan-vulnerabilities` target to make it easier to scan in the future

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -6,7 +6,7 @@ jobs:
     env:
       BUILD_IN_CONTAINER: false 
     container:
-      image: grafana/loki-build-image:0.32.0
+      image: grafana/loki-build-image:0.31.3
     steps:
       - uses: actions/checkout@v4
       - run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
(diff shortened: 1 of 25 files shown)


.github/workflows/checks.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     env:
       BUILD_IN_CONTAINER: false 
     container:
-      image: grafana/loki-build-image:0.32.0
```
dannykopping, line 9: This is weird; howcome we're still publishing 0.31 after 0.32 was published?

dannykopping (approved):
LGTM, thanks!

.github/workflows/checks.yml
trevorwhitney: good catch, I don't know where `0.32.0` was published from as it's not mentioned in the `Makefile`, but I'm going to jump up to `0.33.0` to avoid confusion.

loki-build-image/README.md
```diff
@@ -2,7 +2,11 @@
 
 ## Versions
 
-### 0.30.18
+### 0.31.3
```
dannykopping, line 5: Should we update this too?

MichelHollands (approved):

PR #11622, opened 2024-01-09
Update admonitions to add shortcode

JStickler:
Completing the work that was started in #9409 but not completed because there were multiple variations on how to format notes in the docs source files.  We rolled out different colors for admonitions recently, which highlighted that there were still a few notes not using the admonition shortcode.  I searched the docs for the following words:
* tip
* caution
* important
* warning
* note

So hopefully I've caught all of the admonitions that weren't using the shortcode.


MichelHollands (approved):
LGTM

PR #11645, opened 2024-01-10
Add e2e tests for bloom filtering

chaudum:
**What?**

This PR adds an end-to-end integration test for creating bloom filters and using them in the gateway to filter chunks.

**Why?**

This test helped to identify bugs in the bloom shipper code that would have taken a very long time to discover in other ways, such as deploying to a dev environment or running a long-running docker compose setup.

**Notes**

The following commits of this PR are actual changes to the compactor/gateway code to make e2e test work:

* https://github.com/grafana/loki/pull/11645/commits/2dcc761ba069a377ece1a3a48fa6c6c59039043c
* https://github.com/grafana/loki/pull/11645/commits/84052ddb34c012e12d3e98b86317a6b18bc3d76e
* https://github.com/grafana/loki/pull/11645/commits/3ba44f83c81bb305d3fb08d0141b6f080bde03ea

The bloom gateway code path for processing blocks was cleaned up, because it still contained the unused "sequention processing" path of blocks, which was initially kept to verify the callback based processing works alike:

* https://github.com/grafana/loki/pull/11645/commits/f55d79b84fa6d6118fd66720aa107ba15ef0b862

### DIFF
--- a/integration/loki_micro_services_test.go
+++ b/integration/loki_micro_services_test.go
@@ -3,11 +3,14 @@ package integration
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"math/rand"
 	"strings"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/go-kit/log/level"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
 	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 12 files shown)


pkg/storage/stores/shipper/bloomshipper/shipper.go
```diff
@@ -59,14 +58,8 @@ func (s *Shipper) Fetch(ctx context.Context, tenantID string, blocks []BlockRef,
...
-		case result, ok := <-blocksChannel:
-			if !ok {
-				return nil
-			}
-			err := runCallback(callback, result)
-			if err != nil {
-				return err
-			}
+		case result := <-blocksChannel:
+			return runCallback(callback, result)
```
chaudum, line 62: This will only work for downloading a single block. I want to fix this in a follow up PR.

vlad-diachenko (approved):

integration/loki_micro_services_test.go
```diff
@@ -1056,6 +1059,174 @@ func TestCategorizedLabels(t *testing.T) {
...
+		err := cliDistributor.PushLogLine(line, now.Add(-1*time.Hour).Add(time.Duration(i-len(uniqueStrings))*time.Second), nil, map[string]string{"pod": pod})
+		require.NoError(t, err)
+	}
+
+	// restart ingester to flush chunks and that there are zero chunks in memory
+	require.NoError(t, cliIngester.Flush())
+	require.NoError(t, tIngester.Restart())
+
+	// wait for compactor to compact index and for bloom compactor to build bloom filters
+	time.Sleep(10 * time.Second)
```
vlad-diachenko, line 1209: it would make the test more reliable and faster if we continued it once the compaction was done. maybe use some metrics or something like that.

chaudum: Agree. As I want to do the same for verifying the actual use of bloom filters for querying, I would do that in the separate follow up PR.

pkg/storage/stores/shipper/bloomshipper/shipper.go
chaudum: Good point. I think the actuall problem was only
```
case result, ok := <-blocksChannel:
```
which should be

```
case result := <-blocksChannel:
```

chaudum: > what was the problem with the previous implementation?

The previous implementation never exited from the `for { select {} }` loop, because `runCallback()` should not return an error.

pkg/storage/stores/shipper/bloomshipper/shipper.go
```diff
@@ -164,39 +147,30 @@ func (s *Shipper) findBlocks(metas []Meta, startTimestamp, endTimestamp model.Ti
...
-		return i
-	}
-	return len(s)
-}
-
-func isOutsideRange(b *BlockRef, startTimestamp, endTimestamp model.Time, fingerprints []uint64) bool {
+// isOutsideRange tests if a given BlockRef b is outside of search boundaries
+// defined by min/max timestamp and min/max fingerprint.
+// Fingerprint ranges must be sorted in ascending order.
+func isOutsideRange(b *BlockRef, startTimestamp, endTimestamp model.Time, fingerprints [][2]uint64) bool {
```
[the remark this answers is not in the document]
chaudum: See commit https://github.com/grafana/loki/pull/11645/commits/db61f73b0647bcb4d5cf916dd0e764a031571504

pkg/storage/stores/shipper/bloomshipper/shipper.go
```diff
@@ -164,39 +169,29 @@ func (s *Shipper) findBlocks(metas []Meta, startTimestamp, endTimestamp model.Ti
...
-	// Check if the block range is inside a "gap" in the fingerprint slice
-	// e.g.
-	// fingerprints = [1, 2,          6, 7, 8]
-	// block =              [3, 4, 5]
-	idx := getPosition[[]uint64](fingerprints, b.MinFingerprint)
-	// in case b.MinFingerprint is outside of the fingerprints range, return true
-	// this is already covered in the range check above, but I keep it as a second gate
-	if idx > len(fingerprints)-1 {
-		return true
+	prev := fpRange{0, 0}
```
chaudum, line 187: Could be written as

```suggestion
	var prev fpRange
```

PR #11654, opened 2024-01-11
Cache: correctly check background cache size

dannykopping:
The size check was not being performed atomically, which led to flakiness in the `TestBackgroundSizeLimit` test.

### DIFF
--- a/pkg/storage/chunk/cache/background.go
+++ b/pkg/storage/chunk/cache/background.go
@@ -148,8 +148,13 @@ func (c *backgroundCache) Store(ctx context.Context, keys []string, bufs [][]byt
 		}
 
 		size := bgWrite.size()
-		newSize := c.size.Load() + int64(size)
+		// prospectively add new size
+		c.size.Add(int64(size))
+
+		newSize := c.size.Load()
 		if newSize > int64(c.sizeLimit) {
+			// subtract it since we've exceeded the limit
+			c.size.Add(-int64(size))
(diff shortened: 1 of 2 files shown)


pkg/storage/chunk/cache/background.go
```diff
@@ -148,8 +148,13 @@ func (c *backgroundCache) Store(ctx context.Context, keys []string, bufs [][]byt
 		}
 
 		size := bgWrite.size()
-		newSize := c.size.Load() + int64(size)
+		// prospectively add new size
+		c.size.Add(int64(size))
+
+		newSize := c.size.Load()
```
chaudum, line 154: `Add` atomically adds to the wrapped int64 and returns the new value.                  

```suggestion
		newSize := c.size.Add(int64(size))
```

pkg/storage/chunk/cache/background.go
```diff
@@ -148,8 +148,13 @@ func (c *backgroundCache) Store(ctx context.Context, keys []string, bufs [][]byt
...
 		if newSize > int64(c.sizeLimit) {
+			// subtract it since we've exceeded the limit
+			c.size.Add(-int64(size))
```
chaudum, line 157: `Sub` atomically subtracts from the wrapped int64 and returns the new value.           

```suggestion
			c.size.Sub(int64(size))
```

chaudum (approved):

PR #11657, opened 2024-01-11
fix(log results cache): compose empty response based on the request

ashwanthgoli:
Log results cache when handling a hit composes an empty response based on the cached request.
But the limit or direction fields in the cached request need not match with the current request being served.

This causes the log results cache to return a response with incorrect limit. This incorrect limit could then get applied when merging responses upstream (split by interval mw for ex.)

This pr fixes this by composing the response based on the request being served.

I also thought about updating the cache key to include both limit and direction to have a clear separation, but I left it as is for the following reason: if a time range contains no log lines, that result would not change irrespective of a different limit or direction

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


MichelHollands (approved):
LGTM

chaudum (approved):
LGMT

Don't forget to add the relevant backport labels

PR #11665, opened 2024-01-11
Deprecate and flip -legacy-read-mode to false by default

salvacorts:

This PR addresses a pending TODO to flip the value of the `-legacy-read-mode` to false by default. We also deprecate it as we plan to remove it after Loki 3.0.

### DIFF
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -101,7 +101,7 @@ type Config struct {
 	Tracing       tracing.Config       `yaml:"tracing"`
 	Analytics     analytics.Config     `yaml:"analytics"`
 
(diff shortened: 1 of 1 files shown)


MichelHollands:
@salvacorts Should we have a changelog entry for this?

trevorwhitney (approved):
Looks great, thanks!

trevorwhitney:
oh, probably a good call on the changelog (and the upgrading guide)

salvacorts:
@trevorwhitney @MichelHollands, thanks for the heads up! I updated the changelog and the deprecated config checker.

MichelHollands (approved):
LGTM

PR #11701, opened 2024-01-17
Loki: increase the limit for query length from 5k to 128k

slim-bean:

We originally added this limit from fuzz testing and realizing there should be some maximum limit to an allowed query size. The original limit was 5120 based on some internet searching and a best estimate of what a reasonable limit would be. We have seen use cases with queries containing a lot of filter expressions or long expanded variable names where this limit was too small. Apparently the spec does not specify a limit, and more internet searching suggests almost all browsers will handle 100k+ length urls without issue
Some limit here still seems prudent however, so the new limit is now 128k.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/syntax/parser.go
+++ b/pkg/logql/syntax/parser.go
@@ -31,7 +31,12 @@ var parserPool = sync.Pool{
 	},
 }
 
-const maxInputSize = 5120
+// (E.Welch) We originally added this limit from fuzz testing and realizing there should be some maximum limit to an allowed query size.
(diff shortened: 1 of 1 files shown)


MasslessParticle (approved):

dannykopping (approved):
Seems reasonable, LGTM
Although why do we have it as a constant and not a limit?

owen-d (approved):

PR #11703, opened 2024-01-17
[TSDB][bug] retain previously rotated tsdb heads for one cycle to serve queries

owen-d:
This PR ensures we retain previous TSDB heads on ingesters for one cycle (15m) in order to serve index queries while waiting for index-gws or queriers to sync it from storage. This bug hadn't surfaced yet (for us) because we've historically used ~6m `chunk-retain` periods which masked this problem (chunks were kept around so we didn't see read gaps). Fixing this bug should allow us to avoid retaining chunks post-flush in ingester memory in TSDB since it doesn't utilize the index-cache (which was the other reason for retaining chunks in memory).

This is very attractive because it allows us to reduce memory pressure in ingesters as well as reduces the data they query since now ingesters will only return the chunk references. This allows queriers to download chunks directly rather than ingesters iterating the chunk data for flushed chunks themselves.

Also does some refactoring to make testing validation easier.

### DIFF
--- a/pkg/storage/stores/shipper/indexshipper/tsdb/head_manager.go
+++ b/pkg/storage/stores/shipper/indexshipper/tsdb/head_manager.go
@@ -144,26 +144,66 @@ func NewHeadManager(name string, logger log.Logger, dir string, metrics *Metrics
 	return m
 }
 
-func (m *HeadManager) loop() {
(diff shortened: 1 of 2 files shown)


slim-bean (approved):
LGTM

PR #11716, opened 2024-01-19
jsonnet: fix unicode quotes in jsonnet to avoid tools failing to parse it properly

sandeepsukhani:
We have a linter in place in our internal deployment tooling to catch the usage of Unicode quotes because it has caused some issues due to shells/tooling not recognizing them. The linter has caught one such instance in the jsonnet in the Loki repo, which is fixed in this PR.


MichelHollands (approved):
LGTM

PR #11720, opened 2024-01-19
[bloom-compactor] Fix: blooms returning same checksum with different fp/ts params. 

poyzannur:
We noticed that blooms created for different fingerprint(FPs) ranges and timestamps(Ts) were returning same checksums. 
Block building essentially creates [blooms](https://github.com/grafana/loki/blob/main/pkg/storage/bloom/v1/builder.go#L105C1-L114C2) and [indexes](https://github.com/grafana/loki/blob/main/pkg/storage/bloom/v1/builder.go#L276C1-L290C1). Previously we were calculating checksums based on blooms. However they do not contain information about FP and Ts. Moving checksum calculation to indexes includes that information, and results in different checksums as expected. 

Fixes https://github.com/grafana/loki-private/issues/899

The tests assume the chunkRefs in series are the same, for simplicity I kept this as is. But tested locally with different chunks refs too.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


poyzannur:
> I feel like this misses the mark a bit. Seeing the same checksums after summing the _content_ (blooms) and migrating to checksumming the _metadata_ (index) just seems like we had a bug in the previous checksum code, not that this approach is more attractive. I think checksumming on content|index|both would reasonably work (both is _probably_ better b/c it's all technically "content"), but logically we shouldn't be seeing the same checksums when summing the blooms if their metadata is different (outside of statistically unlikely collisions).
> 
> Does that make sense?

It sure does make sense. I tried to communicate in the last meeting that this was a cheap work around to unblock the team. Bc I could not find a bug in the checksum code, hence i offered this. I'll spend more time looking at the checksum code. But I don't think it harms to merge this while keeping the bug issue open. 
how do you envision implementing checksum over both without adding overhead?

owen-d:
Thanks for addt'l context that this was just unblocking our current approach. A few bits in no particular order:
* Checksumming based on the index _should_ be fine, as long as we also include the block `schema` as part of the input to checksumming code (so different algorithm parameters built against the same inputs produce different checksummed blocks). One benefit of checksumming everything is that we remove the need to think about minutiae like this. Instead, if the block bytes change, so will the checksum. Small benefit now, but can help catch some nasty bugs down the line.
* I'm worried that changing this now, while helpful & expedient in the short term, is covering up issues we don't yet understand.
* Performance wise, this all has a cost, but I don't expect it to be an issue (this is something we can measure/re-evaluate later).

Let's spend some more time uncovering why we're seeing identical checksums before moving to this approach.

poyzannur:
@paul1r and I were debugging this issue and got to the bottom of the why. We calculate the checksum with [BloomPageHeader](https://github.com/grafana/loki/blob/main/pkg/storage/bloom/v1/bloom.go#L144), which does not contain any information about actual bloom data. All fields can turn out to be the same for same number of pages for the same number of series in a bloom.

We calculate checksum on the page data at [flushPage()] (https://github.com/grafana/loki/blob/main/pkg/storage/bloom/v1/builder.go#L192-L198), but we just append it to that page and leave it. We now consider adding that hash to the `BloomPageHeader` which will give some variance on the bloom data per page. I'm closing this PR in favour of this solution.

PR #11736, opened 2024-01-22
De-duplicate overlapping chunks for log volume.

### LINKED ISSUE #11709: Investigate LVE Improvement
Currently the Log Volume Exporter, LVE, has roughly a relative error of 20% for queries more than two hours in the past. Queries for the last two hours have a lot of duplication because the ingesters didn't de-dupe the chunks in memory.

One idea to improve the LVE accuracy is to de-duplicate overlapping chunks by the percentage of the overlap. This is already done for determining how much a chunk is contributing to a given time range: https://github.com/grafana/loki/blob/28a413539e4d14311a2d5d58e7e145e0e1e56ac6/pkg/storage/stores/shipper/indexshipper/tsdb/index/chunk.go#L247-L252

The task is to evaluate the method.

jeschkies:
De-duplicate overlapping chunks for volume endpoint.

Fixes #11709

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/storage/stores/shipper/indexshipper/tsdb/index/chunk.go
```diff
@@ -25,6 +26,39 @@ func (c ChunkMeta) From() model.Time                 { return model.Time(c.MinTi
 func (c ChunkMeta) Through() model.Time              { return model.Time(c.MaxTime) }
 func (c ChunkMeta) Bounds() (model.Time, model.Time) { return c.From(), c.Through() }
 
+// Sub subtracts the other chunk from c if the overlap. If the other is completely
+func Sub(l, r *ChunkMeta) []*ChunkMeta {
```
gubjanos, line 30: note, this is unused and even probably incorrect.

I think the assumption is l.Min > r.Min and l.Max < r.Max, but then on line 37 we should do r.Max-l.Max right?

pkg/storage/stores/shipper/indexshipper/tsdb/index/chunk_test.go
```diff
@@ -145,6 +145,104 @@ func TestChunkMetasFinalize(t *testing.T) {
...
+			chks: []ChunkMeta{
+				{MinTime: 0, MaxTime: 1, KB: 1, Entries: 1},
+				{MinTime: 1, MaxTime: 4, KB: 1, Entries: 1},
+				{MinTime: 2, MaxTime: 3, KB: 1, Entries: 1},
+				{MinTime: 3, MaxTime: 5, KB: 2, Entries: 2},
+				{MinTime: 4, MaxTime: 5, KB: 1, Entries: 1},
+			},
+			from:    0,
+			through: 5,
+			exp: ChunkStats{
```
gubjanos, line 183: I think this testcase is an amazing representation of our current problem.
Let me try to do a different math:
I only take the following chunks (subset of the test's data):
{MinTime: 0, MaxTime: 1, KB: 1, Entries: 1},
{MinTime: 2, MaxTime: 3, KB: 1, Entries: 1}
{MinTime: 3, MaxTime: 5, KB: 2, Entries: 2}
result would be: {Chunks:3, KB: 4, Entries: 4}

If I take a closer look, all I can tell is we have 2 possible duplicates, so 6-2 would end up with 4. What is happening here?

pkg/storage/stores/shipper/indexshipper/tsdb/index/chunk.go
```diff
@@ -105,6 +139,51 @@ func (c ChunkMetas) Finalize() ChunkMetas {
...
+	res := ChunkStats{}
+
+	if len(c) > 1 && deduplicate {
+		sort.Sort(c)
+		last := c[0]
+		n := 1
+		for _, cur := range c[1:] {
+			// Skip chunk if it's a subset of the last one
+			if cur.MinTime < last.MaxTime && cur.MaxTime <= last.MaxTime {
+				continue
```
gubjanos, line 153: Would it make sense to take the max(cur, last) instead of skipping? I am a newbie of how chunks are formed, but if one of them could be a subset of the other, nothing guarantees that the one coming first is the "bigger".

jeschkies: That's true. We should average the size probably.

pkg/storage/stores/shipper/indexshipper/tsdb/index/chunk_test.go
jeschkies: What's with `{MinTime: 1, MaxTime: 4, KB: 1, Entries: 1}` how would we get that out?

jeschkies:
Won't do for now since the use case is limited.

PR #11737, opened 2024-01-23
[logQL] Allow sharding of first and last over time aggregations

cstyan:
Supersedes https://github.com/grafana/loki/pull/11605, implementing both first and last over time sharding.


cstyan:
closing, we've updated #11605

PR #11750, opened 2024-01-23
Extend `TestPartitionFingerprintRange` test case

chaudum:

The test cases now also cover partially out of bounds blocks as well as inconsecutive block ranges.


salvacorts (approved):

vlad-diachenko (commented):
looks great

PR #11755, opened 2024-01-24
[Blooms] Specs out BloomGenerator

owen-d:
This is my attempt at making more testable implementations of the work done by the `bloomcompactor` pkg. It's largely inspired by what already exists there, but without depending on things like object storage, chunk clients, etc, it's much easier to locally prove behavior. I'm planning to build on top of this, continuing to integrate work already done in this pkg.

### DIFF
--- a/pkg/bloomcompactor/v2spec.go
+++ b/pkg/bloomcompactor/v2spec.go
@@ -7,6 +7,7 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
+	"github.com/grafana/loki/pkg/storage/chunk"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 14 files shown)


pkg/bloomcompactor/v2spec.go
```diff
@@ -0,0 +1,142 @@
...
+
+	level.Debug(s.logger).Log("msg", "generating bloom filters for blocks", "num_blocks", len(blocksMatchingSchema), "skipped_blocks", len(skippedBlocks), "schema", fmt.Sprintf("%#v", s.opts.Schema))
+
+	// TODO(owen-d): implement bounded block sizes
+
+	mergeBuilder := v1.NewMergeBuilder(blocksMatchingSchema, s.store, s.populate)
+	writer, reader := s.readWriterFn()
+	blockBuilder, err := v1.NewBlockBuilder(v1.NewBlockOptionsFromSchema(s.opts.Schema), writer)
+	if err != nil {
+		return skippedBlocks, nil, errors.Wrap(err, "failed to create bloom block builder")
```
chaudum, line 133: Do we want to return `skippedBlocks` in case of an error? Does the consumer rely on them?

pkg/bloomcompactor/v2spec.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	_, err = mergeBuilder.Build(blockBuilder)
+	if err != nil {
+		return skippedBlocks, nil, errors.Wrap(err, "failed to build bloom block")
```
chaudum, line 137: Same here

pkg/bloomcompactor/v2spec_test.go
```diff
@@ -0,0 +1,123 @@
...
+			storeItr := v1.NewMapIter[v1.SeriesWithBloom, *v1.Series](
+				v1.NewSliceIter[v1.SeriesWithBloom](data),
+				func(swb v1.SeriesWithBloom) *v1.Series {
+					return swb.Series
+				},
+			)
+
+			gen := dummyBloomGen(tc.toSchema, storeItr, sourceBlocks)
+			skipped, results, err := gen.Generate(context.Background())
+			require.Nil(t, err)
```
chaudum, line 102: `require.NoError()` is a bit more verbose if case the assertion fails.

chaudum (commented):
LGTM

pkg/bloomcompactor/v2spec.go
owen-d: I'm mostly returning them for informational purposes at the moment. It's possible that we'd still want to know incompatible blocks when there's an error, but this is all too early to tell. I expect it to change a bit over time.

chaudum (approved):

poyzannur (approved):

PR #11758, opened 2024-01-24
Release 2.9.4

poyzannur:
Updates changelog and release notes for 2.9.4 patch release

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Unreleased
+## 2.9.4 (2024-01-24)
 
 ### All Changes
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,21 +1,21 @@
-## Unreleased
+## 2.9.4 (2024-01-24)
 
 ### All Changes
 
 #### Loki
 
 ##### Fixes
 
-* [11606](https://github.com/grafana/loki/pull/11606) **dannykopping** Fixed regression adding newlines to HTTP error response bodies which may break client integrations.
```
dannykopping, line 9: Curious why this was removed?

poyzannur: it didn't have the backport to 2.9.x tag, i deduced it wasn't backported? 
Let me double check if it's on the branch.

poyzannur: yep, it's not backported. would you like to include it in the patch release?

dannykopping: My bad! I _think_ this bug hadn't made it into 2.9.x which is why I didn't backport it
Lemme double check (and 2.8.x as well)

dannykopping: OK nvm, I was right, this bug never made it into the other releases 👍 please ignore

na-- (approved):

docs/sources/release-notes/v2-9.md
```diff
@@ -49,3 +49,11 @@ For a full list of all changes and fixes, look at the [CHANGELOG](https://github
 * Fix querying ingester for label values with a matcher (previously didn't respect the matcher).
 * Ensure all lifecycler cfgs ref a valid IPv6 addr and port combination
 
+### 2.9.4 (2024-01-24)
```
DylanGuedes, line 52: nit: I know that we were already listing the different versions in increasing order (ex: first 2.9.1, then 2.9.2, etc), but I think the user experience is better if more recent releases are listed first. So 2.9.4, then 2.9.3, etc

na--: While I tend to agree, especially given the fact that in the 2.8.x branch the newest patch versions were first, see https://github.com/grafana/loki/blob/ned/release-v2.8.8/docs/sources/release-notes/v2-8.md :confused: 

I thought that someone had intentionally flipped the order for 2.9.x and so I didn't comment :sweat_smile:

poyzannur: thanks for catching, i actually didn't realised. 
i changed the order to be reverse.

PR #11768, opened 2024-01-24
Bloom Gateway: Partition requests into multiple tasks

chaudum:

This PR changes at which point a request is split into multiple tasks in case the chunks boundaries of the request overlap multiple days.

This solves the problem that the request handler needs to know for how many responses from the block querier needs to wait.

Previously, a single task that spanned across multiple days was executed multiple times (once for each matching day).


pkg/bloomgateway/bloomgateway.go
```diff
@@ -299,6 +299,18 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
 		return nil, err
 	}
 
+	// start time == end time => empty response
```
salvacorts, line 302: nit: The `=>` here reads like greater-equal rather than an arrow. Can you use `-->` instead?

pkg/bloomgateway/bloomgateway.go
```diff
@@ -315,21 +327,63 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
 		return req.Refs[i].Fingerprint < req.Refs[j].Fingerprint
 	})
 
-	task, resCh, errCh, err := NewTask(tenantID, req)
+	var expectedResponses int
+	swbPerDay := partitionRequest(req)
```
salvacorts, line 331: what does `swb` stand for here?

pkg/bloomgateway/bloomgateway.go
```diff
@@ -315,21 +327,63 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
 		return req.Refs[i].Fingerprint < req.Refs[j].Fingerprint
```
salvacorts, line 327: Is this sort needed other than for the bin-search inside `partitionRequest`?
- If not: I think `partitionRequest` should make the sorting.
- If yes: I think `partitionRequest` should check that the chunks are sorted. And sort them if not.

pkg/bloomgateway/bloomgateway.go
```diff
@@ -315,21 +327,63 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
+	for _, seriesWithBounds := range swbPerDay {
+		task, err := NewTask(tenantID, seriesWithBounds, req.Filters)
+		if err != nil {
+			return nil, err
+		}
+		tasks = append(tasks, task)
+		expectedResponses += len(seriesWithBounds.series)
+	}
+
+	if len(tasks) == 0 {
```
salvacorts, line 343: I think we can have this check with `swbPerDay` right after `partitionRequest` (not a big deal tho)

pkg/bloomgateway/bloomgateway.go
```diff
@@ -342,9 +396,9 @@ outer:
 		case res := <-resCh:
```
salvacorts, line 396: In case of an error executing any of the subtasks, shouldn't we cancel the context so other subtasks stop as well?

pkg/bloomgateway/bloomgateway.go
chaudum: `seriesWithBounds`. it's the return type's name

pkg/bloomgateway/bloomgateway.go
chaudum: We rely on sorted fingerprints throughout the read path. So we sort once in the beginning.

pkg/bloomgateway/bloomgateway.go
chaudum: Cancellation of tasks does not really work yet, that's due to how we rely on getting results from the channel and not passing the context to the task.
Yes, it's an issue indeed, but I would make a separate PR to fix that properly. I thought it would exceed the scope of this PR.

salvacorts (approved):
LGTM

pkg/bloomgateway/bloomgateway.go
```diff
@@ -315,21 +327,64 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
+	}
+
+	tasks := make([]Task, 0, len(seriesWithBloomsPerDay))
+	for _, seriesWithBounds := range seriesWithBloomsPerDay {
+		task, err := NewTask(tenantID, seriesWithBounds, req.Filters)
+		if err != nil {
+			return nil, err
+		}
+		tasks = append(tasks, task)
+		expectedResponses += len(seriesWithBounds.series)
```
vlad-diachenko, line 347: 👍

pkg/bloomgateway/bloomgateway.go
```diff
@@ -315,21 +327,64 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
-	level.Info(g.logger).Log("msg", "enqueue task", "task", task.ID)
-	g.queue.Enqueue(tenantID, []string{}, task, func() {
-		// When enqueuing, we also add the task to the pending tasks
-		g.pendingTasks.Add(task.ID, task)
-	})
 
-	requestCount := len(req.Refs)
-	responses := responsesPool.Get(requestCount)
+	errCh := make(chan error, 1)
+	resCh := make(chan v1.Output, 1)
```
vlad-diachenko, line 353: we know the responses count upfront, so we can create a channel that can keep N responses simultaneously

vlad-diachenko: I believe we are going to do it in the follow up PRs

vlad-diachenko (commented):
looks great!

PR #11717, opened 2024-01-19
Make bloom gateway tasks cancelable (and other fixes)

chaudum:
**Why we need it**:

A bloom gateway request calling `FilterChunkRefs` that timed out caused a deadlock of the workers, because when the request context is cancelled, the request returned, but the task continued to process until the block querier tried to write to the results channel. That writing was blocked, because the request that was supposed to read from the results channel to gather the result was gone.

**How is is solved**:

The PR introduces two changes:

1) The request context is also passed when creating a new `Task`, which allows to check the cancellation of a request also in the task.
2) The channels of the `Task` are replaced by a "pipe channel", which consists of a sender channel and a receiver channel. The pipe reads from the sender channel until it is closed, and forwards the received item to the receiver channel, but only if the internal context of the pipe isn't closed.
3) In the worker, when dequeuing and processing the tests, the context of tasks is checked for error, to shortcut operations and return errors early.

---

Further smaller changes/fixes:

1) Fix a potential `index out of bounds` panic in the code that removes filtered out chunk refs from th original request.
2) Make sure that the `DequeueMany` call on the task queue returns successive tasks based on their enqueue order.
3) Do not remove cancelled task from task list once they have been partitioned by fingerprint ranged.

PR #11770, opened 2024-01-24
Bloom Gateway: Fix panic in chunk removal after receiving query results

chaudum:

While removing not matching chunks from a fingerprint, the for loop over the chunks removes items, and therefore the index of the current iteration needs to be adjusted to accommodate for the removal, otherwise it may panic with `out of bounds`.

**Notes for reviewer**:

Extracted from https://github.com/grafana/loki/pull/11717


paul1r (commented):
lgtm

salvacorts (approved):

PR #11777, opened 2024-01-25
<Promtail>: Fix flaky docker test

wildum:

We ported this fix a few days ago to the Agent via this [PR](https://github.com/grafana/agent/pull/6201) but the added test is often failing in our drone pipeline.

I believe that there are two reasons why this test is flaky:
- the test expects that the 5 last lines of logs are the expected ones but it seems that other lines might be logged as well (https://drone.grafana.net/grafana/agent/16045/4/2)
- we saw that the simulated container did not always have enough time to stop before calling the tgt.startIfNotRunning() to restart it. (https://drone.grafana.net/grafana/agent/16077/4/2)

Fix:
- the test now uses "assert.EventuallyWithT" and checks that within 5 seconds the expected log lines will be amongst the logs without duplicate.
- the test now stops the simulated container before restarting it

**Checklist**

### DIFF
--- a/clients/pkg/promtail/targets/docker/target_test.go
+++ b/clients/pkg/promtail/targets/docker/target_test.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
-	"sort"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


MichelHollands (approved):
LGTM

PR #11784, opened 2024-01-25
Bloom Shipper: Fix bug in block boundary check

chaudum:

The check `isOutsideRange()` returned incorrect results, because the end timestamp of the search interval is not inclusive.

### DIFF
--- a/pkg/storage/stores/shipper/bloomshipper/shipper.go
+++ b/pkg/storage/stores/shipper/bloomshipper/shipper.go
@@ -43,6 +43,15 @@ func (r fpRange) maxFp() uint64 {
 	return r[1]
 }
 
+func (r fpRange) Cmp(other uint64) v1.BoundsCheck {
+	if other < r[0] {
+		return v1.Before
+	} else if other > r[1] {
+		return v1.After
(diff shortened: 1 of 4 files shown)


pkg/storage/stores/shipper/bloomshipper/shipper.go
```diff
@@ -15,6 +15,24 @@ import (
 	"github.com/grafana/loki/pkg/storage/stores/shipper/bloomshipper/config"
 )
 
+type Interval struct {
+	Start model.Time
```
salvacorts, line 19: nit: Maybe we can reuse model.Interval instead?

salvacorts (approved):

pkg/storage/stores/shipper/bloomshipper/shipper.go
chaudum: I started with using `model.Interval`, but I wanted to extend it with `Cmp()` and `String()`.

owen-d (approved):

vlad-diachenko (commented):
looks great

PR #11785, opened 2024-01-25
Add timing metrics for compaction

salvacorts:
This PR modifies the metrics of the bloom compactor to:
- Add new histogram metrics to track the time spent to process compaction cycles, tenants and jobs.
- Replace `_succeeded` and `_failed` metrics with a new `_completed` metric with a status label. 

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/bloomcompactor/bloomcompactor.go
+++ b/pkg/bloomcompactor/bloomcompactor.go
@@ -211,13 +211,16 @@ func (c *Compactor) running(ctx context.Context) error {
 	for {
 		select {
 		case <-ticker.C:
+			start := time.Now()
 			c.metrics.compactionRunsStarted.Inc()
 			if err := c.runCompaction(ctx); err != nil {
-				c.metrics.compactionRunsFailed.Inc()
+				c.metrics.compactionRunsCompleted.WithLabelValues("fail").Inc()
+				c.metrics.compactionRunTime.WithLabelValues("fail").Observe(time.Since(start).Seconds())
 				level.Error(c.logger).Log("msg", "failed to run compaction", "err", err)
 				continue
 			}
-			c.metrics.compactionRunsCompleted.Inc()
+			c.metrics.compactionRunsCompleted.WithLabelValues("success").Inc()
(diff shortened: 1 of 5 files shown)


owen-d (approved):

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -211,13 +211,16 @@ func (c *Compactor) running(ctx context.Context) error {
 	for {
 		select {
 		case <-ticker.C:
+			start := time.Now()
```
chaudum, line 214: convenience: `<- ticker.C` yields the time of the tick

pkg/bloomcompactor/bloomcompactor.go
```diff
@@ -211,13 +211,16 @@ func (c *Compactor) running(ctx context.Context) error {
...
 			c.metrics.compactionRunsStarted.Inc()
 			if err := c.runCompaction(ctx); err != nil {
-				c.metrics.compactionRunsFailed.Inc()
+				c.metrics.compactionRunsCompleted.WithLabelValues("fail").Inc()
```
chaudum, line 217: Do we have common name for the label value in the failure case? I think we should try to be consistent with existing label names.

chaudum:
The e2e test rely on one of these metrics, see https://github.com/grafana/loki/blob/main/integration/loki_micro_services_test.go#L1223-L1230 I think you will need to adjust the assertion there.

pkg/bloomcompactor/bloomcompactor.go
salvacorts: Good call. I see we use "failure" for failed statuses. I've added constants for both statuses.

chaudum (approved):

salvacorts:
The `TestBloomFiltersEndToEnd` started to be flanky today since we set the client time to `Fri Jan 19 2024 00:00` but the server's now is the actual _now_ date. Since the max_table_age is 7 days, the test started to fail today.

This PR fixes it as well, but the `test_target_branch` drone stage will fail until we merge this PR.

PR #11787, opened 2024-01-25
[Blooms] Fix panic when accessing page outside of page bounds

owen-d:
Move variable declaration to only access the desired page when it's valid. Also removes an unused (duplicate) test fn.

### DIFF
--- a/pkg/storage/bloom/v1/index_querier.go
+++ b/pkg/storage/bloom/v1/index_querier.go
@@ -52,8 +52,6 @@ func (it *LazySeriesIter) Seek(fp model.Fingerprint) error {
 		return header.ThroughFp >= fp
 	})
 
-	page := it.b.index.pageHeaders[desiredPage]
-
 	switch {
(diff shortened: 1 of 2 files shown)


pkg/storage/bloom/v1/index_querier.go
```diff
@@ -52,8 +52,6 @@ func (it *LazySeriesIter) Seek(fp model.Fingerprint) error {
 		return header.ThroughFp >= fp
 	})
 
-	page := it.b.index.pageHeaders[desiredPage]
-
 	switch {
 	case desiredPage == len(it.b.index.pageHeaders):
```
chaudum, line 56: Could it be possible that `desiredPage` is greater that `len(it.b.index.pageHeaders)`?

owen-d: No, according to the [search](https://pkg.go.dev/sort#Search) fn used.

chaudum (approved):

PR #11789, opened 2024-01-25
feat: add release please config (#11333)

trevorwhitney:

The new release process is being developed in another repo for now as it enables much faster feedback loops. We are at the stage where we are pulling the Loki repo and building images, so it's time to move the `release-please` config into Loki, and configure it for the various packages we will be releasing.

This PR adds the two basic configs that `release-please` requires. This PR does not change anything about the release process, it just enables us to keep iterating out of band until we're ready to move over.

This was already added to `main`, but adding here so I can test it on the release branch

Re #11332

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -281,7 +281,16 @@ cmd/migrate/migrate:
 #############
 GOX = gox $(GO_FLAGS) -output="dist/{{.Dir}}-{{.OS}}-{{.Arch}}"
 CGO_GOX = gox $(DYN_GO_FLAGS) -cgo -output="dist/{{.Dir}}-{{.OS}}-{{.Arch}}"
+
+SKIP_ARM ?= false
 dist: clean
+ifeq ($(SKIP_ARM),true)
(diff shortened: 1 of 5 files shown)


paul1r (approved):

Makefile
```diff
@@ -281,14 +281,24 @@ cmd/migrate/migrate:
 #############
 GOX = gox $(GO_FLAGS) -output="dist/{{.Dir}}-{{.OS}}-{{.Arch}}"
 CGO_GOX = gox $(DYN_GO_FLAGS) -cgo -output="dist/{{.Dir}}-{{.OS}}-{{.Arch}}"
+
+SKIP_ARM ?= false
 dist: clean
+ifeq ($(SKIP_ARM),true)
```
MichelHollands, line 287: (nit) The name SKIP_ARM is a bit misleading. You're swapping arm for Intel here.

MichelHollands (approved):
One nit about a variable name

Makefile
trevorwhitney: intel is being built in either case, so the guard is to whether _only_ to build intel, or whether to build intel _and_ arm, unless I'm missing something? problem is GitHub actions don't support arm yet, so we need a way to disable those archs specifically

PR #11792, opened 2024-01-25
Bloom Gateway: Make tasks cancelable

chaudum:

This PR refactors the bloom gateway workers so that tasks that have been enqueued by requests do not end up locking the results channel and therefore the worker, in case the request was cancelled (`context cancelled`) or timed out (`context deadline exceeded`).
It also handles errors from the shipper in a way that they are returned to the waiting request asap so it can return and does not need to wait for all tasks to finish.

This PR also fixes the worker shutdown in a way that it now gracefully stops and continues to work off the remaining tasks from the queue.

### DIFF
--- a/pkg/bloomgateway/bloomgateway.go
+++ b/pkg/bloomgateway/bloomgateway.go
@@ -411,9 +411,6 @@ outer:
 func consumeTask(ctx context.Context, task Task, tasksCh chan<- Task, logger log.Logger) {
 	logger = log.With(logger, "task", task.ID)
 
-	task.responses = responsesPool.Get(len(task.series))
-	defer responsesPool.Put(task.responses)
-
 	for res := range task.resCh {
 		select {
 		case <-ctx.Done():
--- a/pkg/bloomgateway/bloomgateway.go
+++ b/pkg/bloomgateway/bloomgateway.go
@@ -371,11 +371,11 @@ outer:
 	for {
 		select {
 		case <-ctx.Done():
-			return nil, errors.Wrap(ctx.Err(), "waiting for results")
+			return nil, errors.Wrap(ctx.Err(), "request failed")
 		case task := <-tasksCh:
 			level.Info(logger).Log("msg", "task done", "task", task.ID, "err", task.Err())
 			if task.Err() != nil {
-				return nil, errors.Wrap(err, "waiting fo result")
+				return nil, errors.Wrap(task.Err(), "request failed")
 			}
 			responses = append(responses, task.responses...)
 			remaining--
@@ -405,6 +405,7 @@ outer:
 // until the task is closed, however it does not forward them any mode.
 func consumeTask(ctx context.Context, task Task, tasksCh chan<- Task, logger log.Logger) {
 	logger = log.With(logger, "task", task.ID)
+
 	task.responses = responsesPool.Get(len(task.series))
 	defer responsesPool.Put(task.responses)
 
@@ -417,6 +418,7 @@ func consumeTask(ctx context.Context, task Task, tasksCh chan<- Task, logger log
 			task.responses = append(task.responses, res)
 		}
 	}
+	level.Debug(logger).Log("task error", task.Err())
 
 	select {
 	case <-ctx.Done():
--- a/pkg/bloomgateway/bloomgateway.go
+++ b/pkg/bloomgateway/bloomgateway.go
@@ -330,17 +330,17 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
 	})
 
 	var numSeries int
-	seriesWithBloomsPerDay := partitionRequest(req)
+	seriesByDay := partitionRequest(req)
 
 	// no tasks --> empty response
-	if len(seriesWithBloomsPerDay) == 0 {
+	if len(seriesByDay) == 0 {
 		return &logproto.FilterChunkRefResponse{
 			ChunkRefs: []*logproto.GroupedChunkRefs{},
 		}, nil
 	}
 
-	tasks := make([]Task, 0, len(seriesWithBloomsPerDay))
-	for _, seriesWithBounds := range seriesWithBloomsPerDay {
+	tasks := make([]Task, 0, len(seriesByDay))
+	for _, seriesWithBounds := range seriesByDay {
 		task, err := NewTask(ctx, tenantID, seriesWithBounds, req.Filters)
 		if err != nil {
 			return nil, err
--- a/pkg/bloomgateway/bloomgateway.go
+++ b/pkg/bloomgateway/bloomgateway.go
@@ -299,6 +299,8 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
 		return nil, err
 	}
 
+	logger := log.With(g.logger, "tenant", tenantID)
+
 	// start time == end time --> empty response
 	if req.From.Equal(req.Through) {
 		return &logproto.FilterChunkRefResponse{
@@ -327,7 +329,7 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
 		return req.Refs[i].Fingerprint < req.Refs[j].Fingerprint
 	})
 
-	var expectedResponses int
+	var numSeries int
 	seriesWithBloomsPerDay := partitionRequest(req)
 
 	// no tasks --> empty response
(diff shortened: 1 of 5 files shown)


pkg/bloomgateway/multiplexing.go
```diff
@@ -70,17 +100,37 @@ func (t Task) Bounds() (model.Time, model.Time) {
...
-		ErrCh:   t.ErrCh,
-		ResCh:   t.ResCh,
+		err:     t.err,
+		resCh:   t.resCh,
 		filters: t.filters,
 		series:  series,
 		bounds:  t.bounds,
 		day:     t.day,
+		ctx:     t.ctx,
+		done:    make(chan struct{}),
```
chaudum, line 133: `done` is never used on a sub-task (task created by `Copy()`).

pkg/bloomgateway/bloomgateway.go
```diff
@@ -339,67 +341,45 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
 		tasks = append(tasks, task)
-		expectedResponses += len(seriesWithBounds.series)
+		numSeries += len(seriesWithBounds.series)
 	}
 
 	g.activeUsers.UpdateUserTimestamp(tenantID, time.Now())
 
-	errCh := make(chan error, 1)
-	resCh := make(chan v1.Output, 1)
+	tasksCh := make(chan Task)
```
vlad-diachenko, line 354: we know tasks count upfront
```suggestion
	tasksCh := make(chan Task, len(tasks))
```

pkg/bloomgateway/bloomgateway.go
```diff
@@ -339,67 +341,45 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
...
-		case res := <-resCh:
-			responses = append(responses, res)
-			// log line is helpful for debugging tests
-			level.Debug(g.logger).Log("msg", "got partial result", "progress", fmt.Sprintf("%d/%d", len(responses), expectedResponses))
-			// wait for all parts of the full response
-			if len(responses) == expectedResponses {
+			return nil, errors.Wrap(ctx.Err(), "request failed")
+		case task := <-tasksCh:
+			level.Info(logger).Log("msg", "task done", "task", task.ID, "err", task.Err())
+			if task.Err() != nil {
```
vlad-diachenko, line 377: in this case, the first failing task will break the execution and we won't read from `tasksCh` anymore and it will mean that all goroutines started here `go consumeTask(ctx, task, tasksCh, logger)` , will be blocked.... 
unless the channel has enough capacity to keep all the responses, regardless they will be read or not...

pkg/bloomgateway/bloomgateway.go
```diff
@@ -415,10 +395,41 @@ outer:
...
+			task.responses = append(task.responses, res)
+		}
+	}
+
+	select {
+	case <-ctx.Done():
+		// do nothing
+	case <-task.Done():
+		// notify request handler about finished task
+		tasksCh <- task
```
vlad-diachenko, line 429: 👍

pkg/bloomgateway/worker.go
```diff
@@ -104,109 +105,118 @@ func (w *worker) starting(_ context.Context) error {
...
+			task, ok := item.(Task)
+			if !ok {
+				// This really should never happen, because only the bloom gateway itself can enqueue tasks.
 				w.queue.ReleaseRequests(items)
-				continue
+				return errors.Errorf("failed to cast dequeued item to Task: %v", item)
 			}
-			w.metrics.dequeuedTasks.WithLabelValues(w.id).Add(float64(len(items)))
+			level.Debug(w.logger).Log("msg", "dequeued task", "task", task.ID)
+			w.pending.Delete(task.ID)
```
vlad-diachenko, line 142: could you please remind me why we track pending tasks?

vlad-diachenko (approved):
looks great. just some minor comments and one important comment about the size of `tasksCh` in bloomgateway

pkg/bloomgateway/bloomgateway.go
chaudum: I explicitly want an unbuffered channel.

pkg/bloomgateway/bloomgateway.go
chaudum: I tried to reproduce this with[ this test case](https://github.com/grafana/loki/pull/11792/files?diff=unified&w=0#diff-2783be6588fd384b385d0daa07cac14a9d48891dbb96b3eca7c11a57f8666b94R187), but it does not fail.

chaudum: I changed it and added a comment.

pkg/bloomgateway/bloomgateway.go
```diff
@@ -415,10 +398,41 @@ outer:
...
+// result channel and stores them on the task.
+// In case the context task is done, it drains the remaining items until the
+// task is closed by the worker.
+// Once the tasks is closed, it will send the task with the results from the
+// block querier to the supplied task channel.
+func consumeTask(ctx context.Context, task Task, tasksCh chan<- Task, logger log.Logger) {
+	logger = log.With(logger, "task", task.ID)
+
+	task.responses = responsesPool.Get(len(task.series))
+	defer responsesPool.Put(task.responses)
```
salvacorts, line 415: We are returning the task (and therefore its responses), but we are also putting back the responses to the pool. I think this may cause a race condition where another task can overwrite the values of the responses.

pkg/bloomgateway/bloomgateway.go
```diff
@@ -415,10 +398,41 @@ outer:
...
+
+	for res := range task.resCh {
+		select {
+		case <-ctx.Done():
+			level.Debug(logger).Log("msg", "drop partial result", "fp_int", uint64(res.Fp), "fp_hex", res.Fp, "chunks_to_remove", res.Removals.Len())
```
salvacorts, line 420: > In case the context task is done, it drains the remaining items until the task is closed by the worker.

By *drain* I understand removing the enqueued items for this task, but I don't see that happening here. At least, shouldn't we break the for loop over the channel or return?

salvacorts: I guess we expect the context to be checked on the worker level and the resCh to be closed?

pkg/bloomgateway/worker.go
```diff
@@ -104,109 +105,118 @@ func (w *worker) starting(_ context.Context) error {
...
-					w.queue.ReleaseRequests(items)
-					return errors.Errorf("failed to cast dequeued item to Task: %v", item)
-				}
-				level.Debug(w.logger).Log("msg", "dequeued task", "task", task.ID)
-				w.pending.Delete(task.ID)
+			tasksPerDay[task.day] = append(tasksPerDay[task.day], task)
+		}
 
-				tasksPerDay[task.day] = append(tasksPerDay[task.day], task)
+		for day, tasks := range tasksPerDay {
```
salvacorts, line 147: I think we should check early if the context is canceled here. Maybe the context was canceled while processing the previous day.

I see it's done after calling `GetBlockRefs` (which will return an err if the task ctx is closed) but I don't see why not doing it before.

pkg/bloomgateway/bloomgateway.go
```diff
@@ -339,67 +341,48 @@ func (g *Gateway) FilterChunkRefs(ctx context.Context, req *logproto.FilterChunk
 
 	tasks := make([]Task, 0, len(seriesWithBloomsPerDay))
 	for _, seriesWithBounds := range seriesWithBloomsPerDay {
```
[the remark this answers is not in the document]
chaudum: :+1: The intention was `seriesWithBoundsPerDay`

pkg/bloomgateway/bloomgateway.go
chaudum: Good point. Gonna remove the pool.

pkg/bloomgateway/worker.go
chaudum: Agree, we can check at multiple places for cancelled context

salvacorts (approved):

pkg/bloomgateway/bloomgateway.go
```diff
@@ -411,9 +411,6 @@ outer:
 func consumeTask(ctx context.Context, task Task, tasksCh chan<- Task, logger log.Logger) {
 	logger = log.With(logger, "task", task.ID)
 
-	task.responses = responsesPool.Get(len(task.series))
```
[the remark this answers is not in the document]
chaudum: See 256015c7827c0680cd30ff19feda714e2c98b6c3

PR #11793, opened 2024-01-25
[Blooms] Refactoring bloom compactor to isolate state from logic

owen-d:
This PR refactors a bunch of code to separate state+I/O related complexity from logic so we can test and extend it more easily. It also adds more logic for the bloom generator to use.

### DIFF
--- a/pkg/bloomcompactor/v2spec.go
+++ b/pkg/bloomcompactor/v2spec.go
@@ -68,7 +68,7 @@ type BloomGenerator interface {
 // Simple implementation of a BloomGenerator.
 type SimpleBloomGenerator struct {
 	store       v1.Iterator[*v1.Series]
-	chunkLoader chunkLoader
+	chunkLoader ChunkLoader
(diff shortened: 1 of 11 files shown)


chaudum (approved):
lgtm

pkg/bloomcompactor/v2spec.go
```diff
@@ -140,3 +169,172 @@ func (s *SimpleBloomGenerator) Generate(_ context.Context) (skippedBlocks []*v1.
...
+	next := b.work[0]
+	batchSize := min(b.batchSize, len(next.chks))
+	toFetch := next.chks[:batchSize]
+	// update work
+	b.work[0].chks = next.chks[batchSize:]
+	if len(b.work[0].chks) == 0 {
+		b.work = b.work[1:]
+	}
+
+	b.batch, b.err = next.fetcher.FetchChunks(b.ctx, toFetch)
```
salvacorts, line 298: Shouldn't we update `b.cur` here?

PR #11797, opened 2024-01-26
DequeueMany fix

vlad-diachenko:
Previously we relied only on the index of the tenant's queue to read the requests from the same tenant's queue. 
However, as long as the queue is aggressively used by the consumers in parallel, there are a few edge cases when knowing the index of last used tenant's queue is not enough to guarantee that we dequeue items from exactly the same tenant's queue. 

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/queue/queue_test.go
```diff
@@ -333,6 +336,208 @@ func TestMaxQueueSize(t *testing.T) {
...
+						if err == nil {
+							//tenant name is sent as task
+							tenant := tasks[0]
+							receivedTasksPerTenant[fmt.Sprintf("%v", tenant)] += len(tasks)
+						}
+						// put the slice back to the pool
+						queue.ReleaseRequests(tasks)
+						if receivedTasksCount.Load() == int32(totalTasksCount) {
+							//cancel the context to release the rest of the consumers once we received all the tasks from all the queues
+							cancel()
```
chaudum, line 495: I don't really understand why we need to call `cancel()` here.
`cancel()` is deferred on line 641, and I don't see the `ctx.Err()` checked anywhere.

pkg/queue/queue_test.go
```diff
@@ -333,6 +336,208 @@ func TestMaxQueueSize(t *testing.T) {
...
+							return fmt.Errorf("error while dequeueing many task by %s: %w", consumer, err)
+						}
+						if err == nil && len(tasks) == 0 {
+							return fmt.Errorf("%s must receive at least one item if there is no error, but got 0", consumer)
+						}
+						if len(tasks) > 1 && !isAllItemsSame(tasks) {
+							return fmt.Errorf("expected all items to be from the same tenant, but they were not: %v", tasks)
+						}
+						time.Sleep(tt.consumerDelay)
+						mtx.Lock()
```
chaudum, line 485: For better readability, maybe defer `mtx.Unlock()` here, instead of multiple times on different code branches.

pkg/queue/queue.go
```diff
@@ -135,80 +136,110 @@ func (q *RequestQueue) ReleaseRequests(items []Request) {
...
-		if len(items) == maxItems {
-			return items, idx, nil
+		idx = newIdx.ReuseLastIndex()
+		if len(items) == maxItems || tenantQueueEmpty {
+			return items, newIdx, nil
 		}
 	}
 }
 
+const anyQueueName = ""
```
chaudum, line 165: nit: move them to the other const at the top of the file. also, maybe name it `anyQueue`
The queue does not really have a concept of "named queues". The names are the tenant ids by coincident.

pkg/queue/queue.go
```diff
@@ -135,80 +136,110 @@ func (q *RequestQueue) ReleaseRequests(items []Request) {
...
+			return nil, last, queue.Name(), false, ErrQueueWasRemoved
+		}
 		// Pick next request from the queue.
-		for {
-			request := queue.Dequeue()
-			if queue.Len() == 0 {
-				q.queues.deleteQueue(tenant)
-			}
+		request := queue.Dequeue()
+		tenantQueueEmpty := queue.Len() == 0
```
chaudum: nit: In Go, boolean variable names are mostly prefixed with `is/Is` or `has/Has` etc.

```suggestion
		isEmpty := queue.Len() == 0
```

pkg/queue/queue.go
vlad-diachenko: done

pkg/queue/queue_test.go
vlad-diachenko: as long as we do not know how many iterations we have, each worker will run this `for` to dequeueMany until the context is cancelled... so, once the counter `receivedTasksCount` reaches the expected number, we need to cancel the context to stop the rest of the goroutines... 
`defere cancel()` is here just to avoid leaking goroutines in case the tests fail. 

and `ctx.Err()` is checked inside the queue.

pkg/queue/queue_test.go
vlad-diachenko: done

pkg/queue/queue.go
vlad-diachenko: done, but I left it as `isTenantQueueEmpty` because I want to add additional meaning that not the whole queue is empty, but only tenant's queue

salvacorts (approved):

chaudum (approved):

PR #11800, opened 2024-01-26
Update Helm chart docs and yaml

poyzannur:
Following recent loki `v2.9.4` and GEL `v1.8.6` patch releases, updates Helm chart docs and yaml.

Loki release drone creates is the [PR as expected](https://github.com/grafana/loki/pull/11769/files), but there seems to be a problem with the GEL drone generating the docs -which is looked at separately. 
Instead I manually updated them in one go.


MichelHollands (approved):
LGTM

trevorwhitney (approved):

JStickler:
I think you're missing a file here?  Do we need to update line 379 in `production/helm/loki/values.yaml`?

poyzannur:
> I think you're missing a file here?  Do we need to update line 379 in `production/helm/loki/values.yaml`?  

Yes I do. Unexpected, the tests were failing if i add that change. Because make doc command doesn't update correctly. I'll debug the make doc script on Monday.

PR #11802, opened 2024-01-26
operator: Add support for running with Azure Workload Identity

xperimental:

This PR adds support for configuring LokiStack to make use of [Azure Workload Identity](https://learn.microsoft.com/en-us/entra/workload-id/workload-identities-overview) when communicating with the Azure Blob Storage.

[LOG-4546](https://issues.redhat.com/browse/LOG-4546)

- [ ] I have only tested this in the `AzureGlobal` environment so far. It might make sense to try to run it in another Azure environment as well to see if the configuration is sufficient.
- [x] I have changed how "missing secret field" is reported, so that the name of the missing field is present in the error. This is currently only changed for the Azure secret but can be extended to other secrets as well.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated


periklis (commented):
Overall looks good to me. Will have a look on a live cluster to get a feeling about it.

operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -165,25 +168,58 @@ func extractAzureConfigSecret(s *corev1.Secret) (*storage.AzureStorageConfig, er
...
+func validateAzureCredentials(s *corev1.Secret) (workloadIdentity bool, err error) {
+	accountName := s.Data[storage.KeyAzureStorageAccountName]
+	accountKey := s.Data[storage.KeyAzureStorageAccountKey]
+	clientID := s.Data[storage.KeyAzureStorageClientID]
+	tenantID := s.Data[storage.KeyAzureStorageTenantID]
+	subscriptionID := s.Data[storage.KeyAzureStorageSubscriptionID]
+
+	if len(accountName) == 0 {
+		return false, fmt.Errorf("%w: %s", errSecretMissingField, storage.KeyAzureStorageAccountName)
+	}
```
periklis, line 196: The validation does not look correct to me at this point. IIRC the only fields we should require for workload identity are: `environment`, `clientID`, `tenantID` and `subscriptionID`. If I am not mistaken we always require `accountName` here but it's not used by the azure SDK afaics.

periklis (approved):

PR #11810, opened 2024-01-27
[Bloom] Keyspace aware bloom creation, library utilities

owen-d:
This PR continues work refactoring the bloom compactor pkg and adds a bunch of utilities for keyspace manipulation/mgmt. A lot of the bloom block logic is contained here as a skeleton.

This code is not complete, but is fine to merge as we don't use this yet and should be easier to review before it gets too large.

### DIFF
--- a/pkg/bloomcompactor/v2_meta.go
+++ b/pkg/bloomcompactor/v2_meta.go
@@ -5,14 +5,16 @@ import (
 	"hash"
 	"path"
 
+	"github.com/pkg/errors"
+
 	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
 	"github.com/grafana/loki/pkg/storage/stores/shipper/indexshipper/tsdb"
 	"github.com/grafana/loki/pkg/util/encoding"
-	"github.com/pkg/errors"
 )
 
 const (
 	BloomPrefix = "bloom"
+	MetasPrefix = "metas"
 )
 
 // TODO(owen-d): Probably want to integrate against the block shipper
@@ -41,15 +43,14 @@ type MetaRef struct {
 	Checksum       uint32
 }
 
-// `bloom/<period>/<tenant>/metas/<start_fp>-<end_fp>-<start_ts>-<end_ts>-<checksum>.json`
+// `bloom/<period>/<tenant>/metas/<start_fp>-<end_fp>-<checksum>.json`
 func (m MetaRef) Address(tenant string, period int) (string, error) {
 	joined := path.Join(
 		BloomPrefix,
 		fmt.Sprintf("%v", period),
 		tenant,
-		"metas",
-		m.OwnershipRange.String(),
-		fmt.Sprintf("%v", m.Checksum),
+		MetasPrefix,
+		fmt.Sprintf("%v-%v", m.OwnershipRange, m.Checksum),
 	)
 
 	return fmt.Sprintf("%s.json", joined), nil
--- a/pkg/bloomcompactor/v2_meta.go
+++ b/pkg/bloomcompactor/v2_meta.go
@@ -5,10 +5,11 @@ import (
 	"hash"
 	"path"
 
+	"github.com/pkg/errors"
+
 	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
 	"github.com/grafana/loki/pkg/storage/stores/shipper/indexshipper/tsdb"
 	"github.com/grafana/loki/pkg/util/encoding"
-	"github.com/pkg/errors"
 )
 
 // TODO(owen-d): Probably want to integrate against the block shipper
--- a/pkg/bloomcompactor/v2_meta.go
+++ b/pkg/bloomcompactor/v2_meta.go
@@ -8,9 +8,18 @@ import (
 	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
 	"github.com/grafana/loki/pkg/storage/stores/shipper/indexshipper/tsdb"
 	"github.com/pkg/errors"
+	"github.com/prometheus/common/model"
 )
 
-type BlockRef struct{}
+// TODO(owen-d): Probably want to integrate against the block shipper
+// instead of defining here, but only (min,max,fp) should be required for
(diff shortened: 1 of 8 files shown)


chaudum (approved):
lgtm. I should be able to use a lot of the new library utilities in https://github.com/grafana/loki/pull/11812

pkg/storage/bloom/v1/bounds.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil
+	}
+
+	return &FingerprintBounds{
+		Min: max(b.Min, target.Min),
+		Max: min(b.Max, target.Max),
+	}
+}
+
+// Union returns the union of the two bounds
```
salvacorts, line 88: I find it a bit strange that when the union of two bounds is empty, this function returns both bounds sorted by FP. I think this should return nil and expect the user to use Cmp, or add a comment explaining that.

pkg/storage/bloom/v1/bounds.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return []FingerprintBounds{
+		{
+			Min: min(b.Min, target.Min),
+			Max: max(b.Max, target.Max),
+		},
+	}
+}
+
+// Unless returns the the subspace of itself which does not intersect with the target bounds
```
salvacorts, line 105:
```suggestion
// Unless returns the subspace of itself which does not intersect with the target bounds
```

pkg/storage/bloom/v1/bounds.go
```diff
@@ -0,0 +1,146 @@
...
+	return b.Clip(FingerprintBounds{Min: min, Max: max})
+}
+
+// Returns whether the fingerprint is fully within the target bounds
+func (b FingerprintBounds) Within(target FingerprintBounds) bool {
+	return b.Min >= target.Min && b.Max <= target.Max
+}
+
+// Intersection returns the intersection of the two bounds
+func (b FingerprintBounds) Intersection(target FingerprintBounds) *FingerprintBounds {
```
salvacorts, line 77: The body is the same as Clip. Make Intersection call Clip

pkg/bloomcompactor/v2_meta.go
```diff
@@ -0,0 +1,131 @@
...
+
+type MetaStore interface {
+	GetMetas([]MetaRef) ([]Meta, error)
+	PutMeta(Meta) error
+	ResolveMetas(bounds v1.FingerprintBounds) ([]MetaRef, error)
+}
+
+type BlockStore interface {
+	// TODO(owen-d): flesh out|integrate against bloomshipper.Client
+	GetBlocks([]BlockRef) ([]interface{}, error)
```
salvacorts, line 129: Why do we take and return `interface{}` instead of `v1.Block` here?

salvacorts (approved):

pkg/bloomcompactor/v2_meta.go
```diff
@@ -0,0 +1,131 @@
...
+
+// `bloom/<period>/<tenant>/metas/<start_fp>-<end_fp>-<start_ts>-<end_ts>-<checksum>`
+func (m MetaRef) Address(userID string, period int) (string, error) {
+	joined := path.Join(
+		"bloom",
+		fmt.Sprintf("%v", period),
+		userID,
+		"metas",
+		m.OwnershipRange.String(),
+		fmt.Sprintf("%v", m.Checksum),
```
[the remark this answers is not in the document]
owen-d: Mostly habit, it's just an alias for default formatting option [here](https://pkg.go.dev/fmt).

pkg/bloomcompactor/v2_meta.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+type TSDBStore interface {
+	ResolveTSDBs() ([]*tsdb.TSDBFile, error)
+}
+
+type MetaStore interface {
+	GetMetas([]MetaRef) ([]Meta, error)
+	PutMeta(Meta) error
+	ResolveMetas(bounds v1.FingerprintBounds) ([]MetaRef, error)
```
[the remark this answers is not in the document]
owen-d: The more I think about it, the more I think we should skip time ranges in the beginning because I'd expect all blocks to cover the entire time range (we put data in blocks based on fingerprint instead of time so it's likely to have data points from across the whole time range).

pkg/storage/bloom/v1/bounds.go
owen-d: Are you thinking `intersection`? [This](https://en.wikipedia.org/wiki/Set_(mathematics)#Basic_operations) may help.

pkg/storage/bloom/v1/bounds.go
owen-d: Haha, thanks. Going to prefer intersection but good catch

pkg/bloomcompactor/v2_meta.go
owen-d: Mainly because i haven't figured out what the return type should be (which may or may not be block)

pkg/bloomcompactor/v2_meta.go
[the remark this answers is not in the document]
chaudum: I like the idea that metas are just for a single day. It makes them definitely easier to handle.

pkg/bloomcompactor/v2_meta.go
```diff
@@ -0,0 +1,134 @@
...
+
+// `bloom/<period>/<tenant>/metas/<start_fp>-<end_fp>-<start_ts>-<end_ts>-<checksum>.json`
+func (m MetaRef) Address(tenant string, period int) (string, error) {
+	joined := path.Join(
+		BloomPrefix,
+		fmt.Sprintf("%v", period),
+		tenant,
+		"metas",
+		m.OwnershipRange.String(),
+		fmt.Sprintf("%v", m.Checksum),
```
chaudum, line 52: `<start>-<end>-<checksum>`

```suggestion
		strings.Join([]string{m.OwnershipRange.String(), fmt.Sprintf("%x", m.Checksum)}, "-")
```

PR #11812, opened 2024-01-29
Bloom Gateway: Extract processing logic from bloom worker into separate struct

chaudum:

The processor executes a set of tasks.
It cleanly separates I/O from logic and therefore is more testable than the current worker implementation.

https://github.com/grafana/loki/blob/9602214abf5f0b016f1cad90e921d1e4d969856c/pkg/bloomgateway/processor.go#L33-L36
    
**Note**:

~~Only commit https://github.com/grafana/loki/pull/11812/commits/cd28931055a38827f3a8a739fe2a64c165b3fc89 is relevant. The rest of the commits are part of different PRs (#11792 and #11809), which need to be merged first.~~

The processor is not used yet in the worker.

**Certain parts must be refactored when #11810 is merged.**

### DIFF
--- a/pkg/bloomgateway/processor.go
+++ b/pkg/bloomgateway/processor.go
@@ -6,9 +6,10 @@ import (
 	"sort"
 
 	"github.com/go-kit/log"
+	"github.com/prometheus/common/model"
+
 	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
 	"github.com/grafana/loki/pkg/storage/stores/shipper/bloomshipper"
-	"github.com/prometheus/common/model"
 )
 
 type tasksForBlock struct {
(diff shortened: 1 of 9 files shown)


owen-d (approved):

pkg/bloomgateway/processor.go
```diff
@@ -0,0 +1,169 @@
...
+	blockLoader
+	metaLoader
+}
+
+type processor struct {
+	store  store
+	logger log.Logger
+}
+
+func (p *processor) run(ctx context.Context, tasks []Task) error {
```
salvacorts, line 38: nit: I feel like `run`, `do` and `process` not descriptive enough names and to me they mean pretty much the same.

salvacorts (approved):
LGTM

pkg/bloomgateway/processor.go
chaudum: Good point. I will give them more descriptive names.

PR #11814, opened 2024-01-29
feat(caching): Support caching for instant metric query results

kavirajk:
Follow up to [metadata results caching](https://github.com/grafana/loki/pull/11539), this PR adds support for instant metric queries.

Changes:
1. Boolean flag to enable this type of caching (`cache_instant_metric_results`)
2. CacheConfig to tweak (usually not needed, have sane defaults) (`instant_metric_cache`)
3. Per-tenant overrides config to set split interval specific to instant metric queries caching. (`split_instant_metric_queries_by_interval`)

Will have follow up PRs (tried doing it in single PR, but turns out really hard for the reviewers) to 
1. Refactor configs  to unify and simplify all the results cache configs
2. Refactor results cache metrics, to avoid lots of duplicates
3. Simplify some protobuf definitions (particularly `stats.proto`) 

Fixes #<issue number>

TODOs before making it ready for review.
1.  Remove offset from subqueries generated by range mapper before generating caching keys.
2. Adding some tests 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


pkg/querier/queryrange/downstreamer.go
```diff
@@ -94,16 +99,50 @@ type instance struct {
...
+		newEnd   = query.Params.End()
+	)
+	expr.Walk(func(e syntax.Expr) {
+		switch rng := e.(type) {
+		case *syntax.RangeAggregationExpr:
+			off := rng.Left.Offset
+
+			if off != 0 {
+				rng.Left.Offset = 0     // remove offset
+				if newStart == newEnd { // instant query
```
ashwanthgoli, line 121: shouldn't we update the end irrespective of the range_type?

ashwanthgoli: also should we always expect a single offset in the expr? I am thinking of binary expressions that could use an offset on both operands. but [this code](https://github.com/grafana/loki/blob/85908fafdfc9261c2f12f630726fa0dad514f886/pkg/logql/rangemapper.go#L319) also expects a most 1 for some reason

pkg/logql/rangemapper.go
```diff
@@ -327,6 +340,74 @@ func (m RangeMapper) getOriginalOffset(expr syntax.SampleExpr) (offset time.Dura
...
+	newRng = m.splitByInterval // [1h]
+
+	// Rest of the subqueries. Always aligned with splitBy
+	for i := 0; i < splits-2; i++ {
+		downstreams = appendDownstream(downstreams, expr, newRng, newOffset)
+		newOffset += m.splitByInterval
+	}
+
+	// last subquery
+	newRng = m.splitByInterval - align // e.g: [24m]
```
ashwanthgoli, line 400: could this be 0 or neg for the last split? thinking if we will always have splits+1 when the query is unaligned.
what do you think about generating sub-queries as long as rangeInterval > 0 decreasing it with each generated query.

1. reduce rangeInterval by `align` for the first query
2. keep adding sub-queries that are aligned to splitInterval and sub `splitByInterval` from `rangeInterval`
3. when the left over `rangeInterval` is smaller than `splitByInterval`, use it to generate the last query

pkg/querier/queryrange/downstreamer.go
ashwanthgoli: instant queries with binary expr that uses multiple offsets may be broken already

kavirajk: > shouldn't we update the end irrespective of the range_type?

Good catch. you are right. End need to be updated irrespective of `range_type`. I think I added this in the very beginning to catch any instances if we use RangeMapper for range queries (turns out we use it only for instant queries).

Will update it 👍 

> also should we always expect a single offset in the expr? I am thinking of binary expressions that could use an offset on both operands. 
> instant queries with binary expr that uses multiple offsets may be broken already

I think [we split binary expressions into two expressions](https://github.com/grafana/loki/blob/kavirajk%2Fcache-instant-queries2/pkg/logql/rangemapper.go#L142) before mapping it into actual downstream `ConcatExpr` type. 

At this level, where we create and append downstream expressions, there cannot be two offsets in a single expression.

pkg/logql/rangemapper.go
kavirajk: > could this be 0 or neg for the last split?

Failing to understand how this could be possible. This is based on the invariant that `splitByInterval` will **always be greater than** `align`.

If you see how we calculate `align` in the first place, 
```
align := m.splitAlignTs.Sub(m.splitAlignTs.Truncate(m.splitByInterval))       // say, 12:34:00 - 12:00:00(truncated) = 34m
```
wondering at what case, this `align` can be greater than or equal to `splitbyInterval`.  Am I missing anything here?

Also, the approach you explained focus only on the `rangeInterval` but not on the `offset` we generate for each subqueries. 

So basically we don't subtract `splitByInterval` from `rangeInterval` for all subqueries, but instead we adjust the offset. keeping `rangeInterval` aligned with `splitByInterval` (except first and last subquery). 

So once we remove offset and adjust start and end time we have sub queries that have aligned both `rangeInterval` and `execTs`.

May be I'm missing something that you are trying to explain?

ashwanthgoli: taking the same example from the comments, if we take the case where orig query uses a  range of 2h34m and executes at 12:34:00 `sum(rate({foo="bar"}[2h34m])`

these would be the expected sub queries:
1. sum(rate({foo="bar"}[34m]))
2. sum(rate({foo="bar"}[1h] offset 34m))
3. sum(rate({foo="bar"}[1h] offset 1h34m))

4th sub-query would not be need in this case. don't think it could be a neg value, but 0 is a possibility but a very unlikely one. we could just as well solve this with a check to see if the left over range is gt 0.

kavirajk: > 4th sub-query would not be need in this case. don't think it could be a neg value, but 0 is a possibility but a very unlikely one. we could just as well solve this with a check to see if the left over range is gt 0.

ACK that. I cleanup up the method to address this part 👍

docs/sources/configure/_index.md
```diff
@@ -886,6 +886,28 @@ volume_results_cache:
...
+instant_metric_results_cache:
+  # The cache block configures the cache backend.
+  # The CLI flags prefix for this block configuration is:
+  # frontend.instant-metric-results-cache
+  [cache: <cache_config>]
+
+  # Use compression in cache. The default is an empty value '', which disables
+  # compression. Supported values are: 'snappy' and ''.
+  # CLI flag: -frontend.instant-metric-results-cache.compression
+  [compression: <string> | default = ""]
```
dannykopping, line 904: Why are we disabling compression by default?

pkg/logql/rangemapper.go
```diff
@@ -57,6 +57,19 @@ type RangeMapper struct {
 	splitByInterval time.Duration
 	metrics         *MapperMetrics
 	stats           *MapperStats
+
+	// SplitAlign
```
dannykopping, line 61: Nit: not seeing the value of this comment

```suggestion
```

pkg/logql/rangemapper_test.go
```diff
@@ -1675,7 +1753,7 @@ func Test_SplitRangeVectorMapping(t *testing.T) {
 		// Non-splittable vector aggregators - should go deeper in the AST
 		{
 			`topk(2, count_over_time({app="foo"}[3m]))`,
-			`topk(2, 
+			`topk(2,
```
dannykopping, line 1756: I'm seeing a few of these changes in this PR where trailing spaces are removed.
I'd suggest you commit with `-p` to only include the changes relevant to the PR; it makes the PR larger and adds some irrelevant detail.

pkg/logqlmodel/stats/stats.proto
```diff
@@ -57,6 +57,10 @@ message Caches {
     (gogoproto.nullable) = false,
     (gogoproto.jsontag) = "labelResult"
   ];
+  Cache instantMetricResult = 8 [
```
dannykopping, line 60: Nit: this should be called `instantMetricQueryResult` IMHO.

pkg/querier/queryrange/codec.go
```diff
@@ -109,7 +109,7 @@ func (r *LokiInstantRequest) GetStart() time.Time {
 
 func (r *LokiInstantRequest) WithStartEnd(s time.Time, _ time.Time) queryrangebase.Request {
 	clone := *r
-	clone.TimeTs = s
+	clone.TimeTs = time.Unix(0, s.UnixNano())
```
dannykopping, line 112: This might be changing some semantics; can you explain what the purpose of this change is pls?

pkg/querier/queryrange/instant_metric_cache.go
```diff
@@ -0,0 +1,83 @@
...
+	return fmt.Sprintf("instant-metric:%s:%s:%d:%d", userID, r.GetQuery(), currentInterval, split)
+}
+
+type InstantMetricCacheConfig struct {
+	queryrangebase.ResultsCacheConfig `yaml:",inline"`
+}
+
+// RegisterFlags registers flags.
+func (cfg *InstantMetricCacheConfig) RegisterFlags(f *flag.FlagSet) {
+	cfg.RegisterFlagsWithPrefix(f, "frontend.instant-metric-results-cache.")
```
dannykopping, line 45:
```suggestion
	cfg.RegisterFlagsWithPrefix(f, "frontend.instant-metric-query-results-cache.")
```

docs/sources/configure/_index.md
```diff
@@ -886,6 +886,28 @@ volume_results_cache:
   # CLI flag: -frontend.volume-results-cache.compression
   [compression: <string> | default = ""]
 
+# Cache instant metric query results.
+# CLI flag: -querier.cache-instant-metric-results
+[cache_instant_metric_results: <boolean> | default = false]
+
+# If a cache config is not specified and cache_instant_metric_results is true,
+# the config for the results cache is used.
+instant_metric_results_cache:
```
dannykopping, line 895:
```suggestion
instant_metric_query_results_cache:
```

to be consistent with `instant_metric_query_split_align`.

pkg/querier/queryrange/roundtrip.go
```diff
@@ -761,19 +775,71 @@ func NewMetricTripperware(cfg Config, engineOpts logql.EngineOpts, log log.Logge
...
+	metrics *Metrics,
+	merger base.Merger,
+	c cache.Cache,
+	cacheGenNumLoader base.CacheGenNumberLoader,
+	retentionEnabled bool,
+	indexStatsTripperware base.Middleware,
+	metricsNamespace string,
+) (base.Middleware, error) {
+	var cacheMiddleware base.Middleware
+	if cfg.CacheInstantMetricResults {
```
dannykopping, line 793: Generally prefer the happy path to be as unindented as possible; so you could invert this `if` and exit early.

dannykopping (commented):
**Love** the detailed description and lots of tests!
Added a few comments, mostly nits. Good job Kavi!

pkg/logql/rangemapper.go
```diff
@@ -327,6 +340,76 @@ func (m RangeMapper) getOriginalOffset(expr syntax.SampleExpr) (offset time.Dura
...
+	newRng = m.splitByInterval // [1h]
+
+	// Rest of the subqueries.
+	for pendingRangeInterval > 0 {
+		if pendingRangeInterval < m.splitByInterval {
+			newRng = pendingRangeInterval // last subquery
+		}
+		downstreams = appendDownstream(downstreams, expr, newRng, newOffset)
+		newOffset += m.splitByInterval
+		pendingRangeInterval -= m.splitByInterval
```
ashwanthgoli, line 401:
```suggestion
		pendingRangeInterval -= newRng
```

pkg/logql/rangemapper.go
```diff
@@ -327,6 +340,76 @@ func (m RangeMapper) getOriginalOffset(expr syntax.SampleExpr) (offset time.Dura
...
+
+	align := m.splitAlignTs.Sub(m.splitAlignTs.Truncate(m.splitByInterval)) // say, 12:34:00 - 12:00:00(truncated) = 34m
+
+	if align == 0 {
+		return m.rangeSplit(expr, rangeInterval, recorder) // Don't have to align
+	}
+
+	var (
+		newRng               = align
+		newOffset            = originalOffset
```
ashwanthgoli, line 380: if the originalOffset is non-zero, we might have to account for it when aligning as well because the effective timestamp after removing the offset in the downstreamer would not align.

if it's unwieldy, we could skip aligning for queries with origOffset for the first pr?

ashwanthgoli:
I need to do another pass to review the tests, rest lgtm. Nice one @kavirajk :heart:

pkg/querier/queryrange/roundtrip.go
kavirajk: true. But I think it applies only if we **could** just return and exit without needing any additional steps. 

Here in `NewInstantMetricTripperware` we do some additional steps in either case where `cfg.CacheInstantMetricResults` is true or false.

If I invert, I would still need more `if` to create other middlewares. Basically we cannot just return or exist directly if it is unhappy path

Did that answer your concern or am I failing to understand something else?

docs/sources/configure/_index.md
kavirajk: same rationale like previously mentioned. Tried to be consistent with other results cache.

```
-bash-5.2$ rg "results_cache" | rg "yaml:\"" | awk '{print $4}'
`yaml:"results_cache,omitempty"`
`yaml:"index_stats_results_cache"
`yaml:"volume_results_cache"
`yaml:"instant_metric_results_cache"
`yaml:"series_results_cache"
`yaml:"label_results_cache"
`yaml:"results_cache"`
```

pkg/querier/queryrange/codec.go
kavirajk: Ah. Good catch. Actually it's not needed. I changed to validate some tests and forgot to revert it. Reverted now.

pkg/logqlmodel/stats/stats.proto
kavirajk: I tried to be in consistent with other cases here.

```
  Cache statsResult = 4 [
    (gogoproto.nullable) = false,
    (gogoproto.jsontag) = "statsResult"
  ];
  Cache volumeResult = 5 [
    (gogoproto.nullable) = false,
    (gogoproto.jsontag) = "volumeResult"
  ];
  Cache seriesResult = 6 [
    (gogoproto.nullable) = false,
    (gogoproto.jsontag) = "seriesResult"
  ];
  Cache labelResult = 7 [
    (gogoproto.nullable) = false,
    (gogoproto.jsontag) = "labelResult"
  ];
```
None of them have `Query` to it. I would prefer to keep it this way for consistency, I can change if you really want to.

docs/sources/configure/_index.md
kavirajk: Actually, this is the auto-generated with `make doc`. Looks like this [default](https://github.com/grafana/loki/blob/kavirajk%2Fcache-instant-queries2/pkg/storage/chunk/cache/resultscache/config.go#L21) for all [other caches ](https://github.com/grafana/loki/blob/e373341ee7f294bdba535f9905543c662ed91de0/docs/sources/configure/_index.md#query_range)as well.

pkg/logql/rangemapper.go
kavirajk: Oh! actually great catch. But this may complicate a bit. I propose to add a todo now and handle it in the follow up PR if that's ok for you?

dannykopping (approved):
LGTM!

kavirajk:
Caching and split align in action

**Do instant query for `3h` range (first time)**
```
-bash-5.2$ ./cmd/logcli/logcli instant-query 'sum(rate({job="varlogs"}[3h]))'
2024/02/20 08:18:37 http://localhost:3100/loki/api/v1/query?direction=BACKWARD&limit=30&query=sum%28rate%28%7Bjob%3D%22varlogs%22%7D%5B3h%5D%29%29&time=1708413517893995000
[
  {
    "metric": {},
    "value": [
      1708413517.893,
      "0.21814814814814815"
    ]
  }
```

How the query is split and cache reqs and cache hits (these logs lines are from `metrics.go` and `engine.go` on query-frontend and queriers)

Quey-frontend (saying 4 requests were actually made for cache, and none of those got hit)
```
latency=fast query="sum(rate({job=\"varlogs\"}[3h]))" query_hash=1737987035 cache_result_req=4 cache_result_hit=0
```
How the above query got split and run

```
msg="executing query" type=instant query="sum(count_over_time({job=\"varlogs\"}[1h]))" 
msg="executing query" type=instant query="sum(count_over_time({job=\"varlogs\"}[1h]))"
msg="executing query" type=instant query="sum(count_over_time({job=\"varlogs\"}[41m22s106ms]))" msg="executing query" type=instant query="sum(count_over_time({job=\"varlogs\"}[18m37s893ms]))" 
```

**Do the instant query for `3h` range (second time)**

```
-bash-5.2$ ./cmd/logcli/logcli instant-query 'sum(rate({job="varlogs"}[3h]))'
2024/02/20 08:19:11 http://localhost:3100/loki/api/v1/query?direction=BACKWARD&limit=30&query=sum%28rate%28%7Bjob%3D%22varlogs%22%7D%5B3h%5D%29%29&time=1708413551869792000
[
  {
    "metric": {},
    "value": [
      1708413551.869,
      "0.02935185185185185"
    ]
  }
```

How the query is split and cache reqs and cache hits (these logs lines are from `metrics.go` and `engine.go` on query-frontend and queriers)

Quey-frontend (saying 4 requests were actually made for cache, **2 of those got hit**)
```
latency=fast query="sum(rate({job=\"varlogs\"}[3h]))" query_hash=1737987035 cache_result_req=4 cache_result_hit=2
```
How the above query got split and run (only two queries that miss cache hit, got run this time)

```
msg="executing query" type=instant query="sum(count_over_time({job=\"varlogs\"}[19m11s869ms]))" 
msg="executing query" type=instant query="sum(count_over_time({job=\"varlogs\"}[40m48s130ms]))"
```

PR #11816, opened 2024-01-29
added the capability to configure Loki without needing to rebuild the…

Gordejj:
… operator

I added the capability to append template for Loki config without needing to rebuild the operator. In our case, we are developing a private cloud and want to provide logging as a service. This means that multiple user spaces will have clusters deployed with the loki-operator.
Therefore, we cannot rebuild the loki-operator for each such project individually. It's easier for us to create a template with the Loki configuration as an external dependency, which can be provided to the operator at runtime to modify the config-map and manually restart the components.
I understand the best way is to develop resources, but it seems that a quick workaround would also be useful. 

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/operator/internal/handlers/lokistack_create_or_update.go
+++ b/operator/internal/handlers/lokistack_create_or_update.go
@@ -59,6 +59,8 @@ func CreateOrUpdateLokiStack(
 		gwImg = manifests.DefaultLokiStackGatewayImage
 	}
 
+	cfgBase64 := os.Getenv(manifests.EnvRelatedLokiConfig)
+
 	objStore, err := storage.BuildOptions(ctx, k, &stack, fg)
 	if err != nil {
 		return err
@@ -110,6 +112,7 @@ func CreateOrUpdateLokiStack(
 		Timeouts:               timeoutConfig,
 		Tenants:                tenants,
 		OpenShiftOptions:       ocpOptions,
+		ConfigLokiBase64:       cfgBase64,
 	}
 
 	ll.Info("begin building manifests")
--- a/operator/internal/manifests/config.go
+++ b/operator/internal/manifests/config.go
@@ -2,9 +2,11 @@ package manifests
 
 import (
 	"crypto/sha1"
+	"encoding/base64"
 	"fmt"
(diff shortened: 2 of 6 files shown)


periklis:
@Gordejj Thanks for reaching out! It's always good to watch out and listen on problem our users might encounter.

Nevertheless I would like to engage in an conversation first on what's so special on your custom config template that is not supported by the present Loki Operator custom resource namely LokiStack. The PR as proposed breaks actually the systematic approach to give users an opinionated custom resource definition on how to configure Loki and in return not having the need to learn all the nitty gritty details of the super powerful but huge Loki config.

I encourage you to provide a sample of your config even a partial if possible for inspection. It might turn out that we either support your case via the LokiStack CRD (See [docs](https://loki-operator.dev/docs/api.md/#loki-grafana-com-v1-LokiStack)) or we can add support for the feature you need.

Gordejj:
@periklis Thank you for the prompt response. 
I understand that it is necessary to pass such parameters to resources as needed (I will return with more specific information about the required parameters and will leave corresponding PRs). 

However, in the case of providing logging as a service, there might be a need to quickly change something for a user's service installation. If the corresponding field is missing in the resource, some fast track method for its swift modification is needed. For instance, the config map is continuously updated based on an embedded template, and it seems that there is currently no way to quickly do this for a specific installation - only to wait for the resource fields to be updated.

periklis:
@Gordejj I understand your case a bit better now, so let me summarize a bit and propose what you can already to. IIUC you need to patch the loki config on production temporarily in ways that the operator does not support yet in order to get your production up and running. For such scenarios you can use the `lokistack.spec.managementState: Unmanaged` field to make the operator skip reconciliation. This is meant to be the escape hatch for situations where the operator might stay on your toes (e.g. identified bug or remediating a production issue with a config switch that is not exposed).

periklis:
Closing this PR as suggested in previous comment that a better supported approach exists to manifest temporarily changes in the Loki configuration w/o interference by the operator.

Issue #9295, opened 2023-04-26
Implement filters for the Ruler `api/v1/rules` prometheus compatible API


gotjosh:
#### Is your feature request related to a problem? Please describe.

https://github.com/grafana/grafana/issues/66701

In Grafana, a user has the ability to view the information of a single rule. When this page is loaded [Grafana calls the `api/v1/rule`](https://grafana.com/docs/mimir/latest/references/http-api/#list-prometheus-rules) which pulls all of the *state* of rules of that the tenant has.

I've recently implemented filtering by rule name, rule group and file (namespace in the case of mimir) in Prometheus. We should do the same for Loki.

#### Describe the solution you'd like

The rules API does a fanout request to all the other rulers of which the tenant has rules and then aggregates the results before creating the response. 

We have two options:

1. We can do the filtering after we're aggregated the results - it's a shorter implementation path but it means we'll still transmit all of the data across the network to then throw it away.
2. We pass the filters down to the GRPC requests to limit the data returned - it's a bit of a larger effort but gets us a significant reduction of network bandwidth given that this query is called quite often (every 20s) per client using the alerting view in Grafana. 

I'd argue we should go for option 2, given that there's a plan to implement pagination and setting a foundation for filtering the data transmitted between rulers will become even more useful.

The Prometheus equivalent changes are https://github.com/prometheus/prometheus/pull/12270

#### Describe alternatives you've considered

N/A

#### Additional context

N/A

```[tasklist]
### Tasks
- [ ] Change the proto definition of `GroupStateDesc` to allow filtering options by `rule_name`, `rule_group` and `file[]` (namespace in the case of mimir)
- [ ] Allow support for these query filters in the Prometheus compatible endpoint API
- [ ] Document the supported filters
```

The mimir equivalent of this is in: https://github.com/grafana/mimir/issues/4792

gotjosh:
Hi @aminesnow, sadly it isn't what you have implemented is what I described as option 1 and I think the implementation should follow a bit more of what I think it's option 2.

aminesnow:
So if I understand correctly, you're suggesting we pass the filters down to the store and do the filtering there, right?

PR #11817, opened 2024-01-29
ruler: add filtering support for prometheus rules api

ashwanthgoli:
Adds `rule_name`, `rule_group`, `file` and `type` query parameters for filtering the response of `/prometheus/api/v1/rules` endpoint. Replicates mimir's functionality: https://github.com/grafana/mimir/pull/5291

- all of them are optional.
- `type` paremeter accepts either `alert` or `record`
- `rule_name`, `rule_group`, `file` can accept multiple values and they filter the response accordingly.

There is a minor change in behavior: `/prometheus/api/v1/rules` endpoint will no longer return empty rule groups which is inline with both [prometheus](https://github.com/prometheus/prometheus/pull/12270/files#diff-315f251cdd7e93fcec1e7e9505744da1d1828f30d2b61d1f4ce963fa26bf1909R1403) and [mimir](https://github.com/grafana/mimir/pull/5291/files#diff-e5424c21c0e827bd1c9d3f669ed605897696bdc27993bc8bfd7113eba787b49dR1120). This is not a breaking change since rule groups with [no rules fail validation](https://github.com/grafana/loki/blob/27fbd62505f4412e3cb9180b1a5a66518bba9752/pkg/ruler/base/manager.go#L295) and cannot be created.

Fixes https://github.com/grafana/loki/issues/9295

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/ruler/base/api.go
```diff
@@ -143,15 +152,34 @@ func (a *API) PrometheusRules(w http.ResponseWriter, req *http.Request) {
 	userID, err := tenant.TenantID(req.Context())
 	if err != nil || userID == "" {
 		level.Error(logger).Log("msg", "error extracting org id from context", "err", err)
-		respondError(logger, w, "no valid org id found")
+		respondServerError(logger, w, "no valid org id found")
 		return
 	}
 
-	w.Header().Set("Content-Type", "application/json")
```
kavirajk, line 150: Rationale for removing content type? Intentional?

ashwanthgoli: this was a duplicate, we set it later in the same function before returning

pkg/ruler/base/ruler.go
```diff
@@ -799,26 +799,69 @@ func RemoveRuleTokenFromGroupName(name string) string {
...
+	if len(fs) == 0 {
+		return false
+	}
+	_, ok := fs[val]
+	return !ok
+}
+
+func (r *Ruler) getLocalRules(userID string, req *RulesRequest) ([]*GroupStateDesc, error) {
+	var (
+		getRecordingRules = true
```
kavirajk, line 837: I would prefer to create and switch these flags more direct way.

1. start with both false.
2. case AlertingRule => `alertRule=true`
3. case RecordingRule => `recordingRule=true`
4. case AnyRule => `alerRule=true; recordingRule=true`

This way it's more intuitive IMHO.

kavirajk (commented):
LGTM. just one minor recommendation. 👍

pkg/ruler/base/ruler.go
kavirajk: Also one other question. Do you think we also should support `exclude_alerts=<bool>`, just like prometheus rules endpoint does.
https://prometheus.io/docs/prometheus/latest/querying/api/#rules

ashwanthgoli: i'll add `exclude_alerts` in a follow-up

PR #11828, opened 2024-01-30
Bloom shipper: Restructure bloom store

chaudum:

This PR changes the structure of how the bloom client

* The `BloomStore` is the top component interface which implements the `Store` and the `Client` interfaces. The store holds a store entry for each schema period.
* The `bloomStoreEntry` implements the `Store` and `Client` interfaces. It holds a bloom client for a single schema period. Additionally, the store entry also exposes a fetcher for metas (and in the future for blocks), which is responsible for getting data from cache or storage (if not available in cache).
* The `BloomClient` implement the `Client` interface. The bloom client uses an object client for the schema period of the bloom client.
* The `Fetcher` can fetch and cache metas.

This structure is very similar to what we use for the chunk store.

**Note**

Before implementing `FetchBlocks()` in the `BloomStore`, I want to implement the new `FingerprintBounds` type also in the shipper and bloom gateway code to be consistent across components, as well as to make use of the new utilities exposed by them.

The store implementation probably needs some specific test cases around using the correct store entry for different schema period configs. The code however is mostly taken from the chunk store implementation.

### DIFF
--- a/pkg/storage/stores/shipper/bloomshipper/client.go
+++ b/pkg/storage/stores/shipper/bloomshipper/client.go
@@ -9,7 +9,6 @@ import (
 	"path"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/go-kit/log"
 	"github.com/grafana/dskit/concurrency"
@@ -18,7 +17,6 @@ import (
 	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
 	"github.com/grafana/loki/pkg/storage/chunk/client"
 	"github.com/grafana/loki/pkg/storage/config"
-	"github.com/grafana/loki/pkg/util/math"
 )
 
 const (
--- a/pkg/storage/stores/shipper/bloomshipper/client.go
+++ b/pkg/storage/stores/shipper/bloomshipper/client.go
@@ -6,15 +6,15 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"path/filepath"
+	"path"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/go-kit/log"
 	"github.com/grafana/dskit/concurrency"
 	"github.com/prometheus/common/model"
 
-	"github.com/grafana/loki/pkg/storage"
 	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
 	"github.com/grafana/loki/pkg/storage/chunk/client"
 	"github.com/grafana/loki/pkg/storage/config"
@@ -75,7 +75,7 @@ type MetaSearchParams struct {
 type MetaClient interface {
 	// Returns all metas that are within MinFingerprint-MaxFingerprint fingerprint range
 	// and intersect time period from StartTimestamp to EndTimestamp.
-	GetMetas(ctx context.Context, params MetaSearchParams) ([]Meta, error)
+	GetMetas(ctx context.Context, metas []MetaRef) ([]Meta, error)
 	PutMeta(ctx context.Context, meta Meta) error
 	DeleteMeta(ctx context.Context, meta Meta) error
 }
@@ -91,7 +91,7 @@ type Block struct {
 }
 
 type BlockClient interface {
-	GetBlock(ctx context.Context, reference BlockRef) (LazyBlock, error)
+	GetBlock(ctx context.Context, ref BlockRef) (LazyBlock, error)
 	PutBlocks(ctx context.Context, blocks []Block) ([]Block, error)
 	DeleteBlocks(ctx context.Context, blocks []BlockRef) error
 }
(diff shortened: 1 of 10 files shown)


pkg/storage/stores/shipper/bloomshipper/store.go
```diff
@@ -0,0 +1,356 @@
...
+	"github.com/prometheus/common/model"
+
+	"github.com/grafana/loki/pkg/storage"
+	"github.com/grafana/loki/pkg/storage/chunk/cache"
+	"github.com/grafana/loki/pkg/storage/chunk/client"
+	"github.com/grafana/loki/pkg/storage/config"
+)
+
+type Store interface {
+	SearchMetas(ctx context.Context, params MetaSearchParams) ([]Meta, error)
```
chaudum, line 21: Eventually, this should become:

```suggestion
	ResolveMetas(ctx context.Context, params MetaSearchParams) ([]MetaRef, error)
```

To match with

https://github.com/grafana/loki/blob/main/pkg/bloomcompactor/v2_meta.go#L125-L129

pkg/storage/stores/shipper/bloomshipper/store.go
```diff
@@ -0,0 +1,356 @@
...
+		for _, object := range list {
+			metaRef, err := createMetaRef(object.Key, params.TenantID, table)
+
+			if err != nil {
+				return nil, err
+			}
+			if metaRef.MaxFingerprint < uint64(params.Keyspace.Min) || uint64(params.Keyspace.Max) < metaRef.MinFingerprint ||
+				metaRef.EndTimestamp.Before(params.Interval.Start) || metaRef.StartTimestamp.After(params.Interval.End) {
+				continue
+			}
```
chaudum, line 56: Use `Overlap()` from https://github.com/grafana/loki/blob/main/pkg/storage/bloom/v1/bounds.go#L50-L52

pkg/storage/stores/shipper/bloomshipper/store.go
```diff
@@ -0,0 +1,356 @@
...
+			refs = append(refs, metaRef)
+		}
+	}
+	return b.fetcher.FetchMetas(ctx, refs)
```
chaudum, line 60: Should this be done in a separate function call?

```go
refs, fetchers, err := store.SearchMetas(ctx, params)
for i := range refs {
    metas, err := fetchers[i].FetchMetas(ctx, refs[i])
}
```

pkg/storage/stores/shipper/bloomshipper/store.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+func (b *BloomStore) storeDo(ts model.Time, f func(s *bloomStoreEntry) error) error {
+	if store := b.getStore(ts); store != nil {
+		return f(store)
+	}
+	return nil
+}
+
+func (b *BloomStore) forStores(ctx context.Context, interval Interval, f func(innerCtx context.Context, interval Interval, store Store) error) error {
```
chaudum, line 309: Taken from the chunk store implementation.

pkg/storage/stores/shipper/bloomshipper/client.go
```diff
@@ -75,7 +73,7 @@ type MetaSearchParams struct {
 type MetaClient interface {
 	// Returns all metas that are within MinFingerprint-MaxFingerprint fingerprint range
 	// and intersect time period from StartTimestamp to EndTimestamp.
-	GetMetas(ctx context.Context, params MetaSearchParams) ([]Meta, error)
+	GetMetas(ctx context.Context, metas []MetaRef) ([]Meta, error)
```
chaudum, line 76: Should probably be

```suggestion
	GetMetas(ctx context.Context, metas []Meta) ([][]Meta, []Fetcher, error)
```

to adhere with the chunk client interface.

owen-d: Yeah, there's some consistency there, although I think I prefer separate structures for items vs their references, as it makes it very clear what's expected of the data type (no confusion about whether it's loaded yet, etc)

pkg/storage/stores/shipper/bloomshipper/client.go
```diff
@@ -102,84 +100,40 @@ type Client interface {
...
-		}
-		periodicObjectClients[periodicConfig.From] = objectClient
-	}
-	return &BloomClient{
-		periodicConfigs:       periodicConfigs,
-		storageConfig:         storageConfig,
-		periodicObjectClients: periodicObjectClients,
-	}, nil
-}
+var _ Client = &BloomClient{}
```
owen-d, line 103: Nit: this isn't immediately clear to everyone that we're using this as a compiler check.

pkg/storage/stores/shipper/bloomshipper/store.go
owen-d: Yeah I'd keep these separate. Resolving Refs independently of fetching them allows us to do things without needing to download the underlying object. I think that's what you mean?

owen-d (approved):
Nice, thanks for this

pkg/storage/stores/shipper/bloomshipper/store.go
chaudum: Right, that's what I meant.

pkg/storage/stores/shipper/bloomshipper/client.go
chaudum: Will add a comment.

PR #11837, opened 2024-01-31
otlp: fix otlp service.name check when no resource attributes are defined

sandeepsukhani:
When no resource attributes are defined, check for `service.name` resource attributes fail with `nil pointer dereference`. This PR fixes the issue by checking the bool returned by the `Get` method on resource attributes.

**Checklist**
- [x] Tests updated


MichelHollands (approved):
LGTM

PR #11840, opened 2024-01-31
feat: Support usage trackers for received and discarded bytes.

jeschkies:
Currently, Loki exposes the `distributor_bytes_received_total` and `discarded_bytes_total` metrics. These only expose the tenant and the discarded reason in the labels. In some cases it is useful to be specific to to certain streams. E.g. one might want to know how many bytes from a certain region were received and dropped. This is where usage trackers come in.

They receive the added or discarded bytes of a stream and its labels. This can be used to build a custom tracker such as Mimir's https://github.com/grafana/mimir/pull/1188 which uses regex matchers and exports the active series as a Prometheus metric.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/distributor/distributor.go
```diff
@@ -347,14 +348,17 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
 					bytes += len(e.Line)
 				}
 				validation.DiscardedBytes.WithLabelValues(validation.InvalidLabels, tenantID).Add(float64(bytes))
+				for _, tracker := range validationContext.customTrackerConfig.MatchTrackers(lbs) {
+					validation.DiscardedBytesCustom.WithLabelValues(validation.RateLimited, tenantID, tracker).Add(float64(bytes))
+				}
```
cstyan, line 353: we could loop once and save the trackers that the labelset matched, and then loop over those and update the discarded (and if needed rate limited) metrics

cstyan (commented):
We should have some mechanism for cleaning these metrics up if they haven't been updated recently, in which case they're just excess cardinality.

Also, it looks like disitrbutors processing `Push` requests is roughly 10% of their CPU time, it would be good to know what kind of additional CPU cost we're incurring when we have some of these custom trackers for a tenant or two.

pkg/distributor/distributor.go
jeschkies: That's a good point.

cstyan: No, you just have the same loop iteration later commented out 
```
/* We would have to filter all streams again here.
		for _, tracker := range 
```

If you're worried about iterating over all the trackers multiple times we could save the matching trackers once and then just iterate over those when we detect that we need to update the `DiscardedBytesCustom` metrics.

jeschkies: Ah, I missed that comment thanks.

jeschkies: Ok, this got a little messy. We cannot report on a labels basis if the labels cannot be parsed. We do want to report on labels basis when we hit the rate limit.

pkg/ingester/limiter.go
```diff
@@ -139,6 +141,10 @@ func (l *Limiter) minNonZero(first, second int) int {
 	return first
 }
 
+func (l *Limiter) CustumTrackersConfig(userID string) *push.CustomTrackersConfig {
+	return l.limits.CustomTrackersConfig(userID)
+}
+
```
cstyan, line 147: misspelling of custom here

pkg/ingester/instance.go
```diff
@@ -282,21 +299,14 @@ func (i *instance) createStream(pushReqStream logproto.Stream, record *wal.Recor
 			bytes += len(e.Line)
 		}
 		validation.DiscardedBytes.WithLabelValues(validation.StreamLimit, i.instanceID).Add(float64(bytes))
+		if i.customStreamsTracker != nil {
+			for _, matchedLbs := range i.limiter.CustumTrackersConfig(i.instanceID).MatchTrackers(labels) {
```
cstyan, line 303: and here

pkg/distributor/distributor.go
```diff
@@ -684,30 +706,31 @@ func (d *Distributor) sendStreamsErr(ctx context.Context, ingester ring.Instance
 }
 
 type labelData struct {
+	ls     labels.Labels
 	labels string
 	hash   uint64
 }
```
cstyan, line 712: could we modify this struct to just have the `ls`? or do we call access the `labels` enough that `ls.String()` would be too expensive?

pkg/loghttp/push/push.go
```diff
@@ -206,12 +234,29 @@ func ParseLokiRequest(userID string, r *http.Request, tenantsRetention TenantsRe
...
+		// Init custom trackers
+		customTrackers := limits.CustomTrackersConfig(userID)
+		var lbs labels.Labels
+		if tenantsRetention != nil || len(customTrackers.source) > 0 {
+			lbs, err = syntax.ParseLabels(s.Labels)
 			if err != nil {
 				return nil, nil, fmt.Errorf("couldn't parse labels: %w", err)
 			}
+		}
+		trackedLabels := customTrackers.MatchTrackers(lbs)
```
vlad-diachenko, line 247: I would call it trackers
```suggestion
		trackers := customTrackers.MatchTrackers(lbs)
```

jeschkies:
@vlad-diachenko and @cstyan I've changed the code to just inject a usage tracker. This is quite generic and allows us to encapsulate the labels filtering in summing in a few simple methods.

cstyan (approved):
LGTM but I still think the additional CPU usage is worth measuring

> Also, it looks like disitrbutors processing Push requests is roughly 10% of their CPU time, it would be good to know what kind of additional CPU cost we're incurring when we have some of these custom trackers for a tenant or two.

vlad-diachenko (commented):
Looks great 💎

PR #11841, opened 2024-01-31
(chore) Bloom shipper: Extend `Interval` struct with utility functions

chaudum:

~~Even though it is mainly used by the bloom shipper and bloom gateway, it should reside next to the FingerprintBounds, which has similar characteristics and utility functions.~~

Update:

I reverted moving the struct, but kept the utility functions.

### DIFF
--- a/pkg/bloomgateway/processor.go
+++ b/pkg/bloomgateway/processor.go
@@ -33,10 +33,7 @@ type processor struct {
 
 func (p *processor) run(ctx context.Context, tasks []Task) error {
 	for ts, tasks := range group(tasks, func(t Task) model.Time { return t.day }) {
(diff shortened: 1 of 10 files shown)


owen-d (commented):
I'm not sure Interval makes sense to put in the v1 pkg -- there's nothing in there that specifically uses it? Not really against it either, but you'll want to actually move the struct definition so it compiles :)

chaudum:
> I'm not sure Interval makes sense to put in the v1 pkg -- there's nothing in there that specifically uses it? Not really against it either, but you'll want to actually move the struct definition so it compiles :)

I don't have strong opinions on where the struct resides. I'll move it back to the shipper then.

owen-d (approved):
LGTM

PR #11844, opened 2024-01-31
quantile sharding: switch to ddsketch sparse constructor and remove quantile vector pooling in Join function

cstyan:
This should reduce our memory usage for our quantile sketches by about half, at least in our current situation we'll OOM after ~160-170 minutes instead of ~90 :cry:

### DIFF
--- a/pkg/logql/quantile_over_time_sketch.go
+++ b/pkg/logql/quantile_over_time_sketch.go
@@ -284,7 +284,8 @@ func JoinQuantileSketchVector(next bool, r StepResult, stepEvaluator StepEvaluat
 	}
 
 	// The result is released to the pool when the matrix is serialized.
-	result := quantileVectorPool.Get(stepCount)
+	//result := quantileVectorPool.Get(stepCount)
+	result := ProbabilisticQuantileMatrix{}
 
 	for next {
 		result = append(result, vec)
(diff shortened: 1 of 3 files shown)


jeschkies (approved):

pkg/logql/quantile_over_time_sketch.go
```diff
@@ -242,10 +240,14 @@ type quantileSketchBatchRangeVectorIterator struct {
...
-	if r.at == nil {
-		r.at = make([]ProbabilisticQuantileSample, 0, len(r.window))
-	}
+	/*
+		if r.at == nil {
+			r.at = make([]ProbabilisticQuantileSample, 0, len(r.window))
+		}
+		clear(r.at)
+	*/
 	r.at = r.at[:0]
```
cstyan, line 249: do we need both this line and the `make` line below?

pkg/logql/quantile_over_time_sketch.go
```diff
@@ -242,10 +240,14 @@ type quantileSketchBatchRangeVectorIterator struct {
...
+	r.at = make([]ProbabilisticQuantileSample, 0, len(r.window))
```
cstyan, line 250: I think if we're not reusing `r.at` then reassigning it via this line should be enough to allow the previous underlying array to be gc'd right?

jeschkies (approved):

PR #11848, opened 2024-02-01
Blooms/bounds integration

owen-d:
Replaces `Ref`'s min/max fingerprint fields with `v1.FingerprintBounds` and improves a bunch of checks that use it. Builds on top of https://github.com/grafana/loki/pull/11847


pkg/storage/stores/shipper/bloomshipper/store.go
```diff
@@ -54,10 +55,20 @@ func (b *bloomStoreEntry) ResolveMetas(ctx context.Context, params MetaSearchPar
 			if err != nil {
 				return nil, nil, err
 			}
-			if metaRef.MaxFingerprint < uint64(params.Keyspace.Min) || uint64(params.Keyspace.Max) < metaRef.MinFingerprint ||
+
+			// LIST calls return keys in lexicographic order.
+			// Since fingerprints are the first part of the path,
+			// we can stop iterating once we find an item greater
+			// than the keyspace we're looking for
```
chaudum, line 62: Good call!

pkg/storage/bloom/v1/bounds.go
```diff
@@ -33,8 +64,16 @@ func (b FingerprintBounds) Hash(h hash.Hash32) error {
 	return errors.Wrap(err, "writing OwnershipRange")
 }
 
+// Addr returns the string representation of the fingerprint bounds for use in
+// content addressable storage.
+// TODO(owen-d): incorporate this into the schema so we can change it,
+// similar to `{,Parse}ExternalKey`
+func (b FingerprintBounds) Addr() string {
+	return fmt.Sprintf("%x-%x", uint64(b.Min), uint64(b.Max))
```
chaudum, line 72: This will change the string representation from a fixed length string for Min/Max to a variable length, right?

Is this what we want? When putting blocks to storage, a fixed length would make them sortable by Min fingerprint.

chaudum: IMO `fmt.Sprintf("%016x-%016x", uint64(b.Min), uint64(b.Max))` is what we want.

pkg/storage/bloom/v1/bounds.go
```diff
@@ -17,6 +20,34 @@ const (
 	After
 )
 
+func ParseFingerprint(s string) (model.Fingerprint, error) {
+	fp, err := strconv.ParseUint(s, 16, 64)
+	if err != nil {
+		return 0, fmt.Errorf("error parsing fingerprint %s : %w", s, err)
+	}
+	return model.Fingerprint(fp), nil
+}
```
chaudum, line 29: The package `github.com/prometheus/common/model` would also provide this functionality.

chaudum (commented):
Left a question regarding the fingerprint string representation.

pkg/storage/bloom/v1/bounds.go
```diff
@@ -33,8 +64,16 @@ func (b FingerprintBounds) Hash(h hash.Hash32) error {
 	return errors.Wrap(err, "writing OwnershipRange")
 }
 
+// Addr returns the string representation of the fingerprint bounds for use in
+// content addressable storage.
+// TODO(owen-d): incorporate this into the schema so we can change it,
+// similar to `{,Parse}ExternalKey`
+func (b FingerprintBounds) Addr() string {
```
salvacorts, line 71: I think naming this method `String` is more idiomatic plus allows you to do formating with %s right away.
https://goplay.tools/snippet/lmt7fpPGa7R

salvacorts (approved):

pkg/storage/bloom/v1/bounds.go
owen-d: oof, good point. Will fix

PR #11852, opened 2024-02-01
feat: update logcli so it tries to load the latest version of the schemaconfig

MichelHollands:

Cloud Logs exporter will write new versions of the schema config in separate files now. This means it will work with object stores that have a policy of write once (ie no updates). Logcli is changed in this PR to take that into account.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


DylanGuedes (approved):
lgtm, just 2 nits

pkg/logcli/query/query.go
```diff
@@ -395,6 +395,41 @@ func maxTime(t1, t2 time.Time) time.Time {
...
 
+func getLatestConfig(client chunk.ObjectClient, orgID string) (*config.SchemaConfig, error) {
+	searchFor := fmt.Sprintf("%s.yaml", schemaConfigFilename) // schemaconfig.yaml for backwards compatibility
+	loadedSchema, err := LoadSchemaUsingObjectClient(client, searchFor)
+	if err == nil {
+		return loadedSchema, nil
+	}
+	if err != errNotExists && err != nil {
+		return nil, err
+	}
```
vlad-diachenko, line 406: we need to move this fallback logic to the end of the function. 

let's say if we do not have any files with the naming convention `schemaconfig-tenant.yaml` or `schemaconfig-tenant-N.yaml`, then we fallback and load `schemaconfig.yaml`.

with the current implementation, it will always load `schemaconfig.yaml` even if newer files exist.

vlad-diachenko (changes requested):

pkg/logcli/query/query.go
MichelHollands: Good catch. This has been moved to the back. Tests have been added as well to check the cases of 1) a single legacy schemaconfig and 2) both a new style one and a legacy one.

Tested it with running logcli against a GCP bucket with bucket retention policy set.

vlad-diachenko (approved):

PR #11855, opened 2024-02-02
Blooms/align compactor to shipper

owen-d:
Does some alignment work between the `bloomcompactor` and the `bloomshipper` pkgs. Notably:

* Uses `bloomshipper.BlockRef` everywhere (removes old bloomshipper struct
* Integrates `v1.FingerprintBounds` in `Ref` struct
* `Location` interface to distinguish local paths vs paths in object storage for certain types (`{Meta,Bloom}Ref`s)
* Introduces `KeyResolver` interface to generate locations from these structs
* Integrates `KeyResolver` into our bloom store. In the future, this will allow us to change key structures across schema boundaries when we want to change|improve them.
* Removes `BlockPath` from `BlockRef` in favor of the new resolving functionality. This is also beneficial because it lets us _calculate_ locations from the pure Ref objects, rather than tie some arbitrary state to them which can change (or not be populated by accident).

### DIFF
--- a/pkg/storage/stores/shipper/bloomshipper/resolver.go
+++ b/pkg/storage/stores/shipper/bloomshipper/resolver.go
@@ -1,0 +1,121 @@
+package bloomshipper
+
+import (
+	"fmt"
+	"path"
+	"path/filepath"
+)
+
+const (
+	BloomPrefix  = "bloom"
+	MetasPrefix  = "metas"
+	BlocksPrefix = "blocks"
+)
+
+// KeyResolver is an interface for resolving keys to locations.
+// This is used to determine where items are stored in object storage _and_ on disk.
+// Using an interface allows us to abstract away platform specifics
+// (e.g. OS path-specifics, object storage difference, etc)
+// TODO(owen-d): implement resolvers that schema-aware, allowing us to change
+// the locations of data across schema boundaries (for instance to upgrade|improve).
+type KeyResolver interface {
+	Meta(MetaRef) Location
+	Block(BlockRef) Location
+}
+
+type defaultKeyResolver struct{}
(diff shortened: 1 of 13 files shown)


pkg/storage/stores/shipper/bloomshipper/resolver.go
```diff
@@ -0,0 +1,121 @@
...
+	BloomPrefix  = "bloom"
+	MetasPrefix  = "metas"
+	BlocksPrefix = "blocks"
+)
+
+// KeyResolver is an interface for resolving keys to locations.
+// This is used to determine where items are stored in object storage _and_ on disk.
+// Using an interface allows us to abstract away platform specifics
+// (e.g. OS path-specifics, object storage difference, etc)
+// TODO(owen-d): implement resolvers that schema-aware, allowing us to change
```
chaudum, line 19:
```suggestion
// TODO(owen-d): implement resolvers that are schema-aware, allowing us to change
```

chaudum (approved):
LGTM

pkg/storage/stores/shipper/bloomshipper/resolver.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (p PrefixedResolver) Block(ref BlockRef) Location {
+	return locations{
+		key(p.prefix),
+		p.KeyResolver.Block(ref),
+	}
+}
+
+type Location interface {
+	Addr() string      // object storage location
```
salvacorts, line 76: nit: what about `RemotePath` instead? I think it's clearer and differentiable from `LocalPath`

salvacorts (approved):

PR #11858, opened 2024-02-02
Update docs to support OTEL GA

JStickler:
Updating the docs in advance of OTEL release to remove "experimental" notes.

Also restructured release notes so that most recent content is first.


JordanRushing (approved):
LGTM

PR #11872, opened 2024-02-05
[Blooms] Stateless archiving

owen-d:
Moves the archival code to an interface based approach, removing the explicit fs dependency. We can now implement with in-memory solutions. Also refactors to use a tar-file iterator and marks tests as parallelizable.

### DIFF
--- a/pkg/storage/bloom/v1/reader.go
+++ b/pkg/storage/bloom/v1/reader.go
@@ -12,6 +12,7 @@ import (
 type BlockReader interface {
 	Index() (io.ReadSeeker, error)
 	Blooms() (io.ReadSeeker, error)
+	TarEntries() (Iterator[TarEntry], error)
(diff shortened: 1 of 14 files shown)


pkg/storage/bloom/v1/reader.go
```diff
@@ -31,6 +32,33 @@ func (r *ByteReader) Blooms() (io.ReadSeeker, error) {
 	return bytes.NewReader(r.blooms.Bytes()), nil
 }
 
+func (r *ByteReader) TarEntries() (Iterator[TarEntry], error) {
```
chaudum, line 35: I'm not sure why this needs to return an `Interator` and not just a slice, because you don't build the `TarEntry`s lazily anyway.

chaudum (approved):
lgtm

PR #11875, opened 2024-02-06
Use correct header name in docs for query fairness in scheduler

chaudum:

The rest of the doc mentions the correct header name.


ashwanthgoli (approved):
lgtm

PR #11876, opened 2024-02-06
(chore) BloomStore: Clean up `FetchBlocks()`

chaudum:

* Removes `FetchBlocksWithQueue` and replace it with `FetchBlocks`
* Operate on single `BlockRef` when loading a block directory from cache/fs/storage.

### DIFF
--- a/pkg/storage/stores/shipper/bloomshipper/fetcher.go
+++ b/pkg/storage/stores/shipper/bloomshipper/fetcher.go
@@ -125,7 +125,12 @@ func (f *Fetcher) writeBackMetas(ctx context.Context, metas []Meta) error {
 	return f.metasCache.Store(ctx, keys, data)
 }
 
-func (f *Fetcher) FetchBlocksWithQueue(ctx context.Context, refs []BlockRef) ([]BlockDirectory, error) {
+// FetchBlocks returns a list of block directories
+// It resolves them from three locations:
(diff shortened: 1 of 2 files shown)


salvacorts (approved):

pkg/storage/stores/shipper/bloomshipper/fetcher.go
```diff
@@ -125,120 +124,96 @@ func (f *Fetcher) writeBackMetas(ctx context.Context, metas []Meta) error {
...
-// It resolves them from three locations:
+// fetchBlock resolves a block from three locations:
 //  1. from cache
 //  2. from file system
 //  3. from remote storage
-func (f *Fetcher) FetchBlocks(ctx context.Context, refs []BlockRef) ([]BlockDirectory, error) {
-	if ctx.Err() != nil {
-		return nil, errors.Wrap(ctx.Err(), "fetch Blocks")
-	}
+func (f *Fetcher) fetchBlock(ctx context.Context, ref BlockRef) (BlockDirectory, error) {
```
owen-d, line 179: nit for later: we should probably implement these GETs over multiple blocks internally to allow us to use the `memcache.GetMulti` functionality to resolve multiple keys from a single request rather than 1:1

owen-d (approved):

PR #11878, opened 2024-02-06
Limit bloom block size

salvacorts:
This PR limits the size of the blocks created by the compactor.

The block builder keeps adding series' blooms to a block until the limit is exceeded, meaning that blocks may grow beyond the configured maximum. This is needed so we avoid having tiny blocks which had space for small blooms but later a bigger blooms didn't fit.

Blocks are built lazily: the generator returns an iterator that builds one block at a time.

The maximum size is currently set to 50 MBs. We will make this configurable on a followup PR.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


chaudum (commented):
lgtm

owen-d (approved):
Going to merge in favor of expediency, but there are a few bits worth mentioning:

* Re-adding the block inputs is going to introduce a lot of performance inefficiencies unless we make sure the `MergeBuilder` can take advantage of `Seek`: https://github.com/grafana/loki/blob/main/pkg/storage/bloom/v1/builder.go#L547
* nit: It feels a bit weird to create a `LazyBlockBuilderIterator` struct if we're not going to independently use or test it and it'd be easier to inline it's logic as a for loop.

pkg/storage/bloom/v1/util.go
```diff
@@ -168,6 +168,10 @@ func (it *SliceIter[T]) At() T {
 	return it.xs[it.cur]
 }
 
+func (it *SliceIter[T]) Reset() {
+	it.cur = -1
```
chaudum, line 172: This `Reset()` has different semantics than the `Reset()` of the `BlockQuerier` ([here](https://github.com/grafana/loki/pull/11878/files#diff-ad299f9ea738f6f6d8e1a3b9fcc651506b3a241dbfbd95a60ab9d96b5b8b4d4aR139))

In the block querier, you can call `At()` after `Reset()` and you'll get the first item of the iterator. Here, you'll need to call `Next()` first, to do the same.

salvacorts: You are right. I'm removing this one tho since it's a leftover from a previous impl that I was trying out

pkg/storage/bloom/v1/builder.go
```diff
@@ -545,8 +549,17 @@ func (mb *MergeBuilder) Build(builder *BlockBuilder) (uint32, error) {
 		nextInBlocks *SeriesWithBloom
 	)
 
+	// reset all the blocks to the start
+	blocksAsPeekIters := make([]PeekingIterator[*SeriesWithBloom], 0, len(mb.blocks))
```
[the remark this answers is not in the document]
salvacorts: Moved here
https://github.com/grafana/loki/blob/1748c29af710c310e24298be8e9398d172f2851d/pkg/bloomcompactor/spec.go#L210-L216

> Do you really need to make a copy of the slice?

Unfortunately yes. We need to pass `[]PeekingIterator` instead of []v1.PeekingResetIter to the builder. I expected Go to infer that they are compatible types but... no :(

pkg/storage/bloom/v1/util.go
salvacorts: Well, I'm actually leaving it, it's useful to test `PeekResetIterator`

salvacorts: > In the block querier, you can call At() after Reset() and you'll get the first item of the iterator

Revisiting this, it's not correct. You are expected to call `Next` after `Reset` (or `Seek(0)`). Here's an test for that:
https://github.com/grafana/loki/blob/1748c29af710c310e24298be8e9398d172f2851d/pkg/storage/bloom/v1/builder_test.go#L276-L283

Anyway, in my last commit I removed this as it's no longer needed.

pkg/storage/bloom/v1/builder.go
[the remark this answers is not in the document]
chaudum: Hm interesting. Can we cast it?

pkg/storage/bloom/v1/builder_test.go
```diff
@@ -104,20 +129,22 @@ func TestBlockBuilderRoundTrip(t *testing.T) {
...
-				require.Equal(t, true, querier.Next(), "on iteration %d", j)
-				got := querier.At()
-				require.Equal(t, halfData[j].Series, got.Series)
-				for _, key := range halfKeys[j] {
-					require.True(t, got.Bloom.Test(key))
+			if !iterHasPendingData {
+				i := numSeries / 2
+				halfData := data[i:]
+				halfKeys := keys[i:]
+				require.Nil(t, querier.Seek(halfData[0].Series.Fingerprint))
```
chaudum, line 136: nit:

```suggestion
				require.NoError(t, querier.Seek(halfData[0].Series.Fingerprint))
```

chaudum (approved):

pkg/storage/bloom/v1/builder_test.go
salvacorts: Done, thank you!

PR #11882, opened 2024-02-06
fix: promtail; clean up metrics generated from logs after a config reload.

ptodev:
It is possible to generate metrics from log lines in Promtail. If the config file is reloaded, those metrics are currently not cleaned up. This causes a few issues:
* If a metric is not necessary anymore, it is still visible on the `/metrics` endpoint.
* A metric with the same name might have a different meaning after the config file reload, so it'd make sense to clean these up.

This will help us fix [a bug](https://github.com/grafana/agent/issues/2754) reported in the Grafana Agent.

I suppose no changelog entry is required because PromQL is already designed to handle counter resets? But if you think it's appropriate, I could add a changelog entry?

I tested this change locally using a config file like this:

<details>
  <summary>Promtail config</summary>

```yaml
server:
  http_listen_port: 9080
  grpc_listen_port: 0
  enable_runtime_reload: true
positions:
  filename: /Users/paulintodev/Desktop/log_metrics_bug/tmp_promtail/positions
clients:
  - url: "https://logs-prod-008.grafana.net/loki/api/v1/push"
    basic_auth:
      username: ""
      password: ""
scrape_configs:
- job_name: log files
  static_configs:
  - labels:
      __path__: /Users/paulintodev/Desktop/log_metrics_bug/*.log
  pipeline_stages:
    - metrics:
        log_lines_total:
          type: Counter
          description: "total number of log lines"
          prefix: my_promtail_custom_
          max_idle_duration: 24h
          config:
            match_all: true
            action: inc
        log_lines_total2:
          type: Counter
          description: "total number of log lines"
          prefix: my_promtail_custom2_
          max_idle_duration: 24h
          config:
            match_all: true
            action: inc
```

</details>

Then I'd do a `curl localhost:9080/reload` and check `http://localhost:9080/metrics`. 

I didn't add a whole lot of unit tests because TBH we should probably focus on adding tests in the Agent, given that we're sunsetting Promtail over time.

### DIFF
--- a/clients/pkg/logentry/stages/metrics.go
+++ b/clients/pkg/logentry/stages/metrics.go
@@ -128,11 +128,11 @@ func newMetricStage(logger log.Logger, config interface{}, registry prometheus.R
 			metrics[name] = collector
 		}
 	}
-	return toStage(&metricStage{
+	return &metricStage{
 		logger:  logger,
 		cfg:     *cfgs,
 		metrics: metrics,
-	}), nil
+	}, nil
 }
 
 // metricStage creates and updates prometheus metrics based on extracted pipeline data
@@ -140,6 +140,19 @@ type metricStage struct {
 	logger  log.Logger
 	cfg     MetricsConfig
 	metrics map[string]prometheus.Collector
+}
+
+func (m *metricStage) Run(in chan Entry) chan Entry {
+	out := make(chan Entry)
+	go func() {
+		defer close(out)
+
+		for e := range in {
+			m.Process(e.Labels, e.Extracted, &e.Timestamp, &e.Line)
+			out <- e
+		}
+	}()
+	return out
 }
 
 // Process implements Stage
@@ -176,6 +189,20 @@ func (m *metricStage) Process(labels model.LabelSet, extracted map[string]interf
 // Name implements Stage
 func (m *metricStage) Name() string {
 	return StageTypeMetric
+}
+
+// Cleanup implements Stage.
+func (m *metricStage) Cleanup() {
+	for _, collector := range m.metrics {
+		switch vec := collector.(type) {
+		case *metric.Counters:
+			vec.DeleteAll()
+		case *metric.Gauges:
+			vec.DeleteAll()
+		case *metric.Histograms:
+			vec.DeleteAll()
+		}
+	}
 }
 
 // recordCounter will update a counter metric
--- a/clients/pkg/logentry/stages/metrics_test.go
+++ b/clients/pkg/logentry/stages/metrics_test.go
@@ -127,6 +127,13 @@ func TestMetricsPipeline(t *testing.T) {
 		strings.NewReader(expectedMetrics)); err != nil {
 		t.Fatalf("mismatch metrics: %v", err)
 	}
+
+	pl.Cleanup()
+
+	if err := testutil.GatherAndCompare(registry,
+		strings.NewReader("")); err != nil {
+		t.Fatalf("mismatch metrics: %v", err)
+	}
 }
 
 func TestNegativeGauge(t *testing.T) {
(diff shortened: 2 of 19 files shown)


clients/pkg/logentry/stages/metrics_test.go
```diff
@@ -127,6 +127,13 @@ func TestMetricsPipeline(t *testing.T) {
 		strings.NewReader(expectedMetrics)); err != nil {
 		t.Fatalf("mismatch metrics: %v", err)
 	}
+
+	pl.Cleanup()
+
+	if err := testutil.GatherAndCompare(registry,
+		strings.NewReader("")); err != nil {
+		t.Fatalf("mismatch metrics: %v", err)
+	}
```
cstyan, line 136: shouldn't we specifically check for the absence of the metric?

cstyan (commented):
@ptodev need to rebase/merge in master to get some CI updates

clients/pkg/logentry/stages/metrics.go
```diff
@@ -142,6 +142,19 @@ type metricStage struct {
...
+	go func() {
+		defer close(out)
+
+		for e := range in {
+			m.Process(e.Labels, e.Extracted, &e.Timestamp, &e.Line)
+			out <- e
+		}
+	}()
+	return out
+}
```
cstyan, line 156: why was this change necessary? just to mimic what `toStage` was giving us before?

clients/pkg/logentry/stages/metrics_test.go
ptodev: After `pl.Cleanup()`, no metrics should be visible. I don't see an advantage in checking individual metrics when there shouldn't be any of them.

clients/pkg/logentry/stages/metrics.go
ptodev: Yes. The metrics stage needs a cleanup function, and `toStage` doesn't add one. Maybe we could avoid the repetition by having another function called `toStageWithCleanup`?

cstyan: I think this way is fine. Even with `toStageWithCleanup` we would still require the addition of `Cleanup` noop functions on all stages except the metric stage.

clients/pkg/logentry/stages/metrics_test.go
cstyan: what if there's originally two metrics stages, and after we reload there should be only one? we should still check for it's existence and correct value, I think right now we're deleting metrics for stages that will still exist after the reload which would not be correct

ptodev: I'd argue that deleting all the metrics is the right thing to do. The fact that a metric exists doesn't mean that it's used in the same way in the new config. It's possible that the user changed the meaning of each metric, but retained the metric names. In that case it would make sense to start with fresh metrics.

In the future we could make the code smart and reload only the stages which changed, but I think that's a separate issue. To do that, we could avoid calling the `Cleanup` method for stages which are completely unchanged as a whole.

cstyan: we should document this clearly in the metrics config section https://github.com/grafana/loki/blob/a00f1f1b0b8f536f2cdac2f8857eb40c716aa696/docs/sources/send-data/promtail/configuration.md?plain=1#L678 and maybe as a note here as well

cstyan (commented):
@ptodev sorry for the delay, I think we just need some docs, also have you guys been using this code downstream in Agent/Alloy already or would merging this be the first usage of this code?

cstyan (approved):
I'm going to approve/merge this with a note that if we get reports of issues I'll revert this change. Hopefully we here from the Alloy team after a short period of time whether they have confirmed usages that are working well or not.

PR #11883, opened 2024-02-06
feat(lambda-promtail): Add possibility to set additional HTTP headers for promtail-client requests

trc-ikeskin:

This PR adds the ability to set additional, custom HTTP headers for requests made by the promtail client in the` lambda-promtail` tool. Additional headers can be passed to the AWS Lambda function via the environment variable `EXTRA_HTTP_HEADERS` in the form `<HTTP header 1 key>,<HTTP header 1 value>,<HTTP header 2 key>,<HTTP header 2 value>, ...`.

This functionality is needed when for example requests pass a AWS Web Application Firewall that expects a certain custom header to be set on every request passed to a load balancer. If this load balancer must be traversed to access the Loki gateway push endpoint, then there is currently no easy way to achieve this.

Simple header key validation was implemented using regex, but there is currently no validation for header values.


trc-ikeskin:
@cstyan would you be able to review this PR?

tools/lambda-promtail/lambda-promtail/main.go
```diff
@@ -25,19 +26,22 @@ const (
...
+	username, password, extraLabelsRaw, dropLabelsRaw, tenantID, bearerToken, extraHeadersRaw string
+	keepStream                                                                                bool
+	batchSize                                                                                 int
+	s3Clients                                                                                 map[string]*s3.Client
+	extraLabels                                                                               model.LabelSet
+	dropLabels                                                                                []model.LabelName
+	skipTlsVerify                                                                             bool
+	printLogLine                                                                              bool
+	extraHeaders                                                                              map[string]string
+	httpHeaderKeyRegex                                                                        = regexp.MustCompile("([A-Za-z0-9-])+")
```
cstyan, line 44: I think the set of characters that are valid for HTTP header keys is much more than this, is it not?

tools/lambda-promtail/lambda-promtail/main.go
```diff
@@ -190,6 +200,29 @@ func checkEventType(ev map[string]interface{}) (interface{}, error) {
...
+	}
+
+	for i := 0; i < len(extraHeadersSplit); i += 2 {
+		if !httpHeaderKeyRegex.MatchString(extraHeadersSplit[i]) {
+			return nil, fmt.Errorf("HTTP header key is invalid: %s", extraHeadersSplit[i])
+		}
+		extractedHeaders[extraHeadersSplit[i]] = extraHeadersSplit[i+1]
+	}
+
+	fmt.Println("extra HTTP headers:", extractedHeaders)
```
cstyan, line 222: we should follow the same pattern as the other existing parser functions and not log here

cstyan (commented):
I'm not sure the regex for the header values is correct, or necessary?

We should also have some tests, see the other tests in `main_test.go` for examples, such as `TestLambdaPromtail_ExtraLabels*`

tools/lambda-promtail/lambda-promtail/main.go
trc-ikeskin: You seem to be right about this from an RFC perspective, since AFAICS there is no real specification for the format of a valid HTTP header field name. Nevertheless some services still impose such restrictions - especially the ones stated by AWS ELB led me to idea to implement this:

- https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#:~:text=default%20is%20defensive.-,routing.http.drop_invalid_header_fields.enabled,-Indicates%20whether%20HTTP
- https://developers.cloudflare.com/rules/transform/request-header-modification/reference/header-format/
- https://pkg.go.dev/golang.org/x/net/http/httpguts#ValidHeaderFieldName

tools/lambda-promtail/lambda-promtail/main.go
trc-ikeskin: You are right, this seems to be an artifact of my tests. I will remove that line.

tools/lambda-promtail/lambda-promtail/main_test.go
```diff
@@ -64,3 +64,33 @@ func TestLambdaPromtail_TestDropLabels(t *testing.T) {
...
+	extraHeaders, err := parseExtraHeaders("A,a,B,b,C,c,D")
+	require.Nil(t, extraHeaders)
+	require.Errorf(t, err, invalidExtraHeadersError)
+}
+
+func TestLambdaPromtail_TestParseHeadersNoneProvided(t *testing.T) {
+	extraLabels, err := parseExtraHeaders("")
+	require.Len(t, extraLabels, 0)
+	require.Nil(t, err)
+}
```
cstyan, line 96: lets simplify a bit and make this a table driven test

tools/lambda-promtail/lambda-promtail/main.go
```diff
@@ -25,19 +26,22 @@ const (
 
 	maxErrMsgLen = 1024
 
-	invalidExtraLabelsError = "invalid value for environment variable EXTRA_LABELS. Expected a comma separated list with an even number of entries. "
+	invalidExtraLabelsError  = "invalid value for environment variable EXTRA_LABELS. Expected a comma separated list with an even number of entries. "
+	invalidExtraHeadersError = "invalid value for environment variable EXTRA_HTTP_HEADERS. Expected a comma separated list with an even number of entries."
```
cstyan, line 30: depending on what we end up for the regex we're choosing to enforce, we should include the regex (or link to where the regex is defined, in the case of either of the 3 links you gave)

trc-ikeskin: I split the error messages and included the actual regex in the corresponding error message.

tools/lambda-promtail/lambda-promtail/main_test.go
trc-ikeskin: I am not really sure how a table driven test simplifies my test, but I am also not too experience with this kind of testing. Maybe you could give me a hint and what you had in mind here?

cstyan: rather than declaring 4 new tests, one for each of these success or failure scenarios, we would have one test with an array of input/expected output, see here https://go.dev/wiki/TableDrivenTests

tools/lambda-promtail/lambda-promtail/main.go
```diff
@@ -25,19 +26,23 @@ const (
 
 	maxErrMsgLen = 1024
 
-	invalidExtraLabelsError = "invalid value for environment variable EXTRA_LABELS. Expected a comma separated list with an even number of entries. "
+	invalidExtraLabelsError  = "invalid value for environment variable EXTRA_LABELS. Expected a comma separated list with an even number of entries. "
+	invalidEvenExtraHeadersError = "invalid value for environment variable EXTRA_HTTP_HEADERS. Expected a comma separated list with an even number of entries."
+	invalidRegexExtraHeadersError = "invalid value for environment variable EXTRA_HTTP_HEADERS. Header key must conform with regex /[-A-Za-z0-9]+/ "
```
cstyan, line 31: can we use `httpHeaderKeyRegex.String()` here rather than hardcoding the regex into this error string, just in case the string for the regex changes in the future

tools/lambda-promtail/lambda-promtail/promtail.go
```diff
@@ -198,6 +198,12 @@ func (c *promtailClient) send(ctx context.Context, buf []byte) (int, error) {
 		req.Header.Set("Authorization", "Bearer "+bearerToken)
 	}
 
+	if len(extraHeaders) > 0 {
+		for key, value := range extraHeaders {
+			req.Header.Set(key, value)
+		}
+	}
+
```
cstyan, line 206: I'm am wondering if, to avoid any potential unexpected issues, we should include some validation here. For example, not allowing overwriting of the headers we set earlier in this function, and logging if we overwrite an existing header value (such as someone not noticing that in their configured extra headers they're setting `SOME-EXTRA-HEADER=a, ... , SOME-EXTRA-HEADER=b`.

cstyan:
@trc-ikeskin IMO there's still some minor changes we should make here

tools/lambda-promtail/lambda-promtail/main.go
trc-ikeskin: I removed the regex from the constant string and added `httpHeaderKeyRegex.String()` to the format string in the fmt.Errorf() call here: https://github.com/grafana/loki/pull/11883/commits/8d90c1003fed78eea21b12771125d2c0a3ec302a

tools/lambda-promtail/lambda-promtail/promtail.go
trc-ikeskin: I implemented a simple check to see if a key has already been set and added a warning for the user to check its env vars: https://github.com/grafana/loki/pull/11883/commits/6a8a19d68626e976e0a0bfe56b1b9825d4dc928d

Not sure about the logging part since I could not really find a clear way the logging is implemented in the code. Maybe you could have a closer look at that.

cstyan: @trc-ikeskin thanks! this is exactly what I had in mind. Logging in this context just meant fmt.Println or using the logger which you've done.

I still think we should guard against people overwriting the headers that are set earlier in this same function, such as `Authorization` or `X-Scope-OrgID`. We can do that here or in a follow up PR. Up to you.

tools/lambda-promtail/lambda-promtail/main.go
```diff
@@ -25,19 +26,23 @@ const (
 
 	maxErrMsgLen = 1024
 
-	invalidExtraLabelsError = "invalid value for environment variable EXTRA_LABELS. Expected a comma separated list with an even number of entries. "
+	invalidExtraLabelsError       = "invalid value for environment variable EXTRA_LABELS. Expected a comma separated list with an even number of entries. "
+	invalidEvenExtraHeadersError  = "invalid value for environment variable EXTRA_HTTP_HEADERS. Expected a comma separated list with an even number of entries."
```
cstyan, line 30: hey @trc-ikeskin either the name here or on line 89 of the tests needs to change, I think `invalidExtraHeadersError` is enough

cstyan:
@trc-ikeskin did you still want to proceed with this PR?

trc-ikeskin:
> @trc-ikeskin did you still want to proceed with this PR?

@cstyan I still want to continue eventually but I cannot find the time right now - sorry.

cstyan:
> > @trc-ikeskin did you still want to proceed with this PR?
> 
> @cstyan I still want to continue eventually but I cannot find the time right now - sorry.

:+1: no worries

JStickler:
@trc-ikeskin as of https://github.com/grafana/loki/pull/18531, the lambda-promtail code has moved to its own repo as another team is taking ownership of it. Could you re-open your PR against the [new lambda promtail repo](https://github.com/grafana/lambda-promtail) please?

PR #11894, opened 2024-02-08
[Blooms] Routing & wiring for bloom building

owen-d:
A few changes wiring up bloom building logic:
* A `Router` as the controlling mechanism for iterating tenants+periods, queueing work for parallel computation, determining which fingerprint ranges and tenants are owned by a replica, and waiting for it's completion
* Updates the `SimplerBloomController` to accept ownership ranges, periods, and tenants as arguments for use by the `Router`
* Builds a `BloomTSDBStore` struct around the `indexshipper/storage.Client` interface in order to iterate relevant TSDBs/etc


slim-bean (approved):
LGTM

pkg/bloomcompactor/router.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/prometheus/common/model"
+
+	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
+	"github.com/grafana/loki/pkg/storage/config"
+)
+
+type DayTable model.Time
+
+func (d DayTable) String() string {
+	return fmt.Sprintf("%d", d.ModelTime().Time().UnixNano()/int64(config.ObjectStorageIndexRequiredPeriod))
```
chaudum, line 24: At some point we need to add the prefix from the schema config.

chaudum: A string as table representation is quite confusing, because it may or may not contain the prefix.

pkg/bloomcompactor/tsdb.go
```diff
@@ -2,15 +2,106 @@ package bloomcompactor
...
+) (v1.CloseableIterator[*v1.Series], error) {
+	data, err := b.storage.GetUserFile(ctx, table, tenant, id.Name())
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get file")
+	}
+
+	buf, err := io.ReadAll(data)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read file")
+	}
```
chaudum, line 92: Do we want to load the full index into memory?

pkg/bloomcompactor/router.go
```diff
@@ -0,0 +1,224 @@
...
+	return newDayRangeIterator(r.minTable, r.maxTable), nil
+}
+
+func (r *router) run(ctx context.Context) error {
+	// run once at beginning
+	if err := r.runOne(ctx); err != nil {
+		return err
+	}
+
+	ticker := time.NewTicker(r.interval)
```
chaudum, line 95: ticker should be stopped on exit

chaudum (commented):
lgtm

pkg/bloomcompactor/tsdb.go
owen-d: I think this is easier (especially for now). Individual TSDBs can vary in size, but are commonly a couple hundred MB. We can optimize this if it becomes a problem, but it's easier to just load these into memory for now.

chaudum: Makes sense

PR #11896, opened 2024-02-08
Bloom gateway: Use multicontext for processing tasks

chaudum:

Multiple tasks that are multiplexed into a single one, may stem from different requests and therefore different request contexts.

Using a multi-context for the task processing ensures client side context cancellation is handled correctly within the processor.

### DIFF
--- a/pkg/bloomgateway/worker.go
+++ b/pkg/bloomgateway/worker.go
@@ -10,11 +10,8 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	"github.com/prometheus/common/model"
-	"golang.org/x/exp/slices"
 
 	"github.com/grafana/loki/pkg/queue"
-	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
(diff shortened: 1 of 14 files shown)


pkg/bloomgateway/worker.go
```diff
@@ -123,9 +124,11 @@ func (w *worker) running(_ context.Context) error {
 			level.Debug(w.logger).Log("msg", "dequeued task", "task", task.ID)
 			w.pending.Delete(task.ID)
 			tasks = append(tasks, task)
+			ctxs = append(ctxs, task.ctx)
 		}
 
-		err = p.run(taskCtx, tasks)
+		multiCtx := multicontext.WithContexts(ctxs...)
```
owen-d, line 130: Do we want all requests to cancel if one is? I think that's what's happening here

chaudum: Up for discussion. The other option is to use `taskCtx`, which is not related to any task and cannot be cancelled.

salvacorts: My expectation would be that all ongoing child operations for a given task are canceled if the ctx of the given task is canceled. IIUC, with multicontext we are doing the other way around: a task ctx canceled cancels the contexes of other tasks.

PR #11897, opened 2024-02-08
feat(metadata): introduce a separate split interval for recent query window

ashwanthgoli:

Metadata queries are split using a 24h interval by default. The same interval gets used in the cache key.
But it is not possible to extract a subset of labels/series from a cached extent, unlike samples they are not associated with a timestamp. To prevent short queries from returning the results of an entire `24h` extent, caching is disabled for the last 24h by default using `max_metadata_cache_freshness`.

But we have noticed in our cloud environments that most queries fall in the last 24h interval. This pr introduces the following changes to help cache the recent metadata query results:
- Use a smaller split interval for caching recent metadata queries.  Portion of the query within `recent_metadata_query_window` is split using a different interval of `split_recent_metadata_queries_by_interval`.
- Only use extents that are entirely within the requested metadata query range. This is done to avoid using extents that contain results from the outside of the requested range.

---
To use a shorter split interval for recent metadata queries, the following needs to be configured:
- `recent_metadata_query_window` to configure the window inside which the shorter split interval gets applied. Disabled by default.
- `split_recent_metadata_queries_by_interval` to configure the split interval for the portion of the query within the recent metadata query window. Defaults to `1h`. Recommended to be configured to a value smaller than `split_metadata_queries_by_interval`.
- reduce `max_metadata_cache_freshness` to control the cache freshness

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/querier/queryrange/splitters.go
```diff
@@ -91,42 +91,67 @@ func (s *defaultSplitter) split(execTime time.Time, tenantIDs []string, req quer
...
+		origEnd             = req.GetEnd().UTC()
+		start, end          = origStart, origEnd
+
+		reboundOrigQuery           bool
+		splitIntervalBeforeRebound time.Duration
 	)
 
-	start, end, needsIngesterSplits := ingesterQueryBounds(execTime, s.iqo, req)
+	switch req.(type) {
+	case *LokiSeriesRequest, *LabelRequest:
```
kavirajk, line 104: curious why aren't we considering `split_ingester_queries_by_interval` for series and labels queries?

kavirajk (commented):
Overall looks good 👍 nice job @ashwanthgoli 

Left few minor comments. Two other things

1. I see you have some tests to lock the behaviour of returning "correct" split either inside or outside recent metadata query window. But not enough tests to check the results of the split I think?. Asking because I'm curious say for smaller recent window (say 30m), if a query comes asking for `5m` of labels, how big returned data time range would be `1h`, `1w`? 
2. Like to see what others think of having "second" window to split metadata queries before merging it.

pkg/querier/queryrange/labels_cache.go
```diff
@@ -11,21 +11,42 @@ import (
...
+	recentMetadataQueryWindow := validation.MaxDurationOrZeroPerTenant(tenantIDs, limits.RecentMetadataQueryWindow)
+	recentMetadataQuerySplitInterval := validation.MaxDurationOrZeroPerTenant(tenantIDs, limits.RecentMetadataQuerySplitDuration)
+
+	// if either of the options are not configured, use the default metadata split interval
+	if recentMetadataQueryWindow == 0 || recentMetadataQuerySplitInterval == 0 {
+		return split
+	}
+
+	// if the query start is not before window start, it would be split using recentMetadataQuerySplitInterval
+	if windowStart := ref.Add(-recentMetadataQueryWindow); !start.Before(windowStart) {
```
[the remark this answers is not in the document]
ashwanthgoli: had to use `!before` because we also have to consider the case where `start` aligns with the `recentMetadataQueryWindow` start.

pkg/querier/queryrange/splitters.go
ashwanthgoli: i feel these features have orthogonal goals.

`split_ingester_queries_by_interval` is introduced to reduce the number of sub-queries we send to ingesters. currently metadata queries send atmost 1 subquery to ingesters with the default split interval.
 
while `split_recent_metadata_queries_by_interval` would increase the number of sub-queries, they should ideally deamplify over time as the results get cached.

 I am not sure if we'd use these two together for metadata queries and what the config would look like if we do so.

kavirajk: Make sense. can you record this somewhere as comment for the future reference?

slim-bean:
Haven't looked to closely at the implementation, but the idea makes good sense to me.

First thought, does this need to be configurable? What circumstances would we ever change these configs?

ashwanthgoli:
> First thought, does this need to be configurable? What circumstances would we ever change these configs?

mostly `recent_metadata_query_window` will be left at 24h unless the orig split interval is changed. Having this configurable also allows disabling recent metadata splits if it's not working as expected.

`split_recent_metadata_queries_by_interval` would be useful to create even smaller cache buckets, smaller the bucket the higher the chance of a hit. To start with 1h seems like a good default.

kavirajk (approved):
LGTM 👍

pkg/querier/queryrange/labels_cache.go
```diff
@@ -11,21 +11,42 @@ import (
...
 type cacheKeyLabels struct {
 	Limits
 	transformer UserIDTransformer
-	iqo         util.IngesterQueryOptions
+}
+
+// metadataSplitIntervalForTimeRange returns split interval for series and label requests.
+// If `recent_metadata_query_window` is configured and the query start interval is within this window,
+// it returns `split_recent_metadata_queries_by_interval`.
+// For other cases, the default split interval of `split_metadata_queries_by_interval` will be used.
```
dannykopping, line 25: I actually think your code is pretty simple and self-documenting; all this comment can do is go stale.

pkg/storage/chunk/cache/resultscache/cache.go
```diff
@@ -334,6 +336,25 @@ func (s ResultsCache) partition(req Request, extents []Extent) ([]Request, []Res
 			continue
 		}
 
+		if s.onlyUseEntireExtent && (start > extent.GetStart() || end < extent.GetEnd()) {
```
dannykopping, line 339: This is awesome 👏

dannykopping (approved):
Great work! Just one minor nit.

PR #11901, opened 2024-02-08
logs query & hashedquery separately to minimize stacktraces

owen-d:
Rather than call all our values together, this PR precomputes a few in `metrics.go` so any issues will result in smaller stack traces.


cyriltovena (approved):
Lgtm

PR #11905, opened 2024-02-09
quantile sharding; fix bug related to # of steps (instant query mem usage) and use CollapsingLowestDense store for ddsketch

cstyan:
**EDIT: I also found the bug that was causing us to OOM when executing instant queries, see the 2nd commit**

This should reduce the memory allocated by the vector iteration by at least 25% but as much as ~80%. The quantile vector iteration accounts for at least 15% of all in use space last time I tested (and during that time queriers we're OOM'ing as a result of quantile sharding). 

However, with the usage of this alternate store type we give up some amount of relative accuracy guarantees for _lower_ quantiles. From their readme:

```
We also provide implementations, returned by LogCollapsingLowestDenseDDSketch(relativeAccuracy, maxNumBins) and LogCollapsingHighestDenseDDSketch(relativeAccuracy, maxNumBins), where the q-quantile will be accurate up to the specified relative error for q that is not too small (or large). Concretely, the q-quantile will be accurate up to the specified relative error as long as it belongs to one of the m bins kept by the sketch. For instance, If the values are time in seconds, maxNumBins = 2048 covers a time range from 80 microseconds to 1 year.
```

Benchmark:
```
goos: linux
goarch: amd64
pkg: github.com/grafana/loki/pkg/logql
cpu: AMD Ryzen 9 5950X 16-Core Processor
                                                      │    base     │                new2                │
                                                      │   sec/op    │   sec/op     vs base               │
QuantileBatchRangeVectorIteratorAt/1-samples-32         153.4n ± 1%   169.8n ± 1%  +10.76% (p=0.002 n=6)
QuantileBatchRangeVectorIteratorAt/1000-samples-32      30.17µ ± 1%   10.63µ ± 1%  -64.76% (p=0.002 n=6)
QuantileBatchRangeVectorIteratorAt/10000-samples-32     281.1µ ± 1%   101.5µ ± 2%  -63.88% (p=0.002 n=6)
QuantileBatchRangeVectorIteratorAt/100000-samples-32    2.355m ± 2%   1.006m ± 2%  -57.26% (p=0.002 n=6)
QuantileBatchRangeVectorIteratorAt/1000000-samples-32   23.50m ± 2%   10.11m ± 1%  -56.99% (p=0.002 n=6)
geomean                                                 148.4µ        71.47µ       -51.84%

                                                      │     base     │                 new2                 │
                                                      │     B/op     │    B/op      vs base                 │
QuantileBatchRangeVectorIteratorAt/1-samples-32          80.00 ±  0%   80.00 ±  0%        ~ (p=1.000 n=6) ¹
QuantileBatchRangeVectorIteratorAt/1000-samples-32      157.00 ±  1%   80.00 ±  0%  -49.04% (p=0.002 n=6)
QuantileBatchRangeVectorIteratorAt/10000-samples-32     155.00 ±  3%   80.00 ±  1%  -48.39% (p=0.002 n=6)
QuantileBatchRangeVectorIteratorAt/100000-samples-32    109.00 ± 76%   83.00 ±  0%  -23.85% (p=0.002 n=6)
QuantileBatchRangeVectorIteratorAt/1000000-samples-32   1082.5 ±  4%   222.0 ± 11%  -79.49% (p=0.002 n=6)
geomean                                                  187.2         98.84        -47.19%
¹ all samples are equal

                                                      │    base    │                new2                 │
                                                      │ allocs/op  │ allocs/op   vs base                 │
QuantileBatchRangeVectorIteratorAt/1-samples-32         3.000 ± 0%   3.000 ± 0%        ~ (p=1.000 n=6) ¹
QuantileBatchRangeVectorIteratorAt/1000-samples-32      6.000 ± 0%   3.000 ± 0%  -50.00% (p=0.002 n=6)
QuantileBatchRangeVectorIteratorAt/10000-samples-32     6.000 ± 0%   3.000 ± 0%  -50.00% (p=0.002 n=6)
QuantileBatchRangeVectorIteratorAt/100000-samples-32    4.000 ± 0%   3.000 ± 0%  -25.00% (p=0.002 n=6)
QuantileBatchRangeVectorIteratorAt/1000000-samples-32   7.000 ± 0%   3.000 ± 0%  -57.14% (p=0.002 n=6)
geomean                                                 4.967        3.000       -39.60%
¹ all samples are equal
```

And profiling
2024-02-08-180659_2040x719_scrot
2024-02-08-180725_2042x650_scrot

### DIFF
--- a/pkg/logql/downstream_test.go
+++ b/pkg/logql/downstream_test.go
@@ -146,7 +146,7 @@ func TestMappingEquivalenceSketches(t *testing.T) {
 		regular := NewEngine(opts, q, NoLimits, log.NewNopLogger())
 		sharded := NewDownstreamEngine(opts, MockDownstreamer{regular}, NoLimits, log.NewNopLogger())
 
-		t.Run(tc.query, func(t *testing.T) {
+		t.Run(tc.query+"_range", func(t *testing.T) {
 			params, err := NewLiteralParams(
 				tc.query,
 				start,
@@ -178,45 +178,15 @@ func TestMappingEquivalenceSketches(t *testing.T) {
 
 			relativeError(t, res.Data.(promql.Matrix), shardedRes.Data.(promql.Matrix), tc.realtiveError)
 		})
-	}
-}
-
-func TestMappingEquivalenceSketches_Instant(t *testing.T) {
-	var (
-		shards   = 3
-		nStreams = 10_000
-		rounds   = 20
-		streams  = randomStreams(nStreams, rounds+1, shards, []string{"a", "b", "c", "d"}, true)
-		start    = time.Unix(0, int64(rounds+1))
-		end      = time.Unix(0, int64(rounds+1))
-		step     = time.Duration(0)
-		interval = time.Duration(0)
-		limit    = 100
-	)
-
-	for _, tc := range []struct {
-		query         string
-		realtiveError float64
-	}{
-		{`quantile_over_time(0.70, {a=~".+"} | logfmt | unwrap value [10m]) by (a)`, 0.03},
-		{`quantile_over_time(0.99, {a=~".+"} | logfmt | unwrap value [10m]) by (a)`, 0.02},
-	} {
-		q := NewMockQuerier(
-			shards,
-			streams,
-		)
-
-		opts := EngineOpts{}
-		regular := NewEngine(opts, q, NoLimits, log.NewNopLogger())
-		sharded := NewDownstreamEngine(opts, MockDownstreamer{regular}, NoLimits, log.NewNopLogger())
-
-		t.Run(tc.query, func(t *testing.T) {
+		t.Run(tc.query+"_instant", func(t *testing.T) {
+			// for an instant query we set the start and end to the same timestamp
+			// plus set step and interval to 0
 			params, err := NewLiteralParams(
 				tc.query,
-				start,
-				end,
-				step,
-				interval,
+				time.Unix(0, int64(rounds+1)),
+				time.Unix(0, int64(rounds+1)),
+				0,
+				0,
 				logproto.FORWARD,
 				uint32(limit),
 				nil,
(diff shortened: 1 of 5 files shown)


pkg/logql/downstream_test.go
```diff
@@ -181,6 +181,70 @@ func TestMappingEquivalenceSketches(t *testing.T) {
 	}
 }
 
+func TestMappingEquivalenceSketches_Instant(t *testing.T) {
```
jeschkies, line 184: What makes this an instant query? What do you think about song another loop to the original equivalence tests?

pkg/logql/quantile_over_time_sketch.go
```diff
@@ -283,6 +283,11 @@ func JoinQuantileSketchVector(next bool, r StepResult, stepEvaluator StepEvaluat
 		if stepEvaluator.Error() != nil {
 			return nil, stepEvaluator.Error()
 		}
+		// Don't continue to evaluate more steps if we already filled the
+		// # of steps we should have based on start/end and step params.
+		if len(result) == stepCount {
```
jeschkies, line 288: How's that done in the original join?

jeschkies (changes requested):
Nice find. We altos take the chance and rename the join methods to merge.

pkg/logql/quantile_over_time_sketch.go
jeschkies: Found it https://github.com/grafana/loki/blob/918e7365dbbe4ac33ec02b803c6cf7155eb4a9e1/pkg/logql/engine.go#L388

pkg/logql/downstream_test.go
cstyan: The query params right after this line, in the case of an instant query we set `start` and `end` to the same value.

jeschkies (approved):
🎉

PR #11917, opened 2024-02-10
Use `bloomcompactor.DayTable` in bloom gateway code

chaudum:
Where previously a `model.Time` was used to define the day of a task.


chaudum:
We should remove `bloomcompactor.DayTable` in favour of `config.Day`.

salvacorts (approved):

PR #11918, opened 2024-02-11
Bloom/running

owen-d:
Includes a few small fixes in order to get the compactor running, including
* Wires up i/o from `v1` pkg and the `bloomshipper` pkg in order to archive+write blocks remotely after building
* config error handling
* fix an error iterating chunks
* handles gzipping when loading tsdbs from storage
* bypasses internal cache in `BloomTSDBStore`'s internal `indexshipper/storage.Client` to get up to date results, which is easier for prototyping
* `ClosableReadSeekerAdapter`


pkg/loki/loki.go
```diff
@@ -648,7 +651,7 @@ func (t *Loki) setupModuleManager() error {
 		Write:   {Ingester, Distributor},
 		Backend: {QueryScheduler, Ruler, Compactor, IndexGateway, BloomGateway, BloomCompactor},
 
-		All: {QueryScheduler, QueryFrontend, Querier, Ingester, Distributor, Ruler, Compactor},
+		All: {QueryScheduler, QueryFrontend, Querier, Ingester, Distributor, Ruler, Compactor, BloomCompactor},
```
chaudum, line 654: Up for discussion, but I don't think that we want to include the bloom compactor in the `all` target.
For testing, it comes handy, I understand.

chaudum (approved):
lgtm :shipit:

pkg/loki/loki.go
salvacorts: I don't think users of the monolith mode will benefit from building blooms significantly, right? In addition to that, it will increase resource usage and operational burden.

salvacorts (approved):

pkg/bloomcompactor/spec.go
```diff
@@ -313,6 +311,12 @@ func (b *batchedLoader) Next() bool {
 	}
 
 	b.batch, b.err = next.fetcher.FetchChunks(b.ctx, toFetch)
+	return b.prepNext()
```
chaudum, line 314: FetchChunks may return an error, then `len(b.batch)` is 0, which will cause an out of bounds error in `prepNext()`.

owen-d: Good catch, will fix. Thanks

pkg/loki/loki.go
owen-d: Fair points all around, but we can leave it to enable better testing then remove it prior to release.

PR #11923, opened 2024-02-13
Update links

JStickler:
Fixes broken links in https://grafana.com/docs/loki/latest/release-notes/cadence/


docs/sources/release-notes/cadence.md
```diff
@@ -8,7 +8,7 @@ weight: 1
 
 ## Stable Releases
 
-Loki releases (this includes [Promtail](/clients/promtail), [Loki Canary](/operations/loki-canary/), etc) use the following
+Loki releases (this includes [Promtail](https://grafana.com/docs/loki/latest/send-data/promtail/), [Loki Canary](https://grafana.com/docs/loki/latest/operations/loki-canary/), etc) use the following
```
jdbaldry, line 11:
```suggestion
Loki releases (this includes [Promtail](https://grafana.com/docs/loki/<LOKI_VERSION>/send-data/promtail/), [Loki Canary](https://grafana.com/docs/loki/<LOKI_VERSION>/operations/loki-canary/), etc) use the following
```

jdbaldry (commented):
If we use `<LOKI_VERSION>` instead of `latest`, then these links will resolve correctly independent of the documentation version.

The example in Writers' Toolkit uses Grafana but the same behavior applies here: https://grafana.com/docs/writers-toolkit/write/links/#examples.

For more information about how version substitution works, refer to [About version substitution](https://grafana.com/docs/writers-toolkit/write/shortcodes/#about-version-substitution)

docs/sources/release-notes/cadence.md
JStickler: I literally do not understand how there is any difference between specifying `latest` in the URL and specifying `latest` as the <LOKI_VERSION>.  It's still "latest".

jdbaldry: Because these docs will also be in `next` and then they will link to `latest` instead of `next`.
And when you release the next version of Loki, the previous docs will also link to `latest` instead of linking within the current version.

You can choose to not use the `<LOKI_VERSION>` syntax but it will mean that you need to update your links as part of every release and it will negatively impact backporting.

JStickler: Trying to think this through.... Going on what is currently documented in the Writers Toolkit, if the root `_index/md` file has `LOKI_VERSION: latest` then you still have the same problem of the links in `next` pointing to `latest`.

It isn't documented, but for links to resolve correctly wouldn't you need the following?
Latest release (Loki 2.9) -> `_index/md` file has `LOKI_VERSION: latest`?
Or even better, root `_index/md` file has `LOKI_VERSION: 2.9`?
Main branch  -> root `_index/md` file has `LOKI_VERSION: next`?

jdbaldry: > The shortcode substitutes the special syntax `<SOMETHING_VERSION>` with the version inferred from the page’s URL. If the page’s URL has the prefix /docs/grafana/latest/, the shortcode replaces the syntax `<SOMETHING_VERSION>` with latest in the final URL.
>
> You can override version inference by including additional metadata in the front matter of the file. To override the value of `<GRAFANA_VERSION>`, set the `GRAFANA_VERSION` parameter in the page’s front matter. For example, to set the version to next irrespective of the source content version, add the following to the front matter: `GRAFANA_VERSION: next`.

If you don't set the front matter, the version is inferred from the page's URL

jdbaldry: The front matter override is primarily there so that other projects can specify the version of the documentation they are interested in linking to. For example, Grafana Cloud sets `GRAFANA_VERSION: next` so that all links to Grafana documentation go to the `next` version. The Mimir helm chart sets [`MIMIR_DOCS_VERSION`](https://github.com/grafana/mimir/blob/776e998a6d9c0dfac0b4cf1ef64d18b15c8eba04/docs/sources/helm-charts/mimir-distributed/_index.md?plain=1#L11) to `v2.10.x` so that they can update a single place when the Mimir helm chart is updated to be compatible with a different version of Mimir and then all links point to the correct version of the Mimir docs.

JStickler: >If you don't set the front matter, the version is inferred from the page's URL

The way that the Writers Toolkit is currently written tells you to set the value for the version in the root _index.md file.  I would never have guessed that that was an optional step.

jdbaldry: PRs are always welcome

JStickler:
I have been unable to update this PR for the past two days.   Closing this and opened a new PR (#11956).

PR #11930, opened 2024-02-13
Tail queue

salvacorts:


pkg/ingester/stream.go
```diff
@@ -288,30 +288,28 @@ func (s *stream) recordAndSendToTailers(record *wal.Record, entries []logproto.E
 	hasTailers := len(s.tailers) != 0
 	s.tailerMtx.RUnlock()
 	if hasTailers {
-		go func() {
```
na--, line 291: Hmm I don't understand the surrounding code, but from the `recordAndSendToTailers()` name and from the previous usage of a goroutine here, I assume this part should be as non-blocking as possible? If so, does it make sense to keep the goroutine here? We'd still handle the actual sending over a queue, but this part won't be blocking either.

salvacorts: We are removing the goroutine and sending the stream to the queue. Since the queue is a buffered channel, it's a non-blocking opt (if chan is full, the stream is dropped)

pkg/ingester/tailer.go
```diff
@@ -102,6 +115,37 @@ func (t *tailer) loop() {
 	}
 }
 
+func (t *tailer) receiveStreamsLoop() {
+	for {
+		select {
+		case <-t.conn.Context().Done():
+			t.close()
```
slim-bean, line 122: I think maybe t.close() should be in a `defer` block so that anywhere we return from this function it would be called?

slim-bean (approved):
one change to close the the tailer if we exit the receive loop for any reason, otherwise LGTM!

pkg/ingester/tailer.go
salvacorts: Done

PR #11941, opened 2024-02-14
meta & removal support for bloom compaction

owen-d:
Building of `meta.jsons` and removal of old `meta.jsons` + bloom blocks in the `bloomcompactor`


slim-bean (approved):
turn and burn baby

PR #11950, opened 2024-02-14
Bloom compactor: Lazy load blocks in batches

chaudum:

This PR implements lazy loading of blocks when building new blocks from existing ones using the MergeBuilder.

The signature of the `NewMergeBuilder()` function changed so it accepts a single `Iterator[*SeriesWithBloom]` instead of a list of `PeekingIterator[*SeriesWithBloom]`.

The required Iterator is built using the `blockLoadingIter` which implements `CloseableIterator` and `ResettableIterator`.

The `blockLoadingIter` uses a `batchedLoader` to fetch grouped overlapping blocks in batches.


pkg/bloomcompactor/batch.go
```diff
@@ -2,94 +2,357 @@ package bloomcompactor
...
+			}
+			return b
+		},
+		v1.NewPeekingIter(mergedBlocks),
+	)
 	return true
 }
+
+func overlappingBlocksIter(inputs []bloomshipper.BlockRef) v1.Iterator[[]bloomshipper.BlockRef] {
+	// can we assume sorted blocks?
```
owen-d, line 341: Yes -- if it's not we can consider that a bug elsewhere and fix

pkg/bloomcompactor/batch.go
```diff
@@ -2,94 +2,357 @@ package bloomcompactor
...
+// Reset implements v1.ResettableIterator.
+// TODO(chaudum) Cache already fetched blocks to to avoid the overhead of
+// creating the reader.
+func (i *blockLoadingIter) Reset() error {
+	if !i.initialized {
+		return nil
+	}
+	i.initialized = false
+	// close loaded queriers
+	return i.Close()
```
owen-d, line 253:
```suggestion
	// close loaded queriers
	err := i.Close()
	i.initialized = false
	clear(i.loaded)
	return err
```

Need to empty the loaded blocks on reset.

owen-d (approved):
One change but lgtm

PR #11960, opened 2024-02-15
Replace min/max token with `TokenRange` in bloom ring utilities

chaudum:
This PR replaces min/max token fields with a `TokenRange` field that uses the `Range[uint32]` type.
The `Range[uint32]` uses similar semantics as the `FingerprintBounds` we use for fingerprint ranges.

### DIFF
--- a/pkg/bloomutils/ring.go
+++ b/pkg/bloomutils/ring.go
@@ -9,16 +9,13 @@ import (
 
 	"github.com/grafana/dskit/ring"
 	"github.com/prometheus/common/model"
+	"golang.org/x/exp/constraints"
 	"golang.org/x/exp/slices"
 
 	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
 )
 
-type integer interface {
-	~uint | ~uint32 | ~uint64 | ~int | ~int32 | ~int64
-}
-
-type Range[T integer] struct {
+type Range[T constraints.Integer] struct {
 	Min, Max T
 }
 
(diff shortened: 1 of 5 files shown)


pkg/bloomutils/ring.go
```diff
@@ -14,20 +14,47 @@ import (
 	v1 "github.com/grafana/loki/pkg/storage/bloom/v1"
 )
 
-type InstanceWithTokenRange struct {
-	Instance           ring.InstanceDesc
-	MinToken, MaxToken uint32
+type integer interface {
```
salvacorts, line 17: I think you can use `constrainsts.Integer` instead:
https://pkg.go.dev/golang.org/x/exp/constraints#Integer

pkg/bloomutils/ring.go
```diff
@@ -14,20 +14,47 @@ import (
...
+	~uint | ~uint32 | ~uint64 | ~int | ~int32 | ~int64
 }
 
-func (i InstanceWithTokenRange) Cmp(token uint32) v1.BoundsCheck {
-	if token < i.MinToken {
+type Range[T integer] struct {
```
salvacorts, line 21: Maybe `GetInstanceWithTokenRange` can be generic and return Range[T]. So we can map tokens to all ok the Integer key-spaces.

salvacorts: Same for `NewInstanceSortMergeIterator`

salvacorts (commented):
I think this is **great**. Left some comments. I think being able to map the tokens to whichever key-space we want by using generics would be super handy.

chaudum:
> I think this is **great**. Left some comments. I think being able to map the tokens to whichever key-space we want by using generics would be super handy.

Yeah, good idea

salvacorts (approved):

PR #11968, opened 2024-02-15
fix(operator): Extend status to show difference between running and ready

xperimental:

Currently the Loki Operator emits the `ReadyCondition` even when there are pods running but not ready yet. It looks like this causes error conditions to get lost by support and users, because they do not notice the pods in unready state.

This PR tries to fix this issue by only emitting the `ReadyCondition` when all LokiStack managed pods are ready and falling back to `PendingCondition` with a different message when all pods are running but some are missing their readiness check.

Fixes [LOG-5160](https://issues.redhat.com/browse/LOG-5160)

- An alternative implementation would be to introduce a new condition instead of including the "running but not ready" status in the `PendingCondition`.
- This PR also contains an unrelated change ("storage status is overwritten"). This can be split into a separate PR.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/operator/CHANGELOG.md
+++ b/operator/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Main
 
+- [11968](https://github.com/grafana/loki/pull/11968) **xperimental**: Extend status to show difference between running and ready
 - [12007](https://github.com/grafana/loki/pull/12007) **xperimental**: Extend Azure secret validation
 - [12008](https://github.com/grafana/loki/pull/12008) **xperimental**: Support using multiple buckets with AWS STS
 - [11964](https://github.com/grafana/loki/pull/11964) **xperimental**: Provide Azure region for managed credentials using environment variable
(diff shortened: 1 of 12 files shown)


operator/internal/status/storage.go
JoaoBraveCoding, line 1: I might have missed something but this change seems unrelated

JoaoBraveCoding (commented):
Overall looks like a great addition

operator/internal/status/storage.go
xperimental: No, you got that right. Is is unrelated (see reviewer notes in description). :slightly_smiling_face:

xperimental: As luck would have it, I just noticed another issue with the status updates. Maybe instead of tacking the change onto this PR I'll group both of the "status fixes" into a separate PR instead...

periklis (approved):

PR #11983, opened 2024-02-19
Various bloom gateway instrumentation changes

chaudum:

See individual commit messages for details.

### DIFF
--- a/pkg/bloomgateway/bloomgateway.go
+++ b/pkg/bloomgateway/bloomgateway.go
@@ -82,10 +82,12 @@ var (
 )
 
 type metrics struct {
-	queueDuration       prometheus.Histogram
-	inflightRequests    prometheus.Summary
-	chunkRefsUnfiltered prometheus.Counter
-	chunkRefsFiltered   prometheus.Counter
+	queueDuration    prometheus.Histogram
+	inflightRequests prometheus.Summary
+	chunksTotal      prometheus.Counter
+	chunksFiltered   prometheus.Counter
+	seriesTotal      prometheus.Counter
+	seriesFiltered   prometheus.Counter
 }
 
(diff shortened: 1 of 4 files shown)


pkg/bloomgateway/bloomgateway.go
```diff
@@ -419,16 +393,18 @@ outer:
...
 
 	for res := range task.resCh {
 		select {
 		case <-ctx.Done():
 			level.Debug(logger).Log("msg", "drop partial result", "fp_int", uint64(res.Fp), "fp_hex", res.Fp, "chunks_to_remove", res.Removals.Len())
+			g.metrics.chunkRemovals.WithLabelValues("dropped").Add(float64(res.Removals.Len()))
 		default:
 			level.Debug(logger).Log("msg", "accept partial result", "fp_int", uint64(res.Fp), "fp_hex", res.Fp, "chunks_to_remove", res.Removals.Len())
 			task.responses = append(task.responses, res)
+			g.metrics.chunkRemovals.WithLabelValues("accepted").Add(float64(res.Removals.Len()))
```
salvacorts, line 407: I don't think the meaning of _dropped_ or _accepted_ is intuitive for the operators. Maybe we can add what they mean in the metric _help_? Or chose different label values?

salvacorts (approved):

pkg/bloomgateway/bloomgateway.go
chaudum: Does this description make sense to you? https://github.com/grafana/loki/pull/11983/commits/0431acd2419d1752529913b9db0f67c2ba46711c

salvacorts: Yes, that makes perfect sense now? :) Thank you!

PR #11987, opened 2024-02-19
refactor: Limit processing of blocks to requested fingerprint ranges in bloom gateway

chaudum:

This PR limits the amount of data being processed for a single multiplexed iteration to the union of the fingerprint bounds of its requests, instead of looking at blocks from the complete fingerprint range.


pkg/bloomgateway/util.go
```diff
@@ -184,3 +184,29 @@ func partitionRequest(req *logproto.FilterChunkRefRequest) []seriesWithBounds {
 
 	return result
 }
+
+type MultiFingerprintBounds []v1.FingerprintBounds
```
chaudum, line 188: Should this belong to the v1 package?

pkg/bloomgateway/util_test.go
```diff
@@ -385,3 +385,108 @@ func createBlockRefsFromBlockData(t *testing.T, tenant string, data []*bloomship
...
+				v1.NewBounds(5, 9),
+				v1.NewBounds(15, 19),
+				v1.NewBounds(25, 29),
+			},
+			target: v1.NewBounds(9, 15),
+			exp: MultiFingerprintBounds{
+				v1.NewBounds(0, 2),
+				v1.NewBounds(5, 19),
+				v1.NewBounds(25, 29),
+			},
```
owen-d, line 484: nit: add a test where there is no overlap but the comparison is between the other two:
```
mb := bounds(0,10) + bounds (30,40)
target = bounds(15,25)
```

pkg/bloomgateway/util.go
```diff
@@ -184,3 +184,29 @@ func partitionRequest(req *logproto.FilterChunkRefRequest) []seriesWithBounds {
...
+
+func (mb MultiFingerprintBounds) Union(target v1.FingerprintBounds) MultiFingerprintBounds {
```
owen-d, line 190: I think for this to behave as expected, the multi-bounds must be
a) sorted already
b) contain disjoint (non-overlapping) ranges

Otherwise you can end up with many overlapping ranges in the result set as well.

Consider:
```
multi = [(0,6), (5,15)]
target = (8,10)

result = [(0,6), (5,15)] // can be further reduced
```

Thinking about it, it's probably not a real concern we have if we can ensure the multi-bounds are ordered and non overlapping.

owen-d (approved):

pkg/bloomgateway/util.go
chaudum: I solved the issue by using a different algorithm.

salvacorts (approved):

PR #12003, opened 2024-02-20
Add o11y for chunk data retrieved from ingesters

dannykopping:
We're currently missing some insight about how much work ingesters are doing as part of read requests.
With this data we can estimate how much query work ingesters are doing and make some scaling inferences from that.


pkg/logql/metrics.go
```diff
@@ -173,6 +173,26 @@ func RecordRangeAndInstantQueryMetrics(
...
+		// The total of chunk reference fetched from index.
+		"ingester_chunk_refs", stats.Ingester.Store.GetTotalChunksRef(),
+		// Total number of chunks fetched.
+		"ingester_chunk_downloaded", stats.Ingester.Store.GetTotalChunksDownloaded(),
+		// Time spent fetching chunks in nanoseconds.
+		"ingester_chunk_fetch_time", stats.Ingester.Store.ChunksDownloadDuration(),
+		// Total of chunks matched by the query from ingesters.
+		"ingester_chunk_matches", stats.Ingester.GetTotalChunksMatched(),
+		// Total ingester reached for this query.
+		"ingester_requests", stats.Ingester.GetTotalReached(),
```
kavirajk, line 185: nice. Didn't know we had this data 👍

pkg/logql/metrics.go
```diff
@@ -200,6 +220,10 @@ func RecordRangeAndInstantQueryMetrics(
 	recordUsageStats(queryType, stats)
 }
 
+func humanizeBytes(val uint64) string {
```
kavirajk, line 223: nit: Some example in the comment would be helpful.

kavirajk (approved):
LGTM 👍 one minor nit.

pkg/logql/metrics.go
```diff
@@ -173,6 +173,26 @@ func RecordRangeAndInstantQueryMetrics(
 		"cache_result_hit", resultCache.EntriesFound,
 		"cache_result_download_time", resultCache.CacheDownloadTime(),
 		"cache_result_query_length_served", resultCache.CacheQueryLengthServed(),
+		// The total of chunk reference fetched from index.
+		"ingester_chunk_refs", stats.Ingester.Store.GetTotalChunksRef(),
+		// Total number of chunks fetched.
+		"ingester_chunk_downloaded", stats.Ingester.Store.GetTotalChunksDownloaded(),
+		// Time spent fetching chunks in nanoseconds.
+		"ingester_chunk_fetch_time", stats.Ingester.Store.ChunksDownloadDuration(),
```
slim-bean, line 181: we may want to take these back out? ingesters don't fetch chunks so I think this will always be zero values?

pkg/logql/metrics.go
```diff
@@ -173,6 +173,26 @@ func RecordRangeAndInstantQueryMetrics(
...
+		// Total ingester reached for this query.
+		"ingester_requests", stats.Ingester.GetTotalReached(),
+		// Total bytes processed but was already in memory (found in the headchunk). Includes structured metadata bytes.
+		"ingester_chunk_head_bytes", humanizeBytes(uint64(stats.Ingester.Store.Chunk.GetHeadChunkBytes())),
+		// Total bytes of compressed chunks (blocks) processed.
+		"ingester_chunk_compressed_bytes", humanizeBytes(uint64(stats.Ingester.Store.Chunk.GetCompressedBytes())),
+		// Total bytes decompressed and processed from chunks. Includes structured metadata bytes.
+		"ingester_chunk_decompressed_bytes", humanizeBytes(uint64(stats.Ingester.Store.Chunk.GetDecompressedBytes())),
+		// Total duplicates found while processing.
+		"ingester_chunk_duplicates", stats.Ingester.Store.Chunk.GetTotalDuplicates(),
```
slim-bean, line 193: this can never be anything other than 0 in an ingester I believe and can probably be removed

PR #11633, opened 2024-01-10
feat: Add tracing integration to profiling.

cyriltovena:
**What this PR does / why we need it**:

Same as https://github.com/grafana/tempo/pull/3276 this adds profiling integration to tracing instrumentation allowing to get profile for a single request removing the noise of everything else.

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

Thanks me later.

**Checklist**
- [x] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [x] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated
  - [ ] If the change is worth mentioning in the release notes, add `add-to-release-notes` label
- [x] Changes that require user attention or interaction to upgrade are documented in `docs/sources/setup/upgrade/_index.md`
- [x] For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)
- [x] If the change is deprecating or removing a configuration option, update the `deprecated-config.yaml` and `deleted-config.yaml` files respectively in the `tools/deprecated-config-checker` directory. [Example PR](https://github.com/grafana/loki/pull/10840/commits/0d4416a4b03739583349934b96f272fb4f685d15)

cc @poyzannur we talked about this for compaction to get profiling per run, all you need is a root span per run !

PR #12004, opened 2024-02-20
Set tracing.profiling-enabled to true by default

salvacorts:

Fixes https://github.com/grafana/loki/pull/11633


cyriltovena (approved):
LGTM

PR #12012, opened 2024-02-20
skips publishing images on PRs

owen-d:
We only need to publish images on `main` -- there's no need to publish PR builds and they can cause us to be rate limited much more easily, increase build times, etc.


trevorwhitney (changes requested):
@owen-d how urgently do you need this? I've also removed this in https://github.com/grafana/loki/pull/11808, as well as moving the checks over to GH actions. Can we wait on that to avoid merge conflicts?

trevorwhitney:
Better yet, I can merge #11808 as soon as it's green and move any additional release pipeline work into a new PR

trevorwhitney:
closing in favor or #11808

PR #12018, opened 2024-02-21
refactor: Simplify fingerprint partitioning in bloom gateway client

chaudum:

Use the recently introduced `tokenRangesForInstance()` function to determine the the ranges for each replication set.

For this purpose, the function moves into the `bloomutils` packages.

This PR eliminates the structs `addrsWithBounds` and `instanceWithFingerprints` (which had somewhat similar usage) and unifies them into `rsWithRanges` that holds both the fingerprint ranges and the assigned GroupedChunkRefs.


owen-d (approved):
skimmed, lgtm

PR #12019, opened 2024-02-21
fix: Ensure working dir for bloomstore exists

chaudum:

Fail startup if directory does not exist or there are not enough permissions. This prevents the bloomstore to fail later in the process once it tries to download and extract blocks.

### DIFF
--- a/pkg/storage/chunk/client/util/util.go
+++ b/pkg/storage/chunk/client/util/util.go
@@ -72,6 +72,8 @@ func EnsureDirectory(dir string) error {
 		return os.MkdirAll(dir, 0o777)
 	} else if err == nil && !info.IsDir() {
 		return fmt.Errorf("not a directory: %s", dir)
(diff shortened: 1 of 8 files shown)


salvacorts (approved):

owen-d (approved):

pkg/storage/chunk/client/util/util.go
```diff
@@ -72,6 +72,8 @@ func EnsureDirectory(dir string) error {
 		return os.MkdirAll(dir, 0o777)
 	} else if err == nil && !info.IsDir() {
 		return fmt.Errorf("not a directory: %s", dir)
+	} else if err == nil && info.Mode()&0700 != 0700 {
```
salvacorts, line 75: I'd maybe set these `0700` masks as const describing which permissions they set.

salvacorts (approved):

PR #12023, opened 2024-02-21
feat: [helm] Support for PVC Annotations for Non-Distributed Modes

onelapahead:

This adds the ability to specify PVC annotations in the volume claim templates of all StatefulSets in the chart (read, write, backend, and singleBinary). 

CSI drivers often allow for customizing PVCs, or modifying existing PVCs, via annotations. For example, in the open-source the AWS EBS CSI driver allows for changing IOPS, storage type, etc. via annotations: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/modify-volume.md.

Supporting PVC annotations is quite common in Helm charts. For example, the `minio` chart that is a dependency of the `loki` chart also allows for such customization and this PR additionally sets that value so it is more clearly documented for `loki` chart users.

N/A

**Checklist**
- [x] Documentation added
- [ ] Tests updated
- [x] `CHANGELOG.md` updated


production/helm/loki/templates/backend/statefulset-backend.yaml
```diff
@@ -267,6 +267,12 @@ spec:
       kind: PersistentVolumeClaim
       metadata:
         name: data
+        {{- if .Values.backend.persistence.annotations }}
```
hfuss, line 270: We wrap the `with` in an `if` for backwards compatibility, to avoid making an unnecessary changes to immutable fields of the `StatefulSets` for existing `loki` installations.

hfuss:
Apologies messed up a rebase, one sec

hfuss:
There's some extra checks / reviewers still on the PR due to the rebase I messed up, but the PR is now ready for review again and up-to-date.

hfuss:
@MichelHollands Any estimate as to where this PR is in the queue among all the other open Helm PRs ?

hfuss:
See https://github.com/grafana/loki/pull/12281 for another example where annotations on persistent storage would be valuable.

danielfrg:
Thanks for the PR. I was facing the same issue and made the same fix.

Hopefully we can move this one forward!

onelapahead:
Noticing @JStickler on https://github.com/grafana/loki/pull/12023/commits/91c5520049ff51812b7954df63f86b86921877de the checks that failed seemed all cosmetic - it says `helm:` isn't an appropriate PR title prefix (though swore I read in the contributor guide to use that) and then that I need to rerun `make -C docs sources/setup/install/helm/reference.md`. When doing that `make` (on MacOS) considered the build target "built" and that it was up-to-date.

I've added a fix to the Makefile to always builds the docs when run but that could be I'm missing something. Look forward to the feedback! And thanks @danielfrg for the other contribution and helping promote the need for this 🙇🏻

docs/Makefile
```diff
@@ -10,6 +10,7 @@ include docs.mk
 PODMAN := $(shell if command -v podman >/dev/null 2>&1; then echo podman; else echo docker; fi)
 BUILD_IN_CONTAINER ?= true
 
+.PHONY: sources/setup/install/helm/reference.md
```
onelapahead, line 13: Otherwise when running `make -C docs sources/setup/install/helm/reference.md` like the CI jobs recommended, my `make` didn't think there was anything to update.. is there a flag I'm missing you'd prefer me use rather than `.PHONY` ?

onelapahead:
@MichelHollands @JStickler any thoughts on testing this one again or initially reviewing in case I missed anything obvious ? 

We could really use the ability to have this (and another user has reported the need for it too) without having to resort to forking the chart for awhile.

I also see PRs like #12434 in-progress where new modes for running the microservices model are being added, so it'd be good to get this persistence annotations pattern in place so its not missed on new components going forward. Thanks.

JStickler:
@onelapahead I'm the technical writer on the team, I've run the build and fixed the title of your PR so that it passes the Conventional Commits test (which might be new since you opened this PR), but I'm not familiar enough with Helm charts to approve code changes.  Hopefully @MichelHollands can take a look at this in the next day or two.

onelapahead:
Awesome no worries @JStickler appreciate you running the checks just so I can get the feedback of anything I should fix before its further reviewed 🙇🏻

danielfrg:
Could we get some eyes on this one please.
Hopefully should be a quick one to review.

docs/Makefile
trevorwhitney: I'm confused why this needs to be a `.PHONY` when this file does exist. maybe you need to `clean` first? But I'm guessing it failed because the file existed even though the contents weren't what they should be, and the `.PHONY` just forces it to always run? I guess that's a fair trade off vs having to remember to `clean` everytime.

production/helm/loki/CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ Entries should include a reference to the pull request that introduced the chang
 
 [//]: # (<AUTOMATED_UPDATES_LOCATOR> : do not remove this line. This locator is used by the CI pipeline to automatically create a changelog entry for each new Loki release. Add other chart versions and respective changelog entries bellow this line.)
 
+## 5.48.0
```
trevorwhitney, line 16: I think we're on a a new major now.

production/helm/loki/templates/backend/statefulset-backend.yaml
trevorwhitney: why would just using a `with` cause changes?

trevorwhitney (approved):
Looks good to me, thanks! Approving to unblock, but I don't understand why we have to wrap the `with` in an `if`?

docs/Makefile
onelapahead: Hey yeah sorry `clean` should have been obvious to me, but I guess I am use to codegen/docgen/chartgen/etc. tooling via Makefile (like kubebuilder's make targets for example) always running / re-generating code since it should be deterministic, and therefore you don't have to remember to run `clean` which is nice.

I recognize this is a bit overstepping in the PR's scope and could have unintended consequences in the CI workflows,etc. potentially so if you'd prefer I back this out that's fine with me!

production/helm/loki/templates/backend/statefulset-backend.yaml
onelapahead: https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with

> Note that we removed the if conditional from the previous exercise because it is now unnecessary - the block after with only executes if the value of PIPELINE is not empty.

Right sorry for some reason I think it'd do:
```
annotations:
  {}
```
w/o the `if` and make a "breaking" change to the STS' volumeClaimTemplate. Confirmed that's not the case so will revert as part resolving conflicts.

onelapahead:
Thanks @trevorwhitney for the feedback and patience, believe everything has been addressed now! Noting that all PVCs / claim templates in the new distributed mode of deployment had annotations support, so this PR is largely just adding this support to single binary and simple scalable modes now.

PR #12029, opened 2024-02-21
feat: add the ability to inject log parsers to the distributor

MasslessParticle:
The facilitate changing parsing behavior in the distributor, this PR adds the ability to wrap the existing log parsers. It also exports the push stats so they're accessible to the wrapping parser.


ashwanthgoli (approved):
lgtm

na-- (approved):

PR #12035, opened 2024-02-22
feat: Convert line filter expressions to a set of bloom tests

salvacorts:
This PR introduces:
- A new interface `BloomTest` with two test methods that take a bloom as an argument and returns a boolean. True if the test passes with the given bloom.
- A new function `FiltersToBloomTest` that given a set of `LineFilterExpr`, it returns a `BloomTest`.

~The logic in this PR is not yet used anywhere since we need to do some refactoring to be able to use it. These refactorings will be done on a set of follow-up PRs:~
~1. Pass `syntax.LineFilterExpr` instead of `syntax.LineFilter` with `logproto.FilterChunkRefRequest`~
~2. Create BloomTest from the line filters passed along through `logproto.FilterChunkRefRequest` and modify the fused querier to use the resulting `BloomTest`.~
3. Map regex line filters to `BloomTest` when possible. The logic would be pretty similar to what's done in the `simplify` function at [pkg/logql/log/filter.go](https://github.com/grafana/loki/blob/60edfe0006b0adf6041b7ac437b8c47d66d1ead6/pkg/logql/log/filter.go#L470-L496). We can probably extract the simplification logic into [pkg/util/regex.go].(https://github.com/grafana/loki/blob/5a08a6bcb9ecc816328cb7859bbdcc7cf9d004c5/pkg/util/regex.go)

- As explained in the bullets above, regexes are not yet tested. all regex filter expressions are mapped to Noop tests which always return true.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/storage/bloom/v1/bloom_tester.go
+++ b/pkg/storage/bloom/v1/bloom_tester.go
@@ -8,24 +8,24 @@ import (
 )
 
 type BloomTest interface {
-	Test(bloom filter.Checker) bool
-	TestWithPrefixBuf(bloom filter.Checker, buf []byte, prefixLen int) bool
+	Matches(bloom filter.Checker) bool
+	MatchesWithPrefixBuf(bloom filter.Checker, buf []byte, prefixLen int) bool
 }
 
 type BloomTests []BloomTest
 
-func (b BloomTests) Test(bloom filter.Checker) bool {
+func (b BloomTests) Matches(bloom filter.Checker) bool {
 	for _, test := range b {
-		if !test.Test(bloom) {
+		if !test.Matches(bloom) {
 			return false
 		}
 	}
 	return true
 }
 
-func (b BloomTests) TestWithPrefixBuf(bloom filter.Checker, buf []byte, prefixLen int) bool {
+func (b BloomTests) MatchesWithPrefixBuf(bloom filter.Checker, buf []byte, prefixLen int) bool {
 	for _, test := range b {
-		if !test.TestWithPrefixBuf(bloom, buf, prefixLen) {
+		if !test.MatchesWithPrefixBuf(bloom, buf, prefixLen) {
 			return false
 		}
 	}
@@ -61,21 +61,21 @@ func simpleFilterToBloomTest(b NGramBuilder, filter syntax.LineFilter) BloomTest
 	case labels.MatchRegexp, labels.MatchNotRegexp:
 		// TODO(salvacorts): Simplify regex similarly to how it's done at pkg/logql/log/filter.go (`simplify` function)
 		// 					 Ideally we want to extract the simplify logic into pkg/util/regex.go
-		return NoopTest
+		return MatchAll
 	default:
-		return NoopTest
+		return MatchAll
 	}
 }
 
-type noopTest struct{}
+type matchAllTest struct{}
 
-var NoopTest = noopTest{}
+var MatchAll = matchAllTest{}
 
-func (n noopTest) Test(_ filter.Checker) bool {
+func (n matchAllTest) Matches(_ filter.Checker) bool {
 	return true
 }
 
-func (n noopTest) TestWithPrefixBuf(_ filter.Checker, _ []byte, _ int) bool {
+func (n matchAllTest) MatchesWithPrefixBuf(_ filter.Checker, _ []byte, _ int) bool {
 	return true
 }
 
@@ -101,7 +101,7 @@ func newStringTest(b NGramBuilder, search string) stringTest {
 	return test
 }
 
-func (b stringTest) Test(bloom filter.Checker) bool {
+func (b stringTest) Matches(bloom filter.Checker) bool {
 	for _, ngram := range b.ngrams {
 		if !bloom.Test(ngram) {
 			return false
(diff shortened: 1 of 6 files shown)


pkg/storage/bloom/v1/bloom_tester.go
```diff
@@ -0,0 +1,156 @@
...
+
+import (
+	"github.com/prometheus/prometheus/model/labels"
+
+	"github.com/grafana/loki/pkg/logql/syntax"
+	"github.com/grafana/loki/pkg/storage/bloom/v1/filter"
+)
+
+type BloomTest interface {
+	Test(bloom filter.Checker) bool
```
chaudum, line 11: nit: `Test()` does not give any clue about what the returned `true` or `false` means.

`BloomTest` is a [predicate](https://en.wikipedia.org/wiki/Predicate_%28mathematical_logic%29) of `Request`.

```go
type MatchesBloom interface {
    ...
}
```

or

```go
struct BloomTest interface {
    Matches(bloom filter.Checker) bool
    ...
}
```

pkg/storage/bloom/v1/bloom_tester.go
```diff
@@ -0,0 +1,156 @@
...
+	case labels.MatchRegexp, labels.MatchNotRegexp:
+		// TODO(salvacorts): Simplify regex similarly to how it's done at pkg/logql/log/filter.go (`simplify` function)
+		// 					 Ideally we want to extract the simplify logic into pkg/util/regex.go
+		return NoopTest
+	default:
+		return NoopTest
+	}
+}
+
+type noopTest struct{}
```
chaudum, line 70: nit like above: this could be named `matchAll`

pkg/storage/bloom/v1/bloom_tester_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+func extractLineFilters(expr syntax.Expr) (lineFilters []syntax.LineFilterExpr) {
+	visitor := &syntax.DepthFirstTraversal{
+		VisitLineFilterFn: func(v syntax.RootVisitor, e *syntax.LineFilterExpr) {
+			lineFilters = append(lineFilters, *e)
+		},
+	}
+	expr.Accept(visitor)
+	return lineFilters
+}
```
chaudum, line 137: I think you can use the function from the `syntax` package, right?

pkg/storage/bloom/v1/bloom_tester_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+type fakeNgramBuilder struct{}
```
chaudum, line 139: this is actually not a fake ngram builder, but a very simple tokenizer, right? 😀

pkg/storage/bloom/v1/tokenizer.go
```diff
@@ -42,8 +42,8 @@ func NewNGramTokenizer(n, skip int) *NGramTokenizer {
 
 // The Token iterator uses shared buffers for performance. The []byte returned by At()
 // is not safe for use after subsequent calls to Next()
-func (t *NGramTokenizer) Tokens(line string) NGramTokenIter {
-	return NGramTokenIter{
+func (t *NGramTokenizer) Tokens(line string) Iterator[[]byte] {
```
chaudum, line 45: Can you add a comment that this function implements the new `NGramBuilder` interface?

pkg/storage/bloom/v1/tokenizer.go
```diff
@@ -97,17 +97,17 @@ type PrefixedTokenIter struct {
 	buf       []byte
 	prefixLen int
 
-	NGramTokenIter
+	Iterator[[]byte]
```
chaudum, line 100: I wouldn't make this field exported.

chaudum (commented):
This is looking really good!

pkg/storage/bloom/v1/tokenizer.go
salvacorts: Done

pkg/storage/bloom/v1/tokenizer.go
salvacorts: Done

pkg/storage/bloom/v1/bloom_tester_test.go
salvacorts: Indeed, leftover from before rebasing

pkg/storage/bloom/v1/bloom_tester.go
salvacorts: Done

pkg/storage/bloom/v1/tokenizer.go
salvacorts: it's indirectly  exported. We need have that way so we don't need to reimplement Next, Err...

pkg/storage/bloom/v1/bloom_tester_test.go
salvacorts: Yes, but w/o much utility other than testing

pkg/storage/bloom/v1/tokenizer.go
salvacorts: I don't have a stong opinion on this, i'm fine with making this private and implementing Err and Next as wrappers. But I think this is more idiomatic

owen-d (approved):

PR #12036, opened 2024-02-22
chore(blooms): Various fixes in blooms read path

chaudum:

* Wire up bloom querier metrics to fix nil pointer panic
* Use `bloomutils.TokenRangesForInstance` instead of `subRing.GetTokenRangesForInstance` to avoid `zone not set` error

### DIFF
--- a/pkg/bloomgateway/querier.go
+++ b/pkg/bloomgateway/querier.go
@@ -61,7 +61,7 @@ type BloomQuerier struct {
 func NewQuerier(c Client, r prometheus.Registerer, logger log.Logger) *BloomQuerier {
 	return &BloomQuerier{
 		c:       c,
-		metrics: newQuerierMetrics(r, constants.Loki, "bloom_gateway_querier"),
+		metrics: newQuerierMetrics(r, constants.Loki, querierMetricsSubsystem),
 		logger:  logger,
 	}
 }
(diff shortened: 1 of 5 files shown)


pkg/bloomgateway/querier.go
```diff
@@ -57,8 +58,12 @@ type BloomQuerier struct {
 	metrics *querierMetrics
 }
 
-func NewQuerier(c Client, logger log.Logger) *BloomQuerier {
-	return &BloomQuerier{c: c, logger: logger}
+func NewQuerier(c Client, r prometheus.Registerer, logger log.Logger) *BloomQuerier {
+	return &BloomQuerier{
+		c:       c,
+		metrics: newQuerierMetrics(r, constants.Loki, "bloom_gateway_querier"),
```
salvacorts, line 64: Add string as constant?

salvacorts (approved):

pkg/bloomgateway/querier.go
chaudum: Can do, but it's really only used here.

chaudum: https://github.com/grafana/loki/pull/12036/commits/99277d2c49bb7fff9b4600318802820690597dd5

PR #12048, opened 2024-02-23
chore(blooms): Add more instrumentation to bloom gateway and bloom store

chaudum:

See individual commit messages.

### DIFF
--- a/pkg/bloomgateway/bloomgateway_test.go
+++ b/pkg/bloomgateway/bloomgateway_test.go
@@ -77,7 +77,8 @@ func setupBloomStore(t *testing.T) *bloomshipper.BloomStore {
 		},
 	}
 
-	store, err := bloomshipper.NewBloomStore(schemaCfg.Configs, storageCfg, cm, nil, nil, logger)
+	reg := prometheus.NewPedanticRegistry()
+	store, err := bloomshipper.NewBloomStore(schemaCfg.Configs, storageCfg, cm, nil, nil, reg, logger)
 	require.NoError(t, err)
 	t.Cleanup(store.Stop)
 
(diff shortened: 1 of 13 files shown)


pkg/bloomgateway/bloomgateway_test.go
```diff
@@ -77,7 +77,8 @@ func setupBloomStore(t *testing.T) *bloomshipper.BloomStore {
 		},
 	}
 
-	store, err := bloomshipper.NewBloomStore(schemaCfg.Configs, storageCfg, cm, nil, nil, logger)
+	reg := prometheus.NewPedanticRegistry()
```
salvacorts, line 80: Why Pedantic registry instead of the usual default registry?

pkg/storage/stores/shipper/bloomshipper/cache.go
```diff
@@ -133,6 +130,12 @@ func (b BlockDirectory) Release() error {
 	return nil
 }
 
+func (b BlockDirectory) Size() int64 {
+	bloomFileStats, _ := os.Lstat(path.Join(b.Path, v1.BloomFileName))
```
salvacorts, line 134: IIUC, the size won't change since we are not modifying the block. Maybe we can do the Lstat on `NewBlockDirectory` save the size inside the  `BlockDirectory` struct? That way we can save a bunch of sys-calls.

salvacorts (commented):
LGTM. Left two minor comments.

pkg/bloomgateway/bloomgateway_test.go
chaudum: There's no reason for using the PedanticRegistry, which, according to docs, is useful for testing implementation of collectors and/or metrics, but we don't do that here.

salvacorts (approved):

PR #12050, opened 2024-02-23
fix(blooms): fix reference leak and resulting race condition in BloomPageDecoder

owen-d:
Removes the `Drop()` function for now and adjusts some pool usage according to the following:

* Minor refactoring
* Reference leak in `BloomPageDecoder`
* Race condition due to the above


slim-bean (approved):
LGTM

PR #12053, opened 2024-02-24
fix(blooms): closes chunk iterator

owen-d:
Ensures we close chunks during bloom compaction after they've been exhausted.


slim-bean (approved):
close those iterators 🥇

PR #12058, opened 2024-02-26
Track and log congestion control latency

dannykopping:
https://github.com/grafana/loki/pull/10140 introduced a congestion control mechanism for GCS. Currently we're missing a per-request latency measure to see how long a query was blocked on congestion control.


slim-bean (approved):
LGTM

PR #12065, opened 2024-02-27
feat(operator): Update categories list for OperatorHub

cahartma:
Implements https://issues.redhat.com/browse/LOG-4691
Adds Loki Operator to the Observability list in the OperatorHub


xperimental (changes requested):
It looks as if this would cause release issues (merged this change in another operator already which now fails a validation)

I'm marking this as "request changes" so it does not get accidentally merged.

periklis:
Closing this for now until we have confirmation that downstream validations work as expected.

PR #12096, opened 2024-02-29
feat: Bloom filter regexes

salvacorts:
This is a followup PR for https://github.com/grafana/loki/pull/12035.

In this PR we:
- Extract the logic to simplify regexes from the log pkg into a new interface
- Use the extracted regex simplifier in the bloom tester to try to simplify regexes into simpler tests. If a regex cannot be simplified, we return the MatchAll bloom test.

Filtering in the `Filterer` interface works by checking if an input string matches what a given filter has. E.g. passing the string `foobar` to the `contains` filter which has `foo` inside will check if `foo` is within `foobar`. The `BloomTest` works the other way around: the test has the line to check, whereas the actual content in the the bloom. E.g. the bloom test will contain the line to match `foo` and the bloom filter has `foobar` inside. In other words, the `Filterer` and the `BloomTester` interface works in opposite ways.  That's why we need to extend the `Filterer` interface to have a new `Matches`  method that takes a new `Checker` interface with Test and TestRegex.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/log/filter.go
+++ b/pkg/logql/log/filter.go
@@ -31,6 +31,9 @@ type trueFilter struct{}
 func (trueFilter) Filter(_ []byte) bool { return true }
 func (trueFilter) ToStage() Stage       { return NoopStage }
 
+// Matches implements MatcherFilter
+func (trueFilter) Matches(_ Checker) bool { return true }
+
 // TrueFilter is a filter that returns and matches all log lines whatever their content.
 var TrueFilter = trueFilter{}
 
@@ -47,6 +50,9 @@ func (e existsFilter) ToStage() Stage {
 		},
 	}
 }
+
+// Matches implements MatcherFilter
+func (e existsFilter) Matches(_ Checker) bool { return true }
 
 // ExistsFilter is a filter that returns and matches when a line has any characters.
 var ExistsFilter = existsFilter{}
@@ -189,12 +195,12 @@ func (a andFilters) ToStage() Stage {
 }
 
 type orFilter struct {
-	left  Filterer
-	right Filterer
+	left  MatcherFilter
+	right MatcherFilter
 }
 
 // newOrFilter creates a new filter which matches only if left or right matches.
-func newOrFilter(left Filterer, right Filterer) Filterer {
+func newOrFilter(left MatcherFilter, right MatcherFilter) MatcherFilter {
 	if left == nil || left == TrueFilter {
 		return right
 	}
@@ -210,7 +216,7 @@ func newOrFilter(left Filterer, right Filterer) Filterer {
 }
 
 // ChainOrFilter is a syntax sugar to chain multiple `or` filters. (1 or many)
-func ChainOrFilter(curr, new Filterer) Filterer {
+func ChainOrFilter(curr, new MatcherFilter) MatcherFilter {
 	if curr == nil {
 		return new
 	}
@@ -227,6 +233,11 @@ func (a orFilter) ToStage() Stage {
 			return line, a.Filter(line)
 		},
 	}
+}
+
+// Matches implements MatcherFilter
+func (a orFilter) Matches(test Checker) bool {
+	return a.left.Matches(test) || a.right.Matches(test)
 }
 
 type regexpFilter struct {
@@ -286,11 +297,16 @@ func (l equalFilter) ToStage() Stage {
 	}
 }
 
+// Matches implements MatcherFilter
+func (l equalFilter) Matches(test Checker) bool {
+	return test.Test(l.match)
+}
+
 func (l equalFilter) String() string {
 	return string(l.match)
 }
 
-func newEqualFilter(match []byte, caseInsensitive bool) Filterer {
+func newEqualFilter(match []byte, caseInsensitive bool) MatcherFilter {
 	return equalFilter{match, caseInsensitive}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/storage/bloom/v1/bloom_tester.go
```diff
@@ -120,6 +191,33 @@ func (b stringTest) MatchesWithPrefixBuf(bloom filter.Checker, buf []byte, prefi
...
+
+// Filter implements the log.Filterer interface
+func (b stringMatcherFilter) Filter(_ []byte) bool {
+	panic("implement me")
+}
+
+// ToStage implements the log.Filterer interface
+func (b stringMatcherFilter) ToStage() log.Stage {
+	panic("implement me")
+}
```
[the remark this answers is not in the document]
salvacorts: The simplifier needs to be able to return both types `Matcher` and `Filterer` so it needs to know how to create a type that implements both. In this case, since we will only be using the `Matcher` path, I think it's safe to just satisfy the `Filterer` interface with these panic impl.

pkg/logql/log/filter.go
```diff
@@ -6,17 +6,22 @@ import (
...
 
+type Checker interface {
+	Test(line []byte, caseInsensitive bool, equal bool) bool
+	TestRegex(reg *regexp.Regexp) bool
+}
+
 // Filterer is a interface to filter log lines.
 type Filterer interface {
 	Filter(line []byte) bool
+	Matches(test Checker) bool
```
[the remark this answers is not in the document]
salvacorts: Done, extracted to a separate interface, and created a new `MatcherFilterer` iface that combines both. The `Simplify` method returns `MatcherFilterer`, but we restrict the type returned by `NewFilter` and `NewLabelFilter` to `Filterer` which is what the rest of the read-path code uses.

pkg/logql/log/filter.go
```diff
@@ -491,22 +570,24 @@ func simplify(reg *syntax.Regexp, isLabel bool) (Filterer, bool) {
 		}
 	case syntax.OpEmptyMatch:
 		return TrueFilter, true
+	default:
+		return nil, false
 	}
 	return nil, false
```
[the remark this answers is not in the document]
salvacorts: I added this because of my IDE linter complaining. I agree it's useless, I'll remove it.

pkg/logql/log/filter.go
```diff
@@ -112,18 +161,21 @@ func (a andFilter) ToStage() Stage {
...
+}
+
 type andFilters struct {
-	filters []Filterer
+	filters []MatcherFilterer
 }
 
-// NewAndFilters creates a new filter which matches only if all filters match
-func NewAndFilters(filters []Filterer) Filterer {
+func NewAndMatcherFilters(filters []MatcherFilterer) MatcherFilterer {
```
chaudum, line 172: Is this anywhere used except in `NewAndFilters`, which returns `Filterer`?

pkg/logql/log/filter.go
```diff
@@ -32,9 +53,30 @@ type trueFilter struct{}
 func (trueFilter) Filter(_ []byte) bool { return true }
 func (trueFilter) ToStage() Stage       { return NoopStage }
 
+// Matches implements MatcherFilter
```
chaudum, line 56:
```suggestion
// Matches implements MatcherFilterer
```

pkg/logql/log/filter.go
```diff
@@ -6,20 +6,41 @@ import (
...
+	Filterer
+}
+
+type matcherFiltererWrapper struct {
+	Filterer
+}
+
+func (m matcherFiltererWrapper) Matches(_ Checker) bool {
+	panic("unimplemented")
+}
```
chaudum, line 42: nit: instead of an "unidirectional" wrapper, we could make a "bidirectional" wrapper, for both `Filterer` and `Matcher`.

That could then also be used in the bloom package, right?

```go

type wrapper struct {
	Filterer
	Matcher
}

func (w wrapper) IsMatcher() bool {
	return a.Matcher != nil
}

func (w wrapper) IsFilterer() bool {
	return a.Filterer != nil
}

func WrapMatcher(f Filterer) MatcherFilterer {
	return adapter{Filterer: f}
}

func WrapFilterer(m Matcher) MatcherFilterer {
	return adapter{Matcher: m}
}
```

chaudum (commented):
As discussed offline, we both have a feeling that there might be a more elegant/clean way, but that would also imply way more changes.

We are kinda happy with the current approach: It hides the `Matches` method from whatever that works with `Filterer` (which is all the rest of the `log` pkg and `ast`).

The changes are mostly minimal, which is good given how critical that code is in both perf and correctness.

Another thing that is great is that the regex simplifier is able to work in both testing/filtering directions.

> In addition to that, perfection is enemy of good.

pkg/logql/log/filter.go
salvacorts: Done! Thanks for the tip :)

pkg/logql/log/filter.go
salvacorts: Done

pkg/logql/log/filter.go
salvacorts: Nope, removed

pkg/logql/log/filter.go
```diff
@@ -6,20 +6,41 @@ import (
...
+	"github.com/grafana/loki/pkg/util"
 	"github.com/prometheus/prometheus/model/labels"
 )
 
+type Checker interface {
+	Test(line []byte, caseInsensitive bool, equal bool) bool
+	TestRegex(reg *regexp.Regexp) bool
+}
+
+type Matcher interface {
```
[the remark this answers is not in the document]
salvacorts: Done

pkg/logql/log/filter.go
```diff
@@ -6,20 +6,41 @@ import (
...
-	"github.com/grafana/loki/pkg/util"
-
 	"github.com/grafana/regexp"
 	"github.com/grafana/regexp/syntax"
 
+	"github.com/grafana/loki/pkg/util"
 	"github.com/prometheus/prometheus/model/labels"
 )
 
+type Checker interface {
```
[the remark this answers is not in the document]
salvacorts: Done

chaudum (approved):

PR #12097, opened 2024-02-29
fix(operator): Fix encoding of blocked query pattern in configuration

btaani:
Prior to this PR, when LokiStack is configured to block queries with a regex pattern, user could still query logs even when they satisfy the regex pattern. This PR fixes the way that the function `yamlBlock` handles the pattern so that regex works.

Fixes [LOG-4927](https://issues.redhat.com/browse/LOG-4927)

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/internal/manifests/internal/config/loki-runtime-config.yaml
```diff
@@ -58,8 +58,7 @@ overrides:
     blocked_queries:
     {{- range $blockedQuery := . }}
     - {{ with $blockedQuery.Pattern -}}
-      pattern: |
-        {{ . | yamlBlock "        "  }}
+      pattern: {{ . | yamlBlock "        "  }}
```
xperimental, line 61: Thinking about the "let's just use a string instead of a YAML multi-line block" topic again, I think we can just replace this with the "quoted Go" `sprintf` output and remove the custom function:

```suggestion
      pattern: {{ . | printf "%q" }}
```

operator/internal/manifests/internal/config/loki-runtime-config.yaml
```diff
@@ -58,8 +58,7 @@ overrides:
     blocked_queries:
     {{- range $blockedQuery := . }}
     - {{ with $blockedQuery.Pattern -}}
-      pattern: |
-        {{ . | yamlBlock "        "  }}
+      pattern: {{ . | printf "%q" }}
```
xperimental, line 61: (can't attach a suggestion to the whole block)

Looking at this configuration block again, I noticed that this is another one of those blocks where every attributes is optional (even though at least one should be present in reality). I wonder if we should make this more robust by rendering at least one attribute in all cases, so that YAML parse errors should not occur. The only sensible choice for me seems to be the `regex` field, because it can be `false` without affecting anything.

@periklis Do you know if Loki treats blocked queries having `regex=false` differently from having `regex` unset? If they are treated the same, I would propose this improvement:

```
    - regex: {{ $blockedQuery.Regex }}
      {{ with $blockedQuery.Pattern -}}
        pattern: {{ . | printf "%q" }}
      {{ end -}}
      {{- with $blockedQuery.Types }}
      types: {{ . }}
      {{- end }}
      {{- with $blockedQuery.Hash }}
      hash: {{ . }}
      {{- end }}
```

xperimental (commented):
Seems to work fine during tests. One thing occurred to me, that we might be able to fix in-passing while we are at this spot (see comment).

operator/internal/manifests/internal/config/loki-runtime-config.yaml
periklis: AFAICS there is no notion on unset regex in the Loki code base: https://github.com/grafana/loki/blob/main/pkg/logql/blocker.go#L55-L66

I agree we should always set at least a single field, but the best option is `pattern: ''` because setting `regex:false` by default will be translated as `pattern:"" regex: true` in the Loki code base. WDYT?

xperimental: During the daily we came to the conclusion, that it probably does not make a difference _technically_ if we use `regex` or `pattern`, but the empty pattern is more obvious to the casual onlooker, so we settled on using that property instead :slightly_smiling_face:

operator/internal/manifests/internal/config/loki-runtime-config.yaml
```diff
@@ -57,13 +57,10 @@ overrides:
     {{- with $l.Blocked }}
     blocked_queries:
     {{- range $blockedQuery := . }}
-    - {{ with $blockedQuery.Pattern -}}
-      pattern: |
-        {{ . | yamlBlock "        "  }}
+    - pattern: {{ $blockedQuery.Pattern }}
+      {{ with $blockedQuery.Regex -}}
+        pattern: {{ . | printf "%q" }}
```
JoaoBraveCoding, line 62: If I'm not mistaken, I think here is regex instead of `pattern` 
```suggestion
    - pattern: {{ $blockedQuery.Pattern }}
      {{ with $blockedQuery.Regex -}}
       regex: {{ . | printf "%q" }}
```

xperimental: Yes, `pattern` is duplicated now, the template also needs a change and there's a slight change to the "zero-length behaviour" of the templates:

```suggestion
    - pattern: {{ $blockedQuery.Pattern }}
      {{- with $blockedQuery.Regex }}
      regex: {{ . }}
```

Don't forget to also change to `{{- end }}` on the next line (not part of the highlight, so can't adjust the suggesstion).

btaani: You're right. I quickly pushed this change yesterday without even testing and now the results are messed up ^^"
It's also not what you suggested though.. There are some use cases when neither `pattern` nor `regex` are used. So I'll update the template to take this into consideration..

xperimental: > It's also not what you suggested though..

Sorry for the confusion. Usually I just provide a code suggestion to avoid misunderstandings.

xperimental (approved):
Just tested with a few blocks, seems to work fine :+1:

PR #12102, opened 2024-03-01
fix: report correct status code for metric and log queries in metrics.go

ashwanthgoli:
`metrics.go` which reports query stats shows incorrect status code for metric and log queries. It is incorrectly reporting successful queries with `500` status code.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/logql/engine.go
+++ b/pkg/logql/engine.go
@@ -256,7 +256,10 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
 	statResult := statsCtx.Result(time.Since(start), queueTime, q.resultLength(data))
 	statResult.Log(level.Debug(spLogger))
 
-	status, _ := server.ClientHTTPStatusAndError(err)
+	status := 200
+	if err != nil {
+		status, _ = server.ClientHTTPStatusAndError(err)
+	}
 
 	if q.record {
 		RecordRangeAndInstantQueryMetrics(ctx, q.logger, q.params, strconv.Itoa(status), statResult, data)


sandeepsukhani (approved):

pkg/logql/engine.go
```diff
@@ -256,7 +256,10 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
 	statResult := statsCtx.Result(time.Since(start), queueTime, q.resultLength(data))
 	statResult.Log(level.Debug(spLogger))
 
-	status, _ := server.ClientHTTPStatusAndError(err)
+	status := 200
+	if err != nil {
+		status, _ = server.ClientHTTPStatusAndError(err)
```
na--, line 261: Hmm does it make sense to add the `err != nil` check in `ClientHTTPStatusAndError()` and return 200 there?

Also, this would benefit from a unit test and I'm not sure how easy it would be to do it here (because of the direct call to `RecordRangeAndInstantQueryMetrics()` that I am not sure we can mock), while doing it in the  `ClientHTTPStatusAndError()` seems much easier.

sandeepsukhani: yeah, it makes sense to add a check there.

cstyan (commented):
@ashwanthgoli bump, I agree with Ned and Sandeep, lets see if moving the check to the server function and be `if err == nil`, and then just confirm that the rest of the tests are still green

PR #12106, opened 2024-03-01
feat(operator): Allow setting explicit CredentialMode in LokiStack storage spec

xperimental:

Currently it is not possible to run the operator in managed mode and then deploy a LokiStack using a different mode for credentials to the object storage (static credentials or non-managed tokens). This is due to the credential-mode being determined only by the environment of the operator and the provided secret, making it an implicit consequence of the configuration instead of an explicit configuration.

This PR adds an optional attribute to the LokiStack's spec that allows the user to override the default credential mode. The main use-case for this is running a non-managed credential with an operator running in a managed-credentials OpenShift cluster, but it can also be used make the credentials-mode explicit for other deployments instead of having the operator decide based on the provided secret.

Fixes [LOG-5105](https://issues.redhat.com/browse/LOG-5105)

- `CredentialMode` is an enum with just three possible values, but it is _optional_, so `""` is a possible value _inside the operator_ as well, although this can not be set on the `LokiStack` resource.
- The `CredentialsRequest` is only created when running in managed-mode. It is also removed when the mode changes and it had been created before.
- [x] I still need to properly test this on Azure & GCP, the last tests were only on AWS.

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/operator/controllers/loki/lokistack_controller.go
+++ b/operator/controllers/loki/lokistack_controller.go
@@ -180,7 +180,7 @@ func (r *LokiStackReconciler) updateResources(ctx context.Context, req ctrl.Requ
 	}
 
 	if r.FeatureGates.OpenShift.ManagedAuthEnv {
-		if err := handlers.UpdateCredentialsRequest(ctx, r.Log, r.Scheme, r.AuthConfig, r.Client, req); err != nil {
+		if err := handlers.CreateUpdateDeleteCredentialsRequest(ctx, r.Log, r.Scheme, r.AuthConfig, r.Client, req); err != nil {
 			return "", err
 		}
 	}
--- a/operator/controllers/loki/lokistack_controller.go
+++ b/operator/controllers/loki/lokistack_controller.go
@@ -180,7 +180,7 @@ func (r *LokiStackReconciler) updateResources(ctx context.Context, req ctrl.Requ
 	}
 
 	if r.FeatureGates.OpenShift.ManagedAuthEnv {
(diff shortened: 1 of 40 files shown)


operator/internal/manifests/storage/options.go
periklis, line 1: Since we use `Options.CredentialsMode` in `storage/configure.go` to determine which mounts/volumes to consider we can probably remove `S3.STS`, `GCS.WorkloadIdentity` and `Azure.WorkloadIdentity`. WDYT?

operator/controllers/loki/lokistack_controller.go
```diff
@@ -180,7 +180,7 @@ func (r *LokiStackReconciler) updateResources(ctx context.Context, req ctrl.Requ
 	}
 
 	if r.FeatureGates.OpenShift.ManagedAuthEnv {
-		if err := handlers.CreateCredentialsRequest(ctx, r.Log, r.Scheme, r.AuthConfig, r.Client, req); err != nil {
+		if err := handlers.UpdateCredentialsRequest(ctx, r.Log, r.Scheme, r.AuthConfig, r.Client, req); err != nil {
```
periklis, line 183: Suggest to rename this to an `CreateUpdateDeleteCredentialsRequest` handler.

periklis (approved):
Looks overall good to me and works fine on an AWS STS managed cluster with an in-cluster Minio installation. Two cleanup suggestions added below, but approved to unblock you when done.

xperimental:
Tested the new code on Azure and GCP as well. Seems to work fine :+1:

periklis (approved):

JoaoBraveCoding (approved):
lgtm

PR #12124, opened 2024-03-04
fix: explicitly serialize and deserialize noop label filters

MasslessParticle:
Noop filters were being ignored by the serialization code. This resulted in `nil` label filters in some circumstances which caused panics on the query path. This PR introduces explicit serialization and deserialization of noop filters. A noop filter always has an underlying matcher so it can be treated the same way as a string filter.


pkg/logql/syntax/serialize_test.go
```diff
@@ -53,6 +53,9 @@ func TestJSONSerializationRoundTrip(t *testing.T) {
 		"multiple post filters": {
 			query: `rate({app="foo"} | json | unwrap foo | latency >= 250ms or bytes > 42B or ( status_code < 500 and status_code > 200) or source = ip("") and user = "me" [1m])`,
 		},
+		"multiple post filters where one is a noop": {
```
MasslessParticle, line 56: before this fix, this test panics

chaudum (commented):
:heavy_check_mark:  LGTM. Does this fix require a changelog entry / is the bug in any publicly released version?

MasslessParticle:
@chaudum I've added a changelog entry. This isn't in a released version, yet

chaudum (approved):

PR #12151, opened 2024-03-06
perf(blooms): removes pendingTasks impl with an atomic.Int64

owen-d:
* pendingTasks only checks the length of outstanding reqs, so instead replace it with a simpler & performant `atomic.Int64`
* Also adds some comments/todos


cyriltovena (approved):
Lgtm

PR #12135, opened 2024-03-06
refactor(bloom-gw): todos and comments

owen-d:
Some TODOs/general comments outlining things that caught my eye in the bloom-gw.

PR #12156, opened 2024-03-07
Check chunk start/end (as well as checksum) on removeNotMatchingChunks

salvacorts:
This PR addresses two TODOs from https://github.com/grafana/loki/pull/12135 regarding potential bugs in `removeNotMatchingChunks`.

We now check the time interval of the chunks, as well as the checksum so responses marking for deletion chunks with same checksum but different time range will only remove the chunks matching (previously we'd have removed all the chunks matching the checksum regardless of the time-range).

On top of that we now use `slices.BinarySearchFunc` which returns whether the FP was actually found instead of `sort.Search` which returns the index where the looked up target would be places regardless of whether it's actually there. 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


owen-d:
I tried to address this more comprehensively in https://github.com/grafana/loki/pull/12162

PR #12158, opened 2024-03-07
chore: Add summary for received filters in bloom gateway

salvacorts:
This PR addresses a TODO from https://github.com/grafana/loki/pull/12135 to add a metric to track the number of filters per request.

The comment suggests adding it to the query frontend but:
- I think this is mainly useful on the blooms path
- This will allow us to see if any of the gateways is receiving more filters than the rest.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


owen-d (commented):
We probably want this in two places:
1) in the query-frontend (as I mentioned in the comment) that logs 1:1 with the original query. This gives one observation per query and ensures we measure queries that have zero filters which are never passed to the bloom-gws.
2) in the bloom-gws that measure an observation once per exectuion (post split+shard in the query frontend). 

Also, we should use histograms and not summaries, which gives us aggregation support. In general, histograms are what you want instead of summaries in most cases. [context](https://prometheus.io/docs/practices/histograms/#:~:text=The%20essential%20difference%20between%20summaries,the%20server%20side%20using%20the)

pkg/bloomgateway/metrics.go
```diff
@@ -66,6 +67,13 @@ func newServerMetrics(registerer prometheus.Registerer, namespace, subsystem str
 			Help:      "Total amount of chunk refs filtered by bloom-gateway",
 			Buckets:   prometheus.ExponentialBucketsRange(1, 100e3, 10),
 		}),
+		receivedFilters: promauto.With(registerer).NewHistogram(prometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "request_filters",
+			Help:      "Number of filters per request.",
+			Buckets:   prometheus.LinearBuckets(0, 1, 20),
```
owen-d, line 75: I don't think we'll need 20 1-width buckets. I've suggested an exponential increasing range up to 256
```suggestion
			// 1->256
			Buckets:   prometheus.ExponentialBuckets(1, 2, 8),
```

pkg/querier/queryrange/metrics.go
```diff
@@ -35,6 +40,57 @@ func NewMetrics(registerer prometheus.Registerer, metricsNamespace string) *Metr
...
+	receivedFilters prometheus.Histogram
+}
+
+func NewMiddlewareQueryMetrics(registerer prometheus.Registerer, metricsNamespace string) *QueryMetrics {
+	return &QueryMetrics{
+		receivedFilters: promauto.With(registerer).NewHistogram(prometheus.HistogramOpts{
+			Namespace: metricsNamespace,
+			Name:      "query_frontend_query_filters",
+			Help:      "Number of filters per query.",
+			Buckets:   prometheus.LinearBuckets(0, 1, 20),
```
owen-d, line 58: Use the same buckets here

owen-d (approved):
approving, but please update the buckets. Thanks!

pkg/bloomgateway/metrics.go
salvacorts: Done

pkg/querier/queryrange/metrics.go
salvacorts: Done

PR #12160, opened 2024-03-07
refactor(bloomstore): Introduce fetch option for blocks fetcher

chaudum:

Requests to the bloom gateway should not wait for blocks to be downloaded into cache first, but should operate on readily available blocks.

This PR introduces fetch options for the bloom store that specify how the store should behave when requesting blocks from block refs.

`WithIgnoreMissing`: Ignore errors from blocks that could not be found in object storage, and instead return a `nil` value in the response.
`WithFetchAsync`: Return only the blocks that are available locally in cache and dispatch downloading of blocks to the blocks downloader, that operates in the background, and make them available for subsequent requests.

### DIFF
--- a/pkg/storage/stores/shipper/bloomshipper/fetcher.go
+++ b/pkg/storage/stores/shipper/bloomshipper/fetcher.go
@@ -175,8 +175,7 @@ func (f *Fetcher) FetchBlocks(ctx context.Context, refs []BlockRef, opts ...Fetc
 	cfg := &options{ignoreNotFound: true, fetchAsync: false}
 	cfg.apply(opts...)
 
-	// first, resolve blocks from cache or from filesystem
-	// enqueue missing blocks to download queue
+	// first, resolve blocks from cache and enqueue missing blocks to download queue
 	n := len(refs)
 	results := make([]*CloseableBlockQuerier, n)
 	responses := make(chan downloadResponse[BlockDirectory], n)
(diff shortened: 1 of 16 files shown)


pkg/storage/stores/shipper/bloomshipper/fetcher.go
```diff
@@ -182,53 +248,70 @@ func (f *Fetcher) processTask(ctx context.Context, task downloadRequest[BlockRef
...
 	}
 
 	// item found in cache
 	if len(fromCache) == 1 {
 		f.metrics.blocksFetchedSize.WithLabelValues(sourceCache).Observe(float64(fromCache[0].Size()))
-		return fromCache[0], nil
+		return fromCache[0], true, nil
 	}
 
-	fromLocalFS, _, err := f.loadBlocksFromFS(ctx, []BlockRef{ref})
```
owen-d, line 222: Do we no longer load blocks from the FS _at all_? Is this an oversight? It looks like we've removed the `fs` from the load path in all cases

I think querying bloom-gws should query against all local blocks (in mem or on disk)

owen-d (approved):
approving, but take a look at my comments about the fs loading

pkg/storage/stores/shipper/bloomshipper/fetcher.go
chaudum: Since we load blocks into cache on startup ([cache.go](https://github.com/grafana/loki/blob/main/pkg/storage/stores/shipper/bloomshipper/cache.go#L55-L59)) we don't need to look at the filesystem at query/runtime.

PR #12169, opened 2024-03-11
feat(bloom-gw): Make num multiplexing tasks configurable

chaudum:

We want to be able to change the amount of dequeued items per iterations to test different configurations.


pkg/storage/stores/shipper/bloomshipper/fetcher.go
```diff
@@ -77,7 +78,7 @@ func NewFetcher(cfg bloomStoreConfig, client Client, metasCache cache.Cache, blo
 		metrics:         newFetcherMetrics(reg, constants.Loki, "bloom_store"),
 		logger:          logger,
 	}
-	q, err := newDownloadQueue[BlockRef, BlockDirectory](1000, cfg.numWorkers, fetcher.processTask, logger)
+	q, err := newDownloadQueue[BlockRef, BlockDirectory](100000, cfg.numWorkers, fetcher.processTask, logger)
```
chaudum, line 81: Should this be configurable?

pkg/storage/bloom/v1/bloom.go
```diff
@@ -239,6 +247,16 @@ func (b *BloomBlock) BloomPageDecoder(r io.ReadSeeker, pageIdx int) (*BloomPageD
 
 	page := b.pageHeaders[pageIdx]
 
+	if page.Len > maxPageSize {
+		// TODO(chaudum): Remove print statement or replace it with a log message.
+		fmt.Fprintf(
```
owen-d, line 252: this should use a logger or bubble up to where a logger is present which we can use.

pkg/storage/bloom/v1/bloom.go
```diff
@@ -12,6 +13,13 @@ import (
 	"github.com/grafana/loki/pkg/util/encoding"
 )
 
+// NB(chaudum): Some block pages are way bigger than others (400MiB and
+// bigger), and loading multiple pages into memory in parallel can cause the
+// gateways to OOM.
+// Figure out a decent maximum page size that we can process.
+var maxPageSize = 32 << 20 // 32MB
```
owen-d, line 20: I think this would be a good thing to keep configurable

pkg/storage/stores/shipper/bloomshipper/fetcher.go
owen-d: I'm not opinionated on this one.

pkg/storage/stores/shipper/bloomshipper/metrics.go
```diff
@@ -56,5 +61,31 @@ func newFetcherMetrics(registerer prometheus.Registerer, namespace, subsystem st
...
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "downloadqueue_enqueue_time_seconds",
+			Buckets:   prometheus.ExponentialBuckets(0.0001, 5, 8), // [0.0001, 0.0005, ... 7.8125]
+			Help:      "Time in seconds it took to enqueue item to download queue",
+		}),
+		downloadQueueSize: r.NewHistogram(prometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "downloadqueue_size",
```
owen-d, line 74:
```suggestion
			Name:      "download_queue_size",
```

owen-d (approved):
I suspect the queue length issue is actually due to two things:
1) The same item can be present in the queue multiple times (probably should not be the case)
2) The key is locked while downloading in
```go
func (q *downloadQueue[T, R]) do(ctx context.Context, task downloadRequest[T, R]) {
	q.mu.LockKey(task.key)
	defer func() {
		err := q.mu.UnlockKey(task.key)
		if err != nil {
			level.Error(q.logger).Log("msg", "failed to unlock key in block lock", "key", task.key, "err", err)
		}
	}()

	q.process(ctx, task)
}
```

If the same keys are regularly requested, workers will both redownload the same blocks multiple times and sit idle waiting to acquire a lock to do duplicate work

We should enable mutex & block profiling to get a better sense for these:
```
runtime.SetBlockProfileRate()
runtime.SetMutexProfileFraction()
```

chaudum:
> I suspect the queue length issue is actually due to two things:
> 
>     1. The same item can be present in the queue multiple times (probably should not be the case)

Yes, that is the case. The reason is that multiple worker can request the same blocks in parallel. Deduplicating the work is done when dequeueing, as I wanted to avoid synchronisation at enqueue time.

> 
>     2. The key is locked while downloading in
> 
> 
> ```go
> func (q *downloadQueue[T, R]) do(ctx context.Context, task downloadRequest[T, R]) {
> 	q.mu.LockKey(task.key)
> 	defer func() {
> 		err := q.mu.UnlockKey(task.key)
> 		if err != nil {
> 			level.Error(q.logger).Log("msg", "failed to unlock key in block lock", "key", task.key, "err", err)
> 		}
> 	}()
> 
> 	q.process(ctx, task)
> }
> ```
> 
> If the same keys are regularly requested, workers will both redownload the same blocks multiple times and sit idle waiting to acquire a lock to do duplicate work

Blocks are not downloaded multiple times, even when there are multiple requests for the same block, since after acquiring the lock for the block key, we check if the item is already in cache. This means that the first goroutine will download the block, the others will get it from cache.

https://github.com/grafana/loki/blob/main/pkg/storage/stores/shipper/bloomshipper/fetcher.go#L251-L267

However, you're right that other worker goroutines that want to download the same block sit idle waiting.

They could immediately discard their work and dequeue the next, but that would not work in conjunction with the fetch option `WithFetchAsync(false)`.


> We should enable mutex & block profiling to get a better sense for these:
> 
> ```
> runtime.SetBlockProfileRate()
> runtime.SetMutexProfileFraction()
> ```

We definitely should.

PR #12180, opened 2024-03-12
feat: Make list of otel blessed attributes configurable

sandeepsukhani:
We have a list of resource attributes which are, by default, picked up index labels. The list is hardcoded which makes it inconvenient to update the list. This PR makes the list configurable.

**Checklist**
- [x] Tests updated

### DIFF
--- a/docs/sources/send-data/otel/_index.md
+++ b/docs/sources/send-data/otel/_index.md
@@ -124,8 +124,9 @@ otlp_config:
   # Configuration for Resource Attributes to store them as index labels or
   # Structured Metadata or drop them altogether
   resource_attributes:
-    # Configure whether to ignore the default list of Resource Attributes to be
-    # stored as Index Labels and only use the given Resource Attributes config
+    # Configure whether to ignore the default list of resource attributes set in
+    # 'distributor.otlp.default_resource_attributes_as_index_labels' to be
+    # stored as index labels and only use the given resource attributes config
     [ignore_defaults: <boolean>]
 
     [attributes_config: <list of attributes_configs>]
--- a/pkg/loghttp/push/otlp.go
+++ b/pkg/loghttp/push/otlp.go
@@ -28,14 +28,6 @@ const (
 	attrServiceName     = "service.name"
 )
 
-var blessedAttributesNormalized = make([]string, len(blessedAttributes))
-
-func init() {
-	for i := range blessedAttributes {
-		blessedAttributesNormalized[i] = prometheustranslator.NormalizeLabel(blessedAttributes[i])
-	}
-}
-
 func newPushStats() *Stats {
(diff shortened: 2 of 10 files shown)


JStickler (commented):
[Docs team]  Docs LGTM

pkg/loghttp/push/otlp.go
```diff
@@ -118,7 +110,7 @@ func otlpToLokiPushRequest(ld plog.Logs, userID string, tenantsRetention Tenants
 			resAttrs.PutStr(attrServiceName, "unknown_service")
 		}
 		resourceAttributesAsStructuredMetadata := make(push.LabelsAdapter, 0, resAttrs.Len())
-		streamLabels := make(model.LabelSet, len(blessedAttributesNormalized))
+		streamLabels := make(model.LabelSet, 30) // we have a default labels limit of 30 so just initialize the map of same size
```
salvacorts, line 113: nit: can we get this number (30) from the defaults somehow?

docs/sources/send-data/otel/_index.md
```diff
@@ -124,8 +124,9 @@ otlp_config:
   # Configuration for Resource Attributes to store them as index labels or
   # Structured Metadata or drop them altogether
   resource_attributes:
-    # Configure whether to ignore the default list of Resource Attributes to be
-    # stored as Index Labels and only use the given Resource Attributes config
+    # Configure whether to ignore the default list of resource attributes set in
```
salvacorts, line 127: To me this reads as: _"if this is set to true, we ignore `distributor.otlp.default_resource_attributes_as_index_labels`" and use whatever is configured in `attributes_config`_. Problem is, the `action` in `attributes_config` only support `structured_metadata` or `drop` so how can you define which attributes to store as index labels?

salvacorts (commented):
After a first pass I have a question. IIUC we now have two configs (from the user perspective):
- The distributor OTLP config (GlobalOTLPConfig). Which configures the blessed attributes, that one is not configurable per tenant.
- The per-tenant OTLP config (OTLPConfig).

I find that a bit confusing. Maybe we can configure the blessed attributes per-tenant and have a default for them (instead of on the distributor)?

pkg/loghttp/push/otlp.go
sandeepsukhani: It is just an optimization, which would work most of the time, so I think it would be better to keep the code clean and not add an unwanted dependency.

docs/sources/send-data/otel/_index.md
salvacorts: Resolving. I was wrong.

salvacorts (approved):
After a conversation in Slack, I agree this is the best way to move forward.

PR #12181, opened 2024-03-12
fix(operator): Improve validation of provided S3 storage configuration

btaani:
This PR adds a validation step to verify that the endpoint URL is of the pattern `https://s3.REGION.amazonaws.com` when Amazon AWS S3 is used.

Fixes [LOG-5052](https://issues.redhat.com/browse/LOG-5052)

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -440,16 +454,37 @@ func extractS3ConfigSecret(s *corev1.Secret, credentialMode lokiv1.CredentialMod
...
+	matched, err := regexp.MatchString(pattern, string(endpoint))
+	if !matched {
+		return errS3InvalidEndpoint
+	}
+	if err != nil {
+		return errS3RegexpMatching
+	}
+
+	return nil
+}
```
JoaoBraveCoding, line 486: Some notes on the proposed changes
- Changed both parameters to string so we don't need to cast them every time
- Changed the validation flow to allow for more specific errors.
- Changed the error messages names so it's easier to read (this might only be for me though so feel free to take it or not)
```suggestion
const awsEndpoint = "amazonaws.com"

func validateS3Endpoint(endpoint string, region string) error {
	parsedURL, err := url.Parse(endpoint)
	if err != nil {
		return errS3EndpointUnparseable
	}
	if parsedURL.Scheme != "http" && parsedURL.Scheme != "https" {
		return errS3EndpointUnsuportedScheme
	}
	if !strings.HasSuffix(endpoint, awsEndpoint) {
		return errS3EndpointNotAWS
	}
	if !strings.HasSuffix(endpoint, fmt.Sprintf("%s.%s", region, awsEndpoint)) {
		return errS3EndpointWrongRegion
	}
	if !strings.HasSuffix(endpoint, fmt.Sprintf("s3.%s.%s", region, awsEndpoint)) {
		return errS3EndpointNotS3
	}
	return nil
}
```

operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -416,17 +422,25 @@ func extractS3ConfigSecret(s *corev1.Secret, credentialMode lokiv1.CredentialMod
 		if len(roleArn) != 0 {
 			return nil, fmt.Errorf("%w: %s", errSecretFieldNotAllowed, storage.KeyAWSRoleArn)
 		}
+
+		err := validateS3Endpoint(endpoint, region)
```
JoaoBraveCoding, line 426: Why not pull the `validateS3Endpoint` out of the switch case? Then if the endpoint is != nil we want to run the validation. For instance, in STS I don't think that endpoint is a required field.

operator/internal/handlers/internal/storage/secrets.go
btaani: This is indeed more descriptive. Thanks!

operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -443,13 +460,35 @@ func extractS3ConfigSecret(s *corev1.Secret, credentialMode lokiv1.CredentialMod
...
+	}
+	if !strings.HasSuffix(endpoint, awsEndpoint) {
+		return errS3EndpointNotAWS
+	}
+	if !strings.HasSuffix(endpoint, fmt.Sprintf("%s.%s", region, awsEndpoint)) {
+		return errS3EndpointWrongRegion
+	}
+	if !strings.HasSuffix(endpoint, fmt.Sprintf("s3.%s.%s", region, awsEndpoint)) {
+		return errS3EndpointNotS3
+	}
```
xperimental, line 488: This effectively removes the possibility to use non-AWS S3 endpoints, which is not what this change is supposed to do. We want to provide some additional validations for when the user _wants to use AWS S3_. If they are using something else, then we do not apply these validations, but we also do not forbid to use it.

```golang
const (
  awsEndpointSuffix      = ".amazonaws.com"
)

var (
  errS3EndpointAWSInvalid       = errors.New("endpoint for AWS S3 must include correct region"
)
```

```suggestion
	if strings.HasSuffix(endpoint, awsEndpointSuffix) {
		validEndpoint := fmt.Sprintf("https://s3.%s%s", region, awsEndpointSuffix)
		if endpoint != validEndpoint {
			return fmt.Errorf("%w: %s", errS3EndpointAWSInvalid, validEndpoint)
		}
	}
```

operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -416,17 +429,21 @@ func extractS3ConfigSecret(s *corev1.Secret, credentialMode lokiv1.CredentialMod
 		if len(roleArn) != 0 {
 			return nil, fmt.Errorf("%w: %s", errSecretFieldNotAllowed, storage.KeyAWSRoleArn)
 		}
+
 		// In the STS case region is not an optional field
 		if len(region) == 0 {
 			return nil, fmt.Errorf("%w: %s", errSecretMissingField, storage.KeyAWSRegion)
+		} else if len(endpoint) != 0 && err != nil {
```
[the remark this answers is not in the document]
JoaoBraveCoding: `endpoint` will only be used in `lokiv1.CredentialModeStatic` so I would move `validateS3Endpoint` there together with the error handling. Another alternative would be to move out of the switch the `cfg.STS = true` and put it in a function. Then we would only call `validateS3Endpoint`  when `cfg.STS == true`

JoaoBraveCoding (commented):
Only one comment to be resolved. Appart from that lgtm

operator/internal/handlers/internal/storage/secrets.go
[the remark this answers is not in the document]
btaani: This makes sense :+1: 
So this means that `endpoint` doesn't need to be validated in the other two cases, even if it was present?

operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -416,17 +425,22 @@ func extractS3ConfigSecret(s *corev1.Secret, credentialMode lokiv1.CredentialMod
...
 	case lokiv1.CredentialModeStatic:
 		cfg.Endpoint = string(endpoint)
 
 		if len(endpoint) == 0 {
 			return nil, fmt.Errorf("%w: %s", errSecretMissingField, storage.KeyAWSEndpoint)
+		} else if len(region) != 0 {
+			err = validateS3Endpoint(string(endpoint), string(region))
+			if err != nil {
+				return nil, err
+			}
```
JoaoBraveCoding, line 443: Just trying to make it more readable WDYT?
```suggestion
		} 
		if len(region) != 0 {
		   if err := validateS3Endpoint(string(endpoint), string(region)); err != nil {
		      return nil, err
		   }
	        }
```

operator/internal/handlers/internal/storage/secrets.go
```diff
@@ -416,18 +425,23 @@ func extractS3ConfigSecret(s *corev1.Secret, credentialMode lokiv1.CredentialMod
...
 		}
-
 		return cfg, nil
 	case lokiv1.CredentialModeStatic:
 		cfg.Endpoint = string(endpoint)
 
 		if len(endpoint) == 0 {
 			return nil, fmt.Errorf("%w: %s", errSecretMissingField, storage.KeyAWSEndpoint)
 		}
+		if len(region) != 0 {
```
JoaoBraveCoding, line 440: @btaani wdyt of moving this if inside of `validateS3Endpoint` and making it an `&&` with the last `if` this way we would also run some minor validation for configurations that don't specify a `region`.

xperimental (approved):

PR #12187, opened 2024-03-12
docs: Update structured metadata docs

JStickler:
Follow up to the weekly team meeting - Adding some clarification around the use case for using structured metadata.

### DIFF
--- a/docs/sources/get-started/labels/structured-metadata.md
+++ b/docs/sources/get-started/labels/structured-metadata.md
@@ -17,6 +17,15 @@ to extract at query time.
 
 Structured metadata can also be used to query commonly needed metadata from log lines without needing to apply a parser at query time. Large json blobs or a poorly written query using complex regex patterns, for example, come with a high performance cost. Examples of useful metadata include trace IDs or user IDs.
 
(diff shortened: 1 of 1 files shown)


docs/sources/get-started/labels/structured-metadata.md
stevendungan, line 1: I think we should replace `trace_id` references with something more representative either infra-specific or an Otel resource attribute

stevendungan (commented):
I think we should replace trace_id references with something more representative either infra-specific or an Otel resource attribute

docs/sources/get-started/labels/structured-metadata.md
JStickler: Looking at the list of "blessed" attributes from #12180 and comparing it to the list of attributes discussed in the design doc, maybe `container.id` or `browser.platform`?

stevendungan: I'd propose `container_id` (replace the `.` with `_`), @slim-bean or @sandeepsukhani speak if you have any objection

slim-bean (approved):

PR #12193, opened 2024-03-13
feat: add support for setting custom grpc interceptors for requests sent to index-gateway

sandeepsukhani:
We support setting custom grpc interceptors for grpc requests sent from queriers to ingesters. The interceptors are useful for setting some additional metadata in the requests. However, support for setting custom grpc interceptors is missing for grpc requests from queriers to index gateways. This PR adds support for the same.


cyriltovena (approved):
LGTM

PR #12194, opened 2024-03-13
fix: fix middlewares being applied for series request when querier is running as a standalone service

sandeepsukhani:
When we run queriers as a standalone service we apply some middleware for the HTTP requests supported by queriers.
We have missed applying the same middleware for series requests as what we have for other request types.
This PR fixes the issue by keeping the middlewares consistent for all the requests.


cyriltovena (approved):
LGTM

PR #12195, opened 2024-03-13
fix: Fixes compactor shutdown log

cyriltovena:
Found this  while debugging something, if an error is there it won't show up because you need key and values.


MichelHollands (approved):
LGTM

PR #12196, opened 2024-03-13
fix(bloom-compactor): Correctly disable the bloom-compactor.

cyriltovena:
I was running main on disk for a while and it kept crashing for a while because it was trying to compact bloom.

This ensure we don't actually start services if we don't want to.


salvacorts (commented):
Should we do the same with `t.Cfg.BloomGateway.Enabled` at `initBloomGateway` and `initBloomGatewayRing`?

salvacorts (approved):

PR #12199, opened 2024-03-13
fix(operator): Add validation on empty per-tenant limits config

btaani:
Prior to this PR, the following Lokistack spec used to be valid:
```
spec:
  limits:
    tenants:
      invalid: {} 
```
However, having an empty per-tenant limits config generates an empty map into the loki config, which may cause some problems in some components.

The goal of this PR is to validate the per-tenant limits config to check on empty configurations, which reflects on the status of the Lokistack as a degraded condition.

Fixes [LOG-5050](https://issues.redhat.com/browse/LOG-5050)

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


operator/internal/handlers/lokistack_create_or_update.go
```diff
@@ -85,11 +86,20 @@ func CreateOrUpdateLokiStack(
...
+				Reason:  lokiv1.ReasonInvalidPerTenantLimitsConfig,
+				Requeue: false,
+			}
+		} else {
+			ll.Error(err, "failed to parse query timeout")
+			return "", &status.DegradedError{
+				Message: fmt.Sprintf("Error parsing query timeout: %s", err),
+				Reason:  lokiv1.ReasonQueryTimeoutInvalid,
+				Requeue: false,
+			}
```
JoaoBraveCoding, line 102: No need to put it under a else
```suggestion
		}
		ll.Error(err, "failed to parse query timeout")
		return "", &status.DegradedError{
			Message: fmt.Sprintf("Error parsing query timeout: %s", err),
			Reason:  lokiv1.ReasonQueryTimeoutInvalid,
			Requeue: false,
```

operator/apis/loki/v1/lokistack_types.go
```diff
@@ -1109,6 +1109,8 @@ const (
 	ReasonFailedCertificateRotation LokiStackConditionReason = "FailedCertificateRotation"
 	// ReasonQueryTimeoutInvalid when the QueryTimeout can not be parsed.
 	ReasonQueryTimeoutInvalid LokiStackConditionReason = "ReasonQueryTimeoutInvalid"
+	// ReasonInvalidPerTenantLimitsConfig when the per-tenant limits configuration is empty.
```
JoaoBraveCoding, line 1112:
```suggestion
	// ReasonInvalidPerTenantLimitsConfig when the per-tenant limits configuration is invalid.
```

operator/internal/manifests/options.go
JoaoBraveCoding, line 1: Maybe this might be overcomplicating but I feel we should define the type `errInvalidTenantConfig` and then you can declare different vars of this type with specific error messages. Because now we are adding validation for empty config but maybe we will want to add some other validation in the future. WDYT?

Also, why is the validation inside `NewTimeoutConfig`? Is that the only entry point that we look into tenant config?

btaani: I personally prefer that it stays simple since we do not have other validations (so far) that we need to check regarding the per-tenant limits. It will be a better approach once we start adding more error messages though.

To answer your second question, `NewTimeoutConfig` was the only place I found that checks on the `spec.limits` field.

operator/internal/manifests/options.go
```diff
@@ -126,6 +127,8 @@ type TLSProfileSpec struct {
 	MinTLSVersion string
 }
 
+var ErrInvalidPerTenantConfig error = errors.New("invalid per-tenant config")
```
JoaoBraveCoding, line 130: While testing I read the error message and it would not be obvious what was the issue 
```suggestion
var ErrPerTenantConfigInvalid error = errors.New("per-tenant config cannot contain empty keys")
```

operator/internal/manifests/options.go
```diff
@@ -195,3 +198,15 @@ func calculateHTTPTimeouts(queryTimeout time.Duration) TimeoutConfig {
 		},
 	}
 }
+
+func ValidatePerTenantConfig(s *lokiv1.LimitsSpec) error {
+	if s != nil && s.Tenants != nil {
+		for _, config := range s.Tenants {
+			c := lokiv1.PerTenantLimitsTemplateSpec{}
+			if config == c {
+				return ErrInvalidPerTenantConfig
```
JoaoBraveCoding, line 207: So it's obvious to the user which key is causing the issue
```suggestion
				return fmt.Errorf("%w: %s", ErrPerTenantConfigInvalid, key)
```

operator/internal/manifests/options.go
```diff
@@ -195,3 +198,15 @@ func calculateHTTPTimeouts(queryTimeout time.Duration) TimeoutConfig {
 		},
 	}
 }
+
+func ValidatePerTenantConfig(s *lokiv1.LimitsSpec) error {
+	if s != nil && s.Tenants != nil {
+		for _, config := range s.Tenants {
```
JoaoBraveCoding, line 204:
```suggestion
		for key, config := range s.Tenants {
```

JoaoBraveCoding (commented):
apart from the linting error lgtm

JoaoBraveCoding (approved):
lgtm

btaani:
Closing this one since we do not need it anymore. A warning is already generated for empty limits config in Loki: https://github.com/grafana/loki/commit/0f2f42de01b1ece5406083bab2e4b42f5884bb33

PR #12203, opened 2024-03-13
chore(bloom-gw): Process blocks in parallel

chaudum:

While processing a single block is rather fast, processing a lot of blocks sequentially can lead to problem that single slowly processed blocks lead to high tail latency.

Example trace:
screenshot_20240313_225517


cyriltovena (approved):
LGTM

make sense to me.

PR #12208, opened 2024-03-14
feat(blooms): bloom integration in query planning

owen-d:
TODO:
* [x] proto extensions
* [x] wiring through storage
* [x] GetShards implementation on indexshipper
* [x] GetShards implementation on idx-gw
* [x] Graceful fallback when not implemented
* [x] integration into queryrange (codec, etc)
* [x] http routes
* [x] Integration into sharding middleware
* [x] Integration into logql engine
* [x] storage matcher compat
* [x] ingester inverted index integration


pkg/storage/stores/composite_store.go
```diff
@@ -206,6 +206,41 @@ func (c CompositeStore) Volume(ctx context.Context, userID string, from, through
...
+		if err != nil {
+			return err
+		}
+		groups = append(groups, shards)
+		return nil
+	})
+
+	switch {
+	case err != nil:
+		return nil, err
```
chaudum, line 230: I would move the error handling into a regular `if err != nil` statement

pkg/storage/stores/shipper/indexshipper/gatewayclient/gateway_client.go
```diff
@@ -237,78 +237,109 @@ func (s *GatewayClient) QueryIndex(_ context.Context, _ *logproto.QueryIndexRequ
...
+		// but can be improved easily in the future by using a stream here.
+		for {
+			resp, err := streamer.Recv()
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				return errors.WithStack(err)
+			}
+			res.Shards = append(res.Shards, resp.Shards...)
```
chaudum, line 334: The `append` is error prone here. `poolDo()` retries the job with the next instance in the replication set of a tenant in case of an error.

owen-d: nice catch, tyvm.

pkg/storage/stores/composite_store.go
```diff
@@ -206,6 +208,83 @@ func (c CompositeStore) Volume(ctx context.Context, userID string, from, through
...
+	}
+
+	switch {
+	case len(groups) == 1:
+		return groups[0], nil
+	case len(groups) == 0:
+		return nil, nil
+	default:
+		sort.Slice(groups, func(i, j int) bool {
```
salvacorts, line 241: Since you are looking for the biggest group, you don't need to sort this slice. Instead, you can use [slices.MinFunc][1] O(n).

[1]: https://pkg.go.dev/slices#MinFunc

pkg/storage/stores/shipper/indexshipper/gatewayclient/gateway_client.go
```diff
@@ -237,78 +241,183 @@ func (s *GatewayClient) QueryIndex(_ context.Context, _ *logproto.QueryIndexRequ
...
+	var strategy sharding.PowerOfTwoSharding
+	shards := strategy.ShardsFor(stats.Bytes, uint64(s.limits.TSDBMaxBytesPerShard(userID)))
+	return &logproto.ShardsResponse{
+		Shards: shards,
+	}, nil
+}
+
+// TODO(owen-d): this was copied from ingester_querier.go -- move it to a shared pkg
+// isUnimplementedCallError tells if the GRPC error is a gRPC error with code Unimplemented.
+func isUnimplementedCallError(err error) bool {
```
salvacorts, line 409: Maybe we can extract this to the common `util` pkg, inside the [errors.go file.](https://github.com/grafana/loki/blob/08e505dd50915d2bdbb678211449e56d6c84ac08/pkg/util/errors.go#L150-L149)

pkg/storage/stores/shipper/indexshipper/gatewayclient/gateway_client.go
```diff
@@ -237,78 +241,183 @@ func (s *GatewayClient) QueryIndex(_ context.Context, _ *logproto.QueryIndexRequ
...
+
+			return nil
+		},
+		func(err error) bool {
+			errCt++
+			return errCt <= maxErrs
+		},
+	); err != nil {
+		if isUnimplementedCallError(err) {
+			return s.getShardsFromStatsFallback(ctx, in)
```
salvacorts, line 366: This should only happen during a rollout, If some of the index gateways are still running an older image which doesn't implement the GetShards method, right?

If so, for the sake of simplicity, I think we should have a TODO here to clean the old way of computing shards once the rollout is completed.

salvacorts: We actually, this is something that would prevent users from rolling out Loki 3.0 w/o panics... Maybe a TODO for a release after 3.0?

pkg/storage/stores/shipper/indexshipper/gatewayclient/gateway_client.go
```diff
@@ -365,6 +474,14 @@ func (s *GatewayClient) clientDoQueries(ctx context.Context, gatewayQueries []*l
 // poolDo executes the given function for each Index Gateway instance in the ring mapping to the correct tenant in the index.
 // In case of callback failure, we'll try another member of the ring for that tenant ID.
 func (s *GatewayClient) poolDo(ctx context.Context, callback func(client logproto.IndexGatewayClient) error) error {
+	return s.poolDoWithStrategy(ctx, callback, func(error) bool { return true })
+}
+
+func (s *GatewayClient) poolDoWithStrategy(
```
salvacorts, line 480: idea: What about using this [retry pkg](https://github.com/avast/retry-go)? You can catch errors to avoid retrying, has exponential backoff, logging and context.

pkg/logproto/indexgateway.proto
```diff
@@ -25,4 +26,45 @@ service IndexGateway {
...
+message ShardsRequest {
+  int64 from = 1 [
+    (gogoproto.customtype) = "github.com/prometheus/common/model.Time",
+    (gogoproto.nullable) = false
+  ];
+  int64 through = 2 [
+    (gogoproto.customtype) = "github.com/prometheus/common/model.Time",
+    (gogoproto.nullable) = false
+  ];
+  string query = 3;
```
salvacorts, line 44: I think when we create this ShardRequest in the QF, we have already parsed the plan form the query. maybe we can pass the Plan here instead of the raw query? This way in the future, GetShards can take advantage of whichever information the plan contains.

pkg/storage/stores/shipper/indexshipper/indexgateway/gateway.go
```diff
@@ -340,6 +344,261 @@ func (g *Gateway) GetVolume(ctx context.Context, req *logproto.VolumeRequest) (*
...
+	ctx := server.Context()
+	log, _ := spanlogger.New(context.Background(), "IndexGateway.GetShards")
+	defer log.Finish()
+
+	instanceID, err := tenant.TenantID(ctx)
+	if err != nil {
+		return err
+	}
+
+	p, err := ExtractShardRequestMatchersAndAST(request.Query)
```
salvacorts, line 357: As my previous comment, we can avoid this by having the Plan inside the request.

pkg/storage/stores/shipper/indexshipper/tsdb/sharding/sharding.go
```diff
@@ -0,0 +1,101 @@
+package sharding
```
salvacorts, line 1: nit: I'd rename the files in this pkg to be more informative. E.g. I'd rename `sharding.go` to `fingerprintsize.go`  or similar.

salvacorts (commented):
Did a first pass. Overall looks reasonable to me but this PR is huge. May we split it into smaller ones? I think we can p[lit it into these PRs:
- Extend protos to have GetShards method for the index gw (leave unimplemented). Adding marshaling/unmarshaling code.
- Implement GetShards for index gw using power of 2.
- Move power of 2 sharding to new pkg 
- Implement new chunk size based sharding

Some of these PRs will still get big, but easier to review and reason about.

salvacorts (approved):
As we discussed, since:
- @chaudum and I already reviewed the PR at a high level and we haven't noticed anything obviously wrong.
- The CI is happy
- This should be safe to rollout: QFs fallback to linear sharding if index-gws don't implement GetShards.
- We have been running this in dev for some days already.

We are merging this PR.

PR #12211, opened 2024-03-14
fix: Read all series in NewTSDBSeriesIter and close index file

salvacorts:

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/bloomcompactor/tsdb.go
+++ b/pkg/bloomcompactor/tsdb.go
@@ -7,8 +7,9 @@ import (
 	"math"
 	"path"
 	"strings"
-	"sync"
 
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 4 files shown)


pkg/bloomcompactor/tsdb.go
```diff
@@ -127,74 +135,21 @@ type forSeries interface {
...
-	go func() {
-		err := t.f.ForSeries(
-			t.ctx,
-			t.bounds,
-			0, math.MaxInt64,
-			func(_ labels.Labels, fp model.Fingerprint, chks []index.ChunkMeta) {
-
+func NewTSDBSeriesIter(ctx context.Context, f forSeries, bounds v1.FingerprintBounds) (v1.Iterator[*v1.Series], error) {
+	// TODO(salvacorts): Create a pool
+	series := make([]*v1.Series, 0, 100)
```
chaudum, line 142: I think we can be a bit more generous with the capacity.

chaudum (approved):
lgtm

pkg/bloomcompactor/tsdb.go
salvacorts: Indeed. I used this one kinda randomly 😅 suggestions are welcome.

pkg/bloomcompactor/tsdb.go
```diff
@@ -85,7 +88,7 @@ func (b *BloomTSDBStore) LoadTSDB(
 	tenant string,
 	id tsdb.Identifier,
 	bounds v1.FingerprintBounds,
-) (v1.CloseableIterator[*v1.Series], error) {
+) (v1.Iterator[*v1.Series], error) {
```
cyriltovena, line 91: nit: I think you should keep the closable even if you don't use it now you can release the pool later.

pkg/bloomcompactor/tsdb.go
cyriltovena, line 142: it's a good idea but imo until it comes to be a problem I like to keep things simple.

pkg/bloomcompactor/tsdb.go
```diff
@@ -207,19 +162,20 @@ func (t *TSDBSeriesIter) background() {
...
-	}()
+				series = append(series, res)
+			}
+		},
+		labels.MustNewMatcher(labels.MatchEqual, "", ""),
+	); err != nil {
+		return nil, err
+	}
+
+	select {
```
cyriltovena, line 173: Simplicity love it.

cyriltovena (approved):
LGTM

pkg/bloomcompactor/tsdb.go
salvacorts: I'd do it only when we use a pool proviced it's "obvious" we need to release the buffer on the close. But no strong opinion.

PR #12251, opened 2024-03-18
chore: make error labels injectable via 'Add'

MasslessParticle:
To control the error flow at the pipeline-level, this PR makes it so label errors can be injected into a line via `Add` in the label builder.


cyriltovena:
Do you have an example where you will use this ? This code is going to be executed for every log line, doesn't seem too bad but I wanted to bring this to your attention.

cyriltovena (approved):

PR #12256, opened 2024-03-19
perf(blooms): Recycle bloom page buffers

chaudum:

When reading a bloom page into memory, a new byte slice with the size of the decoded page was allocated.

screenshot_20240319_084201

This PR changes bloom page decoder to use a byte slice from the `BlockPool`. The caller (`LazyBloomIter` of the block querier) is responsible for closing the page whenever a new one is loaded, so the byte slice is put back to its pool. 

screenshot_20240319_084736

Note that the screeshots are from different time ranges and therefore differ in absolute numbers for allocated memory. However the second screenshot shows that there are no more direct allocations in `github.com/grafana/loki/pkg/storage/bloom/v1.LazyDecodeBloomPage` any more.

Additionally, this PR changes the max bucket of the `BlockPool` byte slice pool to 64MB so that pages bigger than 16MB are correctly put back to the pool, since [slices exceeding the capacity of the largest bucket are not put back](https://github.com/grafana/loki/blob/main/vendor/github.com/prometheus/prometheus/util/pool/pool.go#L74-L87).

### DIFF
--- a/pkg/storage/bloom/v1/bloom.go
+++ b/pkg/storage/bloom/v1/bloom.go
@@ -86,7 +86,7 @@ func LazyDecodeBloomPage(r io.Reader, pool chunkenc.ReaderPool, page BloomPageHe
 	}
 	defer pool.PutReader(decompressor)
 
-	b := make([]byte, page.DecompressedLen)
+	b := BlockPool.Get(page.DecompressedLen)[:page.DecompressedLen]
 
 	if _, err = io.ReadFull(decompressor, b); err != nil {
 		return nil, errors.Wrap(err, "decompressing bloom page")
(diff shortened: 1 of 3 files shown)


salvacorts (approved):
LGTM!

pkg/storage/bloom/v1/bloom.go
```diff
@@ -167,6 +166,13 @@ func (d *BloomPageDecoder) Err() error {
 	return d.err
 }
 
+func (d *BloomPageDecoder) Close() {
+	d.err = nil
+	d.cur = nil
+	d.dec.B = d.data[:0]
```
cyriltovena, line 172: why not setting it to nil ? this is scary to keep a ref that is being put back to the pool.

chaudum: Right, it can be set to `nil`.

PR #12259, opened 2024-03-19
chore: delete request processing improvements

sandeepsukhani:
It does the following 2 changes to make some improvements to the delete request processing
1. By default, we split all delete requests by 24 hours. This, however, does not help much when delete requests are without any line filters since Loki is efficient at processing those requests. We load delete requests in batches, so it, in fact, increases the turnaround time for requests without a line filter. I have changed the code to only split requests with a line filter.
2. We have seen users send delete requests that are way beyond what their retained logs would be based on their retention config. Although this does not cause us to waste any resources, it does increase the turnaround time of requests due to batching. I have updated the code to mark those requests as processed straight away to avoid wasting time on them.

**Checklist**
- [x] Tests updated


MichelHollands (approved):
LGTM

PR #12265, opened 2024-03-19
fix: pass correctly formatted query tags from the ruler to the querier

MasslessParticle:
The querier expects query tags to be in the from `key=value` for them to show up in our `metrics.go` line. This PR brings the `ruler` query tag in line with that so it's properly logged


ashwanthgoli (approved):
lgtm

PR #12271, opened 2024-03-20
ci: pin release to loki-2.8.x tag in loki-release

trevorwhitney:

I've created a tag in `grafana/loki-release` for each release pipeline we need to support, so we can change the tag without having to change Loki if the release code needs to be fixed. This PR changes the release code version to use that tag.


MichelHollands (approved):
Approving to unblock

PR #12273, opened 2024-03-20
ci: pin to loki-2.9.x release tag

trevorwhitney:

Same as https://github.com/grafana/loki/pull/12271 but for 2.9.x releases, pins the release code to the `loki-2.9.x` tag in `grafana/loki-release`.


MichelHollands (approved):
Approving to unblock

PR #12277, opened 2024-03-20
chore: allow lint to be run in container locally

trevorwhitney:
keeping golangci-lint versions in sync with various branches is a PIA

our `golangci-lint` version varies from release to release and it's a PIA keeping the correct binary in-sync locally, so I'd like to just run it in the respective build container.


MichelHollands (approved):
Good idea

PR #12280, opened 2024-03-20
ci: add test-integration make target

trevorwhitney:

When I updated the release pipeline it brought in the `integration` check, this adds the necessary makefile target and go build tags.


MichelHollands (approved):
LGTM

PR #12283, opened 2024-03-21
fix: quantile sharding; fix bug where sharding would always happen

cstyan:
The test in the first commit:
```
{
	in:  `quantile_over_time(0.70, {a=~".+"} | logfmt | unwrap value [1s])`,
	out: `quantile_over_time(0.7,{a=~".+"}|logfmt|unwrapvalue[1s])`,
},
````

Will fail without the change from the 2nd commit.

The fix is relatively simple, before we could return a mapped expression via line 460 if there was no label reduction or grouping operation so instead checking if quantile sharding is enabled should be our very first concern in switch case for quantile operation types.

cc @jeschkies @cyriltovena

### DIFF
--- a/pkg/logql/shardmapper.go
+++ b/pkg/logql/shardmapper.go
@@ -451,6 +451,10 @@ func (m ShardMapper) mapRangeAggregationExpr(expr *syntax.RangeAggregationExpr, 
 		}, bytesPerShard, nil
 
 	case syntax.OpRangeTypeQuantile:
+		if !m.quantileOverTimeSharding {
+			return noOp(expr, m.shards)
+		}
+
 		potentialConflict := syntax.ReducesLabels(expr)
 		if !potentialConflict && (expr.Grouping == nil || expr.Grouping.Noop()) {
 			return m.mapSampleExpr(expr, r)
@@ -460,7 +464,7 @@ func (m ShardMapper) mapRangeAggregationExpr(expr *syntax.RangeAggregationExpr, 
 		if err != nil {
 			return nil, 0, err
 		}
-		if shards == 0 || !m.quantileOverTimeSharding {
+		if shards == 0 {
 			return noOp(expr, m.shards)
 		}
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/logql/shardmapper_test.go [hunk 1]
```diff
@@ -417,6 +421,35 @@ func TestMappingStrings(t *testing.T) {
...
+// sharding options, but whose sharding is turned off, are not sharded on those operations.
+func TestMappingStrings_NoProbabilisticSharding(t *testing.T) {
+	m := NewShardMapper(ConstantShards(2), nilShardMetrics, []string{})
+	for _, tc := range []struct {
+		in  string
+		out string
+	}{
+		{ // This should result in the same downstream sharding of the max and not the inner quantile regardless of whether quantile sharding is turned on.
+			in:  `max by (status)(quantile_over_time(0.70, {a=~".+"} | logfmt | unwrap value [1s]))`,
+			out: `maxby(status)(downstream<maxby(status)(quantile_over_time(0.7,{a=~".+"}|logfmt|unwrapvalue[1s])),shard=0_of_2>++downstream<maxby(status)(quantile_over_time(0.7,{a=~".+"}|logfmt|unwrapvalue[1s])),shard=1_of_2>)`,
```
jeschkies, line 434: Ok, at first I thought this was wrong but this works because `quantile_over_time` has no by clause right?

cstyan: This query has `max` as the top level aggregation, and AFAIK currently we're choosing to only shard the top level aggregation.

pkg/logql/shardmapper.go
```diff
@@ -451,6 +451,10 @@ func (m ShardMapper) mapRangeAggregationExpr(expr *syntax.RangeAggregationExpr,
 		}, bytesPerShard, nil
 
 	case syntax.OpRangeTypeQuantile:
+		if !m.quantileOverTimeSharding {
+			return noOp(expr, m.shards)
+		}
+
```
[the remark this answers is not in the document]
cstyan: It has nothing to do with whether there's a label conflict or not, but about not sharding quantile queries if the feature flag to do so is not enabled.

cyriltovena (approved):
LGTM

jeschkies (approved):

PR #12300, opened 2024-03-21
fix(blooms): leaky ref

owen-d:
Reverts https://github.com/grafana/loki/pull/12256 with the exception of the pool sizing deltas. That PR was very similar to the bug which https://github.com/grafana/loki/pull/12050 previously fixed

### DIFF
--- a/pkg/storage/bloom/v1/util.go
+++ b/pkg/storage/bloom/v1/util.go
@@ -32,10 +32,10 @@ var (
 		},
 	}
 
-	// 4KB -> 64MB
+	// 1KB -> 8MB
 	BlockPool = BytePool{
 		pool: pool.New(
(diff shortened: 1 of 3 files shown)


slim-bean (approved):
LGTM

pkg/storage/bloom/v1/util.go
```diff
@@ -35,7 +35,7 @@ var (
 	// 4KB -> 64MB
 	BlockPool = BytePool{
 		pool: pool.New(
-			4<<10, 64<<20, 4,
+			4<<10, 64<<24, 4,
```
chaudum, line 38: IMO `1<<30` would be easier to read
also, the comment above does not match any more

PR #12305, opened 2024-03-21
fix: Pass query to pipeline wrappers from the AST

MasslessParticle:
This PR makes it so that the pipeline wrappers use a query from the AST. This ensured things like comments and unneeded whitespace is removed.


MichelHollands (approved):
LGTM

PR #12312, opened 2024-03-22
fix: (Bug) correct resultType when storing instant query results in cache

kavirajk:
tldr; Previously the results cache for instant metric query is stored as wrong type `TypeMatrix` so some of the results were ignored during final computation. This PR fixes the correct type `TypeVector`

### Before
````
]-bash-5.2$ ./cmd/logcli/logcli instant-query 'sum(rate({job="varlogs"}[1h]))' -q
[
  {
    "metric": {},
    "value": [
      1711117293.063,
      "0.5336111111111111"
    ]
  }
]-bash-5.2$ ./cmd/logcli/logcli instant-query 'sum(rate({job="varlogs"}[1h]))' -q
[
  {
    "metric": {},
    "value": [
      1711117295.905,
      "0.19944444444444445"
    ]
  }
]-bash-5.2$
````
First query will miss the cache and store it in the cache after downstream requests. Second request fetch from the cache. NOTE both are are same (bug)

### After
```
-bash-5.2$ ./cmd/logcli/logcli instant-query 'sum(rate({job="varlogs"}[1h]))' -q
[
  {
    "metric": {},
    "value": [
      1711117478.337,
      "0.5011111111111111"
    ]
  }
]-bash-5.2$
-bash-5.2$ ./cmd/logcli/logcli instant-query 'sum(rate({job="varlogs"}[1h]))' -q
[
  {
    "metric": {},
    "value": [
      1711117484.699,
      "0.5011111111111111"
    ]
  }
]-bash-5.2$
```

NOTE: both first (without cache) and second (with cache) have same results

### Detail explanation

The bug is the assumption of existing `ResultCache` that it always works with `range` queries everytime it has to store `LokiPromResponse` in the cache. But in reality we share the same response for both instant and range queries. [With adding support for instant queries recently.](https://github.com/grafana/loki/pull/11814)

But I failed to update the case where ResultsCache whenever trying to merge the multiple cache entries (we call `Extent`) it create new bigger response type (`LokiPromResponse`) with `resultType` set to `matrix` (assuming only range queries uses it)

So when the cache is hit and these values are returned for computing the merged responses, it considers only samples with resultType `vector` (because it's instant query). That's the reason, when original query split into multiple subqueries and some of those subqueries even though hit cache, it's value is ignored in the final computation. Hence wrong results.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


sandeepsukhani (approved):
LGTM

PR #12326, opened 2024-03-22
fix: don't shard `quantile_over_time` if it's not the top level aggregation for a query

cstyan:
This PR adds the addition of a `top bool` parameter to mapping functions as well as the `Shardable` function for the Expression interface. This allows us to know whether the expression/operation we're currently deciding whether or not to shard is the top level aggregation for the entire query or not.

Currently we want to use this to signal that `quantile_over_time` is unshardable when it's not the top level query aggregation. This also results in something like `max by (x) (quantile_over_time...)` or  `max by (x) (quantile_over_time... by (y))` being unshardable instead of resulting in a query sharded on the `max`. We can also extend this to other probabilistic sharding operation types in the future, such as `topk`.

In the future we could allow sharding of the inner quantile resulting in pattern
```
maxby(foo)(quantileSketchEval<quantileSketchMerge<downstream<__quantile_sketch_over_time__(...)by(baz),shard=1_of_2>++downstream<__quantile_sketch_over_time__(...)by(baz),shard=0_of_2>>>)
```

but for now this is the safest path forward and unblocks shipping of a stable release + other important features.

### DIFF
--- a/pkg/logql/shardmapper.go
+++ b/pkg/logql/shardmapper.go
@@ -56,7 +56,7 @@ func NewShardMapperMetrics(registerer prometheus.Registerer) *MapperMetrics {
 func (m ShardMapper) Parse(parsed syntax.Expr) (noop bool, bytesPerShard uint64, expr syntax.Expr, err error) {
 	recorder := m.metrics.downstreamRecorder()
 
-	mapped, bytesPerShard, err := m.Map(parsed, recorder)
+	mapped, bytesPerShard, err := m.Map(parsed, recorder, true)
 	if err != nil {
 		m.metrics.ParsedQueries.WithLabelValues(FailureKey).Inc()
 		return false, 0, nil, err
@@ -74,7 +74,7 @@ func (m ShardMapper) Parse(parsed syntax.Expr) (noop bool, bytesPerShard uint64,
 	return noop, bytesPerShard, mapped, err
 }
 
-func (m ShardMapper) Map(expr syntax.Expr, r *downstreamRecorder) (syntax.Expr, uint64, error) {
+func (m ShardMapper) Map(expr syntax.Expr, r *downstreamRecorder, topLevel bool) (syntax.Expr, uint64, error) {
 	// immediately clone the passed expr to avoid mutating the original
 	expr, err := syntax.Clone(expr)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/logql/shardmapper.go
```diff
@@ -89,19 +89,19 @@ func (m ShardMapper) Map(expr syntax.Expr, r *downstreamRecorder) (syntax.Expr,
 	case *syntax.MatchersExpr, *syntax.PipelineExpr:
 		return m.mapLogSelectorExpr(e.(syntax.LogSelectorExpr), r)
 	case *syntax.VectorAggregationExpr:
-		return m.mapVectorAggregationExpr(e, r)
+		return m.mapVectorAggregationExpr(e, r, false)
```
jeschkies, line 92: shouldn't we pass `topLevel` here?

pkg/logql/shardmapper.go
```diff
@@ -89,19 +89,19 @@ func (m ShardMapper) Map(expr syntax.Expr, r *downstreamRecorder) (syntax.Expr,
...
 	case *syntax.LabelReplaceExpr:
-		return m.mapLabelReplaceExpr(e, r)
+		return m.mapLabelReplaceExpr(e, r, topLevel)
 	case *syntax.RangeAggregationExpr:
-		return m.mapRangeAggregationExpr(e, r)
+		return m.mapRangeAggregationExpr(e, r, topLevel)
 	case *syntax.BinOpExpr:
-		return m.mapBinOpExpr(e, r)
+		return m.mapBinOpExpr(e, r, true)
```
jeschkies, line 98: I think we need to pass `topLevel` here.

pkg/logql/syntax/ast.go
```diff
@@ -1337,8 +1337,11 @@ func (e *RangeAggregationExpr) String() string {
 }
 
 // impl SampleExpr
-func (e *RangeAggregationExpr) Shardable() bool {
-	return shardableOps[e.Operation] && e.Left.Shardable()
+func (e *RangeAggregationExpr) Shardable(topLevel bool) bool {
+	if e.Operation == OpRangeTypeQuantile && !topLevel {
+		return false
+	}
```
jeschkies, line 1343: Could you add a comment here as well that we are ignoring the feature flag?

cstyan: The AST code currently doesn't have any knowledge of the feature flag the shardmapper uses, but I can add a comment to clarify the behaviour.

jeschkies (approved):

PR #12327, opened 2024-03-22
feat: Introduce special header that tells Loki not to modify query results

MasslessParticle:
This PR introduces the new `LokiOriginalResults` header and ensures it's propagated to any component that could modify query results.

### DIFF
--- a/pkg/loki/modules.go
+++ b/pkg/loki/modules.go
@@ -86,47 +86,47 @@ const maxChunkAgeForTableManager = 12 * time.Hour
 
 // The various modules that make up Loki.
 const (
-	Ring                          string = "ring"
-	RuntimeConfig                 string = "runtime-config"
-	Overrides                     string = "overrides"
-	OverridesExporter             string = "overrides-exporter"
-	TenantConfigs                 string = "tenant-configs"
-	Server                        string = "server"
-	InternalServer                string = "internal-server"
-	Distributor                   string = "distributor"
-	Querier                       string = "querier"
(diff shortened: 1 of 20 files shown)


pkg/loki/modules.go
```diff
@@ -404,7 +404,7 @@ func (t *Loki) initQuerier() (services.Service, error) {
 	toMerge := []middleware.Interface{
 		httpreq.ExtractQueryMetricsMiddleware(),
 		httpreq.ExtractQueryTagsMiddleware(),
-		httpreq.PropagateHeadersMiddleware(httpreq.LokiEncodingFlagsHeader),
+		httpreq.PropagateHeadersMiddleware(httpreq.LokiEncodingFlagsHeader, httpreq.LokiOriginalQueryResultsHeader),
```
ashwanthgoli, line 407: i think the querier change here only handles requests directly made to queriers http endpoint.

for requests that pass through schedulers, i believe we are encoding these headers within the request itself. 
for [protobuf encoding](https://github.com/grafana/loki/blob/79fd9203ebe434882d308a58a7b0b866f0509843/pkg/querier/queryrange/marshal.go#L364) and [json encoding](https://github.com/grafana/loki/blob/79fd9203ebe434882d308a58a7b0b866f0509843/pkg/querier/queryrange/codec.go#L612)

MasslessParticle: ty, good catch

na-- (approved):

PR #12334, opened 2024-03-25
fix(blooms): Extract only line filters before line format expressions

salvacorts:

This PR fixes a bug from #12035. Before this PR, we'd extract all line filter expression from a query and build bloom tests out of them. This would potentially break queries with a line filter after a line format expression.

For example, in the query `{app="fake"} |= "foo" | line_format "thisNewTextShouldMatch" |= "thisNewTextShouldMatch"` , the last line filter (` |= "thisNewTextShouldMatch"`) should always match since that text was added with the `line_format` expression before the fillter. Before this PR, the bloom tests would match `thisNewTextShouldMatch` against the blooms and would discard all chunks. As a result, the query would return no data.

In this PR we modify how we extract the line filters form a LogQL expression: we only extract (and thus build bloom tests) for all line filters before a line format expression. In the example above, we'd only build a bloom test for the line filter `|= "foo"`.

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/storage/bloom/v1/bloom_tester.go
+++ b/pkg/storage/bloom/v1/bloom_tester.go
@@ -35,6 +35,39 @@ func (b BloomTests) MatchesWithPrefixBuf(bloom filter.Checker, buf []byte, prefi
 	return true
 }
 
+// ExtractTestableLineFilters extracts all line filters from an expression
+// that can be tested against a bloom filter. This will skip any line filters
+// after a line format expression. A line format expression might add content
+// that the query later matches against, which can't be tested with a bloom filter.
+// E.g. For {app="fake"} |= "foo" | line_format "thisNewTextShouldMatch" |= "thisNewTextShouldMatch"
+// this function will return only the line filter for "foo" since the line filter for "thisNewTextShouldMatch"
+// wouldn't match against the bloom filter but should match against the query.
+func ExtractTestableLineFilters(expr syntax.Expr) []syntax.LineFilterExpr {
+	if expr == nil {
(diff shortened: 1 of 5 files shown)


cyriltovena (approved):
LGTM

pkg/storage/bloom/v1/bloom_tester.go
```diff
@@ -35,6 +35,39 @@ func (b BloomTests) MatchesWithPrefixBuf(bloom filter.Checker, buf []byte, prefi
...
+	var filters []syntax.LineFilterExpr
+	var lineFmtFound bool
+	visitor := &syntax.DepthFirstTraversal{
+		VisitLineFilterFn: func(v syntax.RootVisitor, e *syntax.LineFilterExpr) {
+			if e != nil && !lineFmtFound {
+				filters = append(filters, *e)
+			}
+		},
+		VisitLineFmtFn: func(v syntax.RootVisitor, e *syntax.LineFmtExpr) {
+			if e != nil {
```
chaudum, line 59: Is it possible that the visitor visits a `LineFmtExpr` that is `nil`?

chaudum (approved):
Good catch!

pkg/storage/bloom/v1/bloom_tester.go
salvacorts: Probably not but not sure tbh. Still I think it's a good practice to check ptrs. Happy to remove it tho.

PR #12345, opened 2024-03-25
fix: log state of 'disable pipeline wrappers'

MasslessParticle:
Log when disable pipeline wrappers is used


JordanRushing (approved):
LGTM

PR #12368, opened 2024-03-26
chore(bloom-gw): Cleanup tracing implementation

chaudum:

The huge amount of `bloomgateway.ProcessTask` spans caused problem with Tempo's span limit, resulting in dropped/dangling spans.

Since the processing time is now recorded also in a `metrics.go`-like stats log line, we can remove the spans and only add a log event with the summary to the main span of the request handler.

This PR also unifies the event logging in the index gateway, and adds the real processing time (not aggregated processing time) to the request stats.

### DIFF
--- a/pkg/bloomgateway/processor.go
+++ b/pkg/bloomgateway/processor.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 
 	"github.com/grafana/dskit/concurrency"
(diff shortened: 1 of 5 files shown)


pkg/bloomgateway/processor.go
```diff
@@ -162,22 +170,14 @@ func (p *processor) processBlock(_ context.Context, blockQuerier *v1.BlockQuerie
...
-		md, _ := blockQuerier.Metadata()
-		blk := bloomshipper.BlockRefFrom(task.Tenant, task.table.String(), md)
-		sp.LogKV("block", blk.String())
+		// md, _ := blockQuerier.Metadata()
+		// blk := bloomshipper.BlockRefFrom(task.Tenant, task.table.String(), md)
+		//
+		// Why don't we get the span from the task context even though we create it
+		// at the beginning of the request handler?
+		// sp := opentracing.SpanFromContext(task.ctx)
+		// sp.LogKV("process block", blk.String())
```
chaudum, line 180: Needs to be investigated.

salvacorts (approved):

PR #12374, opened 2024-03-27
Promtail: Fix handling of docker logs when a line is split in multiple frames

### LINKED ISSUE #2281: Reconstruct split docker log lines in Promtail
Starting in https://github.com/moby/moby/pull/22982 docker now splits log lines longer than ~16kb

This can totally break JSON processing of these lines in Promtail.

There are a few moving pieces here to support this, the big one is adding support for multi-line logs in promtail which we have been avoiding. Mostly because there were other valid solutions like `show context` in explore.

However, not being able to parse JSON in Promtail because of this split is certainly something we want to fix so I think we are going to finally have to go down the multi-line road.

It does appear docker has some specific metadata to help with this however: https://github.com/moby/moby/commit/0b4b0a7b5d5de8cb575b666312fceaa2cd58e658 which could help special case reconstructing split docker lines.

### LINKED ISSUE #12197: Consistently corrupt log data for long (>16kB) log lines
**Describe the bug**
When a log is >16kB in size, we see "random" injected iso8601-timestamps in the data. 
These timestamps appear every 16k characters (`1<<14`) exactly, and they are equal to the Loki timestamp (to the nanosecond), so they are not actually random but fully consistent.

**To Reproduce**
I have attached a tar file with a small docker compose reproducer project. Using this, it is enough to run `docker compose up -d` and inspect, using Grafana's explore functionality on `localhost:3000`, the logs from the example_logger service, which will be corrupt.

However, the steps can also be explained slightly more detailed:

1. Have an example logger that produces large logs. In my compose example, a small python script repeatedly outputs a large log (>32k).
2. Have a minimal local loki + grafana setup. I am using 2.9.5, also tested on 2.9.0.
3. Use promtail to ship docker logs to loki. Also 2.9.5.
4. Inspect in e.g. grafana explore using the query ``{container=~`.*logger.*`} |~ `\w2024-` ``. This will highlight the timestamps that have been injected.

**Expected behavior**
Logs should appear in their original form. Ie. in my reproducer compose project, i would expect 32k 'a' characters uninterrupted by iso timestamps.

**Environment:**
 - Infrastructure: docker
 - Deployment tool: docker compose

billede

[badlogs.tar.gz](https://github.com/grafana/loki/files/14586092/badlogs.tar.gz)

jonaslb:

Docker splits log lines into 16 kiB frames if a line is longer than that.
When requesting timestamps, every frame is prefixed with the timestamp.
Previously, Promtail would concatenate the frames naively, and only strip the timestamp at the start of a line, and not at the start of every frame.
The result is that logs which were larger than 16 kiB would get corrupted, especially an issue in the case of structured logs that would be parsed downstream somewhere.
This PR changes the docker logs handling to be based around frames in channels instead of just writing into an `io.Write`.

- Fixes #12197
- Fixes #2281 

- The code which is modified from `github.com/docker/docker/pkg/stdcopy` has been put into `pkg/framedstdcopy` and marked as apache2. This is because the bug is (arguably) also present in e.g. the docker CLI, and possibly also in many other places, and I think it is most appropriate then to keep the license for that the same as stdcopy already is.
- I think the new approach of sending slices into a channel is probably more allocation-heavy than the previous approach (with `io.Write`). I don't think it's an issue, but maybe there are strong opinions about it?
-  I put a soft limit on the buffering size when combining frames of 16Mi. Maybe there are opinions about that too?

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


clients/pkg/promtail/targets/docker/target.go
```diff
@@ -149,81 +147,98 @@ func (t *Target) processLoop(ctx context.Context) {
...
 
-	reader := bufio.NewReader(r)
-	for {
-		line, err := readLine(reader)
+	// softMaxBuffer is to prevent unlimited memory growth.
+	// In principle this introduces the same issue that we are compensating for,
+	// i.e. split messages, however there should be some limit and so this number
+	// is chosen to be "very high" just not "OOM high" (which admittedly is not
+	// a universally defined threshold).
+	const softMaxBuffer = 16 * 1024 * 1024
```
cstyan, line 169: if we're using the same 16k as before, how do your changes in this PR get end users a different outcome?

will a log line just be split into multiple log lines instead of being "corrupted" as mentioned in the issue you opened?

cstyan (commented):
I can do a more in depth review later, but I think it would also be a good idea to include copies (modified where needed) of the tests from the docker codes `stdcopy_test.go`

clients/pkg/promtail/targets/docker/target.go
```diff
@@ -109,22 +107,22 @@ func (t *Target) processLoop(ctx context.Context) {
 	}
 
 	// Start transferring
-	rstdout, wstdout := io.Pipe()
-	rstderr, wstderr := io.Pipe()
+	cstdout := make(chan []byte, 1)
+	cstderr := make(chan []byte, 1)
```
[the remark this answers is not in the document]
jonaslb: I thought that there was not much point in making it large. But since each channel element here are mostly small and at most 16 kiB, it would probably not hurt to pick a larger number. It could be e.g. 16 instead?

clients/pkg/promtail/targets/docker/target.go
jonaslb: This is 16 MiB, not 16 kiB :) This message size would likely get rejected at e.g. Loki or other destination anyway. But yes, the proposed behaviour when size goes above 16 MiB is to split the message. I think the only alternative is unbounded buffer growth, which would also seem undesirable, but i can certainly change that if you prefer?

jonaslb:
Thanks for the comments - I have responded to those review comments and will make adjustments (separate tests corresponding to stdcopy_test, removal of changelog entry) sometime this week.

jonaslb:
Requested tests were added and the manually added changelog entry was removed.

clients/pkg/promtail/targets/docker/target.go
cstyan: Ah yes, you're right :facepalm: 

What if we set this to the default `max_line_size` limit for loki of 256kb? https://grafana.com/docs/loki/latest/configure/#limits_config

We definitely shouldn't have an unbounded buffer, but splitting messages more regularly into pieces that Loki _should_ actually accept would be best.

jonaslb: That sounds like a good idea actually. Do you think it should be made a setting then, or should we just pick 256k as the sensible number everyone should use?

I guess this opens up also some kind of hypothetical `max_line_size_exceeded_action` setting which could be e.g. `split`, `truncate`, `discard` (Loki seems to have only a boolean switch for truncate/discard).

jonaslb: Ah, I'm not sure how I missed that setting already existing. But I guess that, since this configuration is already there, we should probably follow it instead of hardcoding something? Even though the default seems to actually be unbounded line size...

clients/pkg/promtail/targets/docker/target.go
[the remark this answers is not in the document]
jonaslb: I've now changed this to be unbuffered because this seems to be the common practice in the rest of promtail, and I don't have a particular motivation for doing this differently.

clients/pkg/promtail/targets/docker/target.go
jonaslb: It's now changed to use the configured `max_line_size` although if it is unlimited (set to 0), then we use a max of 256k instead.

clients/pkg/promtail/targets/docker/target.go
```diff
@@ -185,34 +193,50 @@ func (t *Target) process(frames chan []byte, logStream string) {
...
 		// Add to buffer
 		payloadAcc.WriteString(payload)
 		curTs = ts
+
 		// Send immediately if line ended or we built a very large event
-		if strings.HasSuffix(payload, "\n") || payloadAcc.Len() >= softMaxBuffer {
+		if isEol || payloadAcc.Len() > sizeLimit {
+			discardRemainingLine = !isEol
 			t.handleOutput(logStream, curTs, payloadAcc.String())
 			payloadAcc = new(strings.Builder)
```
cstyan, line 234: same here

clients/pkg/promtail/targets/docker/target.go
```diff
@@ -185,34 +193,50 @@ func (t *Target) process(frames chan []byte, logStream string) {
...
 		}
 	}
 }
 
 func (t *Target) handleOutput(logStream string, ts time.Time, payload string) {
-	// We don't output trailing newlines
-	payload = strings.TrimSuffix(payload, "\n")
-	payload = strings.TrimSuffix(payload, "\r")
```
cstyan: should we be removing these?

cstyan (commented):
:+1: good progress, I think we can merge soon

clients/pkg/promtail/targets/docker/target.go
jonaslb: I moved this into the `process` function which now has a `TrimRight` call for these. The reason is that `process` keeps track of the length of the buffer, to see if it's over the new limit we set, but this length should exclude the trailing newline (because we don't output the newline). And I thought it was just easier to take away the trailing newline earlier in the process for this purpose (otherwise we'd have to keep track of the buffer length in a more complicated manner than just calling `payloadAcc.Len()`).

cstyan:
@jonaslb as a promtail user, do you have any thoughts on the best place to document the behaviour between `max_line_size` and the docker target? Internally we haven't agreed on exactly what to do here. IMO we probably need a note in both sections [here](https://github.com/grafana/loki/pull/12374) and [here](https://github.com/grafana/loki/blob/a00f1f1b0b8f536f2cdac2f8857eb40c716aa696/docs/sources/send-data/promtail/configuration.md?plain=1#L1943-L1955) noting that the docker SD code respects the `max_line_size` config prior to other portion of promtail like pipeline stages.

jonaslb:
Yeah, now that I've integrated `max_line_size` support, I agree that we need to document it. 

I don't think it needs to be explicitly said that the target also respects the limit, because the data flowing from a target to the client (where lines are truncated/discarded according to the setting) is not user observable anyway. So the main thing to document, I think, is the default limit due to memory and then noting that it can be changed with `max_line_size`.

I have pushed a commit with a proposal for how it could be documented.

docs/sources/send-data/promtail/configuration.md
```diff
@@ -1951,6 +1951,12 @@ or [journald](https://docs.docker.com/config/containers/logging/journald/) loggi
 Note that the discovery will not pick up finished containers. That means
 Promtail will not scrape the remaining logs from finished containers after a restart.
 
+The Docker target correctly joins log segments if a long line was split into different frames by Docker.
+To avoid hypothetically unlimited line size and out-of-memory errors in Promtail, this target applies
+a default soft line size limit of 256 kiB. If the buffer increases above this size, then the line will
```
cstyan, line 1956: can we just clarify that that 256k default comes from prometheus' default max line size config

cstyan:
Just one more minor change to the docs and then we can merge imo

docs/sources/send-data/promtail/configuration.md
jonaslb: Assume you meant Loki as we discussed :) It's updated

cstyan (approved):

cstyan (approved):
great :+1: thanks for your patience @jonaslb

PR #12379, opened 2024-03-27
feat: prepare 3.0.0 release candidate again

trevorwhitney:
Release-As: 3.0.0-rc.1

The existing minor release is not picking up our custom version via the `Release-As` commit. Before debugging `release-please` to figure out way, I'm hoping this will force the correct version.


.release-please-manifest.json
```diff
@@ -1,3 +1,3 @@
 {
-  ".": "2.9.6"
+  ".": "2.9.6-rc.1"
```
MichelHollands, line 2: Should this still be 2.9.6?

trevorwhitney: yes it should be 2.9.6, not sure about the `-rc.1` yet though. the `Release-As: 3.0.0-rc.1` comment in the commit body was not being detected as it is supposed to be. This is probably a bug in `release-please`, but given time constraints I'm opting to run a new release pipeline with a versioning strategy of `always-bump-major`.

slim-bean (approved):

PR #12385, opened 2024-03-27
feat(detected_fields): initial plumbing of new endpoint

trevorwhitney:

This sets up the initial plumbing of the new detected fields endpoint. For the time being is is returning hard coded data, but this PR is meant to get the endpoint responding and establishing the API to free up the front end team

* This PR introduces an `experimental` api prefix and config to disable (that defaults to "off")
* establishes request/response for detected fields

Does not fix, but is related to #<issue number>


pkg/loki/modules.go
```diff
@@ -1053,6 +1053,10 @@ func (t *Loki) initQueryFrontend() (_ services.Service, err error) {
 		t.Server.HTTP.Path("/api/prom/tail").Methods("GET", "POST").Handler(defaultHandler)
 	}
 
+  if t.Cfg.Frontend.ExperimentalAPIsEnabled {
```
cyriltovena, line 1056: Not sure if we need a flag if this is an experimental API

cyriltovena (approved):
LGTM

pkg/loki/modules.go
trevorwhitney: I like the idea of just not responding to these endpoints at all when turned off. Makes it very quick to stop the bleeding if there's something terribly wrong.

PR #12392, opened 2024-03-28
feat: discovery of name of services emitting the logs

sandeepsukhani:
otel specs recommends having a service name to identify the service emitting the logs. It would be good to adopt that for non-otel logs as well and add a single label to identify all the services to make it easier to discover the services and filter down the logs by service(s). This PR adds required changes to, by default, always have a `service_name` label if it does not exist.

`service_name` is defined using existing labels. It picks up values from labels defined in `discover_service_name` per-tenant config. The default list is configured to map one of the following labels to `service_name`: `service`, `app`, `application`, `container`, `container_name`, `app_kubernetes_io_name`, `component`, `workload`,`job`. If none of that is defined in the original labels in the stream then the `service_name` label would be set to value `unknown_service`.
The discovery of a service name can be disabled by setting the `discover_service_name` to an empty list.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


docs/sources/configure/_index.md
```diff
@@ -2824,6 +2824,11 @@ The `limits_config` block configures global and per-tenant limits in Loki.
 # CLI flag: -validation.increment-duplicate-timestamps
 [increment_duplicate_timestamp: <boolean> | default = false]
 
+# List of labels to map to service_name label, if not present in the stream
+# labels during ingestion. Empty list disables setting the label.
+# CLI flag: -validation.discover-service-name
+[discover_service_name: <list of strings> | default = [service app application container container_name app_kubernetes_io_name component workload job]]
```
cyriltovena, line 2830: I think container container_name should be after app_kubernetes_io_name

cyriltovena: And we should add `name` after application

cyriltovena (approved):
LGTM

see my comment.

docs/sources/configure/_index.md
```diff
@@ -2824,6 +2824,11 @@ The `limits_config` block configures global and per-tenant limits in Loki.
 # CLI flag: -validation.increment-duplicate-timestamps
 [increment_duplicate_timestamp: <boolean> | default = false]
 
+# List of labels to map to service_name label, if not present in the stream
+# labels during ingestion. Empty list disables setting the label.
+# CLI flag: -validation.discover-service-name
+[discover_service_name: <list of strings> | default = [service app application name app_kubernetes_io_name container container_name component workload job]]
```
JStickler, line 2830: Suggestion that would make it clearer that the label will be one of the items in the list, not the entire list.

```suggestion
[discover_service_name: <list of strings> | default = [One from service, app, application name, app_kubernetes_io_name, container, container_name component, workload, OR job]]
```

sandeepsukhani: The part you have modified is auto-generated. I think it would be hard to update that part. Can we update the help text instead?

JStickler: Sounds like a good workaround to me!

docs/sources/configure/_index.md
```diff
@@ -2824,6 +2824,11 @@ The `limits_config` block configures global and per-tenant limits in Loki.
 # CLI flag: -validation.increment-duplicate-timestamps
 [increment_duplicate_timestamp: <boolean> | default = false]
 
+# List of labels to map to service_name label, if not present in the stream
+# labels during ingestion. Empty list disables setting the label.
```
JStickler, line 2828: @sandeepsukhani  How does this sound?  
```suggestion
# List of labels to map to service_name label, if not present in the stream
# labels during ingestion. Empty list disables setting the label.
# Default is to apply one of the list of possible labels to discover_service_name.
```

PR #12394, opened 2024-03-28
fix(tsdb): correctly use bit prefix calculation in tsdb shard matching

owen-d:
I unwittingly introduced a bug in https://github.com/grafana/loki/pull/12208 by using the old `astmapper` shard's `Match` calculation, which caused us to incorrectly calculate shard inclusion in TSDB. This PR fixes that and I'll look into another to preclude reintroducing this type of error.


slim-bean (approved):
it's poetic how simple this fix is vs how much it took to find

PR #12396, opened 2024-03-28
fix(blooms): logql shard embeds tsdb

owen-d:
followup to https://github.com/grafana/loki/pull/12394 which introduced a tricky bug. This PR removes the `astmapper.ShardAnnotation` entirely from the `logql.Shard` implementation to prevent similar bugs from accidentally happening in the future. TSDB power-of-two shards use bit-prefixes for shard inclusion as opposed to the modulos (bit-suffix) used by the old `astmapper` variant I originally added to Cortex and we ported over here. Thus when iterating TSDB indices but using the other shard calculation logic, we'd erroneously exclude data.


slim-bean (approved):
LGTM

PR #12397, opened 2024-03-29
feat(helm): Allow extraObject items as multiline strings

### LINKED ISSUE #12032: [helm] ingress-secret isn't idempotent

The core issue is that it doesn't seem to be possible right now to render an idempotent ingress-secret.  This matters because CD tools such as Argocd or Flux mandate this -- they can't compare desired state against currently deployed state if desired state changes every time its rendered.  Ultimately this is a limitation of the [helm implementation](https://helm.sh/docs/chart_template_guide/function_list/#htpasswd) of htpasswd which uses the bcrypt algorithm that adds salt to the password which is why the output is always unique.  Both a pro and a con.  The `htpasswd` binary uses MD5 by default which doesn't use salt, but that's not the algorithm the helm implementation is using.

This solution works but only for some ingress controllers.  ingress-nginx supports an alternative format for basic-auth using the annotation [nginx.ingress.kubernetes.io/auth-secret-type: auth-map](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#authentication) (which is actually already [defined](https://github.com/grafana/loki/blob/687978826f065de2d9501b55a41104ad89b3e321/production/helm/loki/values.yaml#L1246)) which means that each username will be stored as a key, and password as the b64enc value.  This must be special handling by the ingress-nginx folks as I see no mention of it in nginx-ingress and native nginx doesn't support this which means `gateway` won't.  An example of how this could be done: https://github.com/TheRealNoob/loki/commit/525f44b68156d1db59879b03bf1cf309b901602e

I wasn't able to find a solution that would work for gateway given that native nginx only supports encrypted passwords that came out of the cryptography library.  Not unless you got into solutions involving init containers passing the b64enc values (as above) through htpasswd.  That way the k8s object is technically idempotent.

**Additional context**

Please let me know your thoughts on this approach -- as stated it won't work for gateway but it seems to be a harmless change allowing `ingress` to at least work.

TheRealNoob:

Allows extraObjects to define multiline string passed through `{{ tpl }}`, useful for templating field names.

Fixes #12032 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated


TheRealNoob:
I've seen examples of enhancements that bumped the helm chart PATCH version.  I bumped the MINOR version because that's what SEMVER dictates, but I'll be happy to change it if you want.

TheRealNoob:
@JStickler why no more review?  I've been waiting on this lately wondering if I shouldn't be updating it?  My hesitation is I don't want to do it if it's not going to get attention.

JStickler:
@TheRealNoob, Michel was getting backed up on other work, removing him as a reviewer just makes it clear to the rest of the team that they should be reviewing this PR.  If you could fix the conflicts on this PR, I'll try to get someone from engineering to review for you (I'm the technical writer on the team, so I don't do code reviews, just docs).

TheRealNoob:
> If you could fix the conflicts on this PR, I'll try to get someone from engineering to review for you

Done.

MichelHollands:
@TheRealNoob Thanks for your contribution. This is failing Helm linting. Can you have a look at this?

TheRealNoob:
@MichelHollands fixed.  I found that the bug was introduced in [this commit](https://github.com/grafana/loki/commit/521d40a96a5c1c65c786c73ec374580fe767dd3b#diff-2296a17a0f4611b539d59e3c89cf6a8f53db863af266d6fdab2846d704c47b52R934) - fortunately it was easy to fix.

PR #12398, opened 2024-03-29
feat: add pattern match line filter

### LINKED ISSUE #12347: Filter with the pattern
Right now you can only use 
```
| pattern `foo <buzz> bar`
``` 

 to extract fields/labels (e.g `buzz` here) but you can't use it for filtering non matching log line of that pattern.

While we can easily make the suggested pattern filtering by replacing fields with regex

```
|~ `foo (.*) bar
``` 

this is by far inefficient.

I suggest we introduce a new way in the language to filter with the pattern language. This will be much more efficient and less problematic with control character which has been proven to be painful in my previous hackathon.

We should be able to filter in and out with pattern.  Part of the task is to suggest the two new language syntax addition.

kolesnikovae:
The PR introduces new line filter operators for pattern matching, `|>` and `!>`. Despite the fact that filtering by patterns is a subset of the regex filter scope, we decided to extend the syntax explicitly.

The filter presumes that placeholders mask arbitrary non-empty literals. Thus, the `"<_>"` pattern matches any non-empty line, and the `""` pattern only matches empty lines. The semantics are identical to the `pattern` parse stage, with an exception that named captures are not allowed: fields are never extracted at filtering.

The PR does not include documentation. At this point, the new feature is somewhat experimental and should be used with caution. The same applies to the frontend query builder and syntax checker.

Fixes #12347

**Checklist**
- [ ] Documentation added
- [x] Tests updated
- [x] `CHANGELOG.md` updated


pkg/logql/log/pattern/pattern.go
```diff
@@ -6,40 +6,67 @@ import (
 )
 
 var (
-	ErrNoCapture   = errors.New("at least one capture is required")
-	ErrInvalidExpr = errors.New("invalid expression")
+	ErrNoCapture         = errors.New("at least one capture is required")
+	ErrCaptureNotAllowed = errors.New("named captures are not allowed")
```
kolesnikovae, line 10: This is debatable, but I think it makes sense: if user tries to parse _fields_, it's better to return an explicit error. In the future, the restriction can be removed.

Note that we're using an unnamed placeholder <_>. I think we could use a new capture identifier (such as <*>) to emphasize the difference with the pattern parse stage. However, it feels like a new syntax for pattern matching, which may confuse users. I would like to hear others' thoughts on this.

cyriltovena: I don't have strong opinitions using _ vs * but I definitively agree we should fail fast.

cyriltovena (approved):
Looks good to me !

pkg/logql/syntax/ast_test.go
```diff
@@ -539,6 +581,10 @@ func TestStringer(t *testing.T) {
 			in:  `{app="foo"} !~ ip("127.0.0.1") or "foo"`,
 			out: `{app="foo"} !~ ip("127.0.0.1") !~ "foo"`,
 		},
+		{
+			in:  `{app="foo"} !> "<_> foo <_>" or "foo <_>" !> "foo <_> baz"`,
```
cyriltovena, line 585: I think the or is a bug here and should not be support but I see it elsewhere. I didn't know about this cc @kavirajk

pkg/logql/log/pattern/pattern.go
slim-bean: I would vote for `<_>` as the behavior between these two examples feels consistent to me:

```
|> "dur=<_> err=<_>"
| pattern "dur=<_> err=<_>"
```

I think of that as telling Loki to ignore any of the content in `<_>`

I'm not sure I see the reason to use `<*>` or maybe I'm misunderstanding something?

pkg/logql/syntax/ast_test.go
kolesnikovae: It looks legit to me: note the De Morgan's law a few lines above (`!(A || B) == !A && !B`):

`not match ("<_> foo <_>" or "foo <_>")` is the same as 
`not match "<_> foo <_>" and not match "foo <_>"`

pkg/logql/log/pattern/pattern.go
kolesnikovae: This popped up in the Slack conversation, so I brought it here.

Thank you, guys, for sharing your thoughts. I keep `<_>`.

cyriltovena:
please add a changelog item btw !

PR #12406, opened 2024-03-29
fix(detected_fields): fix issues with frontend integration

trevorwhitney:

* Makes sure the string `type` of a detected field is always included in the json response.

    When using the protobuf `enum`, this value gets removed from the json response. I tried writing a custom json marshaller, which worked, but broke the unmarshaling back int protobuf.  After bashing my head against this for a few hours, this seemed like quicker solution.

* Remove config flag, endpoint is always on, but nothing happens unless you hit it
* Remove `/experimental` api and put it under `v1` with everything else


cyriltovena (approved):
Lgtm

PR #12414, opened 2024-03-31
perf(bloomstore): Cache metas LIST operation

chaudum:

Reduce the amount of class A list operations on object storage.

### DIFF
--- a/pkg/storage/stores/shipper/bloomshipper/client.go
+++ b/pkg/storage/stores/shipper/bloomshipper/client.go
@@ -8,6 +8,8 @@ import (
 	"hash"
 	"io"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/go-kit/log"
 	"github.com/grafana/dskit/concurrency"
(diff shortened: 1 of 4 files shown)


pkg/storage/stores/shipper/bloomshipper/client.go
```diff
@@ -417,3 +419,88 @@ func findPeriod(configs []config.PeriodConfig, ts model.Time) (config.DayTime, e
...
+		return nil, nil, fmt.Errorf("does not support LIST calls with delimiter: %s", delimiter)
+	}
+	c.mtx.RLock()
+	cached, found := c.cache[prefix]
+	c.mtx.RUnlock()
+	if found {
+		return cached.objects, cached.prefixes, nil
+	}
+
+	c.mtx.Lock()
```
salvacorts, line 482: Should this be moved after the `c.ObjectClient.List`? Right now it's blocking any List of till the inner client finished.
IIUC we might want to do this so two lists doesn't happen at the same time but I think that's not ideal. Would make sense to block prevent two simultaneus calls for the same prefix tho.

chaudum: Right, I would like to avoid multiple concurrent LIST calls. You still need to sync the map access thereafter. So I think it makes sense to lock the whole operation.

Having separate locks for different prefixes makes eviction more complex. Not sure if that's worth it then.

salvacorts: Sounds good. We can re-evaluate if we see lock contention.

salvacorts (approved):

Issue #12339, opened 2024-03-25
detected fields API


trevorwhitney:
**Is your feature request related to a problem? Please describe.**
Since "detected fields" (ie. `logfmt` or `json` fields in a log line) are only known at query time, users exploring their logs can only get information about fields present in the first 1000 (or log limit) log lines. This often pictures a very limited picture. Since indexed label/values are often more about infrastructures, fields in log lines can more important to users debugging an application who may be less knowledgeable about the infrastructure.

To enable these use cases, as well as better visualizations of fields present in logs, it would be nice to have an API that returned information about detected fields. 

**Describe the solution you'd like**
An API that returns information about _useful_ detected fields, where usefulness is determined by:
* for strings, is the cardinality of values below some configured threshold?
* for "unwrappable" values, what can it be unwrapped into (ie. duration, bytes, etc.)

The response should include the label names, and should be ordered by usefulness (ie. common fields come first)

**Describe alternatives you've considered**
* A time range could be broken up and sampled to get a better representation of a wider time range, however this still wouldn't include information about usefulness with regards to cardinality and unwrap type.

**Additional context**
Future work may include storing this information in an index, but that's not a requirement for this iteration.

matyax:
@trevorwhitney  This was closed as completed, but I don't see any new PR related with it.

trevorwhitney:
yeah, I closed it as it's not planned. the real solution to me seems to pre-parse log lines in store them in dataobj columnar format, in which case would we still need this "usefulness" score, since the API would be exhaustive and complete?

PR #12421, opened 2024-04-01
Detected fields data

trevorwhitney:

This PR calculates real responses for the new experimental detected fields endpoint. It does this in an unsustainable, brute force way by parsing each line (up to a limit) in the querier. Ideally, long term, we'll figure out a way to store this data at ingest, but this is a short term solution to vet the API and data we're surfacing to the UI.

**Which issue(s) this PR ~fixes~ references**:
RE: #12339


pkg/loghttp/labels.go
```diff
@@ -99,17 +99,3 @@ func ParseDetectedLabelsQuery(r *http.Request) (*logproto.DetectedLabelsRequest,
 		Query: query(r),
 	}, nil
 }
-
-func ParseDetectedFieldsQuery(r *http.Request) (*logproto.DetectedFieldsRequest, error) {
```
trevorwhitney, line 103: moved to `query.go` in the same package

pkg/loghttp/query.go
```diff
@@ -617,6 +617,93 @@ func ParseVolumeRangeQuery(r *http.Request) (*VolumeRangeQuery, error) {
 	}, nil
 }
 
+func ParseDetectedFieldsQuery(r *http.Request) (*logproto.DetectedFieldsRequest, error) {
```
trevorwhitney, line 620: moved from `labels.go` and additional fields populated

pkg/querier/querier.go
```diff
@@ -900,18 +906,6 @@ func (q *SingleTenantQuerier) Volume(ctx context.Context, req *logproto.VolumeRe
 	return seriesvolume.Merge(responses, req.Limit), nil
 }
 
-func (q *SingleTenantQuerier) DetectedFields(_ context.Context, _ *logproto.DetectedFieldsRequest) (*logproto.DetectedFieldsResponse, error) {
```
trevorwhitney, line 903: moved below `DetectedLables` so so it was next to the private functions it calls

trevorwhitney:
currently not returning any data in `loki-dev-005`, will publish for review once it's working

pkg/querier/querier.go
```diff
@@ -921,3 +916,267 @@ func (q *SingleTenantQuerier) DetectedLabels(_ context.Context, _ *logproto.Dete
...
+
+	if _, err := strconv.ParseBool(value); err == nil {
+		return logproto.DetectedFieldBoolean
+	}
+
+	if _, err := time.ParseDuration(value); err == nil {
+		return logproto.DetectedFieldDuration
+	}
+
+	if _, err := humanize.ParseBytes(value); err == nil {
```
cyriltovena, line 1059: this could be a string.Contains like we do for the level label detection. I suspect a small benchmark with some special log line will help. Might also be true for duration.

cyriltovena (commented):
Love the direction !

PR #12422, opened 2024-04-01
feat(detected_labels): Add pod as a detected label to demo

shantanualsi:
Just adds the pod label to the detected labels skeleton https://github.com/grafana/loki/pull/12390 as per suggestion.


cyriltovena (approved):
Lgtm

PR #12425, opened 2024-04-01
feat: Enable warnings in Loki query responses

MasslessParticle:
Under certain conditions, it might be useful for Loki to be able to return warnings to the frontend. The prometheus response types support this so it's just a matter of wiring it in.


pkg/ingester/ingester.go
```diff
@@ -874,6 +875,7 @@ func (i *Ingester) GetOrCreateInstance(instanceID string) (*instance, error) { /
 func (i *Ingester) Query(req *logproto.QueryRequest, queryServer logproto.Querier_QueryServer) error {
 	// initialize stats collection for ingester queries.
 	_, ctx := stats.NewContext(queryServer.Context())
+	_, ctx = metadata.NewContext(ctx)
```
MasslessParticle, line 878: This creates a new metadata context for the ingester. Any warnings will be attached in the returned batch stats.

pkg/ingester/instance.go
```diff
@@ -971,6 +973,7 @@ func sendBatches(ctx context.Context, i iter.EntryIterator, queryServer QuerierQ
 
 		stats.AddIngesterBatch(int64(batchSize))
 		batch.Stats = stats.Ingester()
+		batch.Warnings = metadata.Warnings()
```
MasslessParticle, line 976: Here's where we attach warnings to the ingester batch response. From here on out, it's a matter of encoding and decoding (and encoding and decoding) in the scheduler and frontend to propagate them to the user

pkg/logql/engine.go
```diff
@@ -266,6 +266,7 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
 		Data:       data,
 		Statistics: statResult,
 		Headers:    metadataCtx.Headers(),
+		Warnings:   metadataCtx.Warnings(),
```
MasslessParticle, line 269: This is the crux of the PR on the frontend. Everything comes through here so this is where we attach the warnings to be propagated to the scheduler and querier.

pkg/querier/queryrange/codec.go
```diff
@@ -1145,7 +1144,8 @@ func decodeResponseJSONFrom(buf []byte, req queryrangebase.Request, headers http
 						ResultType: loghttp.ResultTypeMatrix,
 						Result:     toProtoMatrix(resp.Data.Result.(loghttp.Matrix)),
 					},
-					Headers: convertPrometheusResponseHeadersToPointers(httpResponseHeadersToPromResponseHeaders(headers)),
+					Headers:  convertPrometheusResponseHeadersToPointers(httpResponseHeadersToPromResponseHeaders(headers)),
```
MasslessParticle, line 1147: In this file, we have to attach warnings to each of the response types that might possibly have them

pkg/querier/queryrange/codec.go
```diff
@@ -1962,10 +1965,27 @@ func mergeLokiResponse(responses ...queryrangebase.Response) *LokiResponse {
 		lokiResponses = make([]*LokiResponse, 0, len(responses))
 	)
 
+	uniqueWarnings := map[string]struct{}{}
 	for _, res := range responses {
 		lokiResult := res.(*LokiResponse)
 		mergedStats.MergeSplit(lokiResult.Statistics)
 		lokiResponses = append(lokiResponses, lokiResult)
+
```
MasslessParticle, line 1973: Loki splits a query and then merges the responses into one. Here we make sure all the warnings from subqueries make it into the merged query. We'll do the same below for metrics queries

pkg/querier/queryrange/queryrangebase/query_range.go
```diff
@@ -164,8 +164,28 @@ func (p prometheusCodec) MergeResponse(responses ...Response) (Response, error)
 	// Merge the responses.
 	sort.Sort(byFirstTime(promResponses))
 
+	uniqueWarnings := map[string]struct{}{}
```
MasslessParticle, line 167: Merging metrics queries

pkg/util/marshal/query.go
```diff
@@ -176,11 +176,25 @@ func NewMetric(l labels.Labels) model.Metric {
 	return ret
 }
 
-func EncodeResult(data parser.Value, statistics stats.Result, s *jsoniter.Stream, encodeFlags httpreq.EncodingFlags) error {
+func EncodeResult(data parser.Value, warnings []string, statistics stats.Result, s *jsoniter.Stream, encodeFlags httpreq.EncodingFlags) error {
 	s.WriteObjectStart()
 	s.WriteObjectField("status")
 	s.WriteString("success")
 
+	if len(warnings) > 0 {
```
MasslessParticle, line 184: Custom json marshalling

pkg/loghttp/query.go
```diff
@@ -42,15 +42,25 @@ const (
...
+	Warnings []string          `json:"warnings,omitempty"`
+	Data     QueryResponseData `json:"data"`
 }
 
 func (q *QueryResponse) UnmarshalJSON(data []byte) error {
 	return jsonparser.ObjectEach(data, func(key, value []byte, dataType jsonparser.ValueType, offset int) error {
 		switch string(key) {
 		case "status":
 			q.Status = string(value)
+		case "warnings":
```
MasslessParticle, line 55: Custom json unmarshalling

pkg/logqlmodel/metadata/context.go
```diff
@@ -94,3 +119,19 @@ func ExtendHeaders(dst map[string][]string, src []*definitions.PrometheusRespons
 		dst[header.Name] = header.Values
 	}
 }
+
+func AddWarnings(ctx context.Context, warnings ...string) error {
+	context, ok := ctx.Value(metadataKey).(*Context)
```
na--, line 124: Presumably, most of the time there will be no warnings, and looking up keys in contexts can be surprisingly expensive, not to mention the locking, so a short circuit here is probably a good idea :thinking: 

```suggestion
	if len(warnings) == 0 {
		return nil
	}
	context, ok := ctx.Value(metadataKey).(*Context)
```

pkg/querier/queryrange/codec.go
```diff
@@ -1962,10 +1965,27 @@ func mergeLokiResponse(responses ...queryrangebase.Response) *LokiResponse {
...
+		for _, w := range lokiResult.Warnings {
+			uniqueWarnings[w] = struct{}{}
+		}
+	}
+
+	warnings := make([]string, 0, len(uniqueWarnings))
```
na--, line 1979: Nit, but since [we already import `golang.org/x/exp`](https://github.com/grafana/loki/blob/b66c3433831fbd2c257427fa9a0d6ea52f3ededd/go.mod#L138), we might as well use [`maps.Keys`](https://pkg.go.dev/golang.org/x/exp/maps#Keys) from it

pkg/logqlmodel/metadata/context.go
```diff
@@ -72,6 +75,28 @@ func (c *Context) Headers() []*definitions.PrometheusResponseHeader {
 	return headers
 }
 
+func (c *Context) Warnings() []string {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	warnings := make([]string, 0, len(c.warnings))
```
na--, line 82: Same as https://github.com/grafana/loki/pull/12425/files#r1548053553, we can use `maps.Keys()` here

pkg/logql/accumulator.go
```diff
@@ -353,6 +355,14 @@ func (acc *AccumulatedStreams) Result() []logqlmodel.Result {
 		)
 	}
 
+	warnings := make([]string, 0, len(acc.warnings))
```
na--, line 358: Same, `maps.Keys()` can simplify this

pkg/querier/queryrange/querysharding.go
```diff
@@ -252,6 +252,7 @@ func (ast *astMapperware) Do(ctx context.Context, r queryrangebase.Request) (que
 		return nil, err
 	}
 
+	// TODO(tpatterson): add warnings to response
```
na--, line 255: This `TODO` seems obsolete, you seem to have added the warnings to all of the response types below ?

pkg/loghttp/query.go
```diff
@@ -42,15 +42,25 @@ const (
...
+			var warnings []string
+			if _, err := jsonparser.ArrayEach(value, func(value []byte, dataType jsonparser.ValueType, offset int, err error) {
+				warnings = append(warnings, string(value))
```
na--, line 58: Hmm considering JSON can contain all sorts of junk in an array, this probably needs a bit more validation that we are actually getting strings here, presumably through the `dataType` :thinking: 

I assume that `value` doesn't contain the string quote characters, but I wonder if it contains escape characters, e.g. if the warning itself contains a quote symbol, I assume it will be escaped in `value` here?

na-- (approved):
I don't fully understand how every part of this is threaded and fits together, but overall the code LGTM, with the minor nitpicks I added as inline comment. 

I'll :+1: to unblock, though the only inline comment that is somewhat blocking is [this one](https://github.com/grafana/loki/pull/12425#discussion_r1548078163), though [this one](https://github.com/grafana/loki/pull/12425#discussion_r1548028258) is probably also quite nice to have, considering how `AddWarnings()` seems to be very much on the hot path.

PR #12437, opened 2024-04-02
feat: include the stream we failed to create in the stream limit error message

cstyan:
This has come up a few times with end users/support, it can be difficult to know what to do if you receive the StreamLimit error as it doesn't contain any additional information. Logging of the labels isn't a big improvement, but hopefully it will at least point people in the right direction.


DylanGuedes (approved):

cstyan:
I suppose `feat` is the most accurate for this, not a bug fix but an improvement to the existing logging.

PR #12439, opened 2024-04-02
chore: Minor changes to contributing doc and pr template for conventional commits

cstyan:
Changes based on the fact that we're now using conventional commits, we no longer want contributors making changes to the changelog themselves + we should include the link to conventional commits doc in the PR template.


JStickler (approved):
[docs team] two small tweaks, otherwise LGTM.

.github/pull_request_template.md
```diff
@@ -9,8 +9,7 @@ Fixes #<issue number>
 - [ ] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
 - [ ] Documentation added
 - [ ] Tests updated
-- [ ] `CHANGELOG.md` updated
-  - [ ] If the change is worth mentioning in the release notes, add `add-to-release-notes` label
```
trevorwhitney, line 13: why don't we remove the action that was updating the release notes with this PR as well?

trevorwhitney (approved):
LGTM, but let's also remove the GH workflow that was acting on that `add-to-release-notes` label with this PR as well

.github/pull_request_template.md
cstyan: done, I also added a note that we need to update the docs about updating the release notes as well, I'm not sure what changes we need to make there at the moment

PR #12442, opened 2024-04-03
fix: Fixes detected labels API wiring

cyriltovena:

This was ommited in previous PR causing 

```
curl  -L -d query="{service_name=\"mimir-ingester\"}" -G  http://localhost:3100/loki/api/v1/detected_labels
unsupported response type, got ()%

```

during usage

cc @shantanualsi


MichelHollands (approved):
LGTM

PR #12447, opened 2024-04-03
fix: Fix multi_tenant querier for detected labels

shantanualsi:
Quick fix to multi tenant querier detected labels to fix dev environment
Also returns a default label to avoid confusion


cyriltovena (approved):
LGTM

PR #12448, opened 2024-04-03
refactor!: refactor how we do defaults for runtime overrides

slim-bean:

We introduced a "defaults" section into the runtime overrides for runtime configs, however this prevented us from being able to set a default for these values without providing a runtime config.

we already have a pattern for this we used for limits_config so this PR refactors those defaults to be a first class object in the loki config and follows the same pattern of setting this as defaults.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/pkg/loki/loki.go
+++ b/pkg/loki/loki.go
@@ -99,10 +99,10 @@ type Config struct {
 	TableManager        index.TableManagerConfig   `yaml:"table_manager,omitempty"`
 	MemberlistKV        memberlist.KVConfig        `yaml:"memberlist"`
 
-	RuntimeConfig   runtimeconfig.Config `yaml:"runtime_config,omitempty"`
-	OperationConfig runtime.Config       `yaml:"operation_config,omitempty"`
-	Tracing         tracing.Config       `yaml:"tracing"`
-	Analytics       analytics.Config     `yaml:"analytics"`
+	RuntimeConfig     runtimeconfig.Config `yaml:"runtime_config,omitempty"`
+	OperationalConfig runtime.Config       `yaml:"operational_config,omitempty"`
+	Tracing           tracing.Config       `yaml:"tracing"`
+	Analytics         analytics.Config     `yaml:"analytics"`
 
 	LegacyReadTarget bool `yaml:"legacy_read_target,omitempty" doc:"hidden|deprecated"`
 
@@ -172,7 +172,7 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 	c.BloomCompactor.RegisterFlags(f)
 	c.QueryScheduler.RegisterFlags(f)
 	c.Analytics.RegisterFlags(f)
-	c.OperationConfig.RegisterFlags(f)
+	c.OperationalConfig.RegisterFlags(f)
 }
 
 func (c *Config) registerServerFlagsWithChangedDefaultValues(fs *flag.FlagSet) {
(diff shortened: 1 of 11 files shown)


pkg/loki/loki.go
```diff
@@ -99,9 +99,10 @@ type Config struct {
 	TableManager        index.TableManagerConfig   `yaml:"table_manager,omitempty"`
 	MemberlistKV        memberlist.KVConfig        `yaml:"memberlist"`
 
-	RuntimeConfig runtimeconfig.Config `yaml:"runtime_config,omitempty"`
-	Tracing       tracing.Config       `yaml:"tracing"`
-	Analytics     analytics.Config     `yaml:"analytics"`
+	RuntimeConfig   runtimeconfig.Config `yaml:"runtime_config,omitempty"`
+	OperationConfig runtime.Config       `yaml:"operation_config,omitempty"`
```
slim-bean, line 103: struggling a bit with the name here.....

pkg/runtime/config.go
```diff
@@ -9,7 +11,26 @@ type Config struct {
 	LimitedLogPushErrors bool `yaml:"limited_log_push_errors"`
 }
 
-var EmptyConfig = &Config{}
+// RegisterFlags adds the flags required to config this to the given FlagSet
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.BoolVar(&cfg.LogStreamCreation, "operation-config.log-stream-creation", false, "Log every new stream created by a push request (very verbose, recommend to enable via runtime config only).")
+	f.BoolVar(&cfg.LogPushRequest, "operation-config.log-push-request", false, "Log every push request (very verbose, recommend to enable via runtime config only).")
+	f.BoolVar(&cfg.LogPushRequestStreams, "operation-config.log-push-request-streams", false, "Log every stream in a pus request (very verbose, recommend to enable via runtime config only).")
```
DylanGuedes, line 18:
```suggestion
	f.BoolVar(&cfg.LogPushRequestStreams, "operation-config.log-push-request-streams", false, "Log every stream in a push request (very verbose, recommend to enable via runtime config only).")
```

DylanGuedes (changes requested):
looks awesome.

could you add a changelog entry (since it is a breaking change)?

slim-bean:
> could you add a changelog entry (since it is a breaking change)?

it's only a breaking change for us right now, I don't believe we ever released code with that default section in the runtime config? that didn't make it into 2.9 I don't think?

production/ksonnet/loki/overrides.libsonnet
```diff
@@ -25,14 +25,6 @@ local k = import 'ksonnet-util/kausal.libsonnet';
...
-
-    defaultRuntimeConfigs: {
-      // override the default runtime configs. Useful to change a behavior globally across multiple tenants. See pkg/loki/runtime_config.go
-      //
-      // log_stream_creation: true,
-      // log_push_request: true,
-      // log_push_request_streams: true,
-    },
   },
   local configMap = k.core.v1.configMap,
```
DylanGuedes, line 30: i think you'll have to remove this line too
```diff
  overrides_config:
    configMap.new($._config.overrides_configmap_name) +
    configMap.withData({
      'overrides.yaml': k.util.manifestYaml(
        {
          overrides: $._config.overrides,
          configs: $._config.runtimeConfigs,
-         default: $._config.defaultRuntimeConfigs,
        }
        + (if std.length($._config.multi_kv_config) > 0 then { multi_kv_config: $._config.multi_kv_config } else {}),
      ),
    }),
```

DylanGuedes (changes requested):

production/ksonnet/loki/overrides.libsonnet
slim-bean: ah good catch, ty

DylanGuedes (approved):

Issue #12415, opened 2024-03-31
Bug(logql): Linefilters doesn't work well with `or` expressions


kavirajk:
This[ PR added support for `or` expression to linefilters](https://github.com/grafana/loki/pull//8962), making it easy to queries like following without using `|~` regexp matcher.

`{job="varlogs"} |= "foo" or "bar"`

instead of 

`{job="varlogs"} |~ "foo|bar"`

But looks like it breaks when we have multiple linefilters.

example, `{job="varlogs"} != "test" != "foo" or "bar"`

behaving like `{job="varlogs"} != "test" != "bar"` (buggy)

instead of `{job="varlogs"} != "test" != "foo" != "bar"` (correct. !(A || B) => !A && !B)

Have a working[ example in this gist.](https://gist.github.com/kavirajk/085d4a81b6e913fbb9a013b1e07644d4)

My feeling is, when introducing the PR to support this, we probably missed supporting nested linefilters in linefilters grammer.
```
lineFilters:
    lineFilter                { $$ = $1 }
  | lineFilter OR orFilter    { $$ = newOrLineFilter($1, $3)}
  | lineFilters lineFilter    { $$ = newNestedLineFilterExpr($1, $2) }
  ;

```

PR #12463, opened 2024-04-04
fix(logql): Linefilter behaviour with chained `or` 

kavirajk:
Fixes: https://github.com/grafana/loki/issues/12415

### Background
`Or` in LineFilterExpr works as follows.
	
Case 1: With MatchEqual operators(|= or |~, etc)
example: `{app="loki"} |= "test" |= "foo" or "bar"`
expectation: match "test" AND (either "foo" OR "bar")
	
Case 2: With NotMatchEqual operators (!= or !~, etc)
example: `{app="loki"} != "test" != "foo" or "bar"`
expection: match !"test" AND !"foo" AND !"bar", Basically exactly as if `{app="loki"} != "test" != "foo" != "bar"`.

See LineFilterExpr tests for more examples.

#### 1.  Assuming right most LineFilter never be nested.

NOTE: Before supporting `or` in linefilter, `right` will always be a leaf node (right.next == nil)
After supporting `or` in linefilter, `right` may not be a leaf node (e.g: |= "a" or "b). Here "b".Left = "a")

We traverse the tree recursively untile we make `right` leaf node.
Consider the following expression. `{app="loki"} != "test" != "foo" or "bar or "car"`
It first creates following tree on the left and transformed into the one on the right (as shown in the attached image)
Screenshot 2024-04-03 at 15 33 00

#### 2. Wrong handling with NotEqual operators like `!= and !~` 

When parsing "or" chains in linefilter, particularly variations of NOT filters `(!= or !~)`, we need to transform
say `(!= "foo" or "bar "baz") => (!="foo" != "bar" != "baz")`

But we were missing the last part of the "or" chain before this patch.

#### 3. Not updating the operators of `or` chain when handling NotEqual operators like `!= and !~`

Consider, we have chain of "or", `!= "foo" or "bar" or "baz"` we parse from right to left, so first time `left="bar"`, `right="baz"`, and we don't know the actual `Ty` (equal: |=, notequal: !=, regex: |~, etc). 

So it will have default (0, LineMatchEqual).

We only know real `Ty` in next stage, where `left="foo"`, `right="bar or baz"`, at this time, `Ty` is LineMatchNotEqual(!=).
Now we need to update not just `right.Ty = left.Ty`, we also have to update all the `right.Or`  until `right.Or` is nil.

Fixes #<issue number>

Look at the `ast_test.go` to understand more about the behaviours of `or` filters in Linefilters.

**Checklist**
- [x] Documentation added
- [x] Tests updated
- [ ] `CHANGELOG.md` updated


cyriltovena (approved):
LGTM

MasslessParticle (approved):
LGTM. Appreciate all the test cases!

kavirajk:
Manual testing.

Before:
Screenshot 2024-04-05 at 18 03 31

After

Screenshot 2024-04-05 at 18 02 24

PR #12466, opened 2024-04-04
fix: Bugfixes for detected  labels API

shantanualsi:
Fixes bug in the detected labels API. 

- Query does not work as matches are not recognized
- Blank value increases the cardinality by 1
- Dedupe does not work properly
- panics if ingester labels are nil

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -1389,7 +1389,7 @@ func (i *Ingester) GetDetectedLabels(ctx context.Context, req *logproto.Detected
 	}
 	var matchers []*labels.Matcher
 	if req.Query != "" {
-		matchers, err := syntax.ParseMatchers(req.Query, true)
+		matchers, err = syntax.ParseMatchers(req.Query, true)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


pkg/ingester/ingester.go
```diff
@@ -1403,7 +1403,7 @@ func (i *Ingester) GetDetectedLabels(ctx context.Context, req *logproto.Detected
 	}
 	result := make(map[string]*logproto.UniqueLabelValues)
 	for label, values := range labelMap {
-		uniqueValues := make([]string, len(values))
```
shantanualsi, line 1406: use of make introduced blank values in the slice

pkg/ingester/ingester.go
```diff
@@ -1389,7 +1389,7 @@ func (i *Ingester) GetDetectedLabels(ctx context.Context, req *logproto.Detected
 	}
 	var matchers []*labels.Matcher
 	if req.Query != "" {
-		matchers, err := syntax.ParseMatchers(req.Query, true)
```
shantanualsi, line 1392: := again after declaring zero value above 🧠

pkg/querier/ingester_querier.go
```diff
@@ -370,17 +370,24 @@ func (q *IngesterQuerier) DetectedLabel(ctx context.Context, req *logproto.Detec
...
+				combinedValues = thisIngesterValues.Values
+			}
+			labelMap[label] = combinedValues
 		}
 	}
+
+	// Dedupe all ingester values
 	mergedResult := make(map[string]*logproto.UniqueLabelValues)
 	for label, val := range labelMap {
-		uniqueValues := slices.CompactFunc(val, strings.EqualFold)
```
shantanualsi, line 383: Compact only works for dedupe if values are sorted

cyriltovena (approved):
LGTM

Many bug fixes, few tests proving it's fixes. Please add more in follow up PR.

PR #12470, opened 2024-04-04
perf(blooms): Use jumpmap for client side sharding of the filter requests

chaudum:

Use the same client-side sharding mechanism for bloom gateway filter requests that we use for client side sharding in the memcached client.

This uses DNS discovery to resolve a list of addresses uses as buckets for the jumphash algorithm to shard GroupedChunkRefs across bloom gateway servers.
The ring component is still part of the bloom gateway, but it is not used any more.

https://arxiv.org/pdf/1406.2294.pdf


pkg/bloomgateway/client.go
```diff
@@ -258,31 +231,28 @@ func (c *GatewayClient) FilterChunks(ctx context.Context, tenant string, from, t
 		// overlap on instances to the left and right of the range.
 		firstFp, lastFp := groups[0].Fingerprint, groups[len(groups)-1].Fingerprint
 		pctKeyspace := float64(lastFp-firstFp) / float64(math.MaxUint64)
-		pctInstances := float64(len(servers)) / float64(len(rs.Instances))
+		pctInstances := float64(len(servers)) / float64(len(c.pool.Addrs()))
```
owen-d, line 234: avoid divide by zero if addrs() changes during computation
```suggestion
		pctInstances := float64(len(servers)) / float64(max(1,len(c.pool.Addrs())))
```

owen-d (approved):
nit, lgtm

PR #12478, opened 2024-04-04
docs: move Storage page under Configuration

JStickler:
Moves single Storage topic under Configuration.
Slight revision to move TSDB to be the first storage option instead of the last mentioned in the overview.
Fixes cross-references broken by the move.


slim-bean (approved):
LGTM

jdbaldry:
To avoid the issues this PR caused in future:

We should:

- Check build errors reported locally when you run `make docs`. 
  You'll see something like:

  ```
  WARN [en] REF_NOT_FOUND: Ref "../../storage": "/hugo/content/docs/loki/next/get-started/labels/structured-metadata.md:1:1": page not found
  WARN [en] REF_NOT_FOUND: Ref "../../storage": "/hugo/content/docs/loki/next/operations/storage/retention.md:1:1": page not found
  ```
  It warns of broken links that need to be fixed. Writers' Toolkit documents the Hugo build errors that result from the use of `relref` https://grafana.com/docs/writers-toolkit/review/test-documentation-changes/#example-page-not-found.

- Add an alias for the old page: https://grafana.com/docs/writers-toolkit/write/front-matter/#determine-the-relative-alias.

PR #12482, opened 2024-04-05
fix: lambda-promtail, update s3 filename regex to allow finding of log files from AWS GovCloud regions

Overflow0xFFFF:

A broken regex prevents AWS GovCloud regions from properly being supported with Lambda Promtail. The regex matches only on regions with three parts in their filenames (e.g., us-east-1), and the GovCloud regions have four parts (e.g., us-gov-east-1).

No issue found.

The fix is to add an additional, non-greedy, non-capturing group to match on the fourth parth. Initially, I attempted to use `(?:[\w-]+)` for its simplicity and to match on the region-parsing logic we already have. However, this breaks the tests for vpcflowlogs because of the greediness of the `+` operator. As such, I opted to write something more verbose, but non-greedy.

Tests have been added to prevent future breakages, as this tends to be an edge case.

**Checklist**
- [x] Documentation added
- [x] Tests updated


cstyan (approved):
:+1: thanks @Overflow0xFFFF

PR #12484, opened 2024-04-05
chore(blooms): Remove bloom gateway ring

chaudum:

Remove ring in favour of a client side, DNS discovery of the bloom gateway servers combined with consistent hashing using the jumphash algorithm (#12470).

Instead of configuring the ring, the bloom gateway client needs to be configured with the `-bloom-gateway-client.addresses` flag, which takes a list of comma separated, DNS-discovery-style strings.


salvacorts (approved):
👏 Nice!

PR #12491, opened 2024-04-05
feat: split detected fields queries

### LINKED ISSUE #12339: detected fields API

trevorwhitney:

This PR splits detected fields queries using the existing split by time logic used for log filter queries.

Fixes #12339


pkg/querier/queryrange/codec_test.go
```diff
@@ -1013,7 +1013,7 @@ func Test_codec_EncodeResponse(t *testing.T) {
 			},
 		},
 		{
-			"loki legacy", "/api/promt/query",
+			"loki legacy", "/api/prompt/query",
```
shantanualsi, line 1016: looks like unintended spell check thing

pkg/storage/detected/fields.go
```diff
@@ -0,0 +1,41 @@
+package detected
+
+import (
+	"github.com/grafana/loki/v3/pkg/logproto"
+)
+
+func MergeFields(fields []*logproto.DetectedField, limit uint32) []*logproto.DetectedField {
```
shantanualsi, line 7: this is not yet used right?

trevorwhitney: It should be, so maybe I have a bug? It should be used by the query frontend when merging subqueries

shantanualsi: ah no, my bad.. its called directly in codec.go.. 👍

pkg/storage/detected/fields.go
```diff
@@ -0,0 +1,42 @@
...
+			mergedFields[field.Label] = field
+			foundFields++
+			continue
+		}
+
+		//seeing the same field again, update the cardinality if it's higher
+		//this is an estimation, as the true cardinality could be greater
+		//than either of the seen values, but will never be less
+		if ok {
+			curCard, newCard := f.Cardinality, field.Cardinality
```
cyriltovena, line 29: if you merge multiple timerange which isn't the cardinality additive ?

trevorwhitney: because you don't know if you've seen different values in each timerange, for example

timerange 1 (cardinality 3)
```
flavor=sweet
flavor=sour
flavor=sweet
flavor=spicy
```

timerange 2 (cardinality 2)
```
flavor=sweet
flavor=sweet
flavor=bland
```

the safest bet is to take the highest cardinality seen (though, like in this example, the true cardinality is actually 4). since we're discarding the actual values, there's no way to be 100% accurate here

pkg/querier/queryrange/codec_test.go
trevorwhitney: reverted, thanks!

pkg/storage/detected/fields.go
cyriltovena: Sounds like you might want to run the sketch in the frontend may be. Happy to start like this though.

But if you can stream 100k logs back to frontend and do sketch there then it will be more accurate. may be querier could do the parsing of fields to absord most of the CPU work.

trevorwhitney: Cyril suggest async looking in to some ways to merge the sketches in the frontend, so I'm going to investigate that a bit.

cyriltovena (approved):
LGTM.

Really think we should look at merging sketch for better result in the future.

shantanualsi (approved):
LGTM

pkg/storage/detected/fields.go
trevorwhitney: I've changed this code to merge sketches and get more accurate cardinality estimations

PR #12494, opened 2024-04-06
fix(blooms): fixup bloom testing

owen-d:
I found some problems with how we were testing bloom filter ngram membership which caused us to exclude chunks we shouldn't. This PR has a few fixes:

* Handle skip factors in bloom membership
* Improved testing (previous tests didn't test a lot of the actual logic, were complex, etc)
* removes regexps from bloom testing because the current impls were provably incorrect and I don't have time to fix these now (followup for later)

### DIFF
--- a/pkg/storage/bloom/v1/bloom_tester.go
+++ b/pkg/storage/bloom/v1/bloom_tester.go
@@ -1,6 +1,8 @@
 package v1
 
 import (
+	"unicode/utf8"
+
 	"github.com/grafana/regexp"
 	regexpsyntax "github.com/grafana/regexp/syntax"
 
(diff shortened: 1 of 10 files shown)


slim-bean (approved):
:shipit:

pkg/storage/bloom/v1/bloom_tester.go
```diff
@@ -100,21 +102,7 @@ func simpleFilterToBloomTest(b NGramBuilder, filter syntax.LineFilter) BloomTest
...
-
-		simplifier := log.NewRegexSimplifier(newStringFilterFunc(b), newStringFilterFunc(b))
-		matcher, ok := simplifier.Simplify(reg, false)
-		if !ok {
-			// If the regex simplifier fails, we default to MatchAll
-			return MatchAll
-		}
-
-		return matcherFilterWrapper{filter: matcher}
+		return MatchAll
```
chaudum, line 105: :+1:

chaudum (approved):

PR #12502, opened 2024-04-08
fix: use nanosecond precision for timestamp in compacted tsdb index file names

### LINKED ISSUE #12500: Compacted TSDB data lost because S3 object name not unique between compaction and retention
**Describe the bug**
Logs went missing in the sense that queries would no longer return them for tsdb period 19804 (March 22, with 24h period).

By combining the compactor logs and the logs of our S3 service we found that the compacted tsdb was erroneously DELETEd.

This occurred in our staging environment running `main-406b10c-amd64` with the restructured compactor. It appears that with the two compaction loops (one for pure compaction and one for retention) introduced by the restructured compactor, it is possible for both to process the same table with such an interleaving of execution where the old compacted tsdb object name is equal to the new, that it ends up deleted instead of replaced.

Note that bloom filters are not enabled.

We don't recognise a fix for this in the compactor between that commit and current main.

**To Reproduce**
So far we have a single occurrence in our staging environment.

**Expected behavior**
The compacted tsdb should never end up erroneously deleted instead of updated (after a null update from loki-write?)

**Environment:**
Infrastructure: Vanilla Kubernetes v1.28.5 on Fedora CoreOS 38 using a Scality RING S3
Image: main-406b10c-amd64
Deployment tool: helm

Attaching logs for compactor and S3
[s3.log](https://github.com/grafana/loki/files/14902814/s3.log)
[compactor.log](https://github.com/grafana/loki/files/14902900/compactor.log)

sandeepsukhani:
Do the same changes as PR #11277, i.e., have a creation timestamp in compacted index files with nanosecond precision instead of seconds. I had only made that change for `boltdb-shipper` and skipped changing `tsdb` since `tsdb` index file names have checksums in them. However, as seen in issue #12500, we can hit the bug if compaction happens twice in the same unix second, and it ends up having the same contents in the compacted index(hence the same checksum as well).

Fixes #12500 

**Checklist**
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/shipper/indexshipper/tsdb/identifier.go
+++ b/pkg/storage/stores/shipper/indexshipper/tsdb/identifier.go
@@ -79,7 +79,7 @@ func (i SingleTenantTSDBIdentifier) Hash(h hash.Hash32) (err error) {
 func (i SingleTenantTSDBIdentifier) str() string {
 	return fmt.Sprintf(
 		"%d-%s-%d-%d-%x.tsdb",
-		i.TS.Unix(),
+		i.TS.UnixNano(),
 		compactedFileUploader,
 		i.From,
 		i.Through,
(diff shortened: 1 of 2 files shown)


salvacorts (approved):
LGTM. As far as I can tell, blooms only use `ParseSingleTenantTSDBPath`.

PR #12510, opened 2024-04-08
docs: Attempt to fix whatever broke the docs builds

JStickler:
 Something in PR #12478 broke the docs publication workflow.  The error messages on the action aren't particularly helpful, but these are the two files that were throwing errors when I attempted a local build. My current assumption is that there were either typos in the XREFs or broken links in that PR, so I've replaced all the relrefs in these two files with full URLs, which are now supported.


slim-bean (approved):

jdbaldry (commented):
`latest` should probably be `<LOKI_VERSION>` else all versions of Loki will link to "latest"

PR #12527, opened 2024-04-08
ci: add a github action for linting jsonnet

cstyan:
drone is still running our jsonnet linting job, we can migrate that to a github action

**NOTE: for some reason the drone job doesn't fail, but locally or via the new github action here the linting will fail. I may update our jsonnet files to match the linting in this PR.**


.github/workflows/lint-jsonnet.yml
```diff
@@ -0,0 +1,32 @@
...
+          go-version: '1.22.2'
+      - name: setup jsonnet
+        run: |
+          go install github.com/google/go-jsonnet/cmd/jsonnet@v0.20.0
+          go install github.com/google/go-jsonnet/cmd/jsonnetfmt@v0.20.0
+          go install github.com/google/go-jsonnet/cmd/jsonnet-lint@v0.20.0
+          go install github.com/monitoring-mixins/mixtool/cmd/mixtool@16dc166166d91e93475b86b9355a4faed2400c18
+          go install github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb@v0.5.1
+      - name: run linting
+        run: make BUILD_IN_CONTAINER=false lint-jsonnet
```
trevorwhitney, line 26: we're [already linting jsonnet](https://github.com/grafana/loki-release/blob/124c4d996f9625478a79f1884465e29ea082d224/.github/workflows/check.yml#L184) in the [check action](https://github.com/grafana/loki/actions/runs/8666525844/job/23767501298) that runs for every PR, so we should remove this check again here

trevorwhitney (changes requested):
Looks good other than the duplicate `lint-jsonnet` step. Can we remove that from here? If not we should move the mixin check to the other check pipeline.

.github/workflows/lint-jsonnet.yml
cstyan: IMO we should be moving as much as possible out of the giant `check` action, I can remove the linting from that action in this PR?

trevorwhitney (approved):

PR #12536, opened 2024-04-09
fix: make the tsdb filenames correctly reproducible from the identifier

sandeepsukhani:
In PR #12502, we updated tsdb code to use nanosecond precision in compacted file names.
However, this has caused problems in bloom compactor since the identifier generated while parsing tsdb filenames is not able to regenerate the original filename for old files with unix timestamp since it always uses nanoseconds for generating the filename.

I have fixed the issue by retaining the original timestamp from the file while parsing the names.

**Checklist**
- [x] Tests updated


pkg/storage/stores/shipper/indexshipper/tsdb/identifier.go
```diff
@@ -77,9 +81,13 @@ func (i SingleTenantTSDBIdentifier) Hash(h hash.Hash32) (err error) {
 
 // str builds filename with format <file-creation-ts> + `-` + `compactor` + `-` + <oldest-chunk-start-ts> + `-` + <latest-chunk-end-ts> `-` + <index-checksum>
 func (i SingleTenantTSDBIdentifier) str() string {
+	ts := i.origTS
+	if ts == 0 {
+		ts = i.TS.UnixNano()
```
salvacorts, line 86: nit: I find this a bit hacky. I think a bool flag `writeTSAsSecs`  would read better.

salvacorts (approved):
LGTM. Left a minor nit.

PR #12548, opened 2024-04-09
fix(helm): only default bucket names when using minio

trevorwhitney:

Only default bucket names in helm when using minio.

### DIFF
--- a/production/helm/loki/values.yaml
+++ b/production/helm/loki/values.yaml
@@ -311,6 +311,12 @@ loki:
     compactor_address: '{{ include "loki.compactorAddress" . }}'
   # -- Storage config. Providing this will automatically populate all necessary storage configs in the templated config.
   storage:
+    # Loki requires a bucket for chunks and the ruler. GEL requires a third bucket for the admin API.
+    # Please provide these values if you are using object storage.
+    # bucketNames:
+    #   chunks: FIXME
+    #   ruler: FIXME
+    #   admin: FIXME
     type: s3
     s3:
       s3: null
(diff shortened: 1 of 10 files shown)


production/helm/loki/values.yaml
```diff
@@ -311,10 +311,6 @@ loki:
     compactor_address: '{{ include "loki.compactorAddress" . }}'
   # -- Storage config. Providing this will automatically populate all necessary storage configs in the templated config.
   storage:
-    bucketNames:
-      chunks: chunks
-      ruler: ruler
-      admin: admin
```
slim-bean, line 317: instead of deleting this, maybe comment it out with FIXME for the names?

slim-bean (approved):
LGTM

trevorwhitney:
@luong-komorebi would you mind sharing a bit of information about your install? I think the goal here is to _intentionally_ break any installation using the default buckets with a cloud provider as you likely don't own those buckets and don't mean to send data there? Was your installation using the defaults, and if so, were you only using it to query recent data?

Farfaday:
Hi @trevorwhitney 

Commenting out `loki.storage.bucketNames` in values.yaml currently breaks helm lint, as the keys are not defined anymore but are called 9 times in the _helpers.tpl file.

Helm lint create errors of type  `nil pointer evaluating interface`.

I would propose adding a default value, something like
```
-  bucket_name: {{ $.Values.loki.storage.bucketNames.ruler }}
+  bucket_name: {{ $.Values.loki.storage.bucketNames.ruler | default "ruler" }}
```
What do you think? I can open an MR if you think that makes sense.

Thank you!

thiagowfx:
@Farfaday Do you plan to open a PR? :)

PR #12579, opened 2024-04-11
feat: add warnings to metadata context directly

MasslessParticle:
Looking up a context key can be surprisingly expensive. This PR makes it so we can add warnings to the `*metadata.Context` directly

### DIFF
--- a/pkg/logqlmodel/metadata/context.go
+++ b/pkg/logqlmodel/metadata/context.go
@@ -77,6 +77,12 @@ func (c *Context) Headers() []*definitions.PrometheusResponseHeader {
 	return headers
 }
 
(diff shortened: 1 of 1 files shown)


pkg/logqlmodel/metadata/context.go
```diff
@@ -77,6 +77,12 @@ func (c *Context) Headers() []*definitions.PrometheusResponseHeader {
 	return headers
 }
 
+func (c *Context) AddWarning(warning string) {
```
DylanGuedes, line 80: double checking: you're not using it, is that intentional?

ashwanthgoli (approved):
lgtm

PR #12599, opened 2024-04-12
fix(query sharding): Generalize avg -> sum/count sharding using existing binop mapper

MasslessParticle:
We make `avg` queries shardable by converting them into `sum/count` queries. The conversion didn't handle all possible mappings and sometimes results in the error `SelectSamples unimplemented: the query-frontend cannot evaluate an expression that selects samples. this is likely a bug in the query engine. please contact your system operator`. This happens even when inputting the native `sum/count` query would succeed. For example:

`avg by(foo_extracted) (quantile_over_time(0.95, {foo="baz"} | logfmt | unwrap foo_extracted | __error__=`` [$__auto]))` fails but `sum by(foo_extracted) (quantile_over_time(0.95, {foo="baz"} | logfmt | unwrap foo_extracted | __error__=`` [$__auto])) / count by(foo_extracted) (quantile_over_time(0.95, {foo="baz"} | logfmt | unwrap foo_extracted | __error__=`` [$__auto]))` succeeds.

This happens because our handling of top-level binary-operations is pretty sophisticated but the conversion from `avg -> sum/count` was only a piece of that. This PR removes the one-off mapping of binops in the `avg -> sum/count` case and replaces it with a bin op that is handed by our normal `mapBinOp` logic.

### DIFF
--- a/pkg/logql/shardmapper.go
+++ b/pkg/logql/shardmapper.go
@@ -240,33 +240,21 @@ func (m ShardMapper) mapVectorAggregationExpr(expr *syntax.VectorAggregationExpr
 
 		case syntax.OpTypeAvg:
 			// avg(x) -> sum(x)/count(x), which is parallelizable
-			lhs, lhsBytesPerShard, err := m.mapVectorAggregationExpr(&syntax.VectorAggregationExpr{
-				Left:      expr.Left,
-				Grouping:  expr.Grouping,
-				Operation: syntax.OpTypeSum,
-			}, r, false)
-			if err != nil {
-				return nil, 0, err
+			binOp := &syntax.BinOpExpr{
+				SampleExpr: &syntax.VectorAggregationExpr{
+					Left:      expr.Left,
+					Grouping:  expr.Grouping,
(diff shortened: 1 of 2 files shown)


pkg/logql/shardmapper.go
```diff
@@ -240,33 +240,21 @@ func (m ShardMapper) mapVectorAggregationExpr(expr *syntax.VectorAggregationExpr
...
-			// We take the maximum bytes per shard of both sides of the operation
-			bytesPerShard := uint64(max(int(lhsBytesPerShard), int(rhsBytesPerShard)))
-
-			return &syntax.BinOpExpr{
-				SampleExpr: lhs,
-				RHS:        rhs,
-				Op:         syntax.OpTypeDiv,
-			}, bytesPerShard, nil
-
+			return m.mapBinOpExpr(binOp, r, topLevel)
```
jeschkies, line 257: I would have sworn that I did this. Nice find.

jeschkies (approved):
I just went into the bin op mapping and found that @chaudum and myself fixed it a few months ago.

PR #12602, opened 2024-04-12
docs: Update Grafana Agent to Grafana Alloy

JStickler:
This PR updates the documentation to note that Grafana Agent has been deprecated in favor of Grafana Alloy.
Adds note that Promtail is feature complete.
Also fixes a few typos and updates a few cross-references to use full URLs.

I left the references to Grafana Agent in the following places:
* Helm charts documentation, because I don't think we've updated the charts to use Alloy yet.
* Migration documentation, as it is likely that users who are migrating would still be using the Grafana Agent.

`{{< docs/shared source="alloy" lookup="agent-deprecation.md" version="next" >}}` Will pull in a shared file from the Alloy docs that inserts the following Caution.  (Screen shot from Tempo docs, only the Caution will be imported)

AlloyCaution

### DIFF
--- a/docs/sources/get-started/labels/structured-metadata.md
+++ b/docs/sources/get-started/labels/structured-metadata.md
@@ -21,7 +21,7 @@ Structured metadata can also be used to query commonly needed metadata from log 
 
 You should only use structured metadata in the following situations:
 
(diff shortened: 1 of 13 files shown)


MichelHollands (approved):
LGTM

PR #12603, opened 2024-04-12
ci: add a github workflow to verify that the release workflows are correct

cstyan:
We need to ensure that the github workflows are updated correctly if we change the underlying version of the release repo we are using.


.github/jsonnetfile.json
```diff
@@ -8,7 +8,7 @@
           "subdir": "workflows"
         }
       },
-      "version": "main"
+      "version": "124c4d996f9625478a79f1884465e29ea082d224"
```
trevorwhitney, line 11: I think we should pin this to a tag in `grafana/loki-release` rather than a specific sha. I have a tag for each release branch in Loki, ie. [loki-2.8.x](https://github.com/grafana/loki-release/releases/tag/loki-2.8.x), so maybe we create one for `main` as well? Pinning to a tag allows us to change the release code without having to change Loki, which is nice when we want to fix a pipeline but rerun it for the same Loki commit.

.github/workflows/verify-release-workflow.yaml
```diff
@@ -0,0 +1,24 @@
...
+  check-release-changes:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Get changed files
+      # we need continue on error because the git diff | grep pipe can return a non-zero error code if no result is found
+      continue-on-error: true
+      id: changed-files
+      run: |
+            echo "changed_files=$(git diff --name-only main --  .github/workflows/ | grep releases | xargs)" >> $GITHUB_OUTPUT
```
trevorwhitney, line 13: where do we used `changed_files`? so we mean to block subsequent actions on it's presence?

trevorwhitney (commented):
Awesome, thanks for doing this Callum! Some small feedback.

.github/jsonnetfile.json
cstyan: A tag/release can be used here as well, I just don't know what we want to set it to. Do we have a tag/release that is what we should be using going forward, doesn't look like we have a tag for 3.x yet.

.github/workflows/verify-release-workflow.yaml
cstyan: woops, unnecessary copy/paste from the other action I based this one on, removing thins

trevorwhitney (approved):
LGTM! thanks callum!

PR #12630, opened 2024-04-16
chore: Restructure `indexgateway` component package

chaudum:

This PR moves code for the index gateway server and client components into the `pkg/indexgateway` directory, so it that the index gateway has its own top-level package like other components, such as ingester, distributor, querier, ...


vlad-diachenko (approved):
looks awesome. thanks @chaudum

owen-d (approved):

Issue #12645, opened 2024-04-17
Improve Level detection


cyriltovena:
We've seen case where the log would contains `Warn` as the level and an `error` message but it would still eventually end up as error because we don't properly search for `Warn` with a first capital level. This should be the very least.

We might want to investigate to support logfmt and json level, severity keys to make this more robust.

cc @shantanualsi @sandeepsukhani

grvsoni:
@cyriltovena 
I have some concerns raised by [topicuseducation](https://admin.grafana.com/orgs/topicuseducation) [ARR: $ 155,205.00
] somewhat in the same context. 

Their concern is, 
Loki's detection does not seem to work together with what Grafana is capable of. E.g. Grafana supports 'trace' level, but Loki auto detect does not.

For reference, https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/types/logs.ts#L32C2-L32C19, these are the canonical log levels (critical, warning, error, info, debug, trace en unknown) supported by Grafana (explore view). Loki only supports: debug, info, warn, error and critical (https://github.com/grafana/loki/blob/main/pkg/distributor/distributor.go#L947).

So, what to do with trace and unknown? Shouldn't Loki do the same as Grafana and support trace, as with unknown as default instead of info? Is Grafana supporting too much? And should we preprocess our log levels to match the ones from Loki (e.g. replace trace with debug, which is also less than ideal as debug != trace log lines)? So we can use 'pack' to embed a level hint, which is picked up by Loki and added as label.

They mentioned, they use the 'pack/unpack' option to embed the log level in many of their streams:

```
{"class":"MyClass","level":"TRACE","thread":"default task-1","_entry":"some log line"}
```

Now Loki auto detection adds a label 'level' with the value 'INFO'… Using unpack in Grafana does nothing with the level field embedded with pack, as the level label set by Loki is leading now.

They believe, The log level auto detection in Loki should match with the available log levels in Grafana (we could argue about unknown, but not to bothered about unknown vs info, trace being marked as info does bother me).

So, they want to be sure that Loki handle trace level as well.

bentonam:
I am working on Alloy modules for log processing porting what was done for module v1 in agent to alloy modules v2 this is what i'm currently doing to detect / default a log-level if it helps.  

https://github.com/grafana/alloy-modules/blob/log-modules/modules/utils/logs/log-levels.alloy

jtackaberry:
[...]

Issue #7944, opened 2022-12-15
Loki does not recognize log level `ERRO`


alextes:
**Is your feature request related to a problem? Please describe.**
It seems some applications like to make their log levels all four letters, meaning "warn", "info" are fine but "debug" and "error" break.
**Describe the solution you'd like**
Support these levels or custom level mapping somehow so log levels are appropriately recognized by Loki.

**Describe alternatives you've considered**
Match on the literal string `level: ERRO` then add a static label.

**Additional context**

Thanks for a great piece of monitoring!

alextes:
Using a replace stage might also work here I just realized!

alextes:
Yup, adding a stage like this works quite well:
```
- replace:
    expression: "(ERRO)"
    source: level
    replace: "ERROR"
```

It's good enough for me. Still, leaving this open. I'm noticing quite a few logging libraries, I think even all of golang likes to do things this way. It may still be worth supporting.

MichelHollands:
Thanks for reporting this and the solution you found. I'm going to close it as we can't accommodate every logging library's idiosyncrasies.

niluan304:
@alextes 
Excuse me, I hope to receive your help.

I hope to see `level=debug` in grafana instead of `unknown`
![image-20230320133045059](https://user-images.githubusercontent.com/51920864/226263685-12664701-f89f-4c5d-a71f-428d657e7164.png)
I try to add a stage , but it didn't work:

```yaml
- replace:
    expression: "(?P<level>DEBU)"
    source: level
    replace: "debug"
- replace:
    expression: "(?P<level>ERRO)"
    source: level
    replace: "error"    
```

Here are some logs:

```log
2023-03-20 13:17:21.625 [INFO] (?P<msg>)
2023-03-20 13:17:21.625 [INFO] {08762e5dd8084e176b6780683d9ae896} (?P<msg>)
2023-03-20 13:17:21.629 [DEBU] {6471af92d9084e178dfedd11c98d743d} (?P<msg>)
```

promtail  `config.yml`  more info:

```yaml
scrape_configs:

  - job_name: job
    static_configs:
    - labels:
        __path__: /var/log

    pipeline_stages:

    - multiline:
        firstline: '^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}.\d{3}'
        max_wait_time: 3s
    - regex:
[...]

PR #12651, opened 2024-04-17
fix: Fix log level detection

shantanualsi:
Currently detection of log levels is not truly case insensitive. For eg. "warn" or "WARN" is detected, however "Warn" is not. The implementation uses string.Contains over the entire log line to check for these levels. 

However, checking ignoring case would need us to toLower the entire log line which is inefficient.  This PR detects the `level` first -> `"level":` in case of JSON and `level=` in case of logfmt and then compares the next 12 characters after that ignoring case. 

The benchmarks show similar result on my computer with the fixed version at times few 100 ns faster. 

##### implementation from main branch:
image

##### implementation with Fix (This PR):
image

The PR also supports log level TRACE and FATAL and even 3 letter abbreviations that might be common. 

Edit:
Added more improvements after some discussions-  
1. Using logfmt and jsonparsers to parse each log line. After benchmarking, this performs slightly better than the previous implementation
2. Changed the label from level to detected_level so that it is clear for users
3. Support "Level", "Severity" and  "Lvl" in lowercase, uppercase and Pascal case.

Few caveats:
~- "level" should still be lowercase or block case ("LEVEL"). Documentation for configuration is updated. This does not support labels like Severity yet.~

Ref: https://github.com/grafana/loki/issues/12645

**Checklist**
- [x] Documentation added
- [x] Tests updated


pkg/distributor/distributor.go
```diff
@@ -904,42 +904,46 @@ func extractLogLevelFromLogLine(log string) string {
 	}
 
 	if firstNonSpaceChar == '{' && lastNonSpaceChar == '}' {
-		if strings.Contains(log, `:"err"`) || strings.Contains(log, `:"ERR"`) ||
-			strings.Contains(log, `:"error"`) || strings.Contains(log, `:"ERROR"`) {
+		levelIndex := strings.Index(log, `"level":`)
```
cyriltovena, line 907: problem is that it could be Level, LEVEL, serverity, Serverity.

pkg/distributor/distributor.go
```diff
@@ -904,42 +904,46 @@ func extractLogLevelFromLogLine(log string) string {
...
+		if levelIndex == -1 {
+			return logLevelInfo
+		}
+		compareString := log[levelIndex:min(len(log), levelIndex+20)]
```
cyriltovena, line 911: you might panic is +20 is out of bound

cyriltovena: ho I didn't see the min !

cyriltovena (commented):
I like the idea, I think we should may be support different field name ?

almost wonder if we should use some fast json and logfmt parser really to grab the key !

pkg/distributor/distributor.go
shantanualsi: yeah, i've mentioned it in the caveats in notes.. can we impose this in our documentation? or make this string configurable (case sensetive)?

kjq:
I may of filed this in the wrong place - https://github.com/grafana/grafana/issues/86447

shantanualsi:
I found an [old issue](https://github.com/grafana/loki/issues/7944#issuecomment-1698205383), a comment on which claims that 3 letter abbreviations for levels (dbg, wrn) are common as well (i haven't seen them yet). I've added support for them too in this PR. 

Also, supports "level" or "LEVEL" and updated the documentation.

@cyriltovena If everything else looks okay, we can merge this fix first?

To support different labels such as severity including any other cases of ``level`, I think it should be a part of configuration. We can add support for that in a separate PR.

shantanualsi:
> almost wonder if we should use some fast json and logfmt parser really to grab the key !

this is a good idea, i'll try and benchmark this. My hunch is that it's going to be **much** slower than the current bruteforce way...

cyriltovena:
> > almost wonder if we should use some fast json and logfmt parser really to grab the key !
> 
> this is a good idea, i'll try and benchmark this. My hunch is that it's going to be **much** slower than the current bruteforce way...

don't know about that I think it can be quite good. I'm specially worried for structured log.

pkg/distributor/distributor.go
trevorwhitney: what if we pass a hint to this function? for example, I think we could change [this code](https://github.com/grafana/loki/blob/ffe684c330bcd65f9b07a02d6f93bb475106becc/pkg/distributor/distributor.go#L378) to something like the following

```
firstEntry := stream.Entries[0]
firtsStructuredMetadata := logproto.FromLabelAdaptersToLabels(firstEntry.StructuredMetadata)
prevTs := firstEntry.Timestamp
addLogLevel := validationContext.allowStructuredMetadata &&
				validationContext.discoverLogLevels &&
				!lbs.Has(labelLevel)

logLevelField := detectLogLevelFieldFromLogEntry(firstEntry, firtsStructuredMetadata)
```

this would allow `detectLogLEvelFieldFromLogEntry` to do a case insensitive search for a bunch of different options (ie. `level`, `severity`, etc) since it was only doing it once for the whole stream per push. We could then pass that as a hint to `detectLogLevelFromLogEntry`, allowing it to short-circuit when passed.

trevorwhitney (commented):
What do you think about trying to detect the level field once per stream per push rather than on every line?

cyriltovena:
Actually think it's fine for doing this on every line, I suggest though we make a custom image that runs the detection without applying the change, to test in prod and see the impact.

shantanualsi:
I'm spending some time today using json-format and logfmt and benchmark.

shantanualsi:
is it entirely safe to assume that the label for "LEVEL" will be exactly the same in every line within a stream?

shantanualsi:
Turns out parsing is ~3x slower than the brute force way. 
Screenshot 2024-04-22 at 16 34 32

trevorwhitney:
> is it entirely safe to assume that the label for "LEVEL" will be exactly the same in every line within a stream?

I don't have data to support this, but given a stream is usually a single app, I would be very surprised if it wasn't. also, if we had to make this a requirement of this feature working correctly, I think it would be easy for customers to comply with. doing it per stream should also be cheaper for us to run, which is something to keep in mind as well.

shantanualsi:
Marking this as ready for review again as this PR makes sufficient improvements to level detection that we discussed. 

I think its difficult to cover all possibilities of label x level combinations (eg. some application somewhere might just use L=W or just [W], etc. The possibilities are endless. 

@trevorwhitney I haven't incorporated your suggestion yet. As we discussed, I think a stream could contain multiple labels for "level". We can detect once and fallback to scanning a line, but its difficult to quantify how much we'll gain from that. Open to suggestions here.

cyriltovena:
>Most of the memory allocation in logfmt seems to be coming from strings.NewReader() while initializing the decoder d := logfmt.NewDecoder(strings.NewReader(line)). I looked for options to improve this, however most of the projects use logfmt in a similar way.

You need a pool to reuse the reader. Let's see that later.

pkg/distributor/distributor.go
```diff
@@ -376,7 +386,8 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
 			n := 0
 			pushSize := 0
 			prevTs := stream.Entries[0].Timestamp
-			addLogLevel := validationContext.allowStructuredMetadata && validationContext.discoverLogLevels && !lbs.Has(labelLevel)
+
+			addLogLevel := validationContext.allowStructuredMetadata && validationContext.discoverLogLevels && !hasAnyLevelLabels(lbs)
```
cyriltovena, line 390: Since this is for standardisation I think that if the label exist and is found it should be still added a structured metadata under `detected_level` this way we can have a query that works for everyone.

pkg/distributor/distributor.go
```diff
@@ -876,74 +900,87 @@ func detectLogLevelFromLogEntry(entry logproto.Entry, structuredMetadata labels.
...
+	case "critical":
+		return logLevelCritical
+	case "fatal":
+		return logLevelFatal
+	default:
+		return detectLevelFromLogLine(log)
+	}
+}
+
+func getValueUsingLogfmtParser(line string) string {
```
cyriltovena, line 937: can you return a slice of bytes here to avoid alloc ? I think library both can ?

pkg/distributor/distributor.go
```diff
@@ -876,74 +900,87 @@ func detectLogLevelFromLogEntry(entry logproto.Entry, structuredMetadata labels.
...
+	equalIndex := strings.Index(line, "=")
+	if len(line) == 0 || equalIndex == -1 {
```
cyriltovena, line 939: cc @benclive do you think we could have something more resilient ?

pkg/distributor/distributor.go
```diff
@@ -385,10 +396,10 @@ func (d *Distributor) Push(ctx context.Context, req *logproto.PushRequest) (*log
 				}
 
 				structuredMetadata := logproto.FromLabelAdaptersToLabels(entry.StructuredMetadata)
-				if addLogLevel && !structuredMetadata.Has(labelLevel) {
+				if addLogLevel && !hasAnyLevelLabels(structuredMetadata) {
 					logLevel := detectLogLevelFromLogEntry(entry, structuredMetadata)
 					entry.StructuredMetadata = append(entry.StructuredMetadata, logproto.LabelAdapter{
```
cyriltovena, line 401: should we consider not adding any data if we don't find the level ?

pkg/distributor/distributor.go
benclive: Unfortunately not. According to the formal definition, basically everything is valid logfmt. That means that this filter can discard valid logfmt log lines if none of the keys had values. See [this fluentd issue](https://github.com/fluent/fluent-bit/issues/4927) for some discussion.
I reality, I think anything using logfmt would have at least one key=value pair so this is probably the best you can do to validate/guess.
https://pkg.go.dev/github.com/kr/logfmt#pkg-overview contains the definition.

pkg/distributor/distributor.go
shantanualsi: tried returning a `[]byte` but that always is causing more bytes allocated than returning `string`. 

**returning []byte**

Screenshot 2024-05-06 at 12 07 00

**returning string**

Screenshot 2024-05-06 at 12 07 49


My best guess is because string is immutable, a fixed capacity is allocated on the heap as opposed to []byte

shantanualsi:
sorry, ruined the PR by doing `pull --rebase`. Will try to reflog and revert or raise another PR with changes

pkg/distributor/distributor.go
cyriltovena: > Might break queries or rules if the label is absent for some log lines at random.

I can't think of an example, this label shouldn't be in used.

pkg/distributor/distributor.go
shantanualsi: added this change

pkg/distributor/distributor.go
shantanualsi: pushed the change to not set detected_level if unknown

PR #12655, opened 2024-04-17
test: Add new tests for Drain pattern matching

benclive:
* Adding new tests for Drain pattern matching
* We currently don't have any tests for this so these also act as examples and a starting point in order to begin improving the Drain pattern discovery.
* These tests aren't very exhaustive yet but already highlight some potential improvements such as trying to generate more specific patterns in some cases and less specific in others.


pkg/pattern/drain/drain_test.go
```diff
@@ -1 +1,147 @@
...
+		},
+		{
+			// Lower variation logs in json leads to a high number of patterns with very few matchers
+			name:      "Generate patterns on json formatted logs",
+			drain:     New(drainConfig),
+			inputFile: "testdata/drone-json.txt",
+			patterns: []string{
+				"<_> capacity <_>",
+				"<_> capacity changes <_>",
+				"{\"id\":\"D4Oh1ivB6cdLWa08\",\"level\":\"debug\",\"max-pool\":4,\"min-pool\":0,\"msg\":\"check capacity\",\"pending-builds\":0,\"running-builds\":0,\"server-buffer\":0,\"server-capacity\":0,\"server-count\":0,\"time\":\"2024-04-16T14:48:52Z\"}",
```
cyriltovena, line 89: You can use string with ` to avoid the pain of escaping.

pkg/pattern/drain/drain_test.go
```diff
@@ -1 +1,147 @@
...
+	//  > positions of the log message
+	LogClusterDepth: 8,
+	// SimTh is basically a ratio of matching/total in the cluster.
+	// Cluster tokens: "foo <*> bar fred"
+	//       Log line: "foo bar baz qux"
+	//                  *   *   *   x
+	// Similarity of these sequences is 0.75 (the distance)
+	// Both SimTh and MaxClusterDepth impact branching factor: the greater
+	// MaxClusterDepth and SimTh, the less the chance that there will be
+	// "similar" clusters, but the greater the footprint.
```
cyriltovena, line 33: Super cool thank you for that !

cyriltovena (commented):
Love it, happy to merge as is to incrementally move forward.

I think we could add more and more format as we go. Specially popular ones.

pkg/pattern/drain/drain_test.go
benclive: I can't take credit for this! I copied the descriptions and parameters over from https://github.com/grafana/loki/blob/491d2513e7469809d5b6962359305b9f7e6cb250/pkg/pattern/stream.go#L30 in order to get started. 
But now I think it makes more sense to unify it under the DefaultConfig function inside the drain package, at least until it becomes clear that we'd need different configs for different sets of logs.

cyriltovena (approved):
LGTM

PR #12660, opened 2024-04-17
ci: use gotestsum for running tests

trevorwhitney:

This will help us more easily identify the test runtime of each package so we can start to break them up more in CI and improve build times.


trevorwhitney:
we already set `GOTEST` in the docker image

PR #12661, opened 2024-04-17
ci: test each package separately

trevorwhitney:

This is an experiment to see if we can break up our tests by package


trevorwhitney:
this was used to test the CI change

PR #12664, opened 2024-04-17
feat(blooms): compute chunks once

owen-d:
[WIP] Computes chunks once during planning and passes them to the queriers, reducing a _lot_ of calls to the index.


pkg/logproto/logproto.proto
```diff
@@ -78,6 +78,9 @@ message QueryRequest {
   repeated string shards = 7 [(gogoproto.jsontag) = "shards,omitempty"];
   repeated Delete deletes = 8;
   Plan plan = 9 [(gogoproto.customtype) = "github.com/grafana/loki/v3/pkg/querier/plan.QueryPlan"];
+  // If populated, these represent the chunk references that the querier should
+  // use to fetch the data, plus any other chunks reported by ingesters.
+  ChunkRefGroup storeChunks = 10 [(gogoproto.jsontag) = "storeChunks"];
```
salvacorts, line 83: nit: I think this would be better placed inside the Plan.
Would placing it there might the refactoring more difficult?

chaudum (approved):

pkg/querier/queryrange/codec.go
```diff
@@ -2125,7 +2086,81 @@ func mergeLokiResponse(responses ...queryrangebase.Response) *LokiResponse {
 	}
 }
 
-// In some other world LabelRequest could implement queryrangebase.Request.
+func parseRangeQueryWithStoreChunksExtension(r *http.Request) (*LokiRequest, error) {
```
salvacorts, line 2089: I don't see this method being called conditionally of having chunks. I think this as well as the instant should be renamed to `parseRangeQuery` and `parseInstantQuery`

pkg/querier/queryrange/shard_resolver.go
```diff
@@ -265,37 +269,43 @@ func (r *dynamicShardResolver) ShardingRanges(expr syntax.Expr, targetBytesPerSh
 		if resp, ok := httpgrpc.HTTPResponseFromError(err); ok && (resp.Code == http.StatusNotFound) {
```
salvacorts, line 269: We still need this fallback for ingesters, right?

pkg/storage/stores/composite_store.go
```diff
@@ -28,7 +28,14 @@ type ChunkFetcherProvider interface {
...
 
 type ChunkFetcher interface {
-	GetChunks(ctx context.Context, userID string, from, through model.Time, predicate chunk.Predicate) ([][]chunk.Chunk, []*fetcher.Fetcher, error)
+	GetChunks(
+		ctx context.Context,
+		userID string,
+		from,
+		through model.Time,
+		predicate chunk.Predicate,
+		storeChunksOverride *logproto.ChunkRefGroup,
```
salvacorts, line 37: nit: I understand GetChunks does some extra filtering to the `storeChunksOverride` (validates and filters by the time range and fetcher). Still I think we should avoid calling `GetChunk` refs at all instead of passing the `storeChunksOverride` around. We may have a `GetFetcherForChunks` that returns grouped chunks by fetchers.

pkg/storage/store.go
```diff
@@ -411,7 +417,7 @@ func (s *LokiStore) lazyChunks(ctx context.Context, from, through model.Time, pr
 	stats := stats.FromContext(ctx)
 
 	start := time.Now()
-	chks, fetchers, err := s.GetChunks(ctx, userID, from, through, predicate)
+	chks, fetchers, err := s.GetChunks(ctx, userID, from, through, predicate, storeChunksOverride)
```
salvacorts, line 420: nit: Here's where I think we should do something like:
```go
var chunks [][]chunkRefs
var fetchers []fetchers
if storeChunksOverride > 0 {
   chunks, fetchers := s.GetFetcherForChunks(from, through, storeChunksOverride)
} else {
   chunks, fetchers := s.GetChunks(ctx, userID, from, through, predicate)
}
```

pkg/logql/shards.go
```diff
@@ -176,6 +203,16 @@ func (s Shard) Ptr() *Shard {
...
 }
 
+func (s Shard) Bind(chunks *logproto.ChunkRefGroup) *ShardWithChunkRefs {
+	res := &ShardWithChunkRefs{
+		Shard: s,
+	}
+	if chunks != nil {
+		res.chunks = chunks
+	}
+	return res
```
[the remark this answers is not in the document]
owen-d: Good call 👍

pkg/indexgateway/client.go
```diff
@@ -355,48 +352,11 @@ func (s *GatewayClient) GetShards(
 			return errCt <= maxErrs
 		},
 	); err != nil {
-		if isUnimplementedCallError(err) {
```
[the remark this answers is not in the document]
owen-d: No (the OSS repo is independent from our internal deployments at Grafana Labs). This is because the new `bounded` shards implementation is an intentional choice that needs to be turned on (default=power_of_two), so I removed the fallback for simplicity's sake.

pkg/logql/shards.go
```diff
@@ -176,6 +203,16 @@ func (s Shard) Ptr() *Shard {
 	return &s
 }
 
+func (s Shard) Bind(chunks *logproto.ChunkRefGroup) *ShardWithChunkRefs {
+	res := &ShardWithChunkRefs{
+		Shard: s,
+	}
+	if chunks != nil {
```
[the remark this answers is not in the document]
owen-d: You're right that there's a distinction -- `nil` indicates there is no override, but a zero-length chunk override is still valid and can happen when we lookup labels which don't map to chunks.

pkg/querier/multi_tenant_querier.go
```diff
@@ -448,3 +473,11 @@ func NewTenantSampleIterator(iter iter.SampleIterator, id string) *TenantSampleI
 func (i *TenantSampleIterator) Labels() string {
 	return i.relabel.relabel(i.SampleIterator.Labels())
 }
+
+func partitionChunkRefsByTenant(refs []*logproto.ChunkRef) map[string][]*logproto.ChunkRef {
+	filtered := make(map[string][]*logproto.ChunkRef)
+	for _, ref := range refs {
+		filtered[ref.UserID] = append(filtered[ref.UserID], ref)
```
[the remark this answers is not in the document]
owen-d: I'll probably skip this until I see it being a problem

pkg/querier/queryrange/shard_resolver.go
owen-d: I should probably remove this code for consistency with removing the same in `gatewayclient`

PR #10310, opened 2023-08-22
Series Index Store: fix race in GetSeries

bboreham:
**What this PR does / why we need it**:
~Lock around data that is modified from multiple goroutines concurrently.~
Replaced with the implementation from #9984 since @akhilanarayanan did it more efficiently.

**Which issue(s) this PR fixes**:
Relates to #8586

**Checklist**
- [x] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- NA Documentation added
- NA Tests updated
- [x] `CHANGELOG.md` updated
  - NA If the change is worth mentioning in the release notes, add `add-to-release-notes` label
- NA Changes that require user attention or interaction to upgrade are documented in `docs/sources/setup/upgrade/_index.md`
- NA For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)

PR #12674, opened 2024-04-18
fix: crrect initialization of a few slices

seiyab:
I found suspicious slice initialization so fixed them.
These slices are initialized by `make([]type, size)`. But their contents are filled by `slice = append(slice, newElem)`. It should result leading zero values.
Say,
```go
xs := make([]int, 3)
xs = append(xs, 1)
xs = append(xs, 2)
xs = append(xs, 3)
```
will result `xs = []int{0, 0, 0, 1, 2, 3}`. Probably it's unintentional and means
```go
xs := make([]int, 0, 3)
xs = append(xs, 1)
xs = append(xs, 2)
xs = append(xs, 3)
```

N/A. I consider it's a trivial fix.

Feel free to reject this PR.I'm sorry but I'm not familiar with Loki. I found it just scanning Loki with my experimental static checker.
I've read codes around these lines and callers of them. I suppose that they have confusion between length and capacity but I don't understand what they do in context of entire application.
Anyway, I suppose leading zero values look unintentional and tests pass.

**Checklist**
- [ ] ~Documentation added~
- [ ] Tests updated
    - Need you help if new tests are required
    - Need your help. I'm not sure. I guess it's not user-facing because tests passes without any change. My hypothesis is that zero values are dropped implicitly by callers.
- [ ] ~For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)~
- [ ] ~If the change is deprecating or removing a configuration option, update the `deprecated-config.yaml` and `deleted-config.yaml` files respectively in the `tools/deprecated-config-checker` directory. [Example PR](https://github.com/grafana/loki/pull/10840/commits/0d4416a4b03739583349934b96f272fb4f685d15)~


pkg/querier/ingester_querier.go
```diff
@@ -269,7 +269,7 @@ func (q *IngesterQuerier) TailersCount(ctx context.Context) ([]uint32, error) {
 		return nil, err
 	}
 
-	counts := make([]uint32, len(responses))
```
cstyan, line 272: this seems like a good change, AFAICT atm we could be allocating `# of ingesters * 2` uint32's since we could have a response from every ingester

pkg/storage/stores/series/series_index_store.go
```diff
@@ -305,7 +305,7 @@ func (c *IndexReaderWriter) chunksToSeries(ctx context.Context, in []logproto.Ch
 		return nil, err
 	}
 
-	results := make([]labels.Labels, len(chunksBySeries)) // Flatten out the per-job results.
+	results := make([]labels.Labels, 0, len(chunksBySeries)) // Flatten out the per-job results.
```
cstyan, line 308: this also seems safe, but actually I think it should be `len(perJobResults)` should it not? @bboreham I'm not sure from your original PR ([here](https://github.com/grafana/loki/pull/10310/files)) if you intended to create the slice of `chunkBySeries` length rather than `perJobResults` length

seiyab: No, `len(chunksBySeries)` looks correct. I also feel `len(perJobResults)` is correct at first sight.
Actually, argument of `append` is `innerSlice...`, not `innerSlice`. And sum of their length is probably `len(chunkBySeries)` (`groups := make([]chunkGroup, 0, len(chunksBySeries)/c.chunkBatchSize+1)`.

cstyan: Ah you're right, I misread the append line :+1: good find

cstyan (approved):

seiyab:
Thank you for kindful & speedy review!

PR #12688, opened 2024-04-19
fix(ksonnet): Do not generate rbac for consul if you are using memberlist

sentoz:

Added a condition that disables rbac creation for consul if memberlist is used

**Checklist**


sentoz:
Friends, I would like to consider the issue of adopting this change.
@chaudum @ashwanthgoli

cstyan (approved):
seems like a reasonable change :+1: since we haven't used consul ourselves in quite some time I didn't know what the sidekick was, but I see that the container itself is created by the `consul_deployment` so the rbac object got missed

PR #12703, opened 2024-04-19
feat: Add step param to Patterns Query API

### LINKED ISSUE #12689: Add step into the Pattern API
We should have a step parameter on that new API which aggregate into steps each point.

Very much like the range API.

Some example. https://github.com/grafana/pyroscope/blob/main/pkg/querier/querier.go#L1024

benclive:
* Adds an optional `step` parameter to the Patterns Query API
* This returns Pattern counts aggregated by the step duration instead of the raw samples we currently store.
  * Patterns continue to be stored in 10 second samples in the backend as this PR this only affects the read path.

Fixes #12689 

- There may be some aggregation code elsewhere in Loki that can be re-used here.
- Step is currently unused by the frontend but will default to a minimum of 10 seconds (the sample resolution) if not specified.

**Checklist**
- [x] Documentation added
- [x] Tests updated


pkg/pattern/drain/chunk.go
```diff
@@ -43,9 +43,9 @@ func (c Chunk) spaceFor(ts model.Time) bool {
 }
 
 // ForRange returns samples with only the values
-// in the given range [start:end).
-// start and end are in milliseconds since epoch.
-func (c Chunk) ForRange(start, end model.Time) []logproto.PatternSample {
+// in the given range [start:end) and aggregates them by step duration.
+// start and end are in milliseconds since epoch. step is a duration in milliseconds.
+func (c Chunk) ForRange(start, end, step model.Time) []logproto.PatternSample {
```
benclive, line 48: I'm not 100% sure this is the best place to add this functionality. It could easily be separated into a separate "groupByStep" function or similar.

pkg/pattern/drain/chunk_test.go
```diff
@@ -64,9 +64,9 @@ func TestForRange(t *testing.T) {
 		{
 			name: "No Overlap",
 			c: &Chunk{Samples: []logproto.PatternSample{
-				{Timestamp: 1, Value: 2},
-				{Timestamp: 3, Value: 4},
-				{Timestamp: 5, Value: 6},
+				{Timestamp: 2, Value: 2},
```
benclive, line 67: The stepping functionality shunts values into the step bucket before the sample TS, so these tests all had to be skewed down by to an even numbered millisecond as the step size is 2ms. Let me know if you don't agree with this behaviour.

pkg/pattern/drain/chunk.go
```diff
@@ -66,11 +66,37 @@ func (c Chunk) ForRange(start, end model.Time) []logproto.PatternSample {
 			return c.Samples[i].Timestamp >= end
 		})
 	}
-	return c.Samples[lo:hi]
+	if step == timeResolution {
```
cyriltovena, line 69: Shouldn't we support the case where the step is lower than the ingest resolution ? May be change the step to the ingest resolution in that case.

pkg/pattern/drain/chunk.go
cyriltovena: I think it's fine for now, the only problem I see is if someone uses multiple chunks with different start,end,step. Then it going to be a mess, but right now that's not the case so happy with this direction.

pkg/pattern/drain/chunk.go
```diff
@@ -66,11 +66,37 @@ func (c Chunk) ForRange(start, end model.Time) []logproto.PatternSample {
...
+		return c.Samples[lo:hi]
+	}
+
+	// Re-scale samples into step-sized buckets
+	currentStep := truncateTimestamp(c.Samples[lo].Timestamp, step)
+	aggregatedSamples := []logproto.PatternSample{
```
cyriltovena, line 75: nit: if you can figure in advance the amount of sample, then allocating the right amount of sample will be faster. Even if it's off by 1, growing will always perform poorly.

pkg/pattern/drain/chunk_test.go
cyriltovena: Sounds good, but you could have made the step configurable in the tests and default to previous behaviour.

pkg/pattern/instance.go
```diff
@@ -78,10 +79,14 @@ func (i *instance) Iterator(ctx context.Context, req *logproto.QueryPatternsRequ
 		return nil, httpgrpc.Errorf(http.StatusBadRequest, err.Error())
 	}
 	from, through := util.RoundToMilliseconds(req.Start, req.End)
+	step := req.Step
+	if step < 10*time.Second.Milliseconds() {
```
cyriltovena, line 83: Should this be more `step < timeResolution` ? Easy to miss in the future.

cyriltovena:
I think there updates requires in those files too.

https://github.com/grafana/loki/blob/main/pkg/loghttp/patterns.go#L9
https://github.com/grafana/loki/blob/main/pkg/querier/queryrange/codec.go#L999

The frontend in distributed mode is a huge minefields, I suggest you give a go and test it in dev to make sure it works correctly. Hit me up if you need help for doing so.

pkg/pattern/drain/chunk.go
benclive: This does support lower resolution steps - I added tests for that. This if statement is just to short-circuit the sample generation below for the happy case where we don't need to change the data.

The step param is defaulted to the ingest resolution in the API layer. IMO it makes the code harder to test if `step < TimeResolution` is checked here too, as all our tests need to be based on `TimeResolution` increments which would need updating if the resolution changes in the future. I'm happy to do that if you think its better that way though.

pkg/pattern/instance.go
benclive: Good idea. I exposed drain.TimeResolution to do this. IMO it makes sense to do this at the API layer than to default it later in the call stack where it's harder to surface what happened and why. I don't feel strongly about this though.

pkg/loghttp/patterns.go
```diff
@@ -15,6 +16,7 @@ func ParsePatternsQuery(r *http.Request) (*logproto.QueryPatternsRequest, error)
 	}
 	req.Start = start
 	req.End = end
+	req.Step = (time.Duration(defaultQueryRangeStep(start, end)) * time.Second).Milliseconds()
```
cyriltovena, line 19: Shouldn't the default only be used when there's not step query string ?

cyriltovena (approved):
Lgtm

The http parsing could use some tests but nothing blocking

PR #12726, opened 2024-04-22
feat: Querier: Split gRPC client into two.

DylanGuedes:
Split the gRPC client used by the querier into two, one for the communication with the scheduler, the other for communicating with the query-frontend.
- This change is retrocompatible: you don't have to change anything to keep existing behavior. 
- To configure the custom scheduler grpc client, you can use the new `query_scheduler_grpc_client` config or the new CLI flag `querier.scheduler-grpc-client`
- If you'd like to configure your frontend grpc client using a better named section, you can use the new `query_frontend_grpc_client` instead of the old `grpc_client_config`. Just make sure you don't use both at the same time, it will result in an error.

This work is necessary for configuring custom behavior between `querier<->scheduler` vs `querier<->frontend`. A use case is configuring mTLS when a different certificate is used by queriers, schedulers and frontends. You can only configure a single `server_name` with our current setup, making it impossible.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


docs/sources/shared/configuration.md
```diff
@@ -3557,10 +3557,23 @@ The `frontend_worker` configures the worker - running within the Loki querier -
 # CLI flag: -querier.id
 [id: <string> | default = ""]
 
-# The grpc_client block configures the gRPC client used to communicate between a
-# client and server component in Loki.
+# Configures the gRPC client used to communicate between the querier and the
```
salvacorts, line 3560: I'd reword this the other way around: first what;s the common use-case, later the note about use_separated_clients being true.

pkg/querier/worker/worker.go
```diff
@@ -92,9 +101,19 @@ type querierWorker struct {
...
 
 	metrics *Metrics
+
+	grpcClientConfig      grpcclient.Config
+	maxConcurrentRequests int
 }
 
 func NewQuerierWorker(cfg Config, rng ring.ReadRing, handler RequestHandler, logger log.Logger, reg prometheus.Registerer, codec RequestCodec) (services.Service, error) {
+	if !cfg.UseSeparatedGRPCClients {
+		level.Warn(logger).Log("msg", "Using the same client for frontend and scheduler. This is deprecated and will be removed in the future. Please use separated clients for frontend and scheduler.")
```
salvacorts, line 111: IF this is deprecated, we should add a note to the docs.

salvacorts: Are we sure we want to deprecate this? IIUC 99.% of the users will only need one grpc client. So we will require them to configure two identical grpc configs one for the frontend and another for the scheduler?

DylanGuedes: great point. https://github.com/grafana/loki/pull/12726/commits/35820954f379ea91a62b369a88615125b9d20ed9 reverts it.

docs/sources/shared/configuration.md
```diff
@@ -5329,83 +5328,83 @@ The `swift_storage_config` block configures the connection to OpenStack Object S
 
 ```yaml
 # OpenStack Swift authentication API version. 0 to autodetect.
-# CLI flag: -<prefix>.swift.auth-version
+# CLI flag: -swift.auth-version
```
salvacorts, line 5331: I think these shouldn't be here.

salvacorts (approved):
LGTM! Thank you Dylan :)

docs/sources/shared/configuration.md
```diff
@@ -543,19 +543,19 @@ The `alibabacloud_storage_config` block configures the connection to Alibaba Clo
 
 ```yaml
 # Name of OSS bucket.
-# CLI flag: -common.storage.oss.bucketname
```
DylanGuedes, line 546: this flag docs was bugged this whole time

PR #12731, opened 2024-04-22
docs: delete step to configure Minio from Helm docs

JStickler:

Deletes the step to configure MinIO as it is now configured by default in the Helm chart in Loki 3.0.


JStickler:
I was thinking MinIO, but it is memcached that is now configured by default.  D'oh!

PR #12741, opened 2024-04-23
fix(packaging): Require online network in systemd unit file for Loki and Promtail

### LINKED ISSUE #12724: After=network.target should be After=network-online.target
In all your service files, `After=network.target` should be `After=network-online.target`.  

Failure to do so will lead to flaky, unpredictable behaviour.

See `man systemd.special`, it says of `network.target` that:

> This unit is supposed to indicate when network functionality is available, but it is only very weakly defined

And the rest of the man-page section about `network.target` basically appears to imply you should only rely on it during the shutdown process, not during boot.

Meanwhile the `man systemd.special` on `network-online.target` is pretty clear, e.g.  :

> Units that strictly require a configured network connection should pull in network-online.target (via a Wants= type dependency) and order themselves after it.

chaudum:

From https://systemd.io/NETWORK_ONLINE/:

**How do I make sure that my service starts after the network is really online?**

That depends on your setup and the services you plan to run after it (see above). If you need to delay you service after network connectivity has been established, include

```systemd
After=network-online.target
Wants=network-online.target
```

in the `.service` file.

This will delay boot until the network management software says the network is “up”. For details, see the next question.

### Which issue(s) this PR fixes:

Closes #12724 

### Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/tools/packaging/loki.service
+++ b/tools/packaging/loki.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Loki service
-After=network.target
+After=network-online.target
+Wants=network-online.target
 
 [Service]
 Type=simple
--- a/tools/packaging/promtail.service
+++ b/tools/packaging/promtail.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Promtail service
-After=network.target
+After=network-online.target
+Wants=network-online.target
 
 [Service]
 Type=simple


MichelHollands (approved):
LGTM

PR #11502, opened 2023-12-16
Update installation.md: Install using apt or rpm

autoantwort:
**What this PR does / why we need it**:
Currently you suggest to download the files manually instead of using apt/rpm which means you don't get any updates. I thing using apt/rpm would be better. 

**Special notes for your reviewer**:

**Checklist**
- [x] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [x] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated
  - [ ] If the change is worth mentioning in the release notes, add `add-to-release-notes` label
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/setup/upgrade/_index.md`
- [ ] For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)
- [ ] If the change is deprecating or removing a configuration option, update the `deprecated-config.yaml` and `deleted-config.yaml` files respectively in the `tools/deprecated-config-checker` directory. [Example PR](https://github.com/grafana/loki/pull/10840/commits/0d4416a4b03739583349934b96f272fb4f685d15)

I admit I have not read the full https://grafana.com/docs/writers-toolkit/ guide, seems a lot of work for such a small change 😅

PR #12757, opened 2024-04-23
docs: fix broken link

JStickler:
Fixes broken link reported via docs @ grafana email.
Section added for 3.0 docs, does not exist in 2.9 docs.

Fixes malformed link that was introduced by #11502 and not fixed in #12677 because I didn't look closely enough at the link.  This time I've tested the link, so it should really be fixed this time.  🙄


MichelHollands (approved):
LGTM

PR #12763, opened 2024-04-23
fix(nix): lambda-promtail vendor hash

trevorwhitney:

Fixes the vendor hash for lamda-promtail, looks like the deps have been updated.

### DIFF
--- a/pkg/bloomgateway/processor.go
+++ b/pkg/bloomgateway/processor.go
@@ -63,7 +63,7 @@ func (p *processor) runWithBounds(ctx context.Context, tasks []Task, bounds v1.M
 	return nil
 }
 
-func (p *processor) processTasks(ctx context.Context, tenant string, day config.DayTime, keyspaces v1.MultiFingerprintBounds, tasks []Task) error {
+func (p *processor) processTasks(ctx context.Context, tenant string, day config.DayTime, _ v1.MultiFingerprintBounds, tasks []Task) error {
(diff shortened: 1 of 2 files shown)


owen-d (approved):

pkg/bloomgateway/processor.go
```diff
@@ -63,7 +63,7 @@ func (p *processor) runWithBounds(ctx context.Context, tasks []Task, bounds v1.M
 	return nil
 }
 
-func (p *processor) processTasks(ctx context.Context, tenant string, day config.DayTime, keyspaces v1.MultiFingerprintBounds, tasks []Task) error {
```
cstyan, line 66: I don't think a change here was intended?

cstyan (changes requested):

pkg/bloomgateway/processor.go
trevorwhitney: It was, the linter fails on the unused parameter

cstyan: nevermind, I missed that this was also failing the same Nix linting job

how was it not failing any other linting jobs?

trevorwhitney: I haven't quite got to the bottom of that, but it has something to do with how we only lint on changes (vs nix build linting on everything because there's not concept of version control in the nix build). however, that means these changes do sneak through somehow, and that's the part I haven't figured out. something small like this pops up on almost every small nix PR like this.

cstyan: strange, I still would have assumed the regular golangci-lint job would have failed on the PR that originally added this LOC

PR #12789, opened 2024-04-25
fix: loki-operational.libsonnet

QuentinBisson:
This PR fixes the  loki-operational.libsonnet mixin in case the value of per_cluster_label starts with cluster_.

For historical reasons, we named the cluster label cluster_id so generation actually replaces all cluster_id with cluster_id_id in this dashboard because the replace is incorrect

This PR fixes the loki operational dashboard when the cluster label starts with cluster_

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/loki-mixin/dashboards/loki-operational.libsonnet
+++ b/production/loki-mixin/dashboards/loki-operational.libsonnet
@@ -88,8 +88,8 @@ local utils = import 'mixin-utils/utils.libsonnet';
                                      'cluster="$cluster"',
                                      $._config.per_cluster_label + '="$cluster"'
                                    ),
-                                   'cluster_',
-                                   $._config.per_cluster_label + '_'
+                                   'cluster_job',
+                                   $._config.per_cluster_label + '_job'
                                  )
                                  else
                                    std.strReplace(


QuentinBisson:
@cstyan I would really appreciate the review :)

cstyan:
@QuentinBisson sorry, I'm failing to visualize in my head what the new value of the label replace would end up inserting as the label, can you help me out here with a more concrete example?

QuentinBisson:
Sure, so the issue is that if we set the per_cluster_label now to `cluster_id` then all occurrences of `cluster=$cluster` are perfectly replaced to `cluster_id=$cluster` but when the englobing replace them again then this résults in `cluster_id_id=$cluster`. Limiting thé englobing to cluster_job only renames thé recording rules to cluster_id_job which is what this replace was for

cstyan:
> Sure, so the issue is that if we set the per_cluster_label now to `cluster_id` then all occurrences of `cluster=$cluster` are perfectly replaced to `cluster_id=$cluster` but when the englobing replace them again then this résults in `cluster_id_id=$cluster`. 

Okay this makes sense, thanks :+1: 

> Limiting thé englobing to cluster_job only renames thé recording rules to cluster_id_job which is what this replace was for

can this not cause issues in other places though where we have just `cluster`? I supposed the earlier replace directives handle those cases?

QuentinBisson:
The previous one should be enough but i'll run it an test. From what I understood, there are 4 use cases:
- cluster=$cluster
- cluster=~$cluster
- cluster in templating or by clauses which i'm or sure about
- cluster_job recording rules.
Let me generate this on monday to assert all cluster labels are fixed. This dashboard importing raw json is not the easiest

QuentinBisson:
@cstyan I checked and there is no other use caes for cluster apart from those 3. The one used in the templating section is managed via the addCluster libsonnet function

cstyan (approved):
LGTM, thanks for digging in and providing an explanation for your changes :+1:

PR #12795, opened 2024-04-25
fix(promtail): Fix bug with Promtail config reloading getting stuck indefinitely

ptodev:

Recently, a memory issue was reported with the Agent Static mode. The memory of the Agent was creeping up steadily, until it eventually OOMs. That Agent was having its config reloaded every 30 seconds.

A goroutine dump indicated that these calls have been taking a long time:
```
goroutine 152484 [chan receive, 1214 minutes]:
github.com/grafana/loki/clients/pkg/promtail/targets/file.(*FileTarget).Stop(...)
	/go/pkg/mod/github.com/grafana/loki@v1.6.2-0.20231004111112-07cbef92268a/clients/pkg/promtail/targets/file/filetarget.go:159

goroutine 152424 [chan send, 1220 minutes]:
github.com/grafana/loki/clients/pkg/promtail/targets/file.(*FileTarget).startWatching(0xc0030ab5f0, 0xc002d9fe48?)
	/go/pkg/mod/github.com/grafana/loki@v1.6.2-0.20231004111112-07cbef92268a/clients/pkg/promtail/targets/file/filetarget.go:314 +0x20a

goroutine 152426 [chan send, 1220 minutes]:
github.com/grafana/loki/clients/pkg/promtail/targets/file.(*FileTarget).startWatching(0xc0030ab6c0, 0xc002e31e48?)
	/go/pkg/mod/github.com/grafana/loki@v1.6.2-0.20231004111112-07cbef92268a/clients/pkg/promtail/targets/file/filetarget.go:314 +0x20a

goroutine 152428 [chan send, 1210 minutes]:
github.com/grafana/loki/clients/pkg/promtail/targets/file.(*FileTarget).stopWatching(0xc0030ab790, 0xc002da1d88?)
	/go/pkg/mod/github.com/grafana/loki@v1.6.2-0.20231004111112-07cbef92268a/clients/pkg/promtail/targets/file/filetarget.go:327 +0x20a
```

What is probably happening is that `FileTargetManager` begins a `Stop()`, but doesn't yet close the `targetEventHandler` channel. As a result, `startWatching` and `stopWatching` seem stuck with sending on the channel. This causes the `sync` call to never complete, which on the other hand means that the `FileTarget`'s `Stop()` function can't complete. 

The memory build up is probably due to lots of calls to the config reload function which never complete.

cc @paul1r who recently committed [similar fixes](https://github.com/grafana/loki/pull/12656).

Should I add a changelog entry? And do you think there is a way to test this? Also, I haven't yet tested with the customer. If you think the code looks ok, we could merge it and verify later that it does fix the customer issue?

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/clients/pkg/promtail/targets/file/filetarget.go
+++ b/clients/pkg/promtail/targets/file/filetarget.go
@@ -348,7 +348,6 @@ func (t *FileTarget) startWatching(dirs map[string]struct{}) error {
 			path:      dir,
 			eventType: fileTargetEventWatchStart,
 		}:
-			// continue
 		}
 	}
 	return nil
@@ -368,7 +367,6 @@ func (t *FileTarget) stopWatching(dirs map[string]struct{}) error {
 			path:      dir,
 			eventType: fileTargetEventWatchStop,
 		}:
-			// continue
 		}
 	}
 	return nil
--- a/clients/pkg/promtail/targets/file/filetarget.go
+++ b/clients/pkg/promtail/targets/file/filetarget.go
@@ -25,6 +25,8 @@ const (
 	FilenameLabel = "filename"
 )
 
+var errFileTargetStopped = errors.New("File target is stopped")
+
 // Config describes behavior for Target
 type Config struct {
 	SyncPeriod time.Duration `mapstructure:"sync_period" yaml:"sync_period"`
@@ -223,6 +225,11 @@ func (t *FileTarget) run() {
 			}
 		case <-ticker.C:
 			err := t.sync()
+			if errors.Is(err, errFileTargetStopped) {
+				// This file target has been stopped.
+				// This is normal and there is no need to log an error.
+				return
+			}
 			if err != nil {
 				level.Error(t.logger).Log("msg", "error running sync function", "error", err)
 			}
@@ -291,14 +298,20 @@ func (t *FileTarget) sync() error {
 	t.watchesMutex.Lock()
 	toStartWatching := missing(t.watches, dirs)
 	t.watchesMutex.Unlock()
-	t.startWatching(toStartWatching)
+	err := t.startWatching(toStartWatching)
+	if errors.Is(err, errFileTargetStopped) {
+		return err
+	}
 
 	// Remove any directories which no longer need watching.
 	t.watchesMutex.Lock()
 	toStopWatching := missing(dirs, t.watches)
 	t.watchesMutex.Unlock()
 
-	t.stopWatching(toStopWatching)
+	err = t.stopWatching(toStopWatching)
+	if errors.Is(err, errFileTargetStopped) {
+		return err
+	}
 
 	// fsnotify.Watcher doesn't allow us to see what is currently being watched so we have to track it ourselves.
 	t.watchesMutex.Lock()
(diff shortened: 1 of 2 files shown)


cstyan:
> Should I add a changelog entry? Also, I haven't yet tested with the customer. 

Changelog entries will be auto generated now via conventional commit format for the PR title, see the failed check [here](https://github.com/grafana/loki/actions/runs/8836872358/job/24264414313?pr=12795)

>  And do you think there is a way to test this?

We definitely need a test. I imagine if we added a test that calls `Stop()` or `startWatchers/stopWatchers` in a goroutine and then used a timeout to fail the test if those functions didn't return within X seconds it _should_ fail without the changes in your current commit here, and pass with your changes.

> If you think the code looks ok, we could merge it and verify later that it does fix the customer issue?

It looks okay but why can't we at least verify out of band before we consider merging? I would assume that if the agent or alloy codebase is still pulling in upstream promtail code then this change could be hacked in somehow (via go mod I guess since I think you guys are not using a vendor directory) and deployed so that we can trigger a config reload and see if there's still a deadlock.

ptodev:
@cstyan thank you so much for the quick and thorough feedback!

The problem with the test is that we need to make sure `FileTarget` has already started a `sync`, but has not yet sent all its data to the channel. If we call `Stop()` after it already sent the data on the channel, or before it kicked off a sync, then the test wouldn't be valid.

I updated the PR with a test which I believe works.

> It looks okay but why can't we at least verify out of band before we consider merging?

I'm just not sure if I can replicate the circumstances required for this bug in real life. I think it's most likely to replicate if there is a long list of directories to watch, and a very quick config reload frequency. I could try replicating it next week, but if I'm not successful in a few hours I think we should just merge the PR. I do believe that the PR fixes a real bug anyway.

ptodev:
> The problem with the test is that we need to make sure `FileTarget` has already started a `sync`, but has not yet sent all its data to the channel. 

One way to do this is to call `sync` directly, just like some other tests do. However, I want to avoid this because I don't want to make assumptions about what `sync`'s internals are.

clients/pkg/promtail/targets/file/filetarget_test.go
```diff
@@ -336,6 +336,90 @@ func TestFileTarget_StopsTailersCleanly_Parallel(t *testing.T) {
...
+	// Wait until the file target is waiting on a channel send due to a full channel buffer.
+	requireEventually(t, func() bool {
+		return len(fakeHandler) == 1
+	}, "expected an event in the fakeHandler channel")
+
+	// If FileHandler works well, then it will stop waiting for
+	// the blocked fakeHandler and stop cleanly.
+	// This is why this time we don't drain fakeHandler.
+	target.Stop()
+	ps.Stop()
```
cstyan, line 414: This test still isn't 100% correct IMO.

If you copy/paste it and execute it against the current main Loki branch, it will never succeed or fail except due to the general 10 minute timelimit on `go test` runs, because of the same original issue that `Stop` calls have a deadlock currently. 

It should fail within a few seconds against `main`, and then pass almost instantaneously with the change from your original commit.

cstyan (changes requested):

clients/pkg/promtail/targets/file/filetarget_test.go
ptodev: Just to clarify - you are stating that the test is not correct because it fails due to a timeout, and not because it fails due to a condition somewhere in the test? So you'd like me to make sure we fail the test if Stop() hands for more than a few seconds? I think that's a reasonable ask, but I just want to make sure this is what you'd like me to do.

cstyan: > Just to clarify - you are stating that the test is not correct because it fails due to a timeout, and not because it fails due to a condition somewhere in the test? So you'd like me to make sure we fail the test if Stop() hangs for more than a few seconds? I think that's a reasonable ask, but I just want to make sure this is what you'd like me to do.

More or less, yes. 

Right now against `main` the test fails because in essence the test never actually finishes and reaches an actual success or failure condition. It just reaches the default timeout for `go test` which is 10 minutes. 

The goal of your change is to make `Stop` and related functions not block deadlock via allowing select on the closing of the related channel. By wrapping these two `Stop()` calls in a go routine and then changing the last `require.NoError` to a wrapper with `require.Eventually` the test should fail due to `Stop` not returning in a timely manner, and then with your change the test would pass.

ptodev: I updated the test to wrap `Stop()` with `requireEventually`. I don't think there is a need to run `Stop()` in a goroutine.

cstyan: > I updated the test to wrap Stop() with requireEventually. I don't think there is a need to run Stop() in a goroutine.

your solution is simpler :+1:

clients/pkg/promtail/targets/file/filetarget.go
```diff
@@ -321,32 +334,44 @@ func (t *FileTarget) sync() error {
...
 		level.Info(t.logger).Log("msg", "removing directory from watcher", "directory", dir)
-		t.targetEventHandler <- fileTargetEvent{
+		select {
+		case <-t.quit:
+			return errFileTargetStopped
+		case t.targetEventHandler <- fileTargetEvent{
 			path:      dir,
 			eventType: fileTargetEventWatchStop,
+		}:
+			// continue
```
cstyan, line 371: the bracket here seems wrong? and we don't need the continue

ptodev: Why would it be wrong? We are sending a `fileTargetEvent` to the channel.

The point of the "continue" is just to indicate that we are continuing the loop with the next item. If you think it's obvious, then I could remove the comment :)

cstyan: :facepalm: my mistake, diff was rendering strangely on my screen because my browser was too small. Yeah, lets remove the comment since we're not explicitly `continue`'ing, nor do we want to.

cstyan (commented):
approved :+1: just one last nit to be fixed

:+1: thanks for your patience and continued effort with various promtail upstream work @ptodev

cstyan (approved):

PR #12799, opened 2024-04-26
feat: Generic logline placeholder replacement and tokenization

na--:

This code allows us to preprocess generic logs and replace highly variable dynamic data (timestamps, IPs, numbers, UUIDs, hex values, bytesizes and durations) with static placeholders for easier pattern extraction and more efficient and user-friendly matching by the Drain algorithm.

Additionally, there is logic that splits generic log lines into discrete tokens that can be used with Drain for better results than just naively splitting the logs on every whitespace. The tokenization here handles quote counting and emits quoted strings as a part of the same token. On the other side, it also handles likely JSON logs without any white spaces in them better, by trying to split `{"key":value}` pairs (without actually parsing the JSON).

All of this is done without using regular expressions and without actually parsing the log lines in any specific format. That's why it works very efficiently in terms of CPU usage and alloctions, and should handle all log formats and unformatted logs equally well.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


pkg/pattern/tokenization/tokenization.go
```diff
@@ -0,0 +1,1054 @@
...
+		}
+		n = r.peekNext4()
+	}
+
+	if n[0] == 's' && boundryChars[n[1]] {
+		secsLen = 1
+	} else if n[1] == 's' && (n[0] == 'm' || n[0] == 'n' || n[0] == 'u') && boundryChars[n[2]] {
+		secsLen = 2
+	} else if n[2] == 's' && ((n[0] == 0xC2 && n[1] == 0xB5) || (n[0] == 0xCE && n[1] == 0xBC)) && boundryChars[n[3]] {
+		// This checks for the unicode "µs" (U+00B5 = micro symbol) and "μs" (U+03BC = Greek letter mu)
```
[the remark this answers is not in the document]
na--: That's where I actually got both from, see [this](https://github.com/golang/go/blob/master/src/time/format.go#L1573-L1582) :sweat_smile:

pkg/pattern/tokenization/tokenization.go
```diff
@@ -0,0 +1,1054 @@
...
+
+	r.head += secsLen
+	return true
+
+restore: // should be faster than a defer
+	r.head = curHead
+	return false
+}
+
+var byteSizes = [256]bool{'k': true, 'K': true, 'm': true, 'M': true, 'g': true, 'G': true, 't': true, 'T': true}
```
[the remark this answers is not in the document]
na--: :thinking: Yeah, it doesn't hurt to add PB in there, though probably not a lot of folks work at the scale where it'll be needed :sweat_smile:

pkg/pattern/tokenization/tokenization.go
```diff
@@ -0,0 +1,1054 @@
...
+			r.emit(false, placeholderTimestamp)
+			return true
+		}
+		r.resetHead()
+
+		// ANSIC, UnixDatem, RubyDate e.g
+		// - "Mon Jan  2 15:04:05 2006"
+		// - "Mon Jan  2 15:04:05 MST 2006"
+		// - "Mon Jan 02 15:04:05 -0700 2006"
+		if r.advanceChar(' ') && r.advanceMonthName() && r.advanceChar(' ') && (r.advanceChar(' ') || true) && r.advanceUintUpTo(31, 2) && r.advanceChar(' ') && r.advanceTime(false) && r.advanceStringOrNumericTimeZone(true) && r.advanceChar(' ') && r.advanceYear() {
```
ashwanthgoli, line 692: `&& (r.advanceChar(' ') || true)` this is clever 😻

pkg/pattern/tokenization/tokenization.go
```diff
@@ -0,0 +1,1054 @@
...
+// Integer numbers after these words won't be replaced by `<NUM>`.
+var numPreservingKeys = [][]byte{[]byte("status"), []byte("status_code"), []byte("httpStatus")}
+
+const (
+	placeholderEndOfLine = "<...>"
+	maxYear              = 2100 // anything above that is unlikely to be a timestamp...
+	minHexLen            = 12   // 48 bits min for free-standing hex strings (e.g "0123456789ab") or 42 bits for "0xABCDEF1234" strings
+)
+
+var boundryChars = [256]bool{}
```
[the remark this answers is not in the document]
na--: Resolved by https://github.com/grafana/loki/pull/12799/commits/41e45b7488f07a7502894fbade232405a9a233d0

na--:
Sorry if this makes the code review harder, but I decided that now was probably the best moment to split the `tokenizer` and `replacer` code in different files with https://github.com/grafana/loki/pull/12799/commits/c62c6cfb746392a31e9dad52cd04cca4bf7cd152, doing so after merging this PR would have just added more noise

pkg/pattern/tokenization/tokenization.go
na--: One of the potential minor performance improvements I want to benchmark at some future point is actually getting rid of `advanceOneOf`. We only use it in a couple of places, here: https://github.com/grafana/loki/blob/c62c6cfb746392a31e9dad52cd04cca4bf7cd152/pkg/pattern/tokenization/replacer.go#L191
and here: https://github.com/grafana/loki/blob/c62c6cfb746392a31e9dad52cd04cca4bf7cd152/pkg/pattern/tokenization/replacer.go#L535

And both of these can be easily rewritten like this:
```go
(r.advanceChar('+') || r.advanceChar('-'))
(r.advanceChar('T') || r.advanceChar(' '))
```

I'll actually comment on a few other potential minor performance improvements I've thought of but haven't had time to benchmark, just so I don't forget them in the future :sweat_smile:

pkg/pattern/tokenization/replacer.go
```diff
@@ -0,0 +1,859 @@
...
+// check the length!
+func (r *replacer) advanceUintRet() (val uint, length uint) {
+	var c byte
+	for ; r.head < len(r.source); r.head++ {
+		c = r.source[r.head]
+		if c < '0' || '9' < c {
+			break
+		}
+		length++
+		val = val*10 + uint(c-'0') // TODO: use bitwise trick?
```
na--, line 96: This is what I meant by "bitwise trick": https://www.guidodiepen.nl/2017/03/clever-bit-tricks-with-ascii/#integervaluesinascii

I doubt it's actually faster on modern CPU architectures (and it's certainly less readable), but worth benchmarking to confirm, just in case :sweat_smile:

pkg/pattern/tokenization/replacer.go
```diff
@@ -0,0 +1,859 @@
...
+func (r *replacer) wasNumPreservingKeyword() bool {
+	for _, key := range numPreservingKeys {
+		pPos := r.cur - 1 - len(key)
+		if pPos < -1 {
+			return false // all subsequent keys are longer
+		}
+		if pPos != -1 && !boundaryChars[r.source[pPos]] {
+			continue
+		}
+		if bytes.HasPrefix(r.source[pPos+1:], key) {
```
na--, line 765: This probably can be optimized, the `numPreservingKeys` values can be of a `string` type and if this check is something like `string(r.source[pPos+1:pPos+len(key)]) == key)`, it might be faster. This should not result in an actual string allocation because of [some Go compiler optimizations](https://github.com/golang/go/blob/5419f652b6593a0916f7b2155435b945e8ee0fb4/src/runtime/string.go#L136-L150) for cases like this. Besides, this is basically what `bytes.HasPrefix` does internally as well, just with an extra bounds check and function calls... :sweat_smile:  Which are probably inlined, but it can't hurt to compare the values directly :man_shrugging:

pkg/pattern/tokenization/replacer.go
```diff
@@ -0,0 +1,859 @@
...
+
+		// RFC850, e.g.
+		//  - "Monday, 02-Jan-06 15:04:05 MST"
+		backtrackedSlice := r.source[r.head-3:]
+		var matchedDay []byte
+		for _, dw := range longDayNames {
+			if bytes.HasPrefix(backtrackedSlice, dw) {
+				matchedDay = dw
+				break
+			}
```
na--, line 713: If `longDayNames` was a `map[[3]byte]`, the lookup might be a bit faster, though not sure if that matters a lot for a slice of 7 elements :sweat_smile: Still, while this specific part of the code is probably executed fairly rarely, it couldn't hurt to benchmark :sweat_smile:

pkg/pattern/tokenization/replacer.go
```diff
@@ -0,0 +1,859 @@
...
+	maxYear   = 2100 // anything above that is unlikely to be a timestamp...
+	minHexLen = 12   // 48 bits min for free-standing hex strings (e.g "0123456789ab") or 42 bits for "0xABCDEF1234" strings
+)
+
+var boundaryChars = [256]bool{}
+
+func init() {
+	for i := 0; i < 128; i++ { // for now, we don't consider non-ASCII chars as boundary
+		if i < '0' || (i > '9' && i < 'A') || (i > 'Z' && i < 'a') || i > 'z' {
+			boundaryChars[i] = true
```
na--, line 31: I am not certain if the slice lookups like `if boundaryChars[c] { /* do something */}` in the code below are actually faster than if we had a simple function like this:
```go
func isBoundary(c byte) {
	return c < '0' || (c > '9' && c < 'A') || (c > 'Z' && c < 'a') || (c > 'z' && c < 128)
}
```

I suspect the function will be faster because of how CPU caching works, but I haven't actually benchmarked it. I adopted this approach originally because I had a couple of exceptions here (e.g. I didn't consider `_` to be a boundary character initially).

ashwanthgoli (approved):
LGTM 🚀 thanks again for this @na-- 

not for this pr, but it might be worth having separate tests for the replacer and tokeniser, helps narrowing down the problem if we see any bugs down the line.

pkg/pattern/tokenization/replacer.go
```diff
@@ -0,0 +1,859 @@
...
+		c = r.source[r.cur]
+		switch {
+		// If we're currently not at a boundary, the only thing we need to check
+		// is whether the current char would create a boundary in the next iteration.
+		case !onBoundary:
+			onBoundary = boundaryChars[c]
+
+			// We weren't at a boundary and now we are, so check if we match one
+			// of the special keywords that will preserve numbers
+			if onBoundary {
```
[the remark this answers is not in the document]
na--: Hmm :thinking: Maybe, though I am not sure if this will always hold. For example, the `_` in `status_code=` is technically a boundary character, so it's not inconceivable for `status` to already be "flushed" when we reach the `=`.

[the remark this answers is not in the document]
na--: Yeah, just stepped through the `status_code=123` example with a debugger and that's exactly what happens, because `_` is a boundary and `c` is a potential start of a hex string, we encounter this branch and copy `status_` to the `r.dest` slice (so `r.cur - r.tail` becomes 0) :disappointed: 

Let's keep this thread open though, as a potential performance improvement along with the other ones I opened, because that's an implementation detail that's fixable. If profiling proves that this would make a big difference, we can change the other behavior to be less aggressive in copying the source string

PR #12806, opened 2024-04-26
fix(blooms): Reset error on LazyBloomIter.Seek

salvacorts:

We are skipping all pages after any of the pages on a given block have failed due to being too big:
1. `Seek` a page that is too big. `it.err` is set here  https://github.com/grafana/loki/blob/feb210b4c4250fee622661ce349aa6843c20a394/pkg/storage/bloom/v1/bloom_querier.go#L66-L68

2. then `it.Next` return `false` in this check https://github.com/grafana/loki/blob/feb210b4c4250fee622661ce349aa6843c20a394/pkg/storage/bloom/v1/bloom_querier.go#L87-L89.

3. `Seek` another page that should be valid. Decoding of the new page succeeds but `Seek` won’t reset `it.err`. Therefore, the call to `it.Next` return `false` again.

This PR fixes this by resetting `it.err` on every `Seek`.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/storage/bloom/v1/bloom_querier.go
+++ b/pkg/storage/bloom/v1/bloom_querier.go
@@ -45,6 +45,9 @@ func (it *LazyBloomIter) ensureInit() {
 func (it *LazyBloomIter) Seek(offset BloomOffset) {
 	it.ensureInit()
 
+	// reset error from any previous seek/next
+	it.err = nil
+
 	// if we need a different page or the current page hasn't been loaded,
 	// load the desired page
 	if it.curPageIndex != offset.Page || it.curPage == nil {
--- a/pkg/storage/bloom/v1/bloom_querier.go
+++ b/pkg/storage/bloom/v1/bloom_querier.go
@@ -45,8 +45,10 @@ func (it *LazyBloomIter) ensureInit() {
 func (it *LazyBloomIter) Seek(offset BloomOffset) {
 	it.ensureInit()
 
-	// reset error from any previous seek/next
-	it.err = nil
+	// reset error from any previous seek/next that yield pages too large
+	if errors.Is(it.err, ErrPageTooLarge) {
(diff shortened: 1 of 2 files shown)


pkg/storage/bloom/v1/bloom_querier.go
```diff
@@ -45,6 +45,9 @@ func (it *LazyBloomIter) ensureInit() {
 func (it *LazyBloomIter) Seek(offset BloomOffset) {
 	it.ensureInit()
 
+	// reset error from any previous seek/next
+	it.err = nil
```
chaudum, line 49: I don't think we should reset the error unconditionally, e.g we should not continue seeking after IO error.
Instead, we probably want to check the error type, similar to what the do here https://github.com/grafana/loki/pull/12297/files#diff-ad299f9ea738f6f6d8e1a3b9fcc651506b3a241dbfbd95a60ab9d96b5b8b4d4aR156

salvacorts: Yes, I agree. Thank you!

pkg/storage/bloom/v1/fuse_test.go
```diff
@@ -130,6 +145,92 @@ func TestFusedQuerier(t *testing.T) {
...
+
+		require.True(t, querier.series.Next())
+		series := querier.series.At()
+		require.Equal(t, fp, series.Fingerprint)
+
+		querier.blooms.Seek(series.Offset)
+
+		if fp == 0 || fp == 2 {
+			require.False(t, querier.blooms.Next())
+			require.Error(t, querier.blooms.Err())
```
chaudum, line 225: Should we also check the error type here?

chaudum (approved):
Otherwise LGTM

PR #12807, opened 2024-04-26
fix(blooms): Fully deduplicate chunks from FilterChunkRef responses

chaudum:

This PR aims for full de-duplication of chunks and series from filter requests from the index gateway to the bloom gateway.

Whenever we merge/de-duplicate slices, the inputs need to be sorted. It appears that the `Removals` (chunks) from the `v1.Output` are not guaranteed to be sorted. 

When comparing ShortRefs, both `From`, `Through`, and `Checksum` need to be used.

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/bloomgateway/bloomgateway.go
+++ b/pkg/bloomgateway/bloomgateway.go
@@ -44,6 +44,7 @@ package bloomgateway
 import (
 	"context"
 	"fmt"
+	"slices"
 	"sort"
 	"time"
 
@@ -383,8 +384,10 @@ func (g *Gateway) consumeTask(ctx context.Context, task Task, tasksCh chan<- Tas
 		case <-ctx.Done():
 			// do nothing
 		default:
-			// chunks may not be sorted
-			sort.Slice(res.Removals, func(i, j int) bool { return res.Removals[i].Less(res.Removals[j]) })
+			// chunks may not always be sorted
+			if !slices.IsSortedFunc(res.Removals, func(a, b v1.ChunkRef) int { return a.Cmp(b) }) {
+				slices.SortFunc(res.Removals, func(a, b v1.ChunkRef) int { return a.Cmp(b) })
+			}
 			task.responses = append(task.responses, res)
 		}
 	}
--- a/pkg/bloomgateway/bloomgateway.go
+++ b/pkg/bloomgateway/bloomgateway.go
@@ -383,6 +383,8 @@ func (g *Gateway) consumeTask(ctx context.Context, task Task, tasksCh chan<- Tas
 		case <-ctx.Done():
 			// do nothing
 		default:
+			// chunks may not be sorted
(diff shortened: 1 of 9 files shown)


pkg/bloomgateway/bloomgateway.go
```diff
@@ -383,6 +383,8 @@ func (g *Gateway) consumeTask(ctx context.Context, task Task, tasksCh chan<- Tas
 		case <-ctx.Done():
 			// do nothing
 		default:
+			// chunks may not be sorted
+			sort.Slice(res.Removals, func(i, j int) bool { return res.Removals[i].Less(res.Removals[j]) })
```
salvacorts, line 387: nit: We tend to use sort very often and I'm worried we end up sorting already sorted slices (quicksort can degrade to O(n^2) for sorted inputs). Should we check if the list is sorted before sorting it?

pkg/bloomgateway/client.go
```diff
@@ -324,51 +325,37 @@ func mergeSeries(input [][]*logproto.GroupedChunkRefs, buf []*logproto.GroupedCh
...
+			result = append(result, a)
+			i++
+			j++
+			continue
+		}
+
+		if a.Less(b) {
+			result = append(result, a)
+			i++
+		} else {
```
salvacorts, line 345: nite: more idiomatic to add `continue` inside the if than having an else.

pkg/bloomgateway/bloomgateway.go
```diff
@@ -383,8 +384,10 @@ func (g *Gateway) consumeTask(ctx context.Context, task Task, tasksCh chan<- Tas
 		case <-ctx.Done():
 			// do nothing
 		default:
-			// chunks may not be sorted
-			sort.Slice(res.Removals, func(i, j int) bool { return res.Removals[i].Less(res.Removals[j]) })
+			// chunks may not always be sorted
+			if !slices.IsSortedFunc(res.Removals, func(a, b v1.ChunkRef) int { return a.Cmp(b) }) {
```
salvacorts, line 388: nit: we end up using the same cmp funtion all over the places. Would be nice to have another function next to `func (r *ChunkRef) Cmp(other ChunkRef) int` that does this:
```go
func CmpChunkRefs(a, b ChunkRef) int { return a.cmp(b) }
``` 

So we can use IsSortedFunc and SortFunc as:
```
                        if !slices.IsSortedFunc(a.Refs, CmpChunkRefs) {
				slices.SortFunc(a.Refs, CmpChunkRefs)
			}
			if !slices.IsSortedFunc(b.Refs, CmpChunkRefs) {
				slices.SortFunc(b.Refs, CmpChunkRefs)
			}
```

salvacorts (approved):
Approving but leaving a nit here

PR #12808, opened 2024-04-26
fix(blooms): dont break iterator conventions

owen-d:
Followup to https://github.com/grafana/loki/pull/12806 which exposes skipped pages more explicitly than as an error.
* refactors skip logic for bloom pages that are too large
* s/Seek/LoadOffset/ for LazyBloomIter
* removes unused code

### DIFF
--- a/pkg/storage/bloom/v1/block.go
+++ b/pkg/storage/bloom/v1/block.go
@@ -144,7 +144,7 @@ func (bq *BlockQuerier) Seek(fp model.Fingerprint) error {
 func (bq *BlockQuerier) Next() bool {
 	for bq.series.Next() {
 		series := bq.series.At()
-		if ok := bq.blooms.LoadOffset(series.Offset); !ok {
+		if skip := bq.blooms.LoadOffset(series.Offset); skip {
 			// can't seek to the desired bloom, likely because the page was too large to load
 			// so we skip this series and move on to the next
 			continue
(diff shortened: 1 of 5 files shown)


pkg/storage/bloom/v1/block.go
```diff
@@ -144,14 +144,12 @@ func (bq *BlockQuerier) Seek(fp model.Fingerprint) error {
 func (bq *BlockQuerier) Next() bool {
 	for bq.series.Next() {
 		series := bq.series.At()
-		bq.blooms.Seek(series.Offset)
+		if ok := bq.blooms.LoadOffset(series.Offset); !ok {
```
salvacorts, line 147: I think LoadOffset should return an error. Then we should check here if the error is due to the page being too big in which case we can safely continue. Otherwise, we wont be able to error when there is. for example, an IO error.

pkg/storage/bloom/v1/fuse.go
```diff
@@ -85,26 +94,22 @@ func (fq *FusedQuerier) Run() error {
...
-			}
+			fq.noRemovals(nextBatch, fp)
 			continue
 		}
 
 		// Now that we've found the series, we need to find the unpack the bloom
-		fq.bq.blooms.Seek(series.Offset)
+		ok := fq.bq.blooms.LoadOffset(series.Offset)
+		if !ok {
+			// could not seek to the desired bloom,
```
salvacorts, line 104: Same as previous comment

owen-d:
Good points @salvacorts  -- I inverted the logic and now `LoadOffset` returns a `skip`:
```go
// LoadOffset returns whether the bloom page at the given offset should
// be skipped (due to being too large) _and_ there's no error
func (it *LazyBloomIter) LoadOffset(offset BloomOffset) (skip bool)
```

PR #1275, opened 2019-11-18
pkg/promtail: IETF Syslog (RFC5424) Support

bastjan:
**What this PR does / why we need it**:

Allow promtail to ingest [IETF syslog (RFC5424)](https://tools.ietf.org/html/rfc5424).

**Limitations**:

- No remote time support (tried it but got out-of-order errors due to data races across the network)
- No UDP/ No [BSD syslog (RFC3164)](https://tools.ietf.org/html/rfc3164).

**Performance**:

Tested on my somewhat slow machine (Intel Core i5-3210M). Used the `loggen` tool from syslog-ng with 200 parallel log streams. Minimal relabeling.

<details>
<summary>Configuration</summary>

```yaml
server:
  http_listen_port: 9080
  grpc_listen_port: 0

positions:
  filename: /tmp/positions.yaml

clients:
  - url: http://localhost:3100/loki/api/v1/push

scrape_configs:
- job_name: syslog
  syslog:
    listen_address: localhost:1514
    labels:
      job: "syslog"
  relabel_configs:
  - source_labels: ['__syslog_message_hostname']
    target_label: 'hostname'
```

</details>

```sh
> loggen --syslog-proto --active-connections 200 --rate 10000 --stream localhost 1514
average rate = 78671.04 msg/sec, count=2124253, time=27.001, (average) msg size=260, bandwidth=19963.56 kB/sec
```

**Which issue(s) this PR fixes**:
Fixes #153, Fixes #935 

**Special notes for your reviewer**:

~~Currently tracks the `develop` branch of `github.com/influxdata/go-syslog` since the `v2.0.0` releases go mod support is broken.~~

**Checklist**
- [x] Documentation added
- [x] Tests updated

PR #12810, opened 2024-04-28
feat(promtail): Support of RFC3164 aka BSD Syslog

catap:

This PR introduce support of RFC3164 at promtail with additional hacks to support old and wired implementations.

It includes different BSD system such as OpenBSD and some embeded system. Frankly speaking all of them that I have access: from OpenBSD server to Unifi UA and IP Socket and Synology NAS with Nginx, works with this changes.

Both protocol TCP and UDP works.

It based on https://github.com/grafana/loki/pull/7845 and requries some changes inside used go-syslog library which was accomulated as single PR https://github.com/influxdata/go-syslog/pull/55 but upstream of that library seems to be not so active.

It is tested with config like:
```yaml
server:
  disable: true

positions:
  filename: /tmp/promtail-syslog-positions.yml

clients:
  - url: http://127.0.0.1:3100/loki/api/v1/push

scrape_configs:
  - job_name: syslog
    syslog:
      listen_address: 127.0.0.1:1514
      listen_protocol: tcp # or udp
      is_rfc3164_message: true
      labels:
        job: syslog
    relabel_configs:
      - source_labels: [__syslog_message_severity]
        target_label: level
      - source_labels: [__syslog_message_facility]
        target_label: facility
      - source_labels: [__syslog_connection_hostname]
        target_label: host
      - source_labels: [__syslog_connection_ip_address]
        target_label: ip
      - source_labels: [__syslog_message_hostname]
        target_label: hostname
      - source_labels: [__syslog_message_app_name]
        target_label: app
      - source_labels: [__syslog_message_proc_id]
        target_label: proc
      - source_labels: [__syslog_message_msg_id]
        target_label: msgid
```

A few times was made quite clear that such things aren't supported. So, no issue were opened.

**Checklist**
- [x] Documentation added
- [x] Tests updated


cstyan:
hey @catap we're technically calling promtail feature complete but I think we can make a case for this really being a fix rather than being a feature if promtail didn't work with BSD syslog before, though we would definitely want to wait for your changes to be upstreamed via https://github.com/influxdata/go-syslog/pull/55 first

catap:
Hey @cstyan!

Yes, BSD syslog were nuke when support of syslog was introduced at https://github.com/grafana/loki/pull/1275 since when was one attempt https://github.com/grafana/loki/pull/7845 to do it which I've reused.

As part of that attempt the PR: https://github.com/influxdata/go-syslog/pull/46 with required changes inside go-syslog were created but nobody had reviewed it.

It was December 2022.

I doubt that my PR will be merged some time soon, let be honest.

But this changes which is localized inside not used before code allows me to introduce support of different devices which sends logs to Loki. From BSD system to Nginx with Unifi Networks and so on.

So, what's a plan here? I see that https://github.com/grafana/go-syslog exists and I may open PR into this repository, but as far as I see nobody uses it.

catap:
Just a rebase to the last main.

cstyan:
> Hey @cstyan!
> 
> Yes, BSD syslog were nuke when support of syslog was introduced at #1275 since when was one attempt #7845 to do it which I've reused.
> 
> As part of that attempt the PR: [influxdata/go-syslog#46](https://github.com/influxdata/go-syslog/pull/46) with required changes inside go-syslog were created but nobody had reviewed it.
> 
> It was December 2022.
> 
> I doubt that my PR will be merged some time soon, let be honest.
> 
> But this changes which is localized inside not used before code allows me to introduce support of different devices which sends logs to Loki. From BSD system to Nginx with Unifi Networks and so on.
> 
> So, what's a plan here? I see that https://github.com/grafana/go-syslog exists and I may open PR into this repository, but as far as I see nobody uses it.

Unfortunately I don't have any suggestions here at the moment. Given we're winding down our own work on promtail and (to my knowledge) don't use the syslog target ourselves I don't really want to take on maintenance of this code within the Loki repo. It's likely better to ask the Alloy team to take on maintenance of the code either directly or via a fork, though I doubt that will happen anytime soon either. They have their own open issue tracking support for the same RFC: https://github.com/grafana/alloy/issues/305

My only other suggestion would be opening an issue on the go-syslog repo itself asking if the repo is unmaintained and trying to get in contact with someone from Influx who has worked on it in the past. It's likely it would also get no response, or a response of "yes, unmaintained". In either case, hopefully that would entice someone else in the Go community to take on maintenance of a fork.

catap:
> Unfortunately I don't have any suggestions here at the moment. Given we're winding down our own work on promtail and (to my knowledge) don't use the syslog target ourselves I don't really want to take on maintenance of this code within the Loki repo. It's likely better to ask the Alloy team to take on maintenance of the code either directly or via a fork, though I doubt that will happen anytime soon either. They have their own open issue tracking support for the same RFC: [grafana/alloy#305](https://github.com/grafana/alloy/issues/305)
> 
> My only other suggestion would be opening an issue on the go-syslog repo itself asking if the repo is unmaintained and trying to get in contact with someone from Influx who has worked on it in the past. It's likely it would also get no response, or a response of "yes, unmaintained". In either case, hopefully that would entice someone else in the Go community to take on maintenance of a fork.

From another hand: do you see any objection to merge? If this code will be removed soon enough, when you decided to nuke promtail, why not keep it with this feature?

It doesn't change any existed code, just add a new one feature which should be enabled by hand.

Upstream project looks like to be forgetten to be archived, but this code increases use case of loki right now, that should attract new users for Loki.

cstyan:
> From another hand: do you see any objection to merge? If this code will be removed soon enough, when you decided to nuke promtail, why not keep it with this feature?

Yes, because we're not just getting rid of promtail. It's going to be maintained with security and bug fixes for (I'm just trying to make a reasonable guess here) a year or two if not more. 

What we are not doing is supporting additional features. We could make an argument for this being a fix but the actual changes you're making to `go-syslog` would need to be present in the upstream repo or a fork that someone is committed to maintaining (which we would then need to switch to using instead of influx's repo).

catap:
> What we are not doing is supporting additional features. We could make an argument for this being a fix but the actual changes you're making to `go-syslog` would need to be present in the upstream repo or a fork that someone is committed to maintaining (which we would then need to switch to using instead of influx's repo).

I see your point and I haven't got any resources to promise you to support `go-syslog` for next couple of years.

Frankly speaking after xz-history I understand why you decided to go this way.

But, as I said before, I really think that this PR will benefite Lokie and Grafana ecosystem.

Maybe as better approach you may tag someone from Alloy team? I had quite a fast look to their code and it seems like a fork of promtail.

cstyan:
> I see your point and I haven't got any resources to promise you to support go-syslog for next couple of years.
> But, as I said before, I really think that this PR will benefite Lokie and Grafana ecosystem.

I don't disagree.

> Maybe as better approach you may tag someone from Alloy team?

We can ping them like this @grafana/grafana-alloy-maintainers, but we also had a discussion internally, the consensus was that we don't have the bandwidth to maintain go-syslog ourselves and finding another solution is low priority at the moment.

> I had quite a fast look to their code and it seems like a fork of promtail.

It's a hard fork of some of promtails code, yes.

catap:
> > Maybe as better approach you may tag someone from Alloy team?
> 
> We can ping them like this @grafana/grafana-alloy-maintainers, but we also had a discussion internally, the consensus was that we don't have the bandwidth to maintain go-syslog ourselves and finding another solution is low priority at the moment.
> 

Seems that github included comma into name of group and decided to not ping them. So, here I'm to ping: @grafana/grafana-alloy-maintainers

catap:
@cstyan here an update. @leodido had forked his `go-syslog` which he decided to continue support, see his response here: https://github.com/influxdata/go-syslog/issues/56#issuecomment-2119171616

After that he had merged the PR, and here I'm back with reduced PR which:
 - switch go-syslog to an actual upstream (and the last release);
 - enable support of old syslog at promtail.

cstyan:
> @cstyan here an update. @leodido had forked his `go-syslog` which he decided to continue support, see his response here: [influxdata/go-syslog#56 (comment)](https://github.com/influxdata/go-syslog/issues/56#issuecomment-2119171616)


That's great news :+1:

leodido:
Yup, I decided to keep maintaining it at https://github.com/leodido/go-syslog 
With write access it's way more easy to maintain stuff :)
You can already go get github.com/leodido/go-syslog/v4 (v4.1.0 is the latest atm)

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
```diff
@@ -202,6 +202,9 @@ type SyslogTargetConfig struct {
 	// message should be pushed to Loki
 	UseRFC5424Message bool `yaml:"use_rfc5424_message"`
 
+	// IsRFC3164Message defines wether the log is formated as RFC3164
+	IsRFC3164Message bool `yaml:"is_rfc3164_message"`
+
```
cstyan, line 207: Sorry @catap I thought I left this comment yesterday:

Is there a way we can just detect whether each message is RFC3164 or RFC5424 format, rather than adding another config option? Is it possible that a single target could output logs in both formats?

catap: @cstyan nope, it is impossible to determine which format is used. Usually sender sends in supported format, or it may have a selector where someone may choose the format.

The best that I can do here is use `handleMessageRFC3164` as fallback. If `RFC5424` can't parse it, let use `RFC3164` but it may lead to the case when `RFC5424` parses it wrong way, so, I suggest to avoid that.

cstyan: Okay that's fine. 

Lets keep the logic as is then, except can we change the config option to be `SyslogFormat` which has multiple string possibilities, like one of `RFC5425, RFC3164, another RFC, etc.`? And then have a validation function when creating the loading the syslog config or creating the syslog target. 

And it's default value can be `RFC5424`. That way we don't need to add another config field if at some point the future we add support for yet another format. it just has to be another possible option for the validation function.

We also need a reference for this in the syslog section of `docs/sources/send-data/promtail/configuration.md`

catap: Here the catch: right now we have two options:
```
	// UseRFC5424Message defines whether the full RFC5424 formatted syslog
	// message should be pushed to Loki
	UseRFC5424Message bool `yaml:"use_rfc5424_message"`

	// IsRFC3164Message defines wether the log is formated as RFC3164
	IsRFC3164Message bool `yaml:"is_rfc3164_message"`
```
and I feel that this should be reworked as weel. To something like:
 - SyslogFormat: RFC5424 (default) or RFC3164;
 - UseFullSyslogMessage: default is false

cstyan: > and I feel that this should be reworked as weel. To something like:
> 
> SyslogFormat: RFC5424 (default) or RFC3164;
> UseFullSyslogMessage: default is false

this makes sense to me, we would have to deprecate `UseRFC5424Message` for now and have the actual value for `use syslog` be an or of `UseRFC5424` and `UseFullSyslogFormat: false`  when `SyslogFormat` is `RFC5424`

I can present your PR to the team as a "we should do this as a bug fix, it also technically requires these config changes" and we should be able to get it merged

catap: @cstyan I was wrong, RFC3164 hasn't got string method and implement it isn't something that seems easy. So, let keep it as is.

catap:
@cstyan to make review easy, I've added all remarks as dedicated commit. Feel free to ask, if I need to squash it.

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
```diff
@@ -202,15 +211,19 @@ type SyslogTargetConfig struct {
 	// message should be pushed to Loki
 	UseRFC5424Message bool `yaml:"use_rfc5424_message"`
```
cstyan, line 212: I think we wanted a second new config option `PushFullSyslogMessage`? we still need to keep this one as well, and mark it as deprecated

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
```diff
@@ -202,15 +211,19 @@ type SyslogTargetConfig struct {
...
 
-	// IsRFC3164Message defines wether the log is formated as RFC3164
-	IsRFC3164Message bool `yaml:"is_rfc3164_message"`
+	// Used Sylog format at the target.
```
cstyan, line 214:
```suggestion
        Syslog format used at the target.
```

clients/pkg/promtail/targets/syslog/syslogtarget.go
```diff
@@ -225,7 +225,7 @@ func (t *SyslogTarget) handleMessageRFC3164(connLabels labels.Labels, msg syslog
 }
 
 func (t *SyslogTarget) handleMessage(connLabels labels.Labels, msg syslog.Message) {
-	if t.config.IsRFC3164Message {
+	if t.config.IsRFC3164Message() {
```
cstyan, line 228: I think switching on the `config.SyslogFormat` would be cleaner

cstyan (commented):
No need to squash/force push. It's actually easier to review if you push new commits for new changes. We always squash when merging back into main anyways.

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
catap: I can't do it, because RFC3164 message has lack of `String()` method.  See implementation for RFC5424: https://github.com/leodido/go-syslog/blob/v4.1.0/rfc5424/builder.go#L9348-L9408

The issue that RFC3164 isn't well structured and quite flexible, and rebuild original message is quite a challenge.

Sorry to misslead you.

clients/pkg/promtail/targets/syslog/syslogtarget.go
catap: Here possible, but it is using in few more places, and it leads to `==` and need to export the contant name, or hardcoded strings...

So, I've decided to use trivial method which can be refactored.

Anyway, here only two RFC for Syslog, and I doubt that someone will make 3rd soon enough.

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
catap: Enforced missed `//`

cstyan:
okay, I think we're probably good to go 

I'll take another look in the morning

catap:
@cstyan may I ask you to re-run tests, this time with `go fmt`!

clients/pkg/promtail/scrapeconfig/scrapeconfig.go
```diff
@@ -202,12 +211,19 @@ type SyslogTargetConfig struct {
 	// message should be pushed to Loki
 	UseRFC5424Message bool `yaml:"use_rfc5424_message"`
 
+	// Syslog format used at the target.
```
cstyan, line 214: I think we should clarify the valid options here, and that `rfc5424` is the default.

cstyan: we could implement the default in the parser or like this: https://github.com/grafana/loki/blob/efdae3df14c47d627eb99e91466e0451db6e16f6/clients/pkg/promtail/targets/syslog/syslogtarget.go#L233-L238

cstyan: It looks like we don't have a dedicated config parsing function on this path, so maybe in `NewSyslogTarget` would make the most sense to detect if `SyslogFormat` is empty string and if so set the default

cstyan (commented):
just one last change needed imo

catap:
@cstyan ping

cstyan (approved):
Thanks for your patience @catap! Sorry for the delay, I was on vacation last week.

PR #12836, opened 2024-04-30
fix: fix parsing of default per tenant otlp config

### LINKED ISSUE #12780: `otlp_config` has no effect
**Describe the bug**
I'm trying to utilize `otlp_config` to shift some attributes I was previously dropping into newly introduced `Structured metadata`.
My attempts were ineffective and it's hard to distinguish indexed label and non-indexed attribute in Explore UI, so I decided to test radical removal of attributes with the following config.

```yaml
limits_config:
  otlp_config:
    resource_attributes:
      attributes_config:
      - action: drop
        attributes:
        - k8s.pod.uid
        - k8s.pod.name
        - k8s.deployment.name
    log_attributes:
      - action: drop
        regex: .*
    scope_attributes:
      - action: drop
        regex: .*
```

This still resulted in no changes in observed `Fields` in Explore UI. So I've enabled debug logging on `OTel Collector` to see the data I'm sending to Loki.
Here is an example of log record produced by debug exporter:
```yaml
ResourceLog #15
Resource SchemaURL: 
Resource attributes:
     -> app.name: Str(REDACTED)
     -> build.id: Str(REDACTED)
     -> k8s.namespace.name: Str(REDACTED)
     -> k8s.node.name: Str(REDACTED)
     -> k8s.pod.name: Str(REDACTED)
     -> k8s.pod.uid: Str(REDACTED)
ScopeLogs #0
ScopeLogs SchemaURL: 
InstrumentationScope  
LogRecord #0
ObservedTimestamp: 2024-04-24 17:36:23.589775336 +0000 UTC
Timestamp: 2024-04-24 17:36:23.420370489 +0000 UTC
SeverityText: 
SeverityNumber: Unspecified(0)
Body: Str(REDACTED)
Attributes:
     -> pod: Str(REDACTED)
     -> container: Str(REDACTED)
     -> namespace: Str(REDACTED)
     -> cluster: Str(REDACTED)
     -> alerting_tier: Str(REDACTED)
Trace ID: 
Span ID: 
Flags: 0
```
And here is the list if observed `Fields`
image
So it is still exactly a combination of `resource_attributes` and `log_attributes` plus auto-discovered `level` and `service_name`.

Additionally I've confirmed, that `k8s_pod_name` is indexed label while `k8s_pod_uid` is non-indexed field, which follows default behavior. And mere presence of `pod`, `container`, `namespace`, `cluster` and `alerting_tier` makes me think that `otlp_config` is completely ineffective.

In ingester's debug log I see lines like those:
```log
caller=flush.go:176 component=ingester msg="flushing stream" user=<...> fp=65e0db5cd9e011cf immediate=false num_chunks=1 labels="{k8s_deployment_name=\"<...>\", k8s_namespace_name=\"<...>\", k8s_pod_name=\"<...>\", service_name=\"<...>\"}"
```
so it somehow still produces those labels I'm trying to drop.

**Expected behavior**
I expect majority of `Fields` to be absent in Explore, plus reduced resource consumption by Loki.

**Environment:**
 - Infrastructure: Kubernetes (EKS 1.28)
 - Deployment tool: Helm chart `6.3.4`

sandeepsukhani:
There was a mistake I made in setting the default per tenant otlp config parsed from the overall Loki config, which causes Loki to ignore it. The per tenant otlp config set in overrides works as expected.
This PR fixes the issue and updates the otlp integration test to validate the change.

Fixes #12780 

**Checklist**
- [x] Tests updated

### DIFF
--- a/integration/cluster/cluster.go
+++ b/integration/cluster/cluster.go
@@ -68,6 +68,9 @@ limits_config:
       attributes_config:
         - action: index_label
           attributes: ["service.name"]
+    log_attributes:
+      - action: drop
+        attributes: [email]
 
 storage_config:
   named_stores:
@@ -430,6 +433,7 @@ func (c *Component) run() error {
 		return err
 	}
 
+	config.LimitsConfig.SetGlobalOTLPConfig(config.Distributor.OTLPConfig)
 	var err error
 	c.loki, err = loki.New(config.Config)
 	if err != nil {
--- a/integration/loki_micro_services_test.go
+++ b/integration/loki_micro_services_test.go
@@ -739,7 +739,7 @@ func TestOTLPLogsIngestQuery(t *testing.T) {
 
 	t.Run("ingest-logs", func(t *testing.T) {
 		// ingest some log lines
-		require.NoError(t, cliDistributor.PushOTLPLogLine("lineA", now.Add(-45*time.Minute), map[string]any{"trace_id": 1, "user_id": "2"}))
+		require.NoError(t, cliDistributor.PushOTLPLogLine("lineA", now.Add(-45*time.Minute), map[string]any{"trace_id": 1, "user_id": "2", "email": "foo@bar.com"}))
 		require.NoError(t, cliDistributor.PushOTLPLogLine("lineB", now.Add(-45*time.Minute), nil))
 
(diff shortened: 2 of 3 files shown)


MichelHollands (approved):
lgtm

PR #12851, opened 2024-05-02
chore(blooms): Remove ID field from task struct

chaudum:

We've seen a  few cases where creating the ULID failed for unknown reasons, and the ID is not really used. It was only useful early on in the development for debugging.


ashwanthgoli (approved):
lgtm

PR #12863, opened 2024-05-02
feat(blooms): ignore individual bloom-gw failures

owen-d:
* includes a new metric to record statuses now that we ignore errors
* Also removes old/no longer relevant cache locality score metric since we now resolve bloom-gw membership in a different way

### DIFF
--- a/pkg/bloomgateway/client.go
+++ b/pkg/bloomgateway/client.go
@@ -238,18 +238,6 @@ func (c *GatewayClient) FilterChunks(ctx context.Context, _ string, interval blo
 		}
 	}
 
-	if len(servers) > 0 {
-		// cache locality score (higher is better):
-		// `% keyspace / % instances`. Ideally converges to 1 (querying x% of keyspace requires x% of instances),
-		// but can be less if the keyspace is not evenly distributed across instances. Ideal operation will see the range of
-		// `1-2/num_instances` -> `1`, where the former represents slight
-		// overlap on instances to the left and right of the range.
-		pctKeyspace := float64(lastFp-firstFp) / float64(math.MaxUint64)
-		pctInstances := float64(len(servers)) / float64(max(1, len(c.pool.Addrs())))
-		cacheLocalityScore := pctKeyspace / pctInstances
-		c.metrics.cacheLocalityScore.Observe(cacheLocalityScore)
-	}
-
 	results := make([][]*logproto.GroupedChunkRefs, len(servers))
 	count := 0
 	err := concurrency.ForEachJob(ctx, len(servers), len(servers), func(ctx context.Context, i int) error {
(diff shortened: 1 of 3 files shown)


pkg/bloomgateway/client.go
```diff
@@ -269,8 +257,19 @@ func (c *GatewayClient) FilterChunks(ctx context.Context, _ string, interval blo
...
+				// so instrument & move on
+				c.metrics.clientRequests.WithLabelValues(typeError).Inc()
+				level.Error(c.logger).Log(
+					"msg", "filter failed for instance, skipping",
+					"addr", rs.addr,
+					"series", len(rs.groups),
+					"blocks", len(rs.blocks),
+					"err", err,
+				)
+				return nil
```
chaudum, line 270: In case a request fails, the results entries need to be equal to the request entries, otherwise we assume that all requested chunks have been filtered out.

```suggestion
				results[i] = rs.groups
				count += len(rs.groups)
				return nil
```

pkg/bloomgateway/metrics.go
```diff
@@ -15,21 +15,25 @@ type metrics struct {
...
-		cacheLocalityScore: promauto.With(registerer).NewHistogram(prometheus.HistogramOpts{
+		clientRequests: promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
 			Namespace: constants.Loki,
 			Subsystem: "bloom_gateway_client",
-			Name:      "cache_locality_score",
-			Help:      "Cache locality score of the bloom filter, as measured by % of keyspace touched / % of bloom_gws required",
-			Buckets:   prometheus.LinearBuckets(0.01, 0.2, 5),
-		}),
+			Name:      "requests_total",
+			Help:      "Total number of requests made to the bloom gateway",
```
chaudum, line 35: :+1:

pkg/bloomgateway/client.go
chaudum: Increasing `count` is not strictly needed, but the value is used for the allocation of the result buffer.

pkg/bloomgateway/client_test.go
```diff
@@ -52,7 +52,7 @@ func TestGatewayClient_MergeSeries(t *testing.T) {
 			{Fingerprint: 0x01, Refs: []*logproto.ShortRef{shortRef(0, 1, 3), shortRef(1, 2, 4)}}, // fully overlapping chunks
 			{Fingerprint: 0x02, Refs: []*logproto.ShortRef{shortRef(0, 1, 5), shortRef(1, 2, 6)}}, // partially overlapping chunks
 		},
-		// response 2
+		// response 3
```
chaudum, line 55: nit

chaudum (approved):
Minor nitpicks, but approved to unblock

chaudum (approved):

PR #12876, opened 2024-05-03
docs: specify minimum number of nodes in k8s cluster

a-h:
Minor update. The Helm chart can't be deployed to a single Node k8s setup due to pod affinity rules in the chart.


JStickler:
@a-h Thanks for the update.  Could you please sign the Contributor License Agreement, then we can get this merged.

PR #12878, opened 2024-05-03
fix(helm): Fix GEL image tag, bucket name and proxy URLs

### LINKED ISSUE #12845: Helm chart is using wrong image tag and proxy config for GEL in SSD mode
**Describe the bug**
When deploying GEL 3.0 from the Helm chart, the following happens:
- The chart tries to deploy `enterprise-logs:v3.0.0` but the image tag is actually `enterprise-logs:3.0.0` (without the `v`). So the deployment fails to pull the image from Docker Hub.
- The GEL plugin for Grafana does not show Ring Health properly, because the enterprise gateway is configured with proxy URLs which assume distributed mode, and not the default SSD mode.

**To Reproduce**
1. Create a values.yaml file
2. `helm template --values values.yaml gel grafana/loki | less`
3. Notice that the image tag is `v3.0.0` instead of `3.0.0`. The URLs on the Deployment are set to distributed URLs (distributor, ingester, etc).

**Expected behavior**
The Helm chart should deploy GEL 3.0.0, and the GEL plugin for Grafana should correctly show Ring health.

**Environment:**
 - Infrastructure: laptop -> GKE
 - Deployment tool: helm

N/A

monodot:
The chart doesn't currently seem to deploy GEL without having to do a couple of workarounds. This PR:
- Updates the image tag referenced in the deployment to the correct `3.0.1` (without the `v` prefix). 
- Respects the user's admin bucket name. 
- Adds the correct downstream proxy URLs in the enterprise-gateway for deployments in SSD mode (the Ring Health tab in the Grafana Enterprise Logs plugin was otherwise unable to list the nodes in SSD).

Fixes #12845

- I have updated the image tag manually here to get the chart working, but I think the grafanabot job will also need to be updated (the one that gets the latest tag and inserts it into the Helm chart), otherwise this change will be overwritten on the next release. For that, there's an Issue here: https://github.com/grafana/enterprise-logs/issues/2381
- I have reverted to using the user's configured bucket name for the admin bucket, could you please confirm it looks correct?

**Checklist**
- [x] Documentation added
- [ ] Tests updated


production/helm/loki/values.yaml
```diff
@@ -480,7 +480,7 @@ enterprise:
     admin_client:
       storage:
         s3:
-          bucket_name: admin
+          bucket_name: {{ .Values.loki.storage.bucketNames.admin }}
```
trevorwhitney, line 483: good catch, looks like we broken this in #12548. Would you mind adding a default for this value _only_ in the case where both minio and enterprise are enabled? Otherwise there should be no default so we don't run into the problem fixed in #12548

trevorwhitney (changes requested):
Looks great, thanks for catching these. Can you add a default for the admin bucket _only_ in the case of using minio, and make sure it's empty otherwise?

production/helm/loki/values.yaml
monodot: I've updated this now. I've separated the admin storage config out into its own separate Helm value (`enterprise.adminClientConfig`), which includes a default for Minio only. 

To set any other storage options (e.g. AWS, GCS, etc), users can just override that `adminClientConfig` value.

trevorwhitney (changes requested):
Ideally a user only has to provide `loki.storage.bucketNames` and everything else should just work. I think we should default the admin config to use `loki.storage.bucketNames.admin` in all cases but minio, and in the minio case we can hardcode it. I don't think we want to force a user to have to provide a whole `adminClient` section when they are not using minio.

trevorwhitney (approved):
looks great, thanks for addressing all my feedback!

production/helm/loki/templates/_helpers.tpl
```diff
@@ -454,6 +454,82 @@ ruler:
...
+storage:
+  {{- if .Values.minio.enabled }}
+  backend: "s3"
+  s3:
+    bucketnames: admin
+  {{- else if eq .Values.loki.storage.type "s3" -}}
+  {{- with .Values.loki.storage.s3 }}
+  backend: "s3"
+  s3:
+    bucketnames: {{ $.Values.loki.storage.bucketNames.admin }}
```
monodot, line 468: I think the admin storage config has different value names here? e.g. it uses `bucket_name` not `bucketnames`

vlad-diachenko: you are completely right;) let me fix it asap

vlad-diachenko: done

PR #12889, opened 2024-05-03
chore(blooms): records more bloom iteration stats

owen-d:
This gives us more granular information on why data was added/passed filters/skipped. Also cleans up a bit of bloom-gw code. Gives info like 
image


chaudum (approved):
lgtm

PR #12899, opened 2024-05-06
feat(detectedFields): Support multiple parsers to be returned for a single field

svennergr:
Detected fields across different streams can require different parsers, thus this PR changes the `parser` property to a `parsers: []string` property.

Furthermore, if multiple fields with the same label/name are detected but different types, we set that to `string` now.

I removed the `parsedFields.isTypeDetected` property and will just detect the type for each new stream, as this is most likely to differ there. Let me know if you think that should change.


pkg/querier/querier.go
```diff
@@ -1206,10 +1203,15 @@ func parseDetectedFields(ctx context.Context, limit uint32, streams logqlmodel.S
...
 				}
 
+				if !slices.Contains(df.parsers, *parser) {
+					df.parsers = append(df.parsers, *parser)
+				}
+
 				for _, v := range vals {
 					parsedFields := detectedFields[k]
-					if !parsedFields.isTypeDetected {
+					if detectType {
```
trevorwhitney, line 1212: this is tricky. just to be clear, out assumption here is that fields in the same stream will have the same type, but fields in different streams may have different types which is why we only include the default "string" type logic in the merge code?

I think that's probably a safe assumption, and it save us from having to run the determine type logic on every line, but I want to make sure it's clear to future readers the assumption we're making.

trevorwhitney (approved):
lgtm! one comment to call all the assumption were making about field types in single streams vs. when merging the results. wonder if it's worth a comment in the code?

pkg/querier/querier.go
svennergr: Agree, added a comment!

PR #12900, opened 2024-05-06
fix(blooms): Fix `partitionSeriesByDay` function

chaudum:

The function `partitionSeriesByDay` could yield empty when chunks were outside of the requested time range.
    
This would result in requests for a time range that does not contain any
chunks for a series to filter. These requests would return errors, because a subsequent `partitionSeriesByDay` for would remove that day resulting in no task.

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/bloomgateway/util.go
+++ b/pkg/bloomgateway/util.go
@@ -2,7 +2,6 @@ package bloomgateway
 
 import (
 	"sort"
-	"time"
 
 	"github.com/prometheus/common/model"
 	"golang.org/x/exp/slices"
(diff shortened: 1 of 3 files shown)


pkg/bloomgateway/util.go
```diff
@@ -13,13 +12,8 @@ import (
...
 )
 
-func getDayTime(ts model.Time) time.Time {
-	return ts.Time().UTC().Truncate(Day)
-}
-
 func truncateDay(ts model.Time) model.Time {
-	// model.minimumTick is time.Millisecond
-	return ts - (ts % model.Time(24*time.Hour/time.Millisecond))
+	return model.TimeFromUnix(ts.Time().Truncate(Day).Unix())
```
salvacorts, line 16: ❤️

pkg/bloomgateway/util.go
```diff
@@ -125,7 +119,7 @@ func partitionSeriesByDay(from, through model.Time, seriesWithChunks []*logproto
 			})
 
 			// All chunks fall outside of the range
-			if min == len(chunks) || max == 0 {
+			if min == len(chunks) || max == 0 || min == max {
```
salvacorts, line 122: idea: I feel we have run already too many times into forgetting to check `min==max` maybe we should wrap this search use-case into some Template function or something.

salvacorts (approved):
Excellent catch 👏  LGTM

PR #12918, opened 2024-05-08
feat: Add metrics for number of patterns detected & evicted

benclive:
Adds two metrics for tracking the operation of the Patterns extraction.
- One to count whenever we find a new pattern ("LogCluster")
- One to count whenever a pattern is evicted from the LRU cache (I believe this only happens if a single stream has more than 300 patterns)

Together, they should let us understand the performance & memory use of the pattern-ingester in order to know whether to tweak the number of patterns generated or to prune more often.

### DIFF
--- a/pkg/pattern/drain/drain.go
+++ b/pkg/pattern/drain/drain.go
@@ -151,11 +151,15 @@ func New(config *Config, metrics *Metrics) *Drain {
 		panic("depth argument must be at least 3")
 	}
 	config.maxNodeDepth = config.LogClusterDepth - 2
+	var evictFn func(int, *LogCluster)
+	if metrics != nil {
+		evictFn = func(int, *LogCluster) { metrics.PatternsEvictedTotal.Inc() }
+	}
 
 	d := &Drain{
 		config:      config,
 		rootNode:    createNode(),
-		idToCluster: createLogClusterCache(config.MaxClusters, func(int, *LogCluster) { metrics.PatternsEvictedTotal.Inc() }),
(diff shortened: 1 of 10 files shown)


pkg/pattern/drain/drain.go
```diff
@@ -146,7 +146,7 @@ func DefaultConfig() *Config {
 	}
 }
 
-func New(config *Config) *Drain {
+func New(config *Config, metrics *Metrics) *Drain {
```
benclive, line 149: I'm not sure if this is the best way to pass these metrics around. There are many streams/instances of Drain, so I had to register the metrics in the ingester code and pass it all the way through. Let me know if there's a preferred way to do this!

pkg/pattern/drain/drain.go
```diff
@@ -155,7 +155,8 @@ func New(config *Config) *Drain {
 	d := &Drain{
 		config:      config,
 		rootNode:    createNode(),
-		idToCluster: createLogClusterCache(config.MaxClusters),
+		idToCluster: createLogClusterCache(config.MaxClusters, func(int, *LogCluster) { metrics.PatternsEvictedTotal.Inc() }),
```
shantanualsi, line 158: perhaps needs a nil check for `metrics`?

pkg/pattern/drain/drain.go
cyriltovena: yep this is it.

cyriltovena:
Need a conflict resolution before merging.

PR #12936, opened 2024-05-10
feat: Store structured metadata with patterns

benclive:
- Adds structured metadata from the incoming log stream to the Pattern chunks. 
- Enables the the use of label filters in the Pattern API Query string so filters like `| detected_level="debug"` can be applied to patterns returned by the API.
- No other LogQL queries are supported.

I'm definitely not confident this is the right way to implement this feature, hence leaving this PR in draft mode until I get some early feedback.
It also needs some more tests so I'll follow up with some more commits over the next few days before making undrafting this PR.

### DIFF
--- a/pkg/pattern/drain/chunk.go
+++ b/pkg/pattern/drain/chunk.go
@@ -4,9 +4,12 @@ import (
 	"sort"
 	"time"
 
+	"github.com/grafana/loki/pkg/push"
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/model/labels"
 
 	"github.com/grafana/loki/v3/pkg/logproto"
+	"github.com/grafana/loki/v3/pkg/logql/log"
 	"github.com/grafana/loki/v3/pkg/pattern/iter"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/pattern/drain/chunk.go
```diff
@@ -21,16 +24,20 @@ const (
 type Chunks []Chunk
 
 type Chunk struct {
-	Samples []logproto.PatternSample
+	Samples            []logproto.PatternSample
+	StructuredMetadata map[string]string
 }
 
-func newChunk(ts model.Time) Chunk {
+func newChunk(ts model.Time, structuredMetadata push.LabelsAdapter) Chunk {
```
benclive, line 31: I'm not confident I am using the right types to pass the labels through from the request - any recommendations here?

pkg/pattern/drain/chunk.go
```diff
@@ -94,31 +101,50 @@ func (c Chunk) ForRange(start, end, step model.Time) []logproto.PatternSample {
...
+	var emptyLine []byte
 	for _, chunk := range c {
+		chunkMetadata := make([]labels.Label, 0, len(chunk.StructuredMetadata))
+		for k, v := range chunk.StructuredMetadata {
+			chunkMetadata = append(chunkMetadata, labels.Label{
+				Name:  k,
+				Value: v,
+			})
+		}
+		_, _, matches := labelFilters.Process(0, emptyLine, chunkMetadata...)
```
benclive, line 143: This feels a little hacky since I'm passing empty TS and Line, but it enables re-use of the existing LogQL logic to implement the matching so I think its the best option.

pkg/pattern/ingester_querier.go
```diff
@@ -44,10 +45,24 @@ func NewIngesterQuerier(
 }
 
 func (q *IngesterQuerier) Patterns(ctx context.Context, req *logproto.QueryPatternsRequest) (*logproto.QueryPatternsResponse, error) {
-	_, err := syntax.ParseMatchers(req.Query, true)
+	expr, err := syntax.ParseLogSelector(req.Query, true)
 	if err != nil {
 		return nil, httpgrpc.Errorf(http.StatusBadRequest, err.Error())
 	}
+	var queryErr error
+	expr.Walk(func(treeExpr syntax.Expr) {
```
benclive, line 53: I put this here to guard against anyone passing in more complex queries to this API. Only allowing a subset of LogQL is a bit of an odd use case so I'd appreciate any input on whether there's a better way to do this.

cyriltovena (commented):
The way you did it is fine, but it duplicates data this is why I think we want to improve storage to co-locate all this data together.

trevorwhitney:
The strategy I'm currently working on for storing series/metrics might help here, where we can store the label on the metric series, since that's really what we want from this API (the metric for the counts that is)

benclive:
Pausing this until https://github.com/grafana/loki/pull/13020 is merged, then we'll rethink whether this approach will work.

PR #12976, opened 2024-05-15
perf: Replace channel check with atomic bool in tailer.send()

benclive:
- @cyriltovena discovered that tailer.isClosed was eating a significant amount of CPU in ops (>20 CPUs worth)
image]
- I replaced the select on the channel to try and reduce the amount of CPU time spent on this particular code path.
- I used an atomic bool for the `send()` but retained the `closeChan` to signal the other parts of the tailer to shut down ASAP, otherwise we'd need to check the bool periodically using (for example) `time.After`.

Benchmark before thte change, using select-default:
```
Benchmark_tailer_isClosed
Benchmark_tailer_isClosed-14    	536670529	         2.220 ns/op
PASS
```

Benchmark after the change, using atomic bool:
```
Benchmark_tailer_isClosed
Benchmark_tailer_isClosed-14    	1000000000	         0.2886 ns/op
PASS
```

So this small change is nearly 10x faster and therefore would save ~18-20 cores worth of CPU time in our ops cluster.

### DIFF
--- a/pkg/ingester/tailer.go
+++ b/pkg/ingester/tailer.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"hash/fnv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-kit/log/level"
@@ -46,6 +47,7 @@ type tailer struct {
 	// and the loop and senders should stop
 	closeChan chan struct{}
 	closeOnce sync.Once
+	closed    atomic.Bool
 
 	blockedAt         *time.Time
 	blockedMtx        sync.RWMutex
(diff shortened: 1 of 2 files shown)


MasslessParticle (commented):
Would it make sense to remove `closeChan` altogether? I see it's used in a few other `select` statements that we can't get rid of, but it would keep us from signaling `closed` in two different ways in the same file.

Nevermind. On closer look, those other usages rely on the close chan to interrupt selects blocking

MasslessParticle (approved):

cyriltovena (approved):
LGTM !

For next time, in your PR you can use [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) it gives a much better output.

PR #12992, opened 2024-05-20
feat(blooms): Separate page buffer pools for series pages and bloom pages

chaudum:
    
Series pages are much smaller than bloom pages and therefore can make use of a separate buffer pool with different buckets.

The second commit fixes a possible panic.


salvacorts (approved):
LGTM!

PR #13006, opened 2024-05-21
feat: new stream count limiter

vlad-diachenko:
implemented a new stream count limiter that uses the values from ownedStreamService if the feature flag is enabled. This functionality is needed to take into account only the owned streams while asserting the limit.

- until the feature flag is enabled, these changes should not affect the users
- previously we had a RWLock near the value in the limiter, and that lock was used to disable all the limit while recovering the WAL. However, this lock is not needed because when we replay the WAL we do not assert if we reached the stream count limit: https://github.com/grafana/loki/blob/main/pkg/ingester/instance.go#L288-L290 (record is nil while replaying the WAL). So, I removed this RWLock to simplify the implementation.

**Checklist**
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/ingester/instance.go
+++ b/pkg/ingester/instance.go
@@ -104,7 +104,10 @@ type instance struct {
 	tailers   map[uint32]*tailer
 	tailerMtx sync.RWMutex
 
-	limiter *Limiter
+	limiter            *Limiter
+	streamCountLimiter *streamCountLimiter
+	ownedStreamsSvc    *ownedStreamService
+
 	configs *runtime.TenantConfigs
 
 	wal WAL
@@ -147,11 +150,12 @@ func newInstance(
 	if err != nil {
 		return nil, err
 	}
-
+	streams := newStreamsMap()
+	ownedStreamsSvc := newOwnedStreamService(instanceID, limiter)
 	c := config.SchemaConfig{Configs: periodConfigs}
 	i := &instance{
 		cfg:        cfg,
-		streams:    newStreamsMap(),
+		streams:    streams,
 		buf:        make([]byte, 0, 1024),
 		index:      invertedIndex,
 		instanceID: instanceID,
@@ -159,9 +163,11 @@ func newInstance(
 		streamsCreatedTotal: streamsCreatedTotal.WithLabelValues(instanceID),
 		streamsRemovedTotal: streamsRemovedTotal.WithLabelValues(instanceID),
 
-		tailers: map[uint32]*tailer{},
-		limiter: limiter,
-		configs: configs,
+		tailers:            map[uint32]*tailer{},
+		limiter:            limiter,
+		streamCountLimiter: newStreamCountLimiter(instanceID, streams.Len, limiter, ownedStreamsSvc),
+		ownedStreamsSvc:    ownedStreamsSvc,
+		configs:            configs,
 
 		wal:                   wal,
 		metrics:               metrics,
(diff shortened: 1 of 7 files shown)


pkg/ingester/instance.go
```diff
@@ -333,21 +321,47 @@ func (i *instance) createStream(ctx context.Context, pushReqStream logproto.Stre
...
+}
+
+func (i *instance) onStreamCreated(s *stream) {
 	memoryStreams.WithLabelValues(i.instanceID).Inc()
 	memoryStreamsLabelsBytes.Add(float64(len(s.labels.String())))
 	i.streamsCreatedTotal.Inc()
 	i.addTailersToNewStream(s)
 	streamsCountStats.Add(1)
-
+	i.ownedStreamsSvc.incOwnedStreamCount()
```
JordanRushing, line 357: Should this go in the `LogStreamCreation` conditional?

pkg/ingester/limiter.go
```diff
@@ -76,46 +77,39 @@ func (l *Limiter) UnorderedWrites(userID string) bool {
...
-	return fmt.Errorf(errMaxStreamsPerUserLimitExceeded, userID, streams, calculatedLimit, localLimit, globalLimit, adjustedGlobalLimit)
+	return first
 }
 
 func (l *Limiter) convertGlobalToLocalLimit(globalLimit int) int {
 	if globalLimit == 0 {
 		return 0
 	}
-
+	// todo: change to healthyInstancesInZoneCount() once
```
JordanRushing, line 112: 👍

pkg/validation/limits.go
```diff
@@ -269,6 +270,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Var(&l.CreationGracePeriod, "validation.create-grace-period", "Duration which table will be created/deleted before/after it's needed; we won't accept sample from before this time.")
 	f.IntVar(&l.MaxEntriesLimitPerQuery, "validation.max-entries-limit", 5000, "Maximum number of log entries that will be returned for a query.")
 
+	f.BoolVar(&l.UseOwnedStreamCount, "ingester.use-owned-stream-count", false, "When true, the ingester takes into account only the streams that belongs to this instance according to the ring while applying the stream limit.")
```
JordanRushing, line 273:
```suggestion
	f.BoolVar(&l.UseOwnedStreamCount, "ingester.use-owned-stream-count", false, "When true an ingester takes into account only the streams that it owns according to the ring while applying the stream limit.")
```

JordanRushing (approved):
LGTM 🎉 May I suggest adding a benchmark for using owned streams vs not?

pkg/ingester/limiter.go
JordanRushing: re: https://github.com/grafana/dskit/pull/526

pkg/ingester/instance.go
vlad-diachenko: no, we need to do it all the time... btw, even if UseOwnedStreamCount is set to false, because this option is in runtime config and can be turned on at any moment, the service should always know the count of the streams.

vlad-diachenko:
> May I suggest adding a benchmark for using owned streams vs not?

Stream creation is not executed on every push, only when we create a new stream. Also, the implementation just adds another source of values, which will be just atomic ints. So, the new changes are unlikely to affect the performance, and knowing that this validation is not executed on every push request, makes me think that we could skip it;))) 
WDYT?

PR #13008, opened 2024-05-21
TBD

DylanGuedes:
TBD


pkg/tool/commands/audit.go
```diff
@@ -0,0 +1,60 @@
...
+	if err := util_cfg.DefaultUnmarshal(&auditCfg, []string{configFileArg}, flag.CommandLine); err != nil {
+		fmt.Fprintf(os.Stderr, "failed parsing config: %v\n", err)
+		os.Exit(1)
+	}
+	if err := auditCfg.Validate(); err != nil {
+		fmt.Fprintf(os.Stderr, "failed validating config: %v\n", err)
+		os.Exit(1)
+	}
+
+	found, missing, err := audit.Run(context.Background(), a.path, "19859", auditCfg, logger)
```
MichelHollands, line 37: 19859 is hardcoded. Shouldn't this be a table name?

DylanGuedes: good question. I'll cehck ti again but my thought here was: this table name is only used by tsdb to compute the interval, which is 24h in our case. So instead of asking the user to give the table name I preset it.

That said: I'll check how the TSDB internals use it, maybe we don't even need the table at all :shrug:

DylanGuedes: I've updated the code, WDYT?

pkg/tool/commands/audit.go
```diff
@@ -0,0 +1,64 @@
...
+	configFile string
+
+	extraArgs []string
+}
+
+func (a *AuditCommand) auditIndex(_ *kingpin.ParseContext) error {
+	logger := log.NewLogfmtLogger(os.Stdout)
+
+	var auditCfg audit.Config
+	// configFileArg := fmt.Sprintf("-config.file=%s", a.configFile)
```
MichelHollands, line 30: This can be removed

MichelHollands (approved):
One comment that can be removed but otherwise LGTM

pkg/tool/commands/audit.go
DylanGuedes: oops :sweat_smile:

PR #13010, opened 2024-05-22
fix: Fix panic on requesting out-of-order Pattern samples

benclive:
This PR fixes a panic that can occur when samples are out-of-order in the Pattern chunks.
This happens because the calculated slice length can be negative if the timestamp at `hi-1` is less than the timestamp at `lo` due to subtracting `currentStep` (generated from timestamp@`lo`) from the timestamp@`hi-1`.
```go
	currentStep := truncateTimestamp(c.Samples[lo].Timestamp, step)
	aggregatedSamples := make([]logproto.PatternSample, 0, ((c.Samples[hi-1].Timestamp-currentStep)/step)+1)
```

**Implementation Notes**
- I opted to return nil if the timestamps are out-of-order which may be incorrect behaviour (but does solve the panic).
- The samples should never be out of order now since I added a sort to the `Add` method.
- An alternative is to simply reject samples that are older than the latest during `Add` - let me know if you prefer this approach.

### DIFF
--- a/pkg/pattern/drain/chunk.go
+++ b/pkg/pattern/drain/chunk.go
@@ -51,7 +51,7 @@ func (c Chunk) ForRange(start, end, step model.Time) []logproto.PatternSample {
 	}
 	first := c.Samples[0].Timestamp
 	last := c.Samples[len(c.Samples)-1].Timestamp
-	if start >= end || first >= end || last < start {
+	if last < first || start >= end || first >= end || last < start {
 		return nil
 	}
 	var lo int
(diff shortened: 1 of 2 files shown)


pkg/pattern/drain/chunk.go
```diff
@@ -51,7 +51,7 @@ func (c Chunk) ForRange(start, end, step model.Time) []logproto.PatternSample {
 	}
 	first := c.Samples[0].Timestamp
 	last := c.Samples[len(c.Samples)-1].Timestamp
-	if start >= end || first >= end || last < start {
+	if last < first || start >= end || first >= end || last < start {
```
ashwanthgoli, line 54: this might not cover all the cases since the in-between samples could be out-of-order?

benclive: Good point - I'll add some more test cases

benclive:
Rewrote this so it will reject out-of-order samples for the Pattern instead of trying to sort this chunk.
Firstly, this makes the sample generation code much simpler. Secondly, this is a fairly rare case and since Patterns don't need to be 100% exact, its OK to simply drop these values.

ashwanthgoli (approved):
lgtm

PR #13021, opened 2024-05-23
fix(mixins): remove ingester from reads resources

QuentinBisson:

This PR remove the ingester section from the read path because it is already on the loki-writes-resources file so it makes no sense to have it duplicated

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


QuentinBisson:
@MichelHollands do you think this change makes sense? If it does not, I would advise having only 1 resource dashbord for the ssd deployment that would contain resources for read, write and backend in one place

cstyan:
@QuentinBisson imo we need to keep the ingester on the reads dashboard both for distributed and ssd mode, queries for recent data hit the ingesters and so it's useful to see ingester resource usage right next to other read path components.

QuentinBisson:
Sure, i'll make a loki resources overview for the single scalable deployment only that will have all 3 anyways,  closing here :)

PR #13030, opened 2024-05-24
fix: not owned stream count

vlad-diachenko:

added an additional field to track the count of not-owned streams count, otherwise when recalculating job updates ownedStreamCount and not-owned streams are actually flushed, and the counter will contain the wrong value because it's decremented every time the stream is flushed.


JordanRushing (approved):
LGTM

PR #13039, opened 2024-05-27
perf: Introduce fixed size memory pool for bloom querier

chaudum:

This PR introduces a fixed size memory pool for bloom pages that are loaded in the block querier.

The aim of a fixed size pool of `[]byte` buffers is to reduce the amount of allocations, as well as to control the maximum heap size to prevent OOMing of the bloom gateways.

Also, with the current usage of a `sync.Pool`, the query parallelism (`bloom-gateway.worker-concurrency` and `bloom-gateway.block-query-concurrency`) was defined by the `bloom.max-query-page-size` (and vice versa), because the max page size could be loaded `workers * concurrency` times at the same time. Most of the time, though, smaller pages are loaded, and therefore concurrency is not optimized.

With the new fixed size memory pool, this problem should be solved. The pool is divided into slabs of different sizes holding different amounts of buffers; a larger amount of small sized buffers and a smaller amount of large sized buffers.

**TODO**

- [x] Make slab sizes and their capacity configurable
- [x] Expose metrics for mempool usage

**Notes for the reviewer**

I'm up for suggestions for naming the configuration options, as well as for the interface name. I don't really like the term `Allocator` in this context.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


pkg/storage/bloom/v1/util.go
```diff
@@ -44,24 +44,49 @@ var (
 
 	// buffer pool for bloom pages
 	// 128KB 256KB 512KB 1MB 2MB 4MB 8MB 16MB 32MB 64MB 128MB
-	BloomPagePool = BytePool{
+	BloomPagePool = &BytePool{
```
salvacorts, line 47: IIUC we will no longer be using this one but we are still allocating it. Should we remove it?

docs/sources/shared/configuration.md
```diff
@@ -5378,6 +5378,18 @@ bloom_shipper:
...
     # CLI flag: -bloom.metas-lru-cache.ttl
     [ttl: <duration> | default = 1h]
+
+  memory_management:
+    # One of: simple (simple heap allocations using Go's make([]byte, n) and no
+    # re-cycling of buffers), dynamic (a buffer pool with variable sized buckets
+    # and best effort re-cycling of buffers using Go's sync.Pool), fixed (a
+    # fixed size memory pool with configurable slab sizes, see mem-pool-buckets)
+    # CLI flag: -bloom.memory-management.alloc-type
+    [bloom_page_alloc_type: <string> | default = "dynamic"]
```
salvacorts, line 5388: Is this something we want to expose to users? Or do we want to have it to run some experiments and will later remove it?

pkg/bloomgateway/bloomgateway_test.go
```diff
@@ -216,6 +216,12 @@ func TestBloomGateway_FilterChunkRefs(t *testing.T) {
 	})
 
 	t.Run("request cancellation does not result in channel locking", func(t *testing.T) {
+		t.Skip(`
```
salvacorts, line 219: Shall we add a TODO here so we don't skip it in the future?

pkg/storage/bloom/v1/block.go
```diff
@@ -117,11 +117,11 @@ type BlockQuerier struct {
 // will be returned to the pool for efficiency. This can only safely be used
 // when the underlying bloom bytes don't escape the decoder, i.e.
 // when loading blooms for querying (bloom-gw) but not for writing (bloom-compactor).
-func NewBlockQuerier(b *Block, noCapture bool, maxPageSize int) *BlockQuerier {
+func NewBlockQuerier(b *Block, alloc Allocator, maxPageSize int) *BlockQuerier {
 	return &BlockQuerier{
 		block:  b,
 		series: NewLazySeriesIter(b),
-		blooms: NewLazyBloomIter(b, noCapture, maxPageSize),
+		blooms: NewLazyBloomIter(b, alloc, maxPageSize),
```
salvacorts, line 124: Instead of having to pass the allocator down the chain, what about having the allocator as a pkg-level variable similarly to here?

https://github.com/grafana/loki/pull/13039/files#diff-8572b07b377f1bfae2ec4ffd992cb8d06972b089d2c195996faae0763621a318R18

pkg/storage/bloom/v1/mempool/pool.go
```diff
@@ -0,0 +1,117 @@
...
+
+	ptr := unsafe.Pointer(unsafe.SliceData(buf))
+	s.buffer <- ptr
+}
+
+// MemPool is an Allocator implementation that uses a fixed size memory pool
+// that is split into multiple slabs of different buffer sizes.
+// Buffers are re-cycled and need to be returned back to the pool, otherwise
+// the pool runs out of available buffers.
+type MemPool struct {
```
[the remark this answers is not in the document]
chaudum: Yes, it's an open TODO of this PR

pkg/storage/bloom/v1/mempool/pool.go
```diff
@@ -0,0 +1,117 @@
...
+// Allocating a buffer from an exhausted pool/slab will return an error.
+// Allocating a buffer that exceeds the largest slab size will cause a panic.
+func (a *MemPool) Get(size int) ([]byte, error) {
+	for i := 0; i < len(a.slabs); i++ {
+		if a.slabs[i].size < size {
+			continue
+		}
+		return a.slabs[i].get(size)
+	}
+	panic(fmt.Sprintf("no slab found for size: %d", size))
```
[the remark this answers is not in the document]
chaudum: I don't want to grow the memory pool, but keep it to a predefined max size. Yes, we can return an error instead of panicking.

docs/sources/shared/configuration.md
chaudum: I think this must be configurable, but it's kind of an "expert setting". We could hide it from the docs

pkg/bloomgateway/bloomgateway_test.go
chaudum: I would rather remove the test completely

pkg/storage/bloom/v1/block.go
chaudum: I prefer explicit passing instead of package-global variables. IMO global state should be avoided.

pkg/storage/bloom/v1/util.go
chaudum: We use it as default Allocator type, if you don't specify anything in the config. It's the type "dynamic" (yeah, rather bad name, I know ...)

salvacorts (approved):
LGTM!

PR #13069, opened 2024-05-29
feat: Add tokenizer interface for Drain Training

benclive:
* Added an interface for Tokenizing incoming log lines for Drain
* This is useful because Tokenization is important to how well the algorithm runs, and we may want to test multiple tokenizers or swap tokenizers depending on the stream log formats.

Also took the opportunity to add a filter so Drain will not train on any loglines with less than 4 tokens.


cyriltovena (approved):
Lgtm

PR #13077, opened 2024-05-30
BREAKING CHANGE(HELM): remove unregister on shutdown argument on zone aware ingesters

### LINKED ISSUE #13071: Why is ingester.unregister-on-shutdown=false for Zone aware ingesters hardcoded?
https://github.com/grafana/loki/blob/797bb641736a2355b4f8503c147fc0c8a814f19a/production/helm/loki/templates/ingester/statefulset-ingester-zone-b.yaml#L104

It's respectfully, very annoying to both have argument flags and configuration settings that do the same. This sends people on a chase in their config. Only to later realise that it's set as bin argument. 

I guess this consist of both a question and suggestion:

1. Why is this set as default? Is it required for zone aware ingesters? If so, why? 
2. Please, do not add this as arg, but make the config leading. IF it's required for zone aware ingesters, make an if/else in the values.yaml or some logic that 'forces' it to be `false`. 

For point 2, the whole thing is that helm templates are abstracted from us (users). Config values are leading. Having said value to default `true` and no mention of it being set to `false` is confusing.

Thanks in advance.

wiardvanrij:

Helm templates are abstracted from us (users). Config values should be leading as that's the thing that folks can influence/change. Therefore, removing the hardcoded arg values in favour of being able to set these via the config. Furthermore, the default should be `true`, and I do not see any reason this should be `false` for zone aware ingesters

Fixes #13071

I made it a breaking change, as we are effectively changing the default value if one uses zone aware ingesters. Users will now default to the configuration default, which is `true` instead of the current `false`. There is not really a way to keep backwards compatibility, but my 2 cents is that it's better to default to the actual intended default. 

**Checklist**
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/helm/loki/CHANGELOG.md
+++ b/production/helm/loki/CHANGELOG.md
@@ -13,6 +13,10 @@ Entries should include a reference to the pull request that introduced the chang
 
 [//]: # (<AUTOMATED_UPDATES_LOCATOR> : do not remove this line. This locator is used by the CI pipeline to automatically create a changelog entry for each new Loki release. Add other chart versions and respective changelog entries bellow this line.)
 
+## 6.7.0
+
+- [CHANGE] **BREAKING** Remove `ingester.unregister-on-shutdown=false` for zone aware ingesters in favour of being able to set this via the configuration option `unregister_on_shutdown` (defaults `true`).
+
 ## 6.6.2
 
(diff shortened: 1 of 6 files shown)


vlad-diachenko:
this change is not needed. I explained why this option is so important and how to override the value if needed in [this comment](https://github.com/grafana/loki/issues/13071#issuecomment-2141726827)

PR #13078, opened 2024-05-30
refactor(blooms): Add metrics for per-tenant tasks progress to planner

salvacorts:
This PR adds two new metrics to track the build process of each tenant.

- We already expose a per-tnenat metric for the queue size:
   https://github.com/grafana/loki/blob/5b8d0e666de2221f08a4537d7b5eda0be79f939b/pkg/queue/metrics.go#L16-L21 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


pkg/bloombuild/planner/planner.go
```diff
@@ -149,6 +151,8 @@ func (p *Planner) running(ctx context.Context) error {
 			}
 
 		case <-inflightTasksTicker.C:
+			// TODO(salvacorts): Is think this won't evaluate while the runOne is running.
+			//                   We should run the runOne in a goroutine and check the context inside (as we already do).
```
chaudum, line 155: nit: I would rather run the ticker in a new goroutine

pkg/bloombuild/planner/planner.go
```diff
@@ -272,6 +278,12 @@ func (p *Planner) loadWork(
 				})
 			}
 
+			// Reset progress tracking metrics for this tenant
+			// NOTE(salvacorts): We will reset them multiple times for the same tenant, for each table, but it's not a big deal.
+			//                   Alternatively, we can use a Counter instead of a Gauge, but I think a Gauge is easier to reason about.
+			p.metrics.tenantTasksPlanned.WithLabelValues(tenant).Set(0)
+			p.metrics.tenantTasksCompleted.WithLabelValues(tenant).Set(0)
```
chaudum, line 285: nit: should we remove the metric with the tenant label completely, instead of just resetting it to 0?

chaudum: Also, we could then use a Counter

salvacorts: > should we remove the metric with the tenant label completely, instead of just resetting it to 0?

AFAICT a metric cannot be explicitly removed. Can it?

> we could then use a Counter

That was my first idea but a counter would make it a bit difficult to see for a given iteration, how many "new" tasks would be created. It would require using `increase` but setting the interval is difficult. Instead, by using a gauge, you can see the current set of tasks.

Also note that the gauge doesn't change often, but only when the build is planned.

chaudum (approved):

PR #13079, opened 2024-05-30
fix: Update expected patterns when pruning

benclive:
Fixes a test that is currently failing on `main` due to https://github.com/grafana/loki/pull/13069
* The tokenizer was changed which resulted in a slightly different set of patterns in this test.
* Updated & sorted the output so this test passes and stays stable.


cyriltovena (approved):
LGTM

PR #13084, opened 2024-05-30
feat: Collect duplicate log line metrics

paul1r:
This PR is for adding support to collect metrics and information about duplicate logs that are ingested.  Previously, duplicate logs are just discarded, with no way to actually see the fact that duplicates were received, nor the quantity of the data that was discarded.

This PR adds two runtime configuration options for a tenant:
`log_duplicate_metrics`: To output an ingester metric named `duplicate_log_bytes_total`, by tenant id, which counts how many bytes were discarded due to log line(s) being duplicates.
`log_duplicate_stream_info`: To output details about duplicate logs to insight logs.

**Checklist**
- [x] Documentation added
- [x] Tests updated


pkg/validation/validate.go
```diff
@@ -128,3 +129,13 @@ var LineLengthHist = promauto.NewHistogram(prometheus.HistogramOpts{
 	Help:      "The total number of bytes per line.",
 	Buckets:   prometheus.ExponentialBuckets(1, 8, 8), // 1B -> 16MB
 })
+
+// DuplicateLogEntries is a metric of the total discarded duplicate bytes, by tenant.
+var DuplicateLogEntries = promauto.NewCounterVec(
+	prometheus.CounterOpts{
+		Namespace: constants.Loki,
+		Name:      "duplicate_log_entries_total",
```
slim-bean, line 137: I believe this metric is a byte counter not a line counter so this should probably be `duplicate_log_bytes_total`

pkg/chunkenc/unordered.go
```diff
@@ -135,7 +135,7 @@ func (hb *unorderedHeadBlock) Append(ts int64, line string, structuredMetadata l
 		for _, et := range displaced[0].(*nsEntries).entries {
 			if et.line == line {
 				e.entries = displaced[0].(*nsEntries).entries
-				return nil
+				return ErrDuplicateEntry
```
slim-bean, line 138: I don't think you can do this? returning an Error here I believe will be propagated as a failure of the push of the line, which is a change in semantics as to how Loki processes exact dupes (we currently do not fail on this, rather silently swallow the duplicates)

slim-bean: Instead of incrementing the metric in 2 other places and returning an error, can we increment the metric here? This is really the place were exact dupes are actually handled (at least when running Loki with out of order ingest)

slim-bean:
I think rather than incrementing the metric in 2 places, is it possible instead to increment it inside the unordered head block? this is the "source of truth" if you will for if we store an exact dupe or not. 

Also then we can avoid returning an error from that function which I _think_ is changing some existing behaviors in Loki

pkg/ingester/stream.go
```diff
@@ -342,6 +350,18 @@ func (s *stream) storeEntries(ctx context.Context, entries []logproto.Entry, usa
...
+				if s.configs.LogDuplicateMetrics(s.tenant) {
+					s.metrics.duplicateLogBytesTotal.WithLabelValues(s.tenant).Add(float64(len(entries[i].Line)))
+				}
+				if s.configs.LogDuplicateStreamInfo(s.tenant) {
+					errMsg := fmt.Sprintf("duplicate log entry at timestamp %s for stream %s", entries[i].Timestamp.Format(time.RFC3339), s.labelsString)
+					dupErr := errors.New(errMsg)
+					s.writeFailures.Log(s.tenant, dupErr)
+				}
+			}
+		}
```
DylanGuedes, line 364: could we refactor this part to make it less nested?

pkg/ingester/stream.go
```diff
@@ -342,6 +350,18 @@ func (s *stream) storeEntries(ctx context.Context, entries []logproto.Entry, usa
 			}
 			continue
 		}
+		if dup {
+			if s.configs != nil {
+				if s.configs.LogDuplicateMetrics(s.tenant) {
+					s.metrics.duplicateLogBytesTotal.WithLabelValues(s.tenant).Add(float64(len(entries[i].Line)))
```
DylanGuedes, line 356: this PR adds two CLI configs, one for the metric and one for the logs. WDYT of making both working behind a single one? Asking because I can't think of a scenario were I'd want only one of them. But having both means another CLI flag to Loki.

pkg/distributor/writefailures/manager.go
```diff
@@ -39,7 +39,8 @@ func (m *Manager) Log(tenantID string, err error) {
 		return
 	}
 
-	if !m.tenantCfgs.LimitedLogPushErrors(tenantID) {
+	if !m.tenantCfgs.LimitedLogPushErrors(tenantID) &&
+		!m.tenantCfgs.LogDuplicateStreamInfo(tenantID) {
```
DylanGuedes, line 43: this conditional is redundant (since you're checking this on stream.go already) but moreover, this would mean we won't log any new log lines through the manager if `logDuplicateStreamInfo` is false.

Suggestion: don't add this check here. The manager should just emit the log and ratelimit it. Control over what errors are logged or not should probably be done at an upper level.

docs/sources/shared/configuration.md
```diff
@@ -3778,6 +3778,14 @@ These are values which allow you to control aspects of Loki's operation, most co
 # CLI flag: -operation-config.log-push-request-streams
 [log_push_request_streams: <boolean> | default = false]
 
+# Log metrics for duplicate lines received.
+# CLI flag: -operation-config.log-duplicate-metrics
+[log_duplicate_metrics: <boolean> | default = false]
+
+# Log stream info for duplicate lines received
+# CLI flag: -operation-config.log-duplicate-stream-info
+[log_duplicate_stream_info: <boolean> | default = false]
```
DylanGuedes, line 3787: ditto: I think it makes sense to unify both into a single config.

DylanGuedes (changes requested):
sounds good, just a few small suggestions and one change necessary (the one regarding not logging anything on the manager if the new arg is false)

pkg/ingester/stream.go
paul1r: @slim-bean Argues that we need two, which is why we added it

pkg/ingester/stream.go
```diff
@@ -357,6 +368,21 @@ func (s *stream) storeEntries(ctx context.Context, entries []logproto.Entry, usa
...
 
+func (s *stream) handleLoggingOfDuplicateEntry(entry logproto.Entry) {
+	if s.configs == nil {
+		return
+	}
+	if s.configs.LogDuplicateMetrics(s.tenant) {
+		s.metrics.duplicateLogBytesTotal.WithLabelValues(s.tenant).Add(float64(len(entry.Line)))
+	}
+	if s.configs.LogDuplicateStreamInfo(s.tenant) {
+		errMsg := fmt.Sprintf("duplicate log entry at timestamp %s for stream %s", entry.Timestamp.Format(time.RFC3339), s.labelsString)
```
DylanGuedes, line 379: maybe you want to log the size in bytes too? (even if that's redundant with the metric above)
like `duplicate log entry with size=%d at timestamp %s for stream %s`

DylanGuedes (approved):

docs/sources/shared/configuration.md
slim-bean: i think i would leave these separate, I think generally just the metric is useful?

slim-bean (approved):
LGTM!

I think my only nit would be to add a comment to each of the Append implementations which explains what the bool does

PR #13085, opened 2024-05-30
feat: Add utf8 support to Pattern Lexer to support utf8 chars

benclive:
Drain was producing patterns which did not match the original line in some cases. Example logline:
```
13:25:18.033470 ▶ INFO route ops sending to dest https://graphite-cortex-ops-blocks-us-east4.grafana.net/graphite/metrics: service_is_carbon-relay-ng.instance_is_carbon-relay-ng-c665b7b-j2trk.mtype_is_gauge.dest_is_https_graphite-cortex-ops-blocks-us-east4_grafana_netgraphitemetrics.unit_is_B.what_is_FlushSize.type_is_manual.stat_is_max_999 0.00 1717075518
```

The issue was in the LogQL pattern matching code: The pattern parser was identifying the single UTF8 character `▶` as 3 separate ASCII characters. This is actually a single 3-byte UTF encoded character.
I added support for UTF-8 bytes in the Lexer to correctly identify them as 1 character in order to resolve this.

The lexer code was generated by Copilot, although I found an [alternative implementation](https://gist.github.com/CAFxX/0d15ce5430921cda86d93922a2696c69) to check it against.


pkg/logql/log/pattern/lexer.rl
```diff
@@ -13,11 +13,25 @@ package pattern
     }
 }%%
 
+%%{
+utf8 = (
```
benclive, line 17: this bit was generated by copilot but I've manually checked it and I think its correct. As far as I understand, anyway.

cyriltovena (approved):
Lgtm

PR #13091, opened 2024-05-31
feat: Support negative numbers in LogQL

benclive:
Adds negative number support to the LogQL ~lexer~ parser.
* Works for both ints and floats.
* Negative durations are already supported.
* Negative bytes are not supported because the humanize library does not support them and I don't think negative bytes make sense semantically.
  * This could be probably be supported with more effort, if we want to pursue it.


pkg/logql/syntax/expr.y
```diff
@@ -400,13 +400,13 @@ bytesFilter:
...
 
 numberFilter:
-      IDENTIFIER GT NUMBER      { $$ = log.NewNumericLabelFilter(log.LabelFilterGreaterThan, $1, mustNewFloat($3))}
-    | IDENTIFIER GTE NUMBER     { $$ = log.NewNumericLabelFilter(log.LabelFilterGreaterThanOrEqual, $1, mustNewFloat($3))}
-    | IDENTIFIER LT NUMBER      { $$ = log.NewNumericLabelFilter(log.LabelFilterLesserThan, $1, mustNewFloat($3))}
-    | IDENTIFIER LTE NUMBER     { $$ = log.NewNumericLabelFilter(log.LabelFilterLesserThanOrEqual, $1, mustNewFloat($3))}
-    | IDENTIFIER NEQ NUMBER     { $$ = log.NewNumericLabelFilter(log.LabelFilterNotEqual, $1, mustNewFloat($3))}
-    | IDENTIFIER EQ NUMBER      { $$ = log.NewNumericLabelFilter(log.LabelFilterEqual, $1, mustNewFloat($3))}
-    | IDENTIFIER CMP_EQ NUMBER  { $$ = log.NewNumericLabelFilter(log.LabelFilterEqual, $1, mustNewFloat($3))}
+      IDENTIFIER GT literalExpr      { $$ = log.NewNumericLabelFilter(log.LabelFilterGreaterThan, $1,  $3.Val)}
```
benclive, line 403: This is the crux of the change: numberFilters now accept literal expressions instead of the NUMBER token. This means the filter can include `+1`, `1` or `-1` as equally valid. They get parsed as a float as other literals do, the value is then extracted via Val.

pkg/logql/syntax/lex_test.go
```diff
@@ -95,6 +95,19 @@ func TestLex(t *testing.T) {
 		{`{foo="bar"} | logfmt --strict code"`, []int{OPEN_BRACE, IDENTIFIER, EQ, STRING, CLOSE_BRACE, PIPE, LOGFMT, PARSER_FLAG, IDENTIFIER}},
 		{`{foo="bar"} | logfmt --keep-empty --strict code="response.code", IPAddress="host"`, []int{OPEN_BRACE, IDENTIFIER, EQ, STRING, CLOSE_BRACE, PIPE, LOGFMT, PARSER_FLAG, PARSER_FLAG, IDENTIFIER, EQ, STRING, COMMA, IDENTIFIER, EQ, STRING}},
 		{`decolorize`, []int{DECOLORIZE}},
+		{`123`, []int{NUMBER}},
+		{`-123`, []int{SUB, NUMBER}},
```
benclive, line 99: I reverted the changes in the lexer so all negative numbers now evaluate as SUB, NUMBER. The AST does the work now.

pkg/logql/syntax/parser_test.go
```diff
@@ -499,19 +499,19 @@ var ParseTestCases = []struct {
 	// label filter for ip-matcher
 	{
 		in:  `{ foo = "bar" }|logfmt|addr>=ip("1.2.3.4")`,
-		err: logqlmodel.NewParseError("syntax error: unexpected ip, expecting BYTES or NUMBER or DURATION", 1, 30),
+		err: logqlmodel.NewParseError("syntax error: unexpected ip", 1, 30),
```
benclive, line 502: This is the biggest caveat with this change: The error messages presented to users are less clear. Presumably because numberFilter is now an expression, so there is no specific TOKEN to look for next. 
There is no change to the syntax, just in the error message.

benclive (commented):
@cyriltovena I've rethought my approach and moved the changes to the Parser and out of the Lexer. I've added comments on the relevant pieces, and added tests to clarify & valid the expected behaviour.

chaudum (approved):
:tada:

PR #13117, opened 2024-06-04
fix: Do not filter out chunks for store when `From==Through` and `From==start` 

chaudum:

This PR fixes a bug where chunks are incorrectly filtered out when their `From` timestamp is equal to their `Through` timestamp (which happens when there is a single log line) and the `From` timestamp is equal to the `from` time of the of the request.

How to reproduce:

1. Insert a single log line with a timestamp exactly at point of an hour
2. Flush ingester
3. Query log line with a split interval of 1h

Fixes #<issue number>

* In a query range, the `From` is inclusive and the `Through` is exclusive
* In a ChunkRef, both `From` and `Through` are inclusive, because `Through` represents the timestamp of the last log line in the chunk.

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/composite_store_entry.go
+++ b/pkg/storage/stores/composite_store_entry.go
@@ -92,8 +92,8 @@ func filterForTimeRange(refs []*logproto.ChunkRef, from, through model.Time) []c
 	for _, ref := range refs {
 		// Only include chunks where the query start time (from) is < the chunk end time (ref.Through)
 		// and the query end time (through) is >= the chunk start time (ref.From)
-		// A special case also exists where a chunk can contain a single log line which results in ref.From being equal to ref.Through, but only include this chunk if this log line timestamp is in the timerange of the query
-		if (through >= ref.From && from < ref.Through) || (ref.From == ref.Through && (ref.From >= from && ref.From < through) {
+		// A special case also exists where a chunk can contain a single log line which results in ref.From being equal to ref.Through, and that is equal to the from time.
+		if (through >= ref.From && from < ref.Through) || (ref.From == from && ref.Through == from) {
 			filtered = append(filtered, chunk.Chunk{
 				ChunkRef: *ref,
 			})
--- a/pkg/storage/stores/composite_store_entry.go
+++ b/pkg/storage/stores/composite_store_entry.go
@@ -90,7 +90,10 @@ func (c *storeEntry) GetChunks(
 func filterForTimeRange(refs []*logproto.ChunkRef, from, through model.Time) []chunk.Chunk {
 	filtered := make([]chunk.Chunk, 0, len(refs))
 	for _, ref := range refs {
-		if (through >= ref.From && from < ref.Through) || (ref.From == from && ref.Through == from) {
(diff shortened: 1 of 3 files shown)


pkg/storage/stores/composite_store_entry.go
```diff
@@ -90,7 +90,7 @@ func (c *storeEntry) GetChunks(
 func filterForTimeRange(refs []*logproto.ChunkRef, from, through model.Time) []chunk.Chunk {
 	filtered := make([]chunk.Chunk, 0, len(refs))
 	for _, ref := range refs {
-		if through >= ref.From && from < ref.Through {
+		if (through >= ref.From && from < ref.Through) || (ref.From == from && ref.Through == from) {
```
slim-bean, line 93:
```suggestion
		// Only include chunks where the query start time (from) is < the chunk end time (ref.Through)
		// and the query end time (through) is >= the chunk start time (ref.From)
		// A special case also exists where a chunk can contain a single log line which results in ref.From being equal to ref.Through, but only include this chunk if this log line timestamp is in the timerange of the query
		if (through >= ref.From && from < ref.Through) || (ref.From == ref.Through && (ref.From >= from && ref.From < through) {
```

pkg/storage/stores/composite_store_entry.go
```diff
@@ -90,7 +90,10 @@ func (c *storeEntry) GetChunks(
 func filterForTimeRange(refs []*logproto.ChunkRef, from, through model.Time) []chunk.Chunk {
 	filtered := make([]chunk.Chunk, 0, len(refs))
 	for _, ref := range refs {
-		if through >= ref.From && from < ref.Through {
+		// Only include chunks where the query start time (from) is < the chunk end time (ref.Through)
+		// and the query end time (through) is >= the chunk start time (ref.From)
+		// A special case also exists where a chunk can contain a single log line which results in ref.From being equal to ref.Through, and that is equal to the from time.
+		if (through >= ref.From && from < ref.Through) || (ref.From == from && ref.Through == from) {
```
DylanGuedes, line 96: do you know if there's a relevant difference between using `through.After(x)` vs `through > x`?

chaudum: I don't think so. `through.After()` is an additional function call, but that might get inlined anyway.

chaudum: Quickly checked the following example:


```go
package main

import (
	"fmt"
)

type Time int64

func (t Time) After(other Time) bool {
	return t > other
}

func main() {
	a := Time(1000)
	fmt.Println(a.After(500))
}

```


```console
$ go build -gcflags -m inline.go 
# command-line-arguments
./inline.go:9:6: can inline Time.After
./inline.go:15:21: inlining call to Time.After
./inline.go:15:13: inlining call to fmt.Println
<autogenerated>:1: inlining call to Time.After
./inline.go:15:13: ... argument does not escape
./inline.go:15:21: ~r0 escapes to heap
```

slim-bean (approved):
LGTM, nice find!

PR #13119, opened 2024-06-04
feat(structuredMetadata): add API to return structured metadata fields

svennergr:

Adds an API to return structured metadata API.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [x] Tests updated


pkg/querier/queryrange/codec.go
```diff
@@ -1556,6 +1614,21 @@ func (Codec) MergeResponse(responses ...queryrangebase.Response) (queryrangebase
...
+		headers := resp0.Headers
+
+		fields := []*logproto.StructuredMetadata{}
+		for _, r := range responses {
+			fields = append(fields, r.(*StructuredMetadataResponse).Response.Fields...)
+		}
+
+		return &StructuredMetadataResponse{
+			Response: &logproto.StructuredMetadataResponse{
+				Fields: fields,
```
cyriltovena, line 1628: I think you need to dedupe unique fields.

pkg/querier/queryrange/queryrange.proto
```diff
@@ -226,6 +234,7 @@ message QueryResponse {
     DetectedFieldsResponse detectedFields = 11;
     QueryPatternsResponse patternsResponse = 12;
     DetectedLabelsResponse detectedLabels = 13;
+    StructuredMetadataResponse structuredMetadata = 14;
```
cyriltovena, line 237: I dont think you need to touch pattern response for this.

svennergr:
@cyriltovena @trevorwhitney I decided to go with `/api/v1/structured_metadata/keys` for now. It's the quickest solution for now and would enable structuredmetadata in Explore Logs already.

Could you guide me where I should focus on adding tests, please?

pkg/util/marshal/marshal.go
```diff
@@ -191,6 +191,16 @@ func WriteDetectedFieldsResponseJSON(r *logproto.DetectedFieldsResponse, w io.Wr
 	return s.Flush()
 }
 
+// WriteStructuredMetadataKeysResponseJSON marshals a logproto.StructuredMetadataKeysResponse to JSON and then
+// writes it to the provided io.Writer.
+func WriteStructuredMetadataKeysResponseJSON(r *logproto.StructuredMetadataKeysResponse, w io.Writer) error {
```
trevorwhitney, line 196: I think this is another function we could write a test for.

pkg/querier/queryrange/codec.go
```diff
@@ -1610,6 +1668,21 @@ func (Codec) MergeResponse(responses ...queryrangebase.Response) (queryrangebase
 			},
 			Headers: headers,
 		}, nil
+	case *StructuredMetadataKeysResponse:
+		resp0 := responses[0].(*DetectedFieldsResponse)
+		headers := resp0.Headers
+
+		keys := []string{}
+		for _, r := range responses {
+			keys = append(keys, r.(*StructuredMetadataKeysResponse).Response.Keys...)
```
trevorwhitney, line 1677: do we want to dedupe or sort these? might also be worth adding a test for

pkg/querier/querier.go
```diff
@@ -1148,6 +1149,62 @@ func (q *SingleTenantQuerier) DetectedFields(ctx context.Context, req *logproto.
...
+			},
+		},
+	}
+
+	iters, err := q.SelectLogs(ctx, params)
+	if err != nil {
+		return nil, err
+	}
+
+  //TODO(twhitney): we can improve this in the future as we don't need to iterate lines for structured metadata, just chunks
```
trevorwhitney, line 1175: added this as a reminder... I don't think we have good methods for getting structured metadata per chunk, but since it's stored per chunk and not per line we can improve this in the future

trevorwhitney (commented):
Looking great! I added a test to the querier, and noted 2 other places tests would be nice if there are easy patterns to follow in existing tests.

pkg/querier/querier.go
```diff
@@ -1144,6 +1145,62 @@ func (q *SingleTenantQuerier) DetectedFields(ctx context.Context, req *logproto.
...
+	streams, err := streamsForFieldDetection(iters, 1000)
+	if err != nil {
+		return nil, err
+	}
+
+	structuredMetadata := parseStructuredMetadata(ctx, streams)
+
+	return &logproto.StructuredMetadataKeysResponse{
+		Keys: structuredMetadata,
+	}, nil
```
cyriltovena, line 1181: I'm fine with starting with this, but I think in the future we want to select a sample of chunk over the range and get only metadata from each chunk.

This would make it more accurate and snappier.

pkg/querier/querier.go
```diff
@@ -1144,6 +1145,62 @@ func (q *SingleTenantQuerier) DetectedFields(ctx context.Context, req *logproto.
...
+}
+
+func parseStructuredMetadata(ctx context.Context, streams logqlmodel.Streams) []string {
+	lbls := []string{}
+
+	for _, stream := range streams {
+		level.Debug(spanlogger.FromContext(ctx)).Log(
```
cyriltovena, line 1188: I think you need to remove that line. Too intense and it seems to also span log it.

pkg/querier/querier.go
svennergr: Yea, makes sense

pkg/querier/querier.go
```diff
@@ -1144,6 +1145,62 @@ func (q *SingleTenantQuerier) DetectedFields(ctx context.Context, req *logproto.
...
+			"structured_metadata", "true",
+			"msg", fmt.Sprintf("looking for structured metadata in stream %d with %d lines", stream.Hash, len(stream.Entries)))
+
+		for _, entry := range stream.Entries {
+			for _, l := range entry.StructuredMetadata {
+				if !slices.Contains(lbls, l.Name) {
```
cyriltovena, line 1194: you're better of using a map to ensure unicity. Otherwise you doing a lot of iteration as the slices grows.

Once the map is full, you can then transform into an array and sort it for stability.

cyriltovena: map[string]struct{}{}

pkg/querier/querier.go
svennergr: Also makes sense. Should I remove https://github.com/grafana/loki/blob/main/pkg/querier/querier.go#L1217-L1219 as well?

cyriltovena: Yes please

pkg/querier/querier.go
svennergr: Like https://github.com/grafana/loki/pull/13119/files#diff-a0b881d1b7b99f439716f189b50eef90ccdaea00845b087c0dbce0e148aa2c0eR1184-R1202 ?

svennergr: >Once the map is full, you can then transform into an array and sort it for stability.

do i have to sort it?

pkg/querier/querier.go
```diff
@@ -1144,6 +1145,61 @@ func (q *SingleTenantQuerier) DetectedFields(ctx context.Context, req *logproto.
...
+
+	for _, stream := range streams {
+		for _, entry := range stream.Entries {
+			for _, l := range entry.StructuredMetadata {
+				keysm[l.Name] = struct{}{}
+			}
+		}
+	}
+
+	keys := make([]string, 0, len(keysm))
```
cyriltovena, line 1195: you need to sort it I think to avoid unstable result ?

cyriltovena (approved):
LGTM

pkg/querier/querier.go
svennergr: It will get sorted here https://github.com/grafana/loki/pull/13119/files#diff-0670643837a81e9531f858f619539d92fc5be49b594b406d420215b308b61d97R1675-R1680 I think. Should be enough?

pkg/querier/querier.go
cyriltovena: In my experience anything coming out of a map key is better sorted because the order will be unstable and makes testing or client usage harder.

trevorwhitney:
@svennergr this was replace by #13604 right? can we close

svennergr:
Correct!

PR #13134, opened 2024-06-05
fix: properly return http status codes from ingester to querier for RPC function calls

cstyan:
We noticed this earlier with incorrect query syntax on a recording rule that was then sent to an ingester. The error returned to the end user had an HTTP 500 status code without much information because the grpc server was not seeing a valid status code in the result from the ingester, and was wrapping the error as a generic 500 status code.

@slim-bean and I discussed on slack and we think all functions the Querier can call on the Ingester via RPC should use the same error wrapping as we have in our server utils code. Looking at the generated GRPC code, this should be:
- Query
- QuerySample
- Label
- Series
- Tail
- TailersCount
- GetChunkIDs
- GetStats
- GetVolume
- GetDetectedLabels
- GetDetectedFields (this one can never return an error afaict)

Note that I have just done the easy thing for now, and I'll simplify the duplicated code in the morning

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	server_util "github.com/grafana/loki/v3/pkg/util/server"
 	"math/rand"
 	"net/http"
 	"os"
@@ -22,6 +23,7 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/grafana/dskit/concurrency"
+	"github.com/grafana/dskit/httpgrpc"
 	"github.com/grafana/dskit/modules"
 	"github.com/grafana/dskit/multierror"
 	"github.com/grafana/dskit/ring"
@@ -1041,6 +1043,15 @@ func (i *Ingester) asyncStoreMaxLookBack() time.Duration {
 
 // GetChunkIDs is meant to be used only when using an async store like boltdb-shipper or tsdb.
 func (i *Ingester) GetChunkIDs(ctx context.Context, req *logproto.GetChunkIDsRequest) (*logproto.GetChunkIDsResponse, error) {
+	var status int
+	gcr, err := i.getChunkIDs(ctx, req)
+	status, err = server_util.ClientHTTPStatusAndError(err)
+	err = httpgrpc.Errorf(status, "%s", err.Error())
+	return gcr, err
+}
+
+// GetChunkIDs is meant to be used only when using an async store like boltdb-shipper or tsdb.
+func (i *Ingester) getChunkIDs(ctx context.Context, req *logproto.GetChunkIDsRequest) (*logproto.GetChunkIDsResponse, error) {
 	orgID, err := tenant.TenantID(ctx)
 	if err != nil {
 		return nil, err
@@ -1168,6 +1179,14 @@ func (i *Ingester) Label(ctx context.Context, req *logproto.LabelRequest) (*logp
 
 // Series queries the ingester for log stream identifiers (label sets) matching a set of matchers
 func (i *Ingester) Series(ctx context.Context, req *logproto.SeriesRequest) (*logproto.SeriesResponse, error) {
+	var status int
+	sr, err := i.series(ctx, req)
+	status, err = server_util.ClientHTTPStatusAndError(err)
+	err = httpgrpc.Errorf(status, "%s", err.Error())
+	return sr, err
+}
+
+func (i *Ingester) series(ctx context.Context, req *logproto.SeriesRequest) (*logproto.SeriesResponse, error) {
 	instanceID, err := tenant.TenantID(ctx)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


pkg/ingester/ingester.go
```diff
@@ -1111,21 +1120,28 @@ func (i *Ingester) Label(ctx context.Context, req *logproto.LabelRequest) (*logp
 	pprof.Do(ctx, pprof.Labels("path", "read", "type", "labels", "tenant", userID), func(c context.Context) {
 		resp, err = instance.Label(ctx, req, matchers...)
 		if err != nil {
+			fmt.Println("error: ", err)
 			return
 		}
 		if req.Start == nil {
+			fmt.Println("asdf")
```
cyriltovena, line 1127:
```suggestion
```

pkg/ingester/ingester.go
```diff
@@ -1111,21 +1120,28 @@ func (i *Ingester) Label(ctx context.Context, req *logproto.LabelRequest) (*logp
 	pprof.Do(ctx, pprof.Labels("path", "read", "type", "labels", "tenant", userID), func(c context.Context) {
 		resp, err = instance.Label(ctx, req, matchers...)
 		if err != nil {
+			fmt.Println("error: ", err)
```
cyriltovena, line 1123:
```suggestion
```

pkg/ingester/ingester.go
```diff
@@ -1111,21 +1120,28 @@ func (i *Ingester) Label(ctx context.Context, req *logproto.LabelRequest) (*logp
...
 		if req.Start == nil {
+			fmt.Println("asdf")
+
 			return
 		}
 
 		// Only continue if the active index type is one of async index store types or QueryStore flag is true.
 		asyncStoreMaxLookBack := i.asyncStoreMaxLookBack()
 		if asyncStoreMaxLookBack == 0 && !i.cfg.QueryStore {
+			fmt.Println("qwer")
```
cyriltovena, line 1135:
```suggestion
```

pkg/ingester/ingester.go
```diff
@@ -1111,21 +1120,28 @@ func (i *Ingester) Label(ctx context.Context, req *logproto.LabelRequest) (*logp
...
+
 			return
 		}
 
 		var cs storage.Store
 		var ok bool
 		if cs, ok = i.store.(storage.Store); !ok {
+			fmt.Println("1234")
```
cyriltovena, line 1143:
```suggestion
```

pkg/ingester/ingester.go
```diff
@@ -1137,18 +1153,24 @@ func (i *Ingester) Label(ctx context.Context, req *logproto.LabelRequest) (*logp
 		start := adjustQueryStartTime(maxLookBackPeriod, *req.Start, time.Now())
 		if start.After(*req.End) {
 			// The request is older than we are allowed to query the store, just return what we have.
+			fmt.Println("ttttt")
```
cyriltovena, line 1156:
```suggestion
```

pkg/ingester/ingester.go
```diff
@@ -1137,18 +1153,24 @@ func (i *Ingester) Label(ctx context.Context, req *logproto.LabelRequest) (*logp
...
+
 			return
 		}
 		from, through := model.TimeFromUnixNano(start.UnixNano()), model.TimeFromUnixNano(req.End.UnixNano())
 
 		if req.Values {
 			storeValues, err = cs.LabelValuesForMetricName(ctx, userID, from, through, "logs", req.Name, matchers...)
 			if err != nil {
+				fmt.Println("error2: ", err)
```
cyriltovena, line 1165:
```suggestion
```

pkg/ingester/ingester.go
```diff
@@ -1137,18 +1153,24 @@ func (i *Ingester) Label(ctx context.Context, req *logproto.LabelRequest) (*logp
...
+
 				return
 			}
 		} else {
 			storeValues, err = cs.LabelNamesForMetricName(ctx, userID, from, through, "logs", matchers...)
 			if err != nil {
+				fmt.Println("error3: ", err)
```
cyriltovena, line 1172:
```suggestion
```

pkg/ingester/ingester.go
```diff
@@ -1160,6 +1182,7 @@ func (i *Ingester) Label(ctx context.Context, req *logproto.LabelRequest) (*logp
 	if err != nil {
 		return nil, err
 	}
+	fmt.Println("resp values: ", resp.Values)
```
cyriltovena, line 1185:
```suggestion
```

cyriltovena:
While this is a good idea to have. I think we should investigate how such queries end up on ingester, we should fail fast if this is not good syntax a querier or query-frontend should have caught it.

This was brought up during incident review meeting.

cstyan:
> we should fail fast if this is not good syntax a querier or query-frontend should have caught it.

Agreed, and I will look into this more. Do you have any objections to merging this as is while we continue to investigate why these queries are making it to the ingester in the first place?

cyriltovena (approved):
LGTM

PR #13137, opened 2024-06-05
feat: Tune Patterns query drain instance

benclive:
- Tunes the Patterns API Query Drain instance by setting SimTh to 1.0. This has the effect of merging & deduplicating patterns across streams without changing them or inserting more placeholder tokens which lead to very generic patterns.
- Adds metrics to see how many patterns are being returned & discarded by this Query Drain
  - I'm expecting the number of discarded patterns to be far higher than the number of retained patterns on each query.


cyriltovena (approved):
LGTM

PR #13167, opened 2024-06-07
fix: Fix bloom deleter PR after merge

salvacorts:
Fixes https://github.com/grafana/loki/pull/13153. Reimplemented `outdatedMetas` as it's currently implemented in the bloomcompactor pkg:
https://github.com/grafana/loki/blob/fbe7c559b5ed153fb46a1965c24180011a558b85/pkg/bloomcompactor/versioned_range.go#L213

**nit**: I'm removing the err returned by `outdatedMetas` since it's always nil.

### DIFF
--- a/pkg/bloombuild/planner/planner.go
+++ b/pkg/bloombuild/planner/planner.go
@@ -365,10 +365,7 @@ func (p *Planner) processTenantTaskResults(
 	}
 
 	combined := append(originalMetas, newMetas...)
-	outdated, err := outdatedMetas(combined)
-	if err != nil {
-		return fmt.Errorf("failed to find outdated metas: %w", err)
-	}
+	outdated := outdatedMetas(combined)
 	level.Debug(logger).Log("msg", "found outdated metas", "outdated", len(outdated))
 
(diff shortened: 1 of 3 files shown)


pkg/bloombuild/planner/planner.go
```diff
@@ -476,7 +473,7 @@ func (p *Planner) loadTenantWork(
 
 		// If this is the first this we see this table, initialize the map
 		if tenantTableWork[table] == nil {
-			tenantTableWork[table] = make(map[string][]v1.FingerprintBounds, tenants.Len())
+			tenantTableWork[table] = make(map[string][]v1.FingerprintBounds, tenants.Remaining())
```
grobinson-grafana, line 476: I don't know this code too well, but I assume the behavior of `Len` and `Remaining` here is the same – i.e. `Remaining` doesn't do additional filtering, for example.

chaudum (approved):

grobinson-grafana (approved):
LGTM!

pkg/bloombuild/planner/planner.go
salvacorts: Yup :)

PR #13171, opened 2024-06-07
fix: Correctly encode step when translating proto to http internally

benclive:
Fixes the encoding from a proto to a http request for QueryPatterns & DetectedFields & DetectedLabels
- The step parameter is represented internally as a int64 number of milliseconds but the HTTP API expects either a duration string or a int number of seconds. 
- When using HTTP between internal components, we were adding step to a query string as-is so the step parameter in milliseconds was being treated as seconds, hence the API eventually returned  one big aggregated datapoint instead of many smaller ones.
- This only shows up when **not** using `frontend: protobuf` so it was tough to reproduce locally. With protobuf encoding, everything worked as expected.
- The same bug was affecting DetectedFields too but we just didn't notice it yet.

Fixes https://github.com/grafana/loki/issues/13155

### DIFF
--- a/pkg/querier/queryrange/codec.go
+++ b/pkg/querier/queryrange/codec.go
@@ -919,7 +919,10 @@ func (c Codec) EncodeRequest(ctx context.Context, r queryrangebase.Request) (*ht
 			"end":         []string{fmt.Sprintf("%d", request.End.UnixNano())},
 			"line_limit":  []string{fmt.Sprintf("%d", request.GetLineLimit())},
 			"field_limit": []string{fmt.Sprintf("%d", request.GetFieldLimit())},
-			"step":        []string{fmt.Sprintf("%d", request.GetStep())},
+		}
+
+		if request.Step != 0 {
+			params["step"] = []string{fmt.Sprintf("%f", float64(request.Step)/float64(1e3))}
 		}
 
 		u := &url.URL{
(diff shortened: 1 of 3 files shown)


pkg/querier/queryrange/codec.go
```diff
@@ -919,7 +919,10 @@ func (c Codec) EncodeRequest(ctx context.Context, r queryrangebase.Request) (*ht
 			"end":         []string{fmt.Sprintf("%d", request.End.UnixNano())},
 			"line_limit":  []string{fmt.Sprintf("%d", request.GetLineLimit())},
 			"field_limit": []string{fmt.Sprintf("%d", request.GetFieldLimit())},
-			"step":        []string{fmt.Sprintf("%d", request.GetStep())},
+		}
+
+		if request.Step != 0 {
+			params["step"] = []string{fmt.Sprintf("%f", float64(request.Step)/float64(1e3))}
```
benclive, line 925: An alternative is to format the step as a duration string so 3000 becomes 3000ms instead of 3.000 seconds.

pkg/querier/queryrange/codec_test.go
```diff
@@ -202,6 +202,57 @@ func Test_codec_EncodeDecodeRequest(t *testing.T) {
 			Step:        30 * 1e3, // step is expected in ms; default is 0 or no step
 			AggregateBy: "series",
 		}, false},
+		{"detected_fields", func() (*http.Request, error) {
+			return DefaultCodec.EncodeRequest(ctx, &DetectedFieldsRequest{
```
benclive, line 206: I'm a bit confused as to why we use a wrapped queryrange.DetectedFieldsRequest and not the raw logproto.DetectedFieldsRequest here - is this the right thing to do?

pkg/querier/queryrange/codec_test.go
```diff
@@ -202,6 +202,57 @@ func Test_codec_EncodeDecodeRequest(t *testing.T) {
...
+				logproto.DetectedFieldsRequest{
+					Query:      `{foo="bar"}`,
+					Start:      start,
```
benclive, line 209: Other proto api objects use unix nanos for time, but these 3 (DetectedFields/DetectedLabels/QueryPatterns) use google.protobuf.Timestamp instead. Is there a reason for that? Should we be consistent across all our APIs and use one or the other?

pkg/querier/queryrange/codec_test.go
cyriltovena: the query-frontend is messy for sure I think it's fine.

cyriltovena (approved):
LGTM nice catch

PR #13162, opened 2024-06-06
fix(regression):  reverts grafana/loki#13039 to prevent use-after-free corruptions

owen-d:
Temporarily reverts grafana/loki#13039

Looks like that PR breaks the conventions around pool reuse (to safeguard use-after-free memory corruptions) described [here](https://github.com/grafana/loki/pull/13039/files#diff-ad299f9ea738f6f6d8e1a3b9fcc651506b3a241dbfbd95a60ab9d96b5b8b4d4aL115-L119) and [here](https://github.com/grafana/loki/pull/13039/files#diff-a9098e10e2bb5fb23648a6a08d55797ce96561e613358a4b931dac30c7661400L158-L160).

PR #13172, opened 2024-06-07
perf: Re-introduce fixed size memory pool for bloom querier

chaudum:

This PR re-introduces the fixed size memory pool that was originally introduced with #13039 but reverted with  #13162

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/storage/stores/shipper/bloomshipper/config/config.go
+++ b/pkg/storage/stores/shipper/bloomshipper/config/config.go
@@ -4,11 +4,16 @@ package config
 import (
 	"errors"
 	"flag"
+	"fmt"
+	"slices"
+	"strings"
 	"time"
 
 	"github.com/grafana/dskit/flagext"
 
 	"github.com/grafana/loki/v3/pkg/storage/chunk/cache"
+	lokiflagext "github.com/grafana/loki/v3/pkg/util/flagext"
+	"github.com/grafana/loki/v3/pkg/util/mempool"
 )
 
 type Config struct {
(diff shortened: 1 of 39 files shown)


pkg/storage/stores/shipper/bloomshipper/config/config.go
```diff
@@ -18,6 +23,7 @@ type Config struct {
 	BlocksCache         BlocksCacheConfig         `yaml:"blocks_cache"`
 	MetasCache          cache.Config              `yaml:"metas_cache"`
 	MetasLRUCache       cache.EmbeddedCacheConfig `yaml:"metas_lru_cache"`
+	MemoryManagement    MemoryManagementConfig    `yaml:"memory_management" doc:"hidden"`
```
chaudum, line 26: Explicitly hide this config from the docs.

pkg/storage/stores/shipper/bloomshipper/store.go
```diff
@@ -265,6 +267,13 @@ func (b *bloomStoreEntry) Client(_ model.Time) (Client, error) {
 	return b.bloomClient, nil
 }
 
+// Allocator implements Store.
+// While bloomStoreEntry implements the Store interface, this method must never
+// be used directly and therefore can safely return nil.
+func (b *bloomStoreEntry) Allocator() mempool.Allocator {
+	return nil
```
salvacorts, line 274: nit: I think would be better to no having to satisfy the interface in this way. Maybe we can have 
```go
type StoreWithAllocator interface {
    Store
    Allocator() mempool.Allocator
}
```

Ans only `BloomStore` implements that.

Having said that, if we don't want to do this, I think we should rather panic here.

salvacorts (approved):
LGTM. Left a nit.

pkg/storage/stores/shipper/bloomshipper/store.go
chaudum: Good idea. I will add the interface.

salvacorts (approved):

PR #13180, opened 2024-06-08
fix:  #13124 test-canary template doesn't respect test.enabled flag 

### LINKED ISSUE #13124: [helm] test-canary template doesn't respect test.enabled flag
**Describe the bug**
test-canary template only checks if Loki Canary is enabled (`lokiCanary.enabled`). It should also checks `test.enabled`.

**To Reproduce**
1. Install using Helm and using a value of `test.enabled: false`

**To fix it**

```
# test-canary.yaml
...
{{- if and .enabled $.Values.lokiCanary.enabled }}
...
```

**Expected behavior**
If `test.enabled` is false, `loki-helm-test` should not be deploy.

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: Helm

yincongcyincong:
Fixes  #13124

### DIFF
--- a/production/helm/loki/templates/tests/test-canary.yaml
+++ b/production/helm/loki/templates/tests/test-canary.yaml
@@ -1,5 +1,5 @@
 {{- with .Values.test }}
-{{- if $.Values.lokiCanary.enabled }}
+{{- if and .enabled $.Values.lokiCanary.enabled }}
 ---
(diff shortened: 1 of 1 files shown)


lindhe:
@yincongcyincong I like your PR. Why did you close it?

PR #13203, opened 2024-06-12
fix typo in _helpers.tpl

rdxmb:

fix typo, maybe function


JStickler:
Closing as fixed in current chart.  Thank you for taking the time to submit this and contributing to the Loki project.

PR #13223, opened 2024-06-14
feat: Optimize log parsing performance by using unsafe package

cyriltovena:
This commit modifies the log parsing logic in the `distributor` package to improve performance. It introduces the use of the `unsafe` package to work with log strings as byte slices, reducing memory allocations and improving efficiency.

Benchstat

```
name                                          old time/op    new time/op    delta
_optParseExtractLogLevelFromLogLineLogfmt-10    1.25µs ± 1%    0.66µs ± 7%  -46.65%  (p=0.008 n=5+5)

name                                          old alloc/op   new alloc/op   delta
_optParseExtractLogLevelFromLogLineLogfmt-10    4.15kB ± 0%    0.02kB ± 0%  -99.61%  (p=0.008 n=5+5)

name                                          old allocs/op  new allocs/op  delta
_optParseExtractLogLevelFromLogLineLogfmt-10      5.00 ± 0%      1.00 ± 0%  -80.00%  (p=0.008 n=5+5)

name                                        old time/op    new time/op    delta
_optParseExtractLogLevelFromLogLineJson-10    1.20µs ± 0%    0.93µs ±20%  -22.34%  (p=0.008 n=5+5)

name                                        old alloc/op   new alloc/op   delta
_optParseExtractLogLevelFromLogLineJson-10      293B ± 0%       16B ± 0%     ~     (p=0.079 n=4+5)

name                                        old allocs/op  new allocs/op  delta
_optParseExtractLogLevelFromLogLineJson-10      6.00 ± 0%      1.00 ± 0%  -83.33%  (p=0.008 n=5+5)
```


pkg/distributor/distributor.go
```diff
@@ -917,56 +919,57 @@ func detectLogLevelFromLogEntry(entry logproto.Entry, structuredMetadata labels.
...
+func getValueUsingJSONParser(log []byte) []byte {
 	for allowedLabel := range allowedLabelsForLevel {
-		l, err := jsonparser.GetString([]byte(log), allowedLabel)
+		l, _, _, err := jsonparser.Get(log, allowedLabel)
 		if err == nil {
 			return l
 		}
 	}
-	return logLevelUnknown
+	return nil
```
DylanGuedes, line 972: ditto, might be unknown here?

pkg/distributor/distributor.go
```diff
@@ -917,56 +919,57 @@ func detectLogLevelFromLogEntry(entry logproto.Entry, structuredMetadata labels.
...
+
+	d := logfmt.NewDecoder(line)
+	for !d.EOL() && d.ScanKeyval() {
 		if _, ok := allowedLabelsForLevel[string(d.Key())]; ok {
-			return string(d.Value())
+			return (d.Value())
 		}
 	}
-	return logLevelUnknown
+	return nil
```
DylanGuedes, line 962: fyi: `logLevelUnknown` has value="unknown"

cyriltovena: yes, in that case the function returns nil and the switch will get into the default case.

DylanGuedes (approved):

PR #13224, opened 2024-06-14
refactor: Introduce context cause to our code

DylanGuedes:
Introduce the usage of Go new ' WithCause' functions to make our `context cancellation` errors more descriptive.
I'm not yet changing every `context.WithCancel` here because, from what I learned, places were we immediately cancel the context won't strongly benefit from this.

### DIFF
--- a/pkg/querier/worker/scheduler_processor.go
+++ b/pkg/querier/worker/scheduler_processor.go
@@ -21,6 +21,7 @@ import (
 	"github.com/grafana/dskit/user"
 	otgrpc "github.com/opentracing-contrib/go-grpc"
 	"github.com/opentracing/opentracing-go"
+	"github.com/pkg/errors"
 	"go.uber.org/atomic"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/health/grpc_health_v1"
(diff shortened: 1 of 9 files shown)


pkg/querier/worker/scheduler_processor.go
```diff
@@ -89,7 +90,7 @@ func (sp *schedulerProcessor) processQueriesOnSingleStream(workerCtx context.Con
 	// Run the querier loop (and so all the queries) in a dedicated context that we call the "execution context".
 	// The execution context is cancelled once the workerCtx is cancelled AND there's no inflight query executing.
 	execCtx, execCancel, inflightQuery := newExecutionContext(workerCtx, sp.log)
-	defer execCancel()
+	defer execCancel(errors.New("scheduler processor execution context canceled"))
```
cstyan, line 93: should we include more context here? the comment above makes it pretty clear when we would be cancelling the execution context
```
// The execution context is cancelled once the workerCtx is cancelled AND there's no inflight query executing.
```

cstyan (commented):
these errors are not propagated back to end users, right? except maybe query timeouts?

pkg/querier/worker/scheduler_processor.go
DylanGuedes: > should we include more context here? the comment above makes it pretty clear when we would be cancelling the execution context

yep it is pretty much what is in the comment, but the comment won't be present on the log line so you won't have access to it when querying the logs.

DylanGuedes:
> these errors are not propagated back to end users, right? except maybe query timeouts?

yep, I'm expecting the query timeouts to be propagated back but the others won't happen

cstyan (approved):

PR #12996, opened 2024-05-20
docs: update otlp ingestion with correct endpoint and add endpoint to reference api docs

vitoorgomes:
**What this PR does / why we need it**:
This PRs fixes some parts of the documentation about the otlp ingestion endpoints. 

The original documentation created on PR #11026 gives a great explanation and how to use it, but it was missing the complete endpoint and it was driving me crazy that I was receiving a 404 when trying to push data from otel collector to loki.

So I found this #12709 and things were clearer and after using this endpoint on my otel collector exporter, it worked like it was supposed to.

I also took the liberty to add in the HTTP API Reference documentation, since it's a good place to also check for all endpoints

**Special notes for your reviewer**:

**Checklist**
- [x] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [x] Documentation added

PR #13235, opened 2024-06-17
docs: Update OTLP endpoint

JStickler:
Reverts incorrect endpoint introduced in #12996

### DIFF
--- a/docs/sources/reference/loki-http-api.md
+++ b/docs/sources/reference/loki-http-api.md
@@ -24,7 +24,7 @@ Authorization needs to be done separately, for example, using an open-source loa
 These endpoints are exposed by the `distributor`, `write`, and `all` components:
 
 - [`POST /loki/api/v1/push`](#ingest-logs)
-- [`POST /otlp`](#ingest-logs-using-otlp)
+- [`POST /otlp/v1/logs`](#ingest-logs-using-otlp)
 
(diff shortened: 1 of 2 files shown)


docs/sources/reference/loki-http-api.md
```diff
@@ -264,13 +264,17 @@ curl -H "Content-Type: application/json" \
 ## Ingest logs using OTLP
 
 ```bash
-POST /otlp/v1/logs
+POST /otlp
```
Jayclifford345, line 267: same here I belive. The true route is `/otlp/v1/logs` its just the otel collector adds the last bit for you

docs/sources/reference/loki-http-api.md
```diff
@@ -24,7 +24,7 @@ Authorization needs to be done separately, for example, using an open-source loa
 These endpoints are exposed by the `distributor`, `write`, and `all` components:
 
 - [`POST /loki/api/v1/push`](#ingest-logs)
-- [`POST /otlp/v1/logs`](#ingest-logs-using-otlp)
+- [`POST /otlp`](#ingest-logs-using-otlp)
```
Jayclifford345, line 27: This should still be `/otlp/v1/logs` as if you where to call the endpoint without `/v1/logs` I belive it would fail.  Its only the otel collector which adds the `v1/logs` automatically

Jayclifford345 (approved):
LGTM!

PR #13255, opened 2024-06-18
refactor: Remove unnecessary spanlogger usage

DylanGuedes:
Historically, we've been using `spanlogger` to enrich our spans. That isn't the right usage for it (as of now), as they'll be emitting log lines. Our current usage is causing:
- Noisy/important log lines less visible
- Unnecessary logging volume
- Spread of the bad practice.
With this PR I'm moving away from `spanlogger` all places where the log message is too bad or nonexistent. That's because these cases are indicating we don't care about that log line at all, and only about the data injected in the span.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


pkg/logql/engine.go
```diff
@@ -265,7 +263,7 @@ func (q *query) Exec(ctx context.Context) (logqlmodel.Result, error) {
 	queueTime, _ := ctx.Value(httpreq.QueryQueueTimeHTTPHeader).(time.Duration)
 
 	statResult := statsCtx.Result(time.Since(start), queueTime, q.resultLength(data))
-	statResult.Log(level.Debug(spLogger))
+	statResult.LogWithSpan(sp)
```
slim-bean, line 266: I wonder about a little bit different approach where we leave the original stats.Log() method and use that to print the stats in metrics.go.

But here do: `sp.LogKV(statResult.KVList()...)`

DylanGuedes: i like yours style more but fyi we use `LogWithSpan` at other places. I'll try using `LogKV` there too then.

DylanGuedes: pushed https://github.com/grafana/loki/pull/13255/commits/0c9570ada806d7ab5b99f3fa3ebc5a0c159e1bbb addressing it

slim-bean (approved):
LGTM!

PR #13272, opened 2024-06-20
feat(wal): Benchmark and improve WAL writes using Reset.

cyriltovena:

This adds a benchmark and some improvement to the write path of wal segment.

Fixes https://github.com/grafana/loki-private/issues/1005 https://github.com/grafana/loki-private/issues/1004

```
name       old time/op    new time/op    delta
Writes-16    11.9ms ±29%     6.5ms ± 2%  -44.84%  (p=0.008 n=5+5)

name       old alloc/op   new alloc/op   delta
Writes-16    12.9MB ± 9%     0.1MB ± 1%  -98.84%  (p=0.008 n=5+5)

name       old allocs/op  new allocs/op  delta
Writes-16     2.36k ±19%     0.30k ± 0%  -87.13%  (p=0.008 n=5+5)
```


pkg/storage/wal/segment.go
```diff
@@ -49,35 +57,39 @@ type streamSegment struct {
 }
 
 // NewWalSegmentWriter creates a new WalSegmentWriter.
-func NewWalSegmentWriter() *SegmentWriter {
-	return &SegmentWriter{
-		streams: swiss.NewMap[streamID, *streamSegment](64),
-		buf1:    encoding.EncWith(make([]byte, 0, 4)),
+func NewWalSegmentWriter() (*SegmentWriter, error) {
+	idxWriter, err := index.NewWriter(context.TODO())
```
benclive, line 61: Do we have to pass a ctx to the index? I think we can remove it because we only ever pass `context.TODO()`. We also don't pass a context to the parent SegmentWriter so the utility is very limited.

pkg/storage/wal/segment.go
```diff
@@ -49,35 +57,39 @@ type streamSegment struct {
...
-			// prometheus bucketed pool
-			// https://pkg.go.dev/github.com/prometheus/prometheus/util/pool
-			entries:  make([]*logproto.Entry, 0, 64),
-			lbls:     lbls,
-			tenantID: tenantID,
-		}
+		s = streamSegmentPool.Get().(*streamSegment)
+		s.lbls = lbls
+		s.tenantID = tenantID
+		s.entries = s.entries[:0]
```
benclive, line 89: Should a streamSegment have a `reset()` method?
It looks like only one line to reset them but it would maintain consistency with our other Readers/Writers.

pkg/storage/wal/segment.go
cyriltovena: Yes I'll look into that.

pkg/storage/wal/segment_test.go
```diff
@@ -186,3 +191,172 @@ func TestMultiTenantWrite(t *testing.T) {
...
+	t.Logf("Total chunks %d\n", len(sizes.Series))
+	t.Logf("Index size  %s\n", humanize.Bytes(uint64(sizes.Index)))
+	sizesString := ""
+	for _, size := range sizes.Series {
+		sizesString += humanize.Bytes(uint64(size)) + ", "
+	}
+	t.Logf("Series sizes: [%s]\n", sizesString)
+}
+
+func TestReset(t *testing.T) {
```
benclive, line 261: Could you add/modify this test case so the re-use case adds more data than the initial case?
If we had re-use errors, we'd likely panic when re-using a small buffer buit we'd be less likely to panic when re-using a large buffer.

pkg/storage/wal/segment_test.go
```diff
@@ -186,3 +191,172 @@ func TestMultiTenantWrite(t *testing.T) {
...
+			return writer
+		},
+	}
+
+	b.ResetTimer()
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		writer := pool.Get().(*SegmentWriter)
+
+		dst.Reset()
```
benclive, line 351: This benchmark isn't running in parallel (since dst is reused). You could remove the pool for writers from the benchmark, or add a pool for dst and run the whole thing in parallel :)

benclive (approved):
LGTM! Just one comment on the tests but the benchmark should already cover it so happy to approve regardless.

PR #13273, opened 2024-06-20
chore: Make dedicated iterator package

chaudum:

Iterators of various types are widely used throughout the Loki code base. With the recent code additions of the bloom filters, a new set of utility functions for iterators emerged in `github.com/grafana/loki/v3/pkg/storage/bloom/v1`. The package defines interfaces for various common types, but also provides implementations to create and compose new iterators that implement these interfaces. This new package uses Go Generics.

- However, at the current state, there are multiple iterator interfaces and implementations, which adds cognitive overhead in understanding how they work and which ones should be used. The idea is to unify them into a single iterator lib for Loki.
- As a first step towards a single iterator library for the Loki code base, this PR moves the utilities from the `pkg/storage/bloom/v1` package to the `pkg/iter/v2` package.
- Second, it changes the existing `EntryIterator` and `SampleIterator` iterators to "inherit" `v2.CloseIterator` in order to expose the same function names.
- And lastly, naming conventions of iterator interfaces and structs are unified.

The basic iterator interface (defined in `pkg/iter/v2/interface.go`) looks like so:

```go
// Usage:
//
//	for it.Next() {
//	    curr := it.At()
//	    // do something
//	}
//	if it.Err() != nil {
//	    // do something
//	}
type Iterator[T any] interface {
	Next() bool
	Err() error
	At() T
}
```

**Discussion**

I proposed to use the `Err()` function for getting the optional error, because it was used in the `pkg/storage/bloom/v1` package. However, in existing iterators for entries and samples, this function was named `Error()`.
I don't have strong arguments for or against one or the other name.

**Next steps**

This PR intentionally leaves the duplicate implementation of various iterator types, such a heap sort iterators, slice iterators, or peeking iterators. These should be consolidated in a follow up PR.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


cyriltovena (approved):
LGTM

PR #13288, opened 2024-06-21
perf(blooms): always return bloom pages to allocator

owen-d:


pkg/bloomgateway/processor.go
```diff
@@ -113,16 +114,6 @@ func (p *processor) processTasks(ctx context.Context, tenant string, day config.
...
 func (p *processor) processBlocks(ctx context.Context, bqs []*bloomshipper.CloseableBlockQuerier, data []blockWithTasks) error {
-
-	defer func() {
-		for i := range bqs {
-			if bqs[i] == nil {
-				continue
-			}
-			bqs[i].Close()
-		}
-	}()
```
chaudum, line 124: Maybe I miss something, but we don't seem to close the block remaining block queriers any more in case one `p.processBlock()` fails, since `concurrency.ForEachJob()` returns on first error.

pkg/bloomgateway/processor.go
```diff
@@ -136,15 +127,23 @@ func (p *processor) processBlocks(ctx context.Context, bqs []*bloomshipper.Close
...
 	})
 }
 
-func (p *processor) processBlock(_ context.Context, blockQuerier *v1.BlockQuerier, tasks []Task) error {
+func (p *processor) processBlock(_ context.Context, bq *bloomshipper.CloseableBlockQuerier, tasks []Task) (err error) {
+	defer func() {
+		var errs multierror.MultiError
+		errs.Add(err)
+		errs.Add(bq.Close())
+		err = errs.Err()
```
chaudum, line 143: That looks wrong, since on line 141 you add `err` to the multi-error.

owen-d: That's capturing the `err` from the function signature and adding it to the multi-error (which can now have a main error from the body of the function call and/or an error from cleanup.

chaudum (approved):
lgtm

PR #13305, opened 2024-06-25
chore: Optimize line_tokenizer for better performance

cyriltovena:
The line_tokenizer code has been optimized to improve performance. The `make` function for the `tokens` slice has been modified to use a capacity of 64 instead of the length of the line. 

Hopefully this helps keep allocation under control.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/pattern/drain/line_tokenizer.go
+++ b/pkg/pattern/drain/line_tokenizer.go
@@ -38,7 +38,7 @@ func newPunctuationTokenizer() *punctuationTokenizer {
 }
 
 func (p *punctuationTokenizer) Tokenize(line string) ([]string, interface{}) {
-	tokens := make([]string, len(line))                  // Maximum size is every character is punctuation
+	tokens := make([]string, 0, 64)                      // Maximum size is every character is punctuation
 	spacesAfter := make([]int, strings.Count(line, " ")) // Could be a bitmap, but it's not worth it for a few bytes.
 
 	start := 0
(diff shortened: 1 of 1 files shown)


pkg/pattern/drain/line_tokenizer.go
```diff
@@ -38,7 +38,7 @@ func newPunctuationTokenizer() *punctuationTokenizer {
 }
 
 func (p *punctuationTokenizer) Tokenize(line string) ([]string, interface{}) {
-	tokens := make([]string, len(line))                  // Maximum size is every character is punctuation
+	tokens := make([]string, 0, 64)                      // Maximum size is every character is punctuation
```
benclive, line 41: Was 64 just an educated guess? 
A potential follow up is to add a metric for "characters per token" or "tokens per line" so we can get a better estimate here - we can probably improve perf a bit more with a better start point.

pkg/pattern/drain/line_tokenizer.go
```diff
@@ -51,22 +51,22 @@ func (p *punctuationTokenizer) Tokenize(line string) ([]string, interface{}) {
 		included := char < 128 && p.includeDelimiters[char] != 0
 		if char == ' ' || included || unicode.IsPunct(char) {
 			if i > start {
-				tokens[nextTokenIdx] = line[start:i]
+				tokens = append(tokens, line[start:i])
 				nextTokenIdx++
```
benclive, line 55: You can delete nextTokenIdx completely now you're using append since it was only used to resize the returned slice.
You can potentially do the same for the spaces slice (use append, delete nextSpaceIdx) too.

PR #13306, opened 2024-06-25
refactor(blooms): Implement retry in builder

salvacorts:
This PR implements a retry with backoff in the bloom builder. We retry to connect to the planner at two levels:
- At the top of the receive and build loop
- When we try to send back the task result to the planner

- We do not retry when we send a builder shutdown notification to the planner. Rationale being that this more like a best effort and we don't want to prevent a builder from shutting down.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/bloombuild/builder/builder.go
+++ b/pkg/bloombuild/builder/builder.go
@@ -10,6 +10,7 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/google/uuid"
+	"github.com/grafana/dskit/backoff"
 	"github.com/grafana/dskit/services"
 	"github.com/grafana/dskit/user"
 	"github.com/pkg/errors"
@@ -110,12 +111,38 @@ func (b *Builder) stopping(_ error) error {
 }
 
 func (b *Builder) running(ctx context.Context) error {
+	// Try to re-establish the connection up to 5 times.
+	retries := backoff.New(context.Background(), backoff.Config{
+		MinBackoff: 1 * time.Second,
+		MaxBackoff: 10 * time.Second,
+		MaxRetries: 5,
+	})
+
+	for retries.Ongoing() {
+		err := b.connectAndBuild(ctx)
+		if err == nil {
+			break
+		}
+
+		level.Error(b.logger).Log("msg", "failed to connect and build. Retrying", "err", err)
+		retries.Wait()
+	}
+
+	if err := retries.Err(); err != nil {
+		return fmt.Errorf("failed to connect and build: %w", err)
+	}
+
+	return nil
+}
+
+func (b *Builder) connectAndBuild(
+	ctx context.Context,
+) error {
 	opts, err := b.cfg.GrpcConfig.DialOption(nil, nil)
 	if err != nil {
 		return fmt.Errorf("failed to create grpc dial options: %w", err)
 	}
 
-	// TODO: Wrap hereafter in retry logic
 	conn, err := grpc.DialContext(ctx, b.cfg.PlannerAddress, opts...)
 	if err != nil {
 		return fmt.Errorf("failed to dial bloom planner: %w", err)
--- a/pkg/bloombuild/builder/builder.go
+++ b/pkg/bloombuild/builder/builder.go
@@ -112,10 +112,10 @@ func (b *Builder) stopping(_ error) error {
 
 func (b *Builder) running(ctx context.Context) error {
 	// Retry if the connection to the planner is lost.
-	retries := backoff.New(context.Background(), b.cfg.BackoffConfig)
+	retries := backoff.New(ctx, b.cfg.BackoffConfig)
 	for retries.Ongoing() {
 		err := b.connectAndBuild(ctx)
-		if err == nil {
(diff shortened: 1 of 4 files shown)


pkg/bloombuild/builder/builder.go
```diff
@@ -110,12 +111,33 @@ func (b *Builder) stopping(_ error) error {
 }
 
 func (b *Builder) running(ctx context.Context) error {
+	// Retry if the connection to the planner is lost.
+	retries := backoff.New(context.Background(), b.cfg.BackoffConfig)
```
chaudum, line 115: What's the reason for creating a new context instead of using `ctx`?

pkg/bloombuild/builder/builder_test.go
```diff
@@ -26,6 +26,7 @@ import (
 
 func Test_BuilderLoop(t *testing.T) {
 	logger := log.NewNopLogger()
+	//logger := log.NewLogfmtLogger(os.Stdout)
```
chaudum, line 29: Could be removed I assume

pkg/bloombuild/builder/builder_test.go
```diff
@@ -99,41 +126,56 @@ func Test_BuilderLoop(t *testing.T) {
 
 type fakePlannerServer struct {
 	tasks          []*protos.ProtoTask
-	completedTasks atomic.Int64
+	completedTasks int
 	shutdownCalled bool
 
-	addr       string
+	lisAddr    string
```
chaudum, line 132: nit

```suggestion
	listenAddr    string
```

pkg/bloombuild/builder/builder_test.go
```diff
@@ -99,41 +126,56 @@ func Test_BuilderLoop(t *testing.T) {
 
 type fakePlannerServer struct {
 	tasks          []*protos.ProtoTask
-	completedTasks atomic.Int64
+	completedTasks int
```
chaudum, line 129: Is the data race a false positive?

```
WARNING: DATA RACE
Read at 0x00c00073dc58 by goroutine 79:
  github.com/grafana/loki/v3/pkg/bloombuild/builder.(*fakePlannerServer).CompletedTasks()
      /home/christian/sandbox/grafana/loki/pkg/bloombuild/builder/builder_test.go:204 +0xd2c
```

pkg/bloombuild/builder/config.go
```diff
@@ -3,20 +3,27 @@ package builder
...
 }
 
 // RegisterFlagsWithPrefix registers flags for the bloom-planner configuration.
 func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.StringVar(&cfg.PlannerAddress, prefix+".planner-address", "", "Hostname (and port) of the bloom planner")
 	cfg.GrpcConfig.RegisterFlagsWithPrefix(prefix+".grpc", f)
+
+	f.DurationVar(&cfg.BackoffConfig.MinBackoff, prefix+".backoff.min-backoff", 1*time.Second, "Minimum backoff time")
+	f.DurationVar(&cfg.BackoffConfig.MaxBackoff, prefix+".backoff.max-backoff", 10*time.Second, "Maximum backoff time")
+	f.IntVar(&cfg.BackoffConfig.MaxRetries, prefix+".backoff.max-retries", 5, "Maximum number of times to retry an operation")
```
chaudum, line 26: Is there a reason why you did not use
`f.BackoffConfig.RegisterFlagsWithPrefix()`

pkg/bloombuild/builder/builder.go
salvacorts: Changed. Thank you!

pkg/bloombuild/builder/builder_test.go
salvacorts: I'm leaving this around for debugging if necessary but happy to remove them if you feel otherwise

pkg/bloombuild/builder/builder_test.go
salvacorts: I think it can actually happen. Changed back to use the atomic.

pkg/bloombuild/builder/config.go
salvacorts: I think the defaults should be fine. Changed to use `f.BackoffConfig.RegisterFlagsWithPrefix()`

pkg/bloombuild/builder/builder_test.go
```diff
@@ -149,14 +192,19 @@ func (f *fakePlannerServer) BuilderLoop(srv protos.PlannerForBuilder_BuilderLoop
 		if _, err := srv.Recv(); err != nil {
 			return fmt.Errorf("failed to receive task response: %w", err)
 		}
-		f.completedTasks.Add(1)
+		f.completedTasks.Inc()
+		time.Sleep(10 * time.Millisecond) // Simulate task processing time to add some latency.
```
chaudum, line 196: Should the sleep be before the task is counted as completed?

salvacorts: Not necessarilly but I changed that.

chaudum (approved):

PR #13315, opened 2024-06-25
feat: ingester-RF1 Push endpoint

benclive:
* Implements Push API on the RF1 ingester
* This writes the pushed log lines to a WAL segment. The segment is recreated & the old segment written to the store every 500ms (configurable)
* API requests wait on the successful flushing of the segment they wrote their entries to, or request timeout. 
* Vast majority of the code is copied from the existing ingester & there are still a few areas that could be cleaned up
* None of the tests have been copied over yet (!) as I wanted to get a working baseline to collaborate on further ASAP.
* Should support all object stores via PutObject but I've only tested using the local file store.

Fixes https://github.com/grafana/loki-private/issues/1016


pkg/storage/wal/segment.go
```diff
@@ -28,10 +28,12 @@ var (
 	streamSegmentPool = sync.Pool{
 		New: func() interface{} {
 			return &streamSegment{
+				lock:    &sync.Mutex{},
 				entries: make([]*logproto.Entry, 0, 4096),
 			}
 		},
 	}
+	tenantLabel = "__loki_tenant__"
```
benclive, line 36: Had to copy this from `tsdb.TenantLabel` because of a circular import between tsdb & wal via indexWriter.

pkg/ingester-rf1/ingester.go
```diff
@@ -260,15 +286,20 @@ func New(cfg Config, clientConfig client.Config, store Store, limits Limits, con
...
+		streamRateCalculator: NewStreamRateCalculator(),
 		writeLogManager:      writefailures.NewManager(logger, registerer, writeFailuresCfg, configs, "ingester_rf1"),
 		customStreamsTracker: customStreamsTracker,
 		readRing:             readRing,
+		flushCtx: &flushCtx{
+			lock:            &sync.RWMutex{},
+			flushDone:       make(chan struct{}),
+			newCtxAvailable: make(chan struct{}),
+			segmentWriter:   segmentWriter,
+		},
```
cyriltovena, line 299: Interesting choice here. 

I thought we would need multiple flushes routine per ingester to be able to flush faster, right now if a single flush is getting slow down (happens) all other requests will be affected.

Not a problem though let's put it to the test in dev.

pkg/ingester-rf1/ingester.go
```diff
@@ -540,7 +578,25 @@ func (i *Ingester) loop() {
 	for {
 		select {
 		case <-flushTicker.C:
-			i.sweepUsers(false, true)
+			//i.logger.Log("msg", "starting periodic flush")
+			i.flushCtx.lock.Lock() // Stop new chunks being written while we swap destinations - we'll never unlock as this flushctx can no longer be used.
```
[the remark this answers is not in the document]
cyriltovena: I think the Timer must be created as part of the context/request so that we can handle multiple flush timeout for different set of requests.

pkg/storage/chunk/client/object_client.go
```diff
@@ -105,6 +106,15 @@ func (o *client) Stop() {
 	o.store.Stop()
 }
 
+func (o *client) PutWal(ctx context.Context, segment *wal.SegmentWriter) error {
+	buffer := bytes.NewBuffer(nil)
+	_, err := segment.WriteTo(buffer)
+	if err != nil {
+		return err
+	}
+	return o.store.PutObject(ctx, "wal-segment-"+time.Now().UTC().Format(time.RFC3339Nano), bytes.NewReader(buffer.Bytes()))
```
cyriltovena, line 115: Can the new ingester deal with this interface instead ?  `PutObject`

This way we don't leak our storage into object storage clients.

pkg/ingester-rf1/ingester.go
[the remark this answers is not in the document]
cyriltovena: We probably also want to flush based on  input size to limit block size ? I think we can start with just this for now.

vendor/github.com/grafana/loki/pkg/push/push-rf1.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
```
cyriltovena, line 1: What's that for ?

pkg/ingester-rf1/ingester.go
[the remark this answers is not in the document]
benclive: See my comment below on the multiple flushes.
I do need to add the input size to this, yes! That was my next task but I wanted to get early feedback on this XXL PR first - ideally we can parallelise more work on top of this starting point.

pkg/ingester-rf1/ingester.go
benclive: We do have multiple flush routines - the idea was to have one "active" flushCtx at a time, which all Pushs are Appending to.
Every time the timer fires, it will swap the flushCtx for a new one (new wal segment, etc.) as fast as possible - it's just a variable assignment - and then hand off the old flushCtx to be flushed async.
I fully expect flushing to storage to be the slowest part of this, so I'm reusing the existing flush loops. We can have multiple flushes happening in parallel, and only the Pushes that appended to that flush would be waiting. New Pushes would be unaffected as they'd be Appending to a completely different flushCtx.

Definitely open to suggestions if you have better ideas though!

pkg/storage/chunk/client/object_client.go
benclive: I think thats a good idea but for now I was just copying the existing interfaces which use "PutChunks" everywhere. There is a slightly lower level interface which only has PutObject, but our storage layer seems to switch between them frequently.

vendor/github.com/grafana/loki/pkg/push/push-rf1.proto
benclive: Left over from the base PR. Not sure why git is picking up on it! But it was so we could have a separate grpc service for this new ingester, which is needed so traffic is routed correctly.

cyriltovena (approved):
LGTM

Let's merge and improve from there. Build is still failing.

pkg/storage/chunk/client/object_client.go
vlad-diachenko: I addressed it in https://github.com/grafana/loki/pull/13342/commits/b4ca98be8441f09f095f6edc44648ffcc97e54b3

PR #13316, opened 2024-06-25
fix(retry): fix retries when using protobuf encoding

ashwanthgoli:
`retry.go` expects the Details field of grpc status to be populated which it then uses to read the http response code, it fallbacks to retrying if this field is not set.

But with protobuf encoding Details is not populated and loki retries 4xxs. I do not think it is necessary to set this. 

`httpgrpc.HTTPResponseFromError` gets the Status from the error and [additionally tries to decode the http response](https://github.com/grafana/loki/blob/76a764bdca863734176ea5db714da0817d8a4151/vendor/github.com/grafana/dskit/httpgrpc/httpgrpc.go#L141) from Details field. But all `retry.go` needs is the Code from Status field which would be set for both encoding formats, so replacing this call with `status.FromError()` which only pulls out the Status from error should fix the retry behaviour.

**references:**
- Query response status with protobuf encoding is populated here https://github.com/grafana/loki/blob/76a764bdca863734176ea5db714da0817d8a4151/pkg/querier/worker/util.go#L142
- Status from Query response is decoded here. Details are not populated https://github.com/grafana/loki/blob/76a764bdca863734176ea5db714da0817d8a4151/pkg/querier/queryrange/marshal.go#L208
- http encoding populates the Details field here along with Status.Code
https://github.com/grafana/loki/blob/76a764bdca863734176ea5db714da0817d8a4151/pkg/querier/queryrange/codec.go#L667

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/querier/queryrange/queryrangebase/retry.go
+++ b/pkg/querier/queryrange/queryrangebase/retry.go
@@ -6,8 +6,8 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/gogo/status"
 	"github.com/grafana/dskit/backoff"
-	"github.com/grafana/dskit/httpgrpc"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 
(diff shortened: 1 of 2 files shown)


pkg/querier/queryrange/queryrangebase/retry.go
```diff
@@ -89,8 +89,8 @@ func (r retry) Do(ctx context.Context, req Request) (Response, error) {
 		}
 
 		// Retry if we get a HTTP 500 or a non-HTTP error.
-		httpResp, ok := httpgrpc.HTTPResponseFromError(err)
-		if !ok || httpResp.Code/100 == 5 {
+		status, ok := status.FromError(err)
```
duricanikolic, line 92: Could this approach work for you:
```
  if !IsClientError(err) {
    ...
  }
```
where
```
func isClientError(err error) bool {
	if grpcutil.ErrorToStatusCode(err); code/100 == 4 {
		return true
	}
	return false
}
```

chaudum (approved):
lgtm

PR #13321, opened 2024-06-25
fix: support multi-zone ingesters when converting global to local limits for streams in limiter.go

JordanRushing:

Updates the calculation for stream limits to support multi-zone ingester deployments while maintaining the old logic for single-zone deployments.

re:
- https://github.com/grafana/loki/pull/13006
- https://github.com/grafana/loki/pull/13103
- https://github.com/grafana/loki/pull/13254
- https://github.com/grafana/loki/pull/13030
- https://github.com/grafana/loki/pull/13231
- https://github.com/grafana/loki/pull/13314

N/A

Requires https://github.com/grafana/dskit/pull/526 to be merged and the version of dskit used by Loki to be updated to support `HealthyInstancesInZoneCount()`

**Checklist**
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/ingester/limiter.go
+++ b/pkg/ingester/limiter.go
@@ -20,6 +20,8 @@ const (
 // to count members
 type RingCount interface {
 	HealthyInstancesCount() int
+	HealthyInstancesInZoneCount() int
+	ZonesCount() int
 }
 
 type Limits interface {
@@ -84,7 +86,7 @@ func (l *Limiter) GetStreamCountLimit(tenantID string) (calculatedLimit, localLi
 	// We can assume that streams are evenly distributed across ingesters
 	// so we do convert the global limit into a local limit
 	globalLimit = l.limits.MaxGlobalStreamsPerUser(tenantID)
-	adjustedGlobalLimit = l.convertGlobalToLocalLimit(globalLimit)
+	adjustedGlobalLimit = l.convertGlobalToLocalLimit(globalLimit, localLimit)
 
 	// Set the calculated limit to the lesser of the local limit or the new calculated global limit
 	calculatedLimit = l.minNonZero(localLimit, adjustedGlobalLimit)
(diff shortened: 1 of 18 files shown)


pkg/ingester/limiter.go
```diff
@@ -105,24 +107,33 @@ func (l *Limiter) minNonZero(first, second int) int {
 	return first
 }
 
-func (l *Limiter) convertGlobalToLocalLimit(globalLimit int) int {
+func (l *Limiter) convertGlobalToLocalLimit(globalLimit, maxStreamsPerUser int) int {
```
vlad-diachenko, line 110: we do not need to pass local limit here, local limit is used on line number 92 .... 
local limit is a limit per ingester, regardless how many ingesters we have in the ring.

pkg/ingester/limiter.go
```diff
@@ -105,24 +107,33 @@ func (l *Limiter) minNonZero(first, second int) int {
...
 
-	return 0
+	ingestersInZone := l.ring.HealthyInstancesInZoneCount()
+	if ingestersInZone == 0 {
+		return 0 // Avoid division by zero
+	}
+
+	newLimit := int(float64(globalLimit) * float64(l.replicationFactor) * float64(zonesCount) / float64(ingestersInZone))
+
+	if maxStreamsPerUser > 0 && (newLimit == 0 || maxStreamsPerUser < newLimit) {
```
vlad-diachenko, line 132: we just need to return new limit here

pkg/ingester/limiter.go
vlad-diachenko, line 110: the goal of this function is to convert global limit to local to compare it after this with local limit on line 92

vlad-diachenko (approved):
do not pass local limit into `convertGlobalToLocalLimit` function

pkg/ingester/limiter.go
```diff
@@ -106,22 +108,31 @@ func (l *Limiter) minNonZero(first, second int) int {
...
 	if numIngesters > 0 {
 		return int((float64(globalLimit) / float64(numIngesters)) * float64(l.replicationFactor))
 	}
+	return 0
+}
 
+func calculateLimitForMultipleZones(globalLimit, zonesCount int, l *Limiter) int {
+	ingestersInZone := l.ring.HealthyInstancesInZoneCount()
+	if ingestersInZone > 0 {
+		return int(float64(globalLimit) * float64(l.replicationFactor) * float64(zonesCount) / float64(ingestersInZone))
```
vlad-diachenko, line 134:
```suggestion
		return int(float64(globalLimit) * float64(l.replicationFactor) / float64(zonesCount) / float64(ingestersInZone))
```
technically, I expect zones count to be equal to the replication factor, so we could simplify the expression to :
`int(float64(globalLimit) / float64(ingestersInZone))` because every zone receives exactly one replica of the data.

vlad-diachenko (approved):
please apply minor fix in the comment

PR #13329, opened 2024-06-26
perf: Prune unused pattern branches

benclive:
Adds a Prune command to Drain to clear branches with no associated log clusters.

* We use an LRU cache to limit the number of patterns we match against
* But when we delete a pattern due to age, we don't delete it's corresponding branches in the drain tree.
* This PR adds a step to traverse the tree and delete any nodes that don't have children or log clusters attached.
* It is run periodically, at the same time as we run the current logcluster prune process but it could probably be reduced in frequency.


cyriltovena (approved):
LGTM

PR #12176, opened 2024-03-12
fix(sharding): use without() grouping when merging `avg_over_time` shard results

ashwanthgoli:
**What this PR does / why we need it**:
`avg_over_time` expression that uses either `keep` or `label_format` stage without any aggregation fails on main with the following error: `aggregation operator '"sum"' without grouping`

Given these expressions contain stages that could reduce the labelset, sharding converts it to the following form `sum (sum_over_time()) / sum(count_over_time())`, but vector aggregation expr evaluator always expects a grouping. This pr updates the `avg_over_time` shard mapper to use `without()` grouping for generating vector agg expr when it is not explicitly set.

example query to reproduce this error:
```
avg_over_time({container="query-frontend"} |= `metrics.go` | logfmt | keep splits | unwrap split [$__auto])
```

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Special notes for your reviewer**:

**Checklist**
- [ ] Reviewed the [`CONTRIBUTING.md`](https://github.com/grafana/loki/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] Documentation added
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated
  - [ ] If the change is worth mentioning in the release notes, add `add-to-release-notes` label
- [ ] Changes that require user attention or interaction to upgrade are documented in `docs/sources/setup/upgrade/_index.md`
- [ ] For Helm chart changes bump the Helm chart version in `production/helm/loki/Chart.yaml` and update `production/helm/loki/CHANGELOG.md` and `production/helm/loki/README.md`. [Example PR](https://github.com/grafana/loki/commit/d10549e3ece02120974929894ee333d07755d213)
- [ ] If the change is deprecating or removing a configuration option, update the `deprecated-config.yaml` and `deleted-config.yaml` files respectively in the `tools/deprecated-config-checker` directory. [Example PR](https://github.com/grafana/loki/pull/10840/commits/0d4416a4b03739583349934b96f272fb4f685d15)

Issue #13287, opened 2024-06-21
avg_over_time might trigger `aggregation operator '"sum"' without grouping` error at specified intervals


nickgrafana:
**Describe the bug**
Running `avg_over_time` query fails when run against certain intervals of time
Other functions `_over_time` seem to behave properly without failing.
The job _does not_ need to exist, _no data is needed_ to reproduce the error. The problem is independent of the producer, collector or log format.

**To Reproduce**
1. Run `avg_over_time({job="nojob/nojob"} | json | keep http_route, response_time | unwrap response_time [$__auto])`
2. Alternative query from @slim-bean  that exhibits the same behaviour 
```avg by (http_route) (avg_over_time({job="pi-logs-custom3/pi-logs-custom3"} |json | keep http_route, response_time | unwrap response_time [$__auto]))```
3. Use Grafana UI and set a range of time last 5 min, get empty (or values)
```
{
  "queries": [
    {
      "refId": "A",
      "expr": "avg_over_time({job=\"pi-logs-custom3/pi-logs-custom3\"} |json | keep http_route, response_time | unwrap response_time [$__auto])",
      "queryType": "range",
      "datasource": {
        "type": "loki",
        "uid": "grafanacloud-logs"
      },
      "editorMode": "code",
      "legendFormat": "",
      "datasourceId": 7,
      "intervalMs": 200,
      "maxDataPoints": 1343
    }
  ],
  "from": "1719009561800",
  "to": "1719009861802"
}
```
4. Use Grafana UI and set a range of time last 6 hours, get 500 error `aggregation operator '"sum"' without grouping`
Payload:
```
{
  "queries": [
    {
      "refId": "A",
      "expr": "avg_over_time({job=\"pi-logs-custom3/pi-logs-custom3\"} |json | keep http_route, response_time | unwrap response_time [$__auto])",
      "queryType": "range",
      "datasource": {
        "type": "loki",
        "uid": "grafanacloud-logs"
      },
      "editorMode": "code",
      "legendFormat": "",
      "datasourceId": 7,
      "intervalMs": 15000,
      "maxDataPoints": 1343
    }
  ],
  "from": "1718988150000",
  "to": "1719009750941"
}
```
```
{
    "results": {
        "A": {
            "error": "aggregation operator '\"sum\"' without grouping",
            "errorSource": "downstream",
            "status": 500
        }
    }
}
```

**Expected behavior**
I expect values being returned, or empty if no data is found

**Environment:**
 - Grafana Cloud
[...]

PR #13357, opened 2024-07-01
fix: avg function result int unexpected response

yincongcyincong:

Fixes #12176 #13287
avg_over_time function don't contain default **group** , then an error output.

error code line: 
https://github.com/grafana/loki/blob/main/pkg/logql/evaluator.go#L390


cyriltovena:
cc @ashwanthgoli

ashwanthgoli:
@yincongcyincong thanks for the contribution ❤️ really appreciate it.
But I am closing this in the favour of an earlier pr https://github.com/grafana/loki/pull/12176 that fixes the same issue.

